PR #466, opened 2014-09-01
Added feature to allow user to have a wildcard for selecting "Any Pipeli...

lcs777:
...ne" or "Any Stage" within a pipeline when subscribing to email notifications.

This pull request is to let a user select to get notifications for any event on all pipelines or all stages on a particular pipeline.

This is useful because it allows users to be able to be notified on events for a pipeline when it changes without having to update their notifications everytime a pipeline changes.


jeremymv2:
Is there a way that we could add "Any Pipeline" and "Stage XYZ" ?  I have many many pipelines that all use the same template and I have a particular Stage that I want to be notified on.  It would be really nice if I could specify a rule that allowed my to match any pipeline but a specific stage, like "QA-Passed".

Issue #371, opened 2014-07-24
New dashboard JSON API


arvindsv:
This is a placeholder for the commits to new dashboard API (so that it gets an ID).

PR #663, opened 2014-10-15
#371 - Fix dashboard JSON - can administer and locking

arvindsv:
1. The ability of a user to administer a pipeline is indicated in the JSON now.
2. The pipeline settings path is now at the pipeline level, rather than the instance level.
3. Pipeline unlock operation is now shown at the appropriate times (pipeline is locked, and can be unlocked).

The top part of the JSON looks like this:

```
[
  {
    "name": "pipeline_group_name",

    "pipelines": [
      {
        "name": "pipeline_name",
        "can_administer": true,
        "settings_path": "/go/admin/pipelines/pipeline_name/general",
        "is_locked": true,

        "instances": [
              ...
```

The unlock operation looks like this:

```
   ...
        "available_operations": [
          {
            "operation": "unlock",
            "operation_path": "/go/api/pipelines/pipeline_name/releaseLock"
          },
          {
            "operation": "trigger",
            "operation_path": "/go/api/pipelines/pipeline_name/schedule"
          },
   ...
```


arikagoyal:
Verified on 14.4.0(1412-0838e5ed5b74c1)

PR #669, opened 2014-10-18
More appropriate runlevel start and stop priority in chkconfig init scri...

iliasbartolini:
To my understanding more appropriate runlevel for starting the service is 3 when most network services start.

The other change is related to stop priority. From the results of my searching the stop priority is decreasing. Low numbers stops earlier.

I haven't been able to test that those changes correctly generate the right init script after the package is deployed. I currenlty don't know how to test this process.


jyotisingh:
Works as expected during fresh install and upgrades. Verified on 15.1.0-1643.

Issue #234, opened 2014-06-04
Upgrade to TFS SDK 14.0.3


inforithmics:
To connect to repository [company].visualstudio.com/DefaultCollection the TFS sdk needs to be updated to version 11 Version 10.1 cannot connect to the hosted tfs. Or can I replace the sdk version myself?

Latest available: http://www.microsoft.com/en-in/download/details.aspx?id=40785

The Error message is 

```
The SOAP endpoint https://[company].visualstudio.com/DefaultCollection/Services/v1.0/Registration.asmx could not be contacted. HTTP status: 302
```

joshburt:
This is a big deal for our company.  We are in a hybrid configuration on and off-premise.  The need to integrate with VisualStudio online could make or break our adoption of Go.

mikecg:
I was considering using GO at my company, and just spent a few hours troubleshooting this error to no avail, but now that I have seen this issue, I have no choice but to abandon GO. This is very disappointing.

PR #690, opened 2014-10-31
Partial fix for Tfs sdk 12.0.2

mlapping:
Partial fix for #234. Seems to be working with visualstudio.com with alternate credentials, although I do not have a working build task to try it out on yet. It gets the revision id and commit message in any case.

This code does not use the domain variable. That can added easily by prepending "DOMAIN\" to the username in UsernamePasswordCredentials, according to the javadocs shipped with the SDK. The [Jenkins TFS plugin](https://github.com/jenkinsci/tfs-plugin/blob/master/src/main/java/hudson/plugins/tfs/model/Server.java#L39) supports NTLM authentication by explicitly using DefaultNTCredentials, which has been deprecated in favor of using UsernamePasswordCredentials with the username formatted as "DOMAIN\username" or "username@domain".

Also, no work has been done to support unique workspaces mentioned by @Christopher1065.


arvindsv:
I'm closing this in favor of https://github.com/gocd/gocd/pull/886. Even that one needs to be changed, in my opinion, to make it toggle-able. It's too big a risk to change it in one release. We have had memory issues with TFS before, and have had to make an emergency release.

PR #700, opened 2014-11-10
Make agent init script able to launch multiple instances 

bernardn:


arvindsv:
Probably a question to @arikagoyal. When this change gets in, because of the change to installers/agent/release/default.cruise-agent, an agent upgrade will ask a question to the user who is upgrading it. It looks like this:

image

Is this acceptable? I feel it is because of two reasons:

1. An agent upgrade almost never happens, since it is automatically done by the Go Server and Agent bootstrapper.

2. It is possible to provide an option to prevent this prompt. It'll look like this, then:

image

I can accept this change, if you agree. With apologies to bernardn for me forgetting about this and getting caught up with other things.

arvindsv:
Talked about this briefly with @arikagoyal and @jyotisingh. I think this can be merged, since there is an option to make it choose old or new configs, and the fact that this is not an operation done often (agent upgrade using DEB and RPM). During an RPM upgrade, this happens:

screen shot 2015-02-20 at 1 29 18 pm

RPM has chosen the "use old config" option by default, and has written the new config to a .rpmsave file.

Both these behaviors are standard for DEB and RPM. So, I am accepting this.

arikagoyal:
Verified on  15.1.0(1643-d2aa1fdc52e09e)

PR #768, opened 2014-12-12
form to upload a plugin jar in plugin tab

pwen:
This is the initial commit for "enable plugin install/uninstall on the UI" feature. Work in Progress. Please feel free to give us feedback along the way.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/admin/plugins/plugins_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/admin/plugins/plugins_controller.rb
@@ -26,16 +26,21 @@ class Admin::Plugins::PluginsController < AdminController
   end
 
   def upload
-    @upload_response = default_plugin_manager.addPlugin(java.io.File.new(params[:plugin].path), params[:plugin].original_filename)
-
-    respond_to do |format|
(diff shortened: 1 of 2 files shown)


pwen:
@arvindsv Hey! @pamo and I are having our second hack night, and are excited to contribute to this. 

What do you think should be the next thing in this capability that we should work on? We were thinking about the restart mechanism. If that's the case, we'd like some guidance on this. What are your thoughts?

arvindsv:
Hi @pwen and @pamo. Very cool. Here are some thoughts:

I'd not look at the restart mechanism. I don't think the restart mechanism should be strictly necessary. It should be possible to upload a plugin and load it, without having to restart the server. As I might have told you before, it was done to make sure that the Java perm gen space is properly freed after an older version of a plugin is removed (Java's GC does not garbage collect types, at least the older ones didn't). So, I'd look at the ability to run verifications on a plugin, without placing it on the filesystem, and rejecting the upload. If that doesn't seem interesting enough :), I'd consider reloading plugins on the agent side. Right now, plugins get reloaded only on the server side.

If those two don't interest you, then definitely look at the restart mechanism (maybe you've already started looking at it). Let me know if you need any help.

Issue #774, opened 2014-12-12
Support Hg post commit hook


arikagoyal:
Implement a post commit hook for Hg. 
Implementation:
Have a look at SvnPostCommitHookImplementer.java to get an idea of how is this implemented for Svn.
Documentation for svn and git post commit hooks -> http://www.go.cd/documentation/user/current/api/materials_api.html#notification-api

arikagoyal:
Works fine on 15.1.0(1445-488540650a0611)

PR #781, opened 2014-12-13
Post-commit hook for mercurial

sachinsudheendra:
Follows along the same line as git. Moved GitUrlMatchers to be generic enough to be used by mercurial as well.


arikagoyal:
This closes #774

PR #784, opened 2014-12-15
WIP - demonstration of JSON over HTTP agent

hammerdr:
This is a work in progress. I'm submitting the PR to kick off a discussion before I go too far down this road.

My goal is to have the go-agent talk JSON over HTTP instead of Java Serialization over HTTP (basically JMI). This would enable us to
* Have a far more flexible and decouple architecture for the agent
* Encourage different implementations of a go agent (including building in a system level language like golang so we need no dependencies on our hosts and can start them super-fast without needed to boot a JVM)

These are big wins and this is the groundwork for that.


zabil:
Closing this as it's been dormant, please feel free to open this when there's work.

Issue #790, opened 2014-12-15
stdrerr in not always properly redirected to a log file


deedrah:
When server or agent are forking to background using nohup, only stdout is always redirected to a log file. Stderr is dependent on whether it is a terminal or not. If it is not, nohup leaves stderr as is.

This cause a problem for me when managing go.cd instances with saltstack, because saltstack waits for an init script to close both stdout and stderr, which never happens.

Forking to background command should look more like this:
eval "nohup $CMD >>$LOG_FILE **2>&1** &"

arvindsv:
Fixed through #792. Closing issue.

PR #792, opened 2014-12-16
fix for #790 with stderr redirect

roastiek:
CLA should be signed


jyotisingh:
Verified on 15.1.0-1643.

PR #802, opened 2014-12-22
Fixed issue with the way query params were handled in agents page

jyotisingh:
default_url_options was overridden in agents_controller causing any redirects from this page to contain query params for column and order. This lead to a blank page being displayed when navigating to user summary page from here.
All get and post paths used in this page have been fixed to pass along the necessary params


arikagoyal:
Works fine on 15.1.0(1480-4840f33d6c6129)

PR #803, opened 2014-12-22
#80 allow dots in package repo paths

### LINKED ISSUE #80: Routing of get plugin config fails if plugin id contains a dot
## Description

Plugin framework defaults the file name as the plugin-id if not configured by plugin-dev.

If the plugin xml contains a '.' (dot) in plugin id or if the plugin xml is not providedat all, the plugin frameworkloads the plugin correctly.

However, the dropdown in package-repo edit/new page to fetch config for the given plugin fails to route correctly. MDU works correctly, its just an issue with the clicky admin UI. The constraint for the route needs to updated with the allowable plugin-id characters.

Note: The same behaviour is observed when plugin.xml is not specified and the filename of the jar has othercharacters like '*' etc (whatever is allowed in filename for a given OS) but not allowed by the xsd.

arvindsv:
Fixes #80.


arikagoyal:
Works fine on 14.4.0(1412-0838e5ed5b74c1)

PR #809, opened 2014-12-23
#723 - Change message in mail config UI about 'Use TLS' to 'Use SMTPS'

### LINKED ISSUE #723: Rename "Use TLS" to "Use SMTP over SSL"
The email notification option "Use TLS"  sets `mail.transport.protocol=smtps` which is not true TLS support and is based on a [deprecated specification](http://en.wikipedia.org/wiki/SMTPS). If the functionality to support SMTPS is not removed when resolving #724, the string should be changed to correctly reflect the implemented behavior.

arvindsv:
This resolves #723. There is a tooltip which points to documentation about turning on TLS. Documentation change coming up in its repo.


arvindsv:
Documentation is updated as well: https://github.com/gocd/documentation/pull/85

PR #812, opened 2014-12-26
Removed redundant array variable declaration.

akiellor:


arvindsv:
Makes sense. I can accept it, but it looks like you haven't signed the CLA yet. Can you please [do that](http://www.go.cd/contribute/cla.html)? It's a legal requirement.

srinivasupadhya:
@akiellor thanks for signing :+1:

PR #814, opened 2014-12-30
scm extension point - {1,2,3,4,5}/7

srinivasupadhya:
This PR takes care of 1, 2, 3, 4 & 5 in #818. VSM material node display (image for pluggable SCM) needs to be taken care.

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/AgentController.java
+++ b/agent/src/com/thoughtworks/go/agent/AgentController.java
@@ -20,7 +20,9 @@ import com.thoughtworks.go.agent.service.AgentUpgradeService;
 import com.thoughtworks.go.agent.service.SslInfrastructureService;
 import com.thoughtworks.go.config.AgentRegistry;
 import com.thoughtworks.go.domain.exception.UnregisteredAgentException;
+import com.thoughtworks.go.plugin.access.packagematerial.PackageAsRepositoryExtension;
 import com.thoughtworks.go.plugin.access.pluggabletask.TaskExtension;
+import com.thoughtworks.go.plugin.access.scm.SCMExtension;
 import com.thoughtworks.go.plugin.infra.PluginManager;
 import com.thoughtworks.go.plugin.infra.PluginManagerReference;
 import com.thoughtworks.go.publishers.GoArtifactsManipulator;
@@ -62,13 +64,17 @@ public class AgentController {
     private SubprocessLogger subprocessLogger;
     private final SystemEnvironment systemEnvironment;
     private AgentUpgradeService agentUpgradeService;
+    private PackageAsRepositoryExtension packageAsRepositoryExtension;
+    private SCMExtension scmExtension;
     private TaskExtension taskExtension;
 
     @Autowired
     public AgentController(BuildRepositoryRemote server, GoArtifactsManipulator manipulator, SslInfrastructureService sslInfrastructureService, AgentRegistry agentRegistry,
                            AgentUpgradeService agentUpgradeService, SubprocessLogger subprocessLogger, SystemEnvironment systemEnvironment,
-                           PluginManager pluginManager, TaskExtension taskExtension) {
+                           PluginManager pluginManager, PackageAsRepositoryExtension packageAsRepositoryExtension, SCMExtension scmExtension, TaskExtension taskExtension) {
         this.agentUpgradeService = agentUpgradeService;
+        this.packageAsRepositoryExtension = packageAsRepositoryExtension;
+        this.scmExtension = scmExtension;
         this.taskExtension = taskExtension;
         ipAddress = SystemUtil.getFirstLocalNonLoopbackIpAddress();
         hostName = SystemUtil.getLocalhostNameOrRandomNameIfNotFound();
@@ -177,7 +183,7 @@ public class AgentController {
                 }
             }
             runner = new JobRunner();
-            runner.run(work, agentIdentifier, server, manipulator, agentRuntimeInfo, taskExtension);
+            runner.run(work, agentIdentifier, server, manipulator, agentRuntimeInfo, packageAsRepositoryExtension, scmExtension, taskExtension);
         } catch (UnregisteredAgentException e) {
             LOG.warn(String.format("[Agent Loop] Invalid agent certificate with fingerprint %s. Registering with server on next iteration.", e.getUuid()));
             sslInfrastructureService.invalidateAgentCertificate();
--- a/agent/src/com/thoughtworks/go/agent/JobRunner.java
+++ b/agent/src/com/thoughtworks/go/agent/JobRunner.java
@@ -16,7 +16,9 @@
 
 package com.thoughtworks.go.agent;
 
+import com.thoughtworks.go.plugin.access.packagematerial.PackageAsRepositoryExtension;
 import com.thoughtworks.go.plugin.access.pluggabletask.TaskExtension;
+import com.thoughtworks.go.plugin.access.scm.SCMExtension;
 import com.thoughtworks.go.publishers.GoArtifactsManipulator;
 import com.thoughtworks.go.remote.AgentIdentifier;
 import com.thoughtworks.go.remote.AgentInstruction;
@@ -45,10 +47,10 @@ public class JobRunner {
         handled = true;
     }
 
-    public void run(Work work, AgentIdentifier agentIdentifier, BuildRepositoryRemote server,
-                    GoArtifactsManipulator manipulator, AgentRuntimeInfo agentRuntimeInfo, TaskExtension taskExtension) {
+    public void run(Work work, AgentIdentifier agentIdentifier, BuildRepositoryRemote server, GoArtifactsManipulator manipulator, AgentRuntimeInfo agentRuntimeInfo,
+                    PackageAsRepositoryExtension packageAsRepositoryExtension, SCMExtension scmExtension, TaskExtension taskExtension) {
         this.work = work;
-        work.doWork(agentIdentifier, server, manipulator, environmentVariableContext, agentRuntimeInfo, taskExtension);
+        work.doWork(agentIdentifier, server, manipulator, environmentVariableContext, agentRuntimeInfo, packageAsRepositoryExtension, scmExtension, taskExtension);
     }
 
     public boolean isJobCancelled() {
--- a/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
@@ -296,9 +296,9 @@ public class PluggableSCMMaterial extends AbstractMaterial {
 
         PluggableSCMMaterial that = (PluggableSCMMaterial) o;
 
-        if (filter != null ? !filter.equals(that.filter) : that.filter != null) return false;
-        if (folder != null ? !folder.equals(that.folder) : that.folder != null) return false;
-        if (scmId != null ? !scmId.equals(that.scmId) : that.scmId != null) return false;
+        if (this.getFingerprint() != null ? !this.getFingerprint().equals(that.getFingerprint()) : that.getFingerprint() != null) {
+            return false;
+        }
 
         return true;
     }
@@ -306,9 +306,7 @@ public class PluggableSCMMaterial extends AbstractMaterial {
     @Override
     public int hashCode() {
         int result = super.hashCode();
-        result = 31 * result + (scmId != null ? scmId.hashCode() : 0);
-        result = 31 * result + (folder != null ? folder.hashCode() : 0);
-        result = 31 * result + (filter != null ? filter.hashCode() : 0);
+        result = 31 * result + (this.getFingerprint() != null ? this.getFingerprint().hashCode() : 0);
         return result;
     }
 
--- a/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
@@ -283,20 +283,15 @@ public class PluggableSCMMaterial extends AbstractMaterial {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
-
-        // skipping super.equals()
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
 
         PluggableSCMMaterial that = (PluggableSCMMaterial) o;
 
-        if (this.getFingerprint() != null ? !this.getFingerprint().equals(that.getFingerprint()) : that.getFingerprint() != null) {
-            return false;
-        }
+        if (filter != null ? !filter.equals(that.filter) : that.filter != null) return false;
+        if (folder != null ? !folder.equals(that.folder) : that.folder != null) return false;
+        if (scmId != null ? !scmId.equals(that.scmId) : that.scmId != null) return false;
 
         return true;
     }
@@ -305,6 +300,8 @@ public class PluggableSCMMaterial extends AbstractMaterial {
     public int hashCode() {
         int result = super.hashCode();
         result = 31 * result + (scmId != null ? scmId.hashCode() : 0);
+        result = 31 * result + (folder != null ? folder.hashCode() : 0);
+        result = 31 * result + (filter != null ? filter.hashCode() : 0);
         return result;
     }
 
--- a/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
@@ -110,6 +110,13 @@ public class PluggableSCMMaterial extends AbstractMaterial {
         this.folder = folder;
     }
 
+    public File workingDirectory(File baseFolder) {
+        if (folder == null) {
+            return baseFolder;
+        }
+        return new File(baseFolder, folder);
+    }
+
     public Filter getFilter() {
         return filter;
     }
--- a/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
@@ -1,0 +1,315 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2014 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END***********************************/
+
+package com.thoughtworks.go.config.materials;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import com.thoughtworks.go.config.CaseInsensitiveString;
+import com.thoughtworks.go.config.PipelineConfig;
+import com.thoughtworks.go.domain.MaterialInstance;
+import com.thoughtworks.go.domain.MaterialRevision;
+import com.thoughtworks.go.domain.config.ConfigurationProperty;
+import com.thoughtworks.go.domain.materials.*;
+import com.thoughtworks.go.domain.materials.scm.PluggableSCMMaterialInstance;
+import com.thoughtworks.go.domain.materials.scm.PluggableSCMMaterialRevision;
+import com.thoughtworks.go.domain.scm.SCM;
+import com.thoughtworks.go.util.StringUtil;
+import com.thoughtworks.go.util.command.EnvironmentVariableContext;
+import com.thoughtworks.go.util.command.ProcessOutputStreamConsumer;
+import com.thoughtworks.go.util.json.JsonHelper;
+import com.thoughtworks.go.util.json.JsonMap;
+
+import java.io.File;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static com.thoughtworks.go.util.command.EnvironmentVariableContext.escapeEnvironmentVariable;
+import static java.lang.String.format;
+import static org.apache.commons.lang.StringUtils.isEmpty;
+import static org.apache.commons.lang.StringUtils.upperCase;
+
+public class PluggableSCMMaterial extends AbstractMaterial {
+    public static final String TYPE = "PluggableSCMMaterial";
+
+    private String scmId;
+
+    @Expose
+    @SerializedName("scm")
+    private SCM scmConfig;
+
+    private String folder;
+
+    private Filter filter;
+
+    private String fingerprint;
+
+    public PluggableSCMMaterial() {
+        super(TYPE);
+    }
+
+    public PluggableSCMMaterial(String scmId) {
+        this();
+        this.scmId = scmId;
+    }
+
+    public PluggableSCMMaterial(PluggableSCMMaterialConfig config) {
+        this();
+        this.name = config.getName();
+        this.scmId = config.getSCMId();
+        this.scmConfig = config.getSCMConfig();
+        this.folder = config.getFolder();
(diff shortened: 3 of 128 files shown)


srinivasupadhya:
Use [this](https://github.com/srinivasupadhya/git-poller) sample plugin for testing end-poing.

agent/src/com/thoughtworks/go/agent/AgentController.java
```diff
@@ -62,13 +64,17 @@
...
     private AgentUpgradeService agentUpgradeService;
+    private PackageAsRepositoryExtension packageAsRepositoryExtension;
+    private SCMExtension scmExtension;
     private TaskExtension taskExtension;
 
     @Autowired
     public AgentController(BuildRepositoryRemote server, GoArtifactsManipulator manipulator, SslInfrastructureService sslInfrastructureService, AgentRegistry agentRegistry,
                            AgentUpgradeService agentUpgradeService, SubprocessLogger subprocessLogger, SystemEnvironment systemEnvironment,
-                           PluginManager pluginManager, TaskExtension taskExtension) {
+                           PluginManager pluginManager, PackageAsRepositoryExtension packageAsRepositoryExtension, SCMExtension scmExtension, TaskExtension taskExtension) {
```
arvindsv: Should we have ```Extensions``` or something? I don't see the need to change this every time we add a new kind of extension. Could use ```extensions.forTask()``` and ```extensions.forPackageRepository()```, etc.

arvindsv: Anyway, all that is being done here is that it is passed down to the runner.

agent/src/com/thoughtworks/go/agent/JobRunner.java
```diff
@@ -45,10 +47,10 @@ private void cancelJob(AgentRuntimeInfo agentRuntimeInfo) {
         handled = true;
     }
 
-    public void run(Work work, AgentIdentifier agentIdentifier, BuildRepositoryRemote server,
-                    GoArtifactsManipulator manipulator, AgentRuntimeInfo agentRuntimeInfo, TaskExtension taskExtension) {
+    public void run(Work work, AgentIdentifier agentIdentifier, BuildRepositoryRemote server, GoArtifactsManipulator manipulator, AgentRuntimeInfo agentRuntimeInfo,
+                    PackageAsRepositoryExtension packageAsRepositoryExtension, SCMExtension scmExtension, TaskExtension taskExtension) {
```
arvindsv: Same. Extensions class would help here as well.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+import static org.apache.commons.lang.StringUtils.upperCase;
+
+public class PluggableSCMMaterial extends AbstractMaterial {
+    public static final String TYPE = "PluggableSCMMaterial";
+
+    private String scmId;
+
+    @Expose
+    @SerializedName("scm")
+    private SCM scmConfig;
```
arvindsv: Can you think of a more descriptive name than SCM? Maybe PluggedSCMConfig or even SCMConfig?

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+
+    public void setFolder(String folder) {
+        this.folder = folder;
+    }
+
+    public Filter getFilter() {
+        return filter;
+    }
+
+    public void setFilter(Filter filter) {
```
arvindsv: Why are these methods needed? Ibatis/Hibernate?

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+        this.filter = filter;
+    }
+
+    public String getPluginId() {
+        return getScmConfig().getPluginConfiguration().getId();
```
arvindsv: Can getScmConfig() not be null here? I ask because you're checking for it just below in getFingerprint().

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static com.thoughtworks.go.util.command.EnvironmentVariableContext.escapeEnvironmentVariable;
+import static java.lang.String.format;
+import static org.apache.commons.lang.StringUtils.isEmpty;
+import static org.apache.commons.lang.StringUtils.upperCase;
+
+public class PluggableSCMMaterial extends AbstractMaterial {
```
arvindsv: What is the reason for it being an AbstractMaterial and not an SCMMaterial?

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        // skipping super.equals()
```
arvindsv: Why?

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -0,0 +1,315 @@
...
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = super.hashCode();
+        result = 31 * result + (scmId != null ? scmId.hashCode() : 0);
```
arvindsv: Very dangerous territory. It's almost always a mistake to keep equals() and hashCode() different like this. I know it's done the same way in PackageMaterial, but this should be reasoned about, and if confident that it cannot cause issues when this object is in a hash table, then we should comment on top of this method and the corresponding one in PackageMaterial.

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
```diff
@@ -0,0 +1,76 @@
...
+        return scmPropertyConfiguration;
+    }
+
+    private void populateConfiguration(Configuration configuration, com.thoughtworks.go.plugin.api.config.Configuration pluginConfiguration) {
+        for (ConfigurationProperty configurationProperty : configuration) {
+            pluginConfiguration.add(new SCMProperty(configurationProperty.getConfigurationKey().getName(), configurationProperty.getValue()));
+        }
+    }
+
+    private File workingDirectory(File baseFolder, String destinationFolder) {
```
arvindsv: Why not do it like with [ScmMaterial](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/MaterialAgentFactory.java#L49)? Meaning, this can be in the Material itself, rather than here.

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void prepare() {
+        PluggableSCMMaterial material = (PluggableSCMMaterial) revision.getMaterial();
+        Modification latestModification = revision.getLatestModification();
+        SCMRevision scmRevision = new SCMRevision(latestModification.getRevision(), latestModification.getModifiedTime(), null, null, latestModification.getAdditionalDataMap(), null);
+        File destinationFolder = workingDirectory(workingDirectory, material.getFolder());
+        Result result = scmExtension.checkout(material.getScmConfig().getPluginConfiguration().getId(), buildSCMPropertyConfigurations(material.getScmConfig()), destinationFolder.getAbsolutePath(), scmRevision);
+        if (!result.isSuccessful()) {
+            // handle
```
arvindsv: Hmm. Needs to be done, right?

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        // handle messages
+    }
+
+    private SCMPropertyConfiguration buildSCMPropertyConfigurations(SCM scmConfig) {
```
arvindsv: Shouldn't this be the responsibility of the ScmExtension class? Should probably ask @pshivana.

common/src/com/thoughtworks/go/remote/work/BuildWork.java
```diff
@@ -78,14 +80,13 @@ private void initialize(BuildRepositoryRemote remoteBuildRepository,
...
-                       GoArtifactsManipulator goArtifactsManipulator, EnvironmentVariableContext environmentVariableContext,
-                       AgentRuntimeInfo agentRuntimeInfo, TaskExtension taskExtension) {
+    public void doWork(AgentIdentifier agentIdentifier, BuildRepositoryRemote remoteBuildRepository, GoArtifactsManipulator goArtifactsManipulator,
+                       EnvironmentVariableContext environmentVariableContext, AgentRuntimeInfo agentRuntimeInfo,
+                       PackageAsRepositoryExtension packageAsRepositoryExtension, SCMExtension scmExtension, TaskExtension taskExtension) {
         initialize(remoteBuildRepository, goArtifactsManipulator, agentRuntimeInfo, taskExtension);
         environmentVariableContext.addAll(assignment.initialEnvironmentVariableContext());
         try {
-            JobResult result = build(environmentVariableContext, agentIdentifier);
+            JobResult result = build(environmentVariableContext, agentIdentifier, packageAsRepositoryExtension, scmExtension);
```
arvindsv: An Extensions class, again, will help.

config/config-api/src/com/thoughtworks/go/config/materials/MaterialConfigs.java
```diff
@@ -289,12 +318,15 @@ public void setConfigAttributes(Object attributes) {
...
+        } else if (PluggableSCMMaterialConfig.TYPE.equals(materialType)) {
+            addMaterialConfig(getSCMMaterial(), (Map) attributeMap.get(PluggableSCMMaterialConfig.TYPE));
         }
     }
 
     public boolean scmMaterialsHaveDestination() {
-        for (ScmMaterialConfig scmMaterial : filterScmMaterials()) {
-            if (!scmMaterial.hasDestination()) {
+        for (MaterialConfig scmMaterial : getSCMAndPluggableSCMConfigs()) {
+            String destination = (scmMaterial instanceof ScmMaterialConfig) ? ((ScmMaterialConfig) scmMaterial).getFolder() : ((PluggableSCMMaterialConfig) scmMaterial).getFolder();
```
arvindsv: The MaterialConfig interface has a getFolder() method. Why is this needed? The condition check, I mean.

config/config-api/src/com/thoughtworks/go/config/materials/MaterialConfigs.java
```diff
@@ -155,22 +149,43 @@ public void validate(ValidationContext validationContext) {
...
+        allSCMMaterials.addAll(scmMaterials);
+        allSCMMaterials.addAll(pluggableSCMMaterials);
+        return allSCMMaterials;
+    }
+
+    private void validateDestinationFolder(List<MaterialConfig> allSCMMaterials, MaterialConfig material) {
+        String materialFolder = material.getFolder();
+        for (MaterialConfig otherMaterial : allSCMMaterials) {
+            if (otherMaterial != material) {
+                if (otherMaterial instanceof ScmMaterialConfig) {
```
arvindsv: Can we put all of these if-else ladders somewhere else? I'm getting worried that there are too many places this is being done. Putting them in one place is not very different, but it's at least very visible where it is being used.

config/config-api/src/com/thoughtworks/go/config/materials/ScmMaterialConfig.java
```diff
@@ -238,29 +238,25 @@ public void setDestinationFolderError(String message) {
...
         }
         File myDir = new File(myDirPath);
-        File otherDir = new File(otherDirPath);
+        File otherDir = new File(otherSCMMaterialFolder);
         try {
             if (FileUtil.isSubdirectoryOf(myDir, otherDir)) {
-                String message = "Invalid Destination Directory.Every material needs a different destination directory and the directories should not be nested.";
-                this.setDestinationFolderError(message);
-                otherMaterial.setDestinationFolderError(message);
+                addError(FOLDER, "Invalid Destination Directory.Every material needs a different destination directory and the directories should not be nested.");
```
arvindsv: If you're making another commit, can you add a space between the two sentences?

agent/src/com/thoughtworks/go/agent/AgentController.java
srinivasupadhya: umm..yes, but I thought having separate args is easier to test interms of mocks etc. Im ok to make it extensions containing all extensions in it.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: SCMMaterials as things like `getURL`, `getUserName` etc. which cannot be answered by PluggableSCMMaterial.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: I dint want to use PluggedSCMConfig since Im already using PluggableSCMConfig in Pipeline -> Materials. I have kept it SCMs & SCM on same lines as PackageRepositories & PackageRepository. I dont mind changing it to SCMsConfig & SCMConfig.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: I thought they would be needed for tests.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: on same lines as [packagematerial](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/common/src/com/thoughtworks/go/config/materials/PackageMaterial.java#L91). I think packagematerial isn't valid until packgedefinition is set & hence the same assumption can be made? @pshivana what do you think?

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: because super.equals does
```java
        if (name != null ? !name.equals(that.name) : that.name != null) {
            return false;
        }
        if (type != null ? !type.equals(that.type) : that.type != null) {
            return false;
        }
```
type is checked by `getClass() != o.getClass()` while name will be same since its referring to same copy.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: fingerprint & scm-id will be same & can be assumed? not sure why different things were used in equals & hashcode. Im ok making them same. @pshivana?

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
srinivasupadhya: Yes. I need to check if this requires throwing an exception. Also the plugin needs to be initialised with JobConsole so the plugin could log out/err into job console.

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
srinivasupadhya: not all materials have destination directory. specially DependencyMaterial.

config/config-api/src/com/thoughtworks/go/config/materials/MaterialConfigs.java
srinivasupadhya: Since this is the last change for material Im ok to leave it like this.

config/config-api/src/com/thoughtworks/go/config/materials/MaterialConfigs.java
srinivasupadhya: Ok. I assumed getFolder will be present only in SCMMaterial since the folder support was only in SCM materials. I will change this.

config/config-api/src/com/thoughtworks/go/config/materials/ScmMaterialConfig.java
srinivasupadhya: okies.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
pshivana: getScmConfig() will not return null here. It was done so in getFingerprint() as part of fixing validations and fixing specs

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
pshivana: I agree hashCode() and equals() should be done same way, it should be changed for PackageMaterial as well

common/src/com/thoughtworks/go/domain/materials/scm/PluggableSCMMaterialAgent.java
pshivana: Yes....plugin api classes should not be leaked beyond plugin access

config/config-api/src/com/thoughtworks/go/domain/scm/SCMs.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public void removeSCM(String id) {
+        SCM scmToBeDeleted = this.find(id);
+        if (scmToBeDeleted == null) {
+            throw new RuntimeException(String.format("Could not find SCM with id '%s'", id));
+        }
+        this.remove(scmToBeDeleted);
+    }
+
+    private void validateNameUniqueness() {
```
arvindsv: Do you really need this, since the cruise-config.xsd (below) has a name uniqueness check?

srinivasupadhya:
have added ability to get view template, stored in metadata-store (will be used when UI is built). fixed equals & hashcode, workingDirectory(), folder() & the space in error message.

common/src/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
srinivasupadhya: fixed.

arvindsv:
Sigh. You're not helping this get reviewed, by adding more commits. Especially ones like [this](https://github.com/srinivasupadhya/gocd/commit/e2cfe6a3ad20b9a4da93178e2d9bdbde8e8af696).

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/scm/SCMConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+import com.thoughtworks.go.plugin.api.config.Options;
+import com.thoughtworks.go.plugin.api.config.Property;
+
+public class SCMConfiguration implements Comparable {
+
+    public static final Option<Boolean> REQUIRED = Property.REQUIRED;
+    public static final Option<Boolean> PART_OF_IDENTITY = Property.PART_OF_IDENTITY;
+    public static final Option<Boolean> SECURE = Property.SECURE;
+    public static final Option<String> DISPLAY_NAME = Property.DISPLAY_NAME;
+    public static final Option<Integer> DISPLAY_ORDER = Property.DISPLAY_ORDER;
```
arvindsv: Is the DISPLAY_ORDER needed when the view is coming from a template?

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public void insertLatestOrNewModifications(final Material material, MaterialInstance materialInstance, File folder, Modifications list) {
+        final PluggableSCMMaterialInstance currentMaterialInstance = (PluggableSCMMaterialInstance) materialInstance;
+
+        final PluggableSCMMaterialInstance latestMaterialInstance = (PluggableSCMMaterialInstance) material.createMaterialInstance();
+        if (currentMaterialInstance.shouldUpgradeTo(latestMaterialInstance)) {
+            transactionTemplate.execute(new TransactionCallback() {
+                @Override
+                public Object doInTransaction(TransactionStatus transactionStatus) {
+                    PluggableSCMMaterialInstance materialInstance = (PluggableSCMMaterialInstance) materialRepository.find(currentMaterialInstance.getId());
```
arvindsv: This should be moved into MaterialRepository. If any kind of caching considerations are applicable here (now or sometime later), then, no one would look at this usage, here. I mean, if the material instance is cached somewhere (usually in MaterialRepository), it'll need to be removed from cache. That belongs in MaterialRepository, not here.

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
```diff
@@ -0,0 +1,68 @@
...
+                    materialInstance.upgradeTo(latestMaterialInstance);
+                    materialRepository.saveOrUpdate(materialInstance);
+                    return materialInstance;
+                }
+            });
+        }
+        scmMaterialUpdater.insertLatestOrNewModifications(material, currentMaterialInstance, folder, list);
```
arvindsv: Once upgraded, is currentMaterialInstance the right one to be used? Is that because the actual object of currentMaterialInstance will be updated in line 55?

arvindsv:
Alright, I think I'm done. I have nothing else at this point. Only three more questions for you to answer!

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/scm/SCMConfiguration.java
srinivasupadhya: yes. `DISPLAY_NAME` & `DISPLAY_ORDER` will be used in material listing page to list properties & their values.

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
srinivasupadhya: on same lines as [this](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/server/src/com/thoughtworks/go/server/materials/PackageMaterialUpdater.java#L50). @pshivana what do you think?

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
srinivasupadhya: on same lines as [this](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/server/src/com/thoughtworks/go/server/materials/PackageMaterialUpdater.java#L50). @pshivana what do you think?

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
pshivana: Yes it can be moved to MaterialRepository

server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java
pshivana: I don't it matters which instance being used on line 61. It should work either way.

srinivasupadhya: ok will move both usages (packge-repository & pluggable-scm) to material repository in next PR.
also, will make equals & hashcode for packagematerial consistent in that PR.

srinivasupadhya:
ok, i am done too. thanks @arvindsv & @pshivana for the review.
@arvindsv sorry for adding new commit in middle of the review. it was required for UI & dint want to wait for new PR.

arvindsv:
> ok will in next PR:

> move both usages (packge-repository & pluggable-scm) to material repository
> will make equals & hashcode for packagematerial consistent
> make PackgeAsMaterialExtension, SCMExtension & TaskExtension into Extensions on the agent side.
> see if name uniqueness validity is required, both in packge-repository & pluggable-scm

This should be done too.

Issue #103, opened 2014-04-16
Agent -side plugin doesn't get deleted after the same plugin was
deleted on server side


mdaliejaz:
### Steps to reproduce
1.  plugin called test.jar on server side which is present on the agent side as well
2.  Delete this plugin from server side . Restart the server
3.  Now restart the agent and check the plugins folder on agent side

#### Expected Result
The deleted plugin should not be present on agent side

#### Current Result
The deleted plugin is still present

arikagoyal:
Works fine on 15.1.0(1434-8bfacdfeec4b10)

PR #819, opened 2015-01-02
#103 - Fix agent plugin cleanup

arvindsv:
This fixes the symptom, but not the real root issue. The symptom is that plugins removed on the server-side were not properly removed from the agent-side. The code in this PR fixes that.

However, there needs to be more intelligence in the way newer plugins are introduced to both the server and agent sides. Not just a full copy of all the plugins sent every time. That will remove the need for the agent to restart itself, when new plugins are installed.


arvindsv:
For verifying this, the 5 scenarios tested in the code are:

1. Bundled plugin replaced with a new one (plugin-1.jar replaced with plugin-1.jar with new content).
2. External plugin replaced with a new one (plugin-1.jar replaced with plugin-1.jar with new content).
3. Bundled plugin removed, and replaced with a different one (plugin-1.jar replaced with plugin-2.jar).
4. External plugin removed, and replaced with a different one (plugin-1.jar replaced with plugin-2.jar).
5. Plugin removed from the server side, should be removed from the agent side as well.

PR #824, opened 2015-01-06
#000 - Fix remove_licence.sh to exit on failure.

arvindsv:
It is used during installer generation, to remove the license headers.

Also, made it a little more robust. It depends on the GO-LICENSE-END
line, rather than a set number of lines.

Fixed a lot of incorrect file names too. Now, it fails if it is asked to remove license from wrong file names.


jyotisingh:
All looks ok, the script fails on failures now. 
Now that this actually fails the build on failure, we need to make one small fix in additional-files.txt :). 
Basically, "../../common/test-resources/data/customer/CoverageSummary.xml" needs to be changed to "../../common/test-resources/unit/data/customer/CoverageSummary.xml".

arvindsv:
Thank you! Fixed.

PR #840, opened 2015-01-16
Notification end point

srinivasupadhya:


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationExtension.java
```diff
@@ -0,0 +1,95 @@
...
+    private Map<String, JsonMessageHandler> messageHandlerMap = new HashMap<String, JsonMessageHandler>();
+
+    @Autowired
+    public NotificationExtension(PluginManager defaultPluginManager) {
+        this.pluginManager = defaultPluginManager;
+        this.pluginRequestHelper = new PluginRequestHelper(pluginManager, goSupportedVersions, EXTENSION_NAME);
+        this.messageHandlerMap.put("1.0", new JsonMessageHandler1_0());
+    }
+
+    public List<String> getNotificationsInterestedIn(String pluginId) {
```
jyotisingh: method name does not read very well, think about renaming it

server/src/com/thoughtworks/go/server/messaging/plugin/PluginNotificationListenerFactory.java
```diff
@@ -0,0 +1,49 @@
...
+        this.systemEnvironment = systemEnvironment;
+        this.notificationExtension = notificationExtension;
+        this.serverHealthService = serverHealthService;
+    }
+
+    public void init() {
+        int numberOfListeners = systemEnvironment.getNumberOfPluginNotificationListener();
+
+        for (int i = 0; i < numberOfListeners; i++) {
+            PluginNotificationService pluginNotificationService = new PluginNotificationService(notificationExtension, serverHealthService);
```
jyotisingh: PluginNotificationService is a component, why initialize it here again. You could pass it along through the autowired constructor of PluginNotificationListenerFactory.

srinivasupadhya: on similar lines as [this](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/materials/MaterialUpdateListenerFactory.java#L77). Need to check why it isn't autowired here as well.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationExtension.java
srinivasupadhya: ok. will make it `getNotificationsInterestedInFor(pluginId)`

jyotisingh: Does notificationsOfInterestFor(pluginId) sound about alright?

srinivasupadhya:
done as part of #856

PR #842, opened 2015-01-19
#103 - Fix an agent plugins test.

arvindsv:


arvindsv:
I don't think this will fix the failing test on https://build.go.cd/go/tab/build/detail/build-windows/258/build-non-server-windows/1/agent yet. Closing this, till I do some more analysis on it.

PR #847, opened 2015-01-21
Change init.d script to redirect the output of start_daemon calls.

### LINKED ISSUE #846: Init script - calls to start-daemon should be redirected
https://github.com/gocd/gocd/blob/master/installers/server/release/init.cruise-server#L71-73

Attempting to create a saltstack formula for installing go-server, and pkg install command hangs.  This is cause by the start_daemon call not redirecting output, and salt is waiting for the stdout file descriptor to close.

This also affects any calls from salt to start the service.

stevendgonzales:
Currently the absence of this redirection causes configuration
management tools such as Saltstack to hang on service start, as the
process waits infintely for the file descriptors to close.
Tested with init.d script on Ubuntu 14.04.

closes #846

### DIFF
--- a/installers/agent/release/init.cruise-agent
+++ b/installers/agent/release/init.cruise-agent
@@ -68,9 +68,9 @@ start_go_agent() {
 
     if [ "$JAVA_HOME" != "" ]; then
         if [ "${CUR_USER}" == "root" ]; then
-            start_daemon /bin/su - go -c /usr/share/go-agent/agent.sh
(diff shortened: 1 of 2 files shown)


arvindsv:
Hey @stevendgonzales, can you give me a way to reproduce this? I'm thinking of at least writing the output to a log file or something. Or, maybe redirect to /dev/null with an option? Do you think either of them would work better?

I ask because a user who does run /etc/init.d/go-server start will have no feedback, if this is accepted in its current state.

I think an environment variable as an option might make sense. Let me know if you think so too.

arvindsv:
Haven't heard back. Closing this.

PR #850, opened 2015-01-26
Add SERVER_LISTEN_HOST to /etc/default/go-server.

### LINKED ISSUE #705: Documentation: describe how to bind Go to localhost
We're told how to front-end Go with an Apache proxy, but not how to prevent it from binding to `*:8153` and `*:8154`. Should we edit `config.properties`, `jetty.xml`, or `cruise-config.xml`? How?

stevendgonzales:
Make it obvious how to configure the go-server bind host
and to tell what the default setting is.

Closes #705


arvindsv:
I'm closing this without merging it. I've added some [documentation](https://github.com/arvindsv/documentation/blob/add_options_page/user/advanced_usage/other_config_options.md) (it's a [PR](https://github.com/gocd/documentation/pull/96) still), instead. There are many such options, and deciding which ones to add and which ones not to, I think, is not going to be easy and fruitful.

I'm thinking about adding a pointer to this documentation, as a comment in the /etc/default/go-server file, but haven't done that.

Another reason not to accept this, is that, for every existing user, the upgrade will now stop and ask them to resolve a conflict, if this file is changed.

Fixed #705.

PR #856, opened 2015-01-29
notification end point - incorporate feedback

srinivasupadhya:


srinivasupadhya:
done as part of #856

srinivasupadhya:
done as part of #867

Issue #854, opened 2015-01-28
Allow '#' single hash in config


mdaliejaz:
## Current behaviour

If one tries to save config with a single hash (in param resolvable fields), save fails with the following validation error:
_"... # must be followed by a parameter pattern or escaped by another #"_

> '#{PARAM}' is used for using parameters. '##' is used to specify '#' in URLs.
## Why should the config allow single hash?

Single hash is used in URLs. To specify branches in mercurial the URL used is `<repository URL>#branch-name`. In Go this URL is specified as `<repository URL>##branch-name`.
This behaviour is consistent even if someone wants to curl an URL with # in it.

Go sanitises the URLs before performing operations on it. This leads to the bug #832. But this sanitisation of URLs go hand in hand with de-sanitisation for displaying purpose. Check [here](https://github.com/gocd/gocd/blob/master/config/config-api/src/com/thoughtworks/go/util/command/HgUrlArgument.java).
The current behaviour is consistent. However, to resolve the aforementioned bug, we would need to fix code to not sanitise the URL until mercurial post commit hook is realised by Go. Not something I would appreciate.

The [pipeline creation API](http://www.go.cd/documentation/user/current/api/configuration_api.html#adding-a-new-pipeline) also uses double hash for creating a pipeline with Hg material (with branch). It would be better if single hash is allowed in all the places.
## Expected Behaviour

Single hash should be allowed in the config.
If someone tries to use a parameter in a wrong way (#PARAM and not #{PARAM}), Go neither complains nor resolves such texts. It should be the user's responsibility to specify the parameters with the correct semantics.
### Step I

The current behaviour to sanitise the URLs with two hashes '##' to a single hash '#' before doing any operation continues, so that the existing users don't face any problem. The change is to make Go a little more liberal in this context.
### Step II

Perform config and DB migration to move all double hashes (existing ones) to single hash. The code to sanitise URLs can then be thrown off.

**Go Version:** 14.4.0 and previous

nikita-leonov:
There is another relevant example. We are using regexp in commands, and since single # is not allowed and require escaping whole process of writing, reading and understanding regexp is not really clear as they become not pure regexps.

mdaliejaz:
This would resolve issue #817 as well.

zakjan:
Any progress? It prevents writing Shell scripts as reported in https://github.com/srinivasupadhya/script-executor-task/issues/7

hanej:
We're affected by this too

zabil:
[...]

PR #857, opened 2015-01-29
#854 - Removed validation on Single Hash.

mdaliejaz:
* Validation doesn't fail for a single hash.
* Removed/Modified/Added test for new changes.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/preprocessor/ParamResolver.java
+++ b/config/config-api/src/com/thoughtworks/go/config/preprocessor/ParamResolver.java
@@ -16,20 +16,13 @@
 
 package com.thoughtworks.go.config.preprocessor;
 
+import com.thoughtworks.go.config.*;
+
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-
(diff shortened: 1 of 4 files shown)


mdaliejaz:
Personal went green.

arvindsv:
@mdaliejaz, I'm guessing you'll work with someone to get this accepted. You mentioned migration of config in #854. But, I don't see that happening here. So, with this change, the validation seems to be removed, right? So, # is allowed, as well as ##. I guess ## converts to #. However, for new users of this, the fact that ## converts to # is confusing and unexpected. Right?

Once the config migration is done, then all ## in config will be replaced, and this change is meaningful. What do you think? If you agree, then, should config migration be part of this?

mdaliejaz:
@arvindsv Sorry, I forgot to update here.
Yes, there is more work required. I intended to complete both the steps mentioned in #854, and they should go together in a release.
Since this wasn't going in 15.1, I parked the story for sometime due to other priorities for 15.1, but forgot to update it over here.
I can close this PR and send another PR with all the commits together on a later date, or this PR can be merged later when other commits are pushed :) Let me know what you think.

arvindsv:
Yes, I think closing it for now, makes sense.

mdaliejaz:
Cool. Closing it for now :)

Issue #833, opened 2015-01-12
move 'rake' task into plugin


srinivasupadhya:
Current XML:

``` xml
<rake buildfile="build-file" target="target" workingdir="working-directory">
  <runif status="any" />
  <oncancel>
    <rake buildfile="cancel-build-file" target="cancel-target" workingdir="cancel-working-directory" />
  </oncancel>
</rake>
```

After Migration:

``` xml
<task>
  <pluginConfiguration id="rake" version="1"/>
  <configuration>
    <property>
      <key>build_file</key>
      <value>build-file</value>
    </property>
    <property>
      <key>target</key>
      <value>target</value>
    </property>
    <property>
      <key>working_directory</key>
      <value>working-directory</value>
   </property>
  </configuration>
  <runif status="any"/>
  <oncancel>
    <task>
      <pluginConfiguration id="rake" version="1"/>
      <configuration>
        <property>
          <key>build_file</key>
          <value>cancel-build-file</value>
        </property>
        <property>
          <key>target</key>
          <value>cancel-target</value>
        </property>
        <property>
          <key>working_directory</key>
          <value>cancel-working-directory</value>
        </property>
      </configuration>
    </task>
  </oncancel>
</task>
```

arvindsv:
Hmm, seeing that makes me reconsider. Apart from some removal of code, and quite a big increase in config XML, what does this do? Is this really needed?

srinivasupadhya:
Hmm..I have created a [PR](https://github.com/gocd/go-plugins/pull/9) for new rake task plugin. I will send in the PR to delete built-in rake task soon. Have a look at both & let me know if its worth the change.

IMO, advantages of a plugin vs built-in task:

* easier to fix bugs & add features
* core is smaller, easier maintenance
* we will have a supported task plugin, dog-fooding

srinivasupadhya:
PRs https://github.com/gocd/gocd/pull/858 and https://github.com/gocd/go-plugins/pull/9 are in for this. ready to be reviewed.

PR #858, opened 2015-01-29
Move rake task out into task plugin

srinivasupadhya:
As part of #833 

Discussion: https://groups.google.com/d/topic/go-cd/EM4hCJEwcHc/discussion


srinivasupadhya:
we decided not to do this.

Issue #859, opened 2015-01-29
Git clean unversioned files does not delete .gitignored files


mwalton236:

mwalton236:
Any files listed in the .gitignore file are not deleted when the working copy is cleaned

arvindsv:
Closing as duplicate of https://github.com/gocd/gocd/pull/1059, since there is more conversation there.

PR #860, opened 2015-01-29
#859 - change git clean to -xdff so gitignored files are also cleaned

mwalton236:


arvindsv:
Hi @mwalton236, sorry I don't think I can accept this. This is quite a big change in behavior, which could catch people unawares. I've heard of people using .gitignore exactly like this. I don't agree with it being used that way, but unfortunately it is done. :( Sorry.

It could be done as an option, I suppose. But, changing the default behavior is hard.

Issue #818, opened 2014-12-30
SCM extension point


srinivasupadhya:
This will be on the same lines as Package Repository extension point.

Differences between Package Repository & SCM extension points:

1. there is only one entity (scm) as opposed to 2 (repository & package) in case of Package Repository
2. support for destination folder & filters (as in case of regular SCMs)
3. the `lastModificationSince` will return multiple revisions
4. each revision can return modified files
5. there will ability agent side checkout
6. the view will be defined by plugin (not generated)

To Do:

1. XML CRUD
2. end-point (get config, get view, validate config, check connection, latest revision, latest revisions since, checkout)
3. DB CRUD
4. poller & agent fetch
5. view (materials UI read, build cause pop-up [pipeline dashboard, pipeline history, VSM], compare pipeline, trigger with options pop-up, stage - overview & materials, job - materials, environment dashboard, APIs)
6. materials UI - SCM pop-up (add, edit, use-existing, dis-associate) & new pipeline wizard (add)
7. SCM listing - delete

Sample Config XML:
```
<scms>
    <scm id="3bfc282e-43a6-4795-ba9c-6c50665220dd" name="test">
      <pluginConfiguration id="git" version="0.1" />
      <configuration>
        <property>
          <key>url</key>
          <value>/tmp/git-repo-1</value>
        </property>
      </configuration>
    </scm>
  </scms>
  <pipelines group="first">
    <pipeline name="pipeline">
      <materials>
        <scm ref="3bfc282e-43a6-4795-ba9c-6c50665220dd" dest="dest_dir">
          <filter>
            <ignore pattern="*.log" />
          </filter>
        </scm>
      </materials>
      <stage name="stage">
        <jobs>
          <job name="job">
            <tasks>
              <exec command="ls" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </pipelines>
```

PR #865, opened 2015-01-31
scm end-point {6/7} - ui to add & edit global copy of pluggable scm, choose existing SCM

srinivasupadhya:


srinivasupadhya:
this is ready to be reviewed. needs some UX/UI help.

To Do:
- add 'choose existing', 'add new' radio buttons on top for "add new scm material"
- add 'choose existing', 'add new', 'edit material' radio buttons on top for "edit scm material"
- decide how 'choose existing' is going to work

arvindsv:
I'll probably work with you, on reviewing this. Probably Wednesday or Thursday, if that's ok.

srinivasupadhya:
ok. anytime :+1:

config/config-api/src/com/thoughtworks/go/domain/scm/SCM.java
```diff
@@ -198,34 +221,31 @@ public void applyPluginMetadata() {
...
-            return null;
-        }
-        return new SecureKeyInfoProvider() {
-            @Override
-            public boolean isSecure(String key) {
-                SCMConfiguration configuration = metadata.get(key);
-                return configuration.getOption(SCMConfiguration.SECURE);
+    protected void setPluginConfigurationAttributes(Map attributes) {
+        SCMConfigurations scmConfigurations = SCMMetadataStore.getInstance().getConfigurationMetadata(pluginConfiguration.getId());
+        for (SCMConfiguration scmConfiguration : scmConfigurations.list()) {
```
arvindsv: The scmConfigurations list can be null. Is not handled. Any reason? It was handled earlier ([```metadata == null```](https://github.com/gocd/gocd/pull/865/files#diff-241b146f9f85defaee538d834fcbe881L219)).

server/src/com/thoughtworks/go/server/service/materials/PluggableScmService.java
```diff
@@ -0,0 +1,78 @@
...
+                ConfigurationProperty property = scmConfig.getConfiguration().getProperty(key);
+                String configValue = property == null ? null : property.getValue();
+                if (isRequired && StringUtil.isBlank(configValue)) {
+                    validationResult.addError(new ValidationError(key, localizer.localize("MANDATORY_CONFIGURATION_FIELD")));
+                }
+            }
+        }
+
+        for (ValidationError validationError : validationResult.getErrors()) {
+            scmConfig.getConfiguration().getProperty(validationError.getKey()).addError(validationError.getKey(), validationError.getMessage());
```
arvindsv: If a plugin author returns a validation result with a wrong key, this line will throw a null pointer exception. I think this deserves a check.

server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
```diff
@@ -0,0 +1,212 @@
...
+          cruise_config.pipelineConfigByName(CaseInsensitiveString.new(params[:pipeline_name])).materialConfigs().last()
+        end
+
+        def update(cruise_config)
+          scm = @material.getSCMConfig()
+          scm.setConfigAttributes(params[:material])
+          scm.ensureIdExists
+          params[:material][PluggableSCMMaterialConfig::SCM_ID] = scm.getSCMId
+
+          @pluggable_scm_service.validate(scm)
```
arvindsv: Why is this continuing even after validate might have failed? Is that the pattern in this whole save_popup area? I guess it will fail after, while validating everything. Sigh.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pluggable_scm_check_connection.js
```diff
@@ -0,0 +1,42 @@
...
+
+    _bind = function (form_selector, check_connection_selector, connection_message_selector) {
+        jQuery(form_selector).on('click', check_connection_selector, function (e) {
+            var connection_message_element = jQuery(connection_message_selector);
+            connection_message_element.removeClass("error_message").removeClass("ok_message").text('Checking connection...');
+            jQuery.ajax({
+                type: "POST",
+                url: url,
+                data: jQuery(form_selector).serialize(),
+                success: function (data) {
```
arvindsv: Just like "success", the [error event](http://api.jquery.com/Ajax_Events/) will also need to be handled, and a message shown. Otherwise, if there's an error (server not accessible, internal server error, etc), it will get stuck at "Checking connection ...".

server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
```diff
@@ -0,0 +1,212 @@
...
+    def edit
+      assert_load :material, @pipeline.materialConfigs().getByFingerPrint(params[:finger_print])
+      assert_load :meta_data_store, meta_data_store
+      render layout: false
+    end
+
+    def update
+      material = @pipeline.materialConfigs().getByFingerPrint(params[:finger_print])
+      assert_load :material, material
+      assert_load :index, @pipeline.materialConfigs().indexOf(material)
```
arvindsv: Should this use indexOf? I know we talked about this.

server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
```diff
@@ -0,0 +1,212 @@
...
+        end
+
+        def node(cruise_config)
+          cruise_config
+        end
+
+        def subject(cruise_config)
+          pipeline = cruise_config.pipelineConfigByName(CaseInsensitiveString.new(params[:pipeline_name]))
+          #Since the fingerprint that got submitted could be different from the new fingerprint after the setConfigAttributes, use the index.
+          if @index
```
arvindsv: Why will @index be nil? What is the consequence of not returning material_config, in that case?

server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/show_existing.html.erb
```diff
@@ -0,0 +1,46 @@
...
+    <div class="form_content exec_task_editor">
+        <%= render :partial => "admin/shared/global_errors.html", :locals => {:scope => {}} -%>
+        <div id="material">
+            <%= render :partial => 'shared/form_required_message.html', :locals => {:scope => {}} %>
+
+            <%- if @scms.empty? -%>
+                <p class="warnings">No SCMs found. Please add a SCM first.</p>
+            <%- end -%>
+
+            <div class="fieldset">
```
arvindsv: We talked about this, right? About how this allows any one to edit every SCM configured in the system. Can we (easily) restrict that? So that edit can be done only by admins? I ask because it's easier to go back and give permissions to everyone, than to take it back (should we choose that the behavior is wrong).

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -149,6 +149,14 @@
   get "admin/package_repositories/:plugin/config/" => "admin/package_repositories#plugin_config", constraints: {:plugin => ALLOW_DOTS}, as: :package_repositories_plugin_config
   get "admin/package_repositories/:id/:plugin/config/" => "admin/package_repositories#plugin_config_for_repo", constraints: {:plugin => ALLOW_DOTS}, as: :package_repositories_plugin_config_for_repo
 
+  get "admin/pipelines/:pipeline_name/materials/pluggable_scm/show_existing" => "admin/materials/pluggable_scm#show_existing", constraints: {pipeline_name: PIPELINE_NAME_FORMAT}, as: :admin_pluggable_scm_show_existing
```
arvindsv: Have you tried these routes with a plugin whose name has dots in it? That's allowed and caused problems before. That's why it's used with ALLOW_DOTS. See [this](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/config/routes.rb#L149).

arvindsv:
Alright, I've looked through the whole thing, and I've asked questions inline. One last thing I want to do before I accept this (assuming we talk about all my other questions), is: I want to see why the UI of your sample plugin was not showing properly. Can you point me to the code, and a downloadable (built) version of the plugin, preferably? If it has a pom, then it's ok too.

server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
srinivasupadhya: yes you are right. It needs to go & validate everything before rendering error so user gets all messages together.

server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/show_existing.html.erb
srinivasupadhya: I think we should allow editing with a warning about global entity & ask for feedback from users. Im ok removing/turning off code if others think otherwise.

server/webapp/WEB-INF/rails.new/spec/controllers/admin/materials/pluggable_scm_controller_spec.rb
```diff
@@ -0,0 +1,303 @@
...
+  include ConfigSaveStubbing
+  include MockRegistryModule
+
+  before do
+    controller.stub(:populate_health_messages)
+  end
+
+  describe "routes should resolve and generate" do
+    it "show_existing" do
+      {:get => '/admin/pipelines/pipeline.name/materials/pluggable_scm/show_existing'}.should route_to(:controller => 'admin/materials/pluggable_scm', :action => 'show_existing', :pipeline_name => 'pipeline.name')
```
srinivasupadhya: there is test for pipeline name with `.`

server/webapp/WEB-INF/rails.new/config/routes.rb
srinivasupadhya: the `PIPELINE_NAME_FORMAT` takes care of that. there is test as well.

config/config-api/src/com/thoughtworks/go/domain/scm/SCM.java
srinivasupadhya: fixed.

server/src/com/thoughtworks/go/server/service/materials/PluggableScmService.java
srinivasupadhya: fixed.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pluggable_scm_check_connection.js
srinivasupadhya: fixed

srinivasupadhya:
Here is the plugin jar - https://github.com/srinivasupadhya/git-poller/releases/tag/0.1

I have fixed most of the comments. The plugin UI is working as expected. Have a look & let me know if I need to do something more on this.

server/webapp/WEB-INF/rails.new/config/routes.rb
arvindsv: Ok, cool.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pluggable_scm_check_connection.js
arvindsv: I think it's worth showing "data", instead of just "Error occurred!".

server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
arvindsv: What about this?

server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/show_existing.html.erb
arvindsv: Then, show it to a couple of people and see what they say.

arvindsv:
I tried the plugin. I see this:

image

I can understand JGit. I believe "SCM" is showing up because of the changes in this pull request. Is this right? If so, it is unbelievably confusing. It took me a long time to understand what (I think) is going on. I think, the ability to add a new SCM material, and the ability to edit an existing (unrelated) SCM material have been added into this dropdown. Umm, I don't think we need such confusion in the already quite confusing dropdown. What do you think? Can you please talk to others?

If you really think this is first cut, and should be in, then it has to be feature toggled and turned off. Not "turned off by default, because there is no plugin" kind of turned off. Once this goes in, people will start using it, and it will be very hard to remove or change this. I'd want to get it right, before it goes in.

I think this PR should introduce a dropdown item which says, "Other SCMs" (if an SCM plugin exists), and then show "Choose an existing SCM" or "Create", more or less like "Package". I am willing to work with you on this.

srinivasupadhya:
The `JGit` dropdown allows you to add a JGit material. So as the plugins increase the items in dropdown will get added. The `SCM` (can give other name i guess) is purely to select an existing SCM in global list. This is currently just a dropdown of SCM names which can be improved. But I think this is the cheapest (effortwise not qualitywise :D) to get it out. If there are no plugins then there is only one entry in the dropdown `SCM` that allows discover of SCM plugins like `Package` does today for package material plugins.

arvindsv:
It doesn't show up in pipeline wizard. :(

srinivasupadhya:
Done as part of #818

Issue #142, opened 2014-04-28
Console output with ansi colors


selvakn:
Like https://wiki.jenkins-ci.org/display/JENKINS/AnsiColor+Plugin

scottmuc:
I would very much like this! Here's a snippet of output that we would like to see colourized:

    12:35:10.012 [32m12:35:10 [local] OUT[0m   libp11-kit0 libpopt0 libquadmath0

ketan:
scott - this feature is currently behind a toggle. However it only works when the build is running, not after it is completed.

We're working on an alternative implementation that works both when the build is running as well as after it is completed - #1130

chopraapooja:
I am picking this to test.

chopraapooja:
ANSI colors, switching themes, GUI seems working fine.
Found two issues : [#1196] (https://github.com/gocd/gocd/issues/1196) and [#1221] (https://github.com/gocd/gocd/issues/1221)

srinivasupadhya:
both #1196 and #1221 don't seem to be related to this story. closing this since @chopraapooja is done testing this.

scottmuc:
Awesome! Hope to give go.cd another whirl on my project.

Issue #434, opened 2014-08-08
Migrating job details page from velocity to rails-erb


sachinsudheendra:
This issue is specific to moving the job-details page, which currently is one among the few pages using velocity templates into rails-erb. Also, redesigning of the page might occur in this process.

I've moved the console tab and few details specific to a job - and will be using a combination of rails-erb and angularjs to achieve this. The changes are being done on my [fork](https://github.com/sachinsudheendra/gocd/tree/jobs-rails4) (branch: jobs-rails4) and will be submitted to #gocd when done.

With my limited CSS skills, I've come up with the below first-cut of the page. Any feedback is welcome.

![proposed](https://cloud.githubusercontent.com/assets/52740/3859892/0b27d250-1f1e-11e4-8b87-6974eb923c0c.png)

For comparison, here's the current page

![current](https://cloud.githubusercontent.com/assets/52740/3859895/34050bf2-1f1e-11e4-8e98-3d1e9284bc3c.png)

/cc #427

srinivasupadhya:
[discussion thread](https://groups.google.com/d/topic/go-cd/RE7V5CavdNc/discussion)

arvindsv:
Cleanup: Stale. Closing.

PR #868, opened 2015-02-03
Issue #142 - Colorize console logs

ketan:
Based on code that we've been using on snap-ci.com close to about 2 years now.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/application.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/application.js
@@ -38,6 +38,7 @@
 //= require "lib/trimpath-template-1.0.38.js"
 //= require "lib/ui.core-1.7.3.js"
 //= require "lib/ui.dialog-1.7.3.js"
+//= require "lib/ansiparse.js"
 //= require "build_base_observer.js"
 //= require "json_to_css.js"
 //= require "util.js"
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
@@ -69,14 +69,29 @@ BuildOutputObserver.prototype = {
     _update_live_output: function (build_output) {
         var is_output_empty = !build_output;
         if (!is_output_empty) {
-            var escapedOutPut = build_output.escapeHTML();
-            if (Prototype.Browser.IE) {
-                // Fix for the IE not wrap /r in pre bug
-                escapedOutPut = '<br/>' + escapedOutPut.replace(/\n/ig, '<br\/>');
-            }
-            if($('buildoutput_pre')){
-                $('buildoutput_pre').innerHTML += escapedOutPut;
-            }
+            var tempArea = jQuery('<div></div>');
+            jQuery.each(ansiparse(build_output), function() {
+                var output = jQuery('<span></span>').html(this.text.escapeHTML());
+                if (this.foreground) {
+                    output.addClass("fg-" + this.foreground);
+                }
+                if (this.background) {
+                    output.addClass("bg-" + this.background);
+                }
+
+                if (this.bold) {
+                    output.addClass('bold');
+                }
+                if (this.italic) {
+                    output.addClass('italic');
+                }
+                if (this.underline) {
+                    output.addClass('underline');
+                }
+                tempArea.append(output);
+            });
+
+            jQuery('#buildoutput_pre').append(tempArea.html());
         }
         return is_output_empty;
     },
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/ansiparse.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/ansiparse.js
@@ -1,0 +1,190 @@
+// ansiparse.js (https://github.com/mmalecki/ansiparse)
+// Copyright (C) 2011 by Maciej Małecki
+// MIT License (see LICENSE file)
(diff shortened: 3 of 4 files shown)


ketan:
This is only an initial draft to get comments and feedback from the go devs. There are a few things to consider -

* some sort of color map (to choose between themes)
* too much console log (will blow this up in the browser). Alternative is to implement this on the server end, but that may as well end up chewing up CPU to render logs.
* coloring when the build is complete - currently it appears that gocd will open the console.log in an iframe

arvindsv:
> too much console log (will blow this up in the browser).

This is the one I'm a little worried about. The change looks alright to me. We will need to write a Jasmine spec for this. I don't see one existing today. :(

Hmm. Never realized it's an iframe. Needn't be, I suppose?

ketan:
> > too much console log (will blow this up in the browser).

> This is the one I'm a little worried about. The change looks alright to me. We will need to write a Jasmine spec for this. I don't see one existing today. :(

This is because we're throwing the entire console log into ansiparse which causes the browser to run out of memory.

Because js libs (atleast the versions that gocd is using) can't to chunking/streaming of console log we're in no position to parse it in a streaming/chunking manner which would reduce the memory footprint as well.

**Alternatives -**

* Implement this on the server side - Depending on the log size, and number of users, this may chew up a lot more CPU cycles to render the log.
* When a job is running render only the last few hundred lines of the console log - Jenkins does this. So does snap.

> Hmm. Never realized it's an iframe. Needn't be, I suppose?

I don't have any arguments for or against this, although consistency with how it's rendered when live tailing may give a more consistent l&f and coloring.

arvindsv:
Hey @ketan. I thought I'd accept this, maybe with a feature flag since we don't yet know the performance impact of it, with large build logs. But, I found out what you mean by "coloring when the build is complete - currently it appears that gocd will open the console.log in an iframe". So, the coloring goes away when a build is complete. Ok, that means that I cannot accept it in its current state, I suppose, right?

I took a look at the output and it seems reasonable. It doesn't create a span for each line unnecessarily. It does create a span only when the color changes. We were thinking of using the color changes to change the color of the timestamps which show in the log. So, in that case every line will have at least two spans. Will need to definitely look at the performance then.

I took a look at the code (existing code, not just yours), and I think it's doing too much and ... not in a very nice way, especially with the switch to the iframe after a build is finished. Ideally, it should be responsible for periodically fetching new information, coloring it and once the build is done, to stop the polling. I don't really see why it needs to switch to an iframe. I know Sachin had worked on this earlier. Is probably time to move it out of the land of vm (velocity macros) and modernize it, keeping the look more or less the same.  /cc @srinivasupadhya

ketan:
@arvindsv you are right - things in that area are messier than they have to
be.

I agree about the bit about feature toggles, this should give us an option
to experiment with console log coloring with a fallback for users who have
bigger logs that may cause browser crashes.

As a aside - I found it hard to figure out how to use jquery's $,
apparently because $ seems to point to prototype.

Also jquery seems to be a slightly older version that I could not use some
of the newer APIs available.

srinivasupadhya:
@arvindsv #434 is what @sachinsudheendra was working on. doesn't seem like he is pursuing it. someonelse could pick it up.

@ketan yes, $ defaults to prototype! jquery upgrade may be @jyotisingh will be able comment on better. she worked on assets & test migration to jasmine.

Also, there seems to be some merge conflict you need to take care of.

arvindsv:
jQuery can be accessed using "jQuery", instead of "$". That's because we use [jQuery.noConflict](http://api.jquery.com/jquery.noconflict/).

ketan:
@srinivasupadhya -

There are a couple of IE specific bug fixes and html escaping (https://github.com/gocd/gocd/blob/22f8a67fffdbcc463f0cb1cc1c811b5208cde9b8/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js#L72-L81), which I believe is something that needs to be looked into, I'm happy to rebase, but I don't have VMs to test this on various IE versions.

jyotisingh:
While we would love to be on the latest version of all libraries/frameworks we use, it doesn't happen that way for us because of many reasons. For instance, there is no particular reason for not upgrading jQuery apart from the fact that the core contributors could not make time for this. Also, jQuery has undergone major changes from the version we currently use, so it would require considerable amount of time to test existing javascripts across all browsers, fix if an issue arises. 
If anyone is willing to spend that kind of time, we would be happy to review and merge the PR. 

On a related note, jQuery 2.x does not support IE8. There was a [discussion](https://groups.google.com/d/msg/go-cd/FLuRYmJn8RQ/1w40nHY41n0J) around stopping support for IE8, but nothing conclusive came out of it. Since, not many people commented on the post, could we interpret it as "no one raised a voice so no one cares about IE8", and stop supporting it?

arvindsv:
I've merged this. It's painful to see these sitting in the PR queue for so long. This is definitely a work in progress, since it needs changes we're talking about here. Here's what I think is missing:

1. Coloring goes away as soon as a build finished.
2. I switched to a white background, since the black looks very ... out of place (though I personally prefer black). So, the color codes might need to be tweaked. For instance, yellow looks horrible on white.
3. Tests need to be written at around [build_detail_observer_spec.js](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/spec/javascripts/build_detail_observer_spec.js).
4. Remove feature toggle, when it has been around for a bit, and performance is not a problem.

Here's how you toggle this feature on:

```
curl --user username:password -d toggle_value=on http://yourserver:8153/go/api/admin/feature_toggles/color_logs_feature_toggle_key
```

PR #870, opened 2015-02-03
#869 - append changes to build log element, rather than replace html

### LINKED ISSUE #869: Text highlighting is lost in build log when extra logs are appended
While a build is running, if you highlight some text (for example if you want to show someone something) then the highlight is lost whenever new text is appended to the log.

This appears to be because the entire text is being replaced, rather than just the new text being appended.

I presume this (text replacement) would have performance issues as well, as it would probably force a re-layout of the page, but this is pure speculation.

matt-richardson:
Fixes #869.

Use prototypejs functionality to append to the bottom of the existing element, rather than replacing entire innerhtml.

No new tests added - this is covered by existing tests (which I couldn't figure out how to run).

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
@@ -75,7 +75,7 @@ BuildOutputObserver.prototype = {
                 escapedOutPut = '<br/>' + escapedOutPut.replace(/\n/ig, '<br\/>');
             }
             if($('buildoutput_pre')){
-                $('buildoutput_pre').innerHTML += escapedOutPut;
+                $('buildoutput_pre').insert({bottom: escapedOutPut});
             }
         }
         return is_output_empty;


matt-richardson:
Any feedback on this PR?

arvindsv:
I like it! I think it's a very nice change. I'm accepting it. I tried it out with Chrome and Firefox. I haven't tried it with IE. I also output some HTML, to make sure that it does not convert to a DOM element, and allow and XSS attack. So, I think it looks good.

Performance-wise, I think it's going to be alright. The code is executes is very close to the part [here](https://github.com/sstephenson/prototype/blob/1.6.0.3/src/dom.js#L118). If you trace it, finally, it calls [preTagElement.appendChild(documentFragmentForNewPartOfConsoleLog)](https://github.com/sstephenson/prototype/blob/1.6.0.3/src/dom.js#L1005), which feels to me like it should be better than innerHTML +=. A quick JS profile of both heap and CPU did not show up anything alarming.

arvindsv:
Thank you, @matt-richardson!

arikagoyal:
Works fine on 15.1.0(1535-a3ef28c0b8b42f)

Issue #795, opened 2014-12-18
Performance issues with cctray


arikagoyal:
<details to be filled>

arvindsv:
Some details and ideas [here](https://groups.google.com/d/topic/go-cd-dev/0zw0mHEqlVk/discussion).

arvindsv:
I've run some performance benchmarks on the code of this pull request. The benchmark data and graphs are available [here](https://gist.github.com/arvindsv/b44d181d45133a4224f5/) and for download in the [tar.gz file](https://gist.github.com/arvindsv/b44d181d45133a4224f5/raw/bb54840d4cfb962a1f5fec3ea0de1756f302fa3c/cctray.profile.tar.gz). Some context of the information:

* The microbenchmark code, used to generate the graphs, is in the [profile tar.gz file](https://gist.github.com/arvindsv/b44d181d45133a4224f5/raw/bb54840d4cfb962a1f5fec3ea0de1756f302fa3c/cctray.profile.tar.gz) (0001-CCTray-perf-benchmark.patch). It uses JUnit, stubs out the database meaningfully and brings up a few different threads: One which keeps changing the config, and three which keep fetching CCTray requests. It runs this for 30 seconds at a time.
* When the config changes (to add or remove stages and jobs), the stubs change (while those threads are running). So, sometimes, there are exceptions in the output. I didn't want to setup locking, because that might affect the timings. The exception calls are ignored for timing. Since it runs many times every run, it's ok.
* The initial number of groups used is 20, with 50 pipelines each, with 3 stages each and 5 jobs each (leading to 15000 CCTray projects). This config is changed every few milliseconds, to either add a couple of stages and jobs (up to 20 * 50 * 5 * 7 = 35000 projects) or to remove a couple of stages and jobs (down to 20 * 50 * 1 * 3 = 3000 projects). These config changes are done continuously, about once in every 800 milliseconds.
* As mentioned, the DB is stubbed out (to make it easier to be able to change config). But, the effect of the DB should be minimal in both the old and new code.
* The total run is for 30 seconds. This is done three times for each of the old CCTray code and the new CCTray code.

Some interesting points: Comparing [new.3.png](https://gist.github.com/arvindsv/b44d181d45133a4224f5/#file-new-3-data-png) and [old.1.png](https://gist.github.com/arvindsv/b44d181d45133a4224f5/#file-old-1-data-png) (they're all using logarithmic scale):

[...]

PR #877, opened 2015-02-05
#795 - CCTray Perf Fix

arvindsv:
This is a parallel implementation of cctray, available at /go/new_cctray.xml (for now). It reuses much of the existing code, but implements caching in a better way (as mentioned in the mail referenced in #795).

Still need to verify the performance benefits of this. But, this can be accepted if someone has the time to review it, since it is implicitly feature toggled (since the URL is new).


arvindsv:
Personal build is green: go/pipelines/value_stream_map/personal-arvind/7 on 05.

arvindsv:
Some performance numbers are [here](https://github.com/gocd/gocd/issues/795#issuecomment-73578607).

mrmanc:
Hi… just got here from the release notes for 15.1 and thought I’d mention that the URL mentioned in the first message is no longer correct—would it be worth editing it to indicate that the url is not `/go/new_cctray.xml`?

arvindsv:
Done.

PR #880, opened 2015-02-06
Updated plugin "getting started" URL if Javadoc

stephen-souness-springer:
Debatable whether the URL belongs in the Javadoc, but I have updated it to point to a non-404 page.


arvindsv:
Agree. Maybe it shouldn't be there. But, glad it is corrected. Merging.

Issue #400, opened 2014-07-31
upgrade jetty from 6.1.23 (Apr, 2010) to 9.2.3.v20140905 (September 2014)


srinivasupadhya:
http://mvnrepository.com/artifact/org.mortbay.jetty/jetty
6.1.23

http://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server
9.1.5.v20140505

mrmanc:
It would be awesome if this could be an upgrade to [Jetty 9.2.6](http://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/9.2.6.v20141205), as it includes [Slf4jRequestLog](http://download.eclipse.org/jetty/9.2.6.v20141205/apidocs/org/eclipse/jetty/server/Slf4jRequestLog.html), which we should be able to use to output Jetty’s request log over the network using syslog by configuring the [SyslogAppender](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/net/SyslogAppender.html) in log4j.properties.

This would facilitate all sorts of performance / stability monitoring :+1: :chart_with_upwards_trend: :vertical_traffic_light:

arvindsv:
I think it was also in 9.1.5. See [this](http://grepcode.com/file/repo1.maven.org/maven2/org.eclipse.jetty/jetty-server/9.1.5.v20140505/org/eclipse/jetty/server/Slf4jRequestLog.java?av=f).

ketan:
Tested, verified works on 15.1.0(1825-4f2609acf598d7).

PR #886, opened 2015-02-10
Upgraded TFS SDK supporting Visual Studio online per issue #234

Christopher1065:
Issue #234 has a thread around using Visual Studio online. This was my hack to get it working in my environment. I didn't submit a PR previously because it may break the existing TFS functionality but a user suggested I make a PR anyway so here it is. Any feedback is welcome as this is my first contribution to anything open source.


arvindsv:
I'm closing this for now. As I mentioned in #690, in my opinion, this needs to be made toggle-able, like the jetty fix in #400. It's too big a risk to change it in one release. There have been memory issues with TFS before, and have had to make an emergency release. Needs more thought. The work done here is going to be very useful, though.

PR #888, opened 2015-02-13
#869 - Fix build

arvindsv:
IE 9 does not support 'createContextualFragment'. So, need to use innerHTML += for IE.

This failed a [build](https://build.go.cd/go/tab/build/detail/build-windows/271/build-non-server-windows/1/jsunit-win). So, I'm going to accept this fix myself, to fix the build. I've tested this with IE 9, IE 10 and IE Edge. And, the test runs with IE 9, I think. So, this fix should fix the build.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
@@ -68,14 +68,15 @@ BuildOutputObserver.prototype = {
     },
     _update_live_output: function (build_output) {
         var is_output_empty = !build_output;
-        if (!is_output_empty) {
+        var buildoutputPreElement = $('buildoutput_pre');
+        if (!is_output_empty && buildoutputPreElement) {
             var escapedOutPut = build_output.escapeHTML();
             if (Prototype.Browser.IE) {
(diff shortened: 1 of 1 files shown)


matt-richardson:
Apologies for this @arvindsv. I mistakenly assumed that  prototypejs would handle this situation. Bit confused that it doesn't, to be honest! Thanks for fixing for me.

matt-richardson:
Thinking about it a bit more, I'm really surprised that prototypejs doesn't support it.
Looking into that, it appears that [IE9 support is in 1.7](http://prototypejs.org/2010/10/12/prototype-1-7-rc3-support-for-ie9/), but we are only using 1.6. Looks like we need an upgrade to a newer version to avoid these gotchas. New issue?

arvindsv:
@matt-richardson, you're welcome. I was surprised as well! I'd be much more interested in moving to jQuery instead of Prototype (ideally prototype.js is not used in GoCD at all). Unfortunately, it's a move which was started by someone a while ago, but is stuck in limbo. So, half of the scripts use jQuery and half use Prototype. I'm of the opinion that the JS can be written better, to separate the DOM and the actual scripts as well. So, I'd not spend time on upgrading to Prototype, but would love help/ideas on safely moving to jQuery, if that interests you (or others reading this).

Issue #836, opened 2015-01-14
Plugin reload using env variable does not cause metadata store to update


arikagoyal:
1) Let pluginLocationMonitor.sleepTimeInSecs env variable be set. 
2) Add a new yum plugin (with different id - say yum-a) in plugins/external folder
3) The new added plugin gets reloaded and is visible on the plugins page. Also the metadata store gets updated and the new yum id (yum-a) is visbile in 'Add package repository'  screen
4) Now delete yum-a plugin. After some specified time, the plugin goes away from the plugins page. 

Expected - 
The metadata store also gets updated and yum-a should no longer be visibile in 'Add package repository'  screen

Actual -
The metadata store does not get updated and yum-a continues to be visibile in 'Add package repository'  screen

Another scenario - 

With pluginLocationMonitor.sleepTimeInSecs env variable still set, delete few notification plugins from go-server. You still continue to see plugin missing msg in server health msg.

jyotisingh:
The reason for this to happen was that after the bundle is unloaded, the listeners made some calls to the plugins to check for type. This caused exception to be thrown.

arikagoyal:
This issue happens for api based plugins. Works fine on 15.1.0(1541-0371244c807770)

PR #897, opened 2015-02-18
#836 - Removed check for plugin type while deleting data from metadata s...

mdaliejaz:
...tores.


mdaliejaz:
Personal is green.

PR #899, opened 2015-02-20
#898 - Fix: Add LDAP users with special chars in username.

### LINKED ISSUE #898: No way to add LDAP user(s)
Adding an LDAP user results in the "No users selected" message despite the user is actually selected (checkbook is checked) in the "Add user" dialog.

This misbehavior makes the use of AD accounts impossible.
Please see attached screenshot
Go Version: 14.4.0(1356-86064ef5ee28e0)

go add users

arvindsv:
Allows adding LDAP users with special charcaters in their username, full name, etc. Fixes #898.


server/webapp/WEB-INF/rails.new/spec/views/admin/users/_user_results_table_html_spec.rb
```diff
@@ -28,4 +28,28 @@
...
+    user = User.new('!username-with_special_chars._@#$%^&*()-_ and spaces', '!display-name-with_special_chars_@#$%^&*()-_', "email!me#now*@dot.com")
+    user_search_model.should_receive(:getUser).at_least(:once).and_return(user)
+    user_search_model.should_receive(:getUserSourceType).and_return(com.thoughtworks.go.presentation.UserSourceType::LDAP)
+
+    render :partial => "admin/users/user_results_table.html", :locals => {:scope => { :users => [user_search_model]}}
+
+    expected_selection_id = "selection_button_user__username-with_special_chars_dot___________-__and_spacesRAND"
+    expected_name_field_id = "name_user__username-with_special_chars_dot___________-__and_spacesRAND"
+    expected_full_name_field_id = "display_name_user__display-name-with_special_chars__________-_RAND"
+    expected_email_field_id = "email_user__username-with_special_chars_dot___________-__and_spacesRAND"
```
arvindsv: Rubocop config was wrong. Our line max length is not 80. It's 200.

server/webapp/WEB-INF/rails.new/spec/views/admin/users/_user_results_table_html_spec.rb
```diff
@@ -28,4 +28,28 @@
 
     expect(response.body).to have_selector("input[type='radio'][name='selection'][id*='selection_button_user_username_dot_with_dot_dot']")
   end
+
+  it "should santize all fields when generating a DOM ID, to handle fields with special characters" do
```
arvindsv: Rubocop config was wrong. Our line max length is not 80. It's 200.

PR #901, opened 2015-02-23
Plugin download on agent fix

mdaliejaz:


mdaliejaz:
Personal is green.

mdaliejaz:
The server created zip on every re-start which changed the timestamp of the contents, thus changing the md5 of the new zip. The agent then downloaded the plugin zip due to the changed md5.
The fix was to preserve the timestamp at the time of generating the zip file and thus preventing the md5 from getting changed on every restart of the server.

Issue #825, opened 2015-01-07
Plugin reload issue on agent side


arikagoyal:
If pluginLocationMonitor.sleepTimeInSecs env variable is set, a new plugin added gets reloaded on Go server side dynamically (without needing a Go server restart) but the plugin does not get passed onto agent

arvindsv:
@mdaliejaz, have you considered trying to remove the plugins zip itself, and send agents plugins individually?

mdaliejaz:
@arvindsv I think that could be done but I would rather wait for #118 to be played before :)
What do you think?

arvindsv:
Ok. I think they're two different things, though.

zabil:
@mdaliejaz @arvindsv 
Was thinking more about this. I think the implementation is slightly weird.

Does anybody have context/history onto why we are not using 304 (not modified) status but instead  have two separate urls where we pass and receive checksums in the header and then hit the destination url?

We need only one url for e.g
/agents.zip

which returns 200 if we are to download the zip file or 304 if we are to keep the local copy.

zabil:
Maybe to reduce the chattiness?
There'll be 3 requests (/agent, /agent-launcher, /agent-plugins) instead of one (/latest-agent.status)

mdaliejaz:
> Ok. I think they're two different things, though.

I might have understood it wrong then. I think what you meant was to send plugins individually on the agents instead of zipping it. But I thought you meant to pass agent side plugins (not passing the server side plugins) individually. My mistake.
I think it makes sense on what you are saying. This would mean we wouldn't have to pass all the existing plugins every time due to a plugin addition/update/deletion.

@zabil Initially there was only the agent jar and the checksum was checked so that the jar is not downloaded unless there is a change in the checksum. Later with the introduction of bootstrapper and launcher extra checksums were introduced. As part of this change the new URl */latest-agent.status* was introduced to reduce the chattiness (as you said). And I think to keep it same when */agent-plugins.zip* URL was added it also became a part of */latest-agent.status*.

mdaliejaz:
Have created #996 to take care of individual plugins download. 
Thought it's an enhancement and should be done separately :)

arikagoyal:
Verified. Closing on behalf of Srini

PR #905, opened 2015-02-23
#825 - Fix plugin reload on agent

mdaliejaz:


mdaliejaz:
Personal is [green](http://go02.thoughtworks.com:8153/go/pipelines/value_stream_map/personal-ali/24).

plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/listeners/PluginsZipUpdater.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void pluginJarUpdated(PluginFileDetails pluginFileDetails) {
+        pluginsZip.create();
+    }
+
+    @Override
+    public void pluginJarRemoved(PluginFileDetails pluginFileDetails) {
+        pluginsZip.create();
+    }
+}
```
jyotisingh: The implementation worries me a bit. Reason being, now for every new plugin that gets added/removed/updated the whole jar would be recreated everytime, ie. if 10 new plugins get added, the plugin jar would be created 10 times. This won't be pretty for someone using a lot of plugins.

mdaliejaz: Agreed.
Will change the implementation.

plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/commons/PluginsZip.java
```diff
@@ -0,0 +1,84 @@
...
+            return computeMd5DigestOfPlugins();
+        }
+        return md5DigestOfPlugins;
+    }
+
+    private String computeMd5DigestOfPlugins() {
+        try {
+            String digestOfBundledFolder = md5DigestOfFolderContent(bundledPlugins);
+            String digestOfExternalFolder = md5DigestOfFolderContent(externalPlugins);
+            File tempFileToStoreMd5 = File.createTempFile("tempFileToStoreMd5", ".txt");
```
srinivasupadhya: do you need to create a new file. can we create md5 by string concatenation.

plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/monitor/DefaultPluginJarLocationMonitor.java
```diff
@@ -181,40 +183,50 @@ public void run() {
 
         private Set<PluginFileDetails> loadAndNotifyPluginsFrom(File pluginDirectory, Set<PluginFileDetails> knownPluginFiles, boolean isBundledPluginsLocation) {
             Set<PluginFileDetails> currentPluginFiles = getDetailsOfCurrentPluginFilesFrom(pluginDirectory, isBundledPluginsLocation);
-            notifyListenersOfRemovedPlugins(currentPluginFiles, knownPluginFiles);
-            notifyListenersOfUpdatedPlugins(currentPluginFiles, knownPluginFiles);
-            notifyListenersOfAddedPlugins(currentPluginFiles, knownPluginFiles);
+            boolean notifyListenersOfRemovedPlugins = notifyListenersOfRemovedPlugins(currentPluginFiles, knownPluginFiles);
+            boolean notifyListenersOfUpdatedPlugins = notifyListenersOfUpdatedPlugins(currentPluginFiles, knownPluginFiles);
+            boolean notifyListenersOfAddedPlugins = notifyListenersOfAddedPlugins(currentPluginFiles, knownPluginFiles);
+            if (notifyListenersOfRemovedPlugins || notifyListenersOfUpdatedPlugins || notifyListenersOfAddedPlugins) {
```
srinivasupadhya: can this be made more simpler? computing 3 booleans & checking looks a little clumpsy.

plugin-infra/go-plugin-infra/test/com/thoughtworks/go/plugin/infra/DefaultPluginManagerTest.java
```diff
@@ -416,6 +419,17 @@ public void shouldThrowExceptionIfMatchingExtensionVersionNotFound() throws Exce
...
 
+    @Test
+    public void shouldAddPluginChangeListener() throws Exception {
+        DefaultPluginManager pluginManager = new DefaultPluginManager(monitor, registry, mock(GoPluginOSGiFramework.class), jarChangeListener, applicationAccessor, pluginWriter, pluginValidator, systemEnvironment, pluginsZipListener);
+        pluginManager.startInfrastructure();
+
+        InOrder inOrder = inOrder(monitor);
+
+        inOrder.verify(monitor).addPluginJarChangeListener(jarChangeListener);
+//        inOrder.verify(monitor).addPluginJarChangeListener(pluginsZipUpdater);
```
srinivasupadhya: intentional? if yes delete if you make another commit.

plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/monitor/DefaultPluginJarLocationMonitor.java
mdaliejaz: No.
While it looks like a simple check similar to *currentPluginFiles!=knownPluginFiles* will work, it won't!

We need the three boolean values from the three methods since the current and known plugin files might have same name but different md5 (plugin update case). These methods take care of the md5 changes while comparing the plugin files and it would be duplication and unnecessary work to do that again over here. I thought it better to check on the boolean than do the check again.

plugin-infra/go-plugin-infra/test/com/thoughtworks/go/plugin/infra/DefaultPluginManagerTest.java
mdaliejaz: Oops. Yeah. The behaviour changed. That line is no longer required :)
Will delete that.

plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/commons/PluginsZip.java
mdaliejaz: No. Yes, we can.

PR #913, opened 2015-02-25
scm end-point - support post commit hook

srinivasupadhya:


config/config-api/src/com/thoughtworks/go/domain/PipelineGroups.java
```diff
@@ -158,18 +158,17 @@ public void addError(String fieldName, String message) {
         configErrors.add(fieldName, message);
     }
 
-    public HashSet<MaterialConfig> getAllUniquePostCommitSchedulableMaterials() {
-        HashSet<MaterialConfig> materialConfigs = new HashSet<MaterialConfig>();
-        Set<Map> uniqueMaterials = new HashSet<Map>();
```
srinivasupadhya: @sachinsudheendra do you remember why this was a set of map & not set of string?

sachinsudheendra: @srinivasupadhya It was a `Set<Map>` because it was storing & comparing `sqlCriteria` which is a `Map`. I cannot recall the exact reason for using `sqlCriteria` but it might be because I did not want the overhead of fingerprint computation which is a derived field of `sqlCriteria` anyway. We had enough hotspot on fingerprint computation that introducing one here did not make sense.

srinivasupadhya: cool. thanks :+1:

config/config-api/src/com/thoughtworks/go/domain/PipelineGroups.java
```diff
@@ -158,18 +158,17 @@ public void addError(String fieldName, String message) {
...
-                    if (!uniqueMaterials.contains(materialConfig.getSqlCriteria()) && materialConfig instanceof ScmMaterialConfig) {
-                        if (materialConfig.isAutoUpdate()) {
-                            continue;
-                        }
+                    if ((materialConfig instanceof ScmMaterialConfig || materialConfig instanceof PluggableSCMMaterialConfig)
+                            && !materialConfig.isAutoUpdate()
+                            && !uniqueMaterials.contains(materialConfig.getFingerprint())) {
                         materialConfigs.add(materialConfig);
-                        uniqueMaterials.add(materialConfig.getSqlCriteria());
+                        uniqueMaterials.add(materialConfig.getFingerprint());
```
arvindsv: Can you make a couple of small changes?

1. PluggableSCMMaterialConfig#getFingerprint() is not locally cached, as with other materials.

2. materialConfig.getFingerprint() can be null, especially in the case of pluggable SCM materials. How is this handled? My guess is that this will fail, since it will try to add null into the list. If you think it cannot be null, why is PluggableSCMMaterialConfig#getFingerprint() returning null?

arvindsv:
I looked at the rest. I have nothing more. I can merge it. Just that one comment I made earlier.

config/config-api/src/com/thoughtworks/go/domain/PipelineGroups.java
srinivasupadhya: 1. `PluggableSCMMaterialConfig#getFingerprint()` is not locally cached, as with other materials.

Same as `PackageDefinition#getFingerprint()`. They are cached to handle the case of plugin not available. The fingerprint will change based on wether plugin is available or not (in the later case all fields are treated as part of identity).

2. PluggableSCMMaterialConfig#getFingerprint() returning null

will check if that is possible & how it can be handled.

arvindsv: Ok, see about the null. My worry about the fingerprint is performance, obviously. If you look, ```getAllUniquePostCommitSchedulableMaterials``` is not cached at all (even though it goes through every group, pipeline and material). When config changes, all these objects should get thrown away, so locally caching this should be possible, I suppose. Right?

Only case this can fail is that a plugin which was missing is brought back. I suppose that's a very corner case.

srinivasupadhya: The case it causes problems currently is that `fingerprint` is cached in object & the plugin is removed. I remember it was cached before & was removed. @pshivana or @jyotisingh should be able to comment more about it.

I agree that they should be cached. To fix the other issues may be we can force reload the config when a plugin is updated/removed.

arvindsv: > The case it causes problems currently is that fingerprint is cached in object & the plugin is removed.

Yes, you're right.

> I agree that they should be cached. To fix the other issues may be we can force reload the config when a plugin is updated/removed.

I don't think we can afford to do this, till config reload happens in a reasonable time, for big configs. I mean, till config save perf is fixed.

arvindsv: Alright, handle the null, and I'll accept it. I think there can be performance changes we can make here, but it will be premature to change it, without fixing other issues. And, especially without better profiling.

srinivasupadhya: I tried reproducing the scenario in tests & realised that `fingerprint` can never be null since `scmConfig` can never by null. When the config is read its guaranteed to load the right SCM config to `scmConfig`. The null check is for tests.

arikagoyal:
Endpoint - http://[server]/go/api/material/notify/pluggable_scm?scm_name=

srinivasupadhya:
Done as part of #818

PR #920, opened 2015-03-02
Colorize console log

arvindsv:
Accept Ketan's pull request of console log coloring. Still a work in progress. So, it is toggled off by default.


arvindsv:
Accepting since this is @ketan's PR, with a feature toggle added.

Issue #57, opened 2014-04-16
go-server fails to poll and checkout from a hg repository with
authentication using latest version of hg (version greater than
2.0.1)


mdaliejaz:
Description
-----------

While the steps indicate use of a private bitbucket repo, the bug basically happens when Go is used with any hg repo using a client version \>=2.0.1

#### Steps to reproduce

- Setup a hg repo with basic authentication enabled. You may also use a private mercurial hg repository in bitbucket with security enabled

- Try configuring a hg material for a pipeline by giving right credentials. Click check connection to make sure configuration is correct.

- Now save the pipeline and un pause it

#### Expected result

The pipeline should trigger fine and should go green

#### **Current result**

PIpeline does not get triggered. Material update fails with an error:
>>Failed to run hg pull command: ERROR: warning: bitbucket.org certificate with >>fingerprint 45:ad:ae:1a:cf:0e:73:47:06:07:e0:88:f5:cc:10:e5:fa:1c:f7:99 not verified >>(check hostfingerprints or web.cacerts config setting) ERROR: abort: http >>authorization required.

This issue occurs with hg client version 2.0.1 or higher.
This is because Mercurial from 2.0.1 onwards does not preserve credentials when doing a clone and as a consequence all subsequent calls (hg pull) to the main repo fails (http://mercurial.selenic.com/wiki/WhatsNew\#Mercurial\_2.0.1\_.282011-12-01.29)

agoyal-git:
Works fine on 15.1.0(1470-8daae3aaaeaef8)

arvindsv:
I think there's a problem with this. When you use this config:

```
<?xml version="1.0" encoding="utf-8"?>
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="75">
  <server artifactsdir="artifacts" commandRepositoryLocation="default" />
  <pipelines group="defaultGroup">
    <pipeline name="Test">
      <materials>
        <hg url="repo" />
      </materials>
      <stage name="defaultStage">
        <jobs>
          <job name="defaultJob">
            <tasks>
              <exec command="ls" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </pipelines>
</cruise>
```
and setup a local repo like this:

```
$ ls
[...]

PR #922, opened 2015-03-02
#57 - Move pre- and post- integration test executions into profile.

arvindsv:
So that they're run only when docker-tests profile is active.


arvindsv:
Fixing build.

Issue #831, opened 2015-01-09
Null pointer exception in logs


arikagoyal:
Not sure about the steps to reproduce (i will update the issue if i find) but sometime when i was playing around with plugin upload form i got this error in go-server log (on UI, i started getting blank screen on job configuration and new pipeline wizard screen)

```
2015-01-08 12:16:32,147 FATAL [148765233@qtp-1101722853-18] Rails:? - 
Java::JavaLang::NullPointerException ():
  com.thoughtworks.go.server.service.tasks.TaskViewService.allPluginTasks(com/thoughtworks/go/server/service/tasks/TaskViewService.java:136)
  com.thoughtworks.go.server.service.tasks.TaskViewService.allTasks(com/thoughtworks/go/server/service/tasks/TaskViewService.java:119)
  com.thoughtworks.go.server.service.tasks.TaskViewService.getTaskViewModelsWith(com/thoughtworks/go/server/service/tasks/TaskViewService.java:69)
  com.thoughtworks.go.server.service.tasks.TaskViewService.getTaskViewModels(com/thoughtworks/go/server/service/tasks/TaskViewService.java:55)
  java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:606)
  RUBY.index(app/controllers/admin/tasks_controller.rb:12)
  ActionController::ImplicitRender.send_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/action_controller/metal/implicit_render.rb:4)
  ActionController::ImplicitRender.send_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/action_controller/metal/implicit_render.rb:4)
  AbstractController::Base.process_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/abstract_controller/base.rb:189)
  AbstractController::Base.process_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/abstract_controller/base.rb:189)
  ActionController::Rendering.process_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/action_controller/metal/rendering.rb:10)
  ActionController::Rendering.process_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/action_controller/metal/rendering.rb:10)
  AbstractController::Callbacks.process_action(vendor/bundle/jruby/1.9/gems/actionpack-4.0.4/lib/abstract_controller/callbacks.rb:18)
[...]

PR #927, opened 2015-03-03
#831 - Added a check to avoid NullPointerException.

mdaliejaz:


arvindsv:
There's a test missing. Once that is done, the comment should not not be necessary. Closing this for now.

PR #928, opened 2015-03-03
Fixing build

jyotisingh:
Reverted some usages for ClassPathResource. Implemented it in the right way for  the others. Fixed test-resources path.


arvindsv:
Ran personal. Saw green. Let's see!

PR #936, opened 2015-03-06
#795 - Make cctray toggle-able.

arvindsv:
Till now, the new CCTray implementation was at /new_cctray.xml. This makes
it be toggleable at /cctray.xml itself. More importantly, the new cctray
implementation is on by default, now.

Toggle key is: new_cctray_feature_toggle_key


arvindsv:
This is a little hard to verify, since the two implementations are  supposed to be behaviorally similar. However, there is a small change in output that we can exploit. When there are no pipelines (or user doesn't have any permitted pipelines, then the empty cctray.xml output is different between the old and the new):

```
$ curl http://localhost:8153/go/cctray.xml; echo
<Projects>
</Projects>
$
$ curl -d toggle_value=off http://localhost:8153/go/api/admin/feature_toggles/new_cctray_feature_toggle_key; echo
{"message":"success"}
$ curl http://localhost:8153/go/cctray.xml; echo
<Projects />
```

First one is new, second one is old.

I need to verify whether the feature toggle check itself adds much to the time taken. I doubt it very much, but will check. However, we should remove this feature toggle as soon as 15.1 is released.

arvindsv:
Did a quick perf check of just the FeatureToggleService. With a thread pool of 100 threads, and 1 million calls to the server, with 90% isToggleOn calls and 10% changeValueOfToggle calls. It finished in 20 seconds, with a reasonable throughput of 50000 per second, in isolation. I think that's good enough for cctray. Even if you have 500 CCTray callers (usual 5 second time), it's only a 100 on average, per second. The toggle check shouldn't add any real time to the call.

arikagoyal:
Works fine on 15.1.0(1632-d41a824d8f5c5d)

Issue #938, opened 2015-03-07
add ability to force schedule pipeline for every change


srinivasupadhya:
currently Go merges multiple modifications of a material for scheduling pipeline.

So if pipeline `A` watches git repo `G` and the last run was for `g1`. Now there are 3 new commits into `G` - `g2`, `g3`, `g4`. Go would schedule `A` be merging them into one instance, thus saving time.

But sometimes it would be useful to force Go to schedule 3 different instances of `A` for each change.

This will be useful to support feature-branch / PR build support. [reference](https://groups.google.com/d/msg/go-cd-dev/Rt_Y5G2VkOc/d6GqsIDckCAJ)

alexouzounis:
hey guys!
Are there any plans in integrating this to mainline and end up in a release ?

Alex

msdisme:
+1

uprial:
+1

aselus-hub:
+1

jamescorbett:
+1

virtuald:
+1

juhasipo:
+1
We have a pipeline for Gerrit review builds. Apparently since Go merges commits it doesn't always build all the submitted patch sets and therefore doesn't mark them as verified/failed.

aselus-hub:
i'm just surprised this is not a high priority core feature... since it's integral to the most common workflow of 'check builds before merging to master/dev'...

aleksih:
+1

We're using Go with Gerrit to build feature branches. This commit merging behaviour prevents us from reliably building every change, hence making it unfeasible to track Gerrit changes. This should be fixed with a high priority - a lot of companies use Gerrit and get a LOT of commits into build queue. We need build and test information on EVERY broken commit, not just for the latest one made.

virtuald:
We use jenkins for building pull requests, because of so many problems with building pull requests with GoCD. It's a shame.

hartjus:
+1

darrenr:
+1

fragsalat:
:+1:

juhasipo:
It really is a shame that GoCD doesn't support this yet. We were able to achieve this with a custom plugin but I think that it isn't the ideal way to implement this. The plugin requires an access to the GoCD API which is uses to poll GoCD to check if the pipeline is clear for the next commit.

[...]

PR #939, opened 2015-03-07
#938 - add support to force schedule pipeline for each new change

srinivasupadhya:
Done as part of #938

### DIFF
--- a/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
+++ b/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
@@ -16,17 +16,23 @@
 
 package com.thoughtworks.go.server.scheduling;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
 import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.PipelineConfig;
+import com.thoughtworks.go.config.StageConfig;
 import com.thoughtworks.go.config.materials.MaterialConfigs;
 import com.thoughtworks.go.config.materials.Materials;
+import com.thoughtworks.go.domain.MaterialRevision;
 import com.thoughtworks.go.domain.MaterialRevisions;
 import com.thoughtworks.go.domain.buildcause.BuildCause;
 import com.thoughtworks.go.domain.materials.Material;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
+import com.thoughtworks.go.domain.materials.Modification;
 import com.thoughtworks.go.server.domain.Username;
 import com.thoughtworks.go.server.materials.MaterialChecker;
 import com.thoughtworks.go.server.materials.MaterialUpdateCompletedMessage;
@@ -56,6 +62,7 @@ import com.thoughtworks.go.serverhealth.HealthStateType;
 import com.thoughtworks.go.serverhealth.ServerHealthService;
 import com.thoughtworks.go.serverhealth.ServerHealthState;
 import com.thoughtworks.go.util.SystemEnvironment;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
@@ -211,6 +218,8 @@ public class BuildCauseProducerService {
                 buildCause.addOverriddenVariables(scheduleOptions.getVariables());
                 updateChangedRevisions(pipelineConfig.name(), buildCause);
                 if (materialConfigurationChanged || buildType.isValidBuildCause(pipelineConfig, buildCause)) {
+                    pruneRevisionsIfRequired(buildType, pipelineConfig, buildCause);
+
                     pipelineScheduleQueue.schedule(pipelineName, buildCause);
 
                     schedulingPerformanceLogger.sendingPipelineToTheToBeScheduledQueue(trackingId, pipelineName);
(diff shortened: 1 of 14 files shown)


srinivasupadhya:
Its WIP because we need to decide if https://github.com/gocd/gocd/pull/858 is going in. If it is then config migration will be `77` else it will be `76`.

srinivasupadhya:
This is done. Ready for review. Currently the attribute `forceScheduleForEveryChange` can be edited only through config xml. I can add ui (checkbox on pipeline config page) if someone thinks thats important.

arvindsv:
We should continue talking about the merits of this before going down this path. Also, I don't think think should be a config.xml parameter at all. This should be a plugin-defined setting. I know it sounds like a user feature, but it should not be. Go's bunching up of commits for svn, git, etc. should continue. SCM plugins might not want that in some cases.

srinivasupadhya:
problem if plugin defines if commits need to be bunched or not is user wont be able to stop it. ex: package repository plugin says every commit needs to run separately then all pipelines are mandated to run that way. could be very time consuming.

server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
```diff
@@ -249,6 +258,52 @@ private void updateChangedRevisions(CaseInsensitiveString pipelineName, BuildCau
         materialChecker.updateChangedRevisions(pipelineName, buildCause);
     }
 
+    void pruneRevisionsIfRequired(BuildType buildType, PipelineConfig pipelineConfig, BuildCause buildCause) {
```
arvindsv: I don't think this is the right way to do this. This work should not be done here. It should be done right after [this](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L184) and [this](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L200) and should happen inside [AutoBuild](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/AutoBuild.java#L40). Otherwise, fan-in would have run, inside [AutoBuild#onModifications](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/AutoBuild.java#L59) and would have decided a compatible revision, based on the top-most material revision, which this code happily goes and (possibly) prunes. I think the revisions passed into ```onModifications``` should itself be changed, in the places I mention above, so that checks like ```buildType.shouldCheckWhetherOlderRunsHaveRunWithLatestMaterials()``` etc. would have run against the correct revisions.

I could be wrong, since you actually know more about this area than I do :), however, I am having a very bad reaction to seeing revisions thrown away, just before scheduling. Even if what you've done is correct, consider what I say, about pruning before starting, or just not considering extra revisions before running fan-in.

I think this is a change that we shouldn't try and rush in, in this release. The only part it will affect is that in the GitHub PR, if multiple PRs happen at close to the same time, they will be bunched together, right? That's ok. We will live with this. I'd rather take time, and get it right, since this area is critical, and getting it wrong here could have huge consequences. I want more people to look at it, for instance, @jyotisingh and others.

arvindsv: Also, isn't there some code somewhere in this area, which removes a build from queue and reschedules it, when a new commit happens, and the build hasn't been picked up yet? That'll need to change too, right? It shouldn't remove the old build from queue any more.

srinivasupadhya:
> I don't think this is the right way to do this. This work should not be done here. It should be done right after this and this and should happen inside AutoBuild. Otherwise, fan-in would have run, inside AutoBuild#onModifications and would have decided a compatible revision, based on the top-most material revision, which this code happily goes and (possibly) prunes.

* The revisions resolved by fan-in would not be pruned. In fan-in's case there would be atleast two materials' revisions with `isChanged` set to true & in that case this would not kick in.
* The code is not inside `AutoBuild` because the `isChanged` is updated [here](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L212), after fan-in is run & the decision to prune or not depends on `isChanged` value.

> I think the revisions passed into onModifications should itself be changed, in the places I mention above, so that checks like buildType.shouldCheckWhetherOlderRunsHaveRunWithLatestMaterials() etc. would have run against the correct revisions.

* I will see what i can do about this.

> I think this is a change that we shouldn't try and rush in, in this release. The only part it will affect is that in the GitHub PR, if multiple PRs happen at close to the same time, they will be bunched together, right? That's ok. We will live with this. I'd rather take time, and get it right, since this area is critical, and getting it wrong here could have huge consequences. I want more people to look at it, for instance, @jyotisingh and others.

* umm..actually this feature comes cheap & is pretty safe. it would not kick in unless you explicitly turn it on for a pipeline (by setting `forceScheduleForEveryChange="true"`). even after you turn it on for a pipeline, it would not kick in for pipeline instance for which fan-in has already pruned (there will be atleast 2 materials whose revisions would be marked changed).

> Also, isn't there some code somewhere in this area, which removes a build from queue and reschedules it, when a new commit happens, and the build hasn't been picked up yet? That'll need to change too, right? It shouldn't remove the old build from queue any more.

* if the pipeline is already scheduled then Go does not try to create new build-cause for it. If the build-cause is created but not scheduled & Go tries to create a new build-cause then pruning should happen again & the revisions should match the one created previously & there should be no change. so we are covered.

ranjib:
@srinivasupadhya you have any plan to rework on this? we are looking forward to use this, and can help in testing,

srinivasupadhya:
Will be giving another shot at this. Hoping to get it right this time :+1:

PR #940, opened 2015-03-07
Scm end point fixes

srinivasupadhya:


common/src/com/thoughtworks/go/domain/materials/Modifications.java
```diff
@@ -129,6 +129,11 @@ public boolean shouldBeIgnoredByFilterIn(MaterialConfig materialConfig) {
             return false;
         }
         Set<ModifiedFile> allFiles = getAllFiles(this);
+
+        if (allFiles.isEmpty()) {
```
arvindsv: What's the rationale for this? Why was this not needed till now?

srinivasupadhya: Till now all supported scms' modifications were guaranteed to have modified files. With pluggable scms its not guaranteed. While most plugins should have it, some might not. And in that case auto-schedule pipelines dont run because if modifications dont contain modified files its an ignored modification currently.

arvindsv: Ok.

arvindsv: Thinking about this a bit more, won't this allow a normal git merge commit to actually trigger a build, when the commit should be ignored because of file modifications? What I mean is: modificationSince returns two commits. First is a commit whose filter should cause the commit to be ignored. Second is a merge commit. Earlier, this would be an ignored set of commits and a build wouldn't run for this. Now, it will.

srinivasupadhya: hmm..ok lets discuss more about this. closing the PR & creating a new one with just the rails change.

Issue #838, opened 2015-01-15
Fix cruise config xml save performance issues


arikagoyal:
Go currently has 1 cruise-config.xml and with time the size tends to become very big. This tends to slow down config saves and starts effecting the performance. 
This issue is a placeholder to start thinking about potential solutions for the same. Details would be updated as we get more clarity around this.

zabil:
Initial observations while profiling 
https://github.com/gocd/gocd/blob/master/server/test/performance/com/thoughtworks/go/config/MagicalCruiseConfigXmlPerformanceTest.java

*shouldReadQuickly*

| Name                                                                                                                                       | Time (ms) | 
|--------------------------------------------------------------------------------------------------------------------------------------------|-----------| 
| com.thoughtworks.go.config.GoConfigGraphWalker.walkFields(Object, ValidationContext, GoConfigGraphWalker$Handler) GoConfigGraphWalker.java | 11639     | 
| com.thoughtworks.go.config.parser.GoConfigClassLoader.parse() GoConfigClassLoader.java                                                     | 10210     | 
| com.thoughtworks.go.config.CruiseConfig$ErrorCollectingHandler.handle(Validatable, ValidationContext) CruiseConfig.java                    | 10127     | 
| com.thoughtworks.go.config.parser.GoConfigFieldLoader.parse() GoConfigFieldLoader.java                                                     | 10034     | 
| com.thoughtworks.go.config.CruiseConfig$5.handleValidation(Validatable, ValidationContext) CruiseConfig.java                               | 9987      | 
| com.thoughtworks.go.config.parser.GoConfigClassLoader.parseCollection(Collection) GoConfigClassLoader.java                                 | 9923      | 



*shouldWriteQuickly*

| Name                                                                                                                                                     | Time (ms) | 
|----------------------------------------------------------------------------------------------------------------------------------------------------------|-----------| 
| com.thoughtworks.go.config.GoConfigGraphWalker.walkFields(Object, ValidationContext, GoConfigGraphWalker$Handler) GoConfigGraphWalker.java               | 11082     | 
| com.thoughtworks.go.config.GoConfigGraphWalker.walkCollection(Object, ValidationContext, GoConfigGraphWalker$Handler) GoConfigGraphWalker.java           | 10854     | 
| com.thoughtworks.go.config.CruiseConfig$5.handleValidation(Validatable, ValidationContext) CruiseConfig.java                                             | 9716      | 
[...]

PR #887, opened 2015-02-11
#838 - Removing extra validation step while creating

zabil:
*goConfigFileDao* return a valid config file. Writing to the output stream to create the document root can skip the validation.

PR #941, opened 2015-03-09
#838 - Removing extra validation step while creating

zabil:
goConfigFileDao return a valid config file. Writing to the output stream to create the document root can skip the validation


arvindsv:
Related to #887. Merged.

arikagoyal:
Verified config save functionality. Works fine on 15.1.0(1594-db70c652aa3804)

PR #942, opened 2015-03-09
Upgrading from Jetty6(6.1.23) to Jetty9 (v 9.2.3)

jyotisingh:
The version of jetty used can be switched during server startup, so as of now both the versions get packaged. Upon upgrade, Go will switch over to jetty 9. If you need to switch back jetty 6, add a JVM arg 'app.server' with value 'com.thoughtworks.go.server.Jetty6Server' during server startup. The default value for this property is 'com.thoughtworks.go.server.Jetty9Server'


common/pom.xml
```diff
@@ -165,6 +165,22 @@
                     <groupId>com.sun</groupId>
                     <artifactId>tools</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.mortbay.jetty</groupId>
+                    <artifactId>servlet-api-2.5</artifactId>
```
arvindsv: Really? Does the artifact ID have 2.5 in it? That seems wrong.

common/pom.xml
```diff
@@ -292,6 +308,22 @@
                                 <groupId>com.sun</groupId>
                                 <artifactId>tools</artifactId>
                             </exclusion>
+                            <exclusion>
+                                <groupId>org.mortbay.jetty</groupId>
+                                <artifactId>servlet-api-2.5</artifactId>
```
arvindsv: Same as earlier.

app-server/pom.xml
```diff
@@ -0,0 +1,284 @@
...
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.sun</groupId>
+                    <artifactId>tools</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.mortbay.jetty</groupId>
```
arvindsv: Version number in artifact? It's an exclusion for cobertura, so it doesn't matter much, but still.

common/src/com/thoughtworks/go/util/validators/JettyWorkDirValidator.java
```diff
@@ -29,6 +29,7 @@
     public Validation validate(Validation val) {
         if (SystemEnvironment.getProperty("jetty.home", "").equals("")) {
             new SystemEnvironment().setProperty("jetty.home", new SystemEnvironment().getPropertyImpl("user.dir"));
+            new SystemEnvironment().setProperty("jetty.base", new SystemEnvironment().getPropertyImpl("user.dir"));
```
arvindsv: Is there a possibility that jetty.home is not empty, but jetty.base is? In that case, jetty.base will be left empty.

development-utility/development-server-with-jetty6/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+
+    <properties>
+        <main.dir>${project.basedir}/../../</main.dir>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jetty.version>6.1.23</jetty.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+			<groupId>com.thoughtworks.go</groupId>
```
arvindsv: Looks like there are tabs instead of spaces here. Low prio. :)

installers/server/release/server.sh
```diff
@@ -115,7 +115,7 @@ fi
 SERVER_STARTUP_ARGS+=("-server $YOURKIT")
 SERVER_STARTUP_ARGS+=("-Xms$SERVER_MEM -Xmx$SERVER_MAX_MEM -XX:PermSize=$SERVER_MIN_PERM_GEN -XX:MaxPermSize=$SERVER_MAX_PERM_GEN")
 SERVER_STARTUP_ARGS+=("$JVM_DEBUG $GC_LOG $GO_SERVER_SYSTEM_PROPERTIES")
-SERVER_STARTUP_ARGS+=("-Duser.language=en -Dorg.mortbay.jetty.Request.maxFormContentSize=30000000 -Djruby.rack.request.size.threshold.bytes=30000000")
+SERVER_STARTUP_ARGS+=("-Duser.language=en -Dorg.eclipse.jetty.server.Request.maxFormContentSize=30000000 -Djruby.rack.request.size.threshold.bytes=30000000")
```
arvindsv: In the docs for switching (docs or emergency doc which is not published by default), we need to mention this as well.

common/pom.xml
jyotisingh: Well,  this and others that you mention so far are from the patch I got from Srini. While I made some changes and also moved around the actual jetty dependencies, I completely ignored the exclusions. Artifact id servlet-api-2.5 artifact-id doesnot look right. Thanks for pointing out, will send a patch.

jetty6/src/com/thoughtworks/go/server/util/Jetty6Response.java
```diff
@@ -0,0 +1,22 @@
+package com.thoughtworks.go.server.util;
+
+import org.mortbay.jetty.Response;
+
+public class Jetty6Response implements ServletResponse {
+    private javax.servlet.ServletResponse servletResponse;
+
+    public Jetty6Response(javax.servlet.ServletResponse servletResponse) {
+
```
arvindsv: Empty line.

jetty6/src/com/thoughtworks/go/server/util/Jetty6Request.java
```diff
@@ -0,0 +1,27 @@
+package com.thoughtworks.go.server.util;
```
arvindsv: The copyright header is missing in a few files.

jetty9/src/com/thoughtworks/go/server/AssetsContextHandler.java
```diff
@@ -0,0 +1,65 @@
...
+    public void init(WebAppContext webAppContext) throws IOException {
+        String railsRootDirName = webAppContext.getInitParameter("rails.root").replaceAll("/WEB-INF/", "");
+        String assetsDir = webAppContext.getWebInf().addPath(String.format("%s/public/assets/", railsRootDirName)).getName();
+        handler.setAssetsDir(assetsDir);
+    }
+
+    public void handle(String target, javax.servlet.http.HttpServletRequest request, javax.servlet.http.HttpServletResponse response, int dispatch) throws IOException, javax.servlet.ServletException {
+        if (shouldNotHandle()) return;
+        Request base_request = (request instanceof Request) ? (Request) request : null;
+        if (target.startsWith(this.getContextPath()) && !base_request.isHandled()) {
```
arvindsv: base_request null check is missing.

jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
```diff
@@ -0,0 +1,242 @@
...
+
+    ContextHandler welcomeFileHandler() {
+        return new GoServerWelcomeFileHandler();
+    }
+
+    Handler legacyRequestHandler() {
+        return new LegacyUrlRequestHandler();
+    }
+
+    static class LegacyUrlRequestHandler extends HandlerWrapper {
```
arvindsv: This is the perfect time/place to get rid of this. Worst case, it is being used in some place we don't know of, and there will have to be a switch to Jetty6. I think it can't get better than this, to remove this cruft.

jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
```diff
@@ -0,0 +1,242 @@
...
+
+        webAppContext.setConfigurationClasses(new String[]{
+                WebInfConfiguration.class.getCanonicalName(),
+                WebXmlConfiguration.class.getCanonicalName(),
+                JettyWebXmlConfiguration.class.getCanonicalName()
+        });
+        webAppContext.setContextPath(systemEnvironment.getWebappContextPath());
+        webAppContext.setWar(getWarFile());
+        webAppContext.setParentLoaderPriority(systemEnvironment.getParentLoaderPriority());
+//        wac.setAttribute("org.eclipse.jetty.server.webapp.WebInfIncludeJarPattern", ".*/spring-[^/]*\\.jar$");
```
arvindsv: Is this comment needed, or something we need to talk about?

jetty9/src/com/thoughtworks/go/server/util/GoPlainSocketConnector.java
```diff
@@ -0,0 +1,49 @@
...
+    private final Connector httpConnector;
+
+    public GoPlainSocketConnector(Jetty9Server server, SystemEnvironment systemEnvironment) {
+        this.plainPort = systemEnvironment.getServerPort();
+        httpConnector = plainConnector(server);
+    }
+
+    private Connector plainConnector(Jetty9Server server) {
+        HttpConfiguration httpConfig = new HttpConfiguration();
+        httpConfig.setOutputBufferSize(RESPONSE_BUFFER_SIZE); // 32 MB
```
arvindsv: Isn't this 32KB, not MB? I'd just remove the comment. It seems misleading.

jetty9/src/com/thoughtworks/go/server/util/GoPlainSocketConnector.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.thoughtworks.go.server.Jetty9Server;
+import com.thoughtworks.go.util.SystemEnvironment;
+import org.eclipse.jetty.server.*;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+
+public class GoPlainSocketConnector implements GoSocketConnector {
+    private final int plainPort;
+    private static final int MAX_IDLE_TIME = 30000;
```
arvindsv: Do you mind configuring this (idle time) in SystemEnvironment? Because, the Jetty 6 implementation earlier had no default time specified. So, it must have been using a default time. I don't know if that is 30000. Do you? If you do, then it's ok.

jetty9/src/com/thoughtworks/go/server/util/GoSslSocketConnector.java
```diff
@@ -0,0 +1,101 @@
...
+        connector = sslConnector(server.getServer());
+    }
+
+    private Connector sslConnector(Server server) {
+        if (!keystore.exists()) {
+            storeX509Certificate();
+        }
+
+        HttpConfiguration httpsConfig = new HttpConfiguration();
+        httpsConfig.setOutputBufferSize(RESPONSE_BUFFER_SIZE); // 32 MB
```
arvindsv: Comment should be removed here as well.

rack_hack/src/com/thoughtworks/go/rackhack/DelegatingServlet.java
```diff
@@ -39,10 +38,43 @@ public void init(ServletConfig config) throws ServletException {
...
+        for (Method field : fields) {
+            if (field.getName().equals(name)) {
+                field.setAccessible(true);
+                return field;
+            }
+        }
+        return method(name, klass.getSuperclass());
+    }
+
+    public static Object invoke(Object o, String method, Object... args) throws Exception {
```
arvindsv: Since we know that this invoke is called only from this class (right?), we can remove the variable-length args and just let it be String. I'm thinking about the performance of this, since this is called every request. Reflection is reasonably quick these days, but still, maybe there's no need to make this method so generic, without reason?

server-launcher/pom.xml
```diff
@@ -35,45 +35,9 @@
     <dependencies>
         <dependency>
             <groupId>javax.servlet</groupId>
-            <artifactId>servlet-api</artifactId>
-            <version>2.5</version>
+            <artifactId>javax.servlet-api</artifactId>
+            <version>3.1.0</version>
```
arvindsv: I assume that this version of servlet API works with Jetty 6 as well. However, that is a bit of a change from the older one, so we should probably perf test this well.

arvindsv:
Alright, that's all I have! I'll wait for your comments and probably merge it tomorrow.

arvindsv:
Forgot to mention: It looks like we cannot build using JDK 6 any more. Some Apache Shindig related compile failed, for me. Might want to check. Not that I think we should still be compiling against 6. We should at least move to 7. Maybe this is the impetus. :) We should check though, why it's happening.

common/pom.xml
jyotisingh: Alright, cobertura-2.0.3 depends on servlet-api-2.5(org.mortbay.jetty), hence the exclusion.

common/pom.xml
jyotisingh: same as above

app-server/pom.xml
jyotisingh: same as above

server-launcher/pom.xml
jyotisingh: Hmm..ok.

jyotisingh:
Yes, compiling Go with JDK6 does not work anymore as Jetty9 depends on javax.servlet-api-3.1.0 which requires java 7, it doesnot work with java6. 
Almost a year ago an [announcement](http://www.go.cd/2014/07/09/stopping-support-for-java-jdk-6.html) was made that Go would stop supporting java6. So, I think it should be alright to stop supporting it finally. What do you say?

arvindsv:
> Almost a year ago an announcement was made that Go would stop supporting java6. So, I think it should be alright to stop supporting it finally. What do you say?

Yes, I agree.

arvindsv:
Looks like your rebase brought along older, cctray toggling commits.

jyotisingh:
>> Looks like your rebase brought along older, cctray toggling commits.

Umm, yes :P. I guess that should be harmless as that commit has been merged to master already. I should have merged not rebased.

PR #943, opened 2015-03-10
#796 - Fix Mac installers.

### LINKED ISSUE #120: Plugin log does not get created on mac
### Steps to Reproduce
1) Drop an external plugin on go-server's end (Ensure the plugin does some logging using the logging service)
2) Restart the server

### Expected Result
Plugin log should get created on the agent side

### Actual Result
Plugin log does not get created on the agent side

### LINKED ISSUE #796: Mac installers - "Cannot be opened" message
image

The current Mac installers, for 14.4, show a message when you open them. It says, "Go Agent/Go Server can't be opened because the identity of the developer cannot be confirmed". This happens because of changes in Apple's Mac OSX Gatekeeper around 10.9 and 10.10 (Yosemite). The installer has some files at the top level inside the package, and Gatekeeper flags it as an invalid package because of that, even though it is a signed package.

Workarounds are mentioned [here](http://support.apple.com/en-us/HT202491). You need to go to "Security & Privacy" in System Preferences and allow the application to run, only once. It will work as normal, from then onwards.

### LINKED ISSUE #852: Mac Agent has an incorrect PATH variable values.
PATH variable for job executed by agent does not contain the same PATH values that actual go agent user has. So for example user's values are:
```
gocdagent$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/Server.app/Contents/ServerRoot/usr/bin:/Applications/Server.app/Contents/ServerRoot/usr/sbin
```
while when job executed values are:
```
[Debug Information] Environment variable PATH: /usr/bin:/bin:/usr/sbin:/sbin
```

There was no explicitly defined redefinition of PATH variable anywhere in go.cd configuration on my side.
Is it expected behavior? Where and how it could be reconfigured to include /usr/local/bin as well?

arvindsv:
Details:

* Don't have code inside their top-level .app directory. So, Apple Gatekeeper
  does not deem it an invalid app that should be trashed.

* Allow environment variables like PATH to be overridden (#852).
  Use as: PATH=/my/path open "Go Server.app"

* Can set environment variables during startup (like in /etc/default/go-server
  on Linux). This is in ~/Library/Application Support/Go Agent/overrides.env
  and ~/Library/Application Support/Go Server/overrides.env.

* Logging is corrected (#120)

* They use ~/Library/Application Support/Go Agent and Go Server as their working
  directories.

* Works with Apple and Oracle Java, 6 and 7.

* They give a meaningful message, if Java cannot be found.

This fixes #796, fixes #120 and fixes #852.


kmugrage:
Verified these work on my personal Mac - didn't try to override any variables.

Kens-MacBook-Pro:~ kenmugrage$ uname -a
Darwin Kens-MacBook-Pro.local 14.1.0 Darwin Kernel Version 14.1.0: Mon Dec 22 23:10:38 PST 2014; root:xnu-2782.10.72~2/RELEASE_X86_64 x86_64
Kens-MacBook-Pro:~ kenmugrage$ java -version
java version "1.7.0_75"
Java(TM) SE Runtime Environment (build 1.7.0_75-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.75-b04, mixed mode)

server/src/com/thoughtworks/go/server/util/GoMacLauncher.java
```diff
@@ -41,76 +40,16 @@
...
+    private static final String APPLICATION_SUPPORT_PATHNAME =
+            MessageFormat.format("{0}/Library/Application Support/{1}/", System.getProperty("user.home"), System.getProperty("go.application.name", "Go Server"));
     private static final String GO_CONFIG_DIRECTORY_PATH = APPLICATION_SUPPORT_PATHNAME + "config/";
     private static final String CRUISE_SERVER_URL = "http://localhost:8153/go/";
 
     public static void main(String[] args) {
         new File(GO_CONFIG_DIRECTORY_PATH).mkdirs();
-
-        //TODO: remove this call after a few releases
-        moveFilesFromGoServerDirToConfigDir();
```
jyotisingh: The files movement to config directory was done as a part of a bug fix(Old tracker #5985 - not able to take backup because backup never complete in macosx). Since you have removed moveFilesFromGoServerDirToConfigDir method, one cannot upgrade from Go (pre-12.2) to 15.x or later directly. They need to upgrade to a version between 12.2 and 14.4 first, and then to 15.1. This needs to be documented.

arvindsv: I think no one is (strike that, I mean, should be) using 12.2 (in "prod" on Mac) and considering an upgrade to 15.1. I hope no one is. I don't think we should spend time documenting that, unless someone does hit the issue. 12.2 is over three years old. Let me know, if you think that's ok. If not, I'll document it.

I feel we have too many of those in documentation ("Prior to 12.3.1, Go server stored its data in /Library/Application Support/Go Server. From 12.3.1, it will be in < user-home >/Library/Application Support/Go Server.") and it gets confusing.

Issue #101, opened 2014-04-16
Allow short git sha in pipeline label


mdaliejaz:
## Description
Currently in Go you can name a material, for instance 'git', and reference it in the label template. This however yields to the full, long git revision being used, which is not very nice in the UI.

Go should allow short git revision number to be mentioned as the pipeline label

idapi:
I just discovered the option of including the revision number this way, in our case for our Mercurial material; it will be useful. However, as you say, having a 40+ chars label is not ideal. Would be nice to simply be able to apply some logic in the label template; such as a nice Python slicing - rev[:12] :-)

alexschwartz:
+1 vote for this request.

(I actually filed a ticket for this topic years ago  before GO was open source.)

arvindsv:
Though I agree it would be nice, there's no one who has stepped up and even attempted to take a stab at this. It's now an open-source project, and this would definitely be nice to have. Might be worth bringing it up on the dev mailing list and drumming up some support for it. Maybe someone there is free, and looking for something small to do.

alexschwartz:
I am contributing to implement it; would be my first contribution though.
(Plan: I familiarise myself with the details of the contrib guide, implement it, and generate the merge request.)

Is there a need to discuss the syntax (and/or implementation) on the dev mailing list?
I like the Python slicing syntax.

arvindsv:
> I am contributing to implement it; would be my first contribution though.

That would be wonderful!

> (Plan: I familiarise myself with the details of the contrib guide, implement it, and generate the merge request.)

Sounds good. Ask, if you need any help setting up you environment or anything else. I might not be around much this weekend, but I should be able to respond sometime later. Others can help tool

> Is there a need to discuss the syntax (and/or implementation) on the dev mailing list?

Yes, please do that. If others have no ideas, or no one has any major concerns, then it's your call. You can push for your preferred format, of course. :)

Think about whether it can be made extensible (not necessarily a plugin, in this case, just easily extensible).

alexschwartz:
thanks, I will do so.

paulclarkin:
Note, the Git short hash isn't just the long hash sliced, it attempts to be unique.

See https://www.kernel.org/pub/software/scm/git/docs/git-rev-parse.html under --short.

The length of the short hash can vary as git tries to find a unique hash that's as short as possible.

> Instead of outputting the full SHA1 values of object names try to abbreviate them to a shorter unique name. When no length is specified 7 is used. The minimum length is 4.

[...]

PR #947, opened 2015-03-11
Fix for #101 (short git rev)

alexschwartz:
## Implementation for #101 - Support for short git hash

As discussed in issue #101 and on the dev list (https://groups.google.com/forum/#!topic/go-cd-dev/ZfHNkoumdB0) I implemented support for the following syntax:

  `pipelineLabel = "${COUNT}-${git[:7]}"`

This would truncate the material revision of the material names 'git' such that the first 7 characters are used in the label.

As pointed out by @paulclarkin the short git hash sum can be shorter than 7 character:
> Note, the Git short hash isn't just the long hash sliced, it attempts to be unique.
> See https://www.kernel.org/pub/software/scm/git/docs/git-rev-parse.html under --short.
> The length of the short hash can vary as git tries to find a unique hash that's as short as possible.
>> Instead of outputting the full SHA1 values of object names try to abbreviate them to a shorter unique name. When no length is specified 7 is used. The minimum length is 4.

From my point of the view the truncation is valid to be used in pipeline labels.

##### Implementation details
Most of the logic that was changed is in [PipelineLabel.java](config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java).
It is worth to point out that there was a need to change regular expressions in three independent files:
  * config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java
  * config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
  * config/config-server/resources/cruise-config.xsd

##### TODOs left
* Adapt documentation


common/src/com/thoughtworks/go/domain/MaterialRevisions.java
```diff
@@ -272,12 +272,12 @@ private MaterialRevisions filter(MaterialRevisions other) {
         return filtered;
     }
 
-    public Map<String, String> getNamedRevisions() {
-        Map<String, String> results = new HashMap<String, String>();
+    public Map<CaseInsensitiveString, String> getNamedRevisions() {
+        Map<CaseInsensitiveString, String> results = new HashMap<CaseInsensitiveString, String>();
```
alexschwartz: Changing the return type allows for case-insensitive lookup -- which is required.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
```diff
@@ -236,15 +239,71 @@ public void shouldGetAllTemplateVariableNames() {
...
     public void shouldSupportSpecialCharactors() {
         PipelineConfig pipelineConfig = new PipelineConfig(new CaseInsensitiveString("cruise"), new MaterialConfigs());
         pipelineConfig.setLabelTemplate("pipeline-${COUN_T}-${my-material}${h.i}${**}");
 
         Set<String> variables = pipelineConfig.getTemplateVariables();
-        assertThat(variables.contains("COUN_T"), is(true));
-        assertThat(variables.contains("my-material"), is(true));
-        assertThat(variables.contains("h.i"), is(true));
-        assertThat(variables.contains("**"), is(true));
+        assertThat(variables, contains("COUN_T", "my-material", "h.i", "**"));
```
alexschwartz: I prefer the use of `assertThat(list, contains("elem1", "elem2")` as this yields error message which are more expressive than the one for `assertThat(list.contains("elem1"), is(true)); ...`

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
```diff
@@ -236,15 +239,71 @@ public void shouldGetAllTemplateVariableNames() {
...
     @Test
+    public void shouldValidateCorrectPipelineLabelWithoutAnyMaterial() {
+        PipelineConfig pipelineConfig = new PipelineConfig(new CaseInsensitiveString("cruise"), new MaterialConfigs());
+        pipelineConfig.setLabelTemplate("pipeline-${COUNT}-alpha");
+        pipelineConfig.validate(null);
+        assertThat(pipelineConfig.errors().isEmpty(), is(true));
+        assertThat(pipelineConfig.errors().on(PipelineConfig.LABEL_TEMPLATE), is(nullValue()));
+    }
+
+    private PipelineConfig createAndValidatePipelineLabel(String labelFormat) {
```
srinivasupadhya: This could go to end of file if you do a new commit.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
```diff
@@ -236,15 +239,71 @@ public void shouldGetAllTemplateVariableNames() {
...
+        String labelFormat = "pipeline-${COUNT}-${git}-454";
+        PipelineConfig pipelineConfig = createAndValidatePipelineLabel(labelFormat);
+        assertThat(pipelineConfig.errors().on(PipelineConfig.LABEL_TEMPLATE), is(nullValue()));
+    }
+
+    @Test
+    public void shouldValidatePipelineLabelWithNonExistingMaterial() {
+        String labelFormat = "pipeline-${COUNT}-${NoSuchMaterial}";
+        PipelineConfig pipelineConfig = createAndValidatePipelineLabel(labelFormat);
+        assertThat(pipelineConfig.errors().on(PipelineConfig.LABEL_TEMPLATE), startsWith("You have defined a label template in pipeline"));
```
srinivasupadhya: why is the assertion starts with & not the full message?

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
```diff
@@ -236,15 +239,71 @@ public void shouldGetAllTemplateVariableNames() {
...
+    }
+
+    @Test
+    public void shouldValidateCorrectPipelineLabelWithTruncationSyntax() {
+        String labelFormat = "pipeline-${COUNT}-${git[:7]}-alpha";
```
srinivasupadhya: should there be a tests for cases like `${git[:7}` & `${git[7]}`.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
srinivasupadhya: nice :+1:

config/config-server/resources/cruise-config.xsd
```diff
@@ -712,7 +712,7 @@
     </xsd:simpleType>
     <xsd:simpleType name="labelType">
         <xsd:restriction base="xsd:string">
-            <xsd:pattern value="(([a-zA-Z0-9_\-.!~*'()#:])*[$#]\{[a-zA-Z0-9_\-.!~*'()#:]+\}([a-zA-Z0-9_\-.!~*'()#:])*)+"/>
+            <xsd:pattern value="(([a-zA-Z0-9_\-.!~*'()#:])*[$#]\{[a-zA-Z0-9_\-.!~*'()#:]+(\[:(\d+)\])?\}([a-zA-Z0-9_\-.!~*'()#:])*)+"/>
```
srinivasupadhya: you might want to add a config load test. something like [here](https://github.com/gocd/gocd/blob/4b838fed09498c01b0feb876080de7b74f9f18a8/config/config-server/test/com/thoughtworks/go/config/MagicalGoConfigXmlLoaderTest.java#L1082)

config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
```diff
@@ -164,7 +168,11 @@ private void validateStageNameUniqueness() {
 
     private void validateLabelTemplate() {
         if (XmlUtils.doesNotMatchUsingXsdRegex(LABEL_TEMPLATE_FORMAT_REGEX, labelTemplate)) {
-            addError("labelTemplate", "Invalid label. Label should be composed of alphanumeric text, it should contain material revision or ${COUNT}, or use params as #{<param-name>}.");
+            addError("labelTemplate",
```
srinivasupadhya: should the message be this long? can we do somethingelse?

srinivasupadhya:
documentation needs to be updated [here](http://www.go.cd/documentation/user/current/configuration/admin_use_custom_pipeline_label.html). [code](https://github.com/gocd/documentation)

common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
```diff
@@ -129,11 +161,126 @@ public void shouldCreateLabelIfTemplateIsProvided() {
...
+        String res = assertLabelGroupsMatchingAndReplace("release-${git[:" + GIT_REV_LENGTH + "]}", expectedGroups);
+        assertThat(res, is("release-" + GIT_REVISION));
+    }
+
+    @Test
+    public void canMatchWithOneGitTruncationByOneChar() throws Exception {
+        final int size = GIT_REV_LENGTH - 1;
+        final String[][] expectedGroups = { { "git", size + "" } };
+        String res = assertLabelGroupsMatchingAndReplace("release-${git[:" + size + "]}", expectedGroups);
+        assertThat(res, is("release-" + GIT_REVISION.substring(0, GIT_REV_LENGTH - 1)));
```
srinivasupadhya: This is just `GIT_REVISION`?

common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
```diff
@@ -129,11 +161,126 @@ public void shouldCreateLabelIfTemplateIsProvided() {
...
+    public void shouldNotReplaceTemplateWithoutMaterial() throws Exception {
+        final String label = PipelineLabel.replaceRevisionsInLabel("1.5.0", new HashMap<CaseInsensitiveString, String>());
+        assertThat(label, is("1.5.0"));
+    }
+
+    public static final String SVN_REVISION = "3456";
+    public static final String GIT_REVISION = "c42c0bfa57d00a25496ba899b1f476e6ec8872bd";
+    public static final int GIT_REV_LENGTH = GIT_REVISION.length();
+
+    private String assertLabelGroupsMatchingAndReplace(String labelTemplate, String[][] expectedGroups) throws Exception {
```
srinivasupadhya: these too could go at the bottom for the reason mentioned above.

alexschwartz: resolved.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
alexschwartz: makes sense -- resolved.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
alexschwartz: I tend to prefer to not test for full error message as any minor change of the error message results in a full bunch of strings in the unit tests to be updated. By using startsWith I can ensure that the right error message is used.

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
alexschwartz: maybe it is a good idea to add the unit tests anyway -- of course due to the XSD check there cannot be such a configuration.

common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
alexschwartz: Maybe `canMatchWithOneGitTruncationByOneChar` is not self-explaining, but I currently have not a better idea how to do so.

I aimed for a boundary test: truncating exactly one char.

config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
alexschwartz: To the best of my knowledge this is the message with is shown in UI when I edit a pipeline label in the the pipeline admin form (rather than the settings.xml admin form). 

Hence it should be expressive, and I do not have a clue how to make it shorter and have it still explaining enough. What about pointing to a documentation page or help text?

config/config-server/resources/cruise-config.xsd
alexschwartz: *Status: open*

arvindsv:
I remember this conversation started, so that @alexschwartz can get some feedback about his changes, both from a code perspective (which @srinivasupadhya is doing) and from a build perspective. I ran a personal build against this code and it was largely good. Four failures. See below:

```
Test:	shouldUseUpstreamPipelineLabelForDependencyMaterial
Message:	Expected: is "1.3.0-1234" but: was null

Expected: is "1.3.0-1234"
     but: was null

Expected: is "1.3.0-1234"
     but: was null
  at com.thoughtworks.go.domain.MaterialRevisionsTest.shouldUseUpstreamPipelineLabelForDependencyMaterial(MaterialRevisionsTest.java:507)

============================================

Test:	shouldReturnMapKeyedByGivenMaterialName
Message:	Expected: is "9" but: was null

Expected: is "9"
     but: was null

Expected: is "9"
     but: was null
  at com.thoughtworks.go.domain.MaterialRevisionsTest.shouldReturnMapKeyedByGivenMaterialName(MaterialRevisionsTest.java:443)

============================================

Test:	shouldValidatePipelineLabelWithBrokenTruncationSyntax1
Message:	Expected: a string starting with "You have defined a label template in pipeline" but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."

Expected: a string starting with "You have defined a label template in pipeline"
     but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."

Expected: a string starting with "You have defined a label template in pipeline"
     but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."
  at com.thoughtworks.go.domain.PipelineConfigTest.shouldValidatePipelineLabelWithBrokenTruncationSyntax1(PipelineConfigTest.java:276)

============================================

Test:	shouldValidatePipelineLabelWithBrokenTruncationSyntax2
Message:	Expected: a string starting with "You have defined a label template in pipeline" but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."

Expected: a string starting with "You have defined a label template in pipeline"
     but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."

Expected: a string starting with "You have defined a label template in pipeline"
     but: was "Invalid label. Label should be composed of alphanumeric text, it should contain the build number (as ${COUNT}), a material revision (e.g. ${git} for the full revision or ${git[:6]} for an truncated one), or use params (as #{<param-name>})."
	at com.thoughtworks.go.domain.PipelineConfigTest.shouldValidatePipelineLabelWithBrokenTruncationSyntax2(PipelineConfigTest.java:283)
```

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
srinivasupadhya: @arvindsv what do you think?

common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
srinivasupadhya: yes. you can ignore this.

config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
srinivasupadhya: could be help text [here](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/admin/pipelines/general.html.erb#L16)?

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
arvindsv: I think that's ok.

common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
alexschwartz: *Status: resolved*

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
alexschwartz: *Status: resolved*

common/test/unit/com/thoughtworks/go/domain/PipelineConfigTest.java
alexschwartz: *Status: resolved*

config/config-server/resources/cruise-config.xsd
alexschwartz: added test in commit [bf94e39
](https://github.com/alexschwartz/gocd/commit/bf94e39f958e36cf7639c6ffc82efd104738c720)

alexschwartz:
fixed the broken unit tests with 3a3254b

config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
alexschwartz: I like the message which is used in TOOLTIP_LABEL_TEMPLATE. 
Do you suggest to return TOOLTIP_LABEL_TEMPLATE as localized error message?
For me it looks like as the localized messages are in the server module, so not available for the config-api or common.

srinivasupadhya:
@arvindsv this looks good to go. It would be great if you could have a second look at the regex bits.

@alexschwartz great contribution :+1:  (we will definitely be using this ourselves on http://build.gocd [user: view, password: password]). Good to see all best practices followed religiously :)

alexschwartz:
@srinivasupadhya thank you, I am happy if I can help with a tiny contribution.

alexschwartz:
PR for corresponding documentation update: [PR99](https://github.com/gocd/documentation/pull/99)

arvindsv:
> @arvindsv this looks good to go. It would be great if you could have a second look at the regex bits.

Sorry for the delay. I took a look, earlier. But, I'll take a look early in the morning tomorrow. I think it looks good. Just need one final glance.

arvindsv:
Alright, the regex parts look ok to me. The only feedback I'd have is that ```replaceRevisionsInLabel``` and ```lookupMaterialRevision``` don't really need to be static. ```replaceRevisionsInLabel``` doesn't need to be public. Its real usage is only inside ```PipelineLabel``` and so, it should use the ```label``` field, and it needn't be passed in ```labelTemplate``` as its first argument.

arvindsv:
Meaning, this diff still makes everything pass, without exposing ```replaceRevisionsInLabel```:

```
diff --git a/common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java b/common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
index e48e846..36e2d48 100644
--- a/common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/PipelineLabelTest.java
@@ -16,12 +16,9 @@
 
-import java.util.HashMap;
-import java.util.Map;
-
 import com.thoughtworks.go.config.CaseInsensitiveString;
-import com.thoughtworks.go.config.materials.mercurial.HgMaterial;
 import com.thoughtworks.go.config.materials.ScmMaterial;
+import com.thoughtworks.go.config.materials.mercurial.HgMaterial;
 import com.thoughtworks.go.domain.label.PipelineLabel;
 import com.thoughtworks.go.domain.materials.Modification;
 import com.thoughtworks.go.helper.MaterialsMother;
@@ -30,6 +27,9 @@ import org.hamcrest.core.Is;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import static org.junit.Assert.assertNull;
@@ -155,8 +155,9 @@ public class PipelineLabelTest {
 
     @Test
     public void shouldNotReplaceTemplateWithoutMaterial() throws Exception {
-        final String label = PipelineLabel.replaceRevisionsInLabel("1.5.0", new HashMap<CaseInsensitiveString, String>());
-        assertThat(label, is("1.5.0"));
+        PipelineLabel label = new PipelineLabel("1.5.0");
+        label.updateLabel(new HashMap<CaseInsensitiveString, String>());
+        assertThat(label, is(new PipelineLabel("1.5.0")));
     }
 
     @Test
@@ -246,7 +247,9 @@ public class PipelineLabelTest {
     private String assertLabelGroupsMatchingAndReplace(String labelTemplate, String[][] expectedGroups) throws Exception {
         assertLabelGroupsMatching(labelTemplate, expectedGroups);
 
-        return PipelineLabel.replaceRevisionsInLabel(labelTemplate, MATERIAL_REVISIONS);
+        PipelineLabel label = new PipelineLabel(labelTemplate);
+        label.updateLabel(MATERIAL_REVISIONS);
+        return label.toString();
     }
 
     private void assertLabelGroupsMatching(String labelTemplate, String[][] expectedGroups) {
diff --git a/config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java b/config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java
index c448e3f..0b77ba9 100644
--- a/config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java
+++ b/config/config-api/src/com/thoughtworks/go/domain/label/PipelineLabel.java
@@ -16,15 +16,15 @@
 
 package com.thoughtworks.go.domain.label;
 
+import com.thoughtworks.go.config.CaseInsensitiveString;
+import com.thoughtworks.go.util.StringUtil;
+import org.apache.commons.lang.StringUtils;
+
 import java.io.Serializable;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import com.thoughtworks.go.config.CaseInsensitiveString;
-import com.thoughtworks.go.util.StringUtil;
-import org.apache.commons.lang.StringUtils;
-
 public class PipelineLabel implements Serializable {
     protected String label;
     public static final String COUNT = "COUNT";
@@ -44,8 +44,8 @@ public class PipelineLabel implements Serializable {
 
     public static final Pattern PATTERN = Pattern.compile("(?i)\\$\\{([\\w\\.]+)(\\[:(\\d+)\\])?\\}");
 
-    public static String replaceRevisionsInLabel(String labelTemplate, Map<CaseInsensitiveString, String> materialRevisions) {
-        final Matcher matcher = PATTERN.matcher(labelTemplate);
+    private String replaceRevisionsInLabel(Map<CaseInsensitiveString, String> materialRevisions) {
+        final Matcher matcher = PATTERN.matcher(this.label);
         final StringBuffer buffer = new StringBuffer();
         while (matcher.find()) {
             final String revision = lookupMaterialRevision(matcher, materialRevisions);
@@ -55,7 +55,7 @@ public class PipelineLabel implements Serializable {
         return buffer.toString();
     }
 
-    private static String lookupMaterialRevision(Matcher matcher,  Map<CaseInsensitiveString, String> materialRevisions) {
+    private String lookupMaterialRevision(Matcher matcher,  Map<CaseInsensitiveString, String> materialRevisions) {
         final CaseInsensitiveString material = new CaseInsensitiveString(matcher.group(1));
 
         if (!materialRevisions.containsKey(material)) {
@@ -76,7 +76,7 @@ public class PipelineLabel implements Serializable {
     }
 
     public void updateLabel(Map<CaseInsensitiveString, String> namedRevisions) {
-        this.label = replaceRevisionsInLabel(this.label, namedRevisions);
+        this.label = replaceRevisionsInLabel(namedRevisions);
         this.label = StringUtils.substring(label, 0, 255);
     }
```

alexschwartz:
@arvindsv thanks for the feedback, I applied the changes.

srinivasupadhya:
@alexschwartz - @arvindsv was asking if `${COUNT[:3]}` is a ok. i.e. ability to trim count variable. According to current implementation it is possible & im not sure if it should be allowed. Currently `${COUNT}` is the only variable that guarantees uniqueness of pipeline label.

arvindsv:
I think it makes sense to accept this PR and then talk about it. It doesn't all need to go in at once. I feel this PR has enough value, even with that open question.

alexschwartz:
I can understand the need to either prevent to trim COUNT in general, or prevent to trim it to a short size. 

I suggest I prepare (after the merge?) another commit which rejects to trim COUNT at all.

srinivasupadhya:
Ok :+1: Will accept this as soon as jetty upgrade build goes green. Thanks again :)

alexschwartz:
thanks you :+1:

PR #952, opened 2015-03-15
additionalData map to store material data

srinivasupadhya:


srinivasupadhya:
Moved parsing response out so its done only once.

server/src/com/thoughtworks/go/server/service/materials/PluggableSCMMaterialPoller.java
```diff
@@ -74,6 +96,19 @@ private void populateConfiguration(Configuration configuration, com.thoughtworks
         }
     }
 
+    private void updateAdditionalData(final long materialId, final Map<String, String> materialData) {
```
arvindsv: Why wouldn't you do this inside MaterialRepository? I think it should.

srinivasupadhya: same as [this](https://github.com/gocd/gocd/blob/4b838fed09498c01b0feb876080de7b74f9f18a8/server/src/com/thoughtworks/go/server/materials/PluggableSCMMaterialUpdater.java#L51) & [this](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/server/src/com/thoughtworks/go/server/materials/PackageMaterialUpdater.java#L51). I dont have any opinion, so if you think it should, will do :+1:

arvindsv: :( I think everything that reads or write from material-related tables, should be somewhere in the MaterialRepository or around it.

It's your call, finally.

ketan:
@srinivasupadhya 

Verified - works as expected.

PR #964, opened 2015-03-18
#818 - plugin id may contain '.' & hence the urls need to allow them

srinivasupadhya:


jyotisingh:
Verified on 15.1.0-1643

PR #968, opened 2015-03-18
Notification end point data

srinivasupadhya:
* additional data required to do reporting, status updates etc. load `build-case` from database & `pipeline-group` from config.

* sample data

```json
{
  "pipeline": {
    "group": "pipeline-group",
    "name": "pipeline-name",
    "counter": "1",
    "build-cause": [
      {
        "material": {
          "git-configuration": {
            "branch": "branch",
            "url": "http:\/\/user:******@gitrepo.com"
          },
          "type": "git"
        },
        "changed": true,
        "modifications": [
          {
            "revision": "1",
            "modified-time": "2015-03-19T12:09:32.081Z",
            "data": {
            }
          }
        ]
      }
    ],
    "stage": {
      "name": "stage-name",
      "counter": "1",
      "state": "Passed",
      "result": "Passed",
      "create-time": "2011-07-13T19:43:37.100Z",
      "last-transition-time": "2011-07-13T19:43:37.100Z",
      "jobs": [
        {
          "name": "job-name",
          "schedule-time": "2011-07-13T19:43:37.100Z",
          "complete-time": "2011-07-13T19:43:37.100Z",
          "state": "Completed",
          "result": "Passed",
          "agent-uuid": "uuid"
        }
      ]
    }
  }
}
```

sample JSON for all types of materials:
```json
      {
        "material": {
          "git-configuration": {
            "branch": "branch",
            "url": "http:\/\/user:******@gitrepo.com"
          },
          "type": "git"
        },
        
      },
      {
        "material": {
          "mercurial-configuration": {
            "url": "http:\/\/user:******@hgrepo.com"
          },
          "type": "mercurial"
        },
        
      },
      {
        "material": {
          "svn-configuration": {
            "username": "username",
            "check-externals": false,
            "url": "http:\/\/user:******@svnrepo.com"
          },
          "type": "svn"
        },
        
      },
      {
        "material": {
          "tfs-configuration": {
            "username": "username",
            "project-path": "project-path",
            "domain": "domain",
            "url": "http:\/\/user:******@tfsrepo.com"
          },
          "type": "tfs"
        },
        
      },
      {
        "material": {
          "perforce-configuration": {
            "username": "username",
            "use-tickets": false,
            "view": "view",
            "url": "127.0.0.1:1666"
          },
          "type": "perforce"
        },
        
      },
      {
        "material": {
          "pipeline-configuration": {
            "pipeline-name": "pipeline-name",
            "stage-name": "stage-name"
          },
          "type": "pipeline"
        },
        
      },
      {
        "material": {
          "plugin-id": "pluginid",
          "package-configuration": {
            "k3": "package-v1"
          },
          "type": "package",
          "repository-configuration": {
            "k1": "repo-v1"
          }
        },
        
      },
      {
        "material": {
          "plugin-id": "pluginid",
          "type": "scm",
          "scm-configuration": {
            "k1": "v1"
          }
        },
        
      }
```


server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -16,49 +16,258 @@
...
+            materialRevisionMap.put("modifications", modificationList);
+
+            materialRevisionList.add(materialRevisionMap);
+        }
+        return materialRevisionList;
+    }
+
+    private Map<String, Object> createMaterialDataMap(Material material) {
+        Map<String, Object> materialMap = new LinkedHashMap<String, Object>();
+        if (material instanceof GitMaterial) {
```
arvindsv: All of these should be in their own classes. That's why polymorphic dispatch exists. :) I realize they don't all share an interface. Make them. Something like, InformationProvider, and all kinds of materials implement that interface and they decide what information to put in their own maps.

arvindsv: Also, sleep a little bit.

srinivasupadhya: will do once Go has `feature-branch` support :)

srinivasupadhya: There are already a couple of polymorphic attribute providers like [this](https://github.com/gocd/gocd/blob/ed4536974c8e08d996ca3c8f8daf20d830696fad/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java#L119) and [this](https://github.com/gocd/gocd/blob/ed4536974c8e08d996ca3c8f8daf20d830696fad/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java#L125). I cannot reuse them cause they only populate fields that are part of identity. Adding another for this purpose looked like a worsening the problem.

Also, the attributes of map cannot be changed much since any change would break the API. So there is very little chance of it being reused well.

I know it doesn't look great in the current form. Let me see how I can make this better.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -16,49 +16,258 @@
...
         data.put("stage-name", stage.getIdentifier().getStageName());
         data.put("stage-counter", stage.getIdentifier().getStageCounter());
-        data.put("stage-state", stage.getState());
-        data.put("stage-result", stage.getResult());
-        data.put("create-time", timestampToString(stage.getCreatedTime()));
-        data.put("last-transition-time", timestampToString(stage.getLastTransitionedTime()));
+        data.put("stage-state", stage.getState().toString());
+        data.put("stage-result", stage.getResult().toString());
+
+        Pipeline pipeline = pipelineService.fullPipelineById(stage.getPipelineId());
```
srinivasupadhya: your thoughts on this.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -16,49 +16,258 @@
...
+    private Map<String, Object> createPipelineDataMap(Pipeline pipeline) {
+        Map<String, Object> pipelineMap = new LinkedHashMap<String, Object>();
+        pipelineMap.put("name", pipeline.getName());
+        pipelineMap.put("counter", new Integer(pipeline.getCounter()).toString());
+        pipelineMap.put("label", pipeline.getLabel());
+
+        List<Map> materialRevisionList = createBuildCauseDataMap(pipeline.getMaterialRevisions());
+        pipelineMap.put("build-cause", materialRevisionList);
+
+        List<Map> stageList = new ArrayList<Map>();
```
srinivasupadhya: Also, should it provide info for all stages of pipeline or just the stage in question? the `pipeline-name`, `pipeline-counter`, `stage-name` & `stage-counter` are provided on top for plugin devs. reference (using which he can read the correct stage details in multiple case)

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -16,49 +16,258 @@
...
+    public StageStatusPluginNotifier(NotificationPluginRegistry notificationPluginRegistry, PipelineService pipelineService, PluginNotificationQueue pluginNotificationQueue) {
+        this.notificationPluginRegistry = notificationPluginRegistry;
+        this.pipelineService = pipelineService;
         this.pluginNotificationQueue = pluginNotificationQueue;
     }
 
     @Override
     public void stageStatusChanged(final Stage stage) {
-        Map data = createRequestDataMap(stage);
+        if (notificationPluginRegistry.isAnyPluginInterestedIn(NotificationExtension.STAGE_STATUS_CHANGE_NOTIFICATION)) {
```
srinivasupadhya: Also, i have added a check to make sure all the computation happens if someone is interested (a notification plugin exists)

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
arvindsv: I think it's better to have them polymorphic and have strong tests (maybe at StageStatusPluginNotifierTest itself) to enforce API compatibility. Any change there should break a test here, showing that the API would change.

common/src/com/thoughtworks/go/config/materials/tfs/TfsMaterial.java
```diff
@@ -88,6 +88,10 @@ public MaterialConfig config() {
         return new TfsMaterialConfig(url, userName, domain, getPassword(), projectPath, goCipher, autoUpdate, filter, folder, name);
     }
 
+    public String getDomain() {
```
arvindsv: If/when it becomes polymorphic (other comment here), you should remove this.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationPluginRegistry.java
```diff
@@ -47,6 +47,10 @@ public void removePluginInterests(String pluginId) {
         }
     }
 
+    public boolean isAnyPluginInterestedIn(String notificationName) {
+        return !getPluginsInterestedIn(notificationName).isEmpty();
+    }
+
     public Set<String> getPluginsInterestedIn(String notificationName) {
```
arvindsv: Write a throw-away test on this class, calling these methods in different threads (isAnyPluginInterestedIn and addToMap). I think there will be a concurrent modification exception around notificationNameToPluginsInterestedMap, or at least the Set<String> in it.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
arvindsv: Ok, cool!

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
arvindsv: Yes. Only this stage.

arvindsv: Maybe shallow information about the other stages? Hmm. Can you paste a sample JSON as a gist and link it here? We can take a look at it, and see how it would look to add that.

arvindsv: Definitely worth involving notification plugin authors like @matt-richardson (cc'ed now, I suppose) for thoughts about this. Since he is probably using this, he will know best what is useful. We can then talk about what is performant and possible.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
arvindsv: This is not a cached call. I think it can become problematic. Can you use [pipelineWithModsByStageId](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/PipelineService.java#L77) instead?

matt-richardson:
This looks interesting... Is there more detail about what the actual change is going to look like?
Reason i'm asking is I've put some stuff into https://github.com/matt-richardson/gocd-websocket-notifier to add some pipeline details to a stage change notification... Will this PR remove the need for that part?

arvindsv:
@matt-richardson Yes, this is the discussion about that. There is going to be some more information available. I've [asked](https://github.com/gocd/gocd/pull/968#discussion_r26748989) @srinivasupadhya to put up a JSON of the new information that the Go Server will send, so that we can talk about it.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
srinivasupadhya: have updated with sample JSON

arvindsv: I'd start by removing all the passwords and secure configuration information. I just don't think it makes sense in a notification.

matt-richardson:
Some thoughts on the sample xml:
* very scary that it has the material password in the xml!
* do we get the detail of the build cause? I assume so, but the sample xml doesn't explicitly call it out.

My use case was this app - https://github.com/matt-richardson/gocd-windows-tray-build-notifier (which is about 10% complete) - which i would want to:
* show a mini view of the pipeline so we can see the pipeline in context
* only show toast notifications if its relevant to the current user ("only show builds caused by my commits" kind of thing

So, seems the xml snippet above seems pretty good for this use case (assuming the build cause is there).

srinivasupadhya:
> very scary that it has the material password in the xml!

@matt-richardson yes. The password & every field of material configuration will be available to plugin. Its to enable plugin authors to use the data. use-case: i would like to update GitHub PR status (building,passed,failed). To do this I would would need the whole configuration (url, username, password). I was thinking of adding a warning message on plugins listing page saying "notification plugins get access to all material configuration including the passwords". Not sure if its enough.

arvindsv:
Yeah, I don't think the password should be there. Let the plugin author / users figure out another way to update that information. For instance, github provides API keys for such things. I don't think passwords should be provided. At the end of the day, a plugin runs inside the Go Server and can be malicious. I understand that, and probably there's nothing to stop a determined plugin from causing havoc, even if we do not provide the password.

However, let's remember that this is a notification plugin endpoint and design it towards that. Not towards a specific use-case (GitHub PR status).

common/src/com/thoughtworks/go/config/materials/tfs/TfsMaterial.java
srinivasupadhya: i think this getter is useful. every materials every field is exposed with a getter apart from TFS - domain & project-path.

pshelton-skype:
From our experience using the Notification API, this covers several use cases - namely getting a list of materials that might have caused the stage failure and getting the job that failed the stage. 

We've also seen requests for a pipeline group property in addition to the pipeline name so we can more effectively support teams with multiple pipelines under the same group. Could we get that added as well?

srinivasupadhya:
Yes @ashwanthkumar has asked for it. I will add it for now. Lets see how the performance turns out.

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
srinivasupadhya: I have made it polymorphic for materials with an interface `InformationProvider`.

srinivasupadhya:
I have updated the payload JSON with latest output. Have a look & let me know your thoughts - @matt-richardson, @pshelton-skype, @ashwanthkumar

Done:
* removed secure fields (password), secure plugin properties & masked password in url fields
* added `pipeline-group`
* only the relevant stage is added to JSON as opposed to all stages of pipeline

Also:
* used another query that was well cached.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationPluginRegistry.java
srinivasupadhya: `isAnyPluginInterestedIn` is just doing `!notificationNameToPluginsInterestedMap.get('plugin-id').isEmpty()`. I don think it will cause concurrent modification exception. why do you think that would be the case? Do you think `notificationNameToPluginsInterestedMap` should be `ConcurrentHashMap` instead of just the `HashMap`?

srinivasupadhya:
Im thinking of getting rid of 
```json
"pipeline-name": "pipeline-name",
  "pipeline-counter": "1",
  "stage-name": "stage-name",
  "stage-counter": "1",
  "stage-state": "Passed",
  "stage-result": "Passed"
```
since its redundant now with `pipeline` -> `stage`

pshelton-skype:
I think I'm okay with it either way. It's easier to parse if I'm interested in this stage/status the first way, but not too much more difficult the proposed way.

matt-richardson:
I would prefer (for my use case at least) to have all stages, not just the current one. If its not provided, I'll have to call the history api and get the stages from that.

That said, I understand that you've got to balance load vs downstream use cases...

srinivasupadhya:
@pshelton-skype - I agree its easier. But the parsing from json to map will happen once. After that accessing data will be:
```java
responseMap.get("pipeline").get("name")
responseMap.get("pipeline").get("counter")
responseMap.get("pipeline").get("stage").get("name")
responseMap.get("pipeline").get("stage").get("counter")
```
instead of:
```java
responseMap.get("pipeline-name")
responseMap.get("pipeline-counter")
responseMap.get("stage-name")
responseMap.get("stage-counter")
```
looks ok.

@matt-richardson - yes by having to add just the stage whose status changed I get do reduce queries to be run. Its just one DB query to get `build-cause` for pipeline, so will keep it that way.
I think it should be ok for your use-case as well. You will know the status of other stages already (from previous notifications) & you will need to update the status for just this stage.

srinivasupadhya:
This is ready for final review.

matt-richardson:
@srinivasupadhya Unfortunately, that would mean I have to add a database (or some method of persistance) to my client, which I didn't have to when I had all the stages :(
Or, more likely, continue the approach of calling the history api (as i'm currently doing in https://github.com/matt-richardson/gocd-websocket-notifier).

ashwanthkumar:
@srinivasupadhya Is there a way get `trigger_message` equivalent (one from pipeline history API) in this response?

pshelton-skype:
If your code looks like
```
responseMap.get("pipeline").get("name")
responseMap.get("pipeline").get("counter")
responseMap.get("pipeline").get("stage").get("name")
responseMap.get("pipeline").get("stage").get("counter")
```
rather than 
```
ArrayList<LinkedTreeMap<String, Object>> materialRevisions = (ArrayList<LinkedTreeMap<String, Object>>)((LinkedTreeMap<String, Object>)pipeline.get("build_cause")).get("material_revisions");
```
you know something about using the `GsonBuilder.fromJson` functionality better than I do. What's the secret?

pshelton-skype:
@matt-richardson We're using the APIs extensively to collect more information on the stage, pipeline, and job, even with these changes.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationPluginRegistry.java
arvindsv: True. You're right. I think I was thinking of ```removePluginInterests``` and ```getPluginInterests```, but it looks like ```getPluginInterests``` is used only in tests. I think ```notificationNameToPluginsInterestedMap``` should at least be marked volatile, since it is being accessed from different threads. Otherwise, when a new plugin gets registered, the other threads asking ```getPluginsInterestedIn``` might not see that change.

matt-richardson:
@pshelton-skype, do you have a repo/gist/description of how you are consuming the api?

server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -16,49 +16,145 @@
...
+
+        return pipelineMap;
+    }
+
+    private List<Map> createBuildCauseDataMap(MaterialRevisions materialRevisions) {
+        List<Map> materialRevisionList = new ArrayList<Map>();
+        for (MaterialRevision currentRevision : materialRevisions) {
+            Map<String, Object> materialRevisionMap = new LinkedHashMap<String, Object>();
+
+            materialRevisionMap.put("material", ((InformationProvider) currentRevision.getMaterial()).getAttributes(false));
```
arvindsv: Oh, come on, a small change like this make the material a proper ```InformationProvider``` and doesn't need casts like this all over. You implemented the polymorphism change all over. Why wouldn't you just do this? It could even be marked abstract, since all the descendants of ```Material``` already implement it. Doing it this way just leads to someone seeing the ```Material``` interface, not implementing this method, and it'll fail with a ```ClassCastException``` at runtime, somewhere.

```
diff --git a/common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java b/common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java
index 9fe8fac..bfaef5c 100644
--- a/common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java
@@ -182,4 +182,9 @@ public abstract class AbstractMaterial extends PersistentObject implements Mater
     public MaterialConfig config() {
         throw new RuntimeException("You need to implement this");
     }
+
+    @Override
+    public Map<String, Object> getAttributes(boolean addSecureFields) {
+        return new HashMap<String, Object>();
+    }
 }
diff --git a/common/src/com/thoughtworks/go/domain/materials/Material.java b/common/src/com/thoughtworks/go/domain/materials/Material.java
index 9138ddd..7c21fa5 100644
--- a/common/src/com/thoughtworks/go/domain/materials/Material.java
+++ b/common/src/com/thoughtworks/go/domain/materials/Material.java
@@ -16,10 +16,6 @@
 
 package com.thoughtworks.go.domain.materials;
 
-import java.io.File;
-import java.io.Serializable;
-import java.util.Map;
-
 import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.materials.SubprocessExecutionContext;
@@ -29,7 +25,11 @@ import com.thoughtworks.go.util.command.EnvironmentVariableContext;
 import com.thoughtworks.go.util.command.ProcessOutputStreamConsumer;
 import com.thoughtworks.go.util.json.JsonMap;
 
-public interface Material extends Serializable {
+import java.io.File;
+import java.io.Serializable;
+import java.util.Map;
+
+public interface Material extends Serializable, InformationProvider {
 
     //-- DB and config behaviour

```

arvindsv:
Sorry to get in the middle of your conversation, @matt-richardson and @pshelton-skype. :) Just want to let @srinivasupadhya know that I've looked at this and it looks good to me.

Will wait on @srinivasupadhya to see if there's anything else missing (apart from a small ```InformationProvider``` related change), before merging.

srinivasupadhya:
@pshelton-skype - [here](https://github.com/gocd/go-plugins/pull/13/files#diff-860ed4034357b852c0372082937b1eb1R78) is an example of how the data will be used.

srinivasupadhya:
@arvindsv - there are a lot of implementors of `Material` interface in tests. I did not want to add this to all of them. Also, having a default implementation on top like in `AbstractMaterial` looks like empty `getAttributes()` is ok. Is that ok? I can do a `instanceOf` check before typecasting it i suppose?

arvindsv:
I think an empty one in ```AbstractMaterial``` is ok. In fact, I'd make that throw an exception, just like [this](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java#L183). That would prevent you from having to add that in hundreds of tests.

srinivasupadhya:
@ashwanthkumar `trigger_message` is built off of [build-cause](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/common/src/com/thoughtworks/go/domain/MaterialRevision.java#L204) or `approvalType` & `approvedBy` like [here](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/common/src/com/thoughtworks/go/domain/buildcause/BuildCause.java#L186). Build-cause is already available. Do you think `approvalType` & `approvedBy` info. should be available as well?

srinivasupadhya:
Looks like we are done!
Thanks a lot @matt-richardson, @pshelton-skype  & @ashwanthkumar for the feedback. The payload looks much better now :+1:

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/NotificationPluginRegistry.java
srinivasupadhya: Have made the `notificationNameToPluginsInterestedMap` a `ConcurrentHashMap`

dsenecalb:
@srinivasupadhya 

Is there a plan to include pipeline/stage/job environment variables into the extension point? If they're already there how to gather them. 
Thanks

PR #969, opened 2015-03-18
#496 - Ability to have short-lived sessions for API requests.

### LINKED ISSUE #496: Probable memory leak in go server
After about two weeks the go server java process consumes more than 1.5GB memory. Even though we have more than 50 pipelines this is quite a lot!

arvindsv:
When an API request starts a session, the idle time of that session is set to (a configurable value,
with default of) 300 seconds. This makes sure that sessions created due to API requests (usually
from scripts) time out, even if they do not send back the JSESSIONID cookie that the Go Server
sets, during every login.

Since this timeout affects only sessions started from API requests, it does not affect normal
functioning of the Go Server. Moreover, if a script does use the cookie and return it to the Go
Server, it continues to work, since the session will not be considered idle.

Added cctray.xml to this, as well (Mar 20, 2015). Checked that it works with CCMenu too.

Fixes #496.


arvindsv:
I've checked these scenarios:

#### 1. Session should timeout when it is an API request

```
$ curl -s -v http://localhost:8153/go/api/support 2>&1 | head -n 40
* Connected to localhost (::1) port 8153 (#0)
> GET /go/api/support HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: */*
>
< HTTP/1.1 200 OK
< Expires: Thu, 01-Jan-1970 00:00:00 GMT
< Set-Cookie: JSESSIONID=fj1tz11staq4b4ezfns34mt6;Path=/go;Expires=Wed, 01-Apr-2015 20:11:09 GMT
[... snip ...]

Timestamp:
==========
Wed Mar 18 16:11:09 EDT 2015

$ sleep 360; curl -b 'JSESSIONID=fj1tz11staq4b4ezfns34mt6' -s -v http://localhost:8153/go/api/support 2>&1 | head -n 40
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8153 (#0)
> GET /go/api/support HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: */*
> Cookie: JSESSIONID=fj1tz11staq4b4ezfns34mt6
>
< HTTP/1.1 200 OK
< Expires: Thu, 01-Jan-1970 00:00:00 GMT
< Set-Cookie: JSESSIONID=m7u0h2ock92ujs6tqcbdnpuu;Path=/go;Expires=Wed, 01-Apr-2015 20:17:24 GMT
< X-Frame-Options: SAMEORIGIN
[... snip ...]


Timestamp:
==========
Wed Mar 18 16:17:24 EDT 2015
```

Sending the old session ID after 300 seconds (chose 360 here), results in a response which sends back a new session ID (since the older one expired after 300 seconds).

#### 2. Session should not expire, when it is not idle

Ran the same curl command as above to find a session ID. Got ejydm4hfv2do1t13xf0dz0h0r. Then, ran this for 20 minutes:

```
while sleep 60; do clear; curl -b 'JSESSIONID=ejydm4hfv2do1t13xf0dz0h0r' -s -v http://localhost:8153/go/api/support 2>&1 | head -n 40; done
```

The reply was always:

```
> GET /go/api/support HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: */*
> Cookie: JSESSIONID=ejydm4hfv2do1t13xf0dz0h0r
>
< HTTP/1.1 200 OK
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-UA-Compatible: chrome=1
< Content-Disposition: inline; filename="serverinfo3015023739453643665txt"
< Content-Transfer-Encoding: binary
< Content-Type: text
< Cache-Control: private
< X-Request-Id: f539adf9-614f-4d9b-9a2c-5e966cceda3b
< X-Runtime: 0.085000
< Transfer-Encoding: chunked
* Server Jetty() is not blacklisted
< Server: Jetty()
<
{ [data not shown]
```

Shows that the server accepted that session, and did not return a Set-Cookie for a new session (the old session has not expired, since it is being used every 60 seconds).

#### 3. Toggling works

Ran:

```
$ curl -d toggle_value=off http://localhost:8153/go/api/admin/feature_toggles/api_requests_short_session_feature_toggle_key; echo
$ curl -s -v http://localhost:8153/go/api/support 2>&1 | head -n 40
*   Trying ::1...
* Connected to localhost (::1) port 8153 (#0)
> GET /go/api/support HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: */*
>
< HTTP/1.1 200 OK
< Expires: Thu, 01-Jan-1970 00:00:00 GMT
< Set-Cookie: JSESSIONID=1xkz4ap7ol7t31ews3xtbdip7e;Path=/go;Expires=Wed, 01-Apr-2015 22:31:38 GMT
[... snip ...]
$ sleep 360; curl -b JSESSIONID=1xkz4ap7ol7t31ews3xtbdip7e -s -v http://localhost:8153/go/api/support 2>&1 | head -n 40
```

The second curl, after 360 seconds did not return a new session ID, showing that the session was active beyond 300 seconds (existing behavior).

#### 4. Does not affect non-api requests

Tried with ```curl -s -v http://localhost:8153/go/``` to verify that it does not timeout the session, since it is a non-API request.

#### 5. Tried with a browser

Tried with a browser. Turned off auto refresh (URL: http://localhost:8153/go/admin/pipeline/new?group=defaultGroup&autoRefresh=false). Was able to refresh the page after more than 5 minutes, without having to re-login. Network tab of Chrome inspector showed that the old session ID continued to be used.

arvindsv:
Would be nice if this went in to 15.1. :) It is toggled too. I think it's going to help performance in many installations.

jyotisingh:
Haven't tested all scenarios, going by your word. The code looks alright to me so merging this.

arvindsv:
> going by your word

Going by my word?? That's never wise. :)

Alright. Hope I haven't made a mistake. As long as you think the toggle is fine (and will properly toggle), I think it's ok.

arikagoyal:
Uhmmm...:|

jyotisingh:
Alright, spent sometime testing it. Most things look fine to me so far apart from one.

Say I have this feature toggled off, I access some API, get a cookie. I toggle on the feature, wait for 5 or more mins, and access the API using the cookie that I got in previous step. The cookie had not expired, the same was used for the session. Its basically (!hadSessionBeforeStarting && hasSessionNow) code which causes that behavior, wondering in what situations would we need it? Shouldn't all API requests have a short-lived life, does it really matter if they have already been started?

arvindsv:
The differentiation between ```hadSessionBeforeStarting``` and ```hasSessionNow``` is useful for requests from a browser.

The ```!hadSessionBeforeStarting && hasSessionNow``` implies that this API request started a session. This is important to differentiate since we don't want web sessions to suddenly have their session timeout shortened because they made an API request.

This is the sequence I was trying to prevent:

1. Login: Session timeout is 2 weeks.
2. Access dashboard: Session timeout is 2 weeks.
3. Dashboard makes API call in background: Session is suddenly shortened to 5 minutes.
4. Close dashboard, come back in 6 minutes: Logged out.

Does that make sense?

In this case: ```! hadSessionBeforeStarting``` will not be true. So, it won't reduce the session time.

jyotisingh:
Ok, makes sense. Anyway, don't expect people to keep toggling the feature on and off. Closing this one. Verified on 15.1.0(1813-b9604e0416eff3)

PR #973, opened 2015-03-19
Read only mode

srinivasupadhya:


server/src/com/thoughtworks/go/server/security/ModeAwareFilter.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        if (shouldBlockRequest((HttpServletRequest) servletRequest)) {
+            LOGGER.info("Got a non-GET request: " + servletRequest);
```
arvindsv: LOGGER.info? Maybe LOGGER.warn?

arvindsv: Damn it. I introduced that change. :) Fine. I will fix it.

PR #974, opened 2015-03-21
Fix/properly detect init script name when in rc x.d

bernardn:
Fix a bug that prevents the agents to be automatically started at boot and stopped at shutdown/restart.

A SystemV init script is linked in /etc/rcX.d/ , where X is the runlevel, with its name being prepended with the letter S or K and a two-digit number, if the service has to be respectively started or stopped at this runlevel.

The extracted service name containing these systemV-specific characters didn't match the actual service name. They are now therefore removed if the script is run from one of the /etc/rcX.d/ directories.

### DIFF
--- a/installers/agent/release/init.cruise-agent
+++ b/installers/agent/release/init.cruise-agent
@@ -28,6 +28,9 @@
 ### END INIT INFO
 SERVICE_NAME=${0##*/}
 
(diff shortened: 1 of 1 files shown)


arvindsv:
Yes, this makes sense. Thank you, @bernardn! Sigh, I guess much around this area will need to change due to systemd winning.

ketan:
Verified and works on `15.1.0(1825-4f2609acf598d7)`

PR #977, opened 2015-03-24
Jetty 9 reincarnated

jyotisingh:
Now that JDK7 is installed on all build.go.cd agents, sending the pull request again. 
All commits are exactly the same apart from last two which are merging latest changes from master and fixing a few failing tests.


arvindsv:
Ok, I seem to be a sucker for this every time. Let's try it again.

PR #982, opened 2015-03-25
Scm fixes

srinivasupadhya:


server/webapp/WEB-INF/rails.new/app/controllers/admin/materials/pluggable_scm_controller.rb
```diff
@@ -166,6 +166,8 @@ def update(cruise_config)
 
           @pluggable_scm_service.validate(scm)
 
+          scm.clearEmptyConfigurations()
```
arvindsv: Is this about removing empty parts of the config? Meaning, fields which have been "emptied" by the user, in the config?

Is there a corresponding test for this? Is this why the optional-field was added, in the specs, in this commit?

srinivasupadhya: yes. `optional-field` would not be null if `clearEmptyConfigurations()` doesn't happen.

arvindsv: Ok.

srinivasupadhya:
Done as part of #818

Issue #410, opened 2014-08-04
Create delete artifacts extension point


pravds:
Whenever Go finds disk space to be lower than what is configured, it will try to delete artifacts for each stage run instance in their chronological order (unless stage is marked as 'never delete artifacts for this stage'). 

It would be useful to provide an extension point around this behavior so that plugins can be written to cater different artifact cleanup strategy.

aramprice:
+1 for this. One strategy I would hope to see supported is "never delete artifacts from the last successful build".

thanks

christrotter:
+1 for this.  We have a CI pipeline that rapidly fills up the disk.  When Go purges, all of the non-CI pipelines are dumped.  

Further (this is probably another issue...), when we re-run the necessary 'build' pipeline to re-create the artifact, it tends to still give 404s until I move the files around back into the specific pipeline run directory.  i.e. if run #2 contains the files, Go is looking for the stuff from run #1, and gives 404s until I copy the files in #2 into #1.  

So pretty much if there is a way to have it keep any artifacts that are being used (for example Dev has v3, QA has v2, UAT v1, Prod v1 - Go should keep v1, 2, and 3).

Alternatively, if there was a way to specify artifact retention per-pipeline, this could be alleviated.

rnjn:
We face the same issue, slow moving pipelines get their artefacts purged from Go, since the artefacts are months old. We need to rebuild them for new releases, for they are purged. We can put these artefacts out of Go, but that's not really a good solution. 
It would be great if we can keep last artefacts on all pipelines safe from the purge monster.

tomas-forsman:
Is this meant to be released any time soon?
Actually what we would need is #1207 - just keep the x latest artifacts form every pipeline

NachbarsLumpi:
We were just hit by this because go decided to delete the latest artifacts from several pipelines and thus our downstream pipelines began to fail with 404.

Since such a happening requires a lot of manual intervention, we would like this to be released with the next update.

jsilverman-blispay:
Is this released yet? Is it on the release roadmap? We currently use a find script to manually delete old artifacts, and have been considering writing something "smart" enough to keep X last artifacts of pipelines deployed to production.

Any progress on a smarter artifact cleanup strategy, built into Go.cd itself, would be awesome.

arvindsv:
[...]

PR #987, opened 2015-03-29
Artifact cleanup extension

pshivana:
Done as part of #410


config/config-server/resources/cruise-config.xsd
```diff
@@ -712,7 +721,12 @@
     </xsd:simpleType>
     <xsd:simpleType name="labelType">
         <xsd:restriction base="xsd:string">
+<<<<<<< HEAD
```
srinivasupadhya: needs fixing

zabil:
Closing this for now. Will have to re-implement the feature with support for plugin settings.

Issue #867, opened 2015-02-03
notification end-point


srinivasupadhya:
- [Discussion thread](https://groups.google.com/d/topic/go-cd/yziHSb1MGuw/discussion)
- [Dev Docs](http://www.go.cd/documentation/developer/writing_go_plugins/notification/notification_plugin_overview.html)
- [User Docs](http://www.go.cd/documentation/user/15.1.0/extension_points/notification_extension.html)
- sample plugin - [email notifier](https://github.com/srinivasupadhya/email-notifier)

arikagoyal:
Bug fixes done. Closing the story

PR #994, opened 2015-04-01
notification end-point fixes

srinivasupadhya:
notification should be delivered only when stage is scheduled & completed. currently its delivered on every job completion.

### DIFF
--- a/server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
+++ b/server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
@@ -54,11 +54,19 @@ public class StageStatusPluginNotifier implements StageStatusListener {
 
     @Override
     public void stageStatusChanged(final Stage stage) {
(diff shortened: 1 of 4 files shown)


server/src/com/thoughtworks/go/server/messaging/plugin/StageStatusPluginNotifier.java
```diff
@@ -54,13 +54,21 @@ public StageStatusPluginNotifier(NotificationPluginRegistry notificationPluginRe
...
             pluginNotificationQueue.post(new PluginNotificationMessage(NotificationExtension.STAGE_STATUS_CHANGE_NOTIFICATION, data));
         }
     }
 
+    private boolean isAnyPluginInterestedInStageStatus() {
+        return notificationPluginRegistry.isAnyPluginInterestedIn(NotificationExtension.STAGE_STATUS_CHANGE_NOTIFICATION);
+    }
+
+    private boolean isStageStateScheduledOrCompleted(Stage stage) {
+        return stage.isScheduled() || stage.getState().completed();
```
arvindsv: There's already stage.isCompleted().

arvindsv:
Make that one small change, please. I didn't think it made sense to wait for just that, since it is behaviorally same.

srinivasupadhya:
done as part of #867

Issue #996, opened 2015-04-02
Optimize plugins download on agents


mdaliejaz:
Today if the md5 of any plugin file changes the entire plugins Zip is downloaded on the agents. This download of the entire plugins zip happens on all the agents which can be avoided by downloading only the required plugins.

There should be a way to download only the newly added, removed or updated plugin.

mdaliejaz:
Sorry... looks like this is a duplicate of #121 
Closing this.

PR #997, opened 2015-04-02
#996 - Added listener to generate PluginsList.

mdaliejaz:
* A map is created for bundled and external plugins which is then used to create a JSON for all the plugins. This JSON will be passed to the agents which will compare with it's own local JSON contents. If there is any change, viz addition (new entry), deletion (entry no longer exists) or update (the md5 of the plugin entry changed), that particular plugin should be downloaded.


plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/DefaultPluginManager.java
```diff
@@ -139,6 +142,7 @@ public void startInfrastructure() {
     @Override
     public void registerPluginsFolderChangeListener() {
         monitor.addPluginsFolderChangeListener(pluginsZipUpdater);
+        monitor.addPluginsFolderChangeListener(pluginsListListener);
```
jyotisingh: I don't see a test for these monitors being added. Can you please write one.

Issue #1007, opened 2015-04-07
rename of plugin jar without change in content is not synced to agents


srinivasupadhya:

srinivasupadhya:
also saw:

```
java.io.FileNotFoundException: /Users/sriniup/projects/gocd.master/target/go-server-15.1.0/plugins_work/script-executor-0.2.jar (Is a directory)
        at java.io.FileInputStream.open(Native Method)
        at java.io.FileInputStream.<init>(FileInputStream.java:146)
        at com.thoughtworks.go.util.FileDigester.md5DigestOfFile(FileDigester.java:68)
        at com.thoughtworks.go.plugin.infra.commons.PluginsList$PluginEntry.<init>(PluginsList.java:98)
        at com.thoughtworks.go.plugin.infra.commons.PluginsList$PluginEntry.<init>(PluginsList.java:87)
        at com.thoughtworks.go.plugin.infra.commons.PluginsList$PluginEntries.initialize(PluginsList.java:80)
        at com.thoughtworks.go.plugin.infra.commons.PluginsList$PluginEntries.access$000(PluginsList.java:74)
        at com.thoughtworks.go.plugin.infra.commons.PluginsList.update(PluginsList.java:55)
        at com.thoughtworks.go.plugin.infra.listeners.PluginsListListener.handle(PluginsListListener.java:35)
        at com.thoughtworks.go.plugin.infra.monitor.DefaultPluginJarLocationMonitor$PluginLocationMonitorThread$DoOnPluginsFolderChangeListener$1.execute(DefaultPluginJarLocationMonitor.java:345)
        at com.thoughtworks.go.plugin.infra.monitor.DefaultPluginJarLocationMonitor$PluginLocationMonitorThread$DoOnPluginsFolderChangeListener.handle(DefaultPluginJarLocationMonitor.java:343)
        at com.thoughtworks.go.plugin.infra.monitor.DefaultPluginJarLocationMonitor$PluginLocationMonitorThread.loadAndNotifyPluginsFrom(DefaultPluginJarLocationMonitor.java:200)
        at com.thoughtworks.go.plugin.infra.monitor.DefaultPluginJarLocationMonitor$PluginLocationMonitorThread.run(DefaultPluginJarLocationMonitor.java:184)
```

ketan:
Verified, works on `15.1.0(1823-9db0d814df07de)`

PR #1009, opened 2015-04-07
#1007 - The plugins bundled location was wrong due to which the PluginsL...

mdaliejaz:
...ist was throwing the error while calculating the md5.


ketan:
Fixes #1007

PR #1013, opened 2015-04-09
#831 - Added a check to avoid NullPointerException.

mdaliejaz:


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/pluggabletask/PluggableTaskConfigStore.java
```diff
@@ -18,8 +18,8 @@
 
 import com.thoughtworks.go.plugin.access.config.PluginPreferenceStore;
 
-public final class PluggableTaskConfigStore extends PluginPreferenceStore<TaskPreference> {
```
pshivana: Any reason why final keyword is removed here?

arvindsv:
My guess is that the class is being mocked in the TaskViewServiceTest. Can't mock when it is final. Classes shouldn't be final, IMHO.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/pluggabletask/PluggableTaskConfigStore.java
arvindsv: My guess is that the class is being mocked in the TaskViewServiceTest. Can't mock when it is final. Classes shouldn't be final, IMHO.

pshivana: Alright

mdaliejaz: As @arvindsv mentioned, I needed to mock and hence had to remove final keyword.
Sorry for not replying earlier.

PR #1020, opened 2015-04-10
#930 - Fix timestamp issue.

### LINKED ISSUE #930: Timestamp issue
The Go server and agent are in different time zones. See the console output below. Note there is difference in time here. To me it appears that the time on the right hand side is coming from server. Needs to be checked

timestamp_issue

arvindsv:
Earlier, since dateFormat was a field, and BuildWork instance is created on the server side,
the time zone inside the SimpleDateFormat was set to that of the server. Removed it from
being a field. So, time zone is that of the agent now.

Fixes #930.


ketan:
verified - looks good to me.

PR #1021, opened 2015-04-13
notification end-point: notify plugin when stage is re-scheduled

srinivasupadhya:


srinivasupadhya:
this is ready for review.

ketan:
Fixes #1022

srinivasupadhya:
done as part of #867

PR #1025, opened 2015-04-14
Additional fixes for #973 - Read only mode.

ketan:
This will ensure that in addition to GET requests,
HEAD requests are treated as "read only" and will
pass through (not get filtered) to the application.


arvindsv:
First of many! :)

Issue #980, opened 2015-03-25
Write the console log to a temporary directory on go server


zabil:
The go agent posts the console output to go server who saves it on the artifacts directory setup in the go configuration.

If the artifacts directory is a network drive this might be a problem (performance). Instead, write the console on the server and move it to the artifacts directory when the job completes.

zabil:
@arvindsv @jyotisingh @srinivasupadhya @mdaliejaz 

Appreciate it if you look at these bunch of changes before I send in the final pull request.
A summary of what's being done.

* Write the console.log to a temporary directory on the server under pipeline/stage/counter
* Listen to the change in the job status being completed.
* While the job is running serve the console file from the temporary folder.
* Once the job is being completed serve the log from the final artifacts folder.

The current code incorrectly notifies of job completion so had to change the order in console publisher. 
I am still struggling with figuring out how to clean up the temp directory because deleting the root folder (of a pipeline/stage) will affects jobs that run in parallel.

As of now I am thinking of a separate thread that runs low frequency and deletes empty folders. 
Is there any reason why we are missing a pipeline completed status or listener?

arvindsv:
For the others, you can see all the commits in one place [here](https://github.com/gocd/gocd/compare/master...zabil:tempdir).

Thoughts:

1. The movement of reportCompleted to be after reportCompletedAction. Can you dig into the history of the order of those two for a bit. I remember working on this with Sara a long time ago, where we decided on why the order should be what it is. It had something to do with what will happen if an agent (or n/w communication) fails after one of them happens. One of them leaves the DB in a state which needs a reschedule. One doesn't. Long time. I might be mistaken.

2. You've checked in ```server/config/cruise-config.xml```. It has a path of yours, etc. I think this should be removed.

3. I haven't closely looked at the code, but I get the idea. The problem is because you're putting each console log file into its own directory. Does it need to be? Can it be stored in a file with a GUID name? Does it matter? One problem I see with using a random GUID for the name is if the server restarts in the middle of a job, then the file will be lost (not moved). So, can we hash the pipelines/%s/stage/%s/job/%s ... and come up with a (repeatable) value we can use for the console file of the job, instead of using a directory?

[...]

PR #1037, opened 2015-04-16
#980 Fixes to late writes on log console log and update of console files...

zabil:
... on job cancellation.

@arikagoyal @arvindsv 

These are a few regression fixes. As a side effect this now does some logic to calculate the right location of the log at a given point.

I might be making some changes to earlier fix of reporting job completion to the agent side to keep things consistent.

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/ArtifactsService.java
+++ b/server/src/com/thoughtworks/go/server/service/ArtifactsService.java
@@ -55,6 +55,11 @@ public class ArtifactsService implements ArtifactUrlReader {
     @Autowired
     public ArtifactsService(JobResolverService jobResolverService, StageDao stageDao,
                             ArtifactsDirHolder artifactsDirHolder, ZipUtil zipUtil, SystemService systemService) {
+        this(jobResolverService, stageDao, artifactsDirHolder, zipUtil, systemService, new ArtifactDirectoryChooser());
+    }
+
+    protected ArtifactsService(JobResolverService jobResolverService, StageDao stageDao,
+                               ArtifactsDirHolder artifactsDirHolder, ZipUtil zipUtil, SystemService systemService, ArtifactDirectoryChooser chooser) {
         this.artifactsDirHolder = artifactsDirHolder;
         this.zipUtil = zipUtil;
         this.jobResolverService = jobResolverService;
@@ -62,7 +67,7 @@ public class ArtifactsService implements ArtifactUrlReader {
         this.systemService = systemService;
 
         //This is a Chain of Responsibility to decide which view should be shown for a particular artifact URL
-        this.chooser = new ArtifactDirectoryChooser();
+        this.chooser = chooser;
 
     }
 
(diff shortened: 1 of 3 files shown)


server/src/com/thoughtworks/go/server/service/ArtifactsService.java
```diff
@@ -390,7 +401,7 @@ private boolean deleteFile(File file) {
     }
 
     public void appendToConsoleLog(JobIdentifier jobIdentifier, String text) throws IllegalArtifactLocationException, IOException {
-        File file = findArtifact(jobIdentifier, ArtifactLogUtil.getConsoleLogOutputFolderAndFileName());
```
arvindsv: I think it's time to start thinking about a ```ConsoleArtifactService```. This ```ArtifactService``` is too big, and some methods like ```findConsoleArtifact``` and ```temporaryConsoleFile``` are too confusing to tell apart. However, I won't hold this PR back because of this.

arvindsv: Also, there is almost implicit knowledge hidden in some places, right? Like the fact that ```chooser.findArtifact``` can be costly if the artifact store is on the network. So, try to find the temporary console log location first, rather than trying both, etc.

server/src/com/thoughtworks/go/server/service/ArtifactsService.java
```diff
@@ -156,8 +161,14 @@ public void moveConsoleArtifacts(LocatableEntity locatableEntity) {
         }
     }
 
-    public File temporaryConsoleFile(JobIdentifier jobIdentifier) {
-        return chooser.temporaryConsoleFile(jobIdentifier);
+    public File temporaryConsoleFile(JobIdentifier jobIdentifier) throws IllegalArtifactLocationException {
```
arvindsv: ```temporaryConsoleFile``` is no longer just temporary now. If and when we move this out to a different service, we can think of a more appropriate name?

server/src/com/thoughtworks/go/server/service/ArtifactsService.java
```diff
@@ -55,6 +55,11 @@
     @Autowired
     public ArtifactsService(JobResolverService jobResolverService, StageDao stageDao,
                             ArtifactsDirHolder artifactsDirHolder, ZipUtil zipUtil, SystemService systemService) {
+        this(jobResolverService, stageDao, artifactsDirHolder, zipUtil, systemService, new ArtifactDirectoryChooser());
```
arvindsv: This can be fully autowired, I suppose. There's nothing special in ```ArtifactDirectoryChooser``` that I can see.

server/src/com/thoughtworks/go/server/service/ArtifactsService.java
zabil: Yeah, changing the order will unintentionally break performance. I think we should either have a test explaining and verifying the order or should have methods that explain it.

server/src/com/thoughtworks/go/server/service/ArtifactsService.java
zabil: I am changing it to just consoleLogFile.
The service calls the chooser which branches out to temporaryConsole and findArtifact

server/src/com/thoughtworks/go/server/service/ArtifactsService.java
zabil: Did not autowire it because it in the .view. package (which is not scanned by spring).
A usage check shows that it's only used in the service, will try moving it to service package.
Unless there's some weird rails wiring which will be evident when I test the UI.

PR #1017, opened 2015-04-09
#000 - Remove Java 6 dependency from deb installers.

arvindsv:
It needs Java 7 now. Verified after building installers.

PR #1038, opened 2015-04-16
#1035 - Fix installer upgrade bug: JAVA_HOME.

### LINKED ISSUE #1035: go-server on debian refuses to boot up after upgrade
@arvindsv 

Upgrade fails on a fresh ubuntu/trusty64 machine (https://atlas.hashicorp.com/ubuntu/boxes/trusty64)

``` shell
#install 14.4
apt-get install -y openjdk-6-jre unzip
dpkg -i go-server-14.4.0-1356.deb
ps axu | grep go 
 
# upgrade to 15.1
apt-get install -y openjdk-7-jre-headless
sudo dpkg -i go-server-15.1.0-1780.deb
 
# verify
cat /etc/default/go-server
```

arvindsv:
* Set JAVA_HOME if it has not been set previously by the Go Server installer.
* Update JAVA_HOME if it has been set previously.
* Allow overriding this, by exporting JAVA_HOME after the Go Server installer line.

Fixes #1035.

### DIFF
--- a/installers/server/rpm/go-server.spec
+++ b/installers/server/rpm/go-server.spec
@@ -62,14 +62,13 @@ if [ "$1" = "1" ] ; then  # first install
         fix_server_log4j_properties
     fi
 
-    fix_server_java_home_declaration_for_rpm
-
     fix_server_defaults_ownership
     create_necessary_server_directories_and_fix_ownership
     
     chkconfig --add go-server
 fi
 
+fix_server_java_home_declaration_for_rpm
 fix_go_server_lib_ownership
 go_owned /etc/go
 
--- a/installers/shared/shared_post.sh
+++ b/installers/shared/shared_post.sh
@@ -96,13 +96,24 @@ function fix_agent_defaults_ownership {
 }
 
 function fix_java_home_declaration {
-    grep -q -E '^((\s)*export(\s)+JAVA_HOME=)|^((\s)*JAVA_HOME=)' $1 || JAVA_HOME_SET=false
+    determine_java_home_path
+    if [ "$JAVA_HOME_PATH" = "" ]; then
+        echo "WARNING: Unable to determine JAVA_HOME. Not setting or updating it in \"$1\". You might need to set it manually, after this." >&2
+        return
+    fi
+
+    local file_to_update="$1"
+    local java_home_check_pattern='^\s*export\s+JAVA_HOME=.*# SET_BY_GO_INSTALLER__DONT_REMOVE'
+    local final_java_home_line="export JAVA_HOME=\"$JAVA_HOME_PATH\" # SET_BY_GO_INSTALLER__DONT_REMOVE"
+
+    grep -q -E "$java_home_check_pattern" "$file_to_update" || JAVA_HOME_SET=false
     if [ "$JAVA_HOME_SET" = false ]; then
-        determine_java_home_path
-        if [ "$JAVA_HOME_PATH" != "" ]; then
-            echo "JAVA_HOME=$JAVA_HOME_PATH" >> $1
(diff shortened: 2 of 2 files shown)


arvindsv:
After this fix, I built an installer using:

```
docker run -i -t -v `pwd`/installers:/installers -e REPO=https://github.com/arvindsv/gocd.git -e BRANCH=1035_fix_debian_installer_upgrade gocd/gocd-build-installer deb rpm
```

#### Debian

I tried upgrading from 14.4.0 to 15.1, as mentioned by @ketan [here](https://github.com/gocd/gocd/issues/1035#issue-68908567).

First time, during an upgrade from 14.4.0 to 15.1, it says:

```
Found Java /usr/lib/jvm/java-7-openjdk-amd64/jre in PATH, using it.
Appending to "/etc/default/go-server": export JAVA_HOME="/usr/lib/jvm/java-7-openjdk-amd64/jre" # SET_BY_GO_INSTALLER__DONT_REMOVE
```

To simulate an upgrade from 15.1 to a higher version, I changed the line to use java-6-openjdk and upon removing the go-server package and installing it, it says:

```
Found Java /usr/lib/jvm/java-7-openjdk-amd64/jre in PATH, using it.
Current line with JAVA_HOME in "/etc/default/go-server" is: export JAVA_HOME="/usr/lib/jvm/java-6-openjdk-amd64/jre" # SET_BY_GO_INSTALLER__DONT_REMOVE
Updating it to: export JAVA_HOME="/usr/lib/jvm/java-7-openjdk-amd64/jre" # SET_BY_GO_INSTALLER__DONT_REMOVE
```

So, it seems to work. Am I missing anything else?

#### RPM

From what I can [see](https://github.com/gocd/gocd/blob/master/installers/server/rpm/go-server.spec#L65), with RPM, the JAVA_HOME was set only during first installation, and never updated.

I've changed that now. It tries to update it (same as Debian) if it is not found to be set, or is out of date. Same as DEB. The upgrade message looks like this:

```
Found Java /usr/lib/jvm/java-1.7.0-openjdk-1.7.0.79-2.5.5.1.el7_1.x86_64/jre in PATH, using it.
Current line with JAVA_HOME in "/etc/default/go-server" is: export JAVA_HOME="/usr/lib/jvm/java-1.6.0-openjdk-1.6.0.35.x86_64/jre" # SET_BY_GO_INSTALLER__DONT_REMOVE
Updating it to: export JAVA_HOME="/usr/lib/jvm/java-1.7.0-openjdk-1.7.0.79-2.5.5.1.el7_1.x86_64/jre" # SET_BY_GO_INSTALLER__DONT_REMOVE
```

arvindsv:
Also, if you really want to override whatever is set by this, all you need to do is add an ```export JAVA_HOME=...``` line at the end.

jyotisingh:
/cc @arikagoyal 

Alright, /etc/default/go-server generated by 14.4 would add an entry for JAVA_HOME (with export in next line and without the set-by-Go comment). Upgrade to 15.1 will add another entry for JAVA_HOME now. So, basically I will have the file with these entries:

```properties
JAVA_HOME=/usr/lib/jvm/jre-1.7.0-openjdk.x86_64
export JAVA_HOME
export JAVA_HOME="/usr/lib/jvm/jre-1.7.0-openjdk.x86_64" #SET_BY_GO_INSTALLER__DONT_REMOVE
```

It is harmless, but there would be multiple entries. @arikagoyal what do you think?

arikagoyal:
Had spoken to @jyotisingh. Fine by me

ketan:
Tested, verified works on `15.1.0(1825-4f2609acf598d7)` running on CentOS-6.5.

Closes #1035 #1017

PR #1047, opened 2015-04-20
#1046 - Fix render issue with empty comment.

### LINKED ISSUE #1046: Materials detail view fails to render if commit message is empty
This at least appears to be the issue in testing the latest 15.1 build, 9db0d814df07de1dae0fcda841cb727132973a24. Using a TFS material, and a TFS default checkin policy of allowing empty commits, causes the detail view page: http://servername/go/tab/build/detail/pipelane/20/stage/1/job to be blank.

Checking the server log shows the following backtrace:
```
2015-04-20 20:00:59,572 FATAL [qtp1168232895-31] Rails:? -
ActionView::Template::Error (undefined method `match' for nil:NilClass):
    190:             <dt><%= l.string('COMMENT:') -%></dt>
    191:             <dd>
    192:                 <div class='container'>
    193:                     <%= render_comment(__scope[:modification], __scope[:pipeline_name]) -%>
    194:                 </div>
    195:
    196:             </dd>
  app/helpers/materials_helper.rb:43:in `render_comment_markup_for'
  app/helpers/materials_helper.rb:51:in `render_comment'
  app/views/stages/stage.html.erb:193:in `_app_views_stages_stage_html_erb__1425660302_5638'
  app/views/stages/stage.html.erb:169:in `_app_views_stages_stage_html_erb__1425660302_5638'
  app/controllers/stages_controller.rb:137:in `render_stage'
  app/controllers/stages_controller.rb:136:in `render_stage'
  app/controllers/stages_controller.rb:21:in `overview'
``` 

which leads to me guess that the issue is just the missing comment on the material that triggered the pipeline. For now I have just implemented a checkin policy for comments

arvindsv:
Fixes #1046.

If the SHA for 15.1 is not yet decided (or there are going to be other commits), then this is a small fix that can be included.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/helpers/materials_helper.rb
+++ b/server/webapp/WEB-INF/rails.new/app/helpers/materials_helper.rb
@@ -46,17 +46,17 @@ module MaterialsHelper
   end
 
   def render_simple_comment(comment)
-    if comment.match(/\"TYPE\":\"PACKAGE_MATERIAL\"/)
+    if /\"TYPE\":\"PACKAGE_MATERIAL\"/.match(comment)
       package_comment_map = package_material_display_comment(comment)
       trackback_url = package_comment_map['TRACKBACK_URL'].blank? ? l.string('NOT_PROVIDED') : package_comment_map['TRACKBACK_URL']
       result = package_comment_map['COMMENT'] || "#{l.string('TRACKBACK')}#{trackback_url}"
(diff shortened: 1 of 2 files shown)


arvindsv:
@webframp has also mentioned that a checkin policy disallowing empty commits has been put in place. So, it's not affecting him at this point. So, can be accepted later as well.

Issue #1039, opened 2015-04-16
Builds not triggered on SVN external changes


thenathanjones:
Hey guys,

Our build currently makes use of SVN externals to simulate a single tree structure that we're looking to migrate to as part of a Gradle multi-project build.

We're finding a situation where code changes to the externals are not triggering a build. Changes to the top level repository are. If we manually trigger a build all the changes made to the externals are successfully pulled down.

We're currently using version 1.8.11 of Subversion and both agent/server are on CentOS 7.1.15, of which I have no control over. The version of Go we're currently using is 14.4.0.

Any ideas what it could be? My suspicion is an SVN version issue looking at the code, but would this make sense?

Nathan

arvindsv:
Have you set [checkExternals](http://www.go.cd/documentation/user/current/configuration/configuration_reference.html#svn) to true?

thenathanjones:
Yes the flag is set

arvindsv:
[This](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java#L97) is what it does, to get all external URLs, and then it does [this](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java#L134) on all those externals, to find out what has changed. I think you can try executing those, and see if something fails, maybe?

thenathanjones:
It appears to be because they use the ^ character for root relative paths
to the externals. I made a failing test to confirm, and have a fix, I just
want to try it out in a running system before I submit a pull request.
Thanks for the pointer to the code.

2015-04-18 4:38 GMT+12:00 Aravind SV <notifications@github.com>:

thenathanjones:
Hey guys, any update on when we can look at getting #1051 included? This and a few other bugs are really giving me grief at my current client but i'm not seeing it on any roadmap or had any feedback to the pull request.

zabil:
I believe testing manual testing is over in this. Functional twist tests need to be added?

PR #1048, opened 2015-04-21
Fix for relative root paths in SVN 1.5+ externals

thenathanjones:
As described in #1039 we have a file structure created by pulling multiple repositories together with SVN externals. We had an issue with builds not triggering when changes were made to the external repositories, only the root.

After looking through the code that parses SVN externals, my hunch was that it wasn't parsing the relative root '^' character in the URLs. I made a failing test using the output from running PROPGET and found the parser returned no repositories.

The code attached basically does a substitution on the '^' character with the root path, so it just goes through the regular parser logic. I've only added it to the 1.5 version of the parsers as that was when it was introduced.

### DIFF
--- a/common/src/com/thoughtworks/go/domain/materials/svn/SvnExternalParser.java
+++ b/common/src/com/thoughtworks/go/domain/materials/svn/SvnExternalParser.java
@@ -143,6 +143,10 @@ public class SvnExternalParser {
             }
         }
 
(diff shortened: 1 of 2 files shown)


thenathanjones:
The problem runs a little deeper with this one, it's taken this long to get a proper test environment working again. Will re-submit another pull request once I've done the extra changes

Issue #1049, opened 2015-04-21
API endpoint  to put Go server in active/passive mode without restart


pshivana:

pshivana:
It is possible to Go server in active/passive mode #972 , provide API endpoints to switch behavior without restarting Go server.

ketan:
Verified with `15.1.0-1845(ad23c4b2271c80)`

PR #1050, opened 2015-04-21
#1049 Added endpoints to put Go server in active/passive mode

pshivana:
...e current status


ketan:
Verified on `15.1.0(1839-0f15ee122cda40)`, this works as expected. I'm logging another ticket to improve the API endpoint.

PR #1051, opened 2015-04-21
Svn externals relative root fix

thenathanjones:
As described in #1039 we have a file structure created by pulling multiple repositories together with SVN externals. We had an issue with builds not triggering when changes were made to the external repositories, only the root.

After looking through the code that parses SVN externals, my hunch was that it wasn't parsing the relative root '^' character in the URLs. I made a failing test using the output from running PROPGET and found the parser returned no repositories.

The code attached basically does a substitution on the '^' character with the root path, so it just goes through the regular parser logic. I've only added it to the 1.5 version of the parsers as that was when it was introduced.

I think there is probably scope for doing some refactoring in the area in general, but given how niche my problem is it didn't seem worthwhile.

### DIFF
--- a/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
+++ b/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
@@ -302,6 +302,7 @@ public class SvnCommand extends SCMCommand implements Subversion {
     static class SvnInfo {
         private String path = "";
         private String encodedUrl = "";
+        private String root = "";
         private static final String ENCODING = "UTF-8";
 
 
@@ -324,6 +325,7 @@ public class SvnCommand extends SCMCommand implements Subversion {
             String encodedPath = StringUtils.replace(encodedUrl, root, "");
 
             this.path = URLDecoder.decode(encodedPath, ENCODING);
+            this.root = root;
             this.encodedUrl = encodedUrl;
         }
 
@@ -334,5 +336,7 @@ public class SvnCommand extends SCMCommand implements Subversion {
         public String getUrl() {
             return encodedUrl;
         }
+
+        public String getRoot() { return root; }
     }
 }
--- a/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
+++ b/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
@@ -99,10 +99,12 @@ public class SvnCommand extends SCMCommand implements Subversion {
                 .withArg(repositoryUrl);
         ConsoleResult result = executeCommand(svnExternalCommand);
         String svnExternalConsoleOut = result.outputAsString();
-        String repoUrl = remoteInfo(new SAXBuilder(false)).getUrl();
+        SvnInfo remoteInfo = remoteInfo(new SAXBuilder(false));
+        String repoUrl = remoteInfo.getUrl();
+        String repoRoot = remoteInfo.getRoot();
         List<SvnExternal> svnExternalList = null;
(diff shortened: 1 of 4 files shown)


arvindsv:
@thenathanjones: I think this looks pretty good. I have a couple of questions and thoughts:

* In [this documentation](http://svnbook.red-bean.com/en/1.6/svn.basic.in-action.html#idp5925248), it says:
  ```
  In Subversion 1.6, a new caret (^) notation was introduced as a shorthand for
  “the URL of the repository's root directory”. For example, you can use the
  ^/tags/bigsandwich/ to refer to the URL of the /tags/bigsandwich directory in
  the root of the repository. Note that this URL syntax works only when your
  current working directory is a working copy—the command-line client knows the
  repository's root URL by looking at the working copy's metadata. Also note that
  when you wish to refer precisely to the root directory of the repository, you
  must do so using ^/ (with the trailing slash character), not merely ^.
  ```

  So, it looks like it was introduced in 1.6. From your tests (assuming this worked for you), it looks like there is no [pattern change](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/svn/SvnExternalParser.java#L162) between 1.5 and 1.6. So, can you think about either creating an ```Svn16WithRootMatcher```  or renaming ```Svn15WithRootMatcher``` to something like ```Svn15AndAboveWithRootMatcher```?

* In the ```replaceRootRelativePathWithAbsoluteFor``` method, it replaces ^, even if it is not at the beginning of a path. I hope no one has a ^ in their real path.

* About the part mentioned in the end about ^ and ^/, I don't know if root directories usually end in a /. If they do, then replacing only ^ and not ^/ will mean that the replaced path will have two slashes. This should be alright in most cases, but something to think about.

thenathanjones:
@arvindsv thanks for taking a look mate. It's really proving a pain for us.

* The documentation I was following was [this](http://svnbook.red-bean.com/en/1.7/svn.advanced.externals.html) one which suggests it was added in 1.5. I'll happily rename it though if it makes it easier. 

* I can make it replace ones at the root only. Half of me believes that if you're silly enough to put a special character in your path you deserve it, but sure :)

* This is where the naming is a little confusing, the root as defined in the code, returned as part of the propget is not the actual "root" of the repo. From what I've seen, the root hasn't had a trailing slash. I'll double check my scenarios though.

arvindsv:
> From what I've seen, the root hasn't had a trailing slash. 

Then, it might be useful to replace the existing code with something like this:

```
protected String replaceRootRelativePathWithAbsoluteFor(String external, String repoUrl) {
    return external.replace("^/", repoUrl + "/");
}
```

Behavior remains the same (two slashes if it was going to be, before), but uses "^/" instead, which is what is in both the documentation pages referenced here.

thenathanjones:
@arvindsv Done. I've limited it to carets that aren't preceded by a slash or an alphanumeric character. Should someone really want to do anything other than that they're just asking for trouble. It also makes the regex more confusing as caret means something in regex. This seemed a decent compromise.

As part of that I match on the combo of "^/" and replace as per your suggestion. 

I also renamed the matcher to more more explicit about supporting versions beyond 1.5.

arvindsv:
Looks good! I've merged it. Made [one small change](https://github.com/gocd/gocd/commit/63458780e5316c6fb7020d49d626be2c7843edfa) after merge.

> ... makes the regex more confusing ...

You don't say. :) Lookbehind and all. Oh well, the price paid to handle input.

thenathanjones:
Haha. Thanks mate.

Issue #1045, opened 2015-04-20
Temporary console log issues


arikagoyal:
1) ~~Even though console logs are available in 'Artifacts' tab, yet i see following message there which should not be happen~~

~~"Artifacts for this job instance are unavailable as they may have been purged by Go or deleted externally. Re-run the stage or job to generate them again."~~

This is not an issue [@arikagoyal]

2) a) When a job is cancelled, its o/p looks like this -

[go] Job Completed Console-testing/8/defaultStage/1/job213:30:25.194 [go] Start to execute cancel task: Kills child processes
13:30:25.383 [go] Task is cancelled

Expected : "[go] Job Completed Console-testing/8/defaultStage/1/job2" should show up in the end of console log

b) Also on specifying job timeout, the console o/p looks like this -

[go] Job Completed Simple-console-testing/3/defaultStage/1/defaultJobGo cancelled this job as it has not generated any console output for more than 2 minute(s)10:15:54.829 [go] Start to execute cancel task: Kills child processes
10:15:54.891 [go] Task is cancelled

Expected o/p :

Go cancelled this job as it has not generated any console output for more than 1 minute(s)
[go] Start to execute cancel task: Kills child processes

[go] Task is cancelled

[go] Job completed test/4/defaultStage/1/defaultJob on f9c423f70d26 [/var/lib/cruise-agent] at Wed Apr 15 20:43:26 UTC 2015

Also at times when a job is cancelled, sometimes logs do not get moved to the final artifact location.

3) ~~If temporary console logs fails to make to the final artifact location due to disk space issue or permissions issue, they continue to lie in the temporary location forever. There should be a way to get these temp logs moved over to the actual location once the issue is resolved with the actual artifact location. Some automatic retry mechanism or on demand retry mechanism should be in place ?? Maybe in later releases ?~~ Moved this to #1057

arikagoyal:
Works fine on 15.1.0(1839-0f15ee122cda40)

PR #1052, opened 2015-04-22
#1045 Fixed to console logging issues

zabil:


arvindsv:
In one of these commits, the message ends in "To fix". Can you comment on the issue it fixes? I guess it is 1045. Just so that someone code spelunking sometime later will thank us. :)

zabil:
Strange. I wonder why message got truncated. 

It fixes reporting for cancelled jobs.

PR #1053, opened 2015-04-22
Do not close the stdout/error streams of the go server process

ketan:
Do not close the stdout/error streams of the go server process that the mac launcher executes.

When runnong the mac launcher on jdk8 - this seems to cause the
child process (the go.jar process) to crash out because it writes to stdout.

This fix will perform the equivalent of `java -jar go.jar 2>&1 > go-server.log`


ketan:
Verified on `15.1.0(1839-0f15ee122cda40)` running on `Mac OS X 10.10.3` on both JDK 7 and JDK 8.

PR #1054, opened 2015-04-22
Fixup copyright year for the go server and agent's "about box"

ketan:


ketan:
Verified on `15.1.0(1839-0f15ee122cda40)` running on `Mac OS X 10.10.3` on both JDK 7 and JDK 8.

PR #1059, opened 2015-04-23
Change git clean step to remove ignored files.

raztus:
The man page for git-clean says: "Don't use the standard ignore rules read from
.gitignore (per directory) and $GIT_DIR/info/exclude, but do still use the
ignore rules given with -e options. This allows removing all untracked files,
including build products."

Without this option, files from previous builds are persisted across runs. This can mask issues that aren't discovered until someone clones the repository fresh (or runs `git clean -fxd` locally).

### DIFF
--- a/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
+++ b/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
@@ -188,7 +188,7 @@ public class GitCommand extends SCMCommand {
     }
 
     private void cleanUnversionedFiles(File workingDir) {
-        String[] args = new String[]{"clean", "-dff"};
+        String[] args = new String[]{"clean", "-dffx"};
         CommandLine gitCmd = git().withArgs(args).withWorkingDir(workingDir);
         runOrBomb(gitCmd);
     }


arvindsv:
Hello! I closed [another PR](https://github.com/gocd/gocd/pull/860) which did just this, a little while ago. I feel that the change in (default) behavior is a little too much, given that there are others who use the .gitignore files exactly this way (to keep files across builds).

Of course, to get this behavior a task can be added to do it (a little cumbersome, I know). But, there's no coming back from removing ignored files.

Thoughts?

raztus:
Sorry, I searched open issues to make sure I wasn't duplicating, but should have included closed issues!

While some users may be relying on current behavior (others unwittingly so), I believe many more will expect Go to do a "proper" clean across builds, so that builds are guaranteed to be independent. I understand the rationale for not changing the default behavior, so my recommendation would be to include an option on the Git Material page that is selected by default that says something like "don't remove ignored files" to clarify the default behavior and make it easy to change.

ketan:
@arvindsv I just submitted another PR to this effect and noticed this as well.

Using this "feature" to keep files across builds is dangerous, unpredictable and brimming with potential errors (heisenbugs) for end users that may be hard to explain (short of running a build using a truly clean copy).

If users really need to keep files across builds, they should either be using some temp directory, or artefacts to store files.

But I see your point about this being a potentially breaking change. If I may, I would like to propose a feature toggle that will allow users to turn back to old behaviour `git clean -dff` in the interim while they migrate to other means to persist their files across builds.

Thoughts?

arvindsv:
Yes, I think the feature toggle could make sense. One thing to remember is that this is usually related to the agent side. Feature toggles don't really go over to the agent side, the way they are setup today.

If we do this, then for it to be effective, a user should be able to toggle it across all agents from the server side. This is tricky, but doable.

arvindsv:
What I said [here](https://github.com/gocd/gocd/pull/1155#issuecomment-106845189) is relevant to this kind of feature toggle, which needs to be toggled on both server and agents.

gprayback:
I agree with ketan, by default, the system should not persist any file across builds if you told it to clean up.  Having files persist across builds obscures potential problems from you.  A build may not be functioning correctly, but still passing because a required file was left over from a previous build (as opposed to being pulled or generated by the new build).  I'd unquestionably prefer to have builds fail because I forgot to persist some file in a temp directory or something, than have them pass incorrectly because a file was inadvertently left over.

tiwoc:
I just closed my PR #1437 which was a duplicate of this one. Here's my original comment on how the current default lead to a day lost to debugging some pipeline artifacts:

> This has bitten me hard when trying to let gocd build cURL, which uses an ignored build subdirectory. Some remnants of old builds that survived the cleaning led to a subtly broken library.

ketan:
@wpc Since you worked on the shallow clone, any thoughts on this?

wpc:
@ketan: My thoughts is: the last thing I want would be heisenbugs. It could eat up me hours to fix a tricky build failure because I will never expect clean has this behavior.

Regarding give user a way to revert back to old behavior: how about using a special named environment variable instead of feature toggle? This way we do not need passing yet another thing to the agent and complicate the already complex contract between server and agent.

And just curious: what is the second 'f' do in exiting '-dff'?

raztus:
From `man git-diff`:

> Git will refuse to delete directories with .git sub directory or file unless a second -f is given.

In other words, it allows `git clean` to remove git submodules.

wpc:
@raztus  Good to know. Thanks!

arvindsv:
I've been changing my approach to any (mostly breaking) change to be:

1. Toggle
2. Watch
3. Remove toggle and feature

This way, it allows everyone to go ahead with a potentially breaking change quick, without worrying too much about existing users' expectations. If the new behavior feels wrong to existing users, they'll tell us, and we can have a conversation - and tell them how to temporarily turn the new behavior off, while the conversation is going on. If no one has a problem, we remove code and move on.

In this case, there doesn't seem to be an easy way to toggle on the agent. Even with a special environment variable (I don't see it as any different from a toggle. Anything that allows behavior to be toggled is a feature toggle to me), it will need to be set on each agent separately. That seems to be the bigger problem here.

I'm now all for making the clean work properly, by the way (and removing ignored files).

wpc:
> Even with a special environment variable (I don't see it as any different from a toggle. Anything that allows behavior to be toggled is a feature toggle to me), it will need to be set on each agent separately.

With special environment variable approach, user can set it with the pipeline config.

arvindsv:
Oh, ok, that "special". :) It feels a little unclean to involve a user in that way, but I suppose it's an easier option to move forward. If others agree too, then alright.

gprayback:
I second the idea of something the user could override in the pipeline config and/or in an Ant/NAnt script, something like an environment variable. The issue I see is that in a multi-tenant environment where many teams are using the same agent pool, each team may want different behavior. So letting you set the toggle for each pipeline run seems like a good choice to me.

zabil:
Closing this as there are conflicts.

```GitCommand``` has changed a lot in the past releases. 

Feel free to bring this back from cold storage with a feature toggle.

PR #1070, opened 2015-04-28
#980 Jetty 6 deletes the work directory on exit. 

zabil:
Jetty 9 does not. Since we'll be switching between jetty 9 and jetty 6 for atleast another release it's safer to move this out of work to a top level dir and migrate jetty's work folder or temp folder under that in the next release.

### DIFF
--- a/server/src/com/thoughtworks/go/server/view/artifacts/ArtifactDirectoryChooser.java
+++ b/server/src/com/thoughtworks/go/server/view/artifacts/ArtifactDirectoryChooser.java
@@ -75,7 +75,7 @@ public class ArtifactDirectoryChooser {
     }
 
     public File temporaryConsoleFile(LocatableEntity locatableEntity) {
-        return new File("work/local", format("%s.log", DigestUtils.md5Hex(locatableEntity.entityLocator())));
+        return new File("data/console", format("%s.log", DigestUtils.md5Hex(locatableEntity.entityLocator())));
     }
 
 }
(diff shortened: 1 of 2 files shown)


server/src/com/thoughtworks/go/server/view/artifacts/ArtifactDirectoryChooser.java
```diff
@@ -75,7 +75,7 @@ public File findCachedArtifact(LocatableEntity locatableEntity) {
     }
 
     public File temporaryConsoleFile(LocatableEntity locatableEntity) {
-        return new File("work/local", format("%s.log", DigestUtils.md5Hex(locatableEntity.entityLocator())));
+        return new File("data/console", format("%s.log", DigestUtils.md5Hex(locatableEntity.entityLocator())));
```
srinivasupadhya: the folder structure `data/console` is created if it does not exist?

zabil: Yes at https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/ConsoleService.java#L115

and

https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/ConsoleService.java#L178

srinivasupadhya: cool

PR #1074, opened 2015-04-29
Simplify and bring jetty6's configuration in line with jetty9.

ketan:


ketan:
I'm going to close this PR in favor of #1101

Issue #1061, opened 2015-04-24
UTF-8 characters not getting displayed correctly when the job is running or when it is rescheduled


arikagoyal:
Note: This issue does not show up in raw o/p or console.log in artifacts tab

On console tab o/p shows up like this -

12:42:20.670 [go] Start to execute task: <exec command="/bin/bash" >
<arg>-c</arg>
<arg>echo "??? ??? ?? ???? ?? ????? ??"</arg>
</exec>.
12:42:20.672 [go] setting environment variable 'GO_SERVER_URL' to value 'https://10.4.3.121:8154/go/'
12:42:20.672 [go] setting environment variable 'GO_TRIGGER_USER' to value 'admin'
12:42:20.673 [go] setting environment variable 'GO_PIPELINE_NAME' to value 'Simple-console-testing'
12:42:20.673 [go] setting environment variable 'GO_PIPELINE_COUNTER' to value '19'
12:42:20.674 [go] setting environment variable 'GO_PIPELINE_LABEL' to value '19'
12:42:20.674 [go] setting environment variable 'GO_STAGE_NAME' to value 'defaultStage'
12:42:20.675 [go] setting environment variable 'GO_STAGE_COUNTER' to value '1'
12:42:20.675 [go] setting environment variable 'GO_JOB_NAME' to value 'defaultJob'
12:42:20.676 [go] setting environment variable 'GO_REVISION' to value '3ebe905103e40bbd1715645ce3c39560c66c8070'
12:42:20.677 [go] setting environment variable 'GO_TO_REVISION' to value '3ebe905103e40bbd1715645ce3c39560c66c8070'
12:42:20.677 [go] setting environment variable 'GO_FROM_REVISION' to value '3ebe905103e40bbd1715645ce3c39560c66c8070'
12:42:20.699 ??? ??? ?? ???? ?? ????? ??
12:42:20.798 [go] Current job status: passed.

zabil:
This went unnoticed in earlier builds and releases (14.4 etc). 
It happens during display of console logs while building, preparing etc. To fetch console logs chunks the GET /console.json does not have UTF-8 support.

chopraapooja:
This will be fixed when #1130 is accepted

chopraapooja:
This has been tested in v15.2.0(2029-7af6e1bfa43399). It seems working fine.

PR #1083, opened 2015-04-29
#1061 Set the content-type on header to render UTF-8 characters.

zabil:
Please note, I am not setting the content-type as application json as this appears to be handled by "X-JSON" some legacy prototype library usage.


ketan:
Can we please hold on this commit for some time?

I'm making some more tweaks to fix console log coloring that'll need to address this particular issue.

ketan:
Fixed as part of #1092

PR #1092, opened 2015-05-02
Console log coloring improvements and bug fixes (server side)

ketan:
See #1130 for an improved JS implementation

* Use a dark background by default, but allow switching to a white background
* Get rid of browser side console log parsing since it is only useful when "live tailing"
* Render the console log in color form the server in `text/html` instead of `text/plain`
* Change charset to utf-8
* The same mechanism that is used to render the live console log is used to render the log after the job is complete


server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
```diff
@@ -290,22 +290,63 @@ div.build_detail_container_content {
...
+
+    a {
+        padding-right: 15px;
+        color: yellow;
+    }
+
+    &.white-theme {
+        background-color: white;
+        a {
+            color: #1F66BD;
```
ketan: Does this even make sense? Why do we have a case sensitive nazi here?

arvindsv: Ha ha. Change it: [https://github.com/gocd/gocd/blob/master/.hound.yml](https://github.com/gocd/gocd/blob/master/.hound.yml). I don't think case should matter. I suppose it's the default so that it will be consistent.

arvindsv:
Can we rethink this whole page? I mean, the console tab. We have a chance to rewrite it "properly". The issues I see with this page are: It's written in Prototype in the front-end and uses hacks like X-JSON (for [this](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js#L62)) and has a Spring controller on the backend. I see you removed the iframe, and that's good.

Ideally, this should be written in something other than Prototype, should be a standard Rails controller, and should be written in a way that we can switch to a websocket, instead of the polling (independent of polling mechanism, I mean). Does that make sense?

If the general feeling among you and others is that we can get this in and then do that, I'm fine with it. But then, I or someone else, will need to take a closer look at this change.

thenathanjones:
Speaking as a user I'd be for both pushing ahead with this current implementation if it's an improvement and putting a plan in place for an improved version that does does a better job of making sense of the output. 

I'm stuck in a hard place at the moment with my current client, Go does so much better at visualising the build pipeline which is really important, but TeamCity does a much better job at being a CI server which is just as important.

ketan:
@arvindsv - Re-doing just the tab may involve a lot more than what this particular issue is about "console log coloring". This pull request does exactly what it advertises.

But I hear you about the additional improvements. I'm happy to open separate PRs for each of these (moving to rails, and later to server sent events (I'm sure that's what you meant by websockets))

arvindsv:
I'd meant websockets specifically, but SSE should be good enough, since we don't need a bidirectional connection and it's simpler. So, yes, SSE sometime later.

When I work on something like this, I try to move it over to what it should be, because I know other priorities come up and I'll never get a chance to fully rewrite it again. But, I'm fine with either approach, as I said. I'll look at this PR closer, soon. Maybe the other PRs will actually get done. :)

arvindsv:
caniuse.com tells me that EventSource has come a long way, and is supported quite well (except in IE). Didn't realize that. Websockets are supported in IE as well. Will need a polyfill or will need to take care of it somehow, for IE, since even IE 11 and IE Edge don't support EventSource.

ketan:
I see IE8 as a minimum requirement (http://www.go.cd/documentation/user/15.1.0/installation/system_requirements.html), is that correct?

zabil:
Getting rid of prototype is a quick win. But what are we going to use instead jquery?

ketan:
Yes jquery should be the way forward.

ketan:
Here's a sample script that I'm using to reproduce some coloring --

```bash
#!/bin/bash

echo "<script>console.log('42')</script>"

bash <( curl -s https://raw.githubusercontent.com/fidian/ansi/master/ansi) --color-table

#8 bit color table (256 colors)
echo -en "\n   +  "
for i in {0..35}; do
    printf "%2b " $i
done
printf "\n\n %3b  " 0
for i in {0..15}; do
    echo -en "\033[48;5;${i}m  \033[m "
done
#for i in 16 52 88 124 160 196 232; do
for i in {0..6}; do
    let "i = i*36 +16"
    printf "\n\n %3b  " $i
    for j in {0..35}; do
        let "val = i+j"
        echo -en "\033[48;5;${val}m  \033[m "
    done
done
echo -e "\n"

# something else

T='gYw'   # The test text
echo -e "\n                 40m     41m     42m     43m     44m     45m     46m     47m";
for FGs in '    m' '   1m' '  30m' '1;30m' '  31m' '1;31m' '  32m' '1;32m' '  33m' '1;33m' '  34m' '1;34m' '  35m' '1;35m' '  36m' '1;36m' '  37m' '1;37m';
    do FG=${FGs// /}
    echo -en " $FGs \033[$FG  $T  "
    for BG in 40m 41m 42m 43m 44m 45m 46m 47m;
        do echo -en "$EINS \033[$FG\033[$BG  $T \033[0m\033[$BG \033[0m";
    done
    echo;
done
echo

```

PR #1107, opened 2015-05-06
Bump version numbers and remove duplications where possible.

ketan:


plugin-infra/sample-plugins/curl-plugin-old-api-based/resources/plugin.xml
```diff
@@ -18,8 +18,8 @@
 <go-plugin id="curl.task.plugin.old.api.based" version="1">
     <about>
         <name>Curl Task Plugin - Old API Based</name>
-        <version>1.0</version>
-        <target-go-version>14.3.0</target-go-version>
+        <version>PLUGIN_VERSION_NUMBER</version>
+        <target-go-version>GO_VERSION_NUMBER</target-go-version>
```
srinivasupadhya: this should not be dynamically changed. this is the Go Server release it has been tested to work with. We bump it up when we are sure it works only with the newer release. So if `target-go-version` is 15.1.0 it means it wont work with 14.4.0 which isn't the case. Also because its API based  plugin it wont work in newer releases (once we drop support for Java API based plugins)

ketan: I'll undo this change. Is it ok to bump the plugin version itself to 15.2?

Issue #874, opened 2015-02-05
Stripping space characters from guid.txt when loading


gaqzi:
I'm working on a project where we have multiple environments coming together on one Go server. These environments and their agents needs to be configured in a very specific way, so to help with that we generate our own UUIDs based of the agent instance number and hostname. There's currently around 150 go agents and more are planned.

One thing we noticed is that when we save tour template file out editors would automatically add a newline. Which then gets transferred to Go and saved in the XML as an escaped newline character. This caused Go to start having issues dealing with the agents because the UUIDs were invalid.

Looking at the source it seems like guid.txt is taken and directly dumped into a variable without being cleaned. https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/common/src/com/thoughtworks/go/config/GuidService.java

We have worked around this by stripping away newlines from our guid.txt after creating it, but it would be nicer for this to cleaned at load time.

gaqzi:
Example of the XML config when there's a newline in guid.txt:

```diff
diff --git a/cruise-config.xml b/cruise-config.xml
index 6b59a45..5de1def 100644
--- a/cruise-config.xml
+++ b/cruise-config.xml
@@ -2044,6 +2044,7 @@
         <physical uuid="12f348ae-7381-4200-8e99-b95f319b63fa" />
         <physical uuid="61360a03-3f48-4ea8-8234-33f90ecdd1de" />
         <physical uuid="20ddf620-9845-5230-8aee-ef5f7e1bcbfe&#xA;" />
+        <physical uuid="16453a9a-9b36-53f6-8af5-4f52b3268223&#xA;" />
```

zabil:
Will close this issue once we have some testing done.

arikagoyal:
The current fix ensures that any whitespace that gets introduced in guid in future does not cause issues with Go's cruise-config and agent registration in any way. But what if someone has this issue in an older build and then they upgrade to 15.2. Should we take care of migration as well ?

ketan:
> But what if someone has this issue in an older build and then they upgrade to 15.2. Should we take care of migration as well?

@arikagoyal - I don't think it's particularly worth the effort here.

arikagoyal:
Ok i am fine with this not being done (i don't think it's a very high priority thing issue). If anyone feels otherwise we can create a separate issue and fix it as part of that. Closing this one.

PR #1108, opened 2015-05-06
Trim whitespace characters when loading agent GUID

gaqzi:
This correlates to an earlier bug report by me, #874, and this is my try at fixing that fairly minor issue.

I'm not a Java developer so I've basically muddled this together by looking at other tests, so I hope this'll fit with the Go style. :)


ketan:
Verified and this works as advertised.

arvindsv:
> I'm not a Java developer ...

I wouldn't have guessed. Good one! I think it fits the style. There's a missing copyright header at the top of the file. I didn't see it till after I merged it. Will add now.

Issue #1109, opened 2015-05-06
XSS vulnerability on user settings page


orfjackal:
Go Version: 15.1.0(1863-42e58c22aefd03)

### Steps to reproduce:

1. Go to user preferences
2. Edit email settings
3. Insert `<style>* { background-color: red; }</style>` to "My check-in aliases"
4. Save

### Expected result:

The code shows up in the UI as plain text, without changing the site's style.

### Actual result:

The page becomes red.

arvindsv:
Thank you! I've put in a fix. Will monitor, verify and then close this.

zabil:
Verified on 15.2.0-2022 (b1cad1dbd4794401c566acc3aa01a933daeb7c43)

PR #1110, opened 2015-05-06
#1109 - Fix escaping of values in prefs page.

arvindsv:
Email, matchers and filters are all escaped.


arvindsv:
Accepting it myself, since it is a fix for a security issue.

PR #1112, opened 2015-05-07
Updating JGit version to 3.6.2.201501210735-r

tedops:
All file permissions are not preserved after cloning from Git using GoCD (with JGit). This behavior has since been fixed in JGit v3.6.2.201501210735-r, according to https://bugs.eclipse.org/bugs/show_bug.cgi?id=424395, and so this pull request is to update the JGit version used by GoCD.

### DIFF
--- a/config/config-server/pom.xml
+++ b/config/config-server/pom.xml
@@ -51,7 +51,7 @@
         <dependency>
             <groupId>org.eclipse.jgit</groupId>
             <artifactId>org.eclipse.jgit</artifactId>
-            <version>2.0.0.201206130900-r</version>
+            <version>3.6.2.201501210735-r</version>
         </dependency>
 
         <dependency>


srinivasupadhya:
I think you should do a `./bn clean cruise:prepare SKIP_WAR=Y` and make sure it compiles. The API should have changed b/w 2.0 - 3.6
Also, we use JGit to manage internal `config.git` only (we use git cmd to manage repositories). Can you please elaborate on when & where you are seeing this behavior?

tedops:
Hi @srinivasupadhya - I created a [script to compare file permissions](https://github.com/tedops/perm_map/blob/master/perm_map.py) within two directories, and I ran it against a repo that was cloned via GoCD vs. cloned using Git on the CLI.

Also, I ran the same test using Atlassian Bamboo just to compare tools.

Here are the results:
* GoCD: https://gist.github.com/tedops/a45ba913f0e9ab1e710f
* Bamboo: https://gist.github.com/tedops/9b98ac45851fd8c57fae

This test shows that Bamboo accurately reproduces the permissions of all files in the codebase, while GoCD does not. This is because GoCD uses 1) an outdated JGit, and 2) does not allow the user to configure using the system Git. In this example, Bamboo was configured to use the system Git and not its default JGit.

tedops:
I've been working with @arvindsv in #gocd, and he helped me confirm that GoCD is NOT using JGit. The culprit here could be the default umask setting in supervisord that we've been using to run the go-agent.

But since the original reason for this PR is no longer valid, I'll close this.

EDIT: The supervisord umask was not the problem. We're still investigating why this is happening.

tedops:
Just to close the loop, the issue was finally fixed with some knowledge dropping by @arvindsv. Simply adding ``umask 002`` to the end of the agent config (``/etc/default/go-agent``) fixed the issue.

Issue #851, opened 2015-01-27
Authentication plugin end-point


srinivasupadhya:
There is quite some interest in Go users to use authentication scheme other than LDAP & File based. There is scope for many schemes to be implemented like OAuth - [1](https://groups.google.com/d/topic/go-cd/EZ09NfghHAs/discussion) & [2](https://groups.google.com/d/topic/go-cd/-YLb-61xhZs/discussion), 2-factor authentication - #34, SSO - #733 etc. Makes for a good end-point.

---

Support:

- [x] support web request handling
  - [x] ability to render view
  - [x] ability to redirect to a given URL
- [x] support required request processors
  - [x] ability to put/get/remove data from session
  - [x] ability to authenticate user (i.e. tell Go Server a user is authenticated & can be treated as a logged in user)
- [x] end-point
  - [x] ability to authenticate `username` & `password` (additional methods of authentication - other than LDAP & Password file)
  - [x] ability to search users by a search term

- [ ] To Do:
  - [x] Decide UI/UX for listing authentication plugins on login page & implement it 
  - [x] Decide if [SecurityConfig.isSecurityEnabled()](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/config/config-api/src/com/thoughtworks/go/config/SecurityConfig.java#L67) be true if ldap=false & password-file=false but authentication plugin exists
  - [ ] Decide behavior for error scenarios in username & password login when ldap, password file & plugins are involved. (code: `PluginAuthenticationProvider`, `AuthenticationProcessingFilter`). fix [Twist test](https://build.go.cd/go/tab/build/detail/regression/288/twist/1/firefox-split-runInstance-15)
  - [ ] Decide behavior for error scenarios in user search when ldap, password file & plugins are involved (code: `UserSearchService`)

---

[Developer Docs](http://www.go.cd/documentation/developer/writing_go_plugins/authentication/authentication_plugin_overview.html)
[Sample plugin](https://github.com/gocd/go-plugins/tree/master/plugins-for-tests/test-authentication-plugin)
[GitHub OAuth plugin](https://github.com/srinivasupadhya/gocd-oauth-login)

jyotisingh:
Moved the card back to awaiting development lane, until the bugs get fixed.

zabil:
Verified on 15.2.0(2170-05c70bb21bf9d3)

Closing this for now. 
[...]

PR #1116, opened 2015-05-07
#851 - Authentication end-point (first cut - RFC)

srinivasupadhya:
* add web-request handling capability to plugin

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
@@ -1,0 +1,62 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2014 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END***********************************/
+
+package com.thoughtworks.go.plugin.access.authentication;
+
+import com.thoughtworks.go.plugin.access.authentication.model.AuthenticationPluginConfiguration;
+import com.thoughtworks.go.plugin.api.GoPlugin;
+import com.thoughtworks.go.plugin.infra.PluginChangeListener;
+import com.thoughtworks.go.plugin.infra.PluginManager;
+import com.thoughtworks.go.plugin.infra.plugininfo.GoPluginDescriptor;
+import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import static org.apache.log4j.Logger.getLogger;
+
+@Component
+public class AuthenticationPluginRegistrar implements PluginChangeListener {
+    private static Logger LOGGER = getLogger(AuthenticationPluginRegistrar.class);
+
+    private AuthenticationExtension authenticationExtension;
+    private AuthenticationPluginRegistry authenticationPluginRegistry;
+
+    @Autowired
+    public AuthenticationPluginRegistrar(PluginManager pluginManager, AuthenticationExtension authenticationExtension, AuthenticationPluginRegistry authenticationPluginRegistry) {
+        this.authenticationExtension = authenticationExtension;
+        this.authenticationPluginRegistry = authenticationPluginRegistry;
+        pluginManager.addPluginChangeListener(this, GoPlugin.class);
+    }
+
+    @Override
+    public void pluginLoaded(GoPluginDescriptor pluginDescriptor) {
+        if (authenticationExtension.isAuthenticationPlugin(pluginDescriptor.id())) {
+            try {
+                AuthenticationPluginConfiguration configuration = authenticationExtension.getPluginConfiguration(pluginDescriptor.id());
+                authenticationPluginRegistry.registerPlugin(pluginDescriptor.id(), configuration);
+            } catch (Exception e) {
+                LOGGER.warn("Error occurred during plugin registration.", e);
+            }
+        }
+    }
+
+    @Override
+    public void pluginUnLoaded(GoPluginDescriptor pluginDescriptor) {
+        if (authenticationExtension.isAuthenticationPlugin(pluginDescriptor.id())) {
+            authenticationPluginRegistry.deregisterPlugin(pluginDescriptor.id());
+        }
+    }
+}
--- a/server/src/com/thoughtworks/go/server/controller/PluginController.java
+++ b/server/src/com/thoughtworks/go/server/controller/PluginController.java
@@ -1,0 +1,89 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2014 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END***********************************/
+
+package com.thoughtworks.go.server.controller;
+
+import com.thoughtworks.go.plugin.api.request.DefaultGoPluginApiRequest;
+import com.thoughtworks.go.plugin.api.response.DefaultGoApiResponse;
+import com.thoughtworks.go.plugin.api.response.GoPluginApiResponse;
+import com.thoughtworks.go.plugin.infra.PluginManager;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.View;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.HashMap;
+import java.util.Map;
+
+@Controller
+public class PluginController {
+    private PluginManager pluginManager;
+
+    @Autowired
+    public PluginController(PluginManager defaultPluginManager) {
+        this.pluginManager = defaultPluginManager;
+    }
+
+    @RequestMapping(value = "/{pluginId}/{requestName}", method = RequestMethod.GET)
+    public ModelAndView handleAboutRequest(
+            @PathVariable String pluginId,
+            @PathVariable String requestName,
+            HttpServletRequest request) {
+        DefaultGoPluginApiRequest apiRequest = new DefaultGoPluginApiRequest(null, null, requestName);
+        apiRequest.setRequestParams(getParameterMap(request));
+        GoPluginApiResponse response = pluginManager.submitTo(pluginId, apiRequest);
+        if (DefaultGoApiResponse.SUCCESS_RESPONSE_CODE == response.responseCode()) {
+            return renderPluginResponse(response);
+        }
+        if (302 == response.responseCode()) {
+            return new ModelAndView("redirect:" + response.responseHeaders().get("Location"));
+        }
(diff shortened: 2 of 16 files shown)


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
```diff
@@ -0,0 +1,62 @@
...
+                authenticationPluginRegistry.registerPlugin(pluginDescriptor.id(), configuration);
+            } catch (Exception e) {
+                LOGGER.warn("Error occurred during plugin registration.", e);
+            }
+        }
+    }
+
+    @Override
+    public void pluginUnLoaded(GoPluginDescriptor pluginDescriptor) {
+        if (authenticationExtension.isAuthenticationPlugin(pluginDescriptor.id())) {
```
arvindsv: I thought we can't make this call, since the plugin is unloaded.

server/src/com/thoughtworks/go/server/controller/PluginController.java
```diff
@@ -0,0 +1,89 @@
...
+public class PluginController {
+    private PluginManager pluginManager;
+
+    @Autowired
+    public PluginController(PluginManager defaultPluginManager) {
+        this.pluginManager = defaultPluginManager;
+    }
+
+    @RequestMapping(value = "/{pluginId}/{requestName}", method = RequestMethod.GET)
+    public ModelAndView handleAboutRequest(
```
arvindsv: Tsk, tsk. Copied from AboutController. :) Name needs to be changed.

server/src/com/thoughtworks/go/server/controller/PluginController.java
```diff
@@ -0,0 +1,89 @@
...
+            @PathVariable String pluginId,
+            @PathVariable String requestName,
+            HttpServletRequest request) {
+        DefaultGoPluginApiRequest apiRequest = new DefaultGoPluginApiRequest(null, null, requestName);
+        apiRequest.setRequestParams(getParameterMap(request));
```
arvindsv: I don't think we should send the whole map. It should be a structured (known) subset (or a subset which is prefixed by the plugin ID or something). Sometimes there's a lot of information that is there in it.

If the information there is only coming from the login page which the plugin controls, then it's fine. If there is other info that jetty puts there, we might want to look at getting rid of it before sending to the plugin.

ketan:
Why do we continue to write old skool controllers?

server/src/com/thoughtworks/go/server/security/providers/PluginAuthenticationProvider.java
```diff
@@ -0,0 +1,59 @@
...
+    protected void additionalAuthenticationChecks(UserDetails userDetails, UsernamePasswordAuthenticationToken authentication) throws AuthenticationException {
+    }
+
+    protected UserDetails retrieveUser(String username, UsernamePasswordAuthenticationToken authentication) throws AuthenticationException {
+        Set<String> plugins = authenticationPluginRegistry.getPluginsThatSupportsPasswordBasedAuthentication();
+        for (String pluginId : plugins) {
+            Result result = authenticationExtension.authenticateUser(pluginId, username, (String) authentication.getCredentials());
+            if (result.isSuccessful()) {
+                com.thoughtworks.go.plugin.access.authentication.model.User user = authenticationExtension.getUserDetails(pluginId, username);
+                return new GoUserPrinciple(user.getUsername(), String.format("%s %s", user.getFirstName(), user.getLastName()), "", true, true, true, true, authorityGranter.authorities(username));
```
arvindsv: Ha! I never realized it is called ```GoUserPrinciple```. I guess the idea was to call it ```GoUserPrincipal```.

server/src/com/thoughtworks/go/server/controller/PluginController.java
srinivasupadhya: this is spike code :D

server/src/com/thoughtworks/go/server/controller/PluginController.java
srinivasupadhya: its just plugin related params as far as i know. will double check.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
srinivasupadhya: same as [this](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/packagematerial/PackageMaterialMetadataLoader.java#L68)

server/src/com/thoughtworks/go/server/service/plugins/processor/AuthenticateUserRequestProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+public class AuthenticateUserRequestProcessor implements GoPluginApiRequestProcessor {
+    private static final String AUTHENTICATE_USER_REQUEST = "authenticate-user";
+    private AuthorityGranter authorityGranter;
+    private UserService userService;
+
+    @Autowired
+    public AuthenticateUserRequestProcessor(DefaultGoApplicationAccessor goApplicationAccessor, AuthorityGranter authorityGranter, UserService userService) {
+        this.authorityGranter = authorityGranter;
+        this.userService = userService;
+        goApplicationAccessor.registerProcessorFor(AUTHENTICATE_USER_REQUEST, this);
```
arvindsv: So, this is a processor for the plugin to use the GoPluginApiRequestProcessor interface and make a call to ```authenticate-user```, right? It looks like this is to allow "pre-authentication" of some users. What is the idea here?

If the idea is to pre-authenticate, can it be done through a specific pre-auth call from the Go Server, sometime early, during load of the plugin?

server/src/com/thoughtworks/go/server/security/providers/PluginAuthenticationProvider.java
srinivasupadhya: i dint get why its called any of that! but choose to stick with what was [here](https://github.com/gocd/gocd/blob/7958ac8b613031660ee00f0591af792c06c0f74c/server/src/com/thoughtworks/go/server/security/providers/FileAuthenticationProvider.java#L100)

server/src/com/thoughtworks/go/server/web/IgnoreResolver.java
```diff
@@ -76,6 +77,10 @@ private boolean isGoingToAbout(HttpServletRequest request) {
         return StringUtils.contains(request.getRequestURI(), request.getContextPath() + "/about");
     }
 
+    private boolean isGoingToPlugin(HttpServletRequest request) {
+        return StringUtils.contains(request.getRequestURI(), request.getContextPath() + "/plugins/interact");
```
arvindsv: Where did the "/interact" come from? Is ```{requestName}``` above supposed to be "interact"? If so, why use ```{requestName}```?

server/webapp/WEB-INF/vm/auth/login.vm
```diff
@@ -123,6 +123,11 @@
                 </div>
                 #end
                 <div class='clear-both'><!-- Clear floats --></div>
+
+#foreach ($authentication_plugin_id in $authentication_plugin_registry.getAuthenticationPlugins())
```
arvindsv: I guess we should think about the UI of this, as well. Should the administrator be able to set the default way of login?

server/src/com/thoughtworks/go/server/service/plugins/processor/AuthenticateUserRequestProcessor.java
srinivasupadhya: It is to `authenticate-user`, sample [usage](https://github.com/gocd/go-plugins/pull/14/files#diff-8285148299afe93bde971ae9ebb96a67R138).
dint get the pre-authenticate bit.

server/src/com/thoughtworks/go/server/service/plugins/processor/SessionDAORequestProcessor.java
```diff
@@ -0,0 +1,96 @@
...
+
+@Component
+public class SessionDAORequestProcessor implements GoPluginApiRequestProcessor {
+    private static final String STORE_IN_SESSION = "store-in-session";
+    private static final String GET_FROM_SESSION = "get-from-session";
+    private static final String REMOVE_FROM_SESSION = "remove-from-session";
+
+    @Autowired
+    public SessionDAORequestProcessor(DefaultGoApplicationAccessor goApplicationAccessor) {
+        goApplicationAccessor.registerProcessorFor(STORE_IN_SESSION, this);
```
arvindsv: More processors from plugin to Go, right? What is the idea for this? If this is about auth, why does the plugin need to access session? Is it for some state it *might* want? Would it not want to store that during authentication, rather than at some other time?

I'm sure there's a reason. :) Just trying to find out what the thought behind this is.

server/src/com/thoughtworks/go/server/security/providers/PluginAuthenticationProvider.java
arvindsv: :) Yeah. ```UserPrincipal``` is a concept in authentication. See [this](https://docs.oracle.com/javase/7/docs/jre/api/security/jaas/spec/com/sun/security/auth/UserPrincipal.html) and [this](https://msdn.microsoft.com/en-us/library/ftx85f8x(v=vs.110).aspx).

server/src/com/thoughtworks/go/server/web/IgnoreResolver.java
srinivasupadhya: `/plugins/interact` comes from [here](https://github.com/gocd/gocd/pull/1116/files#diff-52263857e788ae1666b5b0b9c5b552d9R210). `{requestName}` maps to what request name for plugin, like [here](https://github.com/gocd/go-plugins/pull/14/files#diff-8285148299afe93bde971ae9ebb96a67R81)

server/src/com/thoughtworks/go/server/service/plugins/processor/SessionDAORequestProcessor.java
srinivasupadhya: right. this is to store state across web requests like [1](https://github.com/gocd/go-plugins/pull/14/files#diff-8285148299afe93bde971ae9ebb96a67R108) & [2](https://github.com/gocd/go-plugins/pull/14/files#diff-8285148299afe93bde971ae9ebb96a67R121). Plugin couldn't have written to file etc. since the right file needs to be mapped back for same session.

server/webapp/WEB-INF/vm/auth/login.vm
srinivasupadhya: I thought about it. If we list all available authentication schemes on the default login page then the user can choose to use what he prefers. admins can prefer authentication schemes by configuring LDAP/Password/adding the right plugin. I suppose that much flexibility should be enough?

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
arvindsv: [Nope](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/packagematerial/PackageMaterialMetadataLoader.java#L75).

arvindsv: Same bug was fixed [here](https://github.com/zabil/gocd/commit/db430bf3ffe42861a8095ffcd46c4de3c7ea36ad).

server/src/com/thoughtworks/go/server/web/IgnoreResolver.java
arvindsv: Ok, so it is ```/plugins/interact/{pluginId}/{requestName}```. Alright. There was a valid question about old-school controllers, from @ketan. Any reason we cannot move all of this to Rails?

server/webapp/WEB-INF/vm/auth/login.vm
arvindsv: But, will Go's default login page be allowed to change? If not, why shouldn't it? Why shouldn't an admin be able to say, "All our auth has to happen through blah plugin. So, the default login page should redirect to that". Feasible? Makes sense, or not?

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistrar.java
srinivasupadhya: got it! will remove :+1:

server/src/com/thoughtworks/go/server/web/IgnoreResolver.java
srinivasupadhya: ya i forgot to reply to @ketan's question. its not listed on this page for some reason.

either way is fine by me. spring seemed easier to get through to for spike.

server/webapp/WEB-INF/vm/auth/login.vm
srinivasupadhya: yes, it does makes sense. But with the current login page & plugins showing up on that page (as buttons/links) it removes the requirement for plugin to handle welcome page. If we decide to go that way we will have to redirect to plugin (which should mandatorily handle that) along with a nice looking UI. Currently with OAuth, Single sign-on etc work off of redirects so there is no need to build a UI (for plugin dev.)

PR #1117, opened 2015-05-08
Remove some old cruft sitting there.

ketan:


arvindsv:
Can ```run_rspec_tests.rake``` be removed as well, then? /cc @srinivasupadhya

ketan:
> Can run_rspec_tests.rake be removed as well, then? /cc @srinivasupadhya

I think so, I looked around, but could not find references. I'd rename the `run_rspec_tests_use_new_rails.rake` to `run_rspec_tests.rake` as well.

arvindsv:
Yes, that makes sense to me. Might need to change usages in build.go.cd config.

srinivasupadhya:
sure :+1:

arvindsv:
I've also had my eye on idea-settings.jar and killableprocess.exe, but they can be in a different PR.

arvindsv:
Not merging this yet, till ```run_rspec_tests_use_new_rails.rake``` is removed, since there is a reference to old.go.jruby in it.

arvindsv:
Removed the old file. Renamed the use_new_rails bit. Changed it in all the configs of the Go Servers I know of (including build.go.cd).

Issue #1100, opened 2015-05-05
[RFC] - API improvements and guidelines


ketan:
This RFC is to gather feedback on fixing some of the inconsistencies with existing APIs (see #572) and establish guidelines for new APIs going forward. A lot of this is inspired from GitHub's APIs described at https://developer.github.com/v3/ with a combination of JSON HAL spec described at http://stateless.co/hal_specification.html.

API Endpoint
============

All APIs will be accessed from `https://go-server-url:8154/go/api`. All data will be sent and recieved as JSON. You may access the APIs over plain text, but for security reasons we suggest that users use SSL.

Authentication
==============

All APIs require you to authenticate yourself using your username and password. A cookie will also be returned as a result of the first authentication call. You may use this cookie for any further authentication attempts. If authentication fails, Go will return status `401`, `403` or `404`.

Basic authentication
--------------------

**Example:**

``` bash
curl -u 'username:password' -H 'Accept: application/vnd.go.cd.v1+json' https://build.example.com/go/api
```

**Response:**

```
HTTP/1.1 200 OK
Content-Type: application/vnd.go.cd.v1+json; charset=utf-8
Set-Cookie: JSESSIONID=JDQz5YSF6W4LzG9JuqPC7g;Path=/go;Expires=Tue, 19-May-2015 09:41:09 GMT
```

```json
{
  "pipelines_url": "https://build.example.com/api/pipelines",
  "users_url": "https://build.example.com/api/users",
}
```

Cookie/Session authentication
-----------------------------

Using the cookie/session returned from the previous API call, one can make further API calls. Using a cookie will dramatically improve performance of API calls especially if go is authenticating against an external source like LDAP.

``` bash
curl --cookie 'JSESSIONID=JDQz5YSF6W4LzG9JuqPC7g' -H 'Accept: application/vnd.go.cd.v1+json' https://build.example.com/go/api
```

**Response:**

```
HTTP/1.1 200 OK
Content-Type: application/vnd.go.cd.v1+json; charset=utf-8
Set-Cookie: session_id=some-session-id
```

```json
{
  "pipelines_url": "https://build.example.com/api/pipelines",
  "users_url": "https://build.example.com/api/users",
}
```

Versioning
==========

By default, all requests receive the **v1** version of the API. It is recommended to explicitly request this version via the `Accept` header.

```
Accept: application/vnd.go.cd.v1+json
```

If you specify an unsupported, `Accept` header, Go will respond with the current default JSON representation of the resource.

[...]

Issue #959, opened 2015-03-17
Api method for assigning resources and environments to agents


reidca:
Would be useful to have API methods to assign resources and environments to agents.

Looking at the documentation (http://www.go.cd/documentation/user/current/api/agent_api.html)  I don't see anyway to do this.

ketan:
I've picked this up to implement.

zabil:
Verified on experimental release15.2.0-2149 (d0369fb00f85d395f6534a3a434b7a612d286f7b)

Issue #958, opened 2015-03-17
API method for renaming agent


reidca:
The existing agent API does not allow agents to be renamed.

This can be achieved by modifying the hostname property in the XML configuration of the agent however this is entirely a manual task.

It would be great to get a new method added to the Agent Api to allow the agent to be given a semantic name.

#124

ketan:
I've picked this up to implement.

zabil:
Use a PATCH instead of PUT for doing this. 
More information here [June 3 2015 4:57 PM](https://gitter.im/gocd/gocd?at=556ee4bf463d0c7c066dd838)

ketan:
> Use a PATCH instead of PUT for doing this.

Opened #1183 to address this, also updating docs directly

ketan:
@zabil close #959 as well?

PR #1122, opened 2015-05-11
API improvements to manage agents

ketan:
This API is compliant with the RFC (#1100) specified at https://github.com/gocd/gocd/issues/1100

* API method for renaming agent (#958)
* API method for assigning resources to agents (part of #959)


ketan:
As a first cut, this is ready to review should someone want to look into it. Once this is merged in, I can work towards improving the documentation.

Here is what is missing in order to completely get rid of the old controller (http://git.io/vUr5g) -

* Bulk actions to manage agents (from the agents tab)
   This one is simple to implement, but I'd rather keep this PR small and make improvements.
* `job_run_history` api call. This one needs a bit of re-thinking on multiple fronts
  * is this valuable to have?
  * do we consider changing how pagination works (we currently suffer from the problem described here - http://www.sitepoint.com/paginating-real-time-data-cursor-based-pagination/)

Issue #124, opened 2014-04-23
Add configuration property to specify the agent name that will be registered with the server


reidca:
The names of the agent when registered with the GO server is the hostname (or is some cases the FQDN) that is obtained by Java. This not configurable however if the hostname itself does not contain any semantic meaning then this name if not very useful when assigning resources or environments.

It would be great if the agent has a configurable property that is used when first registering itself to the server. This could be specified on the install command line.

Additionally having a user interface option on the server to modify the agent name would be useful.

piyush0101:
:+1: 

Even if this is similar to the Jenkins Swarm Slaves, it would be quite useful. With swarm slaves, you can specify the name of the slave that is going to be registered to the Jenkins master.

reidca:
This is still a pain point for us. We have automated the install,. registration and configuration of GO Agents onto our workstations and servers via Puppet and Powershell. When the servers boot they install GO and auto-register themselves.

We still have to manually edit the xml to change the name of the agent and provide some kind of semantic name. Otherwise diagnosing pipelines or just seeing what is going on is very hard.

mikeycmccarthy:
Hi @rusio, that looks like it would be great for us, were there are any limitations around what the GUID could be? We spin up agents as docker containers as well, and we get the same issue when we spin up a new version of an agent.

ketan:
Thank you all for the feedback (and patience).

I've just about wrapped up #958 (API method for renaming agent). This'll allow renaming an agent via the API. I'm looking to make a few more quick improvements to set the agent name when auto-registering agents.

I'm considering adding a property called `agent.auto.register.hostname` inside the agent's `autoregister.properties`. This property, if set, will set the hostname of the agent, and should be very useful when auto-registering agents.

I know that this is what has been asked a couple of times here, I just wanted to get a quick show of hands to see if anyone has a better idea about how registration could be done.

reidca:
Hi Ketan, thank you so much for taking this on. This will be invaluable for us. I already use the autoregister.properties file for setting up the pre-registration key, environments and resources. The missing link has always been the agent name. Having this property will be very useful and will mean agent setup and registration will be fully-automatic.

The only other idea I had was to have an environment variable which could be used to control the hostname however with a combination of the API method and the auto register file I think we have enough.

rusio:
[...]

PR #1124, opened 2015-05-12
Setup agent hostname when auto-registering (#124)

ketan:
* Add a new property `agent.auto.register.hostname` to supply a hostname when registering a hostname
* The registration controller will use the above preferred hostname if autoregisteration key matches, else falls back to the old mechanism


ketan:
> We have logic all over the place to figure out the host name to pick up for auto registration.

I thought of doing this, and two things came to mind -

`AgentRegistrationPropertiesReader` would not just be reading a property, it'll be discovering the hostname based on the local hostname. This could cause confusion because the hostname in a lot of linux setups is set to "localhost.localdomain".

The idea behind this PR was to allow users to explicitly set the hostname via the auto-register mechanism. This mechanism means that registration keys must match before anything that the agent provides is trusted in any form. This was done to prevent malicious submission of hostname in the pretext of auto-registration.

Does that answer your question?

Issue #1121, opened 2015-05-11
Provide ability for plugin to store global configuration


srinivasupadhya:
some plugins might need ability to store configuration. eg: github plugin might need username & password for updating PR status, artifact cleanup plugin might want to take configuration of "how many instances to keep" etc.

Currently this is being done through system property, configuration file (in location specified by plugin) etc. If Go server provides standardized approach to store configuration all plugins could take advantage of this.

---

Implementation:

- [x] Plugin provides "configuration" keys & settings template
- [x] Go Server on the plugins listing page provides a "configure" icon (super admins only), on click of which it renders pop-up with template & data (if its already stored)
- [x] On save the data (key-value pairs) is passed to plugin for validation. On success they are stored as JSON to database
- [x] Plugin will be able to access the configuration through `GoApplicationAccessor` message "get-configuration"

The data will be stored in a seperate table `plugin` which has id as `plugin-id` & column `data` which is JSON representation of key-value pairs

mdaliejaz:
Verified. Works fine on 15.2.0(2093-c20772b).

PR #1127, opened 2015-05-14
Plugin settings

srinivasupadhya:
done as part of #1121


srinivasupadhya:
I ran all the rspec tests locally and they are passing. The configuration for template changed but the code change required isn't in the PR & hence the failure.

This is good to go! :+1:

srinivasupadhya:
Reviewed it with @zabil today. Merging it.

PR #1130, opened 2015-05-15
Console log coloring improvements and bug fixes (client/browser side)

ketan:
See #1092 for a server side implementation written in java.

* Use a dark background by default, but allow switching to a white background
* Render the console log from the server in using UTF-8 charset
* The same mechanism that is used to render the live console log is used to render the log after the job is complete
* In order to not blow up the browser from parsing too much text, we parse console log in chunks of 1000 lines


arvindsv:
Alright. I took a look. I like it! I like this much better than the server-side implementation. From the code and the chunking, I guess it does have the small limitation we talked about, where text with a huge chunk of color will lose its color after the 1000th line. But, that's just being extremely nitpicky. :)

I think the bright colors (bold - ```echo -e "\033[01;32mHELLO^[\033[00m"```) don't have corresponding CSS. Is this something you want to change?

I did try with your test script, and the tables look like this:

#### In my terminal

1_terminal

#### In my browser, with white theme:

2_browser_white

#### In my browser, with black theme:

3_browser_black

I happen to prefer my terminal, but I think this the browser output is better than black and white. :) I had to stop myself from going to town, getting the colors to show up exactly like my terminal. :)

```01;32m``` is quite common, I believe. So, I think that could be changed. Apart from that, it's good. The intensity table not showing up, I'm not very worried about. It's probably used much less frequently. I'm tempted to accept it the way it is, but since this is a small change, I'll wait.

**Comment by arvindsv** [comment]
Ah, ok. The middle table is full of spaces. That's why it is not showing up.

File: server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js [hunk 1]
```diff
@@ -24,7 +24,7 @@ BuildOutputObserver.prototype = {
         this.was_building = false;
         this.is_output_empty = false;
         this.is_completed = false;
-
+        this.ansi_up = ansi_up.ansi_to_html_obj();
```
ketan: Keep state across invocations of ansi_up so that we can improve colouring across chunk boundaries.

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
```diff
@@ -290,35 +290,72 @@ div.build_detail_container_content {
...
+            background-color: $color;
+        }
+
+        .ansi-bright-#{$color}-fg {
+            color: lighten($color, 33%);
+        }
+
+        .ansi-bright-#{$color}-bg {
+            background-color: lighten($color, 33%);
+        }
```
ketan: Added foreground and background colors for regular and bright (intensity).

ketan:
> Alright. I took a look. I like it! I like this much better than the server-side implementation. From the code and the chunking, I guess it does have the small limitation we talked about, where text with a huge chunk of color will lose its color after the 1000th line. But, that's just being extremely nitpicky. :)

Fixed and rebased/squashed PR.

> I think the bright colors (bold - echo -e "\033[01;32mHELLO^[\033[00m") don't have corresponding CSS. Is this something you want to change?

Fixed and rebased/squashed PR.

> 01;32m (I mean, the 01 - bold - series) is quite common, I believe. So, I think that could be changed. Apart from that, it's good. The intensity table not showing up, I'm not very worried about. It's probably used much less frequently. I'm tempted to accept it the way it is, but since this is a small change, I'll wait.

This was just missing CSS for bright bg/fg colors, that is fixed as well.

Here is how the test script https://gist.github.com/ketan/2866a236596636311d64 looks in my terminal -

screen shot 2015-06-01 at 11 09 19 am

vs how it looks in go -

screen shot 2015-06-01 at 11 13 25 am

This is missing support for <blink>, underline, and inverse effects, please don't ask me to add those :)

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/vm/structure_for_old_ui.scss
```diff
@@ -945,9 +945,7 @@ a.xlink_as_button {
 .sub_tabs_container {
     background: #999 image_url('g9/backgrounds/overlay_bottom_inside_shadow.png') bottom repeat-x !important;
     border: 0 !important;
-    padding: 10px;
-    padding-bottom: 0;
-    margin-bottom: 10px !important;
```
arvindsv: With this margin-bottom removed, the other tabs look a little weird(er), don't you think?

### Old:

old1
old2

### New:

new1
new2

arvindsv:
> This is missing support for , underline, and inverse effects, please don't ask me to add those :)

I'm not that bad. :)

arvindsv:
I've put the margin back for now, since I didn't want to wait for you to respond before accepting it. If you think it should not be there, we can talk.

ketan:
That margin was put in to accomodate for the 10px white area, but seems to have broken other stuff.

screen shot 2015-06-02 at 10 28 09 am

I've made the following change to offset your change, if you're OK with that, can you apply this change directly?

```patch
diff --git a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
index 33c74ed..30c2a3d 100644
--- a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
@@ -263,6 +263,7 @@ ul.summary li strong {
 }

 div.build_detail_container {
+    margin-top: -10px;
     padding: 0;
     zoom: 1;
     _width : 95%;
```

arvindsv:
@ketan: Done.

PavelAgejkin:
Sorry for lame question, but how to enable this colorful output?
We use powershell scripts and want to see colorful output.

ketan:
As of 15.2 this was turned on by default, so as long as your build is generating ansi colors, you should be all set.

https://www.go.cd/releases/#15.2

BrewRunner:
Is there a way to make the white theme the default?  I'm getting compaints about the white text on black background.

PR #1131, opened 2015-05-15
Remove dead code

ketan:


server/src/com/thoughtworks/go/config/GoConfigFileDao.java
```diff
@@ -177,36 +141,6 @@ public String md5OfConfigFile() {
         return cachedConfigService.currentConfig().getMd5();
     }
 
-    public void updatePipelineGroup(final PipelineConfigs pipelineConfigs, final String groupName, final String md5) {
```
zabil: You can edit the pipeline group through the ui when the admin sets pipeline group permissions. Are you sure this API is not called?

ketan: nope, for every method that is removed, I executed this code -

```
git grep -i updatePipelineGroup && git grep 'update_pipeline_group'
```

srinivasupadhya:
@zabil reviewed this. accepting on his behalf.

zabil:
The GoConfigAdministrationController routes all saves through "updateConfig" in the dao.

PR #1135, opened 2015-05-18
Authentication end-point

srinivasupadhya:


srinivasupadhya:
reviewed it with @zabil

Issue #602, opened 2014-09-30
Auto scroll-down on console view


varac:
When you have a long console output of a build that takes long, you need to manually scroll down in order to see latest output. It would be great to let the browser always display the latest messages.

matthauck:
:+1:

ketan:
This one is fairly simple to implement, and I have a working prototype. However, I'm not sure where this button can go.

arvindsv:
I remember it used to be like this, before. If you were at the bottom of the page, it would keep scrolling to the bottom, and if not, it wouldn't. How about doing that, and not having a button? I haven't fully looked at 1138, yet. I see there's a button there.

srinivasupadhya:
@ketan - is this done? can you move it in zenhub

ketan:
@srinivasupadhya, @zabil - can you look into #1281 and merge it? I know we have a small merge window, but there are a couple of PRs that could use a push.

arvindsv:
This is done. It's been like this for a year or so now.

PR #1138, opened 2015-05-18
Some console log improvements

ketan:
This branch is based off #1130, so review that before this one

* a button to automatically "follow logs" (#602)
* another button to go back to top
* hide the global back to top button when on the console log page.


ketan:
I've rebased this on top of master (with changes from #1130) so this is just one commit now.

ketan:
Ping - anyone have any feedback about the usefulness of the feature?

matt-richardson:
I definitely want a "follow the tail" feature. I had a go at implementing, but never quite finished it. 

I would probably have gone with "if the user has scrolled to the bottom of the page, then keep scrolled to the bottom", rather than a button, but that part isn't that important.

arvindsv:
I took a look. Thoughts:

1. The button "Scroll to end of logs" turns green and then the text goes away. I had to do it a few times to see if it was something that was meant to happen.

2. There are now two back-to-top buttons. I don't mind the new one, but we need to have one.

3. When it is in "Scroll to end" mode, trying to scroll up is a very jarring experience. You scroll up and it shakes the page violently, but stays at the bottom. The green button would have switched to white by now (you're no longer in "Scroll to end" mode). Now, you can scroll up.

4. Looking at the code as well, I wonder if it can be simpler. Ideally, just when new text has come from the server, we should be able to check whether the vertical offset is in the high 90s, then we scroll to the bottom, while appending that text to the end of the file. Right? Does it really need yet another jQuery plugin and an extra button? I could be wrong. It's just an initial thought.

ketan:
> The button "Scroll to end of logs" turns green and then the text goes away. I had to do it a few times to see if it was something that was meant to happen.

What do you feel would be a better user experience?

> There are now two back-to-top buttons. I don't mind the new one, but we need to have one.

There's supposed to be one of those buttons on screen, I don't know why you're seeing 2 buttons.

> When it is in "Scroll to end" mode, trying to scroll up is a very jarring experience. You scroll up and it shakes the page violently, but stays at the bottom. The green button would have switched to white by now (you're no longer in "Scroll to end" mode). Now, you can scroll up.

This would involve having hooks in BuildOutputObserver which would have caused a bunch of merge conflicts with other PRs. I can rebase this on top of #1195, meaning that #1195 must be merged first.

> Looking at the code as well, I wonder if it can be simpler. Ideally, just when new text has come from the server, we should be able to check whether the vertical offset is in the high 90s, then we scroll to the bottom, while appending that text to the end of the file. Right? Does it really need yet another jQuery plugin and an extra button? I could be wrong. It's just an initial thought.

I'll think about this a bit more.

ketan:
@arvindsv - can I have answers to the questions I asked, so I can proceed with this?

arvindsv:
@ketan: Ok, I thought about this some more. Sorry for the delay. I think:

- The "Job History" at the top right should be made click-to-show. I don't think it is important enough to show up all the time, especially, since the background of the rest of the page stops being black when it reaches it. I think something like this (just for demonstration purposes):
```javascript
jQuery("#build-status-panel .sidebar_history").attr("style", "position: absolute; z-index: 1000; right: 10px; top: 220px;");
jQuery(".job_details_content .sub_tabs_container").attr("style", "margin-right: 0px;");
jQuery(".job_details_content .build_detail_container").attr("style", "margin-right: 0px");
jQuery(".console-action-bar div").prepend("<a id='toggle-job-history' href='javascript:void(0);'>Hide job history</a>");
jQuery("#toggle-job-history").live("click", function(evt) {
                                       jQuery("#build-status-panel .sidebar_history").toggle();
                                       evt.target.text = "Show job history"
                                   });
```
Of course there are many problems with the snippet above (not the least of which is that it changes CSS in JS and that it doesn't work properly in other tabs), but I think the page looking black all over is nicer.

- About the scrolling, it should be possible to turn on live tailing when it is at the "end", as I said earlier, right? Especially now that the rest of the code is in.

arvindsv:
Actually, I took a look at the latest commit you had, and I think this is a nice implementation. A couple of things:

1. The button is not very discoverable. You need to know that it's there.

2. It does both scrolling to the end and live tailing. That's not very obvious.

However, I thought about it, and anything we do around this will be non-obvious. So, my question is: Do you think it makes sense to turn on live tailing by default? I think it does. What do you and the others think?

Then, it'll be a pulsing green button, which is a little more discoverable and can be turned off, if needed.

arvindsv:
I'm merging this. But, there are some obvious missing specs. This should be done. But, showing this working might help others see it and get some ideas. So, I'm accepting this.

PR #1139, opened 2015-05-19
Cleanup and remove some unused (or non-value-add) scripts.

ketan:


ketan:
I'd spoken to @arvindsv about this - https://gitter.im/gocd/gocd?at=553f7e46c07398745ca56fcf

Also, I've changed the config on build.go.cd (and other CI servers) to not invoke these scripts.

PR #1142, opened 2015-05-21
#1141 - Plug in admin screen has malformed URL

### LINKED ISSUE #1141: Plugin Author Link is malformed
On the plug in admin screen (/go/admin/plugins), the URL behind some author links is malformed.
It is being generated as http://https//github.com/srinivasupadhya for example.

paulclarkin:
Check for both http and https before prepending http


zabil:
Thanks!

PR #1143, opened 2015-05-21
Make Keystore password configurable by ENV

timepieces141:
Before creating the AppServer instance, it checks for the environment variable KEYSTORE_PASSWORD. If there is a password found, it uses that, if not (backwards compatible) it uses the hardcoded value stored 'password'. This should allow folks to change the password on the keystore for security sake where needed, just add the environment variable to /etc/default/go-server (or the Windows equivalent) - while maintaining compatibility with those who are fine with the historic hardcoded value.

### DIFF
--- a/server/src/com/thoughtworks/go/server/GoServer.java
+++ b/server/src/com/thoughtworks/go/server/GoServer.java
@@ -76,7 +76,9 @@ public class GoServer {
 
     AppServer configureServer() throws Exception {
         Constructor<?> constructor = Class.forName(systemEnvironment.get(SystemEnvironment.APP_SERVER)).getConstructor(SystemEnvironment.class, String.class, SSLSocketFactory.class);
-        AppServer server = ((AppServer) constructor.newInstance(systemEnvironment, password, sslSocketFactory));
+        String value = System.getenv("KEYSTORE_PASSWORD");
+        String pass = (value != null)?value:password;
+        AppServer server = ((AppServer) constructor.newInstance(systemEnvironment, pass, sslSocketFactory));
         server.configure();
         server.addExtraJarsToClasspath(getExtraJarsToBeAddedToClasspath());
         server.setCookieExpirePeriod(twoWeeks());


timepieces141:
Instructions for folks to change the password of the keystore installed with Go:

http://stackoverflow.com/questions/2889238/keystore-change-passwords

server/src/com/thoughtworks/go/server/GoServer.java
```diff
@@ -76,7 +76,9 @@ protected void startServer() throws Exception {
 
     AppServer configureServer() throws Exception {
         Constructor<?> constructor = Class.forName(systemEnvironment.get(SystemEnvironment.APP_SERVER)).getConstructor(SystemEnvironment.class, String.class, SSLSocketFactory.class);
-        AppServer server = ((AppServer) constructor.newInstance(systemEnvironment, password, sslSocketFactory));
+        String value = System.getenv("KEYSTORE_PASSWORD");
+        String pass = (value != null)?value:password;
+        AppServer server = ((AppServer) constructor.newInstance(systemEnvironment, pass, sslSocketFactory));
```
ketan: Thanks a lot for raising this issue.

We would be happy to allow users to configure this password. Before we merge this pull request, we'd like to make sure that the code confirms to some of the conventions and guidelines. To that effect, here is some feedback that would ensure that the PR gets merged.

The `SystemEnvironment` class is the one true place to configure things and for users to know what can be configured.

Could you make a change to pull this value from the `systemEnvironment` variable instead, like this.

``` java
String value = systemEnvironment.keyStorePassword();
```

for an implementation of `keyStorePassword()`, checkout this https://github.com/gocd/gocd/blob/e7c3a0ceb8d00eecd99c1927fb8a0a23608583e5/base/src/com/thoughtworks/go/util/SystemEnvironment.java#L521-523

Another secondary change that would be great to have.

Also since the `AppServer` class is being passed the `SystemEnvironment` instance, it'd be great if we could kill the password parameter to all the constructors in the `AppServer` hierarchy.

zabil: Also once in SystemEnvironment you can use

https://github.com/gocd/gocd/blob/e7c3a0ceb8d00eecd99c1927fb8a0a23608583e5/base/src/com/thoughtworks/go/util/SystemEnvironment.java#L482

timepieces141:
No problem - I can do this. Quick question, since both options were put forth - would you prefer this be a property? Or an environment variable?

Seems that @ketan's suggestion leads to a `System.getProperty()` call, whereas @zabil's suggestion leads to a `System.getenv()` call.

If it helps, I also like the idea of removing this from the `AppServer` constructor and having that class leverage its reference to the `SystemEnvironment`.

timepieces141:
Now that I have dug a little deeper (I looked through the `server.sh` file), I realize that most *properties* are actually getting their values from the environment anyway - such as `cruise.server.ssl.port` being set to `$GO_SERVER_SSL_PORT` which can be found set in the `/etc/default/go-server` file sourced at the top of `server.sh`. (That said, I do not know how this is trickling down when the server is running on Windows). I'll follow that model unless there are objections.

zabil:
@timepieces141 I got it wrong.
You need to use the example @ketan mentioned

https://github.com/gocd/gocd/blob/e7c3a0ceb8d00eecd99c1927fb8a0a23608583e5/base/src/com/thoughtworks/go/util/SystemEnvironment.java#L521-523

To set up keystore password as a system property and not an environment variable.

ketan:
@timepieces141 - are you still working on this PR? would you need some help from us?

timepieces141:
Sorry for being away for so long. Truth is, I am working under a contract that, apparently, doesn't allow me to contribute publicly. I can't make anymore contributions - I shouldn't have even made the one I did (all three lines of it!). But I will say, it seems like you guys know exactly what needs to happen and it's a fairly light lift. So I hope I have shed light on something that is a good idea and someone else can take it from here.

zabil:
No worries.
We'll take it from here.

ketan:
On second thought - having the system property inject via `-Dgo.keystore.password=foo` is not the most secure thing to do.

zabil:
Closing this pull request, we'll look at #1344 to provide a better way to implement this.

PR #1150, opened 2015-05-25
Various build script cleanup operations

ketan:
I'm trying to get the build time down and realized a few things that 
are either wrong, or unused, so I'm going to create multiple small 
commits (so that they are easy to review)


common/pom.xml
```diff
@@ -338,22 +333,10 @@
...
-            </plugin>
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <version>2.4</version>
-                <configuration>
-                    <archive>
-                        <manifestEntries>
-                            <Go-Version>${go.version}</Go-Version>
```
zabil: I see that you've writing the version to the manifest entry. 
Who's doing this now? I don't see it in the root pom

ketan: I think I did not rebase things correctly, fixing this now.

ketan: This is removing the dup from here https://github.com/gocd/gocd/blob/3b6aac252eef6061016e5cecde6aa73ef5da04d0/common/pom.xml#L343-L362, which is why you see an additional `<goal>jar</goal>` in the lines that follow. Ditto with `config/config-api/pom.xml`

PR #1155, opened 2015-05-26
Upgrade TFS to 12.0.2 (See #234)

juhijariwala:


arvindsv:
Like the Jetty upgrade change, what does it take to make this fully switch-able? I ask because TFS upgrade has caused issues before. There was once an upgrade and then an emergency build to downgrade it, since there was a memory leak inside TFS SDK.

I know the packaging is a little weird, but it should be possible, I think?

ketan:
I'm honestly not sure how that'll fly given that there's a "native" component involved -

https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/tfs/TfsSDKCommandBuilder.java#L104-L108

arvindsv:
That doesn't mean that it cannot be cleanly separated, right? The separation should just be about making that dependency available at runtime, rather than at compile time.

ketan:
Are you suggesting that users can switch tfs sdk versions between reboots, because I *think* that is the only way this could be achieved in a manner that is simple, and does not involve too much of yak shaving.

arvindsv:
Yes, that's what I mean. I don't actually think any users will want to (and should) switch between versions across reboots, but having that choice allows us to upgrade TFS without so much trepidation. Having been bitten by it a couple of times, I see that everyone is apprehensive about upgrading it. I'd like to remove that mental block. :)

arvindsv:
I don't know if you mean machine reboots. I just mean Go server restarts. The native component should just be a DLL or a shared object, which is used by the TFS client. There should be no daemon running (I think the TFS client might start one). But, the choice of switching SDKs is what I'm talking about. At runtime.

ketan:
I did mean a go server reboot.

I think we can address your concern and feedback tomorrow after we wrap up adding secure environment vars on jobs/stages/environments.

arvindsv:
About [this commit](juhijariwala/gocd@81dd5af4dbb02bc74af7db73678cfe12a3ab0bbf): I understand why it is needed. In this case, you were lucky because TFS was initialized in the middle of the flow, after a material was sent across the wire. However, if TFS was supposed to be initialized before (during agent startup, say), this wouldn't have worked.

I think it makes sense to have a mechanism to send relevant system properties from the server to the agent, as a part of the ping between them, for instance. A more formal mechanism, if you will, instead of a hack, as you say. What do you think?

If you think that should be separate from this, I can see that would make sense. But, we should move quick to remove that hack, if we all agree about this approach.

arvindsv:
Apart from that, I have no more questions. I've reviewed this and can be merged.

ketan:
> I think it makes sense to have a mechanism to send relevant system properties from the server to the agent, as a part of the ping between them, for instance. A more formal mechanism, if you will, instead of a hack, as you say. What do you think?

I feel that this information is needed for a plugin to execute, as such I feel that it should should come along with the work that gets assigned to the agent.

Pulling configuration in between pings to setup "global state" feels rather nasty.

WDYT?

arvindsv:
Hmm, this is not a plugin, it's a part of the core. But, I get what you're saying. My point is: It is some state that is needed. In this case, the state can come along with work that gets assigned. However, if this state was something to do with initialization of the agent itself, (say, the property of the plugin infrastructure, which allows plugins to be loaded without a restart), then it cannot be sent along with work. It needs to be sent before it.

But sure, I see the point about pulling configuration between pings being hacky and nasty. It can be even done during agent registration. I brought up pings because a server can restart and system properties might get changed, and the agents would not restart. So, their properties cannot be changed, unless they do. A ping happens irrespective. If we stop thinking of it as a ping, and think of it as a constant connection for communication between the agent and server, we might be able to do a lot using that usually useless back and forth.

This is getting a little complicated and I feel I'm not explaining my thought process well. Tell me what parts I should explain better.

Finally, I felt the need to hijack pings when trying to implement [proper SSH key support](https://github.com/arvindsv/gocd/commits/ssh_key_management). When a key is uploaded to the server, it needs to propagate to the agents. We have no mechanism for that, apart from pingThread and workThread. That's when I wanted to [overhaul the ping mechanism](https://github.com/arvindsv/gocd/commit/68f32a014d4d85cb1761335a382c894a570602a5). The ping mechanism is (ab)used for canceling a job as well, since it is the only out of band communication when an agent is doing some work. I thought it made sense to generalize it.

Thoughts?

ketan:
> However, if this state was something to do with initialization of the agent itself, (say, the property of the plugin infrastructure, which allows plugins to be loaded without a restart), then it cannot be sent along with work. It needs to be sent before it.

I hear you about the need for having a formal mechanism to send configuration data between servers and agents.

My assumption was that if the server dies, the agent dies, but the bootstrapper keeps it alive. Based on your explanation, that does not seem to be the case.

In the situation that you explain, where the agent process does not reboot between server reboots, my understanding is that that TFS reloading would continue to be impossible because of lack of "re-initialization" and unloading capabilities of the "TFS bootstrapper" that we have. This feature toggle itself would need a bit more thought than what I'd initially anticipated.

We probably need to have a bigger conversation about improving how the initialisation and configuration of the agent happen, and I'd like us to put this upgrade on hold pending that conversation.

arvindsv:
> We probably need to have a bigger conversation about improving how the initialisation and configuration of the agent happen, and I'd like us to put this upgrade on hold pending that conversation.

Alright.

zabil:
Moving this to cold storage.

PR #1156, opened 2015-05-26
Some build script cleanups

ketan:


bn
```diff
@@ -22,5 +22,4 @@ else
 fi
 export TF_CLC_HOME=$project_root/tfs-tool
 
-$(dirname $0)/tools/bin/go.jruby -J-Xmx2048m -S buildr $*
-
+$(dirname $0)/tools/bin/jruby -S buildr "$@"
```
arvindsv: Might need to setup JRUBY_OPTS env var on some agents, if they fail due to OOM.

ketan: I'm aware of that - the reason I choose to remove that from the top level is because jruby is calling maven which is causing jruby and that is running near the threshold of how much memory is available on the machines.

So we can definitely bring back the `-Xmx` param as and when needed for specific operations that need that amount of memory.

PR #1160, opened 2015-05-27
Return the right codes for material notify api

zabil:
204 - No content success response when the material is notified.
404 - When material is not found
content-type as text

### DIFF
--- a/server/test/unit/com/thoughtworks/go/server/materials/MaterialUpdateServiceTest.java
+++ b/server/test/unit/com/thoughtworks/go/server/materials/MaterialUpdateServiceTest.java
@@ -28,11 +28,13 @@ import com.thoughtworks.go.config.CruiseConfig;
 import com.thoughtworks.go.config.materials.ScmMaterial;
 import com.thoughtworks.go.config.materials.svn.SvnMaterial;
 import com.thoughtworks.go.config.materials.svn.SvnMaterialConfig;
+import com.thoughtworks.go.domain.PipelineGroups;
 import com.thoughtworks.go.domain.materials.Material;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
 import com.thoughtworks.go.helper.MaterialConfigsMother;
 import com.thoughtworks.go.helper.MaterialsMother;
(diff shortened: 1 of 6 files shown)


arvindsv:
The change looks good to me. Was this the one we talked about trying to return a 202? Should the content-type be JSON or plain text?

zabil:
yes. have to make that change too.

--
Typed using fat thumbs




On Mon, Jun 1, 2015 at 6:29 AM -0700, "Aravind SV" <notifications@github.com> wrote:












The change looks good to me. Was this the one we talked about trying to return a 202? Should the content-type be JSON or plain text?

arvindsv:
Ok. Won't merge, then. Unless you plan for it to be in an other PR.

zabil:
Made the changes.

server/test/unit/com/thoughtworks/go/server/materials/MaterialUpdateServiceTest.java
```diff
@@ -188,6 +205,11 @@ public void shouldReturnImplementerOfSvnPostCommitHookAndPerformMaterialUpdate_W
         spyService.notifyMaterialsForUpdate(username, params, result);
         verify(svnPostCommitHookImplementer).prune(anySet(), eq(params));
         verify(spyService).updateMaterial(svnMaterial);
+
+        HttpLocalizedOperationResult acceptedResult = new HttpLocalizedOperationResult();
+        acceptedResult.accepted(LocalizedMessage.string("MATERIAL_SCHEDULE_NOTIFICATION_ACCEPTED"));
+
+        assertThat(result, is(acceptedResult));
```
ketan: Was there a reason to not be consistent with the rest of the testcase -

```
        assertThat(result.isAccepted(), is(false));
        assertThat(result.httpCode(), is(HttpStatus.SC_ACCEPTED));
        assertThat(result.hasMessage(), is(true));
```

zabil: The equals was not implemented earlier. We have to do that for checking the localizedmessage. Will change it consistent at other places and synchronize this.

Issue #194, opened 2014-05-20
Allow secure variables (environment) on environments, stages and jobs level


agileassistance:
(enhancement)
Is there a good reason for only allowing secure variables (environment) on the pipeline level? If not it would be great also to have this option on pipeline group, stage and job level.

edit: pipeline group level does not support env. vars

agileassistance:
I would really appreciate this feature in the upcoming release; it's vital for keeping sensitive information secured in the build and the lack of it leads to some convoluted/unnecessary workarounds that makes the pipelines hard to maintain.

petergillardmoss:
+1 to this.

At the moment we have pipeline explosion as we need to create new pipelines every time we need to run the same job but with different secure variables.

juhijariwala:
Picking this up!

srinivasupadhya:
https://github.com/gocd/gocd/pull/1161 - implements support for environments, stage & job level secure env. vars.

PR #1161, opened 2015-05-27
Allow secure variables (environment) on stages, jobs and environments level (#194).

juhijariwala:


server/webapp/WEB-INF/rails.new/app/assets/javascripts/environment_variables.js
```diff
@@ -138,8 +137,12 @@ EnvironmentVariables = function() {
         });
     };
 
-    function createWrapperElement(tag_type) {
-        return jQuery(document.createElement(tag_type));
+    function createWrapperElement(tag_type, template_text) {
+        if(tag_type){
+          return jQuery(document.createElement(tag_type)).html(template_text);
+        } else {
+          return jQuery(template_text);
```
jyotisingh: Not sure if I understand this well. Did u really mean to return jQuery(template_text). Template text would normally be some html and not a type or a selector.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/jquery.environmentVariableEdit.js
```diff
@@ -0,0 +1,65 @@
+(function ($) {
+  $.resetEnvironmentVariableOn = function (options) {
```
jyotisingh: I would recommend against writing global functions such as this one. Could you try encapsulating this in a wrapper class instead.

jyotisingh: I do not see any tests for this javascript. Could you write them and update the PR.

jyotisingh: Also, we use snake case to name javascript files.

jyotisingh:
This fixes most part of #194. The original ticket also suggested support for secure environment variables at pipeline group level. Currently, there is no concept of environment variables (even plain text ones) at pipeline-group level so I am totally fine not implementing that part of it. @arvindsv - what do you say?

server/webapp/WEB-INF/rails.new/app/assets/javascripts/environment_variables.js
juhijariwala: JQuery is smart about detecting if the "selector" is actually HTML text and creates an element accordingly. http://jsfiddle.net/r16muadb/1/

ketan:
@jyotisingh - I have addressed your feedback and rebased the commit. Would you mind looking at this again?

PR #1162, opened 2015-05-28
Add an environment variable to disable asset compression to speed up server builds locally.

juhijariwala:


arvindsv:
Hmm, if it is local builds, shouldn't we be using a different environment? development or even a new one, where Rails.env is ```local_build``` or something of that sort, rather than putting logic to switch stuff around in production.rb? What am I missing?

ketan:
Assets for prod environment need to be (and [are being](https://github.com/gocd/gocd/blob/1afe499e5c7e819b3783e5301b5def4f9dfafb08/server/prepare_go_server_webapp.rake#L114-L125)) compiled down using production specific settings. (compress, digest, etc.)

Since we're having to do repeat compiles and packaging of the server jar for testing by building a production distribution, we figured that not compressing assets saved us a bunch of time.

jyotisingh:
@ketan - your point about saving time while re-building for local testing is very valid and it would be nice if we could make it faster. However, I am with @arvindsv on using a different environment config for this. 

It is fairly straightforward to do that anyway. All you need to do is move over the node code you have written (ie. class definition for NoCompression and the assignments) to a new environment config file (say server/webapp/WEB-INF/rails.new/config/environments/funky.rb) and set RAILS_ENV to *funky* while running the build. 
Well, there is one more change thats required in server/prepare_go_server_webapp.rake:

```patch
+  environment = ENV['RAILS_ENV'] || "production"
+
   if Gem.win_platform?
-    ENV['RAILS_ENV'] = "production"
+    ENV['RAILS_ENV'] = environment
     sh <<END
     cd #{File.expand_path(File.join(File.dirname(__FILE__), "webapp", "WEB-INF", "rails.new"))} &&
 END
   else
-    sh "cd #{File.join("webapp/WEB-INF/rails.new")} && RAILS_ENV=production #{ruby} -S rake assets:
+    sh "cd #{File.join("webapp/WEB-INF/rails.new")} && RAILS_ENV=#{environment} #{ruby} -S rake ass

```

What do you think?

ketan:
seems fine, can you close this issue?

PR #1163, opened 2015-05-28
Moved about page to rails controller.

juhijariwala:
* Replaced AboutService.java with AboutHelper.rb
* Removed old css and replaced it with minimal about.scss


server/webapp/WEB-INF/rails.new/app/helpers/about_helper.rb
```diff
@@ -0,0 +1,69 @@
...
+  def os_information
+    [system_property('os.name'), system_property('os.version')].join(' ')
+  end
+
+  def system_property(property)
+    system_environment.getPropertyImpl(property)
+  end
+
+  def available_space
+    artifact_dir = Spring.bean(com.thoughtworks.go.server.service.ArtifactsDirHolder.java_class).getArtifactsDir
```
arvindsv: I think you should just put ```:artifacts_dir_holder``` in [services.rb](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/lib/services.rb#L29) and use that. That's what it is for.

arvindsv:
I think you should put those circles over Hakan's name back. I'm sure he'll mind.

arvindsv:
Rest of it looks good to me. Can merge.

ketan:
> I think you should put those circles over Hakan's name back. I'm sure he'll mind.

You mean Håkan's name? I'm sure he'll mind ;)

ketan:
See #1167, you may close this out.

PR #1164, opened 2015-05-28
Simplify jruby and rails setup

ketan:


arvindsv:
I'm afraid to click on "Files changed - 8863". :) Even though I know what the change is.

ketan:
Don't - it'll certainly crash your browser :)

arvindsv:
In this:

```
diff --git a/tools/rails/bin/jruby b/tools/rails/bin/jruby
new file mode 100755
index 0000000..4b75f19
--- /dev/null
+++ b/tools/rails/bin/jruby
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+GO_ROOT=$(cd $(dirname $0)/../../.. > /dev/null && pwd)
+JRUBY_BASE="$GO_ROOT/tools/jruby-1.7.11"
```

Can you change that line to something like:

```
GO_ROOT=$(cd "$(dirname "$0")"/../../.. > /dev/null && pwd)
```

Otherwise, when used with the full path of a path with spaces, it fails. Admittedly, a corner case.

arvindsv:
In [this](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/prepare_go_server_webapp.rake#L123), do you know why it is ```v[0..100]```?

arvindsv:
Can [this line](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/webapp/WEB-INF/rails.new/config/application.rb#L6) be switched out, if Rails.env is production?

arvindsv:
Some files like [this one](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/tools/jruby-1.7.11/bin/htmldiff#L3) have a path on your machine. I guess those lines can be removed.

arvindsv:
Everything else I care about seems like a rename. Even if they're not, it's too many for me to take a look at. So, I'm fine accepting this, and seeing what the tests show up.

arvindsv:
One thing for a later PR is to remove the test gems from being packaged (as we talked earlier, I think). I see that [init.rb](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/webapp/WEB-INF/init.rb#L13) has BUNDLE_WITHOUT, but test gems are packaged, still.

arvindsv:
Maybe @jyotisingh or @zabil can take a quick look? Or you (@ketan) can do a quick code walkthrough? In case I missed something important.

ketan:
> Can you change that line to something like:
> 
> GO_ROOT=$(cd "$(dirname "$0")"/../../.. > /dev/null && pwd)
> Otherwise, when used with the full path of a path with spaces, it fails. Admittedly, a corner case.

Done

> In [this](https://github.com/ketan/gocd/blob/5c31ae80bc416d489faf0a3cc7773f7e5af9959d/server/prepare_go_server_webapp.rake#L123), do you know why it is v[0..100]?

Yes - a $CLASSPATH containing 100s of jars with absolute paths, is pretty useless to look at every time you run a command.

> Can [this line](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/webapp/WEB-INF/rails.new/config/application.rb#L6) be switched out, if Rails.env is production?

I'm not sure, rails [by default](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/railties-4.0.4/lib/rails/generators/rails/app/templates/config/application.rb#L3-L12) includes [`rails/all`](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/railties-4.0.4/lib/rails/all.rb) or `rails/test_unit/railtie`. Either way, that line is present. I can submit a separate PR to remove that line if I'm able to run things fine.

> Some files like [this one](https://github.com/ketan/gocd/blob/cleanup-jruby-mess/tools/jruby-1.7.11/bin/htmldiff#L3) have a path on your machine. I guess those lines can be removed.

It's not me, [it's jruby](https://github.com/ketan/gocd/blob/12224341b494978e176662bb1ff5c689ac784e10/tools/jruby-1.7.11/bin/htmldiff) doing its thing, I suppose. That line is commented out, so I think it should be fine.

ketan:
I've updated the PR with another commit https://github.com/ketan/gocd/commit/a13278f0cc8065c1f0dc9df189bd3628ae28f1de to remove unwanted gems in production

PR #1165, opened 2015-05-28
#000 - Velocity exception page shows generic message.

arvindsv:
Also, header is not shown in the exception page, since the localizer is not always available.

[Before](https://build.go.cd/go/tab/build/detail/WRONG-build-linux/464/build-server/1/server-split-runInstance-8) this change:

old_message

After this change:

new_message


ketan:
Works as advertised, good to merge - :+1:

PR #1167, opened 2015-05-29
Moved about page to rails controller

ketan:


arvindsv:
Did you really need ```new-css```?

ketan:
We could not reuse the old CSS, because too much of it was tied to the old layouts. So we just created a really simple CSS with no classes or IDs.

Not having that would apply reset.css and the text would look horrible.

Issue #1169, opened 2015-06-01
Upgrade the version of apache shindig for mingle gadgets


zabil:
https://gitter.im/gocd/gocd?at=556c25448d1a52c22e86e4e3

ketan:
@jyotisingh said she had a partial patch sitting somewhere to upgrade shindig. I know that part of the code, so I can finish it off quickly.

This would also be a good opportunity for someone to pair with me and get to know a bit more about gadgets.

jyotisingh:
>> @jyotisingh said she had a partial patch sitting somewhere to upgrade shindig

Well what I said was while trying to introduce a library for json schema validations, I faced some runtime dependency conflict with shindig which would require upgrading shindig. I think I did make an attempt at upgrading shindig, but that also required upgrading gadgets which I wasn't keen on diving into at that point so that's where I stopped.

arvindsv:
If you have that, at the point you stopped, let's finish it off? I don't mind pairing with you and learning about it. I know a bit of it, but there's lots I don't know.

arvindsv:
If you don't, that's ok. We can start on it and get to that point.

zabil:
Make this a plugin and remove it from core

ketan:
Closing, not relevant anymore.

PR #1171, opened 2015-06-02
#000 Removed/Upgraded libs that's pulling older versions of xerces

zabil:
This to fix the problem of 
```
WARNING:  'org.apache.xerces.jaxp.SAXParserImpl: Property 'http://javax.xml.XMLConstants/property/accessExternalDTD' is not recognized.'
Warning:  org.apache.xerces.parsers.SAXParser: Property 'http://javax.xml.XMLConstants/property/accessExternalDTD' is not recognized.
Warning:  org.apache.xerces.parsers.SAXParser: Property 'http://www.oracle.com/xml/jaxp/properties/entityExpansionLimit' is not recognized.
Warning:  org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser: Property 'http://www.oracle.com/xml/jaxp/properties/entityExpansionLimit' is not recognized.
```

These warnings are printed when you have xerces on the classpath. A newer version of the JDK(8) does not fix this. The only solution is to get rid of xerces  (which now comes bundled with the idk) as a transitive dependency. 

This pull request fix it only on some modules.

Apache shindig (used on our app) hard wires the xerces dependency and modules referring shindig will still print this error. For a long term solution we should upgrade the version of Apache shindig.

### DIFF
--- a/app-server/pom.xml
+++ b/app-server/pom.xml
@@ -177,30 +177,6 @@
             <version>${project.parent.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>com.tlb</groupId>
-            <artifactId>tlb-java</artifactId>
-            <version>0.3.2-90-g6df47e3</version>
-            <scope>test</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-io</groupId>
-                    <artifactId>commons-io</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>commons-codec</groupId>
-                    <artifactId>commons-codec</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>log4j</groupId>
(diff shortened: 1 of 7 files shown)


jyotisingh:
This changes make sense, but would not fix the warnings when Go server runs. Probably affects test runs only. Anyway, could you sync your changes with upstream so I could proceed with the merge.

ketan:
> This changes make sense, but would not fix the warnings when Go server runs. Probably affects test runs only. Anyway, could you sync your changes with upstream so I could proceed with the merge.

@jyotisingh I've seen these warnings in production code as well.

zabil:
Will be solved in production code once we finish this #1169

jyotisingh:
@ketan - Yes the warnings show up in production as well, thats exactly what I meant when I said `when Go server runs`. I am okay to merge these changes as the dependencies which were removed in the pr did not make sense in the first place.

Issue #1089, opened 2015-04-30
Rescheduled Jobs not being marked as completed on completion


zabil:
Jobs that are rescheduled needs to be marked as complete once the job passes/fails. A cache issue is causing the job to be shown as rescheduled. 

Restarting the server displays the job as complete.

zabil:
Picking this up

chopraapooja:
STEPS to reproduce :-
1) Have two agents connected to the server.
2) Create a pipeline with a job that will end up longer [to get time to disconnect agent before job completes] say "sleep 120".
3) when job is in the middle of execution, just disconnect the building agent.
4) Job will get rescheduled to another agent after some time. 
5) Keep eye at the job details page . It should show the correct status (without refreshing) and also check console logs.

You should get jobstatus in this order :-
Scheduled,
Assigned
Preparing
Building
Rescheduled,
Scheduled,
Assigned
Preparing
Building
Completing,
Completed

This is the ideal behaviour. But the bug is that job details page stop showing updated status after job gets Reschedule and hang upon "Reschedule" state. 
**Note**:-
To check whether status is updating **without refresh** it is recommended that you should stay on the job details page throughout job execution cycle , don't go back and forth. Use another browser window or tab if you want to goto another page.

juhijariwala:
I reproduced the scenario as above, but it is **not showing the rescheduled state.**

### Current Job status order ###
Scheduled
Assigned
Preparing
Building (killed the assigned agent)
Scheduled
Assigned
Preparing
Building
Completing
Completed

Observed on **15.2.0(2067-07d65665c731a6) **

juhijariwala:
Verified with 15.2.0-2067 (07d65665c731a615f6a6943cce0fd03440eb4a9e).This works as expected.

PR #1174, opened 2015-06-03
#1089 Request for jobStatus.json finds the actual latest job Id using…

chopraapooja:
… the requested job Id which is resceduled job Id

### DIFF
--- a/server/src/com/thoughtworks/go/server/controller/JobController.java
+++ b/server/src/com/thoughtworks/go/server/controller/JobController.java
@@ -16,38 +16,20 @@
 
 package com.thoughtworks.go.server.controller;
 
-import java.util.HashMap;
-import java.util.Map;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
 import com.thoughtworks.go.config.AgentConfig;
 import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.Tabs;
 import com.thoughtworks.go.config.TrackingTool;
-import com.thoughtworks.go.domain.JobIdentifier;
-import com.thoughtworks.go.domain.JobInstance;
-import com.thoughtworks.go.domain.JobInstances;
-import com.thoughtworks.go.domain.Pipeline;
-import com.thoughtworks.go.domain.Properties;
-import com.thoughtworks.go.domain.Stage;
-import com.thoughtworks.go.domain.StageIdentifier;
+import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.i18n.Localizer;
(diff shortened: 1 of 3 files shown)


server/src/com/thoughtworks/go/server/controller/JobController.java
```diff
@@ -154,10 +141,12 @@ public ModelAndView handleRequest(@RequestParam(value = "pipelineName")String pi
                                       HttpServletResponse response) {
         Json json;
         try {
-            JobInstance instance = jobInstanceService.buildByIdWithTransitions(jobId);
-            JobStatusJsonPresentationModel presenter = new JobStatusJsonPresentationModel(instance,
-                    goConfigService.agentByUuid(instance.getAgentUuid()),
-                    stageService.getBuildDuration(pipelineName, stageName, instance));
+            JobInstance requestedInstance = jobInstanceService.buildByIdWithTransitions(jobId);
+            JobInstance mostRecentJobInstance = jobDetailService.findMostRecentBuild(requestedInstance.getIdentifier());
```
arvindsv: I think someone closer should review this. :) I think it looks good. The only thing I'd think about is the impact on performance. This is called from the jobs detail page, right? I see that after #1130 was merged, the jobStatus.json call never stops, even after a build has gone green. It might be useful to fix that too, so that this call does not turn out to be a bottleneck.

/cc @ketan, what do you think?

zabil: @arvindsv true
Also note the calls
```
JobInstance requestedInstance = jobInstanceService.buildByIdWithTransitions(jobId);
JobInstance mostRecentJobInstance = jobDetailService.findMostRecentBuild(requestedInstance.getIdentifier());
```
are cache aware

arvindsv:
> One option we've considered is posting the identifier instead of the id from the ui but it's too long

Even if you post it, on a reschedule it has to get changed, right?

I think the cache should be ok, but we should reconsider sending jobStatus.json calls after the build is finished. If that is fixed, there will be lesser load on this anyway.

server/src/com/thoughtworks/go/server/controller/JobController.java
ketan: > I see that after #1130 was merged, the jobStatus.json call never stops, even after a job has gone green or red. It might be useful to fix that too, so that this call does not turn out to be a bottleneck.

Logged #1204 to track this.

zabil:
> Even if you post it, on a reschedule it has to get changed, right?

Sorry, yes. In short, we cannot avoid the call to find the most recent build. 
Should I fix the jobStatus polling in this pull request or #1204?

arvindsv:
I think doing that in #1204 is completely ok. So, we can merge this PR.

PR #1175, opened 2015-06-03
Moving artifacts fix

chopraapooja:
* Will not move console artifacts when temporary console-logs are not present


chopraapooja:
This behavior is no longer required.

PR #1176, opened 2015-06-03
#NA - Setting plugin location monitor interval to 5secs for Developme…

jyotisingh:
…nt Server

This makes it convenient to test plugins and end-points while developing.

### DIFF
--- a/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
+++ b/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
@@ -52,6 +52,7 @@ public class DevelopmentServer {
 
         systemEnvironment.setProperty(SystemEnvironment.PARENT_LOADER_PRIORITY, "true");
         systemEnvironment.setProperty(SystemEnvironment.CRUISE_SERVER_WAR_PROPERTY, webApp.getAbsolutePath());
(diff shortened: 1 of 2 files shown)


development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
```diff
@@ -52,6 +52,7 @@ public static void main(String[] args) throws Exception {
 
         systemEnvironment.setProperty(SystemEnvironment.PARENT_LOADER_PRIORITY, "true");
         systemEnvironment.setProperty(SystemEnvironment.CRUISE_SERVER_WAR_PROPERTY, webApp.getAbsolutePath());
+        systemEnvironment.set(SystemEnvironment.PLUGIN_LOCATION_MONITOR_INTERVAL_IN_SECONDS, 5);
```
zabil: What about the agent?

arvindsv:
Also, does a plugin that needs to be sent to the agent (like the SCM or task plugins) actually get sent to the agent, when it is updated on the server? I thought it wasn't sent. Would be nice if it was.

jyotisingh:
@arvindsv - That bug was [fixed](https://github.com/gocd/gocd/issues/825) so, download on agent should work. 
@zabil - Will update the PR.

arvindsv:
> @arvindsv - That bug was fixed, so download on agent should work. 

Woo hoo!

zabil:
looks good to me.

PR #1180, opened 2015-06-04
Change min value to 0 on graphs tab.

nfisher:
Given that build numbers monotonically increment and cannot be negative it is desirable to have the graph start at 0 for the x-axis.


zabil:
Thanks!

PR #1183, opened 2015-06-04
Use `PATCH` instead of `PUT` to update an agent properties.

ketan:


zabil:
as discussed here :point_right: [June 9 2015 11:58 AM](https://gitter.im/gocd/gocd?at=5576877427d2203776cd08ea)

use agent_name instead of hostname to modify the agent resource
```
curl --request PATCH -d agent_name=something -H "Accept: application/vnd.go.cd.v1+json" http://admin:badger@localhost:8153/go/api/agents/{uuid}
```

zabil:
Also,
It returns 200 status for an invalid patch 

```
curl --request PATCH -d nonexistant=patch -H "Accept: application/vnd.go.cd.v1+json" -v http://admin:badger@localhost:8153/go/api/agents/82d31545-de97-493a-9c68-b54df23b5c59
```
http://tools.ietf.org/html/rfc5789#section-2.2

Issue #1129, opened 2015-05-15
Technical Debt: Remove prototype.js usage


zabil:
Wherever used migrate from prototype.js to something else, preferably vanilla js.

zabil:
Periodical executor #1304

PR #1195, opened 2015-06-05
Fix multiple issues because of console log coloring

ketan:


arvindsv:
Can you put some information about the issues fixed? I think I see what it is, but will help later, I feel. Also, because of other merges, this is conflicting.

ketan:
Rebased, all information you asked for was always in the commit message. I find it more useful to have it there than this issue, esp when doing a `git blame`.

arvindsv:
> all information you asked for was always in the commit message

Damn. Didn't see. Sorry. You're right.

server/webapp/WEB-INF/vm/build_detail/_buildoutput.vm
```diff
@@ -14,30 +14,25 @@
...
-            if (jQuery('#buildoutput_pre').length === 0){
-                return; // there's no theme to change
-            }
-            jQuery('#buildoutput_pre').toggleClass('white-theme');
-            jQuery('#console-action-bar').toggleClass('white-theme');
-        });
-    });
+    if(navigator && navigator.userAgent && navigator.userAgent.include('Firefox/2') && $('buildoutput_pre')){
+        (function(){
+            $('buildoutput_pre').up('div').setStyle({overflow: 'visible'});
```
zabil: shouldn't this be ```$('.buildoutput_pre')``` ?

ketan: Classic issue when switching between jQuery and prototype. jQuery is set to `.noConflict()`, while prototype maps to `$`. So the following are roughly equivalent -

|jquery selector  |prototype selector|
|---------------------|------------------------|
|`jQuery('.foo')`  | $('foo')                  |
|`jQuery('#foo')` | $$('foo')                |

ketan: also, I think we can kill that section altogether, it's testing for really old versions of firefox that I don't think we need to care for anymore

ketan: I think there's a problem, we never intended for those lines to come there. I don't know why they are there.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
```diff
@@ -72,15 +72,21 @@ BuildOutputObserver.prototype = {
...
+        // on on the console tab, and another on failures tab
+        var buildoutputPreElement = $$('.buildoutput_pre');
         if (!is_output_empty && buildoutputPreElement) {
             var escapedOutPut = build_output.escapeHTML();
             if (Prototype.Browser.IE) {
                 // Fix for the IE not wrap /r in pre bug
                 escapedOutPut = '<br/>' + escapedOutPut.replace(/\n/ig, '<br\/>');
-                buildoutputPreElement.innerHTML += escapedOutPut;
+                buildoutputPreElement.each(function(e, i){
+                    e.innerHTML += escapedOutPut;
```
zabil: When used like this might want to consider the append() or the html() method?
http://stackoverflow.com/questions/18393981/append-vs-html-vs-innerhtml-performance 

Also jquery takes care of multiple element matches right?

```
if (Prototype.Browser.IE) {
    // Fix for the IE not wrap /r in pre bug
    escapedOutPut = '<br/>' + escapedOutPut.replace(/\n/ig, '<br\/>');
    buildoutputPreElement.append(escapedOutPut);
} else {
    buildoutputPreElement.insert({bottom: escapedOutPut});
}

```
Works for me

ketan: That block of code uses prototype, not jQuery.

Also, I was not willing to make bigger changes there knowing that this is old code that is going to be superseded by `_update_live_output_color`.

zabil: $$ sigh

#1129

PR #1201, opened 2015-06-08
Changed session tracking mode to cookie only

### LINKED ISSUE #1185: Accessing login url for auth endpoint for the first time throws a 404 page
Related to #851

Steps:
1. Install and configure github-oauth plugin in Go
2. open a browser, clear all cache and cookies.
3. Access the login url -> localhost:8153/go/plugin/interact/github.oauth.login/index

It should redirect to github login page, but it throws a 404 page from github instead. If you hit the url again from browser, it starts working as expected. 

Findings:
Basically, first time around jsessionid gets generated and attached to the url causing github to throw a 404, eg: 
```url
https://github.com/login/oauth/authorize;jsessionid=px4zl3mmyjrx1toq7bw5e2v0d?client_id=6c4ce62f2e0d2fb1dde1&response_type=code&redirect_uri=https%3A%2F%2Flocalhost%3A8154%2Fgo%2Fplugin%2Finteract%2Fgithub.oauth.login%2Fauthenticate&scope=user,user:email
```

On second attempt, jessionid is not a part of the url (as a JESSIONID cookie is already set) and hence github responds with the authentication page as expected.

jyotisingh:
By default jetty sets it to url and cookie (http://www.eclipse.org/jetty/documentation/9.2.8.v20150217/session-management.html#session-tracking-modes)
Also, updated the xsd version used by the web xml files

Fixes #1185 and [this](https://groups.google.com/forum/#!topic/go-cd/xU8aPOYGBSs/discussion) issue reported on mailing lists.

### DIFF
--- a/server/test/integration/com/thoughtworks/go/server/web/WebappSessionConfigIntegrationTest.java
+++ b/server/test/integration/com/thoughtworks/go/server/web/WebappSessionConfigIntegrationTest.java
@@ -1,0 +1,67 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2015 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END***********************************/
+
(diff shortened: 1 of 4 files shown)


arvindsv:
Checked using this config:

```
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="75">
  <server artifactsdir="artifacts" commandRepositoryLocation="default" serverId="91c3e632-5405-4287-968e-1de4e715db08">
    <security>
      <passwordFile path="/path/to/a/valid/password.properties" />
    </security>
  </server>
</cruise>
```

with a build without this change, and with a build with this change. Accessing: http://localhost:8153/go in an incognito window will show jsessionid in the URL in the older builds, and won't in this.

arvindsv:
I don't think this needs any new functional tests, or changes to existing ones.

ketan:
Verified, works on `15.2.0(2055-c3c306dc03a122)`. Please close this issue out.

PR #1203, opened 2015-06-09
Various changes to the agents API -

ketan:
* Use `PATCH` instead of `PUT` to update an agent properties.
* Fold an agent's `enable` and `disable` into the `update` call.
    When users do not specify a value for enabled,
    treat it as unset, and don't update the enabled flag (TriState)
* Rename some attributes
  * `agent_name` => `hostname`
  * `os` => `operating_system`

### DIFF
--- a/base/src/com/thoughtworks/go/util/TriState.java
+++ b/base/src/com/thoughtworks/go/util/TriState.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2015 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


matt-richardson:
Why the change of agent_name? 
I would've thought that this would cause issues with multiple agents on one host?

ketan:
What has changed is the name of the attribute to represent the `hostname` attribute in the `<agent/>` xml tag. The value continues to be the same. In short the following xml snippet - 

```xml
    <agent hostname="go-agent-119.example.com" ipaddress="192.168.68.73" uuid="01cbed36-0e83-11e5-b41a-6003088b7702" />
```

will map to the following json -

``` json
{
  "uuid": "01cbed36-0e83-11e5-b41a-6003088b7702",
  "hostname": "go-agent-119.example.com",
  "ip_address": "192.168.68.73",
  ...
```

timoa:
Personally, I use 3 agents by host (Mac Mini) because I have some tasks with node.js and it use only 1 core by job. With 3 agents, I use 3 cores and keep 1 for OSX.

I'm not sure if it's a good idea to rename it hostname...

timoa:
Oh OK :)

ketan:
Also, I must add that this affects the [new API](https://github.com/gocd/gocd/issues/1100), and not the old API.

zabil:
Reviewed :+1:

PR #1224, opened 2015-06-15
Cross browser testing with jasmine

jyotisingh:
Review the PR, but do not merge it yet. 
As a part of this PR, some of tasks are deleted and command to run jasmine test has changed, so build will start failing. I need to update pipelines in build.go.cd to reflect the changes.


server/webapp/WEB-INF/rails.new/config/environments/test.rb
```diff
@@ -5,7 +5,7 @@
   # test suite. You never need to work with it otherwise. Remember that
   # your test database is "scratch space" for the test suite and is wiped
   # and recreated between test runs. Don't rely on the data there!
-  config.cache_classes = true
+  config.cache_classes = false
```
ketan: Is this change intentional? This is known to slow down tests quite a bit.

ketan:
Looks good - can you also look into this -

`$ git grep jsunit | grep -v 'test-resources' | grep -v 'test/data'`

and this - 

```
% find -iname '*jsunit*'
./server/jsunit.xml
./server/webapp/WEB-INF/rails.new/spec/javascripts/helpers/jsunitHelpersForMigratedTests.js
```

Issue #1216, opened 2015-06-12
secure env. vars for environments dont seem to work


srinivasupadhya:
related to #194 

steps:

create new environment with a secure value:
![1](https://cloud.githubusercontent.com/assets/1587661/8128054/55aa5de6-1119-11e5-8545-f6087e32833c.png)

config:
![2](https://cloud.githubusercontent.com/assets/1587661/8128055/56cb4ab4-1119-11e5-8d70-91276b9264b6.png)

open edit pop-up & save again (no changes made):
![3](https://cloud.githubusercontent.com/assets/1587661/8128057/587bf2b4-1119-11e5-9b2b-888cf5b545a5.png)

config:
![4](https://cloud.githubusercontent.com/assets/1587661/8128058/5a67c9ea-1119-11e5-8500-693165f29167.png)

open edit pop-up & save again (no changes made):
![5](https://cloud.githubusercontent.com/assets/1587661/8128061/5cade34c-1119-11e5-9266-a2e81357af99.png)

config:
![6](https://cloud.githubusercontent.com/assets/1587661/8128062/5ebb25f0-1119-11e5-8517-f57ef56d91c4.png)

values for 1st, 2nd & 3rd change:
![7](https://cloud.githubusercontent.com/assets/1587661/8128064/6075cae4-1119-11e5-9dd2-95c970c08f07.png)

PR #1225, opened 2015-06-15
Fix for the bug #1216 (related to #194)

juhijariwala:


server/webapp/WEB-INF/rails.new/app/assets/javascripts/environment_variable_edit.js
```diff
@@ -79,3 +79,35 @@ var EnvironmentVariableEdit = function (options) {
         resetLink.toggle();
     })
 };
+
+var EnvironmentVariableAddRemove = function(parentElement, options) {
```
mdaliejaz: missing test? :)

PR #1227, opened 2015-06-15
Default to using jetty6 server in prod mode.

ketan:


zabil:
This PR needs to merged only if performance tests on jetty 9 fails.

ketan:
I'm going to close this PR since we're going to be defaulting to jetty9 for 15.2.

PR #1228, opened 2015-06-15
Upgrade jetty to 9.2.11.v20150529 from 9.2.3.v20140905

ketan:


zabil:
@jyotisingh has a patch with the version she's using for testing along with some jetty changes. 

We'll should be merging that one instead of this. Can you close this?

PR #1231, opened 2015-06-16
#851 - list authentication plugins on login page

srinivasupadhya:


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRegistry.java
```diff
@@ -44,6 +48,7 @@ public void registerPlugin(String pluginId, AuthenticationPluginConfiguration co
 
     public void deregisterPlugin(String pluginId) {
         pluginToConfigurationMap.remove(pluginId);
+        pluginsThatSupportsWebBasedAuthentication.remove(pluginId);
         pluginsThatSupportsPasswordBasedAuthentication.remove(pluginId);
         pluginsThatSupportsUserSearch.remove(pluginId);
```
jyotisingh: There are too many of these internal lists being maintained for different things like supportsUserSearch, supportWebBasedAuth or basic-auth etc. Firstly, I think all auth-plugins should allow user search, I do not see a point in making it optional as far as the end-point is concerned. Sure, if the plugin does not want to implement it, it could always return an empty list. 

pluginsThatSupportsWebBasedAuthentication would come into picture for displaying login image for plugins on the login screen. Since its a user driven action, does it really make sense to cache that info?

I am still undecided on pluginsThatSupportsPasswordBasedAuthentication list, as I do realise that it will help with the performance when making curl requests.

PR #1238, opened 2015-06-17
Fixup agent API (old and new) authorization and improve response body on API errors.

ketan:


arvindsv:
You decided to not create a new role, to pass on auth to Rails?

server/src/com/thoughtworks/go/server/security/BasicProcessingFilterEntryPoint.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.List;
+
+public class BasicProcessingFilterEntryPoint implements AuthenticationEntryPoint, InitializingBean {
+    private String realmName;
+
+    public void afterPropertiesSet() throws Exception {
+        Assert.hasText(realmName, "realmName must be specified");
```
zabil: Can't we inject this in the constructor, instead of having an unused get and assertion for this set.

server/src/com/thoughtworks/go/server/security/BasicProcessingFilterEntryPoint.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public void commence(ServletRequest request, ServletResponse response, AuthenticationException authException)
+            throws IOException, ServletException {
+        HttpServletResponse httpResponse = (HttpServletResponse) response;
+        httpResponse.addHeader("WWW-Authenticate", "Basic realm=\"" + realmName + "\"");
```
zabil: Can we use String.format for readability? Applicable at other parts too

server/src/com/thoughtworks/go/server/security/BasicProcessingFilterEntryPoint.java
```diff
@@ -0,0 +1,79 @@
...
+
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.List;
+
+public class BasicProcessingFilterEntryPoint implements AuthenticationEntryPoint, InitializingBean {
```
zabil: No tests?

zabil:
Can we also have some spec tests for authentication filters as this deals with security.

ketan:
> You decided not to create a new role, to pass on auth to Rails?

That is right, the spring filters were doing multiple things (authentication against 3rd party services, authorization etc). It was (unfortunately) easier to add a filter in spring than delegate what the other filters in the chain were doing into rails.

zabil:
Can we set up some logging on new the filters (debug and info) I think it's important to  know about the the authentication failures.

Issue #1212, opened 2015-06-11
Agent API improvements


zabil:
Brand new API for handling agents in Go
Covers parts of 
#849, #264, #958, #1100 

 #1122

Adds support for, changing host name, adding resources, enabling or disabling agents.

zabil:
Testing notes on 15.2.0-2067 (07d65665c731a615f6a6943cce0fd03440eb4a9e)

* PATCH to /api/agents/{uuid} with invalid param { enable : "invalid" } returns a 200 and does not change the state.

@arvindsv @ketan We need to take a call on how to handle addition of resources to the agent. In in the current PATCH for resources there's no locking.

Option 1
If we are to implement optimistic locking. We have to force a GET and PUT operation with etags (throwing a conflict if the etag version changes) 
  
Limit operations with PATCH
e.g 
{ command: 'addResource', data : resource }
{ command: 'deleteResource', data : resource }

But that means coming up with a consistent PATCH operation all operations on the resource.

arvindsv:
> IMO it's a call if we need to support optimistic or pessimistic locking on a resource based on it's usage.

I don't mind leaving it with no locks. We have a standard way to upgrade APIs now (with the negotiation), and we can add it if we need it. Otherwise, we can try and make it consistent by having only one command. Say, a "changeTo" command, which will change the current resources to whatever the data shows.

I don't mind leaving it the way it is, knowing we can upgrade it if needed.

ketan:
> Limit operations with PATCH
> e.g 
> { command: 'addResource', data : resource }
> { command: 'deleteResource', data : resource }

I thought about it a bit, and it feels more like perform an RPC than a PATCH via REST. I'm not a fan of this approach.

ketan:
> I don't mind leaving it with no locks. We have a standard way to upgrade APIs now (with the negotiation), and we can add it if we need it.

In the interest of consistency – in some config change operations, we will need to perform this locking. I'd much rather start now than have multiple versions of APIs for various config resources which vary by just having a config header. This IMO is a very small change to make, and I can submit a PR if @zabil  and @arvindsv are on board.

arvindsv:
> This IMO is a very small change to make

Can you explain how this will be, with an example like @zabil did? I ask because you said you don't like the approach @zabil suggested.

I'd also like to put @jyotisingh on the spot and make her the final decision maker on decisions such as this. :) I'm just someone with opinions.

ketan:
[...]

PR #1254, opened 2015-06-22
Improved error handling when submitting a non-boolean for a boolean value.

ketan:
See https://github.com/gocd/gocd/issues/1212#issuecomment-114008418


ketan:
From @zabil on https://github.com/gocd/gocd/issues/1212#issuecomment-114414861

> Verified on experimental release 15.2.0-2149 (d0369fb)

PR #1260, opened 2015-06-23
Use `#to_hash` instead of `#to_json` for conversion to json and hashes for testing.

ketan:


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/agent_representer.rb
```diff
@@ -38,8 +38,16 @@ class AgentRepresenter < ApiV1::BaseRepresenter
     property :getStatusForDisplay, as: :status
     property :getOperatingSystem, as: :operating_system
     property :free_space, exec_context: :decorator
-    property :getResources, as: :resources
-    property :getEnvironments, as: :environments
+    property :resources, exec_context: :decorator
+    property :environments, exec_context: :decorator
+
+    def resources #because you know - java.util.ArrayList
```
arvindsv: :)

PR #1261, opened 2015-06-23
Port over users API.

ketan:
Merge this after #1260.

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/UserService.java
+++ b/server/src/com/thoughtworks/go/server/service/UserService.java
@@ -16,22 +16,8 @@
 
 package com.thoughtworks.go.server.service;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
-
 import com.thoughtworks.go.config.*;
-import com.thoughtworks.go.config.GoConfigFileDao;
-import com.thoughtworks.go.domain.NotificationFilter;
-import com.thoughtworks.go.domain.NullUser;
-import com.thoughtworks.go.domain.StageConfigIdentifier;
-import com.thoughtworks.go.domain.User;
+import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.domain.Users;
 import com.thoughtworks.go.domain.exception.ValidationException;
 import com.thoughtworks.go.i18n.LocalizedMessage;
@@ -53,11 +39,14 @@ import com.thoughtworks.go.server.transaction.TransactionTemplate;
 import com.thoughtworks.go.serverhealth.HealthStateScope;
 import com.thoughtworks.go.serverhealth.HealthStateType;
 import com.thoughtworks.go.util.Filter;
+import com.thoughtworks.go.util.TriState;
 import com.thoughtworks.go.util.comparator.AlphaAsciiCollectionComparator;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.support.TransactionCallbackWithoutResult;
+
+import java.util.*;
 
 @Service
 public class UserService {
@@ -129,6 +118,40 @@ public class UserService {
             }
         }
         return false;
+    }
+
+    public User update(final User user, TriState enabled, TriState emailMe, String email, String checkinAliases, LocalizedOperationResult result) {
+        if (enabled.isTrue()) {
+            user.enable();
+        }
+
+        if (enabled.isFalse()) {
+            user.disable();
+        }
+
+        if (email != null) {
+            user.setEmail(email);
+        }
+
+        if (checkinAliases != null) {
+            user.setMatcher(checkinAliases);
+        }
+
+        if (emailMe.isTrue()) {
+            user.setEmailMe(true);
+        }
+
+        if (emailMe.isFalse()) {
+            user.setEmailMe(false);
+        }
+
+        try {
+            saveOrUpdate(user);
+        } catch (ValidationException e) {
+            result.badRequest(LocalizedMessage.string("USER_FIELD_VALIDATIONS_FAILED", e.getMessage()));
+        }
+
+        return user;
     }
 
     public void enable(List<String> usernames, LocalizedOperationResult result) {
--- a/server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
+++ b/server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
@@ -16,18 +16,8 @@
 
 package com.thoughtworks.go.server.service;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import javax.servlet.http.HttpServletResponse;
-
 import com.thoughtworks.go.config.*;
-import com.thoughtworks.go.config.GoConfigFileDao;
-import com.thoughtworks.go.domain.NotificationFilter;
-import com.thoughtworks.go.domain.NullUser;
-import com.thoughtworks.go.domain.StageConfigIdentifier;
-import com.thoughtworks.go.domain.StageEvent;
-import com.thoughtworks.go.domain.User;
+import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.domain.Users;
 import com.thoughtworks.go.domain.config.Admin;
 import com.thoughtworks.go.domain.exception.ValidationException;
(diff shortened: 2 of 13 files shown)


ketan:
PR for api docs - gocd/api.go.cd#6

zabil:
I noticed that a bad request  (400) , if you send malformed json returns as an html and not json.

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -70,6 +71,7 @@
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 import static org.junit.matchers.JUnitMatchers.containsString;
+import static org.mockito.Mockito.verify;
```
zabil: The verify shows up as unused

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -609,6 +611,73 @@ public void shouldDisableAdminSelectionWhenUserIsMemberOfAdminRole() throws Exce
         assertThat(userService.getAdminAndRoleSelections(Arrays.asList("foo", "quux")).getAdminSelection(), is(new TriStateSelection(Admin.GO_SYSTEM_ADMIN, TriStateSelection.Action.add, false)));
     }
 
+    @Test
+    public void updateShouldUpdateEnabledState() throws Exception {
```
zabil: Can we stick to the convention of "shouldUpdateUser..."

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -609,6 +611,73 @@ public void shouldDisableAdminSelectionWhenUserIsMemberOfAdminRole() throws Exce
...
+        userDao.saveOrUpdate(new User("admin-user"));
+        userDao.saveOrUpdate(new User("user-1"));
+        userService.enable(Arrays.asList("user-1"), new HttpLocalizedOperationResult());
+
+        HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
+        userService.update(new User("user-1"), TriState.FALSE, TriState.UNSET, null, null, result);
```
zabil: Ok, I am not user we should pass the user object over here.
We are are re-assigning this parameter in the service. Why can't we just take in the user name.
We don't reload the user name in the service.

server/src/com/thoughtworks/go/server/service/UserService.java
```diff
@@ -131,6 +120,40 @@ private boolean userNameListContainsAdmin(List<String> enabledUserNames) {
...
+    public void update(User user, TriState enabled, TriState emailMe, String email, String checkinAliases, LocalizedOperationResult result) {
+        if (enabled.isTrue()) {
+            enable(Arrays.asList(user.getName()), result);
+        }
+
+        if (enabled.isFalse()) {
+            disable(Arrays.asList(user.getName()), result);
+        }
+
+        user = findUserByName(user.getName());
```
zabil: Prefer us not doing this as the controller has loaded the user.
Or change the method to take in the username and then do the load.

ketan: We're having reload the user because the enable/disable happens in a transaction of its own, the other updates happen in a different transaction.

I'm happy to hear how to fix this properly though.

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
ketan: Can I change the implementation to this -

```java
    public User update(User user, TriState enabled, TriState emailMe, String email, String checkinAliases, LocalizedOperationResult result) {
        if (enabled.isTrue()) {
            user.enable();
        }

        if (enabled.isFalse()) {
            user.disable();
        }

        if (email != null) {
            user.setEmail(email);
        }

        if (checkinAliases != null) {
            user.setMatcher(checkinAliases);
        }

        if (emailMe.isTrue()) {
            user.setEmailMe(true);
        }

        if (emailMe.isFalse()) {
            user.setEmailMe(false);
        }

        try {
            saveOrUpdate(user);
        } catch (ValidationException e) {
            result.badRequest(LocalizedMessage.string("USER_FIELD_VALIDATIONS_FAILED", e.getMessage()));
        }

        return user;
    }
```

Notice that it updates and mutates the same object and attempts to to save it in a single transaction.

This prevents a double load from the controller as well as from inside the service. The service does exactly what it advertises "update", no more.

zabil: Yes this makes sense.
I'd also make the user param final because this way of saving the user sadly unconventional because of our jruby way of doing things .

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -609,6 +611,73 @@ public void shouldDisableAdminSelectionWhenUserIsMemberOfAdminRole() throws Exce
...
+        userService.update(new User("user-1"), TriState.UNSET, TriState.UNSET, "foo@example.com", null, result);
+        assertThat(result.isSuccessful(), is(true));
+
+        assertThat(userService.findUserByName("user-1").getEmail(), is("foo@example.com"));
+
+        result = new HttpLocalizedOperationResult();
+        userService.update(new User("user-1"), TriState.TRUE, TriState.UNSET, "", null, result);
+        assertThat(userService.findUserByName("user-1").getEmail(), is(""));
+        assertThat(result.isSuccessful(), is(true));
+    }
```
zabil: The new line police is upset.

ketan:
@zabil - addressed all your feedback. Ready to merge. Also rebased the PR for backup api on top of this.

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -609,6 +611,73 @@ public void shouldDisableAdminSelectionWhenUserIsMemberOfAdminRole() throws Exce
...
+
+    @Test
+    public void updateShouldUpdateEmailMeState() throws Exception {
+        userDao.saveOrUpdate(new User("user-1"));
+
+        User user = userService.findUserByName("user-1");
+        user.setEmailMe(false);
+        userDao.saveOrUpdate(user);
+
+        userService.enable(Arrays.asList("user-1"), new HttpLocalizedOperationResult());
```
zabil: Should we enable the user for an update?

I think there are too many asserts in the test cases can't we split it up like this?

```
    @Test
    public void shouldUpdateEmailMeStateToTrue() throws Exception {
        User user = new User("user-1");
        user.setEmailMe(false);
        addUser(user);

        userService.update(user, TriState.UNSET, TriState.TRUE, null, null, result);

        assertThat(result.isSuccessful(), is(true));
        assertThat(userService.findUserByName("user-1").isEmailMe(), is(true));

    }
    @Test
    public void shouldUpdateEmailMeStateToFalse() throws Exception {
        User user = new User("user-1");
        user.setEmailMe(false);
        addUser(user);

        userService.update(user, TriState.UNSET, TriState.FALSE, null, null, result);

        assertThat(result.isSuccessful(), is(true));
        assertThat(userService.findUserByName("user-1").isEmailMe(), is(false));
    }
```

Also use the ```addUser``` method

ketan: Technically it's a tristate, we'd need 3 different tests for each of the tristates, plus 2 additional tests for EACH for email and checkinAliases (with null values and without null) I feel these tests are being too verbose, and perhaps don't as much value as I'd thought.

WDYT?

zabil: I know,
 But I found it tough to read one method doing a lot of operations.

 I feel it's cleaner to have one method call and one assertion for the
result of that method call. So that I can easily map what the assertion is
testing.

server/test/integration/com/thoughtworks/go/server/service/UserServiceIntegrationTest.java
```diff
@@ -341,7 +337,7 @@ public void shouldEnableUsers() throws Exception {
         HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
```
zabil: Noticed that the 
```
userService.enable(Arrays.asList("user_one"), new HttpLocalizedOperationResult());
```

On this method takes a new instance of ```new HttpLocalizedOperationResult()``` it should be taking in ```result``` instead.

ketan: That is not my change :)

zabil: True,
but a broken window ;-)

ketan: Can we call this done now?

zabil:
Verified on experimental release15.2.0-2219 (dae53017b0071763841d23c48a0c1d5e29dc661e)

PR #1262, opened 2015-06-24
15.3 - Improve perf logging

arvindsv:
* Better logging of scheduling subsystem. Using trackingID to corelate across threads.
* Agent communication is now logged.
* Web request logging now works even for gzipped content.
* Job identifiers at the end (in final log statements) to help parsing.

### DIFF
--- a/jetty6/src/com/thoughtworks/go/server/util/Jetty6Response.java
+++ b/jetty6/src/com/thoughtworks/go/server/util/Jetty6Response.java
@@ -18,6 +18,8 @@ package com.thoughtworks.go.server.util;
 
 import org.mortbay.jetty.Response;
 
+import javax.servlet.ServletResponseWrapper;
+
 public class Jetty6Response implements ServletResponse {
     private javax.servlet.ServletResponse servletResponse;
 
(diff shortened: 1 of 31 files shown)


arvindsv:
@jyotisingh: I've had to make some changes to the perf logging to help visualization. I've parsed the output (on the way to converting it to a graphical format) using the output after these changes.

However, since this contains some code changes in the scheduler, I'll understand if you want to hold on to this change for a bit. Let me know. That'll give me some time to finish the visualization. Otherwise, I'll need to hurry and finish it in the next day or two, so that we can use it this time.

jyotisingh:
At first glance the changes do not look risky, so I think you could continue with the visualisation work. I will need to have a closer look at the changes before merging them.

ketan:
@arvindsv @jyotisingh do we still need this?

jyotisingh:
Yes, we need this. @arvindsv is still working on visualisation for these logs so I was expecting some more changes to this.

arvindsv:
Worked on this a bit with about 500MB of logs. It was a little too much for a browser to visualize. I'm thinking of extracting important pieces from the logs and visualizing those instead. I'll talk to some of you to see what you all think, later.

arvindsv:
Closing this. I'll reopen it when I have something better.

PR #1263, opened 2015-06-24
945

mdaliejaz:


common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -140,9 +134,9 @@ public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
         return allowed;
     }
 
-    boolean hasOneMaterialUseBaseFolder() {
+    public boolean hasOneMaterialUseBaseFolder() {
         for (Material material : this) {
-            if (material.getFolder() == null) {
+            if (material.getFolder() == null && material instanceof SvnMaterial) {
```
arvindsv: Can we do something other than an instanceof check? The tests are not very useful. For instance, they don't say what the behavior is when there are multiple materials. What does hasOneMaterialUseBaseFolder mean? If there's a dependency material and a git material, what should the behavior be? What about dependency and svn?

mdaliejaz:
We might not require this code at all. But before removing, I need to investigate it a little more.
Closing this PR. Will send another PR.

PR #1265, opened 2015-06-24
fix #993 by skipping empty svn log entries

### LINKED ISSUE #993: "Modification check failed for material" with restrictive subversion permissions
If I configure a subversion material using credentials that do not grant permission to access all subpaths of the material URL, then go fails to determine the current revision of the material, showing the error message "Unable to parse svn log output".

This happens because some revisions in the history of my subversion repository do not modify any subpaths to which the authenticated user has access.

In such a case, `svn log --xml` produces output like

    <logentry
       revision="123">
    </logentry>

which causes `SvnLogXmlParser` to fail [here] (https://github.com/gocd/gocd/blob/cef918f39f513616be24fef34ec9c944ac193556/common/src/com/thoughtworks/go/util/SvnLogXmlParser.java#L66) by assuming the presence of child elements in the `<logentry>`.

norcnorc:


common/src/com/thoughtworks/go/util/SvnLogXmlParser.java
```diff
@@ -56,29 +56,36 @@
         for (Iterator iterator = logEntries.iterator(); iterator.hasNext();) {
             Element logEntry = (Element) iterator.next();
 
-            modifications.add(parseLogEntry(logEntry, path));
+            modification = parseLogEntry(logEntry, path);
```
zabil: Missing a variable declaration here.
```
modification = parseLogEntry(logEntry, path);
```
Should be
```
Modification modification = parseLogEntry(logEntry, path);
```

common/src/com/thoughtworks/go/util/SvnLogXmlParser.java
```diff
@@ -56,29 +56,36 @@
...
     }
 
     private Modification parseLogEntry(Element logEntry, String path) throws ParseException {
+        Element logEntryPaths = logEntry.getChild("paths");
+        if (logEntryPaths == null) {
+            /* Path-based access control forbids us from learning
+             * details of this log entry, so skip it. */
+            return null;
+        }
+
```
zabil: Can we write a test for this at?
https://github.com/gocd/gocd/blob/81701742adecc398e11d2a04443eff3f7bfc9cff/common/test/unit/com/thoughtworks/go/util/SvnLogXmlParserTest.java

zabil:
Closing this due to inactivity. Please feel free to open it again with fixes.

Issue #1012, opened 2015-04-09
Unable to change configuration after making one of the pipelines use a template


zabil:
https://groups.google.com/d/msg/go-cd/fgLSGOavVFo/fZT_3zIZlUoJ

zabil:
Finding it tough to replicate this.

ketan:
Steps to reproduce from the original [mail thread](https://groups.google.com/forum/#!msg/go-cd/fgLSGOavVFo/iTM71QRQqSYJ) - 

1.  Create a template that has no environment variables defined.  
2.  Create two pipelines off of that template. 
3.  Add a parameter to ONE of the pipelines using that template.  
4.  Update the template and create an environment variable, and assign the parameter from the pipelines to that environment variable on the template.  When you try to save the template, it will start throwing the error.  

At this point, try going to the pipeline that doesn't have the parameter defined and try creating it.  Go back to the template and try to create the environment variable again, it should still throw the error and every time you press save, it will append the variable names and the error will get longer and longer.

zabil:
this can be replicated on experimental release15.2.0-2153 (c935e43e651a2a3591792f167ac87023c5491b2e)

chopraapooja:
verified on Go v15.2.0-2177

PR #1269, opened 2015-06-25
#1012 The ReaderState enums are static. 

zabil:
We are not cleaning up the pattern when there's an exception in parameters substitution. This causes the pattern to remain in the string builder resulting in an endless append cycle.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/preprocessor/ParamStateMachine.java
+++ b/config/config-api/src/com/thoughtworks/go/config/preprocessor/ParamStateMachine.java
@@ -34,8 +34,11 @@ public class ParamStateMachine {
         IN_PATTERN {
             ReaderState interpret(char ch, ParamHandler paramsHandler) {
                 if (ch == CHAR_CURL_CLOSE) {
-                    paramsHandler.handlePatternFound(pattern);
-                    pattern.setLength(0);
+                    try {
+                        paramsHandler.handlePatternFound(pattern);
+                    } finally {
+                        pattern.setLength(0);
+                    }
                     return NOT_IN_PATTERN;
                 } else {
                     pattern.append(ch);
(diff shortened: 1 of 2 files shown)


arvindsv:
Checked. Works. Tried with a config like this:

```xml
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="75">
  <server artifactsdir="artifacts" commandRepositoryLocation="default" serverId="d12da3c1-c09d-4fe4-a06c-716f5a0ca461" />
  <pipelines group="defaultGroup">
    <pipeline name="P1" template="Temp1">
      <materials>
        <git url="/tmp/test.git" />
      </materials>
    </pipeline>
    <pipeline name="P2" template="Temp1">
      <params>
        <param name="PARAM1">paramvalue</param>
      </params>
      <materials>
        <git url="/tmp/test.git" />
      </materials>
    </pipeline>
  </pipelines>
  <templates>
    <pipeline name="Temp1">
      <stage name="defaultStage">
        <jobs>
          <job name="defaultJob" />
        </jobs>
      </stage>
    </pipeline>
  </templates>
</cruise>
```

If you try to add an environment variable when the config is like this, it fails. Trying again, at this point looks ok.

If you go back to pipeline P1, add that param and come back, you can now create that environment variable (`ENV => #{PARAM1}`) in the template.

However, every time the save fails in the template, the entered environment variable goes away. Might be related to the secure environment variable story played earlier.

zabil:
>However, every time the save fails in the template, the entered environment variable goes away. Might be related to the secure environment variable story played earlier.

That's what we thought, but this is there in older builds too (14, 15.1), Will log it as a bug.

Issue #1199, opened 2015-06-08
could not fetch artifact with + symbol in file name


rarunkum:
The automation failure screenshots are timestamped with +0800 in them to represent the timezone. These artifacts are listed under the menu, but when we try to fetch it throughs an error. And from the error its clear that the + has been replaced with a space.
![screenshot 2015-06-08 09 29 08](https://cloud.githubusercontent.com/assets/1950489/8027088/78129830-0dc2-11e5-9e80-fff9cdabe25a.png)
![screenshot 2015-06-08 09 30 04](https://cloud.githubusercontent.com/assets/1950489/8027089/7e93cfb2-0dc2-11e5-9c7e-6bda7a8ca684.png)

ketan:
If you were to URL encode the `+` to `%2B` does that work as expected?

ketan:
Oh, never mind my previous commit. I can see what's happening.

ketan:
Since @juhijariwala and me paired on this. Would someone mind testing this out?

The following script may be useful. Publish the directory `foo` as an artefact.

```
mkdir -p foo/bar foo/bar+boo
uuid | tee foo/bar/foo+bar.txt
uuid | tee foo/bar+boo/foo.txt
uuid | tee foo/bar+boo/foo+bar.txt
uuid | tee foo/bar+boo/foo%bar.txt
```

chopraapooja:
Testing in progress

chopraapooja:
Verified

PR #1274, opened 2015-06-26
Could not fetch artifact with '+' symbol in filename(#1199)

juhijariwala:
Solution:
Added escape function in the URLRewrite rule to 
match encoded character.


zabil:
Can't we write a test for this [here](https://github.com/gocd/gocd/blob/master/server/test/integration/com/thoughtworks/go/server/web/UrlRewriterIntegrationTest.java)?

PR #1276, opened 2015-06-27
1133 api validations rails

tomzo:
1st part of #1133 

This contains config-api changes so that configuration repositories can be understood by [services higher](https://github.com/gocd/gocd/issues/1133#issuecomment-116810715).

First 25 commits were reviewed by @arvindsv [here](https://github.com/gocd/gocd/issues/1133#issuecomment-112180237)

I have fixed most of that in next few. There is no duplication at this point. However MergeEnvironmentConfig and MergePipelineConfigs are still here. Now they are fully implemented.

### Major API changes

 * CruiseConfig is an interface
 * BasicCruiseConfig is the only class implementing CruiseConfig. It has nested strategy class
 * EnvironmentConfig is an interface. Implemented by BasicEnvironmentConfig and MergeEnvironmentConfig
 * PipelineConfigs is an interface. Implemented by BasicPipelineConfigs and MergePipelineConfigs

#### Added

 * PartialConfig class, it is like a smaller CruiseConfig. But contains only environments and pipelines in groups. This is what we agree to keep in separate configuration sources, there can be many instances used at once.
 * added new interface ConfigOrigin, which indicates where configuration instance comes from.
 * added new ConfigOrigin member in PipelineConfig, BasicPipelineConfigs, BasicEnvironmentConfig, PartialConfig

### Work done here

 * I have updated config-api. There are new unit tests to cover more test cases and unit tests for the new types.
 * Merge* classes and merge strategy are implemented to handle updates of configuration so that modifications are stored in local (file) part of configuration.
 * New **validations** are added because of extended domain model. Each is described in commit message. E.g. do not reference elements of config repository from a local config file.
 * Finally **rails** is updated so that it uses constructors of proper classes instead of what is now interface. E.g. `BasicCruiseConfig.new` instead of `CruiseConfig.new`

### Effects

This should not break **any** existing functionality and does not actually add any new features.
*As far as I could test* everything passes. 
This makes merely a room for bringing #1133 to life.
None of the Merge* classes or merge strategy are actually used during any service operation. They are only unit tested to some extent.

### Review

 * Please **ignore** any MergeCruiseConfig code which is gone in the end.
 * There are rather long and explanatory comments in each commit message.
 * I'd like to get some feedback, so I could do some corrections so that we can consider this *accepted*. All other branches are on top of it which is why I'd like to consider this stable.

If you do not want to review all commits at once then these would be good parts:
 * https://github.com/tomzo/gocd/compare/1133-api...tomzo:1133-api2 - fixes after first review + implemented editing of merged configurations
 * https://github.com/tomzo/gocd/compare/1133-api2...tomzo:1133-validations - added extra validations, one per commit, usually
 * https://github.com/tomzo/gocd/compare/1133-validations...tomzo:1133-api-validations-rails-pr - rails updates to fit with new api


srinivasupadhya:
you can ignore the [rspec](https://build.go.cd/go/tab/build/detail/build-linux-PR/22/build-server/1/rspec) failure (need to be setup yet) but [server](https://build.go.cd/go/tab/build/detail/build-linux-PR/22/build-server/1/server-split-runInstance-12) (look at 'Tests' tab) test is a genuine failure. (user: view, password: password)

```java
Failure	shouldThrowUpOnConfigSavePostValidationError_ViaMergeFlow
Unit Test Failure and Error Details (1)
Test:	shouldThrowUpOnConfigSavePostValidationError_ViaMergeFlow
Type:	Failure
Message:	Expected: is <false> but: was <true>
Expected: is <false>
     but: was <true>
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:237)
	at org.springframework.test.context.junit4.statements.RunBeforeTestClassCallbacks.evaluate(RunBeforeTestClassCallbacks.java:61)
	at org.springframework.test.context.junit4.statements.RunAfterTestClassCallbacks.evaluate(RunAfterTestClassCallbacks.java:71)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.run(SpringJUnit4ClassRunner.java:174)
Expected: is <false>
     but: was <true>
	at com.thoughtworks.go.server.service.GoConfigServiceIntegrationTest.shouldThrowUpOnConfigSavePostValidationError_ViaMergeFlow(GoConfigServiceIntegrationTest.java:659)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:74)
	at org.springframework.test.context.junit4.statements.RunAfterTestMethodCallbacks.evaluate(RunAfterTestMethodCallbacks.java:83)
	at org.springframework.test.context.junit4.statements.SpringRepeat.evaluate(SpringRepeat.java:72)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.runChild(SpringJUnit4ClassRunner.java:231)
```

tomzo:
Thanks. I'm on it already

tomzo:
Should be fine now. I fixed it and covered this failure by unit test.

common/test/unit/com/thoughtworks/go/domain/BasicPipelineConfigsTest.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static com.thoughtworks.go.util.DataStructureUtils.m;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.core.IsNull.nullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Created by tomzo on 6/23/15.
```
srinivasupadhya: not sure these comments should stay

common/test/unit/com/thoughtworks/go/domain/BasicPipelineConfigsTest.java
```diff
@@ -0,0 +1,59 @@
+package com.thoughtworks.go.domain;
```
srinivasupadhya: we add a license in all java files

tomzo:
It might be easier to review what is in the config-api knowing why is it here. I described services which rely on this [here](https://github.com/gocd/gocd/issues/1133#issuecomment-116810715)

arvindsv:
While reviewing this, I noticed the file `server/webapp/WEB-INF/rails.new/spec/views/admin/pipeline_groups/possible_groups_popup_sepc.rb` and renamed it to `b/server/webapp/WEB-INF/rails.new/spec/views/admin/pipeline_groups/possible_groups_popup_spec.rb`, which is probably why this PR is not merge-able without conflicts now.

config/config-api/src/com/thoughtworks/go/config/ValidationContext.java
```diff
@@ -110,6 +110,21 @@ public JobConfig getJob() {
         } else if (immediateParent.getClass().equals(klass)) {
             return (T) immediateParent;
         }
+        else
+        {
+            // added because of higher hierarchy of configuration types.
```
arvindsv: Will this be changed? I know you said "ignore changes related to Merge", but just wondering.

tomzo: No, it won't. It is still needed by MergeEnvironmentConfig and MergePipelineConfigs for the same reason

tomzo:
> ... this PR is not merge-able without conflicts now. 

I will pull from master and solve such things when you are ready to merge. Just let me know.

arvindsv:
After being quite intimidated by the number of files and commits ("Oh wow, I'm never going to be able to review this"), I pruned down the list to something manageable, ignoring tests and specs for now, and focusing on the non-test changes:

```
git diff --numstat  master..HEAD | egrep -v '_spec.rb|/test/' | sort -k1 -n | cut -f3 | xargs -I{} git --no-pager diff --color=always master..HEAD -- {} | less -R
```

This is manageable. As I see some code, I switch over to the tests using something like: `git diff master..HEAD config/config-api/test/com/thoughtworks/go/config/materials/MaterialConfigsTest.java` to see if it makes sense.

You're making me have to come up with new ways of reviewing code. :)

I haven't reached the end, yet, but it looks like most of the rest of the files I have are renames. So, shouldn't take a lot of time. I'll do them tomorrow, though.

arvindsv:
Once done with that, I'll take a look at the test code.

config/config-api/src/com/thoughtworks/go/config/ValidationContext.java
arvindsv: Ok.

tomzo:
> I haven't reached the end, yet, but it looks like most of the rest of the files I have are renames. So, shouldn't take a lot of time. I'll do them tomorrow, though.

Yes. This is true for next branches as well. A lot of renames in services as well.
On github you can compare to revision instead of branch to branch. That might help as well.

arvindsv:
Hey @tomzo,

I took another look at this, and came up with some thoughts. I don't think any of them are going to involve a big change. Just some food for thought:

- In some places like [this](https://github.com/tomzo/gocd/blob/1133-api-validations-rails-pr/config/config-api/src/com/thoughtworks/go/config/FetchTask.java#L353), [this](https://github.com/tomzo/gocd/blob/1133-api-validations-rails-pr/config/config-api/src/com/thoughtworks/go/config/materials/MaterialConfigs.java#L163) and [this](https://github.com/tomzo/gocd/blob/1133-api-validations-rails-pr/config/config-api/src/com/thoughtworks/go/config/BasicEnvironmentConfig.java#L65), the `toString` is implicitly used in a message. Can you change those to use a proper method, such as `forDisplay` or something like that? I feel that using `toString` can be a problem if someone changes it to be more descriptive. Especially because this is a value shown to the user. It's even alright if `forDisplay` calls `toString` for now.

- Is [this](https://github.com/tomzo/gocd/blob/1133-api-validations-rails-pr/config/config-api/src/com/thoughtworks/go/config/BasicEnvironmentConfig.java#L56) comment statement true? Because, just below that, you check for `isReferenceAllowed`, and not `equal`.


- `BasicEnvironmentConfig#validate` does not check for duplicate environment variables, whereas `MergeEnvironmentConfig#validate` does. I see that the old `EnvironmentConfig` code (in master) did not, as well. I guess it is somewhere else, then?

- `MergeEnvironmentConfig#setVariablesAttributes` is quite complicated. Is there something that can be done to simplify it? You asked that I ignore any MergeCruiseConfig code. Did that include MergeEnvironmentConfig as well? Is it work in progress too? What makes it so complicated? If the two methods `setPipelineAttributes` and `setVariablesAttributes` are so complicated, I don't mind moving them to a different class and testing them thoroughly. Will at least be good documentation for them.

- In MergePipelineConfigs, `getPartWithIndex` is not used, `getPartWithIndexForInsert` can be private. Also, that for loop showing up in many places there is quite irritating, but I understand why.

- Are the CruiseStrategy implementations going out into their own files? I think they should. Are you planning to have strategy implementations for environments and groups? Or, are they too different? I haven't seen your later changes. So, I don't know. But, I think removing the concept of a MergeEnvironmentConfig class can make usage simpler, I feel. If not, I suppose if the fact that one of those objects is a Merge class, and another is a Basic one is hidden from the eventual user of those classes, then it's ok. I just don't want code checking for `instanceOf MergeEnvironmentConfig`. That would make it hard to change.

I'm just going to give 15.2 a couple of days to finish, but I've finished review of this for all practical purposes.

Cheers,
Aravind

arvindsv:
In MaterialConfigsTest, the fact that a pipeline in one repo config can reference a pipeline in another repo config doesn't seem to be made clear. The closest one I can find is `shouldAllowToDependOnPipelineDefinedInConfigRepository_WhenInConfigRepository`, which is the same repo. Would be nice to have a test which uses a "different" repo (even though the code just uses isLocal). What do you think? Or, maybe the test mentioned earlier can itself change one of its repos?

tomzo:
Hi @arvindsv 

> In some places like this, this and this, the toString is implicitly used in a message.

I'l fix that to be forDisplay. I'll have to check higher as well because I recall writing something like it.

> Is this comment statement true? Because, just below that, you check for isReferenceAllowed, and not equal.

Perhaps I wasn't clear enough there. I meant that every `EnvironmentPipelineConfig pipelineRefConfig` in the for loop is defined in the configuration repository as pointed by this.origin. Then for each of pipelineRefConfig I search for `PipelineConfig` anywhere in the current configuration. Just to see where the `PipelineConfig` is defined. Error message is more clear I think.

> BasicEnvironmentConfig#validate does not check for duplicate environment variables, whereas MergeEnvironmentConfig#validate does. I see that the old EnvironmentConfig code (in master) did not, as well. I guess it is somewhere else, then?

I was guessing the same. But now I see I should check. Is it possible that in few places validation relies on xsd validation instead of checking at object level? 
I will add validation in `BasicEnvironmentConfig#validate` if it is not there.

### MergeEnvironmentConfig

> You asked that I ignore any MergeCruiseConfig code. Did that include MergeEnvironmentConfig as well? Is it work in progress too?

I meant MergeCruiseConfig only. I was planning on leaving MergeEnvironmentConfig and MergePipelineConfigs as they are now.

> MergeEnvironmentConfig#setVariablesAttributes is quite complicated. Is there something that can be done to simplify it? ...  What makes it so complicated? If the two methods setPipelineAttributes and setVariablesAttributes are so complicated, I don't mind moving them to a different class and testing them thoroughly. Will at least be good documentation for them.

They could be refactored and have more tests.  As far as I understand, these methods are used from UI when user has selected pipelines or env. variables for the environment. This [commit message](https://github.com/tomzo/gocd/commit/b5900755429740989ec3e12a22a33e7f03c17ff0) explains what these methods do.
Why they are complicated? Because UI (or any other service) does not know about `MergeEnvironmentConfig`. It talks to `EnvironmentConfig` only. When `EnvironmentConfig` is `MergeEnvironmentConfig` then there are parts which cannot be edited - e.g. user cannot remove a pipeline from environment when it is defined in configuration repo. These methods find out what user is trying to change.

I'd like to postpone working on this method until UI. I only initially tested it manually on development server. Is that OK?

> Are the CruiseStrategy implementations going out into their own files? I think they should. I haven't seen your later changes. So, I don't know.

They could be. But just to be clear - the strategies are not public. Only `BasicCruiseConfig` handles which strategy to use based on which constructor was used. 
I can move them to separate files. But do you think we can do it in next branches? Otherwise I will have to merge it 5-7 times higher and I will get conflicts almost each time because I added more to both these strategies later.

> Are you planning to have strategy implementations for environments and groups? Or, are they too different? 

It could be done and it would not impact anything above. As I said no service knows about alternative implementations of the EnvironmentConfig, PipelineConfigs or CruiseConfig
I'm not planning to do that though. Almost entire implementation would have to be in those strategies.

> ...But, I think removing the concept of a MergeEnvironmentConfig class can make usage simpler, I feel. If not, I suppose if the fact that one of those objects is a Merge class, and another is a Basic one is hidden from the eventual user of those classes, then it's ok. I just don't want code checking for instanceOf MergeEnvironmentConfig. That would make it hard to change.

As I said above - nobody is aware of Merge* classes. I agree that there should not be code that does `instanceOf MergeEnvironmentConfig` or alike. If we have to add such code then I think the strategies should be implemented for BasicEnvironmentConfig and BasicPipelineConfigs. I did all work except for UI already and such checking wasn't needed at any moment.

> In MergePipelineConfigs, getPartWithIndex is not used, getPartWithIndexForInsert can be private. Also, that for loop showing up in many places there is quite irritating, but I understand why

The implementation of MergePipelineConfigs and MergeEnvironmentConfig could be simplified by using something like [here](http://stackoverflow.com/questions/4896662/combine-multiple-collections-into-a-single-logical-collection) but that requires to add new dependency in config-api. 

> In MaterialConfigsTest, the fact that a pipeline in one repo config can reference a pipeline in another repo config doesn't seem to be made clear. The closest one I can find is shouldAllowToDependOnPipelineDefinedInConfigRepository_WhenInConfigRepository, which is the same repo. Would be nice to have a test which uses a "different" repo (even though the code just uses isLocal). What do you think? Or, maybe the test mentioned earlier can itself change one of its repos?

OK, I'll add something like that for clarity.

arvindsv:
> [About MergeEnvironmentConfig#setVariables] I'd like to postpone working on this method until UI. I only initially tested it manually on development server. Is that OK?

Sure.

> [About moving out the strategy classes] But do you think we can do it in next branches?

Yes. Let's remember to do that at the end, then.

> It could be done and it would not impact anything above. As I said no service knows about alternative implementations of the EnvironmentConfig, PipelineConfigs or CruiseConfig. I'm not planning to do that though. Almost entire implementation would have to be in those strategies.

> As I said above - nobody is aware of Merge* classes. I agree that there should not be code that does instanceOf MergeEnvironmentConfig or alike. If we have to add such code then I think the strategies should be implemented for BasicEnvironmentConfig and BasicPipelineConfigs. I did all work except for UI already and such checking wasn't needed at any moment.

Makes sense. As long as they're "hidden" and not known by most things, then it's ok.

tomzo:
Last changes I did
 * use `displayName()` instead of `toString()`
 * more clear `MaterialConfigsTests` case

Environment variable duplicated name checks are [here](https://github.com/gocd/gocd/blob/26852c9880a2d56bb0c238b4bdb5c560867f9c4f/config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L159). I did not have to add them in basic config.

**I am done with changes in this PR.
This can be merged.**

arvindsv:
Hi @tomzo, I am running tests before merging and I see a failure. `CruiseConfigTest#shouldCollectOriginErrorsFromMaterialConfigs_InMergedConfig` is failing. I guess one of the origins is null, and since there's now a call to displayName, it's failing. The tests are still running, and I'll let you know about any other failures.

tomzo:
OK,  I see it now.

I will fix in all places so that instead of
```
myOrigin.displayName()
```
there will be something like:
```
displayNameFor(myOrigin)
```

arvindsv:
Alright. Everything else seems to have passed.

tomzo:
Fixed.
Sorry about that. I thought I ran all tests before last commit.

arvindsv:
Done! Let's see how it goes. :) Congratulations!

Issue #945, opened 2015-03-11
Material clean up (eg: git clean -dff) does not happen if one of the materials is pipeline / package repository


webmaster128:
As seen earlier in https://github.com/gocd/gocd/issues/438, the option `Clean Working Directory` not seem to work as expected.

I have different C++ projects where `Clean Working Directory` is always unchecked. In some of these projects no directory cleanup is performed at all

```
[go] Start to prepare myproject/1142/build/1/Linux32 on ubuntu32 [/var/lib/go-agent] at Wed Mar 11 16:28:40 CET 2015

[go] Start updating farcaster at revision 3d5ef877cd3af8cf3e7b06c74ed47b487a8f84c1 from git@bitbucket.org:myuser/myproject.git
```

In other projects, build outputs are deleted

```
[go] Start to prepare myproject/97/my-stage/1/Linux32 on ubuntu32 [/var/lib/go-agent] at Wed Mar 11 16:44:16 CET 2015

[go] Keeping folder pipelines/myproject/cruise-output

[go] Keeping folder pipelines/myproject/somelib-source

[go] Deleting folder pipelines/myproject/build-myproject

[go] Keeping folder pipelines/myproject/myproject-source

[go] Start updating myproject at revision 2fa8b97dd9417e2a8b1b9dabb2fc2ad7d5c245d5 from git@bitbucket.org:myuser/myproject.git
```

The logic of what is deleted and what is kept looks good assuming I want my working directory to be cleaned. But I don't because re-building the entire C++ project takes 15-20 minutes.

srinivasupadhya:
[Here](https://github.com/gocd/gocd/blob/2b6c017b72ad427f517b324c3d633cd1410f7dac/common/src/com/thoughtworks/go/remote/work/BuildWork.java#L291) we simply delete the whole working directory (eg: `/var/lib/go-agent/pipelines/<pipeline-name>`) and prints `Cleaning working directory 'x' since stage is configured to clean working directory` if `cleanWorkingDir` is selected. Agent then does a fresh clone for all materials required. If its not selected we simply update the materials (if its already present) to the correct revision & run tasks. Are you seeing something other than this behavior?

webmaster128:
[...]

PR #1278, opened 2015-06-29
#945 - Clean up of unversioned folders happen as expected.

mdaliejaz:


common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -140,9 +134,9 @@ public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
         return allowed;
     }
 
-    boolean hasOneMaterialUseBaseFolder() {
+    public boolean hasOneMaterialUseBaseFolder() {
         for (Material material : this) {
```
zabil: I see this being public only for testing. 
It's not used anywhere else other than this class. Can't we modify existing integration tests? Or better modify the clean up method to be testable.

common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -140,9 +134,9 @@ public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
...
-            if (material.getFolder() == null) {
+            if (material.getFolder() == null && (material instanceof ScmMaterial || material instanceof PluggableSCMMaterial)) {
```
zabil: Shouldn't this logic be inside material?
e.g ```material.isSCM```
I'd even push ```material.getFolder() == null``` into a method ```material.hasDestinationFolder()``` if that's what the check is implying.

common/src/com/thoughtworks/go/config/materials/Materials.java
mdaliejaz: Sure, can do that. I just wanted to have extensive unit test around it. But I agree with you.

common/src/com/thoughtworks/go/config/materials/Materials.java
mdaliejaz: Material is an interface. Will do it through AbstractMaterial.

common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -142,7 +136,7 @@ public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
 
     boolean hasOneMaterialUseBaseFolder() {
         for (Material material : this) {
-            if (material.getFolder() == null) {
+            if (((AbstractMaterial) material).requiresDestinationFolder() && !((AbstractMaterial) material).hasDestinationFolder()) {
```
zabil: Can't we fuse this into one condition?
Check this inside hasDestinationFolder()

common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java
```diff
@@ -42,13 +39,15 @@
...
     protected CaseInsensitiveString name;
     protected String type;
+    private boolean requiresDestinationFolder;
     private Map<String, Object> sqlCriteria;
     private Map<String, Object> attributesForXml;
     private String pipelineUniqueFingerprint;
     private String fingerprint;
 
-    public AbstractMaterial(String typeName) {
+    public AbstractMaterial(String typeName, boolean requiresDestinationFolder) {
```
zabil: Handle this as a class definition and not an instance definition.
For e.g. the scm or pluggable scm instance cannot be created with false as it will break behavior. 
This must be a protected method which can be overriden in the class definition.

common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -142,7 +136,7 @@ public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
 
     boolean hasOneMaterialUseBaseFolder() {
         for (Material material : this) {
```
zabil: Can we change the name of this method?

common/test/unit/com/thoughtworks/go/domain/materials/AbstractMaterialTest.java
```diff
@@ -16,13 +16,13 @@
```
zabil: As discussed we should not have an abstract material test.
Instead, ```hasDestinationFolder ``` must be tested as a behavior of the test cases ```PluggableSCMMaterialTest``` or ```GitMaterialTest``` etc.

common/src/com/thoughtworks/go/config/materials/Materials.java
mdaliejaz: The two checks cannot be combined. For a pipeline & scm material with a destination folder we don't want to return true.

common/test/unit/com/thoughtworks/go/domain/materials/AbstractMaterialTest.java
mdaliejaz: :+1:

common/src/com/thoughtworks/go/config/materials/AbstractMaterial.java
```diff
@@ -187,4 +185,12 @@ public MaterialConfig config() {
     public Map<String, Object> getAttributes(boolean addSecureFields) {
         throw new RuntimeException("You need to implement this");
     }
+
+    public boolean hasDestinationFolder() {
+        return !StringUtil.isBlank(getFolder());
+    }
+
+    protected boolean requiresDestinationFolder() {
+        return requiresDestinationFolder;
```
zabil: Inline this variable

PR #1281, opened 2015-06-30
Get rid of some code that we can delegate down to scrollToFixed

ketan:


arvindsv:
@ketan: Do you want me to take a look at this before 15.3? Apart from the back to top link overlap thing, if the rest is just a refactor to use scrollToFixed, I'd rather leave it to 15.3, especially since this is not well tested.

If it needs to go in, in 15.2, I'd need to test it quite thoroughly (different browsers, different window sizes and seeing whether that  arrow hides itself properly, etc). I can do it. Or, if you can show it to someone else, that's fine too.

ketan:
Sounds good. I'm going to cherry-pick the fix for the back-to-top link overlap and we'll look at this for 15.3.

zabil:
@ketan is there any work pending on this?

ketan:
This is obsolete, closing

PR #1282, opened 2015-06-30
NA - turn off fall back on triangle resolution on fan-in error scenario

srinivasupadhya:


jyotisingh:
Please update this in release notes as well.

PR #1292, opened 2015-07-01
Port over backup API

ketan:
Merge this after #1261


ketan:
The API docs are here - https://github.com/gocd/api.go.cd/pull/9

PR #1294, opened 2015-07-01
Fanin performance

srinivasupadhya:


server/test/integration/com/thoughtworks/go/server/service/FanInPerformanceTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        configHelper.onTearDown();
+        dbHelper.onTearDown();
+    }
+
+    @Test(timeout = 240000)
+    @RunIf(value = DatabaseChecker.class, arguments = {DatabaseChecker.H2})
```
jyotisingh: I don't understand why do we need this check. Could you remove it?

Issue #1293, opened 2015-07-01
Validation errors are missing on editing template's environment variables.


chopraapooja:
**How to reproduce ?**
- Create a template with default stage.
- Make a pipeline use that template.
- Try to edit environment variables of template's default stage.

When I put the value of the variable without the name of variable(validation error). After clicking save, I got this -
![screen shot 2015-07-01 at 11 18 25 am](https://cloud.githubusercontent.com/assets/10117539/8449768/efde08fe-1ff1-11e5-981f-b17085eb8228.jpg)

**Ideally it should be like this -**

![screen shot 2015-07-01 at 5 40 50 pm](https://cloud.githubusercontent.com/assets/10117539/8454112/80113718-2018-11e5-9ab3-438fb2e38744.png)

The same thing happened again when I tried to put invalid environment variable name and trying to save it -
![screen shot 2015-07-01 at 1 12 01 pm](https://cloud.githubusercontent.com/assets/10117539/8449891/f2869f34-1ff2-11e5-93c1-547c61b4f45e.png)

**I am using Go v15.2.0-2177**

chopraapooja:
Bug - https://github.com/gocd/gocd/issues/1321

chopraapooja:
Verified in Go v15.2.0-2237

PR #1296, opened 2015-07-02
Fix for (#1293): Rendering validation errors on editing environment variables.

juhijariwala:


ketan:
If this is such a simple fix, can we see if we can improve the errors on other admin pages?

ketan:
Juhi tells me that there are no other pages where there's a similar problem. Merging AS-IS.

Issue #1018, opened 2015-04-10
Fail validation for label length of 0


mdaliejaz:
If a material name is used in the label like ${material_name[:0]}, the length of label is zero and the pipeline runs without a label. 

This should not be allowed. The length should be greater than 0.

arvindsv:
Can we make this a documented "warning"? Does this cause an issue?

mdaliejaz:
@arvindsv I think we can document it.

juhijariwala:
Use below XML to reproduce this issue: 
```xml
<?xml version="1.0" encoding="utf-8"?>
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="75">
  <server artifactsdir="logs" commandRepositoryLocation="default" serverId="dev-id">
    <security>
      <passwordFile path="../manual-testing/ant_hg/password.properties" />
      <roles>
        <role name="xyz" />
      </roles>
      <admins>
        <user>admin</user>
      </admins>
    </security>
  </server>
  <pipelines group="first">
    <authorization>
      <view>
        <user>operate</user>
      </view>
      <admins>
        <user>operate</user>
      </admins>
    </authorization>
    <pipeline name="pipeline1" labeltemplate="${material_name_1[:0]}${material_name_2[:00]}" isLocked="false">
      <materials>
        <hg url="../manual-testing/ant_hg/dummy" dest="dest_dir" materialName="material_name_1" />
        <git url="https://github.com/juhijariwala/biblioteca_library.git" dest="abc" materialName="material_name_2" />
      </materials>
      <stage name="up42_stage">
        <jobs>
          <job name="up42_job">
            <tasks>
              <exec command="ls" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
    <pipeline name="pipeline2" labeltemplate="${material_name_1[:1]}${material_name_2[:10]}" isLocked="false">
      <materials>
        <hg url="../manual-testing/ant_hg/dummy" dest="dest_dir" materialName="material_name_1" />
        <git url="https://github.com/juhijariwala/biblioteca_library.git" dest="abc" materialName="material_name_2" />
      </materials>
      <stage name="up42_stage">
        <jobs>
          <job name="up42_job">
            <tasks>
              <exec command="ls" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </pipelines>
</cruise>
[...]

PR #1299, opened 2015-07-03
(#1018) Added validation to check label template should not allow label length of zero.

juhijariwala:


config/config-api/src/com/thoughtworks/go/config/PipelineConfig.java
```diff
@@ -174,6 +176,11 @@ private void validateLabelTemplate() {
             return;
         }
 
+        if(validateLabelTemplateTruncation(labelTemplate)){
+            addError("labelTemplate", String.format("Invalid label. %s length of zero not allowed",labelTemplate));
```
ketan: Can we have a message that reads -

```
Length of zero not allowed on label "some-label" defined on pipeline "pipelineName".
```

PR #1300, opened 2015-07-03
JSON should be rendered with content type `application/json`.

ketan:
Currently being rendered as `text/plain`


arvindsv:
Some [tests](https://build.go.cd/go/tab/build/detail/build-linux-PR/36/build-server/1/server-split-runInstance-17) are failing.

Issue #1086, opened 2015-04-30
Support for Java 8


zabil:
Public [http://www.oracle.com/technetwork/java/eol-135779.html#Java6-end-public-updates](updates) for JDK 7 ends April 2015.

We need to start supporting Go on JDK 8

zabil:
We'll have to support both JDK7 and JDK8 for a few releases because our agents are run on a global JDK setup.

On the long run @ketan suggested we decouple our server or agents from running on a global JDK. With java 8 they've introduced [profiles](https://blogs.oracle.com/jtc/entry/a_first_look_at_compact) maybe try them out?

ketan:
There is another alternative in that we force a dependency on the system provided jdk8 (`java-1.8.0-openjdk` on centos) ubuntu still hasn't sorted its openjdk8 story yet, so I'm not sure what the solution there is (short of using `ppa:openjdk-r/ppa`)

zabil:
Strategy to be adopted
develop on Java 8 and cross compile to Java 7

```
-source 1.7 -target 1.7
```

Problems so far #1248

zabil:
Agents on build.go.cd keep running out of memory on JDK8
I suspect changes how the permgen is handled
https://dzone.com/articles/java-8-permgen-metaspace

This means we'll have to set an alternate flag to control the native space usage. Moving this to next release as it requires further testing.

zabil:
Fixed with https://github.com/gocd/gocd/commit/f768e5d7891720dafe6bd9f419ffaf28503e2e64

PR #1302, opened 2015-07-06
Remove cipher limitation

jyotisingh:
By default there are no restrictions for cipher suites.
When allow.ssl.config is set to Y, all the restrictions are gone by default (this is controlled by server/ssl.config file). Users could override ssl config to make it more restrictive or lenient, based on their needs by creating user.ssl.config under go-server config folder. The overrides file is backed up when a backup operation is initiated from Go. 

EDIT: Changed the way ssl settings can be configured based on feedback. It basically uses system properties now. The flags used are also renamed (Look in SystemEnvironment class for properties named as "go.ssl.<something>")


base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -160,6 +160,11 @@
     public static GoSystemProperty<String> GO_LANDING_PAGE = new GoStringSystemProperty("go.landing.page", "/pipelines");
 
     public static GoSystemProperty<Boolean> FETCH_ARTIFACT_AUTO_SUGGEST = new GoBooleanSystemProperty("go.fetch-artifact.auto-suggest", true);
+    public static GoSystemProperty<String> GO_SSL_CONFIG_FILE_PATH = new GoStringSystemProperty("go.ssl.config.path", "/ssl.config");
+    public static GoSystemProperty<String> USER_CONFIGURED_SSL_CONFIG_FILE_PATH = new GoStringSystemProperty("user.ssl.config.path", "user.ssl.config");
+    public static GoSystemProperty<String> TRANSPORT_PROTOCOL_TO_BE_USED_BY_AGENT = new GoStringSystemProperty("agent.transport.protocol", "TLSv1.2");
```
srinivasupadhya: i thought `SSLv3` was going to be the default

ketan: SSLv3 is no longer considered secure, but kept for backwards compatibility with older clients.

@jyotisingh - do we know what is set of user agents (browsers) that are supported by the current transport protocol and ciphers?

A good way to check would be to spin up a server on ec2 and point ssllabs at port 8154. Also ssllabs mentions a bunch of browser capabilities here - https://www.ssllabs.com/ssltest/clients.html

ketan: Also Java 7u25 does not support TLS v1.2, so we'd need to be careful about wording java 1.7 support.

jyotisingh: >> i thought SSLv3 was going to be the default

As @ketan mentioned, SSLv3 is not considered secure anymore. I am tempted to disable it by default (on server side), but decided against doing it as setups using java7 would stop working.  

>> do we know what is set of user agents (browsers) that are supported by the current transport protocol and ciphers? 

By default, since there are no restrictions, the server and client (agent/browser) could potentially decide on using an obsolete or a less secure cipher. Firefox and chrome usually complain about it. 
Anyway, I will try to put up the details you ask, it would be good to have that documented as I wont remember it after a few days :stuck_out_tongue: 

>> Also Java 7u25 does not support TLS v1.2, so we'd need to be careful about wording java 1.7 support.

Ack.

ketan:
@jyotisingh - should we consider upgrading the JRE that comes pre-packaged with the windows installer?

jetty9/src/com/thoughtworks/go/server/config/ConfigurableSSLSettings.java
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            String systemConfiguredSslConfigFile = systemEnvironment.get(SystemEnvironment.GO_SSL_CONFIG_FILE_PATH);
+            InputStream sslConfigStream = getClass().getResourceAsStream(systemConfiguredSslConfigFile);
+            String json = IOUtils.toString(sslConfigStream);
+            config = new Gson().fromJson(json, Config.class).setDefaults();
+
+            if (!StringUtil.isBlank(systemEnvironment.get(SystemEnvironment.USER_CONFIGURED_SSL_CONFIG_FILE_PATH))) {
+                File userConfiguredSslConfigFile = new File(systemEnvironment.configDir(), systemEnvironment.get(SystemEnvironment.USER_CONFIGURED_SSL_CONFIG_FILE_PATH));
+                if (userConfiguredSslConfigFile.exists()) {
+                    Config userDefinedConfig = new Gson().fromJson(IOUtils.toString(new FileInputStream(userConfiguredSslConfigFile)), Config.class);
```
ketan: Is there any reason that this is read from a config file instead of system properties? Unlike most other things that are populated directly using system environments, this seems to be populated from a json file.

Is there any reason to break consistency here?

While I don't mind reading from a config file, I think we should probably consider "one file to rule them all" sort of a thing, or drop this file (for the current release)

arvindsv: I guess this is a little more complicated than a property. Unless you're thinking the value of that property should be a JSON value. For 15.2, I don't mind going with defaults chosen by us, but I think @jyotisingh was mentioning someone who wanted to disable some ciphers, etc.

ketan: I was not suggesting a single property containing JSON data. I was suggesting multiple properties. I'm fine with this going in as long as we get rid of the file as the first thing in 15.3.

jyotisingh:
>>Is there any reason that this is read from a config file instead of system properties?

System properties are usually simple string based properties, one being fairly independent of the other. SSL config is more of a combination of a few, so it would feel convenient to edit them all at one place. Also, they are somewhat sacred in nature. Being a file placed under Go's config folder, it can be backed-up, synced etc. `/etc/default/go-server` and `wrapper-properties.conf` files are not backed up.

However, if you guys feel using a system property is the right thing to do, I can change the implementation. We should not wait until 15.3 as it will only complicate things such as having to support migration of user's overridden ssl config to system properties etc.

jyotisingh:
>> should we consider upgrading the JRE that comes pre-packaged with the windows installer?

Once #1086 is complete, we could do this as well. Doubt that will happen in 15.2 though.

ketan:
>> should we consider upgrading the JRE that comes pre-packaged with the windows installer?

> Once #1086 is complete, we could do this as well. Doubt that will happen in 15.2 though.

Have you tested your changes with the windows installer? Our JRE is very old. I was merely suggesting bumping it to the latest 1.7 update.

jetty9/src/com/thoughtworks/go/server/config/ConfigurableSSLSettings.java
zabil: I prefer a flat style of options over the verbose way of config to make things simpler.
```
ssl.config.ciphers.include = STRONG_CIPHER
ssl.config.ciphers.exclude = WEAK_CIPHER
ssl.config.protocols.include = STRONG_PROTOCOL
ssl.config.protocols.excludes = WEAK_PROTOCOL
```
Not very sure on system properties or file however.
Also aren't these value set on startup. So even if the files synced don't we have to restart for the server to pick up these properties?

jyotisingh:
>> Have you tested your changes with the windows installer? 

Tested on linux, windows, and across linux server and windows agent (not the other way around). Did not test on browsers on windows (IE). Tested on latest versions of chrome and firefox. 

>> Our JRE is very old. I was merely suggesting bumping it to the latest 1.7 update.

Created and issue for it, #1306

jetty9/src/com/thoughtworks/go/server/config/ConfigurableSSLSettings.java
jyotisingh: >>Is there any reason that this is read from a config file instead of system properties?

System properties are usually simple string based properties, one being fairly independent of the other. SSL config is more of a combination of a few, so it would feel convenient to edit them all at one place. Also, they are somewhat sacred in nature. Being a file placed under Go's config folder, it can be backed-up, synced etc. `/etc/default/go-server` and `wrapper-properties.conf` files are not backed up.

However, if you guys feel using a system property is the right thing to do, I can change the implementation. We should not wait until 15.3 as it will only complicate things such as having to support migration of user's overridden ssl config to system properties etc.

jyotisingh: >> Also aren't these values set on startup. So even if the files synced don't we have to restart for the server to pick up these properties?

Yes to both. But then, as of now you cannot switch to the new primary without restarting it. Even the config xml is not recognised by new primary until a restart.

jyotisingh:
Also created #1313(Move Go's system properties to a file-system based config for user's to override)

PR #1304, opened 2015-07-07
Dashboard_periodical_executer is modified as follow :-

chopraapooja:
- Replaced old prototype style call (Class.create) for creating executer
    - Replaced Ajax call for JSON to jQuery.ajax call
    - Added pause condition for executor


server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
```diff
@@ -43,7 +43,6 @@ BuildOutputObserver.prototype = {
 
         if (this.is_completed && this.is_output_empty) {
             this.update_page(json);
-            dashboard_periodical_executer.pause();
```
jyotisingh: Why did you get rid of this call? Won't it start making calls to update job status of a completed build?

chopraapooja: Basically here, build_output_observer is pausing the executer, it is not the responsibility of observer to pause it's executor. Ideally executor should decide itself to pause on certain condition, I have moved this call [here] (https://github.com/chopraapooja/gocd/blob/job-status-json-polling-fix/server/webapp/WEB-INF/rails.new/app/assets/javascripts/dashboard_periodical_executer.js#L70-L72)

jyotisingh:
Could you sync the the PR

chopraapooja:
#1387 :+1:

PR #1305, opened 2015-07-07
Tweak URL to the API docs.

ketan:


ketan:
Merged it myself since this is a cosmetic change.

PR #1310, opened 2015-07-08
Couple of fixes with console log page.

ketan:
* When switching between tabs, the event handler that switches tabs is
  before the one that hides the back to top button. Since we're using
  the end result of the first callback to determine the result of the
  second, this does not work as expected.
* Also scrolling up from the very bottom of the log when live tailing is
  turned on does not disable "scroll to bottom" feature because the
  scroll event handlers were only installed when the autoscroll button
  was clicked.


arvindsv:
Two thoughts:

1. During a long-running job (in progress), the "Back to top" link does not stick to the bottom of the screen. It is at the bottom of the page, where it is not as useful.

2. The little job history arrow does not show up when there is only one previous job instance (Label = 1 or 2).

Issue #1314, opened 2015-07-08
Config repository modification diff api returns bad content type


juhijariwala:
``` shell
curl http://localhost:8153/go/api/config/diff/d2f9091145e2e4b3b22d938c4819610467ae20c1/f27b48594645c9f15f55bb56d0100dd60e4cbacc \
    -uadmin:badger \
    -is | grep 'Content-Type:'
Content-Type: text/html; charset=UTF-8
```

Expected : `Content-Type: text/plain; charset=UTF-8`

zabil:
Verified and working on 15.3.0-2361 (15f38854eea8741f5eaa9d262cdae4fcb9a29c45)

PR #1315, opened 2015-07-08
fix for #1314. Return text/plain content-type.

juhijariwala:


ketan:
@juhijariwala - can you add a test to assert on the content type so it does not regress in the future.

Issue #1319, opened 2015-07-09
Port over backup API


jyotisingh:
#1292

zabil:
on experimental release15.2.0-2237 (2fd095a561193826f445f9627cf6861c8983f0dc)

The new back up api reports 500 for a successful backup

To replicate try the old api
```
/T/go-server-15.2.0 curl -X POST http://localhost:8153/go/api/admin/start_backup -v
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8153 (#0)
> POST /go/api/admin/start_backup HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 09 Jul 2015 11:46:36 GMT
< Set-Cookie: JSESSIONID=2ncubs4ownqn1gaargg4mj5s4;Path=/go;Expires=Thu, 23-Jul-2015 11:46:36 GMT
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-UA-Compatible: chrome=1
< Content-Type: text/html; charset=UTF-8
< ETag: "6bafc8b96cf6034afea1b77a7a41570f"
< Cache-Control: max-age=0, private, must-revalidate
< X-Request-Id: a258f8be-f46d-4012-8c2c-78bd4fb4bf36
< X-Runtime: 0.079000
< Transfer-Encoding: chunked
* Server Jetty(9.2.z-SNAPSHOT) is not blacklisted
< Server: Jetty(9.2.z-SNAPSHOT)
< 
Backup completed successfully.
* Connection #0 to host localhost left intact
~/T/go-server-15.2.0 
```

and then the new API

```
curl -X POST -H 'Accept: application/vnd.go.cd.v1+json' http://localhost:8153/go/api/backups -v
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8153 (#0)
> POST /go/api/backups HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8153
> Accept: application/vnd.go.cd.v1+json
> 
< HTTP/1.1 500 Server Error
< Date: Thu, 09 Jul 2015 11:47:40 GMT
[...]

PR #1326, opened 2015-07-10
#1319 Summarizing user for the back API response

zabil:
This also fixes problem with serialization when go server is used without authentication.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/backup_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/backup_representer.rb
@@ -16,14 +16,7 @@
 
 module ApiV1
   class BackupRepresenter < ApiV1::BaseRepresenter
-    attr_reader :user
     alias_method :backup, :represented
-
-    def initialize(backup_and_user)
-      backup, user = *backup_and_user
-      super(backup)
-      @user = user
-    end
 
     link :doc do
       'http://api.go.cd/#backups'
@@ -31,6 +24,11 @@ module ApiV1
 
     property :getTime, as: :time
     property :getPath, as: :path
-    property :user, exec_context: :decorator, decorator: UserRepresenter
+    property :user, exec_context: :decorator, decorator: UserSummaryRepresenter
+
+    def user
+      backup.getUsername
+    end
+
   end
 end
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/user_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/user_representer.rb
@@ -15,23 +15,14 @@
 ##########################################################################
 
 module ApiV1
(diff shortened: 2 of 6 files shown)


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/backup_representer.rb
```diff
@@ -16,13 +16,11 @@
...
 module ApiV1
   class BackupRepresenter < ApiV1::BaseRepresenter
-    attr_reader :user
     alias_method :backup, :represented
 
-    def initialize(backup_and_user)
-      backup, user = *backup_and_user
+    def initialize(backup)
+      backup
       super(backup)
```
ketan: @zabil - Don't need the constructor, in this case

ketan:
In the world of hypermedia - I'm curious why this was removed? If you're rendering information about a different type of resource, why not render the resource (or a url to that resource).

An API without hyperlinks is not where we want to go again. Ever.

zabil:
Not sure if I understand.

Why should the backup API return the entire user representation (with email, aliases etc) a hyperlink to the user resource makes sense but when go is used without auth there's no user resource how do we handle that?

ketan:
> Why should the backup API return the entire user representation (with email, aliases etc) a hyperlink to the user resource makes sense but when go is used without auth there's no user resource how do we handle that?

That is a certainly a problem to fix - but I don't think removing the user resource for everyone is a solution for that problem.

zabil:
>  but I don't think removing the user resource for everyone is a solution for that problem.
OK

Here's an attempt to fix that

Is changing  
```
{
  "_links": {
    "doc": {
      "href": "http://api.go.cd/#backups"
    }
  },
  "time": "2015-07-01T14:21:46.593+05:30",
  "path": "/var/lib/go-server/serverBackups/backup_20150701-142146",
  "user": {
    "_links": {
      "doc": {
        "href": "http://api.go.cd/#users"
      },
      "self": {
        "href": "https://ci.example.com/go/api/users/admin"
      },
      "find": {
        "href": "https://ci.example.com/go/api/users/:login_name"
      }
    },
    "login_name": "admin",
    "display_name": "admin",
    "enabled": true,
    "email": null,
    "email_me": false,
    "checkin_aliases": [

    ]
  }
```

to 

```
{
  "_links": {
    "doc": {
      "href": "http://api.go.cd/#backups"
    }
  },
  "time": "2015-07-01T14:21:46.593+05:30",
  "path": "/var/lib/go-server/serverBackups/backup_20150701-142146",
  "user": {
    "_links": {
      "doc": {
        "href": "http://api.go.cd/#users"
      },
      "self": {
        "href": "https://ci.example.com/go/api/users/admin"
      },
      "find": {
        "href": "https://ci.example.com/go/api/users/:login_name"
      }
    }
  }
```

right?

And what to do with user when unauthenticated?

ketan:
That sounds reasonable. Rendering a link to the user object, without rendering the user attributes sounds good.

PR #1331, opened 2015-07-10
1133 low services

tomzo:
This is third part of #1133

The goal of work here is to make Go services accept more than one source of configuration.

### Changes

Best reference to what I did is [here](https://github.com/gocd/gocd/issues/1133#issuecomment-116810715)

#### Effects

This does not make #1133 operational to any extent. Adding a configuration repository to xml would have no effect because there is no embedded plugin or extension point (they are in next PRs).

But it does make the most dramatic changes in all PRs so it is essential to prove that nothing is broken when this is merged.

I have well covered test cases for backwards compatibility - meaning all test cases that ran on old services now pass on new services. But these cases have no remote configuration repositories. I have added (too) few cases where configuration repository is defined.

### Review

Commits have very long, explanatory messages. Soon after I was done with these service I have written a summary which might be helpful (look below 'Merged config for edits')  - https://github.com/gocd/gocd/issues/1133#issuecomment-113937503

Best comparison is at https://github.com/tomzo/gocd/compare/1133-schema76-pr...tomzo:1133-services

#### Ignore

Services which I added but removed in the end (please ignore them):
 * `ScmCheckoutService`
 * `PartialConfigStatusService`

Other misleading code:
 * `GoConfigPluginService` is not implemented and just returns null. This is taken care of in next PRs.
 * `PartialConfigLoadContext` is empty. Will be done in next PRs

**Many lines** - there are some renames and refactoring that caused it:
 * `GoConfigFileDao` to `GoConfigDao` - so to not mislead anyone that we are accessing just a file. Dao is talking to `MergedGoConfig` now.
 * `GoConfigDataSource` to `GoFileConfigDataSource`
 * many tests called `cachedGoConfig.onTimer()` to reload configuration file. Now they call `cachedGoConfig.forceReload()`. Effect is the same, but they talk to `MergedGoConfig`


arvindsv:
This might not take me as long as I had thought it would. I spent a day on this, and here are some thoughts. I still have a few parts left to review (I'll mention which, in the next comment):

- The GoRepoConfigDataSource does not seem to be used, [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/MaterialUpdateListener.java#L35).

- What are [these three lines](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java#L85-L87) added for? I don't see a test for that, so I don't know what it is for.

- If you want, [this](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java#L158-L170) can be simplified to:
```
public boolean hasSameMaterial(MaterialConfig config) {
    MaterialConfig materialConfig = this.getMaterialConfig();
    return materialConfig != null && materialConfig.equals(config);
}

public boolean hasMaterialWithFingerprint(String fingerprint) {
    MaterialConfig materialConfig = this.getMaterialConfig();
    return materialConfig != null && materialConfig.getFingerprint().equals(fingerprint);
}
```
- Also, there's no test for those two methods above.
- I'm curious about the null check case in the methods above. It looks to me like `repo` can (should) never be null.
- In `hasSameMaterial()`, if `this.getMaterialConfig()` is null, then the return value is always `false`. What if the one passed in is null too? Should we even expect and have code for null config?


- In [these tests](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/test/unit/com/thoughtworks/go/config/MergedGoConfigTest.java#L107-L115), it's not entirely clear where the listeners are setup (and why). I'd change the name to something like `shouldListenForFileChangesUponCreation` to give a clue that it is in the constructor. Do you think it makes sense to check the effect of that, rather than the fact that the listener exists. Maybe a test whose name is `shouldTryToAssembleMergedConfigWhenFileConfigChanges`, to show that the listener exists, but without really calling `hasListeners`.

- [This](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L79) is quite a complex method, and it seems to have no tests. Am I missing something? Is it tested through some other class?

- [These two methods](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/config/config-api/src/com/thoughtworks/go/config/remote/ConfigReposConfig.java#L77-L92) don't have tests too, but they're quite simple. So, I'm reasonably ok with them not being tested. Just putting this out there, in case you meant to write them and missed it.

- The change to [this method](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L57) is not tested. I think [this line](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L78) can be removed, and [this line](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L67) changed to have that. Something like: ```revision = newChanges.isEmpty() ? lastRevision.getRevision() : newChanges.get(newChanges.size() - 1).getRevision();```, since [this line](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L62) already has the ```newChanges.size - 1``` bit and it looks like [this block](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L69-L79) was meant to be just a logging block. Also, the fact that `scmMaterialCheckoutService` is called, with the right params isn't checked. What do you think? It seems like it didn't have a test class before, as well. Now might be a good time to start, I feel. Even if it is a test only for this change.

- I think [this](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/PartialConfigChangedListener.java#L23) can be removed. git history will show it is you. :) I'm just making fun of you a bit. I know that the IDE probably added it by itself.


- [ScmMaterialCheckoutService](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/ScmMaterialCheckoutService.java#L34) doesn't have a test. Maybe it is a work in progress? For instance, I see that it is not passing along the modifications (they're not being used at all).

- [This variable](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoConfigWatchList.java#L75) can be inlined.

arvindsv:
I'm yet to see the changes made to this set of files:

```
config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
server/src/com/thoughtworks/go/config/GoPartialConfig.java
server/src/com/thoughtworks/go/server/service/PartialConfigStatusService.java
server/src/com/thoughtworks/go/config/MergedGoConfig.java
server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java

server/test/unit/com/thoughtworks/go/config/CachedFileGoConfigTest.java
common/test/unit/com/thoughtworks/go/domain/CruiseConfigTest.java
server/test/integration/com/thoughtworks/go/util/GoConfigFileHelper.java
config/config-api/test/com/thoughtworks/go/helper/ConfigFileFixture.java
server/test/unit/com/thoughtworks/go/config/MergedGoConfigTest.java
server/test/unit/com/thoughtworks/go/config/GoPartialConfigTest.java
server/test/unit/com/thoughtworks/go/config/GoRepoConfigDataSourceTest.java
```

I don't expect that to take more than a day or two.

tomzo:
Hi @arvindsv 

Please notice that `ScmCheckoutService` is added temporarily (as I mentioned above in PR description). I remove it in next PRs. Same case for `PartialConfigStatusService` - don't even review it, it is not used and not tested. This is not the final layout of services, because there is more work in #1333

In general, I'll write some of missing tests that you pointed. But here are some immediate issues

This is quite important:

> What are [these three lines](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java#L85-L87) added for? I don't see a test for that, so I don't know what it is for.

It is because merged config can be passed to magical writer. But it should not save remotely defined elements of configuration to xml file. To deal with that - magical writer is aware of possibility that merged config might be passed to be serialized. If so then it takes out only locally defined configuration elements in the 3 lines mentioned. Of course I'll add the tests here as well.

> Changes in ScmMaterialUpdater

I have reverted all changes in `ScmMaterialUpdater` (in `insertLatestOrNewModifications`) after removing `ScmCheckoutService`. [Here](https://github.com/tomzo/gocd/blame/7529619d98eef3bbde88a31bd968844bea818436/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L54) you can see it is back to previous state.

> [ConfigRepoConfig] In hasSameMaterial(), if this.getMaterialConfig() is null, then the return value is always false. What if the one passed in is null too? Should we even expect and have code for null config?

Null config may happen only in tests. But perhaps we should throw when it is null and just fix all tests.

> The GoRepoConfigDataSource does not seem to be used, [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/MaterialUpdateListener.java#L35)

It is removed in next branch

I'd like to wait for the rest of review before I fix all that.

arvindsv:
> Please notice that `ScmCheckoutService` is added temporarily (as I mentioned above in PR description). I removed it in next PRs. Same case for `PartialConfigStatusService` - don't even review it, it is not used and not tested. This is not the final layout of services, because there is more work in #1333

Ok. I didn't realize you meant `ScmMaterialCheckoutService`. I didn't find anything called `ScmCheckoutService`. No problem. I won't review `PartialConfigStatusService`.

> It is because merged config can be passed to magical writer. But it should not save remotely defined elements of configuration to xml file. To deal with that - magical writer is aware of possibility that merged config might be passed to be serialized. If so then it takes out only locally defined configuration elements in the 3 lines mentioned. Of course I'll add the tests here as well.

Ok, I understand. Isn't it safe to always do a `.getLocal` then? Maybe that check is unnecessary?

> I have reverted all changes in `ScmMaterialUpdater` (in `insertLatestOrNewModifications`) after removing `ScmCheckoutService`. [Here](https://github.com/tomzo/gocd/blame/7529619d98eef3bbde88a31bd968844bea818436/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L54) you can see it is back to previous state.

I see. But, if I merge this PR in and take some time to merge the other one, then this code will be in. Is it going to be too painful (too many conflicts) to revert the changes to this file, in this PR? If so, let me know.

> Null config may happen only in tests. But perhaps we should throw when it is null and just fix all tests.

Yes, please. If possible, I think that'll be better.

>> The GoRepoConfigDataSource does not seem to be used, [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/server/materials/MaterialUpdateListener.java#L35)

> It is removed in next branch

Ok.

> I'd like to wait for the rest of review before I fix all that.

Sure.

arvindsv:
Ok! I'm done. Here's what I have. None of it changes much of your code, I think. I couldn't find much to say about the code changes. :) So, most of it is related to tests. There are a lot of TODOs in the code.

- About changes to BasicCruiseConfig: Tests like `CruiseConfigTest#shouldAddPipelineToMain`, `CruiseConfigTest#shouldAddPipelineToNewGroup_InMergeAndLocalScope` and `CruiseConfigTest#shouldAddPipelineToExistingGroup_InMergeAndLocalScope` might need to move to the corresponding strategy tests, since the code moved there. Even if they're not moved, the fact that `MergeStrategy#addPipelineWithoutValidation` and `MergeStrategy#addPipeline` do more now (like validation) needs to have a test.

- GoPartialConfigTest and GoRepoConfigDataSourceTest have some methods like `SetUp` and `SetOneConfigRepo`, which start with an uppercase letter. I know it's a bit of a nit pick.

- Why does [`GoPartialConfig#lastPartials()`](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoPartialConfig.java#L74) return an array? Isn't it easier to deal with just a list? I realize it might need to be converted into an array later, to pass into the constructor of BasicCruiseConfig. But, from an interface point of view, I think list makes more sense.

- I think there's a test or two missing in `GoPartialConfigTest`. Something like, `shouldNotifyListenersAfterUpdatingMapOfLatestValidConfig` and `shouldNotifyListenersAfterRemovingPartial`.

- In [`GoPartialConfig#onSuccessPartialConfig`](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoPartialConfig.java#L86): Why is there a check against whether the `configWatchList` has the material? Right now, nothing will happen if it is not in the list. Is that even a valid scenario?

- [This](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L94) is unnecessary. Also, there can be a test written, with a name like this: `shouldNotNotifyListenersWhenMergeFails`.

- If there are no partials, [MergedGoConfig#tryAssembleMergedConfig](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L86) seems to ignore the config passed in, and just uses the one in the CachedGoConfig. Is this intentional? I guess, in most cases they will be in sync, but it feels a little weird to ignore the passed-in argument.

- Should we use a different status [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L113)? If the main config is invalid, it would probably not reach this point. We have more information here (that the merge failed, or a partial is invalid).

- Shouldn't [this](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L136) set currentConfig to `fileService.load()` and then return that instead?

- The name of [this](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java#L53) shouldn't be `fingerprintLatestConfigMap` right? It should be something like `fingerprintOfPartialToItsLatestParseResultMap`, I think.

- It looks like [GoRepoConfigDataSource#latestPartialConfigForMaterial](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java#L82) is used only in tests. It should be possible to remove this, and use `notifyFailureListeners` and `notifySuccessListeners` instead, right? Trying to see if any test-only code can be removed. I can make this change and send it to you, if you like. Unless it is planned to be used somewhere else.

- If this [TODO](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java#L107) is not going to happen, then maybe `removedAny` variable can be removed. If it is going to happen, then keeping it makes sense. Sorry if this has been taken care of in another PR.

- Should there be an else block [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java#L125), which throws an exception or logs an unexpected material? Or, is this called back for every checkout of a material, and this is the way of filtering out non-config-repo materials?

- It looks like [this test](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/test/unit/com/thoughtworks/go/config/GoRepoConfigDataSourceTest.java#L97-L118) is testing [`PartialConfig#setOrigins`](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/config/config-api/src/com/thoughtworks/go/config/remote/PartialConfig.java#L67) rather than GoRepoConfigDataSource. It doesn't hurt, but maybe it can be combined with the previous test. It's your choice.

- The method `Load` in `PartialConfigProvider` should be renamed to `load`.

- [This test](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/test/unit/com/thoughtworks/go/config/GoRepoConfigDataSourceTest.java#L178) can be written like this as well:

```
@Test(expected = BrokenConfigPluginException.class)
public void shouldThrowWhenGettingLatestPartialConfig_WhenPluginHasFailed() throws  Exception
{
    // use broken plugin now
    when(configPluginService.partialConfigProviderFor(any(ConfigRepoConfig.class)))
        .thenReturn(new BrokenConfigPlugin());

    ScmMaterialConfig material = new GitMaterialConfig("http://my.git");
    cruiseConfig.setConfigRepos(new ConfigReposConfig(new ConfigRepoConfig(material, "myplugin")));
    configWatchList.onConfigChange(cruiseConfig);

    repoConfigDataSource.onCheckoutComplete(material, folder, "7a8f");

    assertTrue(repoConfigDataSource.latestParseHasFailedForMaterial(material));

    repoConfigDataSource.latestPartialConfigForMaterial(material);
    fail("should have thrown BrokenConfigPluginException");
}
```

- Are all the `load()` calls in `GoConfigFileHelper` changed to `loadForEdit()` because it is the right thing to do, since they're being changed? Just wondering.

tomzo:
Today I'll be fixing all things you mentioned, probably I will finish tomorrow. I agree with most of it. Here are some immediate answers and issues from my point of view:

> In GoPartialConfig#onSuccessPartialConfig: Why is there a check against whether the configWatchList has the material? Right now, nothing will happen if it is not in the list. Is that even a valid scenario?

In short - It prevents a situation when for *some reason* parsing repository contents has finished but the repository is no longer in `<config-repos>` section. E.g. parsing was very slow, user removed config-repo, parsing has finished and got merged to current config.
I do not think this situation is possible in this PR because all parsing, merging, file edits are wired on the same events. But in next PRs, after implementing polling, second queue, threading etc. properly, like we discussed, it could happen IMO. Therefore I added a guard against it.

> Should we use a different status [here](https://github.com/tomzo/gocd/blob/a0ad09841df7e4497f3e96b8c427e0490d71c6da/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L113)? If the main config is invalid, it would probably not reach this point. We have more information here (that the merge failed, or a partial is invalid).

I wanted to ask about that. It seems to me as well that we should have something like `INVALID_CRUISE_CONFIG_XML` and `INVALID_CRUISE_CONFIG_MERGE`. But I am very much unaware **what components actually rely on this status?**. Is it used only in UI to display error and warning messages? Or are there other services which for example - will **ask health service** if config is valid before operating? Can I just add a new status and hope that it will work?

> Should there be an else block here, which throws an exception or logs an unexpected material? Or, is this called back for every checkout of a material, and this is the way of filtering out non-config-repo materials?

Yes, this is actually filtering out config repos from all materials. It is done differently though, in next PR when there are 2 polling queues.

> Are all the load() calls in GoConfigFileHelper changed to loadForEdit() because it is the right thing to do, since they're being changed? Just wondering.

Yes. I renamed them because `load` actually implemented behavior of `loadForEdit`. Just to avoid confusion.

> [Upper Cases everywhere]

I am from C# world. It's an old habit:)

arvindsv:
> I wanted to ask about that. It seems to me as well that we should have something like INVALID_CRUISE_CONFIG_XML and INVALID_CRUISE_CONFIG_MERGE. But I am very much unaware what components actually rely on this status?. Is it used only in UI to display error and warning messages? Or are there other services which for example - will ask health service if config is valid before operating? Can I just add a new status and hope that it will work?

Yes, I think so. As far as I know, that status is not used for some logic.

> I am from C# world. It's an old habit:)

Thought so. :)

tomzo:
Hi @arvindsv 
I have gone through everything you said and fixed that. Here is what did **not** do:

> If there are no partials, MergedGoConfig#tryAssembleMergedConfig seems to ignore the config passed in, and just uses the one in the CachedGoConfig. Is this intentional? I guess, in most cases they will be in sync, but it feels a little weird to ignore the passed-in argument.

This is the best way to ensure that MergedGoConfig behaves exactly like the CachedFileGoConfig below it. Which is highly desirable when there are no partials. We actually do not care about the argument, rather we care about the fact that there is only one config, and above all we want to return exactly that. 
I have changed [method signature](https://github.com/tomzo/gocd/blob/f734564ab5dadcf682832f1694104214b20f2f4b/server/src/com/thoughtworks/go/config/MergedGoConfig.java#L79) a little bit and moved ignoring the argument higher.
BTW I believe that `tryAssembleMergedConfig` is the right place to ask feature toggle service if #1133 should be enabled.

> Changes in ScmMaterialUpdater

Cannot remove all changes from `insertLatestOrNewModifications` without making #1133 inoperable. It is was meant to be a quick and dirty way to hook into MDU thread. In #1332 we can have minimalistic working config repos feature. If I revert it then we won't. 
**But** maybe it doesn't matter at this point? Assuming that #1333 will be merged eventually, do we care that we cannot use config-repos in master until then?
Anyway I applied the suggestions you mentioned. But did not write extra tests. Let me know what do you think. 

> It looks like GoRepoConfigDataSource#latestPartialConfigForMaterial is used only in tests. It should be possible to remove this, and use notifyFailureListeners and notifySuccessListeners instead, right? Trying to see if any test-only code can be removed. I can make this change and send it to you, if you like. Unless it is planned to be used somewhere else.

I really do not want to remove `latestPartialConfigForMaterial`. It is probably true that it is used only in tests. I have also used it in new integration tests in #1333 where it was really useful. I can remove it if you insist. Although I'd prefer to do it in #1333 then.

### Fixes

I have found a few bugs when running server internally in last weeks. I have "backported" them here already.

I have also found and fixed some bugs when writing the missing tests you mentioned.

### `BasicCruiseConfig` strategy tests

I have split tests of cruise config the way I did for services - there is an abstract class with 2 implementations.
 * common behavior is declared in `CruiseConfigTest`.
 * single (xml-only) tested in `BasicCruiseConfigTest` - tests Basic strategy
 * merge (xml + partials) tested in `MergeCruiseConfigTest` - tests Merge strategy

So strategies are tested indirectly by testing behavior of `BasicCruiseConfig`. I hope that is what you meant. I do not think it would be practical to test strategy against a mock of `BasicCruiseConfig`.

----

I'll fix the conflicts from this branch to my others. What do we do with conflicts to master? Will you solve them when merging this PR?

Let me know if there is stll something unacceptable here.

arvindsv:
Hi @tomzo,

> > About MergedGoConfig#tryAssembleMergedConfig

> This is the best way to ensure that MergedGoConfig behaves exactly like the CachedFileGoConfig below it ... I have changed method signature a little bit and moved ignoring the argument higher.

Ok, that's fine.

> > Changes in ScmMaterialUpdater

> **But** maybe it doesn't matter at this point? Assuming that #1333 will be merged eventually, do we care that we cannot use config-repos in master until then?
Anyway I applied the suggestions you mentioned. But did not write extra tests. Let me know what do you think. 

Ok. I'll take a look at the changes sometime early tomorrow. What I care about, at this point, in making sure that if this PR goes in, and a release has to be made before #1333 goes in, that nothing untoward happens in the merged code. Since this change is in the MDU, I'm just being a little extra cautious.

> I really do not want to remove `latestPartialConfigForMaterial`. It is probably true that it is used only in tests. I have also used it in new integration tests in #1333 where it was really useful. I can remove it if you insist. Although I'd prefer to do it in #1333 then.

I won't insist. There are other places where this is done. So, it feels wrong to insist. :) But, I usually try not to have them, if possible. In this case, even though it is not used outside of tests, it's not behavior that feels out of place. Because most code usages are through the listeners, it has not yet been needed in non-test code. So, it's ok.

> `BasicCruiseConfig` strategy tests
> So strategies are tested indirectly by testing behavior of `BasicCruiseConfig`. I hope that is what you meant.

I *think* that's what I meant. I'll take a look.

> I'll fix the conflicts from this branch to my others. What do we do with conflicts to master? Will you solve them when merging this PR?

Yes, I'll take care of them.

arvindsv:
Hi @tomzo,

Nearly there. I have much fewer things this time. I'm trying to merge this PR in before @jyotisingh gets in some changes in this area. :) Hoping to get her to merge her conflicts herself. Also, I'm away most of next week.

So, here are my "final" thoughts. Let me know what you think:

- Do you really need [this change](https://github.com/tomzo/gocd/commit/ee268d9f3c0a5c5a853dbb3a9a9b934d22800202)? It can be extremely expensive. That method does a costly clone and then validates. Then, after configForEdit is replaced, it does another preprocess and
  validate. [This line](https://github.com/tomzo/gocd/commit/ee268d9f3c0a5c5a853dbb3a9a9b934d22800202#diff-56efae43574f3988ca374a9bbe6237cfR91) also potentially throws away a part of the config, right? All the validation errors in that part go away with that, since validation errors are stored on the object.

- About [this commit](https://github.com/tomzo/gocd/commit/4be5d63588d1019bccd08a791d0ee5d0cbb3a176): I think the signature of validateGroupName is a little weird, and makes it really hard to see what is going on. I think the signature should be `public void validateGroupNameAndAddErrorsTo(ConfigErrors errors)`. Then, from the `validate` method in `BasicPipelineConfigs`, it can be called as `this.validateGroupNameAndAddErrorsTo(this.configErrors)` and from `MergePipelineConfigs`, it can be called as `this.parts.get(0).validateGroupNameAndAddErrorsTo(this.configErrors)`. No need to get the message text and then check if it is null and then add it, etc.

  If you agree, I can do it just before checking in. Should be a small change.

- [This](https://github.com/tomzo/gocd/commit/ca23fb442a89b88b4ed2af021a863e6808353fdb#diff-b82dab9582a19e2a031ddd803b2ebefcR126) seems quite dangerous? Are you sure this should be done this way? Shouldn't we change how pipeline edit screen loads or something? This seems like a violation of the contract. It's used in many places. I see you've mentioned that [here](https://github.com/tomzo/gocd/commit/ca23fb442a89b88b4ed2af021a863e6808353fdb#diff-1d0cb58daa749b03d9e36b15e0cf30bbR245) as well, but still ... it makes me feel a little queasy.

- I'm a little apprehensive of this [one line](https://github.com/tomzo/gocd/blob/ead11ff9a93fbc7b843b5a4890cc0bb005d9547c/server/src/com/thoughtworks/go/server/materials/ScmMaterialUpdater.java#L86) because, it is in a critical flow. I know it gets short-circuited [here](https://github.com/tomzo/gocd/blob/ead11ff9a93fbc7b843b5a4890cc0bb005d9547c/server/src/com/thoughtworks/go/config/GoRepoConfigDataSource.java#L125), when there are no remote config repos. I considered putting it behind a toggle, but I will leave it in for now.

If you can take a look at this on or before Saturday, and let me know what you think, I'll try and run the tests, fix conflicts and merge it in.

Thank you!

tomzo:
Hi @arvindsv 

> [About merged config validation in magical writer]

> Do you really need this change? It can be extremely expensive. That method does a costly clone and then validates.

The rationale for validating merged config before saving right here is the same as for basic config that was validated there. I am sure that it is need. I can give you an example use case to clear things up: User has valid merged configuration with `localPipe1` and `remotePipe1`. Then via UI (or worse yet via xml file) user adds a new locally defined pipeline and names it `remotePipe1`. This is obviously a configuration error because pipeline `remotePipe1` will be defined remotely and locally.
If we let user save the main, xml configuration with `remotePipe1` then next thing that will happen is `tryAssembleMergedConfig` will reject current config repository because it has a duplicate pipeline.
This is just one example of possible errors. Recall those new validations I added back in api branch? There were a few error messages that actually pop-up now, thanks to this change. Things like *Do not reference a remote pipeline from local pipeline*. 

In UI branch I did proper changes to make it less likely that user can make such errors. But these are not enough, final validation is mandatory.

> Then, after configForEdit is replaced, it does another preprocess and
validate.

On the one hand if merged config is valid then basic config is valid as well. So we could change:
```diff
                 configForEdit = configForEdit.getLocal();
             }
-            if (!skipPreprocessingAndValidation) {
+            else if (!skipPreprocessingAndValidation) {
                 loader.preprocessAndValidate(configForEdit);
                 LOGGER.debug("[Serializing Config] Done with cruise config validators.");
             }

```

The only potential problem here is that `getLocal` in some cases (in UI branch) actually creates a basic config with only local elements. So the new config object, could, theoretically be invalid because of bad `getLocal` implementation. Having an extra validation of config saved object prevents saving invalid config. It's your decision if you want to change it into `else if`. In this branch it does not matter anyway. IMO this is the cost of having a merged config.

> This line also potentially throws away a part of the config, right? All the validation errors in that part go away with that, since validation errors are stored on the object.

If config is invalid then exception is thrown by `loader.preprocessAndValidate`. Throwing away the remote part was intentional.

> [signature of validateGroupName]

OK. Changed it as you suggested.

> [Merged config for edit]

In UI branch I changed it a little. There is third merge strategy for edits. But I am pretty sure that **config for edit must have all local and remote elements**. I think we should discuss it further in UI branch.

arvindsv:
I decided to merge this, but some tests failed, and I ran out of time. I'm away for the next four days. That means @jyotisingh and @juhijariwala win, and will make me have to fix conflicts. Damn! :/ Fine. I will, as soon as I get back.

@tomzo, here are the details of the failures. Looks like only two small issues causing most failures:

```
Expected: is null
     but: was "defaultGroup"
Expected: is null
     but: was "defaultGroup"
	at com.thoughtworks.go.domain.BasicPipelineConfigsTest.shouldUpdateName(BasicPipelineConfigsTest.java:89)



Tests run: 421 , Failures: 51 , Not run: 0 , Time: 69.920 seconds.

Error	shouldReturnTrueIfThereAreTwoPipelineGroups
Error	shouldAllowCleanupOfNonExistentStages
Error	shouldRemovePackageRepositoryById
Error	shouldReturnConfigRepos
Error	shouldDeleteAdminRoleFromAllApprovalsWhenARoleIsDeleted
Error	shouldErrorOutWhenTwoPipelinesDependsOnEachOther
Error	shouldErrorOutWhenDependsOnItself
Error	shouldRemoveUserFromRoleWhenRoleIsDeleted
Error	shouldNotConsiderEqualObjectsAsSame
Error	shouldReturnFalseWhenDoesNotHaveGroup
Error	shouldOfferNothingToVisitorsIfThereAreNoTasks
Error	getAllErrors_shouldIgnoreErrorsOnElementToBeSkipped
Error	shouldReturnSelectedGroupNameIfGroupNameIsSpecified
Error	shouldRemoveSystemAdminPermissionsForARoleWhenARoleIsDeleted
Error	shouldAddPipelineWithoutValidationInAnExistingGroup
Error	getAllErrors_shouldCollectAllErrorsInTheChildren
Error	shouldReturnTrueWhenHasGroup
Error	shouldBeAbleToExplicitlyLockAPipeline
Error	shouldCollectAllTheErrorsInTheChildren
Error	shouldKnowWhenBuildPlanNotInConfigFile
Error	shouldTellIfSMTPIsEnabled
Error	shouldDecideIfRepoCanBeDeleted_BasedOnPackageRepositoryBeingUsedByPipelines
Error	shouldFindBuildPlanWithStages
Error	shouldFindNextStage
Error	shouldReturnFalseIfNoMailHost
Error	shouldFindAllAgentResources
Error	shouldOfferAllTasksToVisitors
Error	shouldNotLockAPipelineWhenItIsAddedToAnEnvironment
Error	shouldErrorOutWhenThreePipelinesFormACycle
Error	shouldFindDownstreamPipelines
Error	shouldClearPackageRepositoryConfigurationsWhichAreEmptyWithNoErrors
Error	shouldReturnFalseIfThereIsOnePipelineGroup
Error	shouldReturnAMapOfAllTemplateNamesToListOfAssociatedPipelinesIfUserIsSuperAdmin
Error	shouldUpdatePackageRepository
Error	shouldReturnTrueIfMailHostIsConfigured
Error	shouldGetJobConfigByName
Error	shouldDeleteAdminRolesInAllPipelineGroupWhenARoleIsDeleted
Error	shouldNotDuplicateErrorWhenPipelineDoesnotExist
Error	shouldIgnoreConstantFieldsWhileCollectingErrorsToAvoidPotentialCycles
Error	getAllErrors_shouldRetainAllErrorsWhenNoSubjectGiven
Error	shouldReturnFalseForEmptyCruiseConfig
Error	shouldAddPackageRepository
Error	shouldReturnAMapOfTemplateNamesToListOfAssociatedPipelinesBasedOnUserPermissions
Error	shouldDecideIfPluggableSCMMaterialCanBeDeleted_BasedOnPluggableSCMMaterialBeingUsedByPipelines
Error	shouldReturnDefaultGroupNameIfNoGroupNameIsSpecified
Error	shouldAddPackageDefinitionToGivenRepository
Error	shouldThrowExceptionWhenThereIsNoGroup
Error	shouldFindPreviousStage
Error	shouldBuildTheValidationContextForAnOnCancelTask
Error	shouldLoadPasswordForGivenMaterialFingerprint
Error	shouldReturnAMapOfTemplateNamesToListOfAssociatedPipelinesCaseInsensitively
Unit Test Failure and Error Details (51)
Test:	shouldReturnTrueIfThereAreTwoPipelineGroups
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldAllowCleanupOfNonExistentStages
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldRemovePackageRepositoryById
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnConfigRepos
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldDeleteAdminRoleFromAllApprovalsWhenARoleIsDeleted
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldErrorOutWhenTwoPipelinesDependsOnEachOther
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldErrorOutWhenDependsOnItself
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldRemoveUserFromRoleWhenRoleIsDeleted
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldNotConsiderEqualObjectsAsSame
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnFalseWhenDoesNotHaveGroup
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldOfferNothingToVisitorsIfThereAreNoTasks
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	getAllErrors_shouldIgnoreErrorsOnElementToBeSkipped
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnSelectedGroupNameIfGroupNameIsSpecified
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldRemoveSystemAdminPermissionsForARoleWhenARoleIsDeleted
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldAddPipelineWithoutValidationInAnExistingGroup
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	getAllErrors_shouldCollectAllErrorsInTheChildren
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnTrueWhenHasGroup
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldBeAbleToExplicitlyLockAPipeline
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldCollectAllTheErrorsInTheChildren
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldKnowWhenBuildPlanNotInConfigFile
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldTellIfSMTPIsEnabled
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldDecideIfRepoCanBeDeleted_BasedOnPackageRepositoryBeingUsedByPipelines
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldFindBuildPlanWithStages
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldFindNextStage
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnFalseIfNoMailHost
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldFindAllAgentResources
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldOfferAllTasksToVisitors
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldNotLockAPipelineWhenItIsAddedToAnEnvironment
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldErrorOutWhenThreePipelinesFormACycle
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldFindDownstreamPipelines
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldClearPackageRepositoryConfigurationsWhichAreEmptyWithNoErrors
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnFalseIfThereIsOnePipelineGroup
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnAMapOfAllTemplateNamesToListOfAssociatedPipelinesIfUserIsSuperAdmin
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldUpdatePackageRepository
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnTrueIfMailHostIsConfigured
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldGetJobConfigByName
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldDeleteAdminRolesInAllPipelineGroupWhenARoleIsDeleted
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldNotDuplicateErrorWhenPipelineDoesnotExist
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldIgnoreConstantFieldsWhileCollectingErrorsToAvoidPotentialCycles
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	getAllErrors_shouldRetainAllErrorsWhenNoSubjectGiven
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnFalseForEmptyCruiseConfig
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldAddPackageRepository
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnAMapOfTemplateNamesToListOfAssociatedPipelinesBasedOnUserPermissions
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldDecideIfPluggableSCMMaterialCanBeDeleted_BasedOnPluggableSCMMaterialBeingUsedByPipelines
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnDefaultGroupNameIfNoGroupNameIsSpecified
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldAddPackageDefinitionToGivenRepository
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldThrowExceptionWhenThereIsNoGroup
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldFindPreviousStage
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldBuildTheValidationContextForAnOnCancelTask
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldLoadPasswordForGivenMaterialFingerprint
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
Test:	shouldReturnAMapOfTemplateNamesToListOfAssociatedPipelinesCaseInsensitively
Type:	Error
Message:	
java.lang.InstantiationException
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
```

tomzo:
Ah, I might have changed how null group name was handled. It should take 20 minutes to repair, however...

We've had a little disaster over the weekend and I will not be able to take care of this, because among many things my workstation and dev. environment for gocd is gone. I will need a few days before I can start programming. Sorry, but I cannot do much more to speed this up. I hope it's fine, especially that you are not available anyway. I'll push a fix as soon as I can.

tomzo:
Hi @arvindsv 

Sorry that this took so long. I hope you are not angry.

I fixed this error:
```
Expected: is null
     but: was "defaultGroup"
Expected: is null
     but: was "defaultGroup"
    at com.thoughtworks.go.domain.BasicPipelineConfigsTest.shouldUpdateName(BasicPipelineConfigsTest.java:89)
```

I cannot reproduce the other ones. 
But after looking a while at those, they all seem to be **test methods** from `CruiseConfigTest`. Since last review I changed `CruiseConfigTest` to an abstract class, hence the instantiation exception.
Now they are split into `BasicCruiseConfigTest` and `MergeCruiseConfigTest`.
Perhaps you have some unclean environment where test runner is trying to run tests which no longer exist? 

Just to be clear - I ran tests like this on last commit:

Clean build:
```
./bn clean cruise:prepare
```
Ran `mvn test` in
 * config/config-api
 * config/config-server
 * server/

All passed in first 2, in server I got some errors, but these seem unrelated (I am missing svn and hg commands).

Let me know how you got the errors you mentioned.

arvindsv:
> Sorry that this took so long. I hope you are not angry.

Oh my GOD! Please don't ever say that. :)

I just got back. I'll take a look in a couple of days. I'm the one slowing this all down, and should apologize. Trying to balance too many things, sorry.

arvindsv:
Worked on this a bit. I found out what the problem is. All files whose name end in Test.java are considered test files by a script which sends the list to TLB (test load balancer). It's what splits our test suite so that it can be run in parallel. Since CachedGoConfigBaseTest.java and CruiseConfigTest.java both end in Test.java, it picked them up and ran them as real test files. I've renamed them to CachedGoConfigTestBase.java and CruiseConfigTestBase.java. Running tests now.

I could have pushed it earlier today, but a Windows build was failing. I don't know if you are (un)fortunate enough to have to debug tests in a small VM window, which RDPs into a Windows VM. It's not fun and slow. So, that took time. Once the main build passes, and assuming the tests I changed pass, I'll merge it in.

tomzo:
>  I found out what the problem is. All files whose name end in Test.java are considered test files by a script which sends the list to TLB (test load balancer). It's what splits our test suite so that it can be run in parallel. Since CachedGoConfigBaseTest.java and CruiseConfigTest.java both end in Test.java, it picked them up and ran them as real test files. I've renamed them to CachedGoConfigTestBase.java and CruiseConfigTestBase.java. Running tests now.

Good catch. I recall making many tests abstract in this way. However they are spread across all my branches. I'll definitely fix those in next PRs. In this branch there is also:
 * GoConfigFileDaoTest.java → GoConfigDaoBaseTest

> I could have pushed it earlier today, but a Windows build was failing. I don't know if you are (un)fortunate enough to have to debug tests in a small VM window, which RDPs into a Windows VM. It's not fun and slow. So, that took time

I know the pain and appreciate your efforts.

tomzo:
Hurray!

I see the tests passed so I'm really happy (and relieved) to see this in master.
@arvindsv thanks for everything, I know it required a lot of work from you as well.

arvindsv:
> I see the tests passed so I'm really happy (and relieved) to see this in master.

Yes, they passed! A few agent VMs slowed down a lot suddenly and failed the build once. But, I could rerun them, and they ran as quick as they usually do and passed! I was worried it was some performance related issue in the code, but it looks like the VM host misbehaved.

> @arvindsv thanks for everything, I know it required a lot of work from you as well.

Not as much as you. :) You're welcome though. We move forward.

PR #1332, opened 2015-07-11
1133 minimal

tomzo:
This is fourth part of #1133 . There will be 4 more PRs after this:
 * extension point (done)
 * minimal UI changes to disable editing remote configuration, but view it properly. (TODO)
 * proper configuration repository polling (done) - #1333 
 * scm-config consistency (done)

---

It contains `gocd-xml` pseudo-plugin that makes it possible to use configuration repositories which are defined in current Go's xml format. Plugin is implemeted on top of magical loader.
Then there are some minor changes in services so that this plugin is actually available.

### Effects

When a section like this is added
```xml
<config-repos>
   <config-repo>
     <git url="https://github.com/tomzo/gocd-indep-config-part.git" />
   </config-repo>
</config-repos>
```
Then Go will poll this repository and create 3 new pipelines.

This is a minimalistic feature, there are limitations:
 * polling of the configuration repository is done by a hack in MDU. This is done properly in next PR.
 * there is no guarantee of scm-config consistency. Also done in next PR
 * no UI adjustments - so you will get weird errors when trying to edit a remote pipeline

### Review

Whole thing is just 11 commits.

Comparison 3rd to 4th part - https://github.com/tomzo/gocd/compare/1133-services-pr...tomzo:1133-minimal-pr


server/src/com/thoughtworks/go/config/GoConfigPluginService.java
```diff
@@ -23,8 +27,18 @@
...
 @Component
 public class GoConfigPluginService {
+
+    private final XmlPartialConfigProvider embeddedXmlPlugin;
+
+    @Autowired public GoConfigPluginService(ConfigCache configCache,ConfigElementImplementationRegistry configElementImplementationRegistry,
+                                 MetricsProbeService metricsProbeService)
+    {
+        MagicalGoConfigXmlLoader loader = new MagicalGoConfigXmlLoader(configCache, configElementImplementationRegistry, metricsProbeService);
+        embeddedXmlPlugin = new XmlPartialConfigProvider(loader);
```
jyotisingh: I see that `XmlPartialConfigProvider` is not really a Go *plugin*. Do you plan to introduce an extension point and move this to a plugin in one of your later PRs?

tomzo:
@jyotisingh I did implement an extension point. But I am postponing a PR with it until I get a chance to use and test it.  I have written a JSON configuration plugin so far.
However I am **not planning on moving xml plugin behind extension point** for the following reason:
The plugin would have to implement loading xml file without using existing Go code (`MagicalGoConfigXmlLoader`) then it would serialize it into JSON to pass through extension point, then it would be (eventually) converted to config objects that `MagicalGoConfigXmlLoader` can load straight from xml. 
This would be a lot of work without any benefit (I think).

IMO we can either keep this xml plugin like it is now or throw it away once extension point is merged.
As you probably realize, until then this is the only way to load configuration from repository.

I'll fix the other comments you mentioned.

jyotisingh:
Fair enough

Issue #1082, opened 2015-04-29
Elastic agents


zabil:
More details here.

https://arvindsv.github.io/elastic_agents_for_gocd.html

---
### Tasks:

* Move to websockets for agent-server communication. 
  Take care of build assignments and rescheduling flows
* Server could invoke the plugin right after scheduling based on the agent requirements(resources). 
  Plugins are tagged with special resources stating the kind of plugin they handle, as well as the resources they are capable of handling. eg. : For a docker plugin, the plugin settings would be something like "linux, mvn" maps to build/dockerfile and "ab, linux" maps to performance/dockerfile. So when the plugin is told about "ab, linux" it know it has to create a performance agent. The plugin should pass along these resources for the agent to auto-register. What about environments?
* plugins brings up the required agents(if it has to)
* agent will now run as regular agent, so work assignment flow would be the same as regular plugins. 
* periodically, the plugin decides to kill the agents it has spun up based on whatever arbitrary rule that it fancies, say run duration, idle-state etc.  The same needs to be tracked by the plugin. We probably need to notify the plugin about the current status of the agent ie. idle or not. No other info would be made available in Go to allow the plugin to make this decision.
* auto-registration of such agents is already taken care of. The pending task here would be for plugins to make sure that autoregister.properties file is available for the agent with the right values. Question: How do we pass the autoregister key to such agents? Same for environments key?

### Open questions so far, with some answers

1. Who provisions the agent, once the agent machine is provisioned, how is the registration handled. Who manages accounts/ssh keys for machines running elastic agents?

  Ans. Agent provisioning managed by plugin implementations. Registration is managed by the existing `autoregister.properties` mechanism.

2. When an agent boots up and (re-)registers, how does the server tell between an elastic agent (and what kind, ec2 vs docker) and "normal agent". How does a specific plugin know about its agent vs an agent belonging to another plugin?

  Ans. 
  * Based on the conversation here. We felt that the right way to address this issue is to have an additional property (agent.auto.register.plugin.id) in the autoregister.properties. This property SHOULD indicate which particular plugin owns an agent. This will be a runtime property that will NOT be persisted in the <agent/> tag in cruise-config.xml

  * ~~We're going with resources (for now). So if a EC2 plugin is tasked with bring up agents with resources "centos65, ec2, firefox, java" any agents that come up with those **exact** resources will be managed by the ec2 plugin. Plugins will be tasked with ensuring that `autoregister.properties` contain the correct resources.~~

[...]

PR #1359, opened 2015-07-16
Scrub out the sensitive information from `autoregister.properties` file after reading it.

ketan:
The current implementation removes the `autoregister.properties` file
after reading it. Since we're going to be piggy-backing on the auto
registration mechanism to identify additional properties to associate
an agent with the specific plugin that owns it, we would like to keep
the file around across agent restarts.


agent/test/unit/com/thoughtworks/go/agent/service/AutoregistrationPropertiesTest.java
```diff
@@ -0,0 +1,101 @@
...
+        AutoregistrationProperties properties = new AutoregistrationProperties(configFile);
+        properties.scrubRegistrationKey();
+
+        String newContents = "" +
+                "#\n" +
+                "# file autogenerated by chef, any changes will be lost\n" +
+                "#\n" +
+                "# the registration key\n" +
+                "# The autoregister key has been intentionally replaced by Go as a security measure.\n" +
+                "agent.auto.register.key = ***** Replaced by Go *****\n" +
```
arvindsv: If the key itself is not removed or commented, won't it use `***** Replaced by Go *****` as the key next time?

ketan: Hmm, doesn't seem like there's a way to comment out a property :(

arvindsv: If an agent is already registered, does that key matter any more, any way? The server does not need to use it at all. The correct certificates have been put into the correct trust stores.

srinivasupadhya:
you don't need to scrub since the file gets deleted after first read?

arvindsv:
Some conversation related to this is [here](https://gitter.im/gocd/gocd?at=55a7bc83e39909763178959e). It is also related to [this](https://github.com/gocd/gocd/issues/1082#issuecomment-120886348). So, we need to find some way.

ketan:
@arvindsv - try now? I've implemented a (low level) filter that'll intercept what's being written and comment it out if needed.

agent/test/unit/com/thoughtworks/go/agent/service/AutoregistrationPropertiesTest.java
ketan: The call to `register` is guarded by [`registerIfNecessary`](https://github.com/gocd/gocd/blob/b0dbaf0acfc8d4411c21773e8da56b7aefe88fe5/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java#L89-99).

The current implementation assumes registered implies that the agent has a GUID and the keystore has the correct certs.

So to answer your question, there is no concept of "re-register" or "un-register" that an agent can possibly initiate.

arvindsv:
Looks good to me. I can merge it. However, how do we handle the case @srinivasupadhya brought up in the [earlier comment](https://github.com/gocd/gocd/pull/1359#issuecomment-122139597).

ketan:
I'm not suggesting that the agent re-registers. Did you see my comment here. https://github.com/gocd/gocd/pull/1359#discussion_r34869456

arvindsv:
Oops. Hadn't seen that. Ok, so in that case, the autoregister file will have resources, etc. which are unused beyond the first time? Could be a little misleading. But, we have to keep it around, for us to keep reusing some plugin-related settings. Right?

ketan:
Yes, that's correct. One of the properties (`agent.elastic-agent.pluginId`) is a runtime property, the others are used only during registration.

ketan:
We could probably scrub out the other properties if needed, and to avoid future confusion.

PR #1367, opened 2015-07-20
Handle delete of working directory when it's locked by a process

chopraapooja:
Solution : Let Go to make noise(throw STDERR) when this situation happens

### DIFF
--- a/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
@@ -40,9 +40,9 @@ import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import static com.thoughtworks.go.util.ExceptionUtils.bomb;
-import static com.thoughtworks.go.util.ExceptionUtils.bombIfFailedToRunCommandLine;
+import static com.thoughtworks.go.util.ExceptionUtils.*;
 import static com.thoughtworks.go.util.FileUtil.createParentFolderIfNotExist;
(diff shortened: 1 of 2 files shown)


arvindsv:
Happens on Windows, where file locking is usually mandatory, not advisory. This happens when "Clean Working Directory" is set on a stage and a git material has its working directory locked.

With a config such as this:

```xml
<pipeline name="pipeline1">
  <materials>
    <git url="http://my.git.server" materialName="test" />
  </materials>
  <stage name="pipeline1_stage" cleanWorkingDir="true">
    <jobs>
      <job name="pipeline1_job">
        <tasks>
          <exec command="ls" />
        </tasks>
      </job>
    </jobs>
  </stage>
</pipeline>
```

Open a command prompt at the location `C:\Program Files\Go Agent\pipelines\pipeline1`, so that it is locked and cannot be deleted. Without this change, the error will be something like:

```
[go] Cleaning working directory "C:\Program Files (x86)\Go Agent\pipelines\pipeline1" since stage is configured to clean working directory
...
[GIT] Cleaning all unversioned files in working copy 
Error performing command: --- Command --- 
git clean -dff 
--- Environment --- 
{} 
--- INPUT ---- 


--OUTPUT --- 

--- ERROR --- 
ERROR: fatal: Not a git repository (or any of the parent directories): .git 
---
```

With this change, it'll say something like this:

```
[go] Cleaning working directory "C:\Program Files (x86)\Go Agent\pipelines\pipeline1" since stage is configured to clean working directory
[go] Failed to delete directory: C:\Program Files (x86)\Go Agent\pipelines\pipeline1
```

In both cases, the build will fail to start. In the new case, the error is more obvious.

Issue #1334, opened 2015-07-12
Startup Scripts should use Exec, not Eval


grahamc:
The startup scripts on the agent and server both call `eval` instead of `exec`. This has the side effect of signals being sent to the parent process (the bash script when not using nohup) do not get sent to the child java process. Using exec`s process replacement, java correctly receives the signal.

 - https://github.com/gocd/gocd/blob/master/installers/server/release/server.sh
 - https://github.com/gocd/gocd/blob/master/installers/agent/release/agent.sh

ketan:
Thank you for reporting this issue, we're at the tail end of 15.2, so we'll be considering this issue for 15.3 instead.

chopraapooja:
Using exec instead of eval to start java process in foreground looks fine here :-
- https://github.com/gocd/gocd/blob/master/installers/server/release/server.sh#L139 
- https://github.com/gocd/gocd/blob/master/installers/agent/release/agent.sh#L124 

I think, there is no need  of ```exec``` during starting java process in background (DAEMON), ```nohup``` will take cares of that.

rajiesh:
this has been QA tested on 15.3.0-2597 build on MAC OSX, linux(centOS and Ubuntu) and solaris

PR #1379, opened 2015-07-23
 #1334 Replaced 'eval' with 'exec'

chopraapooja:


installers/agent/release/agent.sh
```diff
@@ -121,5 +121,5 @@ if [ "$DAEMON" == "Y" ]; then
     eval "nohup $CMD >>$LOG_FILE 2>&1 &"
```
ketan: Can you replace this with exec as well? You may need to do `exec setsid nohup...`?

Issue #1357, opened 2015-07-16
Stages API improvements.


juhijariwala:
As we are following new API improvement and guidelines(#1100), I would like to propose a new structure for stages API response.

**Stage instance API**

URL : `curl 'http://localhost:8153/go/api/stages/myPipeline/myStages/instance/19/2' -u 'username:password'`

**Old API response**

``` json
{
  "name": "defaultStage",
  "clean_working_directory": false,
  "approved_by": "admin",
  "jobs": [
    {
      "agent_uuid": "278fb0b6-d3b8-47e1-9443-67f26bfb5c15",
      "name": "job1",
      "job_state_transitions": [
        {
          "state_change_time": 1437027001502,
          "id": 289,
          "state": "Scheduled"
        },
        {
          "state_change_time": 1437027015582,
          "id": 291,
          "state": "Assigned"
        },
        {
          "state_change_time": 1437027025589,
          "id": 292,
          "state": "Preparing"
        },
        {
          "state_change_time": 1437027027308,
          "id": 293,
          "state": "Building"
        },
        {
          "state_change_time": 1437027029321,
          "id": 294,
          "state": "Completing"
        },
        {
          "state_change_time": 1437027029339,
          "id": 295,
          "state": "Completed"
        }
      ],
      "scheduled_date": 1437027001502,
      "original_job_id": null,
      "pipeline_counter": null,
      "rerun": false,
      "pipeline_name": null,
      "result": "Passed",
      "state": "Completed",
      "id": 52,
      "stage_counter": null,
      "stage_name": null
    }
  ],
  "pipeline_counter": 19,
  "pipeline_name": "mypipeline",
  "approval_type": "manual",
  "result": "Passed",
  "counter": 2,
  "id": 33,
  "rerun_of_counter": null,
  "fetch_materials": true,
  "artifacts_deleted": false
}
```

**Improved API response**

``` json
{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/stages/mypipeline/defaultstage/instance/19/2"
    },
    "doc": {
      "href": "http://api.go.cd/#get-stage-instance"
    }
  },
  "name": "defaultStage",
  "result": "Passed",
  "counter": 2,
  "stage_type": "manual",
  "rerun_of_counter": null,
  "approved_by": {
    "_links": {
      "doc": {
        "href": "http://api.go.cd/#users"
      },
      "self": {
        "href": "http://localhost:8153/go/api/users/admin"
      },
      "find": {
[...]

PR #1395, opened 2015-07-28
Ported stage instace and history API (#1357)

juhijariwala:


server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -228,6 +228,9 @@
         patch :update, on: :member
       end
 
+      get 'stages/:pipeline_name/:pipeline_counter/:stage_name/:stage_counter/instance' => 'stages#show', constraints: {pipeline_name: PIPELINE_NAME_FORMAT, pipeline_counter: PIPELINE_COUNTER_FORMAT, stage_name: STAGE_NAME_FORMAT, stage_counter: STAGE_COUNTER_FORMAT}, as: :stage_instance_by_counter_api
+      get 'stages/:pipeline_name/:stage_name/history' => 'stages#history', constraints: {pipeline_name: PIPELINE_NAME_FORMAT, stage_name: STAGE_NAME_FORMAT}, as: :stage_history_api
+
```
ketan: @arvindsv - do we really need these constraints?

ketan: Have you given it a though about how all the new URLs (pipeline, job, stage) look like? It feels like this is a replica of the old url format. 

Is `stages/:pipeline_name/:pipeline_counter/:stage_name/:stage_counter` and `stages/:pipeline_name/:stage_name` not enough?

arvindsv: > @arvindsv - do we really need these constraints?

I guess so. I mean, it's not wrong. Also, it does make sense to get rid of invalid names at this level, than getting to the config or db and throwing them out, I suppose. What am I missing?

ketan: Fair point @arvindsv, I'm curious why the counters are defined as `PIPELINE_COUNTER_FORMAT = STAGE_COUNTER_FORMAT = /-?\d+/` When it can just be `/[0-9]+/`

Is this overloaded for some other purposes? I'd like to simplify the counter constraints if at all possible.

arvindsv: It surprised me as well, when I saw it first. However, negative numbers are valid for counters, for some reason. I've seen them on old Go Servers. After digging into the history a bit (it was waaay before the time I joined), I decided to let sleeping dragons lie (not proud of it). As far as I remember, it wasn't a negative value in DB.

ketan: I'd like to just make it an integer then. If someone files an issue, about negative values, we can address it. Sound reasonable?

arvindsv: I don't think so. It'll break something for sure. If we know we're going to break something, we should put some effort towards finding out why it was the way it was.

common/test/unit/com/thoughtworks/go/helper/StageMother.java
```diff
@@ -231,6 +233,20 @@ public static Stage cancelledStage(String stageName, String jobName) {
         return stage;
     }
 
+    public static StageInstanceModel toStageInstanceModel(Stage stage) {
```
zabil: Move this to it's own StageInstanceModelMother with a descriptive method on what it does 
Refer https://github.com/gocd/gocd/blob/22492eec4137f160f3054b4e5b9e300388023e37/server/test/unit/com/thoughtworks/go/server/ui/helper/AgentsViewModelMother.java

server/webapp/WEB-INF/rails.new/lib/extensions/java_util_date.rb
```diff
@@ -27,4 +31,5 @@ def to_long_display_date_time
     pattern = java.text.SimpleDateFormat.new("dd MMM, yyyy 'at' HH:mm:ss [Z]" )
     pattern.format(self)
   end
+
```
ketan: I know we removed `to_json` on `java.util.Date`. This may likely break existing APIs (user, backup). Can we bring this back until we call `.utc` on dates in those APIs?

Alternatively, can we implement `as_json` as follows - 

```ruby
def as_json(options={})
    self.utc.as_json(options)
end
```

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/stage_representer.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    link :doc do |opts|
+      'http://api.go.cd/#get-stage-instance'
+    end
+
+    property :getName, as: :name
+    property :result, exec_context: :decorator
+    property :getCounter, as: :counter
+    property :getApprovalType, as: :stage_type
+    property :rerun_of_stage, exec_context: :decorator, decorator: StageSummaryRepresenter
```
zabil: @arvindsv @jyotisingh 
Can you provide context for how this value is set.

In the the old API
* When the stage is re-run this value is null
* When a job is re-run the value points to first counter of the stage

It's the same in the new API. Is that correct?

zabil:
While trying to do this API right shouldn't we refer to materials that were a part of this stage?

common/test/unit/com/thoughtworks/go/helper/StageMother.java
juhijariwala: This method is not being used. I have removed it.

Issue #1186, opened 2015-06-04
Login fails with 500 error on invalid LDAP configuration


zabil:
On Experimental *release15.2.0-2019 (2358016d368f9069c48c8c6800dde93874f1cf04)*
Related to #851 (edit: might not be)

Steps to reproduce

* Admin -> Server Settings
* Enter an invalid LDAP configuration (fuzz the url)
* Save

Login with any username password

Stare with horror at

HTTP ERROR 500

Problem accessing /go/api/support. Reason:

    Server Error

Caused by:

```java
java.lang.RuntimeException: Invalid or empty ldap config, Error creating DefaultSpringSecurityContextSource
	at com.thoughtworks.go.util.ExceptionUtils.bomb(ExceptionUtils.java:40)
	at com.thoughtworks.go.server.security.LdapContextFactory.initializeDelegator(LdapContextFactory.java:68)
	at com.thoughtworks.go.server.security.LdapContextFactory.delegator(LdapContextFactory.java:50)
	at com.thoughtworks.go.server.security.LdapContextFactory.getReadOnlyContext(LdapContextFactory.java:81)
	at org.springframework.ldap.core.LdapTemplate.executeReadOnly(LdapTemplate.java:770)
	at org.springframework.security.ldap.SpringSecurityLdapTemplate.searchForSingleEntry(SpringSecurityLdapTemplate.java:190)
	at org.springframework.security.ldap.search.FilterBasedLdapUserSearch.searchForUser(FilterBasedLdapUserSearch.java:118)
```

Expected to not login and display error message on the login screen.

arvindsv:
Should fix [this](https://github.com/gocd/gocd/issues/900) too, when this is fixed?

zabil:
This issue also exists on supported release14.4.0-1356 (86064ef5ee28e0d0d12911f46fd603d15246fdd5)

chopraapooja:
Picking up for QA

rajiesh:
This issue verified on 15.3.0-2559 and working as expected. Exception is handled and user friendly message shown on browser rather than the entire stack trace.

This issue only addressed the exception handling, does not address #900

PR #1399, opened 2015-07-30
Fixes for #1186-Login fails with 500 error on invalid LDAP configuration

juhijariwala:


server/webapp/localize_ja.xml
```diff
@@ -171,6 +171,7 @@
     <entry key="USER_SUCCESSFULLY_ADDED">User ''{0}'' successfully added.</entry>
     <entry key="USER_ALREADY_EXISTS">Failed to add user. User ''{0}'' already exists.</entry>
     <entry key="USER_DOES_NOT_EXIST_IN_DB">User ''{0}'' does not exist in the database.</entry>
+    <entry key="INVALID_LDAP_ERROR">There was an error authenticating you. Please check the server logs, or contact your the go administrator.</entry>
```
zabil: Can we shorten this message to
```
Authentication failed, please contact the administrator.
```

matt-richardson: But what if I'm the administrator? (personally, I hate error messages that say "contact the administrator" when I'm the administrator, and I have to resort to googling to figure out what to do)

zabil: Administrators are all knowing super humans :-)
But yeah, what's a better message here? This error happens when an admin makes a mistake in configuring we can drop the contact administrator.

How about
```
Cannot authenticate, this might be a problem with security configuration or setup.
```

ketan:
IMO the existing message (apart from a small grammatical error) is spot on -

```
There was an error authenticating you. Please check the go server logs, or contact the go server administrator.
```

It tells the user to contact the server admin, and tells the server admin to check the server logs, where the error message along with the backtrace is being logged.

zabil:
Let's keep it at that then.
@juhijariwala can you fix the typo
```
There was an error authenticating you. Please check the server logs, or contact your the go administrator.
```

to

```
There was an error authenticating you. Please check the go server logs, or contact the go server administrator.
```

PR #1400, opened 2015-07-30
Get rid of jetty6

jyotisingh:


base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -153,7 +153,6 @@
     public static GoSystemProperty<Boolean> SHOULD_VALIDATE_XML_AGAINST_DTD = new GoBooleanSystemProperty("validate.xml.against.dtd", false);
     public static GoSystemProperty<String> JETTY_XML_FILE_NAME = new GoStringSystemProperty("jetty.xml.file.name", JETTY_XML);
 
-    public static final String JETTY6 = "com.thoughtworks.go.server.Jetty6Server";
     public static final String JETTY9 = "com.thoughtworks.go.server.Jetty9Server";
     public static GoSystemProperty<String> APP_SERVER = new CachedProperty<String>(new GoStringSystemProperty("app.server", JETTY9));
```
ketan: I don't suppose we still need this constant?

app-server/src/com/thoughtworks/go/server/util/ServletHelper.java
```diff
@@ -25,13 +25,9 @@
...
 
-    public static void init(Boolean usingJetty9) {
+    public static void init() {
         try {
-            if (usingJetty9) {
-                instance = getAppServerHelper("com.thoughtworks.go.server.util.Jetty9ServletHelper");
-            } else {
-                instance = getAppServerHelper("com.thoughtworks.go.server.util.Jetty6ServletHelper");
-            }
+            instance = getAppServerHelper("com.thoughtworks.go.server.util.Jetty9ServletHelper");
```
ketan: Do you think we can simplify this by just instantiating the class, instead of all reflection magic that is going on?

ketan: Ignore what I just said above, makes sense to have this if and when we move to a newer jetty.

ketan:
@jyotisingh - can you get rid of the jetty6 pipelines on build.go.cd?

PR #1409, opened 2015-08-03
Remove UTC conversion from as_json method as Application.rb already has Timezone configured to UTC.

juhijariwala:


rajiesh:
verified on 15.3.0-2686(6922654e1949e3e19d0ac9b8426c418ba495c001). working as expected

Issue #1392, opened 2015-07-28
New Dashboard API


jyotisingh:
#1205

rajiesh:
Sanity tested the dashboard API on 15.3.0-2663

rajiesh:
Verified the Dashboard API on 15.3.0 - 2672(47fd7343016e4a9cda3f8a871539007c2d1cf1b9). Identified a bug raised an issue #1644

Issue #37, opened 2014-04-08
Set start value for pipeline labels


agileassistance:
It would be great if we can set the start value for a pipeline label. 

This would allow us to e.g. get the revision number to continue the sequence on a release branch pipeline when we use the pipeline label for the revision number.

PR #1414, opened 2015-08-05
Dashboard API improvements(#1392)

juhijariwala:


ketan:
There are some more changes coming. Please hold off merging this.

zabil:
@ketan @juhi this is good to go right?

server/webapp/WEB-INF/rails.new/spec/helpers/application_helper_spec.rb
```diff
@@ -40,29 +40,29 @@
 
     it "should handle default_url_options" do
       url = url_for_path("/foo")
-      url.should == "/go/quux/foo?x"
+      url.should == "/foo"
```
ketan: None of these tests should have changed.

ketan:
@juhijariwala - can you also submit additional PRs to -

* remove old "undocumented" dashboard API, associated controllers, views, helpers and any java code. Looking at @arvindsv's code (#373 and #371( that should tell you what you'd need to revert `git log --grep '#37[13]\b' --name-only` and `git log --grep '#37[13]\b' -p` should tell you most relevant code :)
* documentation over at api.go.cd

In addition, please make sure that the documentation links in the current PR are correct (or ensure that the future API docs will be available at the URL)

Issue #1406, opened 2015-07-31
Make embedded SQL cross compatible


sandopolus:
There is some SQL embedded in the server project java code and some of it is not cross compatible with different databases. I am currently working on getting GoCD working with MySQL SQL compatible databases and i have identified some SQL which is currently not compatible.

MaterialRepository has 2 SQL statements at [SQL 1](https://github.com/gocd/gocd/blob/15.2.0/server/src/com/thoughtworks/go/server/persistence/MaterialRepository.java#L144-L148) and [SQL 2](https://github.com/gocd/gocd/blob/15.2.0/server/src/com/thoughtworks/go/server/persistence/MaterialRepository.java#L174-L179) which is a simple fix of changing the alias 'mod' as it is a reserved word in MySQL.

PipelineRepository has [SQL 3](https://github.com/gocd/gocd/blob/15.2.0/server/src/com/thoughtworks/go/server/persistence/PipelineRepository.java#L111-L116) which i would like to ask a question about the Cast statements.
My main question is do we need these cast statements at all?
According to the DB deltas Pipelines.name is already a VARCHAR and Materials.fingerprint is a CHAR.

It appears this code hasn't been changed since the initial commit to the github repo and so i don't know the reason/history for having it this way.
I have done some initial testing of the modified code and it seems to operate fine without the cast statements.
But thought i would get some feedback before issuing a pull request with the changes.

Thanks

arvindsv:
I've replied on the [mailing list thread](https://groups.google.com/d/msg/go-cd-dev/DxJo-z0QpyY/li0X9Rc-AwAJ).

sandopolus:
I have created pull request #1415 after the discussion on the [gocd dev mailing list](https://groups.google.com/forum/#!msg/go-cd-dev/DxJo-z0QpyY/li0X9Rc-AwAJ)

This extends the QueryExtensions interface to allow overriding of the retrieveTimeline query and fixes the queries which use the alias 'mod'

PR #1415, opened 2015-08-05
Making core server SQL cross compatible #1406

sandopolus:
* Changing the alias 'mod' as it is a reserved word for MySQL
* Adding retrieveTimeline query to the QueryExtensions interface so it can be override for different DBs
* Adding the QueryExtensions interface to PipelineRepository to remove the embedded SQL

### DIFF
--- a/server/test/unit/com/thoughtworks/go/server/persistence/PipelineRepositoryTest.java
+++ b/server/test/unit/com/thoughtworks/go/server/persistence/PipelineRepositoryTest.java
@@ -17,11 +17,13 @@
 package com.thoughtworks.go.server.persistence;
 
 
+import com.thoughtworks.go.database.QueryExtensions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 
 import com.thoughtworks.go.server.cache.GoCache;
+import com.thoughtworks.go.server.database.DatabaseStrategy;
 import com.thoughtworks.go.server.domain.user.PipelineSelections;
 import com.thoughtworks.go.util.SystemEnvironment;
 import org.hibernate.SessionFactory;
@@ -40,6 +42,8 @@ public class PipelineRepositoryTest {
     private HibernateTemplate hibernateTemplate;
     private PipelineRepository pipelineRepository;
     private SystemEnvironment systemEnvironment;
+    private DatabaseStrategy databaseStrategy;
+    private QueryExtensions queryExtensions;
 
     @Before
     public void setup() {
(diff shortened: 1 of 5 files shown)


ketan:
@sandopolus - We will try to review this PR sometime next week. Based on preliminary investigations, this PR looks good to us.

server/test/unit/com/thoughtworks/go/server/persistence/PipelineRepositoryTest.java
```diff
@@ -47,7 +51,10 @@ public void setup() {
         hibernateTemplate = mock(HibernateTemplate.class);
         goCache = mock(GoCache.class);
         systemEnvironment = mock(SystemEnvironment.class);
-        pipelineRepository = new PipelineRepository(sessionFactory, goCache,systemEnvironment);
+        databaseStrategy = mock(DatabaseStrategy.class);
+        queryExtensions = mock(QueryExtensions.class);
+        when(databaseStrategy.getQueryExtensions()).thenReturn(queryExtensions);
+        pipelineRepository = new PipelineRepository(sessionFactory, goCache,systemEnvironment, databaseStrategy);
```
zabil: None of the unit tests use ```queryExtensions``` for simplicity can we change this to

```
    ...
    private DatabaseStrategy databaseStrategy;

    @Before
    public void setup() {
       ...        
        databaseStrategy = mock(DatabaseStrategy.class);
        pipelineRepository = new PipelineRepository(sessionFactory, goCache, systemEnvironment, databaseStrategy);
        pipelineRepository.setHibernateTemplate(hibernateTemplate);
    }
```
?

sandopolus: Yes that will be fine. Will make the change the now.

sandopolus:
I have updated the Pull Request to remove the unused mocks.
If you have any other changes that you want done then let me know.
Ta

lenucksi:
This sounds quite good. @sandopolus could you please give an estimation what would  be required to create a e.g. MySQL DB backend plugin for Go.CD?

sandopolus:
I have been working through this exact problem so that i can run GoCD against an AWS Aurora DB which is MySQL compatible. 
It is not very hard to do. Just a little time consuming going through the sql queries in the ibatis sql files and working through the SQL Schema to make it MySQL compatible.

Our current implementation seems to be running ok currently with our MySQL plugin, but we haven't migrated our pipeline to GoCD yet and are still in the dev stages. So it hasn't been tested in anger yet and i know there is still a few minor things to look into.

I will see if i can put the work that has been done in the public section in GutHub so that it can be used & contributed to by others.

lenucksi:
@sandopolus That sounds very interesting. I'd highly appreciate if you could share the code! And something tells me that I could test the code 'in anger'. And also in fire. ;)

Refresh06:
@sandoplus We would be very interested in your MySQL plugin. We are hoping to boost go performance by putting the Config in a database other than H2, but we only have SqlServer and MySQL in house and setting up PostgreSQL servers for this purpose alone seems unreasonable/undesireable.

ivanmoore:
@sandopolus Is your MySQL plugin available as open source anywhere? Even if it isn't finished, I'd be interested to see what it takes to implement as I'm considering writing my own (not for MySQL).

ketan:
For anyone looking to implement a new database driver — you'd start with implementing the `Database` interface. There's already an implementation for [H2](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/database/H2Database.java#L41) that you could look at (and port to whatever database of your choice).

Depending on how different the sql dialect for your DB of choice is from H2, you may need to extend the `QueryExtensions` to add a few more methods.

PR #1423, opened 2015-08-08
Use label tag helper for auto update checkbox

gaqzi:
When using the label tag directly it didn't work in the create new
pipeline view. You had to explicitly click the checkbox instead of the label.

add_pipeline_-_go

The old code did work in the materials tab of an already created pipeline though, and it works now as well from my manual testing.

There's [another location](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/_form.html.erb#L41) that uses a `<label>` tag manually, but I couldn't figure out which page was using that code so I've left it alone for now. 

I did a search through the code and at a cursory glance it seemed to me like the other uses of the label tag without using the Rails helpers were fine.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/_form.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/_form.html.erb
@@ -38,7 +38,7 @@
 
                         <div class="form_item_block checkbox_row material_options">
                             <%= f.check_box(com.thoughtworks.go.domain.scm.SCM::AUTO_UPDATE, {:include_blank => true, :class => "form_input"}, "true") -%>
-                            <label for="material_<%= com.thoughtworks.go.domain.scm.SCM::AUTO_UPDATE -%>"><%= l.string('SHOULD_AUTO_UPDATE') -%></label>
+                            <%= f.label(com.thoughtworks.go.domain.scm.SCM::AUTO_UPDATE, l.string('SHOULD_AUTO_UPDATE')) %>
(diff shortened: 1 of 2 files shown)


ketan:
A quick git grep reveals that this is used in views rendered via `PluggableScmController`. The routes.rb file will tell you the routes that can be used to access that controller. But I think your existing change looks good, so please go ahead and change the other location.

```
% git grep 'pluggable_scm/form'
server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/edit.html.erb:<%= render :partial => "admin/materials/pluggable_scm/form.html", :locals => { :scope => { :material => @material, :url => admin_pluggable_scm_update_path(:plugin_id => @material.getPluginId()), :method => "PUT" , :meta_data_store => @meta_data_store}} -%>
server/webapp/WEB-INF/rails.new/app/views/admin/materials/pluggable_scm/new.html.erb:<%= render :partial => "admin/materials/pluggable_scm/form.html", :locals => { :scope => { :material => @material, :url => admin_pluggable_scm_create_path(:plugin_id => @material.getPluginId()), :method => "POST" , :meta_data_store => @meta_data_store}} -%>
```

gaqzi:
@ketan thanks! :)
I've changed the pluggable scm part to also use the rails helper. I installed the GitHub SCM plugin and then I managed to click my way to see it properly rendered in place.

juhijariwala:
This PR looks fine to me. Checked the "Poll for new changes" label for normal material as well as GitHub SCM plugin material. @ketan Can you merge this?

Issue #1404, opened 2015-07-31
Redesign the pipeline configuration page


zabil:
Linked and part solution to #838 

The current admin screen posts every operation (save stage, jobs etc) to the server. This is turning out to be a painful experience for the user when the size of the config file is large as we validate, merge, save and reload the config.

To solve the problem we propose 
- Converting the page to have a save operation that will post the whole pipeline configuration when stages/jobs/tasks etc are modified.
- Exposing an API that will validate and save the pipeline configuration intelligently without having to validate the whole of config xml.
-  Maintaining the state of Added/Modified/Deleted of stages, jobs or tasks in the browser till the save/cancel button is clicked.

@ketan @arvindsv @srinivasupadhya please feel free to add more details.
#### Checklist of Items to be done
- #1722 General Options
- #1723 Project Management
- #1724 Materials
- #1725 Stages
- #1726 Stage
- #1605 Auto-complete for resources
- #1727 Tasks
- #1729 Navigation  
- #1728 Save
- #1733 Job Configuration
### Technical
- Trim down the list of routes in `js-routes`
- Handle situations when a plugin is removed, but configuration exists 
  ![screen shot 2015-09-03 at 3 00 47 pm](https://cloud.githubusercontent.com/assets/10598/9655109/8fd14286-524c-11e5-81fe-fa7257578d07.png)

zabil:
Showdown for UI framework to be adopted happening at
https://github.com/gocd/spikes

zabil:
Post initial spike we are convinced on using Mithril.
We'll use it to model a better config edit page and see how it goes. 

Some notes on Mithril
* Better handling of binding and intelligent redraws using the virtual dom.
* Support for traditional MVC, routing and template in a uncompressed 12kb library
* Cleaner template system even without the use of JSX/MSX   
* Support for using plain javascript idioms without being tied to a framework.
* Excellent documentation (partly due to the simplicity of the framework)

Mithril code: [app.js](https://github.com/gocd/spikes/blob/4277dc1c0ec43c6fd4d13001653d8f6f2918d9a7/ui/frameworks/mithril/app.js)
React code: [jobs-view.jsx] (https://github.com/gocd/spikes/blob/4277dc1c0ec43c6fd4d13001653d8f6f2918d9a7/ui/frameworks/react/jobs-view.jsx)

* Fewer lines of code without using magic or conventions

srinivasupadhya:
Possible Config Validation Improvements:

*major*:

[...]

PR #1430, opened 2015-08-11
#1404 - Add some basic structure to render a single page app using require.js

ketan:
* Modularize the SPA using require.js
* Use mithril.js to create a very basic SPA.
* Add some tests using jasmine-rails

#### Checklist of things, so I don't forget about them

* [ ] Handle situations when a plugin is removed, but configuration exists 
screen shot 2015-09-03 at 3 00 47 pm

* [ ] Trim down the list of routes in `js-routes`
* [ ] Are all go servers prefixed with `/go` even when fronted by apache?
* [ ] Do we want to use [pretty cron](https://github.com/azza-bazoo/prettycron) to help users figure out cron syntax? See [this page](https://snap-ci.com/gocd/api.go.cd/branch/master/configuration/cron) for an example.
* [ ] Perform simple cron validation? Nothing comprehensive, just basic stuff like maybe if there are sufficient number of fields?

* [ ] Handle material plugins
* [ ] Handle pipeline templates
* [ ] Custom command look ups (discuss on if need)
* [ ] Upstream pipeline dependencies for pipeline material
* [ ] Roles and permissions on stage trigger
* [ ] Auto-complete for resources
* [ ] Automatic pipeline scheduling checkbox (map to stage)
* [ ] Field level tooltip help
* [ ] Testing connection of materials
* [ ] Ordering of tasks
* [ ] Job run information and timeouts


ketan:
@all - We're pretty close to "done" wrt stabilising on the models and a basic view with some interactions.

At this point, I think the PR is ready to be reviewed by someone even if just to get a feel for whether you like the code and some of the patterns we've implemented.

arvindsv:
Took a *very* quick look. Some things that I felt were done consistently (patterns) were `Mixins.HasMany`, `Mixins. fromJSONCollection`, the `fromJSON` method and the view files are named `_widget.js.msx`. The specs are nice and simple too. Why don't you mention a few others which can help people start? Seeing 2800 files changed is sure to be daunting. :)

A few questions (remember I don't have enough context about the implementation, my questions might already be discussed):

1. Routing: I didn't see any routes added. Are you planning to add routes to sections like the materials, etc. so that we can provide a link from somewhere in the UI to a specific part of this? I think that would be nice. Maybe later, if not now.

2. What about handling conflicts (other user has edited this pipeline)? That'll come up later?

3. What about server-side validations? Do you have a good idea about how you'll handle that?

zabil:
> Mixins.HasMany, Mixins.fromJSONCollection, the fromJSON


Yup, we are also considering the idea of making this discoverable, by implementing the a model wrapper

```
var Pipeline = m.model(function(data){
 this.name = m.prop(data.name);
 this.stages = m.prop(data.name);
});

Pipeline.hasMany(Stages).withUnique('name');

var pipeline = new Pipeline.create(data);
```

> Routing: I didn't see any routes added


Although mithril has [routing](https://lhorie.github.io/mithril/routing.html) we are not using it on htis page. The whole page is rendered and elements are hidden/shown based on interactions. We render the stages and jobs etc via the pipeline model and view (as child widgets) and bind them for updates. 
Routing would have forced us to render the widgets as top level components making it tougher to manage updates to the top level model.

> What about handling conflicts (other user has edited this pipeline)? That'll come up later?

Yes, we haven't figured out the UX yet. But theoretically we can can re-render the model on another dom element

```
    var pipelineConfigElem = $('#pipeline-config-new');
    m.mount(pipelineConfigElem.get(0), PipelineConfigWidget(pipelineConfigElem.attr('data-pipeline-api-url')));
```

The user can then compare without the view losing his changes.

> What about server-side validations? Do you have a good idea about how you'll handle that?

PR #1437, opened 2015-08-14
Also remove ignored files when cleaning a git working dir

tiwoc:
If `git clean` is used without -x, all files that are listed in .gitignore survive the cleaning.

This has bitten me hard when trying to let gocd build cURL, which uses an ignored build subdirectory. Some remnants of old builds that survived the cleaning led to a subtly broken library.


zabil:
This change is the same as #1059

tiwoc:
Thanks! I'll add a comment on the other issue.

PR #1464, opened 2015-09-01
#795 - Remove old CCTray code and its toggle

arvindsv:
It has been toggled on and has not caused problems in the last two releases.

Run on go05 passed.


ketan:
:+1: my favorite kind of change :)

rajiesh:
CCTray tests are part of regression and passed in 15.3.0-2682 build. Also the toggle is removed

PR #1476, opened 2015-09-03
Move to using uglifier for javascript and css compression, instead of the deprecated yui

ketan:


juhijariwala:
Verified on [build.go.cd](http://build.go.cd)  Go Version: 15.3.0 https://github.com/gocd/gocd/commit/fb9075d3155d82d481ba062ead6fe36262c7785c
It works fine. No issue observed on the [build.go.cd](http://build.go.cd) after change got merged.

Issue #1343, opened 2015-07-14
Cannot delete user with special character 


Ehekatl:
I try this: 
    
    curl -k 'https://******.com/go/api/users/********@gmail.com'       -u '******:******'       -H 'Accept: application/vnd.go.cd.v1+json'       -X DELETE
     {
      "message": "Either the resource you requested was not found, or you are not authorized to perform this action."
    }
     

# 
    
    curl -k 'https://******.com/go/api/users/********%40gmail%2Ecom'       -u '******:******'       -H 'Accept: application/vnd.go.cd.v1+json'       -X DELETE
     {
      "message": "Either the resource you requested was not found, or you are not authorized to perform this action."
    }
     
Why not just add a button for delete user?

ketan:
This API is only available starting 15.2 (which is going to be out in a couple of days).

What version are you on?

srinivasupadhya:
This is continued from here - https://github.com/srinivasupadhya/gocd-oauth-login/issues/2
i assume he used v15.2 since the GitHub Authentication plugin is working for him

Ehekatl:
My version is 15.2.0-2246, and if I want to remove a environment, I need edit config file...which is very inconvenient, if I want to remove a pipeline and add a new one with the same name, old pipeline just magically appear again...so the pipeline data didn't delete in db?

arvindsv:
Yes, pipeline delete [does not delete](https://github.com/gocd/gocd/issues/13#issuecomment-73181595) from the DB, since there can be dependencies on the pipelines. Here is an [open issue](https://github.com/gocd/gocd/issues/879) about providing that ability. However, let's keep this specific issue about one problem. :) Deleting a user.

lorijoan:
+1 to this - same problem is affecting my instance with just full stops included in the username.

lorijoan:
Has the API moved on from v1?  Removing t -H 'Accept: application/vnd.go.cd.v1+json' \ from the suggested curl notation (http://api.go.cd/current/#update-a-user) has allowed me to delete successfully.

ketan:
Can someone change the routes.rb wrt UsersController to this to verify if it works as expected?

```ruby
      resources :users, param: :login_name, only: [:index, :show, :destroy], constraints: { login_name: USER_NAME_FORMAT } do
        patch :update, on: :member
      end
```

arvindsv:
Checked. It works.

rajiesh:
[...]

PR #1477, opened 2015-09-03
#1343 - Allow accessing user resources for login names containing special characters

ketan:


ketan:
Merging since @arvindsv confirmed that the fix works. Unfortunately it is not possible to check the "routing" with api endpoints because it is managed by rack middleware.

PR #1479, opened 2015-09-04
minor improvements to plugin java api

ketan:


config/config-api/src/com/thoughtworks/go/domain/config/ConfigurationProperty.java
```diff
@@ -187,6 +187,10 @@ public ConfigErrors errors() {
         return configErrors;
     }
 
+    public boolean hasErrors(){
+        return !configErrors.isEmpty();
```
srinivasupadhya: is this being used somewhere?

server/src/com/thoughtworks/go/server/service/tasks/PluggableTaskService.java
```diff
@@ -75,5 +75,7 @@ public void validate(final PluggableTask modifiedTask) {
         for (ValidationError validationError : validationResult.getErrors()) {
             modifiedTask.getConfiguration().getProperty(validationError.getKey()).addError(validationError.getKey(), validationError.getMessage());
         }
+
+        return validationResult.isSuccessful();
```
srinivasupadhya: is this being used somewhere?

config/config-api/src/com/thoughtworks/go/domain/config/ConfigurationProperty.java
ketan: The next set of commits use it, it is used in one of the (new) controllers to validate config properties of plugin tasks in an AJAXy manner.

server/src/com/thoughtworks/go/server/service/tasks/PluggableTaskService.java
ketan: Same as above, used by the next set of commits.

juhijariwala:
Verified on Go Version: 15.3.0(2682-566a58ea0395320403d43b5adb10ca6f99c308e6).
Works fine.

PR #1499, opened 2015-09-16
Move to using therubyrhino for a JS runtime that works

ketan:
... instead of relying on node or windows scripting host


ketan:
Closing in favor of just having nodejs on machines.

PR #1502, opened 2015-09-17
Improve error message grammar

sghill:


ketan:
Thanks for the fix :)

PR #1503, opened 2015-09-17
Revert "#1404 - Add some basic structure to render a single page app …

arvindsv:
Reverts gocd/gocd#1430. Failing on Windows and sometimes on go05.


arvindsv:
I'll run tests against this locally, and then merge it in. build should go back to green after this.

PR #1510, opened 2015-09-21
#M8104 Using the full path for plugin controller

zabil:


arvindsv:
Will need to change auth plugins too.

jyotisingh:
@arvindsv - We don't need to change the plugins, they should already be using the correct url. Or, are you referring to some controller that handles that?

arvindsv:
Ah ok, if the URL was already `/plugin/interact/...` then it's fine. Sorry.

jyotisingh:
Yes, none of the urls were changed. We just need to explicitly specify them in the controllers.

PR #1512, opened 2015-09-22
Removal of pipeline groups in search when there are no pipelines that match

justinholmes:
It's messy/annoying having to scroll through pipeline groups if the search terms do not match any pipelines.


grahamc:
:+1: a feature like this would be quite nice.

alex-leonhardt:
:+1:

rajiesh:
verified on 15.3.0 - 2672 (47fd7343016e4a9cda3f8a871539007c2d1cf1b9). Search now shows only the groups which has pipeline names matching the search string

PR #1520, opened 2015-09-25
Removing sample-plugins from this repo

zabil:
it's now maintained at http://github.com/gocd/sample-plugins


ketan:
@zabil - can you look at the build failures?

zabil:
@ketan I the fix in. Can't get to trigger the build with the right changes because I force pushed an amend.

ketan:
@zabil - it sometimes take a couple of minutes for the github plugin to poll all available heads. I've triggered another build

Issue #1523, opened 2015-09-27
Pipeline status API to also get comment and paused by


therako:
current pipeline status api only gives Paused true/false. It's useful to also get the comment on the paused pipeline and who paused the pipeline.

ketan:
Reopening, someone on the team will verify this (and add additional tests, if needed)

ketan:
PR #1524

rajiesh:
verified pipeline status api for paused and non- paused pipelines with cause message on 15.3.0-2663. Paused By user and paused cause both are availble now. The story can be closed

PR #1524, opened 2015-09-27
#1523 - Adding pipeline paused by and comment to the pipeline status API.

therako:


ketan:
@juhijariwala - do we have these attributes published in the new API as well?

PR #1528, opened 2015-09-29
Modified StageXmlViewModel to expose duration property via API xml format

ffafara:
Simple change to expose stage duration via XML feed in the API:
GET /go/api/stages/:stage_id.xml

The duration property is the same as is shown in the UI.

### DIFF
--- a/server/src/com/thoughtworks/go/server/domain/xml/StageXmlViewModel.java
+++ b/server/src/com/thoughtworks/go/server/domain/xml/StageXmlViewModel.java
@@ -64,7 +64,7 @@ public class StageXmlViewModel implements XmlRepresentable {
 
         root.addElement("approvedBy").addCDATA(stage.getApprovedBy());
 
-        root.addElement("duration").addText(stage.getDuration().duration(RunDuration.PERIOD_FORMATTER));
+        root.addElement("duration").addText(getStageDurationForXml(stage));
 
         Element jobs = root.addElement("jobs");
         for (JobInstance jobInstance : stage.getJobInstances()) {
@@ -76,5 +76,13 @@ public class StageXmlViewModel implements XmlRepresentable {
 
     public String httpUrl(String baseUrl) {
         return httpUrlFor(baseUrl, stage.getId());
+    }
+
+    public String getStageDurationForXml(Stage stage) {
+        try {
+            return stage.getDuration().duration(RunDuration.PERIOD_FORMATTER);
+        } catch (NullPointerException e) {
+            return "N/A";
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/gocd/gocd/pull/1528" target="_blank">ReviewNinja

ketan:
@ffafara - can you fix some of the build failures?

ffafara:
I created a fix for java.lang.NullPointerException that was causing the build errors.
@ketan - can you run it through the pipeline?

server/src/com/thoughtworks/go/server/domain/xml/StageXmlViewModel.java
```diff
@@ -74,4 +77,12 @@ public Document toXml(XmlWriterContext writerContext) {
     public String httpUrl(String baseUrl) {
         return httpUrlFor(baseUrl, stage.getId());
     }
+
+    public String getStageDurationForXml(Stage stage) {
+        try {
+            return stage.getDuration().duration(RunDuration.PERIOD_FORMATTER);
+        } catch (NullPointerException e) {
```
zabil: @ffafara thanks a lot for helping out with this but I think we should handle this better.

If you check [here] (https://github.com/gocd/gocd/blob/ad8fc27d552f038fc8609c27cd0d6e4eadd90739/common/src/com/thoughtworks/go/domain/Stage.java#L124) it mentions that the ```lastTransitionedTime``` will not be null when a job is completed. So this is a problem with test data set up.  

```
    public RunDuration getDuration() {
        if (!state.completed()) {
            return RunDuration.IN_PROGRESS_DURATION;
        }
        Date start = new Date(createdTime.getTime());
        Date end = new Date(lastTransitionedTime.getTime());
        return new RunDuration.ActualDuration(new Duration(end.getTime() - start.getTime()));
    }
```

The above piece of code will not throw a null pointer. We also do not want to return a undefined status like "N/A"

Also while you are at it  there are a few problems with [```RunDuration```](https://github.com/gocd/gocd/blob/ad8fc27d552f038fc8609c27cd0d6e4eadd90739/common/src/com/thoughtworks/go/domain/RunDuration.java) we need to fix.

I think duration should not expose the formatter as it does now. so

```
public static final PeriodFormatter PERIOD_FORMATTER 
```
should be

```
public static final PeriodFormatter PERIOD_FORMATTER 
```

```
public abstract String duration(PeriodFormatter formatter);

``` 
should be

```
public abstract String duration();
```

```
       public String duration(PeriodFormatter formatter) {
            return "In Progress";
        }
```

Should be

```
       public String duration() {
            return "In Progress";
        }
```

and 

```
       public String duration(PeriodFormatter formatter) {
            return formatter.print(duration.toPeriod());
        }
```

should be

```
       public String duration() {
            return PERIOD_FORMATTER.print(duration.toPeriod());
        }
```

So when serializing to xml it ends up like this

```
root.addElement("duration").addText(stage.getDuration().duration())
```

ketan:
@ffafara  — would you mind looking into @zabil's comments so we can merge this long pending PR?

Thanks

zabil:
Will re-open if when the responses are fixed.

PR #1534, opened 2015-10-01
Remove @Deprecated PipelineInstanceGroupModel#getConfig()

sghill:
* Switch callers over to getStages(), which returns the same object
* Rename PipelineInstanceGroupModel's `config` -> `stages` for consistency with the getter


sghill:
Hi Go Team!

I realize a pull request like this with no explanation may look a little odd. I've been trying to get familiar with the codebase and removing this deprecated method looked like a safe way to contribute. If it's riskier than I thought or now isn't the right time, I'm happy to hold off.

Thanks!
Steve

ketan:
> 16 additions and 22 deletions

Hey this is a great contribution. The best ones are the ones that just delete code :) Less code is better for the next person to read things. The less the better :)

PR #1333, opened 2015-07-11
1133 config repo polling, second queue

tomzo:
This is fifth part of #1133

This introduces new MDU workers and second material queue. Details in links.

### Discussions
We talked about this:
 * https://github.com/gocd/gocd/issues/1133#issuecomment-115288970
 * https://github.com/gocd/gocd/issues/1133#issuecomment-115306750
 * https://github.com/gocd/gocd/issues/1133#issuecomment-115809761 - I actually managed to get this 3 ideally points.
 * https://gitter.im/gocd/gocd?at=558d93ea3e94a76222eaf48c

(Good) **Description of final design which is implemented here** - https://github.com/gocd/gocd/issues/1133#issuecomment-116826923

### Review

Comparison 4th to 5th part - https://github.com/tomzo/gocd/compare/1133-minimal-pr...tomzo:1133-pollers-pr

PR #1536, opened 2015-10-02
1133 scm-config consistency

tomzo:
This should be merged after #1333 , until then [here](https://github.com/tomzo/gocd/compare/1133-pollers-pr...tomzo:1133-scm-cons-pr) is the best view of work related only to this topic.

Changes are aimed at the problem we called *scm-config consistency*. There is a very interesting case when pipeline configuration is stored in the same repository that one of pipeline materials. In these cases we want to make sure that *any* pipeline is run on the configuration and material from same revision.

I have covered also other cases (including manual triggers, re-runs and when pipeline configuration is in repository different than one of materials) and I will put them in description of #1133 soon.

Most of the changes are new integration tests.

Services which required changes are all related to pipeline scheduling and build cause producer.

This could be reviewed by @jyotisingh @srinivasupadhya or @arvindsv .


server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
```diff
@@ -126,14 +129,14 @@ public void autoSchedulePipeline(String pipelineName, ServerHealthStateOperation
         }
     }
 
-    public void manualSchedulePipeline(Username username, PipelineConfig pipelineConfig, ScheduleOptions scheduleOptions, OperationResult result) {
-        long trackingId = schedulingPerformanceLogger.manualSchedulePipelineStart(CaseInsensitiveString.str(pipelineConfig.name()));
+    public void manualSchedulePipeline(Username username, String pipelineName, ScheduleOptions scheduleOptions, OperationResult result) {
```
jyotisingh: could you change `String pipelineName` to `CaseInsensitiveString pipelineName` instead as WaitForPipelineMaterialUpdate seems to use CaseInsensitiveString anyhow.

server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
```diff
@@ -246,6 +251,46 @@ ServerHealthState newProduceBuildCause(PipelineConfig pipelineConfig, BuildType
...
+                                           boolean materialConfigurationChanged,MaterialRevisions peggedRevisions) {
+        if(buildCause == null)
+            return false;
+
+        boolean validCause = buildType.isValidBuildCause(pipelineConfig, buildCause);
+
+        if(pipelineConfig.isConfigOriginSameAsOneOfMaterials())
+        {
+            /*
+            if(!peggedRevisions.isEmpty())
```
jyotisingh: could you remove the commented out code

config/config-server/src/com/thoughtworks/go/config/parts/XmlPartialConfigProvider.java
```diff
@@ -0,0 +1,126 @@
...
+    {
+        this.loader = loader;
+    }
+
+    @Override
+    public PartialConfig load(File configRepoCheckoutDirectory, PartialConfigLoadContext context) {
+        File[] allFiles = getFiles(configRepoCheckoutDirectory, context);
+
+        // if context had changed files list then we could parse only new content
+
```
jyotisingh: Please update method names to conform to java naming convention for - ParseFiles, CollectFragments, ParseFile

server/src/com/thoughtworks/go/server/materials/ConfigMaterialUpdater.java
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+        }
+        finally {
+            // always post the original message further
+            // this will remove material from inProgress in MUS
+            topic.post(message);
+        }
+    }
+
```
jyotisingh: So, `onMessage(MaterialUpdateCompletedMessage message)` would happen on a successful/failed MDU of a config-repo-material. In case of a successful MDU, the code seems to be calling `UpdateConfigurationFromCheckout ` which calls `poller.checkout` all over again. The folder here is the flyweight folder which would have been updated during the MDU already. Is that intentional? I am still trying to get my head around this, so there is a chance I might have missed out on one of your comments.

tomzo: > The folder here is the flyweight folder which would have been updated during the MDU already. Is that intentional?

You think that flyweight folder has the most recent checkout after MDU is done? I thought so too, initially... 
Apparently MDU does not perform checkouts. It does not have to. It only queries the remote repository for recent commits. E.g. in svn there only the very first checkout in flyweight directory.
I had to add call to `checkout` it after debugging why flyweight directory has old revisions.
Is that what you mean?

server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
```diff
@@ -246,6 +251,46 @@ ServerHealthState newProduceBuildCause(PipelineConfig pipelineConfig, BuildType
         }
     }
 
+    private boolean isGoodReasonToSchedule(PipelineConfig pipelineConfig, BuildCause buildCause, BuildType buildType,
+                                           boolean materialConfigurationChanged,MaterialRevisions peggedRevisions) {
+        if(buildCause == null)
+            return false;
+
+        boolean validCause = buildType.isValidBuildCause(pipelineConfig, buildCause);
+
```
jyotisingh: Any specific reason for ignoring `materialConfigurationChanged` flag when `pipelineConfig.isConfigOriginSameAsOneOfMaterials()` == true?
If not, would rewriting this if-else block to be something like this make sense:

```
        if (pipelineConfig.isConfigOriginSameAsOneOfMaterials()) {
            // then we need config and material revisions to be consistent
            if (!buildCause.isForced() && !buildCause.pipelineConfigAndMaterialRevisionMatch(pipelineConfig)) {
                return false;
            }
        } else {
            return materialConfigurationChanged || validCause;
        }
```

jyotisingh: Also, please get rid of the commented-out code and peggedRevisions parameter from the method, if its not required.

server/src/com/thoughtworks/go/server/materials/ConfigMaterialUpdater.java
jyotisingh: Yes, thats what I was talking about. In case of git and Hg it would always fetch the latest changes during an MDU, but you are right about svn. `modificationSince` calls for SVN just queries the svn repo to report back new commits. 
So far I wasn't thinking much about other scms, but this makes me think, as of now there is no flyweight folder created for Svn, TFS and P4. The flyweight folder really comes into picture for git and hg alone. `updateTo` method in `Material` interface was called only on the agent side so far. With this PR, the server would start checking out these other material types as well, ofcourse for the config-repo alone.

tomzo: > So far I wasn't thinking much about other scms, but this makes me think, as of now there is no flyweight folder created for Svn, TFS and P4. The flyweight folder really comes into picture for git and hg alone. updateTo method in Material interface was called only on the agent side so far. With this PR, the server would start checking out these other material types as well, ofcourse for the config-repo alone.

I can't find code that explicitly creates the flyweight folder. I did touch the code that returns path of it. Do you suggest that we should create flyweight directory for svn, TFS and P4? IMO when server will call `updateTo` then `svn checkout` command will create the directory.

jyotisingh: Possibly, but P4 and TFS work a little differently in the sense that they create some workspaces that they work out of. I am not expert at them, so would go ahead with what we have here and test/fix those cases separately.

server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
tomzo: > ignoring materialConfigurationChanged flag when pipelineConfig.isConfigOriginSameAsOneOfMaterials() == true

When configuration and pipeline material are in the same repository and there is **some** build cause then it can be either manual trigger or a change in one of the materials. The behavior we want is to run build always when manually forced or build a *scm-config consistent* revision. If revision changed then config has changed.
What you have written is the same as
```
 if (buildCause.isForced() || buildCause.pipelineConfigAndMaterialRevisionMatch(pipelineConfig)) {
                return false;
            }
```
I guess you meant `return true`
But this is not desired.

I'd like to stick with:
```
            if(buildCause.isForced())
            {
                // build is manual - skip scm-config consistency
                return validCause;
            }
            // then we need config and material revisions to be consistent
            if(!buildCause.pipelineConfigAndMaterialRevisionMatch(pipelineConfig))
            {
                return false;
            }

            return validCause;
```
Let me know if it is not OK.

jyotisingh:
@tomzo  - There are some [tests failing](https://build.go.cd/go/tab/build/detail/build-linux-PR/120/build-server/3/server-split-runInstance-2), will merge once you provide a fix.

server/src/com/thoughtworks/go/server/materials/ConfigMaterialUpdater.java
arvindsv: Does it make sense to spend effort specifically for P4 and TFS, given that they're more cumbersome to manage and can potentially cause problems if we don't get the workspaces right? If it's little effort (and it might be), then, it's ok.

tomzo: I wanted to play nice and not exclude any types of materials just because I don't use them. The effort is few lines for each type.
But there are integration tests which require some significant effort in which I totally ignored them. We'll see if anyone complains. I guess it is better to provide some, perhaps working, support than just throwing exception.

tomzo:
I have fixed the 2 failing tests and synced with master. These failures probably were not a regression, I must have skipped them when developing.
This is ready to merge from my perspective.

jyotisingh:
Alright, reviewed all over again since its been a while, changes in this PR looks fine to me. There is just one open scenario that @tomzo and I do not agree on yet(ie. when there is a valid main-config that gets checkedin but invalid merged config, then the checkedin version of the config is not reflected in the cache MergedGoConfig.currentConfig so even though the config is checked-in to config repo, the UI and APIs would not reflect the change), but that change is not part of PR anyway. Merging it.

jyotisingh:
Ah did not see https://github.com/gocd/gocd/pull/1827 earlier which addresses the last comment

PR #1537, opened 2015-10-03
Start services at runlevel 2 on Debian-like systems.

jmkeyes:
While trying to automate the deployment of Go using Puppet on a Debian 7 (Wheezy) server, I found that the init script would always be marked as disabled by Puppet, even when specifically marked as enabled, and the service would not start on boot without manual intervention:

```
root@localhost:~# puppet resource service go-server enable=true
Notice: /Service[go-server]/enable: enable changed 'false' to 'true'
service { 'go-server':
  ensure => 'running',
  enable => 'false',
}
```

I tracked this down to commit 1efeb1b, as part of PR #669, which altered the default runlevels for the ini-script so it would be enabled on runlevels to 3-5 and disabled on runlevel 2 on all systems.

This is fine for Redhat-like systems (as they boot to runlevel 3 by default) but will prevent any Debian-like system from enabling the service (as they boot to runlevel 2 by default).

This patch updates the init-script metadata on Debian-like systems to include runlevel 2 by default.

### DIFF
--- a/installers/agent/release/init.cruise-agent
+++ b/installers/agent/release/init.cruise-agent
@@ -22,8 +22,8 @@
 # Provides: go-agent
 # Required-Start: $network $remote_fs
 # Required-Stop: $network $remote_fs
-# Default-Start: 3 4 5
-# Default-Stop: 0 1 2 6
+# Default-Start: 2 3 4 5
+# Default-Stop: 0 1 6
 # Description: Start the Go Agent
 ### END INIT INFO
 SERVICE_NAME=${0##*/}
--- a/installers/server/release/init.cruise-server
+++ b/installers/server/release/init.cruise-server
@@ -21,8 +21,8 @@
 # Provides: go-server
 # Required-Start: $network $remote_fs
 # Required-Stop: $network $remote_fs
-# Default-Start: 3 4 5
-# Default-Stop: 0 1 2 6
+# Default-Start: 2 3 4 5
+# Default-Stop: 0 1 6
 # Description: Start the Go Server
 ### END INIT INFO
 


ketan:
@jmkeyes - Would you mind signing the [CLA](http://www.go.cd/contribute/cla.html).

Additionally you'd need to change the line `chkconfig` to `2345 90 10` instead of `345 90 10`, so that `chkconfig` behaves correctly. Please squash the commits once you do.

Thanks

---

**tl;dr;** (for the next person who comes to this issue) –

As a matter of [policy](http://www.debian.org/doc/debian-policy/ch-opersys.html#s9.3.3.1) (debian and therefore ubuntu) do not define any differences between run levels 2-5.

On RH and derivatives (like centos), level 2 is undefined.

I looked at a few init scripts on centos and here is what I found –

```
# nginx
# chkconfig: - 85 15
# Default-Start: 2 3 4 5
# Default-Stop: 0 1 6
```

```
# ssh
# chkconfig: 2345 55 25
# Default-Start: 2 3 4 5
# Default-Stop: 0 1 6
```

```
#postfix
# chkconfig: 2345 80 30
# Default-Start: 2 3 4 5
# Default-Stop: 0 1 6
```

jmkeyes:
In that case the patch that I'd make just reverts 1efeb1b. I don't consider that to be a contribution.

ketan:
@jmkeyes - you are right about reverting the change.

@arvindsv since you're the one who originally merged this. WDYT about reverting the change, based on the info in the comment above?

Most networky services on ubuntu seem to be using the same runlevels as as centos just like my comment above.

I'm +1 to revert the original commit.


/cc @iliasbartolini (how do you do :)

arvindsv:
Shouldn't we then have different values for DEB and RPM?

ketan:
@arvindsv - as per my [comment above](
https://github.com/gocd/gocd/pull/1537#issuecomment-145200792), we should not need different runlevels.

arvindsv:
Hmm, ok. Revert it. I'll have to check the docker images, which are based on debian.

PR #1542, opened 2015-10-06
Add API endpoint to create a user.

ketan:
The contract is same as that of the `PATCH` API call, with an
additional REQUIRED FIELD `login_name`.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/users_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/users_controller.rb
@@ -28,16 +28,35 @@ module ApiV1
     end
 
     def update
-      result          = HttpLocalizedOperationResult.new
-      checkin_aliases = if params[:checkin_aliases].is_a?(Array)
-                          params[:checkin_aliases].join(',')
-                        else
-                          params[:checkin_aliases]
-                        end
-
-      @user_to_operate = user_service.update(@user_to_operate, to_tristate(params[:enabled]), to_tristate(params[:email_me]), params[:email], checkin_aliases, result)
+      result           = HttpLocalizedOperationResult.new
+      @user_to_operate = save_user(result, @user_to_operate)
       if result.isSuccessful
         render :json_hal_v1 => UserRepresenter.new(@user_to_operate).to_hash(url_builder: self)
+      else
+        render_http_operation_result(result)
+      end
+    end
+
+    def create
+      result = HttpLocalizedOperationResult.new
+
+      user, created = user_service.withEnableUserMutex do
+        user = user_service.findUserByName(params[:login_name])
+        if user.instance_of?(com.thoughtworks.go.domain.NullUser)
+          user = save_user(result, com.thoughtworks.go.domain.User.new(params[:login_name]))
+          [user, true]
+        else
+          [user, false]
+        end
+      end
+
+      unless created
+        return render_message("The user `#{params[:login_name]}` already exists.", :conflict)
+      end
+
+      if result.httpCode == 200
+        response.location = apiv1_user_url(login_name: params[:login_name])
+        render :json_hal_v1 => UserRepresenter.new(user).to_hash(url_builder: self), status: :created
       else
         render_http_operation_result(result)
       end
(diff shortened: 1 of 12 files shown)


ketan:
I'll submit a PR for the api docs when this is merged.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/users_controller.rb
```diff
@@ -43,12 +37,41 @@ def update
       end
     end
 
+    def create
+      @user_to_operate = user_service.findUserByName(params[:login_name])
+
+      unless @user_to_operate.instance_of?(com.thoughtworks.go.domain.NullUser)
```
jyotisingh: Checking for existence of user over here is fine, but there could possibly be a race condition. [UserService.update](https://github.com/gocd/gocd/blob/956f8baab3b01eaea52fa345a802e0568933ce0b/server/src/com/thoughtworks/go/server/service/UserService.java#L124) internally calls [saveOrUpdate](https://github.com/gocd/gocd/blob/956f8baab3b01eaea52fa345a802e0568933ce0b/server/src/com/thoughtworks/go/server/service/UserService.java#L367) which does not check if the user already exists. Have a look at [addUserIfDoesNotExist](https://github.com/gocd/gocd/blob/956f8baab3b01eaea52fa345a802e0568933ce0b/server/src/com/thoughtworks/go/server/service/UserService.java#L350) instead.

ketan: I'm aware of this problem with possible race condition. However `addUserIfDoesNotExist` is not what I'm looking to do – I want to raise an error if a user already exists, or create if none exists, unless you mean to use `enableUserMutex` to acquire a lock.

jyotisingh: I din't mean to say use `addUserIfDoesNotExist`, but use `enableUserMutex` as used in this method to avoid race-conditions.

ketan:
@jyotisingh - please review again.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/users_controller.rb
```diff
@@ -43,12 +37,48 @@ def update
       end
     end
 
+    def create
+      result = HttpLocalizedOperationResult.new
+
+      user, created = user_service.withEnableUserMutex do
```
arvindsv: Shouldn't this whole block really be in `UserService`?

ketan: As much as I'd love for it to be, the method `save` from the controller cannot be moved into java without writing about 100 lines of java code :(

arvindsv: Because of tristate? Anyway, I'm sure you've thought about it more than I have. My question was based on how a user is created, when called [from the admin page](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/controllers/admin/users_controller.rb#L29-L40).

ketan: Hmm, the `create` method on `UserService` is interesting, I think I should be able to use it.

patrick92100:
I did not find any details on how to create user in the API doc (https://api.go.cd/current/#users)? did I miss something?
Regards

ketan:
@patrick92100 users are created when they first login, therefore there is no way to create users.

arvindsv:
If you need to add them to roles, you can do it in the config. Even if they don't exist (not created), I believe.

patrick92100:
ok, I am a bit confusing, because I saw in the 15.3 change log "Add API endpoint to create a user." but in the 15.2, user was also created at the first login...

ketan:
Hmm... I'm mistaken, the user create endpoint actually exists. It was never documented, because we forgot to do it.

If you POST to the endpoint `go/api/users` with the user attributes in JSON, you should be OK. You will see the attributes in the update user api docs, you will need to submit an additional login name parameter.

I'll document the endpoint later tomorrow.

ketan:
@patrick92100 — the new api docs are published - https://api.go.cd/current/#create-a-user

PR #1544, opened 2015-10-06
Missed part of #1539.

ketan:


ketan:
Merging myself, since there's nothing interesting here.

PR #1549, opened 2015-10-09
Remove confusion when creating `ArtifactPlan` instances.

ketan:
Meat of the changes in `ArtifactPlan` and `TestArtifactPlan`. The rest
of it is supplimental changes to adapt to the change in their interfaces.

ArtifactPlan has a configurable `artifactType` instance. It was therefore
possible to create an instance of `ArtifactPlan` with `artifactType` set to
`ArtifactType.unit` - this was not only confusing, but incorrect since the
behavior of both artifact types is different.

This PR changes that by ensuring that the `artifactType` instance cannot
be set by clients that create an instance of `ArtifactPlan`.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/TestArtifactPlan.java
+++ b/config/config-api/src/com/thoughtworks/go/config/TestArtifactPlan.java
@@ -1,5 +1,5 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2015 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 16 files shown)


config/config-api/src/com/thoughtworks/go/config/TestArtifactPlan.java
```diff
@@ -48,6 +48,10 @@ public TestArtifactPlan(ArtifactPlan plan) {
         setDest(TEST_OUTPUT_FOLDER);
     }
 
+    public TestArtifactPlan(String src, String dest) {
+        super(ArtifactType.unit, src, dest);
+    }
+
     public void add(ArtifactPlan plan) {
         plans.add(plan);
```
jyotisingh: Why do we still need this?

PR #1550, opened 2015-10-09
Go on IE Intranet

sghill:
When the GoCD server is hosted internally, IE treats it as an intranet site by
default. Being an intranet site, it is automatically rendered in IE7
compatibility mode, which causes a few problems:

* pipelines dashboard doesn't line up as in Chrome and Firefox
* console log in job details page does not show line breaks, making it very
  difficult to read

By providing the X-UA-COMPATIBLE meta tag, IE shows the site in standards mode
and everything looks very similar to Chrome and Firefox.

### Before
pipelines-before
job-console-before

### After
pipelines-after
job-console-after

## Reproducing

I grabbed the Windows 7 IE 11 VM from https://dev.modern.ie/tools/vms/mac/ and
added go to the list of intranet sites through Settings -> Internet Options ->
Security Tab -> Highlight Local Intranet -> Sites -> Advanced -> Add

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/views/layouts/admin.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/layouts/admin.html.erb
@@ -11,7 +11,6 @@
 <!--<![endif]-->
 
 <head>
-    <meta http-equiv="X-UA-Compatible" content="IE=9">
     <%- @current_tab_name = "admin" -%>
     <%- @view_title = @view_title || "Administration" -%>
     <%= render :partial => "shared/head" %>
--- a/server/webapp/WEB-INF/rails.new/app/views/shared/_head.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/shared/_head.html.erb
@@ -1,3 +1,4 @@
+<meta http-equiv="X-UA-COMPATIBLE" content="IE=edge,chrome=1">
 <title>
     <%= @view_title %> - Go
 </title>
--- a/server/webapp/WEB-INF/vm/shared/_header.vm
+++ b/server/webapp/WEB-INF/vm/shared/_header.vm
@@ -14,8 +14,7 @@
  * limitations under the License.
  *************************GO-LICENSE-END***********************************#
 
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
-        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<!DOCTYPE html>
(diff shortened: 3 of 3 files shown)


ketan:
The change looks good, someone will need to verify if a few pages render OK in ie9+. Since that is what we support.

sghill:
I downloaded a few more of those http://modern.ie machines and took screenshots of 3 pages I tend to spend a lot of time on. Happy to provide other pages if that helps.

## Windows 7/IE9
ie9-admin
ie9-job-console
ie9-pipelines
## Windows 8/IE10
ie10-admin
ie10-job-console
ie10-pipelines

ketan:
@sghill - FYI we're unable to run the jasmine specs with IE11 as of now. I'm having to force IE9 compatibility mode to get tests to run. See #1564.

sghill:
@ketan - Is there a command I can run them with? The [docs][d] point to a jsunit task that isn't in the current codebase.

[d]: http://www.go.cd/documentation/developer/6/6.2.html#6-2-3-javascript-testing-using-jsunit-tests

ketan:
If you have a server running locally in dev mode. You can hit `http://localhost:8153/go/jasmine-specs` and `http://localhost:8153/go/jasmine-specs-new`. The new specs don't rely on the DOM very much so I don't think you'll have issues there.

ketan:
also - https://build.go.cd/go/tab/build/detail/build-windows/561/build-non-server-windows/3/jasmine_tests

sghill:
Interestingly the specs seem to run fine in IE10 mode as well.

For what it's worth, forcing IE9 or IE10 instead of Edge would also solve the problem, since intranet mode defaults to IE7 (where it looks funny). Happy to update the PR to force IE9 or IE10 instead if that's easier to pull in.

ketan:
It's very much possible that the specs run on IE10 — we've not really spent any time on modernising the tech stack (or testing if it works on modern browsers among other things modern)

Most of the team is out on vacation for the next few weeks so we can't really offer you an answer in terms of what's the pragmatic thing to do. Additionally given the amount of "big stuff" going on (performance of pipeline config save and a new single page app for config edits), we suspect the team won't have much cycles to spend on this one.

I'm all for modernising stuff, but it feels like this particular issue is in all likelihood going to be post 15.3.

zabil:
@sghill sorry about putting this on cold storage.
We'll try to get this merged next release, can you rebase to resolve the conflicts?

sghill:
@zabil No worries. Out of curiosity, has the team heard this compliant from other users? If it's just me, I'd be OK with closing it since there are lots of workarounds, it seems to impact very few users, and it requires a lot of manual testing before merging.

zabil:
Surprisingly, other users haven't complained yet.

PR #1551, opened 2015-10-10
Test deprecation cleanup

sghill:
This pull request looks quite big, but shouldn't change any behavior.

1. `BuildOutputLoggerTest` is the only file I've come across with CRLF line endings, so it was converted to unix-style in its own commit.
2. Converted the few tests extending `TestCase` to the annotation style present in the rest of the codebase
3. Replace the deprecated `junit.framework` asserts with preferred `org.junit` asserts
4. Replace the deprecated `JUnitClassRunner` with preferred `BlockJUnit4ClassRunner`
5. Replace the deprecated `org.junit.matchers.JUnitMatchers.containsString` with preferred `org.hamcrest.CoreMatchers.containsString`
6. Replace deprecated `new ArgumentCaptor` with preferred `ArgumentCaptor.forClass`


sghill:
Is it possible to get view permissions on the PR pipelines?

screen shot 2015-10-11 at 8 53 41 am

ketan:
user: view
password: password

I'm tweaking the pipeline configs to make things go a bit faster; so expect stuff to fail. That may not be your fault :)

sghill:
Thanks very much!

Looks like the builds that failed were because of a cancellation and `Could not find artifact com.thoughtworks.go:database:jar:15.3.0-101`.


Is it possible to trigger the builds again or should I add another commit to the branch?

arvindsv:
I can trigger it again, but will wait for @ketan to reply, since he knows the latest state of them better. If he's not around (or doesn't reply in a while), we can re-trigger it and see what happens.

ketan:
Go ahead and trigger the build @arvindsv

arvindsv:
https://build.go.cd/go/pipelines/build-linux-PR/103/build-non-server/2 is building.

ketan:
Thank you very much for the cleanup - @sghill

arvindsv:
Way to go, @sghill!! What you said [here](https://github.com/gocd/gocd/pull/1534#issuecomment-145925793) also makes a lot of sense. If you need help or are confused by something, let one of us know!

sghill:
Hey thanks for making it so easy to contribute @ketan and @arvindsv. Much appreciated!

PR #1566, opened 2015-10-16
Do not check for execjs runtime during server boot

zabil:
execjs runtime check is required only in development mode. 
This flag prevents the check on production mode or server boot.

### DIFF
--- a/server/webapp/WEB-INF/init.rb
+++ b/server/webapp/WEB-INF/init.rb
@@ -1,5 +1,4 @@
 # This file is sourced by jruby-rack and is used to perform initialization of the jruby environment
-
 # because jruby-rack does not respect GEM_HOME/GEM_PATH set in web.xml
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/init.rb
```diff
@@ -10,4 +9,6 @@
 end
 
 ENV['RAILS_ENV']      ||= (ENV['RACK_ENV'] || 'production')
+
+ENV['EXECJS_RUNTIME']='Disabled' if ENV['RAILS_ENV'] == 'production'
 ENV['BUNDLE_WITHOUT'] = 'development:test:assets' if ENV['RAILS_ENV'] == 'production'
```
ketan: Can we collapse the statements into a single `if` block?

zabil:
@ketan collapsed the if

Issue #1132, opened 2015-05-15
The entirety of JUnit XML files is skipped if they are minified


underyx:
I'm using py.test and its built-in JUnit XML generator to make reports of my tests. Go.cd always fails to display any of the contents, logging this:

```
13:16:44.907 Unable to publish test properties. Error was javax.xml.transform.TransformerException: com.sun.org.apache.xml.internal.utils.WrappedRuntimeException: The element type "all-results" must be terminated by the matching end-tag "</all-results>".
13:16:44.912 Could not publish property org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.
13:16:44.913 Could not publish property org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.
13:16:44.914 Could not publish property org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.
13:16:44.915 Could not publish property org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.
```

Now, bear with me here, cause I have zero knowledge of Java, but skimming through the source I think the issue is that the py.test JUnit files have no newlines in them, so [this line](https://github.com/gocd/gocd/blob/ed4536974c8e08d996ca3c8f8daf20d830696fad/config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java#L158) skips the complete file, with its 200 test cases. ~~If that's not the problem, please let me know and I'll edit the title of this issue.~~ _Edit:_ I tried reformatting the same XML file and it works now. This means that there's a workaround:
# Workaround
1. Install the `xmllib2-utils` package on your agents
2. Change your py.test config to output to `junit.min.xml` instead of `junit.xml`
3. Add the following step to the end of the pipeline:

``` xml
<exec command="/bin/bash">
  <arg>-c</arg>
  <arg>xmllint --format junit.min.xml > junit.xml</arg>
  <runif status="passed" />
</exec>
```

srinivasupadhya:
can you post sample XML file so someone can reproduce the issue?

underyx:
Sure, something like this:

```xml
[...]

PR #1570, opened 2015-10-17
Parsing minified JUnit XML

sghill:
Attempting to address #1132. Included a JUnit test that uses the file contents specified in a comment on that thread.

The first commit doesn't change any behavior, but converts JMock to Mockito. It looks like Mockito is used much more heavily throughout the codebase, so I thought now would be a good time to make things more consistent.

The second commit stops checking for `<?xml` in the first line, and instead removes the prolog, then checks the processed line to see if it has any content.


config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
```diff
@@ -154,8 +156,8 @@ private void pumpFileContent(File file, PrintStream out) throws IOException {
         BufferedReader bufferedReader = null;
         try {
             bufferedReader = new BufferedReader(new FileReader(file));
-            String line = bufferedReader.readLine();
-            if (!line.contains("<?xml")) { // skip prolog
+            String line = PROLOG_PATTERN.matcher(bufferedReader.readLine()).replaceFirst("");
+            if (!line.isEmpty()) { // skip if line was only prolog
```
ketan: I think your change looks good! Just one nitpick, perhaps it's my discomfort with not knowing if the Regex implementation in java are greedy in nature. Also I'm not sure what'd happen if the prolog string happened to be inside a `CDATA` section (remember `CDATA` does not do escaping)

Could we instead change this to something like (may not compile, but gives you a general idea) –

```java
// string that begins with the prolog, ignoring any whitespace
if (line.matches(/^\s*<\?xml\b/)){
  int prologStartIndex = line.indexOf("<?xml ");
  int prologEndIndex = line.indexOf("?>");
  // not sure if we need to handle a `-1` return value for any of the above?
  if (endIndex > prologStartIndex) {
    prologString = line.substring(prologStartIndex, prologEndIndex);
    line = line.replace(prologString, "");
  }
}
```

WDYT?

sghill: I agree, it looks like this could go in `XmlUtils` and be tested in isolation there as well.

sghill:
Hmm...This rebase went awry. Is it better to close and reopen?

config/config-api/src/com/thoughtworks/go/util/XmlUtils.java
```diff
@@ -95,4 +98,16 @@ public static boolean doesNotMatchUsingXsdRegex(Pattern pattern, String textToMa
     public static boolean matchUsingRegex(Pattern pattern, String textToMatch) {
         return pattern.matcher(textToMatch).matches();
     }
+
+    public static String stripProlog(String line) {
+        if (line.contains(PROLOG_START)) {
```
ketan: @sghill - Like I'd said in [my comment](https://github.com/gocd/gocd/pull/1570#discussion_r42316842) before -

Ideally what you'd like to look for is line that `startsWith` ignoring any leading whitespaces that lead upto the prolog.

This is needed because of files that possibly look like this (formatted for readability, but could be all on one line) would be parsed incorrectly with the implementation that you have. Perhaps it's just a matter of calling `strip` on the line before parsing it?

```xml
<testsuite>
  <testcase>
    <stdout>
      <![CDATA[....<?xml version="1.0" encoding="utf-8"?>...]]>
    </stdout>
  </testcase>
</testsuite>
```

sghill: @ketan - great point. I added this to the test cases and implemented your regex suggestions. Thanks for your patience!

ketan: > ~~~Thanks for your patience!~~~

No — on the contrary, thank YOU for your patience.

ketan:
@sghill –

PR looks good, I'm waiting for a green build to merge. We're in the midst of re-building all our agents so we're being a bit cautious.

ketan:
@sghill - there's one test failure that you may want to fix and squash -

```
Error	shouldGenerateReportForNUnitXmlWithByteOrderMark
Unit Test Failure and Error Details (1)
Test:	shouldGenerateReportForNUnitXmlWithByteOrderMark
Type:	Error
Message:	Wanted but not invoked: publisher.setProperty( Name=tests_total_count Value=18 ); -> at com.thoughtworks.go.domain.UnitTestReportGeneratorTest.shouldGenerateReportForNUnitXmlWithByteOrderMark(UnitTestReportGeneratorTest.java:87) However, there were other interactions with this mock: -> at com.thoughtworks.go.domain.UnitTestReportGenerator.generate(UnitTestReportGenerator.java:75) -> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126) -> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126) -> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126) -> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126) -> at com.thoughtworks.go.domain.UnitTestReportGenerator.generate(UnitTestReportGenerator.java:82)
Wanted but not invoked:
publisher.setProperty(
    Name=tests_total_count Value=18
);
-> at com.thoughtworks.go.domain.UnitTestReportGeneratorTest.shouldGenerateReportForNUnitXmlWithByteOrderMark(UnitTestReportGeneratorTest.java:87)

However, there were other interactions with this mock:
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.generate(UnitTestReportGenerator.java:75)
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126)
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126)
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126)
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.addProperty(UnitTestReportGenerator.java:126)
-> at com.thoughtworks.go.domain.UnitTestReportGenerator.generate(UnitTestReportGenerator.java:82)

	at com.thoughtworks.go.domain.UnitTestReportGeneratorTest.shouldGenerateReportForNUnitXmlWithByteOrderMark(UnitTestReportGeneratorTest.java:87)
```

sghill:
This is related to the byte order mark, which maybe explains why it was a `contains` check to begin with. It looks like there are at least a few different options that I'd like some feedback on.

1. Update the regex to `"^\\uFEFF?\\s*<\\?xml\\b.*$"`, which makes the test pass but is a little naive since there are different BOMs out there
2. Update the regex to `"^.?\\s*<\\?xml\\b.*$"`, which matches on much more than a BOM
3. Read the files in with commons-io's [BOMInputStream][1], which handles discarding the BOM on its own. I think I like this option, but it does change the way files are read from
```java
BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
```
to
```java
BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(new BOMInputStream(new FileInputStream(file))));
```

[1]: https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/input/BOMInputStream.html

ketan:
CC: @arvindsv and @jyotisingh

You're right about this, what makes me wonder next is if things blow up when different XML files have different BOMs in them.

Fundamentally, what this transformer does is to output a single [`index.html`](https://build.go.cd/go/files/regression/419/twist/1/firefox-split-runInstance-10/testoutput/index.html) that summarizes all the test failures and gets rendered on the [failures tab](https://build.go.cd/go/tab/build/detail/regression/419/twist/1/firefox-split-runInstance-10#tab-failures) of a job like this -

screen shot 2015-11-04 at 10 24 52 am

My question — given that we know what the expected output should be (an index.html that renders the test counts and errors/failures, if any) – Can we fundamentally change how this class works, as long as the end result is the same (or better). See ant-junit's report format here - https://www.google.com/search?q=junit+xml+report+html&source=lnms&tbm=isch&sa=X as an inspiration

ketan:
Closing in favor of #1685

Issue #1522, opened 2015-09-25
Add Environment to the parameters of the Update Agent API


rmarcinik:
Would it be possible to add the environment value to the API for updating an agent?  It is currently possible [with this method](http://www.go.cd/documentation/user/current/advanced_usage/agent_auto_register.html) but it is not totally reliable.  Sometimes the agent is still in a pending state afterwards, and has no resources or environment.  It would also be beneficial towards automation to control agents fully without the UI.
Thank you.

arvindsv:
Auto register should be reliable. I haven't heard of it not working, unless the agent has the same GUID as another, or something of that sort. The issue with doing what you're talking about is that unlike any other part of the update agent API call, this one will need to send that information all the way to the agent. Also, it *might* need a restart of the agent, etc.

Not something that cannot be done, but it's a bit of work. If you're thinking about making such a change, let me know and I can help. But, in case you're looking for someone from the core team to work on it, it's probably not going to happen soon.

arvindsv:
Actually, please ignore everything I said. I read what you wrote as "environment variable".

ketan:
@rmarcinik - I'm assuming that you mean to say that you want to update add/remove agents from environments in go.

Ideally this api call should belong to environments, and not on an agent. That said, if you're willing to build the API, someone on the team will be very happy to work with you.

rmarcinik:
Thanks for responding, I agree that updating the Environments would be a separate endpoint.  However I dont think there is an endpoint for environments yet, and I am not familiar with what it would take to develop.  
The autoregister method works for initializing agents but it depends on having access to the agents local filesystem, and is less than ideal for modifying the properties of an already enabled agent (or many agents).  I forgot to add, this is for windows agents; so it may work better on linux.

The goal in the end would be to:
* modify agent properties, without access to the local filesystem 
* modify agent properties, without going through the ui

If I had an idea of what this would take to accomplish I could make a proposal to include this in one of our sprints.

Thanks again for your time.

ketan:
On second thought –

I feel that my previous statement a result of my understanding of the structure (less than ideal) `cruise-config.xml` and how `<agent/>s` belong to `<environment/>s` in there.

If I were to think of it in a slightly different manner — in that agents are associated with an environment (and that is really what I feel best describes the relationship between the two, irrespective of how the config file makes it look like).

I'm in favor of allowing the `environments` attribute on an agent to be modified by the agents API – any takers?

arvindsv:
[...]

PR #1573, opened 2015-10-20
Allow agents API to update resources that an agent is associated with (#1522)

ketan:
v2 of the Agents API now accepts `environments` param, and the behavior
is identical to the behavior of the `resources` param.

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/GoConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/GoConfigService.java
@@ -1,5 +1,5 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2015 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -12,10 +12,11 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.server.service;
 
+import com.google.caja.util.Sets;
 import com.rits.cloning.Cloner;
 import com.thoughtworks.go.config.*;
 import com.thoughtworks.go.config.exceptions.*;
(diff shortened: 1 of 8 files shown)


server/src/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -581,17 +585,33 @@ public void updateAgentAttributes(String uuid, String userName, String hostname,
...
         if (hostname != null) {
             command.addCommand(new GoConfigDao.UpdateAgentHostname(uuid, hostname, userName));
         }
+
         if (resources != null) {
             command.addCommand(new GoConfigDao.UpdateResourcesCommand(uuid, new Resources(resources)));
         }
 
+        if (environments != null) {
+            Set<String> existingEnvironments = cruiseConfig().getEnvironments().environmentsForAgent(uuid);
```
arvindsv: I think this should be a method, by itself, especially if we're thinking of an environments API.

server/src/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -581,17 +585,33 @@ public void updateAgentAttributes(String uuid, String userName, String hostname,
...
+
+            Set<String> environmentsToRemove = Sets.difference(existingEnvironments, newEnvironments);
+            Set<String> environmentsToAdd = Sets.difference(newEnvironments, existingEnvironments);
+
+            for (String environmentToRemove : environmentsToRemove) {
+                command.addCommand(new GoConfigDao.ModifyEnvironmentCommand(uuid, environmentToRemove, TriStateSelection.Action.remove));
+            }
+
+            for (String environmentToAdd : environmentsToAdd) {
+                command.addCommand(new GoConfigDao.ModifyEnvironmentCommand(uuid, environmentToAdd, TriStateSelection.Action.add));
```
arvindsv: This can fail and throw an exception, if the environment doesn't exist. I'm not sure that will be handled gracefully. I don't see any test for that.

ketan: If the environment does not exist `GoConfigDao::ModifyEnvironmentCommand` will throw a `NoSuchEnvironmentException` exception via `EnvironmentsConfig#named`. This is already tested well.

This exception will be caught in `AgentService#updateAgentAttributes` along with the relevant message. As of now the server responds with a `500`, but that is something I'll change to respond with a `422` instead. Here's the response as of now -

```shell
$ curl -X PATCH \
    -H "Accept: application/vnd.go.cd.v2+json" \
    -H "Content-Type: application/json" \
    -d '{
      "environments": ["foo"]
    }' \
    'http://localhost:8153/go/api/agents/10c357bf-f7ae-42ff-a61f-fe3a8575a9a7'

{
  "message": "Updating agents failed:  { Environment [foo] does not exist. }"
}
```

ketan: I've addressed the issue about status codes and updated the PR. Specifying an environment that does not exist returns a `422` instead of a `500`.

arvindsv:
Looks good to me now. Someone can accept this, or I can when the build is green.

ketan:
@arvindsv the PR build is green, I'm not sure if you're referring to that build or the end-to-end acceptance tests going green.

arvindsv:
I meant the full set, not the PR build. It's nearly there (except upload installers). So, I'll accept this.

PR #1577, opened 2015-10-22
added recursive option to fetch submodules-status recursively

saschaglo:
I think it's necessary to fetch the status of submodules recursively because otherwise nested submodules won't be checked out correctly.


arvindsv:
Hi @saschaglo. That method is used to just print the status of submodules. [Checkout](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java#L133-L136) and [sync](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java#L375-L383) do happen recursively.

Even if this change needed to get into the codebase, you'd need to update the corresponding tests.

arvindsv:
Haven't heard anything. Closing.

Issue #1567, opened 2015-10-16
Integrate periodic garbage collection for config.git 


lenucksi:
In case of a larger Go server one can incur extreme growth of config.git.
This is due to the fact that all changes lead to commits to config.git which in turn lead lead to loose objects in the git repository that are not compacted or garbage collected automatically.
This leads to high commit or checkout times and also to extreme delay or dysfunction in the "view difference" functionality with the configuration change notices.
More functionality might be impacted, however I cannot provide exact or reproducible data on them.

A shrinkage of two orders of magnitude was observed with a Go server that ran for about two years. 
Here are a few details on the server and its configuration:
* Pipelines: >500
* Lines in config.xml: >80k
* Size of config.xml: >3.5MB
* Size of the config.git **prior to git gc**: 33GB
* Size of the config.git **after git gc**: 336MB
* Time to view a config diff in the web interface **prior to git gc**: Exceeding two minutes, mostly failing with HTTP 50x
* Time to view a config diff in the web interface **prior to git gc**: about 4 seconds.

* git count-objects -v -H **prior to git gc**
 * count: 541298
 * size: 32.44 GiB
 * in-pack: 0
 * packs: 0
 * size-pack: 0 bytes
 * prune-packable: 0
 * garbage: 0
 * size-garbage: 0 bytes

* git count-objects -v -H **after git gc**
 * count: 222
 * size: 11.80 MiB
 * in-pack: 516835
 * packs: 1
 * size-pack: 324.73 MiB
 * prune-packable: 0
 * garbage: 0
 * size-garbage: 0 bytes

Measurements yield that config.git increases by about 52MB loose objects per day. 

I would therefore highly appreciate if Go would gain an automatic garbage collection on the config.git repository.

lenucksi:
The mentioned GC example took about 35 minutes on a 36 core machine with 60GB of RAM.

rajiesh:
verified it on 15.3.0-2724(0deb63cff58810076760a17bba4e6d855b12aa5f) with auto gc on and off and modifying the gc check intervals and thresholds and its working as expected

PR #1600, opened 2015-11-06
Added support for periodic GC on config.git repo

jyotisingh:
* jgit upgraded to 4.1
* periodic GC that triggers at 1am, can be overridden using 
``` properties
-Dgo.config.gc.cron=<cron expression>
```
* GC happens in aggressive mode by default. Can be turned off using 
``` properties
-Dgo.config.repo.gc.aggressive=false
```
* using SLF4j logger instead


config/config-server/src/com/thoughtworks/go/service/ConfigRepository.java
```diff
@@ -390,8 +390,23 @@ void cleanAndResetToMaster() throws IOException {
...
         }
     }
+
+    public void garbageCollect() throws Exception {
+        doLocked(new VoidThrowingFn<Exception>() {
+            public void run() throws Exception {
+                try {
+                    LOGGER.info("Before GC: {}", git.gc().getStatistics());
+                    git.gc().setAggressive(systemEnvironment.get(SystemEnvironment.GO_CONFIG_REPO_GC_AGGRESSIVE)).call();
+                    LOGGER.info("After GC: {}", git.gc().getStatistics());
```
ketan: `Before GC` and `After GC` could be confusing (heap GC?). Can we clarify this a bit?

jyotisingh: The log file entry would look something like this:
```
2015-11-05 19:31:00,029  INFO [pool-1-thread-1] ConfigRepository:403 - Before running GC: {sizeOfPackedObjects=0, sizeOfLooseObjects=6640456, numberOfPackedObjects=0, numberOfPackFiles=0, numberOfPackedRefs=0, numberOfLooseRefs=2, numberOfLooseObjects=2037}
2015-11-05 19:31:02,199  INFO [pool-1-thread-1] ConfigRepository:405 - After running GC: {sizeOfPackedObjects=415262, sizeOfLooseObjects=0, numberOfPackedObjects=2037, numberOfPackFiles=2, numberOfPackedRefs=1, numberOfLooseRefs=1, numberOfLooseObjects=0}
```
That calls out the class name `ConfigRepository`, so I guess that should be ok.

ketan:
@jyotisingh on second thought can the properties be —

````
-Dgo.config.repo.gc.cron=<cron expression>
-Dgo.config.repo.gc.aggressive=false
```

lenucksi:
That probably also tackles #1567.

jyotisingh:
@ketan - #1601

PR #1616, opened 2015-11-16
update some dependencies that seemed low risk to update

ketan:


config/config-api/pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>uk.com.robust-it</groupId>
             <artifactId>cloning</artifactId>
-            <version>1.7.3</version>
+            <version>1.9.2</version>
```
srinivasupadhya: you might want to double check. but i have seen clone failures for newer releases of this lib.

ketan: it was the immediate next version that broke things.

Issue #1576, opened 2015-10-21
scheduling pipelines at a given revision of a material fails on phabricator-type staging repos


drmuey:
Some development systems have the idea of a staging repo that pushes local changes to build/test as a single tagged SHA[1]. The problem is that Go can’t see those SHAs when scheduling a specific one.

For example, given this:

```
$ arc diff
…
* [new tag]         8d09dc41096c9e5a26a609cb3d0a064f7d5f4ead -> phabricator/diff/72576
STAGING PUSHED  Pushed a copy of the changes to tag "phabricator/diff/72576" in the staging area.
Updated an existing Differential revision:
…
```

Now I'd like to start a build of 8d09dc41096c9e5a26a609cb3d0a064f7d5f4ead but when you:

`POST http:/…/go/api/pipelines/phabricator-build/schedule?materials[test_staging]=8d09dc41096c9e5a26a609cb3d0a064f7d5f4ead`

you get a 202 w/ “[Request to schedule pipeline](https://github.com/gocd/gocd/blob/ad8fc27d552f038fc8609c27cd0d6e4eadd90739/server/src/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L300) phabricator-build accepted” but it never starts. Checking the log you see:

```
2015-10-20 12:24:37,564 ERROR [60@MessageListener for MaterialUpdateStatusNotifier] BuildCauseProducerService:244 - Error while scheduling pipeline: phabricator-build
java.lang.RuntimeException: Unable to find revision [8d09dc41096c9e5a26a609cb3d0a064f7d5f4ead] for material [GitMaterial{url=http://….git, branch='master', submoduleFolder='null'}]
        at com.thoughtworks.go.server.materials.MaterialChecker.findSpecificRevision(MaterialChecker.java:52)
```

~~If it did `git fetch --tags` first it’d have 8d09dc41096c9e5a26a609cb3d0a064f7d5f4ead~~ **Update**:

[...]

PR #1619, opened 2015-11-17
Issue #1576: Add --tags to to git pull

drmuey:
@ketan here is the pull request you asked for thanks! I am in the process of building the installer to test (docker/VM issues … blarg)


drmuey:
Work fine but does not solve problem:

https://github.com/gocd/gocd/issues/1576#issuecomment-164443889

Issue #1593, opened 2015-11-04
Failure to upload artifacts when Go server runs out of disk space is not reported to the user


jyotisingh:
If server runs out of disk-space, upload artifact from a given job is reported as success but the upload does not really happen. The logs entries are created but there is no visual indication on the dashboard or build that the upload really failed.

PR #1622, opened 2015-11-19
Fix - Failure to upload artifact upload is not reported when go runs out of disk space bug(#1593)

juhijariwala:


server/src/com/thoughtworks/go/server/service/ArtifactsService.java
```diff
@@ -126,6 +133,19 @@ public boolean saveOrAppendFile(File dest, InputStream stream) {
         }
     }
```
jyotisingh: I see very similar code in [ArtifactsDiskCleaner#deleteOldArtifacts](https://github.com/gocd/gocd/blob/5422c89361c109eb29e6d38088e8c8f0125e285c/server/src/com/thoughtworks/go/server/service/ArtifactsDiskCleaner.java#L62). Can you try to reuse it.

jyotisingh: Also, I don't see a test covering check for diskSpaceMonitor.isLowOnDisk(), could you add that

juhijariwala: Since [ArtifactDiskCleaner#deleteOldArtifacts](https://github.com/gocd/gocd/blob/5422c89361c109eb29e6d38088e8c8f0125e285c/server/src/com/thoughtworks/go/server/service/ArtifactsDiskCleaner.java#L62) has different logic, I can't reuse it. I have added test which covers the ```checkDiskSpace```. 
The ```diskSpaceMonitor.isLowOnDisk()``` method is being tested in ```GoDiskSpcaeMonitorTest```.

ketan:
Closing since there does not seem to be any interest or a bullet-proof-way to ensure that multiple threads uploading artefacts will not see this problem.

Fixing this would basically require that each agent notifies the server about the disk space usage before it uploads artefacts, the server will then "reserve" that storage and wait for the upload to complete.

There are of-course other issues like agents and servers using different block sizes so the size that the agent reports may actually be different than the amount of space the artefacts will consume on the server, which makes this reservation idea not-so-bullet-proof.

Issue #1628, opened 2015-11-20
Upgrade commons-collections to library version 3.2.2 


zabil:
Fix the [serialization](http://www.go.cd/2015/11/09/deserialization-vulnerability-commons-collections.html) vulnerability.

version 3.2.2 of commons-collection has patched [this](https://issues.apache.org/jira/browse/COLLECTIONS-580)

As a part if this change we need to set up a pipeline on build.go.cd to check that older versions are not pulled down as a transitive dependency in the future.

PR #1629, opened 2015-11-20
#1628 Upgrade commons-collection library to 3.2.2

zabil:


ketan:
```
ketanpadegaonkar@ketanpkr  ~gocd pr/1629 $ mvn dependency:tree | grep commons-collections | grep -v 3.2.2
[INFO] |  |  |  +- commons-collections:commons-collections:jar:3.2.1:compile
```

Looks like it's pulling in 3.2.1 from somewhere, and my maven foo is not strong enough to understand where from.

zabil:
You need to run the verbose mode to get more details of which version is omitted or picked.
https://maven.apache.org/plugins/maven-dependency-plugin/examples/resolving-conflicts-using-the-dependency-tree.html

3.2.1 does strangely show up as dependency on the "util" project. But there's an explicit include for 3.2.2 in the pom.

But then again, if you do a ```.bn clean cruise:prepare dist``` and the following
```
sh-3.2$ jar -xvf go.jar 
sh-3.2$ jar -tvf defaultFiles/cruise.war | grep commons-collections
-rwxrwxrwx  0 0      0      588337 Nov 20 12:01 WEB-INF/lib/commons-collections-3.2.2.jar
```
there's only one version of the jar plus the running server plugs the exploit (if you post the exploit)

PR #1637, opened 2015-11-24
Gc support on config git repo

jyotisingh:


ketan:
Are you sure we need aggressive GC kicking in every day?

Can we instead reduce the frequency of aggressive GC to be weekly instead?

Issue #1605, opened 2015-11-09
Auto-complete for resources


ketan:
Part of #1404

varshavaradarajan:
Verified that SPA provides auto-complete for resources.

PR #1647, opened 2015-11-25
More enhancements to the single page app (#1404) —

ketan:
* Add autocomplete of resources (#1605)
* Add support for stage approvals and authorization
* Add some placeholder text where tooltips are not needed
  (environment variables, for e.g.)


server/src/com/thoughtworks/go/server/service/UserService.java
```diff
@@ -203,6 +203,10 @@ public void modifyRolesAndUserAdminPrivileges(final List<String> users, final Tr
         return roles;
     }
 
+    public Collection<String> allRoleNames() {
+        return allRoleNames(goConfigService.cruiseConfig());
+    }
+
     public Collection<Role> allRoles(CruiseConfig cruiseConfig) {
```
zabil: This method is only used in specs and the scope change it introduces is used only is specs is there a way to avoid that?

server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/string-plus.js
```diff
@@ -15,12 +15,14 @@
...
 define(['string', 'lodash'], function (s, _) {
+  var POSITIVE_INTEGER = /^\d+$/;
+
   var mixins = {
     defaultToIfBlank: function (value, defaultValue) {
       return s.isBlank(value) ? defaultValue : value;
     },
 
-    overrideToJSON: function (prop) {
+    collectionToJSON: function (prop) {
```
zabil: This method is an odd ball in string-plus. We should get it out of here.

server/src/com/thoughtworks/go/server/service/UserService.java
ketan: Not correct, this is used in the controller to list all roles for autocomplete in the authorizations form fields

server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/string-plus.js
ketan: Yes, I have it on my todo list. We probably need a generic "utils" object somewhere. I'll do it in a later commit, if that's OK?

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/application.js
```diff
@@ -14,7 +14,13 @@
  * limitations under the License.
  */
 
-requirejs(['jquery', 'mithril', 'pipeline_configs/models/pluggable_tasks', 'pipeline_configs/views/pipeline_config_widget', 'foundation.topbar'], function ($, m, PluggableTasks, PipelineConfigWidget) {
+requirejs([
+  'jquery', 'mithril',
+  'pipeline_configs/models/pluggable_tasks', 'pipeline_configs/models/resources', 'pipeline_configs/models/users', 'pipeline_configs/models/roles',
+  'pipeline_configs/views/pipeline_config_widget', 'foundation.topbar'
+], function ($, m,
```
zabil: Is "foundation.topbar" initialized this way? It's not be passed as an argument,

ketan: requiring `foundation.topbar` registers itself in the jquery foundation namespace (somewhere) so it can be initialized via `$(document).foundation()`

Issue #1614, opened 2015-11-13
Fetch latest release details from client(browser)


maheshp:
As part of #1308 the server at a given cadence checks if there is a new version of GO available. There might be instances where the server would not be able to make calls to fetch the latest release details from go.cd as it might be behind a firewall. This leaves with the client(browser) making a call to go.cd to fetch the latest release and update the server.

Acceptance Criteria
  1. The latest release details should be fetched with a cadence of 1 day.
  2. Clients(Browser) should be used to fetch the latest release if not updated by server.
  3. Since browser would be used as a proxy, ensure only one request is made to fetch the latest release.
  4. This feature should be turned off on development Server and non production distributions(Developer boxes)
  5. Users should have an option to disable this feature.
  6. There should be visual cue on webpages to highlight a new GO version.
  7. Verify messages from the update server are not tampered.
  8. On failure of update, retry every half an hour.

arvindsv:
Related PR is: #1651

rajiesh:
verified it on 15.3.0-2757 (9a2bdcf7063d685239ccf7f9eeb63850bd007658) and update check and update notification works as expected

Issue #1308, opened 2015-07-07
Implement a plugin and Go update check


arvindsv:
Need to check at regular intervals, and notify users (or admins only) about newer releases of Go which are available.

Also, and maybe more importantly, do this for plugins. Plugins can and should be updated more frequently and at a cadence which is different from Go's own. This check should include version incompatibility checks, etc. This can be tricky, because, as it stands now, go.cd site does not have plugin files. They're all available in external repositories of the authors. Doing this right can eventually make a "Plugin Repository" to install plugins from within Go, easier.

jyotisingh:
@kmugrage 's comment from From #1386 :

We hear on a regular basis that people do not know when a new version of Go is released. Go should let people know if there's a new version available so that they can download and install it. The notification should be fairly visible, but not overwhelming.

I think the first version of this is a simple notification.

norcnorc:
Please do make this "phone home" feature opt-in. Not all users may be excited to have even the information that they are using Go broadcast into the world. Also, this can be a slippery slope leading to increasing amounts of "usage statistics" gathered and transmitted by ever more insidious means, such as https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/UsageStatistics/footer.jelly#L26-L28

maheshp:
Tasks 
    1. Add a timer in GO server with a cadence of 1 day to fetch the latest GO version, this data should be available through a go.cd endpoint
        and persist the latest version in the DB.
    2. Add a visual clue on the web page to indicate if a new version is available.

arvindsv:
Closing as the update check is fine. The plugin update check can be made a new issue when it needs to be done (maybe along with [plugin central](https://github.com/gocd/gocd/issues/156)).

PR #1651, opened 2015-11-26
Go Update Check - #1614

maheshp:
- GO update check is done periodically using the browser as proxy
- Users are notified if a new update is available
- Users should be able to turn off this feature if required by setting the flag "go.check.updates" to false


server/test/unit/com/thoughtworks/go/domain/GOVersionTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test(expected = VersionFormatException.class)
+    public void shouldErrorOutIfStringVersionIsNotInCorrectFormat(){
+        new GOVersion("12.abc.1");
+    }
+
+    @Test
+    public void shouldBeAbleToCompareVersions(){
+        assertThat(new GOVersion("15.1.0-123").compareTo(new GOVersion("15.1.0-123")), is(0));
```
ketan: Did you try this out when a distribution is built. The version numbers look like `15.3.0(2597-GIT_SHA) `

server/webapp/WEB-INF/rails.new/app/assets/javascripts/version_updater.js
```diff
@@ -0,0 +1,87 @@
...
+    },
+
+    canUpdateVersion: function(){
+        var versionCheckInfo = localStorage.getItem('versionCheckInfo');
+        if(jQuery.isEmptyObject(versionCheckInfo)){
+            return true;
+        }
+        lastUpdateAt = new Date(versionCheckInfo.last_updated_at);
+        var halfAnHourBack = new Date(Date.now() - 30*60*1000)
+        return halfAnHourBack > lastUpdateAt;
```
ketan: `halfHourAgo`

server/webapp/WEB-INF/rails.new/app/assets/javascripts/version_updater.js
```diff
@@ -0,0 +1,87 @@
...
+    this.staleVersionInfoUrl = staleVersionInfoUrl;
+    this.updateServerVersionInfoUrl = updateServerVersionInfoUrl;
+}
+
+jQuery.extend(VersionUpdater.prototype, {
+    update: function () {
+        if(!this.canUpdateVersion()){
+            return;
+        }
+        this.fetchStaleVersionInfo();
```
ketan: Invert this?

```javascript
if (this.canUpdateVersion()){
  this.fetchStaleVersionInfo();
}
```

Also please use 2 spaces :)

server/webapp/WEB-INF/rails.new/app/assets/javascripts/version_updater.js
```diff
@@ -0,0 +1,87 @@
...
+    },
+
+    fetchStaleVersionInfo: function(){
+        _this = this;
+        jQuery.ajax({
```
ketan: Referring to `jQuery` is a PITA. Using a javascript self invoking function makes this a bit easier –

```javascript
(function ($) {
  ...
  $.ajax()
  ...
})(jQuery);
```

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/version_infos_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  class VersionInfosController < ApiV1::BaseController
+    def update_server
+      result = HttpLocalizedOperationResult.new
+
+      render :nothing => true, :status => 400 and return if message_tampered?
```
ketan: There's a `render_message` helper, I'd suggest sending a message back with a status code. The preferred way to write this is to —

```ruby
return ... if message_tampered?
```

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/version_infos_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      message = JSON.parse(params['message'])
+      version_info = version_info_service.updateServerLatestVersion(message['latest-version'], result)
+
+      if result.httpCode == 200
```
ketan: use `result.isSuccessful`

ketan: Also, I've been thinking, we've used this pattern in a few places. Does this call for a prettier way to write this (just thinking out aloud). This need not be done right now.

```ruby
return if_success(result) do
  render :json_hal_v1 => to_json_hal(version_info)
end


def if_success(result, &blk)
  if result.isSuccessful?
    yield
    true
  else
    render_http_operation_result(result)
    false
  end
end
```

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/version_info_representer.rb
```diff
@@ -0,0 +1,50 @@
...
+module ApiV1
+  class VersionInfoRepresenter < ApiV1::BaseRepresenter
+    def initialize(version_info, system_environment)
+      @version_info = version_info
+      @system_environment = system_environment
+      super(@version_info)
+    end
+
+    link :doc do
+      'http://api.go.cd/#versionInfo'
```
ketan: @arvindsv — are we going to call this "API"?

We do have a special route for internal API if that is what we're going to call this.

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/version_infos_controller_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+require 'spec_helper'
+
+describe ApiV1::VersionInfosController do
+  describe :update_server do
+    before(:each) do
+      @version_info_service = double('version_info_service')
+      @system_environment = double('system_environment')
+
+      @public_key_path = "update_server_public_key_path"
+      @message = "{\n  \"latest-version\": \"15.3.0-123\",\n  \"release-time\": \"2015-07-13 17:52:28 UTC\"\n}"
```
ketan: If you use `%Q{}` you'd not need to use the ugly escaping :)

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/version_infos_controller_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      @signature = "ajdakjdhakjd"
+      installed_version = com.thoughtworks.go.domain.GOVersion.new("1.2.3-1")
```
ketan: See `java_imports.rb` to get rid of the `com.tw.go.domain` namespace everywhere

server/webapp/WEB-INF/rails.new/spec/javascripts/version_updater_spec.js
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+
+describe("VersionUpdater", function(){
+    beforeEach(function () {
+        localStorage.clear();
+    });
+
+    describe("update", function(){
+        it("should_update_the_latest_version", function(){
```
ketan: Using underscores is the old way of writing tests (these were migrated by @jyotisingh, AFAIK). Please use sentences.

server/webapp/WEB-INF/rails.new/spec/javascripts/version_updater_spec.js
```diff
@@ -0,0 +1,107 @@
...
+            });
+
+            versionUpdater.update();
+
+            expect(jQuery.ajax).toHaveBeenCalled();
+            expect(versionUpdater.fetchLatestVersion).toHaveBeenCalledWith(versionInfo);
+       });
+
+        it("should_not_update_the_latest_version_if_updated_in_last_half_an_hour", function(){
+            spyOn(localStorage, 'getItem').and.callFake(function(key){
```
ketan: Is there a need to spy on this when you can actually test what's in the localstorage?

server/webapp/WEB-INF/rails.new/lib/message_verifier.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+class MessageVerifier
+  def self.verify(message, signature, public_key_path)
```
ketan: Can we add another helper method that accepts a public key string, to make it easier for testing? I see you're having to stub file operations.

server/webapp/WEB-INF/rails.new/spec/lib/message_verifier_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      public_key = 'update_server_public_key'
+      decoded_public_key = 'decoded_public_key'
+      digest = double('OpenSSL::Digest::SHA512')
+      pkey = double('OpenSSL::PKey::RSA')
+
+      File.should_receive(:read).with(public_key_path).and_return(public_key)
+      OpenSSL::Digest::SHA512.should_receive(:new).and_return(digest)
+      Base64.should_receive(:decode64).with(public_key).and_return(decoded_public_key)
+      OpenSSL::PKey::RSA.should_receive(:new).with(decoded_public_key).and_return(pkey)
+      pkey.should_receive(:verify).with(digest, signature, message)
```
ketan: I don't see a reason to actually test what's being invoked, we already have a public key, we can test actual verification with a good and bad message/digest.

I feel that's a much "safer" test in case things break.

server/src/com/thoughtworks/go/server/dao/VersionInfoSqlMapDao.java
```diff
@@ -0,0 +1,75 @@
...
+                sessionFactory.getCurrentSession().saveOrUpdate(versionInfo);
+            }
+        });
+    }
+
+    @Override
+    public VersionInfo findByComponentName(final String name) {
+        return (VersionInfo) transactionTemplate.execute(new TransactionCallback() {
+            @Override
+            public Object doInTransaction(TransactionStatus transactionStatus) {
```
jyotisingh: Currently you fetch/save the data in dao and then let `ServerVersionInfoManager` manage/update the cache. Would suggest leaving all cache related business to the dao. Let ServerVersionInfoManager make calls to the dao to get the data, and leave it up to the dao to decide if it should fetch it from cache or make a trip to db. Ditto for saveOrUpdate.

server/test/integration/com/thoughtworks/go/server/dao/VersionInfoSqlMapDaoIntegrationTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Autowired
+    private VersionInfoSqlMapDao versionInfoSqlMapDao;
+
+    @Autowired
+    private DatabaseAccessHelper dbHelper;
+
+    @Before
+    public void setup() throws Exception {
+        dbHelper.onSetUp();
+        versionInfoSqlMapDao.deleteAll();
```
jyotisingh: This is not required as `dbHelper.onSetUp()` and `dbHelper.onTearDown()` should already take care of rolling back all data inserted as part of the test. Could you check.

server/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -148,6 +148,14 @@
         <set name="rails_bound">true</set>
     </rule>
 
+
+    <rule>
```
jyotisingh: can you club it with the `Rails API` rule

```xml
    <rule>
        <name>Rails API</name>
        <from>^/(api/(?!admin/config).*)|(api/version_infos/.*)$</from>
        <to last="true">/rails/$1</to>
        <set name="rails_bound">true</set>
    </rule>

```

server/src/com/thoughtworks/go/server/service/ServerVersionInfoManager.java
```diff
@@ -0,0 +1,122 @@
...
+    public void initialize() {
+        this.serverVersionInfo = builder.getServerVersionInfo();
+
+        addGoUpdateToCacheIfAvailable();
+    }
+
+    public VersionInfo versionInfoForUpdate() {
+        if (isDevelopmentServer() || isVersionInfoUpdatedToday() || isUpdateInProgress()) return null;
+
+        versionInfoUpdatingFrom = clock.currentDateTime();
```
jyotisingh: this method is not thread safe

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/version_info_representer.rb
arvindsv: Is this something provided by Go? A representation of the current version the server is on? If so, it's possible to make this part of the "APIs". I'd keep this unauthenticated though.

arvindsv:
I'd call it "GoVersion", not "GOVersion".

What is the plan for when we want to update the key, in case it is compromised, etc?

arvindsv:
Usually, there will be a very securely stored signing-key which signs other public keys. It is assumed to be never compromised. CAs do this.

ketan:
> What is the plan for when we want to update the key, in case it is compromised, etc?

There's a `SystemEnvironment` for that.

arvindsv:
In case you haven't seen videos like [this one](https://www.youtube.com/watch?v=b9j-sfP9GUU), I find it very cool.

arvindsv:
The problem with a SystemEnvironment is that it needs a release for a change.

ketan:
> The problem with a SystemEnvironment is that it needs a release for a change.

Not quite, the variable points to a file that can be supplied externally. Am I missing something?

arvindsv:
My point is: Imagine on Jan 1, 2016 the private key used by updates.go.cd to sign the messages gets compromised. Now, we update the key. All Go Servers trying to check for updates will fail. Till someone *manually* (?) goes and updates the key or we roll the new public key as a part of the release. Right?

arvindsv:
Instead, we should have a certification hierarchy. Not much should change in the way it is used, but in the way the keys are generated and verified. I'll find some documentation about what I'm talking about, instead of writing it all here. If I can't find something, I will. Essentially, it's a take on the process CAs use with master keys and the actual keys shared with the public.

arvindsv:
This seems reasonable:

https://technet.microsoft.com/en-us/library/cc740209(WS.10).aspx
https://technet.microsoft.com/en-us/library/cc738786(v=ws.10).aspx

I'm sure there are other (better) guides, but the ideas are here.

arvindsv:
If we can split the master key properly away from the keyring and store it securely, and then use only subordinate signing keys in GPG, then it should be ok too. Otherwise, we should create a proper master key and store it securely and use it only to sign another key (the signing key) which can change. We distribute the public key portion of the master key along with the server and it *never* changes.

As a part of the message, we send the public part of the signing key along with its signature (signed using the master key) along with the real message. Verification goes like this:

1. Verify that the signing key has been signed by the master key (for which the server has a public key).

2. Verify the real message was signed using the signing key.

Revocation goes like this (assumption is that the signing key is compromised):

1. Revoke the signing key.

2. Generate a new signing key and sign it using the offline master key.

3. Start generating messages using the new signing key.

Everything should continue working, with no change from the server side.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/version_updater.js
```diff
@@ -0,0 +1,83 @@
...
+
+var VersionUpdater = function (staleVersionInfoUrl, updateServerVersionInfoUrl) {
+    this.staleVersionInfoUrl = staleVersionInfoUrl;
+    this.updateServerVersionInfoUrl = updateServerVersionInfoUrl;
+};
+
+(function ($) {
+    $.extend(VersionUpdater.prototype, {
+        update: function () {
+            if (this.canUpdateVersion())
```
ketan: @maheshp - can you use 2 spaces for javascript?

We should be using 4 spaces only for java code. Also, single if statements are expected to have a paren —

```javascript
if (foo) {
  bar()
}
```

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css_sass/components/notifications.scss
```diff
@@ -0,0 +1,58 @@
...
+  margin:7px 10px 0;
+  position: relative;
+
+.hover-container{
+  display:none;
+  position: absolute;
+  right:-7px;
+  top:100%;
+  padding-top: 5px;
+}
```
ketan: This file does not look like it is indented properly?

server/webapp/WEB-INF/rails.new/spec/javascripts/version_updater_spec.js
```diff
@@ -0,0 +1,111 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("VersionUpdater", function(){
+    beforeEach(function () {
+        localStorage.clear();
+    });
+
+    describe("update", function(){
```
ketan: 2 spaces?

server/webapp/WEB-INF/rails.new/spec/lib/message_verifier_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+                      creq3UW+rFeglirU6LdylDQb6o0v74OPLtFaOzrIed2xynqJEDeq0FnhATLjKCHU
+                      L8pOClq8AkFs+EssHomiBzICRiIvvlk0cE6a7ZRs1y5xYlUxWAy9canCkSkvEE5u
+                      UGfsMNmioktmULZH6Vwd0fuKOhjGz2aa4JiggVV1O0gj1oMg9GXnatlVyVXg8PRu
+                      hD1fOkB51yN8Up7TjLg7CcS2BaY8GZdie8HaeDYwgyoCa+ihOaA7jsCrJjMHao6r
+                      JVCwue68YHLuJDtxier1ESWt2QSIP8IMIVAguNkwdvcn9sId78WnnMwqD09Hoyr4
+                      Bn5Rj/P3lKWJGe5j3suHUAQeiI5nnS/+aSNHa5Uxxq/lpgzd/DFb2eOyib8O5GPY
+                      DctbUWGUkQdj+PJY6kvx1BN7KjphdchGwV9N8ebAStfbOSiJIFw/dGPNuYQVpPJy
+                      kri6psdhcgjNNxT07txS2eB8DyBYfc8A3oV7eIvX8P80Y84mUduxpBJZ5mW9dBmG
+                      HSdbeCUajLTGwhLcHYuhmCcCAwEAAQ==\n
+                      -----END RSA PUBLIC KEY-----\n)
```
ketan: Are you sure you need a `\n` here?

server/test/integration/com/thoughtworks/go/server/dao/VersionInfoSqlMapDaoIntegrationTest.java
maheshp: Looks like dbHelper.onSetUp() and dbHelper.onTearDown() does not roll back, the tests fail upon removing the explicit deleteAll

server/src/com/thoughtworks/go/server/dao/VersionInfoSqlMapDao.java
maheshp: A GO update is available if the latest_version is greater than the installed_version, this needs to be computed for each request to GO server so that a notifications is shown in case of an update available. Instead of computing this on each request a GO update version(if available) is added in cache. This logic does not seem to be part of DAO, so retaining it in the Manager.

arvindsv:
Cool!

Some thoughts:

1. It looks like the server itself never does an update check. Is this true? 

2. Some logging would be nice: "Checked for updates at: ...". At INFO level maybe.

3. [https://update.go.cd/channels/experimental/latest.json](https://update.go.cd/channels/experimental/latest.json) works, but [https://update.go.cd/channels/supported/latest.json](https://update.go.cd/channels/supported/latest.json) doesn't.

4. It doesn't look like a user can switch between the two easily (admin will need to make a property change). Just an observation.

5. An individual user cannot disable it. Admin will need to, for the whole system. Since we're using local storage, this can be done easily I guess, if needed.

6. If you're making another commit sometime before the release, can you change the message to something like: "A new release of GoCD - 15.3.0-2276 is available. Released on: 10 Dec 2015"? Or at least talk to others about whether a message like that looks ok?

arvindsv:
Also, it looks like if the update check fails, the browser/AJAX script doesn't tell the server about the failure to access it. The server will wait 30 mins, thinking a response will come back. It will not be able to log the fact that an update check was performed and failed.

maheshp:
1. It looks like the server itself never does an update check. Is this true?
- Its not implemented yet, its part of https://github.com/gocd/gocd/issues/1308 which needs to be done

2. Some logging would be nice: "Checked for updates at: ...". At INFO level maybe.
 - Make sense, will add it

3. https://update.go.cd/channels/experimental/latest.json works, but   https://update.go.cd/channels/supported/latest.json doesn't.
- Will follow up on it, supported needs to be up

4. It doesn't look like a user can switch between the two easily (admin will need to make a property change). Just an observation.
- Yes, admin will have to make a property change for this. Since the user is not aware of the update check being done dint see a need to add a switch. 

5. An individual user cannot disable it. Admin will need to, for the whole system. Since we're using local storage, this can be done easily I guess, if needed.
- I assume you are talking about browser making a update check, if yes, just wondering if we need to disable at user level.

6. If you're making another commit sometime before the release, can you change the message to something like: "A new release of GoCD - 15.3.0-2276 is available. Released on: 10 Dec 2015"? Or at least talk to others about whether a message like that looks ok?
- I will talk to others about the message. Adding the date will be a bit of work, can we not do it for now.

arvindsv:
> I assume you are talking about browser making a update check, if yes, just wondering if we need to disable at user level.

I was talking about the notification only, at the user level. As a part of that notification drop-down. Something like, "Ignore this version" and/or "Never notify me about updates". The check should continue to happen, unless disabled at the server level (through the switch, which is fine).

This is not urgent. Can be done later. For next release.

> I will talk to others about the message. Adding the date will be a bit of work, can we not do it for now.

I see why. Because the string that is cached is just the latest version. I'd think about caching the whole "message" part of the update response. It'll allow us to show important security vulnerabilities, etc. as well. We can talk about it later. It's fine done in the next release, again.

We should also talk about how having it show multiple messages (update + security message) can conflict with the "Never notify me about updates" thing I talked about earlier.

Issue #1498, opened 2015-09-16
Agent config save/edit api


jyotisingh:
agent-autoregistration goes through the same set of validations as rest of the config. This causes a lot of disruption when a number of agents are added at the same time esp with a really large config

lenucksi:
I can absolutely confirm this. 
This can also lead to the Go-Server not reacting anymore or becoming timeout-slow. 
Also together with non-Go-server coordinated agent scaling solutions this can lead to issues.

It might be valueable to have a look into the time taken to commit to the config.git repository, especially on larger, older Go servers as this can end up taking quite some time.

scottmuc:
One issue that we're having is that we get too much "config changed" noise around the creation and removal of agents. This happens several dozen times a day (soon to be more). Not sure how you would resolve this... it's not really a problem, it just means that we cannot get much value from seeing that event that configuration has changed.

Stono:
Hey
I can confirm this is an extremely limiting factor for us at the moment, we have 2500 pipelines and are trying to use ephemeral agents but we can't spin up more than about 10 at a time (which happens frequently due to the dependency graph).

PR #1655, opened 2015-11-30
Agent config api - #1498

jyotisingh:
Some of this also affects the internals of pipeline config api, both of these APIs use the same mechanism to validate and save config.

### DIFF
--- a/common/src/com/thoughtworks/go/validation/ConfigUpdateValidator.java
+++ b/common/src/com/thoughtworks/go/validation/ConfigUpdateValidator.java
@@ -1,0 +1,23 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2015 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END***********************************/
+
+package com.thoughtworks.go.validation;
+
+import com.thoughtworks.go.config.CruiseConfig;
+
+public interface ConfigUpdateValidator {
+    boolean isValid(CruiseConfig preprocessedConfig);
+}
--- a/common/test/unit/com/thoughtworks/go/domain/AgentConfigTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/AgentConfigTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 ThoughtWorks, Inc.
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,36 +17,25 @@
 package com.thoughtworks.go.domain;
 
 import com.thoughtworks.go.config.*;
-import com.thoughtworks.go.helper.GoConfigMother;
-import org.junit.Before;
 import org.junit.Test;
 
+import static com.thoughtworks.go.util.TestUtils.contains;
 import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsNull.nullValue;
 import static org.junit.Assert.assertThat;
 
 public class AgentConfigTest {
-
-    private CruiseConfig cruiseConfig;
-    private AgentConfig agentConfig;
-
-    @Before
-    public void setUp() {
-        cruiseConfig = GoConfigMother.configWithPipelines("dev", "qa");
-        agentConfig = new AgentConfig("uuid", "hostname", "10.10.10.10");
-        cruiseConfig.agents().add(agentConfig);
-    }
-
     @Test
     public void agentWithNoIpAddressShouldBeValid() throws Exception {
         CruiseConfig cruiseConfig = new BasicCruiseConfig();
-        AgentConfig agent = new AgentConfig();
+        AgentConfig agent = new AgentConfig("uuid", null, null);
         cruiseConfig.agents().add(agent);
 
         assertThat(cruiseConfig.validateAfterPreprocess().isEmpty(), is(true));
     }
 
     @Test
-    public void shouldValidateIpV4AndIpV6() throws Exception {
+    public void shouldValidateIpCorrectIPv4Address() throws Exception {
         shouldBeValid("127.0.0.1");
         shouldBeValid("0.0.0.0");
         shouldBeValid("255.255.0.0");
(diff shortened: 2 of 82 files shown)


common/test/unit/com/thoughtworks/go/domain/AgentConfigTest.java
```diff
@@ -55,9 +57,9 @@ public void shouldValidateIpV4AndIpV6() throws Exception {
 
     @Test
     public void shouldDetectInvalidIPAddress() throws Exception {
-        shouldBeInvalid("blahinvalid", "IpAddress is invalid.");
-        shouldBeInvalid("blah.invalid", "IpAddress is invalid.");
-        shouldBeInvalid("399.0.0.1", "IpAddress is invalid.");
+        shouldBeInvalid("blahinvalid", "'blahinvalid' is an invalid IP address.");
+        shouldBeInvalid("blah.invalid", "'blah.invalid' is an invalid IP address.");
+        shouldBeInvalid("399.0.0.1", "'399.0.0.1' is an invalid IP address.");
```
ketan: Get rid of or replace all `shouldBeInvalid` and `shouldBeValid` tests with the new test?

config/config-api/src/com/thoughtworks/go/config/AgentConfig.java
```diff
@@ -72,13 +86,13 @@ private void validateIpAddress() {
             return;
         }
         if (address.trim().isEmpty()) {
```
ketan: Change to `StringUtils.isBlank`?

config/config-api/src/com/thoughtworks/go/config/exceptions/GoConfigInvalidException.java
```diff
@@ -12,23 +12,34 @@
...
+    public GoConfigInvalidException(CruiseConfig cruiseConfig, List<ConfigErrors> allErrors) {
+        this(cruiseConfig, ListUtil.join(ListUtil.map(allErrors, new ListUtil.Transformer<String>() {
+            @Override
+            public String transform(Object obj) {
+                ConfigErrors errors = (ConfigErrors) obj;
+                return ListUtil.join(errors.getAll());
+            }
+        })));
+    }
+
```
zabil: Looking for usages of this class and constructor. I feel we need a better API
for e.g. to construct this exception one has to do this 
```
    private GoConfigInvalidException getGoConfigInvalidException() {
        ConfigErrors configErrors = new ConfigErrors();
        configErrors.add("command", "command cannot be empty");
        ArrayList<ConfigErrors> list = new ArrayList<ConfigErrors>();
        list.add(configErrors);
        return new GoConfigInvalidException(new BasicCruiseConfig(), list);
    }

```
(this is a test)

But other usages goes like
```
throw new GoConfigInvalidException(preprocessedConfig, new ArrayList<ConfigErrors>());
```

Also ideally we should have just wrapped/extended the ```List<ConfigErrors>``` onto something like ```AllConfigErrors``` and wrapped this method onto that

```
AllConfigErrors.errorAsString
 {Fancy loops}
```

Atleast

```
              ConfigErrors errors = (ConfigErrors) obj;
               return ListUtil.join(errors.getAll());
```
Could have been

```
              ConfigErrors errors = (ConfigErrors) obj;
               return errors.asString();
```

common/src/com/thoughtworks/go/validation/ConfigUpdateValidator.java
```diff
@@ -0,0 +1,23 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2015 ThoughtWorks, Inc.
```
ketan: You mean 2016, right :)

config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
```diff
@@ -140,7 +140,7 @@ private CruiseConfig validateCruiseConfig(CruiseConfig config) throws Exception
         try {
             List<ConfigErrors> allErrors = validate(config);
             if (!allErrors.isEmpty()) {
-                throw new GoConfigInvalidException(config, allErrors);
+                throw new GoConfigInvalidException(config, allErrors.get(0).firstError());
```
ketan: Why do we throw an exception with just the first error?

server/src/com/thoughtworks/go/server/service/AgentConfigService.java
```diff
@@ -47,45 +68,446 @@ public void register(AgentChangeListener agentChangeListener) {
...
+                    "agentConfig=" + agentConfig +
+                    '}';
+        }
 
-    public void modifyResources(AgentInstance[] agentInstances, List<TriStateSelection> selections) {
-        goConfigService.modifyResources(agentInstances, selections);
     }
 
+    @Deprecated
+    public void updateAgentResources(final String uuid, final Resources resources) {
```
ketan: This does not seem to be used, outside of tests, can we delete this method (or move it to a test helper)?

server/webapp/WEB-INF/rails.new/app/controllers/api_v2/agents_controller.rb
```diff
@@ -33,13 +33,16 @@ def show
...
-      agent_service.updateAgentAttributes(current_user, result, params[:uuid], params[:hostname], maybe_join(params[:resources]), maybe_join(params[:environments]), to_enabled_tristate)
+
+      agent = agent_service.updateAgentAttributes(current_user, result, params[:uuid], params[:hostname], maybe_join(params[:resources]), maybe_join(params[:environments]), to_enabled_tristate)
 
       if result.isSuccess
         load_agent
         render json_hal_v2: agent_presenter.to_hash(url_builder: self)
       else
-        render_http_operation_result(result)
+        @agent_instance = agent
```
ketan: This feels a bit redundant, if the `@agent_instance` instance variable were set on line 37 above, at worst it'll be nil.

server/webapp/WEB-INF/rails.new/app/controllers/api_v2/agents_controller.rb
```diff
@@ -84,7 +87,7 @@ def agent_presenter
     end
 
     def agent_view_model
-      com.thoughtworks.go.server.ui.AgentViewModel.new(@agent_instance, environment_config_service.environmentsFor(@agent_instance.getUuid()))
+      com.thoughtworks.go.server.ui.AgentViewModel.new(@agent_instance, environment_config_service.environmentsFor(@agent_instance.getUuid())) if @agent_instance
```
ketan: Is there a need to be defensive here?

server/webapp/WEB-INF/rails.new/app/presenters/api_v2/agent_representer.rb
```diff
@@ -83,6 +84,14 @@ def free_space
       end
     end
 
+    def errors
+      mapped_errors = agent.errors
+      if ip_errors = mapped_errors.delete('ipAddress')
+        mapped_errors['ip_address'] = ip_errors
+      end
+      mapped_errors
+    end
```
ketan: @juhijariwala has a nice pattern for doing this —

```
class FooRepresenter
  error_representer('ipAddress' => 'ip_address')
end
```

The `error_representer` will take care of the translation, and defining the `errors` property that you declared above.

You may need to copy some code from `ApiV1::BaseRepresenter`

server/webapp/WEB-INF/rails.new/app/presenters/api_v2/base_representer.rb
```diff
@@ -33,7 +33,7 @@ def property(name, options={})
     end
 
     def to_hash(*options)
-      super.deep_symbolize_keys
+      super.deep_symbolize_keys unless @represented.nil?
```
ketan: Why was this needed here, but not in apiv1?

server/webapp/localize_ja.xml
```diff
@@ -780,6 +780,7 @@
     <entry key="SCM_USED_IN_PIPELINES_LINK">Show pipelines using this SCM</entry>
     <entry key="STALE_PIPELINE_CONFIG">Someone has modified the configuration for pipeline ''{0}''. Please update your copy of the config with the changes.</entry>
     <entry key="PIPELINE_CONFIG_VALIDATION_FAILED">Validations failed for pipeline ''{0}''. Please correct and resubmit.</entry>
```
ketan: Duplication? Can `PIPELINE_CONFIG_VALIDATION_FAILED` be implemented via `ENTITY_CONFIG_VALIDATION_FAILED` ?

server/src/com/thoughtworks/go/config/EntityConfigSaveResult.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public T getEntityConfig() {
+        return entityConfig;
+    }
+
+    public GoConfigHolder getConfigHolder() {
+        return configHolder;
+    }
+
+    public ConfigSaveState getConfigSaveState() {
```
ketan: This method seems unused.

ketan:
@jyotisingh - done with the review. Please look at the comments and update all copyright headers with 2016.

PR #1658, opened 2015-12-01
Updating the copy of serverconfig when the config changes as the current copy is stale after an update.

zabil:


arvindsv:
Shouldn't this just be holding on to the old site URL and secure site URL? Why hold on to the whole config?

zabil:
Agreed, that's what i thought too and started off as. Didn't have context to why the config was passed directly. Will make that change.

zabil:
Ok remembered why I didn't do it in the first place. It still end up referring 'ServerSiteUrlConfig' there by holding on the old cruise config. Updated the pull request.

Issue #1660, opened 2015-12-02
Check SCM connection throws incorrect error on pipeline creation wizard


ketan:
Carried over from https://github.com/gocd/gocd/issues/1654#issuecomment-161224928 because this is a separate issue

@ketan identified another issue related to this bug -  When we create a new pipeline from 'Admin -> Pipeline -> Create new pipeline within Group' page and specify material details and perform "Check Connection" throws up error as below,

![image](https://cloud.githubusercontent.com/assets/6533924/11526441/87288162-9900-11e5-97a4-b232fe137d00.png)

The Check Connection is successful on pipeline edit page

rajiesh:
verified this defect on 15.3.0-2698(369b66909146ad43e7291032a27a19454252ac8f). The SCM check connection throws valid error on connection failure. However the error shown is very verbose(more of error log output than a user message) as here

![image](https://cloud.githubusercontent.com/assets/6533924/11586627/7629d500-9a9a-11e5-9cf4-ad995fb34b44.png)


The fix is put this way intentionally since user can get more detail of why it failed. Ketan is of view that its useful as user can get more info rather than showing a vague message like "Connection Failed" or something like that.

I agree on improving the message to be less vague, but showing up message which are gibberish for user can be avoided

Please put your comments so that we can decide how to improve the messaging

jyotisingh:
@rajiesh - this has been the behaviour even in previous releases of Go and intentionally so. The only thing that seem to have changed is the css. The error used to show up as :
<img width="710" alt="screen shot 2015-12-15 at 11 32 11" src="https://cloud.githubusercontent.com/assets/1046342/11803640/c9451854-a322-11e5-9692-b97f07086bff.png">

and now shows up as:
<img width="713" alt="screen shot 2015-12-15 at 11 31 54" src="https://cloud.githubusercontent.com/assets/1046342/11803651/ddbf098e-a322-11e5-9a39-a401c82d2f4e.png">

varshavaradarajan:
Closing this. Because @jyotisingh mentioned that the huge message shown is intentional.

PR #1661, opened 2015-12-02
Check SCM connection throws incorrect error on pipeline creation wizard  (#1660)

ketan:
If the pipeline_name parameter submitted is blank, then don't attempt to lookup that pipeline and throw a 404.


jyotisingh:
@ketan - It doesn't really throw a 404, seems to be handling it. I think this version is fine too unless you think pipeline_name param should be mandatory?

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/material_test_controller_spec.rb
```diff
@@ -96,6 +96,23 @@
 
         expect(response).to have_api_message_response(200, 'Connection OK.')
       end
+
+      it 'performs parameter expansion if pipeline_name param is blank' do
```
jyotisingh: "performs material test" instead of "performs parameter expansion"?

Issue #1630, opened 2015-11-20
Issues Observed in Pipeline config API 


rajiesh:
Tested pipeline config API on 15.3.0-2663 and below issues are observed,

~~Issue 1:~~
~~environment variables: Not validates for unique values, does not validate for special chars.if '#' char is present then it checks and throws error but other chars are not checked

eg:
```
"environment_variables": [
    {
      "secure": false,
      "name": "java%",
      "value": "java21"
    },
    {
      "secure": false,
      "name": "java",
      "value": "java2"
    }
  ]
```

```
"environment_variables": [
    {
      "secure": false,
      "name": "java",
      "value": "java21"
    },
    {
      "secure": false,
      "name": "java",
      "value": "java2"
    }
  ]
```

~~Issue 2~~:

working directory validation do not look for special chars: Below value is accepted
>"working_directory": "some../..rm” 

**Issue 3**:
* Create a pipeline with template(vis UI, config API still do not support create templates)
* Perform a GET on the config API
Json contains below stage value which needs to be removed in case of pipelines using template

> "stage" : null

* Similarly when pipeline is defined by stages, the template property should not be returned by the config API. Currently the response for GET API has

> "template": null

~~Issue 4:~~
* Add github-pr-poller plugin to the go server
* create a pipeline with GitHub material and no path blacklisted
* perform a GET on config API for this pipeline
* In the Json response Material contains the SCM ref and also "filter". Filter value is set to null
* Now make modifications to the pipeline and perform PUT 
* Response received is "500 Server Error" - since the filter value was null and API does not accpet null value for SCM filter
* 'null' value should be accepted for SCM filter ~~

Additional issues identified:

**Issue 5**:

GET Pipeline Config API do not have scm material destination.
* Create a pipeline with 2 materials - github material(precondition: add github.pr plugin to server) and another one
* Specify a destination folder for scm material
* Perform a GET pipeline config request
* response json do not contain the destination folder. here is the snippet of the response,

> {
      "type": "plugin",
      "attributes": {
        "ref": "b337590b-919f-41f7-91de-55aa5d805147",
        "filter": null
      }
    }

* Take this response json and do a PUT on pipeline config API -  Response is 422(Unprocessible entity) but the response body do not have any details of what is the failure
* Actually the failure is that since there are 2 materials and the scm material do not have destination set the request is not processed

**Issue 6**:

[...]

Issue #1417, opened 2015-08-06
Configure a pipeline via API


zabil:
Related to #1404 

Build a new API to save pipeline configuration. 

PR => #1491 

**GET**:  `curl -i 'http://<go-server>:8153/go/api/admin/pipelines/<pipeline-name>' -u 'user:password' -H 'Accept: application/vnd.go.cd.v1+json' -H 'Content-Type: application/json'  -H "If-None-Match: <Etag>"`

**PUT**: `curl -i 'http://<go-server>:8153/go/api/admin/pipelines/<pipeline-name>' -u 'user:password' -H 'Accept: application/vnd.go.cd.v1+json' -H 'Content-Type: application/json' -H 'If-Match: <Etag>' -X PUT -d '{<pipeline json>}'`

**POST**: `curl -i 'http://<go-server>:8153/go/api/admin/pipelines' -u 'user:password' -H 'Accept: application/vnd.go.cd.v1+json' -H 'Content-Type: application/json' -X POST -d '{"group": "<group-name>", "pipeline": <pipeline json>}'`

Here's a draft structure without HAL:

```JSON
{
  "name": "yourproject",
  "label_template": "foo-1.0.${COUNT}-${svn}",
  "locked": true,
  "template_name": null,
  "params": [
    {
      "name": "COMMAND",
      "value": "echo"
    },
    {
      "name": "WORKING_DIR",
      "value": "/repo/branch"
    }
  ],
  "tracking_tool": {
    "type": "mingle",
    "attributes": {
      "base_url": "http://mingle.example.com",
      "project_identifier": "my_project",
      "grouping_conditions": "status > 'In Dev'"
    }
  },
  "timer": "0 0 22 ? * MON-FRI",
  "materials": [
    {
      "type": "svn",
      "attributes": {
        "name": "svn_material",
        "url": "http://your-svn/",
        "dest": "dest_folder",
        "auto_update": true
      }
    }
  ],
  "environment_variables": [
    {
      "name": "MULTIPLE_LINES",
      "value": "multiplelines",
      "secure": true
    },
    {
      "name": "COMPLEX",
      "value": "This has very <complex> data",
      "secure": false
    }
  ],
  "stages": [
    {
      "name": "ut",
      "fetch_materials": true,
      "clean_working_directory": false,
      "never_clean_artifacts": true,
      "approval": {
        "type": "manual|success",
        "authorization": {
          "roles": [],
          "users": []
        }
      },
      "environment_variables": [
        {
[...]

PR #1667, opened 2015-12-04
Fixed multiple bugs on pipeline config save API - #1417, #1630

jyotisingh:


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/tasks/on_cancel_representer.rb
```diff
@@ -27,7 +27,7 @@ def from_hash(hash, options={})
 
         def to_hash(*options)
           return nil if on_cancel_config.getTask.getTaskType.eql?("killallchildprocess")
-          TaskRepresenter.new(on_cancel_config.getTask).to_hash(url_builder: UrlBuilder.new)
+          TaskRepresenter.new(on_cancel_config.getTask).to_hash(url_builder: nil)
```
zabil: Why was this done?

config/config-api/src/com/thoughtworks/go/config/materials/PluggableSCMMaterialConfig.java
```diff
@@ -315,11 +315,11 @@ public boolean equals(Object o) {
             return false;
         }
 
-        if (scmConfig == null || that.scmConfig == null) {
+        if (scmConfig != null ? !scmConfig.equals(that.scmConfig) : that.scmConfig != null) {
             return false;
         }
 
-        return this.getFingerprint().equals(that.getFingerprint());
```
zabil: Ok, why are we removing this finger print check?

jyotisingh: fingerprint is a computed field which internally checks for scmConfig itself. If you see, I have included scmConfig field for equals comparison and removed fingerprint, as is done for the other materials. Also, while saving a config xml, the scmConfig object could be null, and only the ref/scmId would be set. In such a case, equals comparison returned false for materials pointing to the same ref/scmId

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/tasks/on_cancel_representer.rb
jyotisingh: `UrlBuilder` is  a test-only class. For production usage we need to pass along the controller. This was possibly added in error,  hence removing it. Also, a value of `nil` is sufficient as there are no embedded urls in TaskRepresenter or anything under it.

Issue #1650, opened 2015-11-26
Get All Users API returns 500 response when a user has space in the login name


rajiesh:
the get all users API returns a 500 if there is a space in the login name, below exception seen in the log,

```
2015-11-26 14:08:57,331 FATAL [qtp1453889274-29] Rails:? -
ActionController::UrlGenerationError (No route matches {:action=>"update", :controller=>"api_v1/users", :login_name=>"test !"} missing required keys: [:login_name]):
  app/controllers/application_controller.rb:200:in `url_for'
  app/presenters/api_v1/user_summary_representer.rb:27:in `UserSummaryRepresenter'
  app/presenters/api_v1/base_representer.rb:74:in `to_hash'
  app/presenters/api_v1/base_representer.rb:74:in `to_hash'
  app/controllers/api_v1/users_controller.rb:23:in `index'
  lib/jetty_weak_etag_middleware.rb:13:in `call'
  lib/catch_json_parse_errors.rb:8:in `call'
```

rajiesh:
verified it on 15.3.0-2753(7aa885fe5c2d71de7544ed776f2c6e16ccc9a79a). GET all users returns user details with details of users with space in the login name. However we cannot do a single user GET with login name containing space(other commonly used special chars works fine - dot(.), &, colon(:) , lt(<) gt(>), ! , @). GET / DELETE / PATCH API cannot be invoked on users with login name containing space.

rajiesh:
created the issue #1706 for the login name with space, so this issue can be closed for other special chars

PR #1668, opened 2015-12-04
Fix for Users API where login_name has special characters #1650

maheshp:
Verified the following scenarios 
?foo?bar?
.foo.bar.
/foo/baz/
*foo*baz*
foo!foo
?foo&bar
+foo@bar&q+
;foo;bar;
"foo"bar"
foo#bar#
>foo<bar<
@foo<bar<
foo?bar<
(foo~*bar+&)
foo+bar

### DIFF
--- a/server/webapp/WEB-INF/urlrewrite.xml
+++ b/server/webapp/WEB-INF/urlrewrite.xml
@@ -134,6 +134,21 @@
         <set name="rails_bound">true</set>
     </rule>
 
+    <!-- Rule for Rails Users API should be above the generic rule 'Rails API' -->
+    <rule>
+        <name>Rails Users Index and Delete API</name>
+        <from>^/api/users$</from>
+        <to last="true">/rails/api/users</to>
+        <set name="rails_bound">true</set>
+    </rule>
+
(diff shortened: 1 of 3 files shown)


ketan:
Can you write a test under urlrewriteintegrationtest to make sure this does not regress?

server/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -136,12 +136,19 @@
 
     <rule>
         <name>Rails API</name>
-        <from>^/(api/(?!admin/config).*)$</from>
+        <from>^/(api/((?!admin/config)(?!users/)).*)$</from>
```
ketan: @maheshp - can we pull this `/users` into a separate rule — while this would be a bit noisy, IMO it'll help when removing individual paths from the rewrite.

WDYT?

maheshp: Ok, I can move it into a separate rule.

I was unable to run the tests even for a single rule, it takes a long time and always times out.  Anyways will add a test, will have to check if it breaks on CI.

Issue #1649, opened 2015-11-26
feature toggle keys to be removed


rajiesh:
The below feature toggle keys are available from last release and turned on by default. No issues reported so far, so its time to remove them

* color_logs_feature_toggle_key
* api_requests_short_session_feature_toggle_key

rajiesh:
verified this fix on 15.3.0-2753(7aa885fe5c2d71de7544ed776f2c6e16ccc9a79a) - the toggles are removed and the regression is pass

PR #1673, opened 2015-12-04
removed feature toggle color_logs_feature_toggle_key - #1649

alexschwartz:
tried to remove feature toggle 'color_logs_feature_toggle_key' - #1649
builds fine locally

### DIFF
--- a/server/resources/available.toggles
+++ b/server/resources/available.toggles
@@ -12,11 +12,6 @@
          "value": false
       },
       {
-         "key": "color_logs_feature_toggle_key",
-         "description": "Enables colored logs in job detail. Default: on.",
-         "value": true
-      },
-      {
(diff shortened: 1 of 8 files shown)


ketan:
Hi, there is also some code in build_output_observer.js, that may be related. I have not seen it in a while, but when implementing this feature, I recalling making some changes there.

alexschwartz:
Hi, thanks for mentioning it. There is indeed more code to be removed -- I just passed "true" for a parameter which is actually obsolete.

ketan:
hey @jyotisingh @zabil — could you look into why the PR plugin is not reporting status here?

ketan:
@alexschwartz thank you for your contribution. The next experimental build at http://www.go.cd/download will have your changes in it.

Issue #1335, opened 2015-07-13
Console log - scroll to end of logs does not happen by default


jyotisingh:
Hosted Go server on AWS, accessing it using google chrome v43 on fedora machine.
When I navigate to console logs from dashboard for a completed build, the "scroll to end of logs" is "green" by default, but the scrolling does not happen. One needs to click on the button twice for it to work.

Similar issue exist in IE9 and FF on fedora.

zabil:
Also noticed this on Go Version: 15.2.0(2246-acdef8b3cb2771)
Mac: 10.10.4 (14E46)
Safari: Version 8.0.7 (10600.7.12)
Chrome: Version 43.0.2357.132 (64-bit)

rajiesh:
verified on 15.3.0-2726(74fa3313ce429737cdc00a94d420b1fcc58dafe7) and is working as expected

PR #1679, opened 2015-12-07
Fixed console tab not scrolling to end on page load #1335

maheshp:


ketan:
@maheshp - to be clear this JS kicks off on the console page as well as the failures tab, so make sure that the selector works there as well.

maheshp:
@ketan verified both on console and failures tab, it looks ok

server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
```diff
@@ -31,12 +31,16 @@ BuildOutputObserver.prototype = {
...
         this.autoScrollButton = jQuery('.auto-scroll');
         this.autoScrollButton.toggleClass('tailing', this.enableTailing);
         this.autoScrollButton.on('click', function(){
             self.enableTailing = !self.enableTailing;
             self.initializeScroll();
         });
+        this.consoleTabElement.on('click', function(){
+            self.initializeScroll();
+        });
         this.initializeScroll();
```
zabil: you can shorten this to 
```
this.consoleTabElement.on('click', self.initializeScroll);
```

PR #1684, opened 2015-12-08
Miscellaneous fixes to SPA —

ketan:
* Improve handing of password/encrypted fields
* Improve look and feel and feedback when saving
* Use opensans font throughout
* Refactor to improve performance based on browser profiling


arvindsv:
> Use opensans font throughout

I haven't seen how it looks yet, but my first reaction was "Huh??". I'm sure it looks good. :) Is it what we use elsewhere?

arvindsv:
Ok, I take that back. It looks nice. I like it much better than before. I really like the nice help tool tips too. The stages are well separated from the materials.

I have some thoughts.

1. Parameters section: The "?" doesn't say anything.

2. I think "Clean working directory" should be called "Delete working directory every time". Go does a clean all the time, but that option causes it to delete and re-check out. People check it all the time, but they shouldn't need to.

3. The text around "Number of Jobs" should be: "Run on one agent", "Run on all agents" and "Run ___ instances". The fact that it says: "Run on ____ agents" is wrong. We have no guarantees that the "Run on X instances" thing actually runs on different agents.

4. It doesn't resize well, but I suppose that needs more thought.

5. When a material has no name, it just shows an empty bar. That bar can be used to show so much more relevant information than just the name of the material.

6. "Ignore pattern" really needs a tooltip or a link to better documentation.

7. When I tried the save, the button went red, with no message. Console showed a 422 (Unprocessable entitiy). This is the response I got with it (didn't show up anywhere):

```
{
  "message": "Validations failed for pipeline 'Test'. Please correct and resubmit.",
  "data": {
    "_links": {
      "self": {
        "href": "http://localhost:8153/go/api/admin/pipelines/Test"
      },
      "doc": {
        "href": "http://api.go.cd/#pipeline_config"
      },
      "find": {
        "href": "http://localhost:8153/go/api/admin/pipelines/:name"
      }
    },
    "label_template": "${COUNT}",
    "enable_pipeline_locking": false,
    "name": "Test",
    "template": null,
    "parameters": [

    ],
    "environment_variables": [

    ],
    "materials": [
      {
        "type": "git",
        "attributes": {
          "url": "abc",
          "destination": null,
          "filter": null,
          "name": "Abc",
          "auto_update": true,
          "branch": "master",
          "submodule_folder": null
        },
        "errors": {
          "folder": [
            "Destination directory is required when specifying multiple scm materials"
          ]
        }
      },
      {
        "type": "hg",
        "attributes": {
          "url": "abc",
          "destination": null,
          "filter": null,
          "name": "Blah",
          "auto_update": true
        },
        "errors": {
          "folder": [
            "Destination directory is required when specifying multiple scm materials"
          ]
        }
      }
    ],
    "stages": [
      {
        "name": "defaultStage",
        "fetch_materials": true,
        "clean_working_directory": false,
        "never_cleanup_artifacts": false,
        "approval": {
          "type": "success",
          "authorization": {
            "roles": [

            ],
            "users": [

            ]
          }
        },
        "environment_variables": [

        ],
        "jobs": [
          {
            "name": "defaultJob",
            "run_instance_count": null,
            "timeout": null,
            "environment_variables": [

            ],
            "resources": [

            ],
            "tasks": [
              {
                "type": "ant",
                "attributes": {
                  "working_directory": "",
                  "build_file": "",
                  "target": ""
                }
              }
            ],
            "tabs": [

            ],
            "artifacts": [

            ],
            "properties": null
          }
        ]
      },
      {
        "name": "Test",
        "fetch_materials": true,
        "clean_working_directory": false,
        "never_cleanup_artifacts": false,
        "approval": {
          "type": "success",
          "authorization": {
            "roles": [

            ],
            "users": [

            ]
          }
        },
        "environment_variables": [

        ],
        "jobs": [
          {
            "name": "",
            "run_instance_count": null,
            "timeout": null,
            "environment_variables": [

            ],
            "resources": [

            ],
            "tasks": [

            ],
            "tabs": [

            ],
            "artifacts": [

            ],
            "properties": null,
            "errors": {
              "name": [
                "Name is a required field"
              ]
            }
          }
        ]
      }
    ],
    "tracking_tool": null,
    "timer": null
  }
}
```

arvindsv:
The jobs section is still too busy.

It has:

Name
Some options
Tasks
Some more options

It should be, IMHO:

Tasks
Options (hidden in accordion, including name there - name should default to something)

arvindsv:
Bug?

1. Click "Add Stage"

2. Immediately click "Manual" in the "Stage Type".

It gets stuck and you can't move between stages or to materials. I think it's because the "Users" and "Roles" part doesn't show up.

Would also be nice if the arrow between stages changes to show that the next stage is manual.

arvindsv:
The whole of "On Cancel" tasks are missing, right?

arvindsv:
Also, "Run If" conditions.

ketan:
Authorization users and roles should always show up, independent of whether the stage is triggered on success or manually

ketan:
> The text around "Number of Jobs" should be: "Run on one agent", "Run on all agents" and "Run ___ instances". The fact that it says: "Run on ____ agents" is wrong. We have no guarantees that the "Run on X instances" thing actually runs on different agents.

Does it make sense to rephrase the "Number of jobs" to "job parallelism" (or something to that effect)?

arvindsv:
> Does it make sense to rephrase the "Number of jobs" to "job parallelism" (or something to that effect)?

No, that might cause more confusion, since jobs are run in parallel anyway.

ketan:
@jyotisingh - I know we're not shipping by calling this as DONE, but this PR has a lot of fixes that allows me to use the SPA to test the pipeline API. Could you merge this on priority?

This PR contains only JS and CSS changes, and no changes to the server, so nothing that we are calling DONE will break.

arvindsv:
Bug: If you have only one job, it's not possible to delete it (and it shows no feedback about why).

Console prints this out:

image

arvindsv:
Also, currently errors don't show up at all. Having said all this, I think this PR is the best version of the SPA we have right now, and should be merged in soon (just thinking from a UI/UX perspective, not from code. @jyotisingh or someone else will need to take a deeper look, I guess).

ketan:
Does not handle and distinguish between `srcfile` and `srcdir` attributes of `<fetchartifact />` task.

PR #1685, opened 2015-12-08
fix #1132: The entirety of JUnit XML files is skipped if they are minified

### LINKED ISSUE #1132: The entirety of JUnit XML files is skipped if they are minified

xli:
Fix  #1132, changes:
* Use Jdom to parse out unit test report xml file instead of using regex expression, when merging all test files
* Output message when a test report xml file can't be parsed as XML document

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
+++ b/config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
@@ -16,12 +16,10 @@
 
 package com.thoughtworks.go.domain;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -37,10 +35,13 @@ import javax.xml.xpath.XPathFactory;
 
 import com.thoughtworks.go.util.FileUtil;
 import com.thoughtworks.go.util.TestFileUtil;
+import com.thoughtworks.go.util.XmlUtils;
 import com.thoughtworks.go.util.XpathUtils;
 import com.thoughtworks.go.work.GoPublisher;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.jdom.Document;
+import org.jdom.JDOMException;
 
 public class UnitTestReportGenerator implements TestReportGenerator {
     private final File folderToUpload;
@@ -151,18 +152,12 @@ public class UnitTestReportGenerator implements TestReportGenerator {
     }
 
     private void pumpFileContent(File file, PrintStream out) throws IOException {
-        BufferedReader bufferedReader = null;
         try {
-            bufferedReader = new BufferedReader(new FileReader(file));
-            String line = bufferedReader.readLine();
-            if (!line.contains("<?xml")) { // skip prolog
-                out.println(line);
-            }
-            while ((line = bufferedReader.readLine()) != null) {
-                out.println(line);
-            }
-        } finally {
-            IOUtils.closeQuietly(bufferedReader);
+            String content = FileUtil.readToEnd(file).trim().replaceFirst("^\\W+<", "<");
+            final Document document = XmlUtils.buildXmlDocument(content);
+            XmlUtils.writeXml(document.getRootElement(), out);
+        } catch (JDOMException e) {
+            publisher.consumeLine(MessageFormat.format("The file {0} could not be parsed as XML document: {1}", file.getName(), e.getMessage()));
         }
     }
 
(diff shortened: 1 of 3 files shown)


arvindsv:
@xli: Nice! First PR! It makes sense to me (the changes) and the PR build passed (though it did not update GitHub). Need to check why.

I don't really understand why the previous code was excluding the prolog. Maybe having the prolog will cause a problem somewhere? Let's see what others think.

Congratulations!! Hopefully, this is the first of many PRs. :)

ketan:
The intent of this code is to "concatenate" multiple XML test result files into one, so that go can run one giant XSLT operation on it. This XSLT output then seems to go to the failures tab.

Since each individual test report XML would have the XML prolog in it, a simple concat would not work. Hence the stripping of the prolog. I think @xli's solution is much neat; so wow; very awesome. I'll look at this by EOD.

Welcome on board, good to have you!

config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
```diff
@@ -151,18 +154,12 @@ private void pumpFileContentIfValid(PrintStream out, File testFile) throws IOExc
...
-                out.println(line);
-            }
-            while ((line = bufferedReader.readLine()) != null) {
-                out.println(line);
-            }
-        } finally {
-            IOUtils.closeQuietly(bufferedReader);
+            String content = FileUtil.readToEnd(file).trim();
+            final Document document = XmlUtils.buildXmlDocument(content);
+            out.println(XmlUtils.asXml(document.getRootElement()));
```
ketan: If you use `XmlUtils.writeXml(Document, OutputStream)`, this new impl (`XmlUtils.asXml(Element)`) is redundant. Feel free to rename that method if you feel it appropriate.

common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
ketan: I think you may have removed the BOM there, was that intentional?

**before:**

```
$ file common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml
common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml: XML 1.0 document, UTF-8 Unicode (with BOM) text
```

**after:**

```
$ file common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml
common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml: XML 1.0 document, ASCII text
```

config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
xli: writeXml(Document, OutputStream) will output "<?xml version="1.0" encoding="utf-8"?>"

common/test-resources/unit/data/NunitTestResultWithByteOrderMark.xml
xli: With BOM, it is not a valid xml file.

arvindsv: I think the test name suggests that the BOM was left there intentionally.

xli:
Reverted NunitTestResultWithByteOrderMark.xml changes. Handles BOM after read file content. Using regex /^\W+</ to clean it up is the simplest one in our case.

ketan:
Hey @xli - a commit message containing `fix(es) #number` will close the original issue when the PR gets merged, which means it falls through the cracks as far as QA is concerned. For the future please don't write `fix(es) or close(s)` in the commit message :) Just mentioning the original issue `#number` should be sufficient to tie PRs to issues. #githubGotchas

PR #1690, opened 2015-12-10
Added logging for Go Update check, changed notification message - #1614

maheshp:


arvindsv:
This looks good to me.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/version_infos_controller.rb
```diff
@@ -20,8 +20,10 @@ class VersionInfosController < ApiV1::BaseController
 
     def update_server
       go_latest_version = ApiV1::GoLatestVersion.new(params, system_environment)
-
-      return render_message(message_tampered_error_message, 400, {}) unless go_latest_version.valid?
+      unless go_latest_version.valid?
+        Rails.logger.error('[GO Update Check] Latest version update failed, version information from update server tampered.')
```
ketan: Just `Go` not `GO`. Also `GoCD`,  not `GOCD`

server/src/com/thoughtworks/go/server/service/ServerVersionInfoBuilder.java
```diff
@@ -21,7 +21,8 @@
 import com.thoughtworks.go.domain.exception.VersionFormatException;
 import com.thoughtworks.go.server.dao.VersionInfoDao;
 import com.thoughtworks.go.server.util.ServerVersion;
-import org.apache.log4j.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
ketan: And thank you for using `slf4j` :)

maheshp: thanks to @jyotisingh for slf4j

PR #1691, opened 2015-12-10
Update UrlUtil.java

ziscloud:
Check length of queryFragmentSplit before access it.

2015-12-10 16:08:41,264 FATAL [qtp2121744517-17] Rails:? -
Java::JavaLang::ArrayIndexOutOfBoundsException (1):
  com.thoughtworks.go.util.UrlUtil$QueryTuple.parse(com/thoughtworks/go/util/UrlUtil.java:122)
  com.thoughtworks.go.util.UrlUtil.urlWithQuery(com/thoughtworks/go/util/UrlUtil.java:57)
  java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:497)

### DIFF
--- a/common/src/com/thoughtworks/go/util/UrlUtil.java
+++ b/common/src/com/thoughtworks/go/util/UrlUtil.java
@@ -119,7 +119,9 @@ public class UrlUtil {
             String[] split = query.split(QUERY_SEPERATOR);
             for (String queryFrag : split) {
                 String[] queryFragmentSplit = queryFrag.split(QUERY_KEY_VAL_SEPERATOR);
-                parsed.add(new QueryTuple(queryFragmentSplit[0], decode(queryFragmentSplit[1])));
+                if (queryFragmentSplit.length == 2) {
+                    parsed.add(new QueryTuple(queryFragmentSplit[0], decode(queryFragmentSplit[1])));
+                }
             }
             return parsed;
         }


zabil:
Can we add a test for this at UrlUtilTest ?

common/src/com/thoughtworks/go/util/UrlUtil.java
```diff
@@ -119,7 +119,9 @@ public static String getQueryParamFromUrl(String url, String paramName) {
             String[] split = query.split(QUERY_SEPERATOR);
             for (String queryFrag : split) {
                 String[] queryFragmentSplit = queryFrag.split(QUERY_KEY_VAL_SEPERATOR);
-                parsed.add(new QueryTuple(queryFragmentSplit[0], decode(queryFragmentSplit[1])));
+                if (queryFragmentSplit.length == 2) {
+                    parsed.add(new QueryTuple(queryFragmentSplit[0], decode(queryFragmentSplit[1])));
+                }
```
zabil: This will delete query param failing this test
```
assertThat(UrlUtil.urlWithQuery("http://baz.quux?bang&hello=world", "foo", "bar"), is("http://baz.quux/?bang&hello=world&foo=bar"));
```

It's valid to have query param without a value. 
A discussion about that http://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value

Here's how it's used

https://github.com/gocd/gocd/blob/53de3722953779500d73f6ed6102dab08f348b8c/server/webapp/WEB-INF/rails.new/app/controllers/admin_controller.rb#L42

I think we should modify the code to **not** drop the query param.
Also, do you have more of the stack trace so that we can figure out the url being parsed?

zabil:
Moving this to cold storage.
Will open if there's any update

PR #1692, opened 2015-12-10
#1630 , #1417 - scmConfig and packageDefinition were not being set for material while…

jyotisingh:
… saving through pipeline config api, causing a NullPointerException when browsing through page on dashboard

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/domain/packagerepository/PackageRepositories.java
+++ b/config/config-api/src/com/thoughtworks/go/domain/packagerepository/PackageRepositories.java
@@ -112,7 +112,7 @@ public class PackageRepositories extends BaseCollection<PackageRepository> imple
         }
     }
 
-    public PackageDefinition findPackageWith(String packageId) {
+    public PackageDefinition findPackageDefinitionWith(String packageId) {
         for (PackageRepository packageRepository : this) {
             for (PackageDefinition packageDefinition : packageRepository.getPackages()) {
                 if (packageDefinition.getId().equals(packageId)) {
--- a/config/config-api/src/com/thoughtworks/go/domain/packagerepository/PackageRepositories.java
+++ b/config/config-api/src/com/thoughtworks/go/domain/packagerepository/PackageRepositories.java
@@ -1,18 +1,18 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2015 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.domain.packagerepository;
 
@@ -110,6 +110,17 @@ public class PackageRepositories extends BaseCollection<PackageRepository> imple
                 packageDefinition.validateFingerprintUniqueness(map);
             }
         }
+    }
+
+    public PackageDefinition findPackageWith(String packageId) {
+        for (PackageRepository packageRepository : this) {
+            for (PackageDefinition packageDefinition : packageRepository.getPackages()) {
+                if (packageDefinition.getId().equals(packageId)) {
+                    return packageDefinition;
+                }
+            }
+        }
+        return null;
     }
 }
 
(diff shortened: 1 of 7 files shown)


config/config-api/src/com/thoughtworks/go/domain/packagerepository/PackageRepositories.java
```diff
@@ -111,5 +111,16 @@ private void validateFingerprintUniqueness() {
             }
         }
     }
+
+    public PackageDefinition findPackageWith(String packageId) {
```
zabil: Can we keep name and behavior consistent with the other method
```
findPackageRepositoryHaving
```

Maybe findPackageDefinitionWith
Also how the return us handled 
```
        throw new RuntimeException(format("Could not find repository for given package id:[%s]", packageId));
```

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/package_material_representer.rb
```diff
@@ -15,21 +15,17 @@
...
-                  end
-
-                def package_id=(value)
-                    material_config.setPackageId(value)
-                  end
-              end
-          end
+        property :packageId, as: :ref, setter: lambda { |value, options|
+          package_definition = options[:config].getPackageRepositories().findPackageWith(value)
+          self.setPackageDefinition(package_definition) if (package_definition)
```
zabil: Is this why a null is being returned?

jyotisingh: Yes, throwing an exception will cause this to fail. When a null is returned, the validation takes care of adding appropriate error. I wouldn't go about throwing an exception in this case. And fixing the other method would have some implications to handle, for instance [this](https://github.com/gocd/gocd/blob/000fc0f7c60902dd05699a182310cfe4690b059c/server/webapp/WEB-INF/rails.new/app/controllers/admin/package_definitions_controller.rb#L79) does not handle null.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/pluggable_scm_material_representer.rb
```diff
@@ -20,7 +20,12 @@ module Materials
             class PluggableScmMaterialRepresenter < ApiV1::BaseRepresenter
                 alias_method :material_config, :represented
 
-                property :scm_id, as: :ref
+                property :scmId, as: :ref, setter: lambda { |value, options|
+                  scm = options[:config].getSCMs().find(value)
+                  self.setSCMConfig(scm) if(scm)
+                  self.setScmId(value)
+                }
```
ketan: Should this  `raise ApiV1::UnprocessableEntity` error if the SCM is not found? —


```ruby
property :scmId, as: :ref, setter: lambda { |val, opts|
  if scm = options[:config].getSCMs().find(value)
    self.setSCMConfig(scm)
    self.setScmId(value
  else
    raise ApiV1::UnprocessableEntity, "A pluggable SCM with identifier `#{value}` was not found!"
  end
}
```

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
```diff
@@ -50,7 +50,7 @@ def update
 
       def get_pipeline_from_request
         @pipeline_config_from_request ||= PipelineConfig.new.tap do |config|
-          ApiV1::Config::PipelineConfigRepresenter.new(config).from_hash(params[:pipeline])
+          ApiV1::Config::PipelineConfigRepresenter.new(config).from_hash(params[:pipeline], {config: go_config_service.getCurrentConfig()})
```
ketan: Can we call this `cruise_config` instead of `config` to reduce ambiguity?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/package_material_representer.rb
```diff
@@ -15,21 +15,17 @@
...
-                def package_id=(value)
-                    material_config.setPackageId(value)
-                  end
-              end
-          end
+        property :packageId, as: :ref, setter: lambda { |value, options|
+          package_definition = options[:config].getPackageRepositories().findPackageDefinitionWith(value)
+          self.setPackageDefinition(package_definition) if (package_definition)
+          self.setPackageId(value)
+        }
```
ketan: `Raise ApiV1::UnprocessableEntity`?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/pluggable_scm_material_representer.rb
jyotisingh: @ketan - Well, the validation does happen at the server side and the response json would contain this error. The issue with throwing exceptions in a places like this is, it wouldn't try to validate the rest of the entity, so a user would get one error at a time. Thats frustrating from a user's point of view.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/package_material_representer.rb
jyotisingh: No, same reason as above

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
jyotisingh: sure, will rename

PR #1694, opened 2015-12-11
Pipeline Config API bug fixes #1630, #1417

### LINKED ISSUE #1630: Issues Observed in Pipeline config API

juhijariwala:


config/config-api/src/com/thoughtworks/go/config/ValidationContext.java
```diff
@@ -49,5 +51,9 @@
     SecurityConfig getServerSecurityConfig();
 
     boolean doesTemplateExist(CaseInsensitiveString template);
+
+    SCM findScm(String scmID);
```
ketan: Can we rename this to `findScmById`?

juhijariwala: I will do that.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/artifact_representer.rb
```diff
@@ -39,17 +39,21 @@ def type
...
-        end
+      private
+      def error_keys
+        {"src" => "source", "dest" => "destination"}
+      end
 
-        mapped_errors
+      def matching_error_key key
+        return error_keys[key] if error_keys[key]
+        key
```
ketan: `error_keys[key] || key` reads better :)

I like this pattern. Can we extract this into a class with usage that looks like this, and use it everywhere?

```ruby
def errors
  ErrorMapper.new(artifact.errors, 'src' => 'source', 'dest' => 'destination').mapped_errors
end
```

ketan: Also, this is a good place to use `inject` instead of `each` —

ketan:
```ruby
artifact.errors.inject({}) do |memo, (key, value)|
  translated_key = error_keys[key] || key
  memo[translated_key] = value
  memo
end
```

config/config-api/src/com/thoughtworks/go/config/ValidationContext.java
```diff
@@ -49,5 +51,9 @@
...
+
+    PackageRepository findPackage(String packageId);
```
ketan: `byId` here too?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/artifact_representer.rb
ketan: On second thought — this DSL probably reads better —

```ruby
class FooRepresenter < BaseRepresenter
    has_errors('src' => 'source', 'dest' => 'destination')
end
```

and let `has_errors` meta-program the necessary methods. Ping me or @maheshp if you'd like to understand this better.

PR #1695, opened 2015-12-11
Gracefully handle cases where ambiguous environment variable definition is flagged as an error (#1630, #1417)

ketan:
When users submit both `value` and `encrypted_value` for a `secure` variable,
the input will be considered as ambiguous and flagged as such

```json
{
  "secure": true,
  "name": "PASSWORD",
  "value": "clearText"
  "encrypted_value", "c!ph3rt3xt"
}
```

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
@@ -16,12 +16,6 @@
 
 package com.thoughtworks.go.config;
 
-import java.io.Serializable;
-import java.util.HashMap;
-import java.util.Map;
-
-import javax.annotation.PostConstruct;
-
 import com.thoughtworks.go.domain.ConfigErrors;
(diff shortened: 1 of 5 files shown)


config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
```diff
@@ -263,28 +277,51 @@ public void setConfigAttributes(Object attributes) {
...
 
-    public String getValueForDisplay(){
-        if(isSecure){
+    public void deserialize(String name, String value, boolean isSecure, String encryptedValue) throws InvalidCipherTextException {
+        setName(name);
+        setIsSecure(isSecure);
+
+        if (isSecure) {
+            if (isNotBlank(value) && isNotBlank(encryptedValue)) {
+                System.out.println("lo = ");
```
jyotisingh: missed this?

jyotisingh: missed this?

PR #1697, opened 2015-12-11
The entirety of JUnit XML files is skipped if they are minified (#1132)

xli:
* Changed to use BOMInputStream from commons-io to ignore BOM in test report xml file.
* Use Jdom to parse out unit test report xml file instead of using regex expression, when merging all test files
* Output message when a test report xml file can't be parsed as XML document

(This reverts commit 84f1458363eccb3efc205da90f3bf58b69e24a9b.)

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
+++ b/config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
@@ -16,12 +16,10 @@
 
 package com.thoughtworks.go.domain;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


config/config-api/src/com/thoughtworks/go/domain/UnitTestReportGenerator.java
```diff
@@ -151,18 +153,12 @@ private void pumpFileContentIfValid(PrintStream out, File testFile) throws IOExc
...
-            String line = bufferedReader.readLine();
-            if (!line.contains("<?xml")) { // skip prolog
-                out.println(line);
-            }
-            while ((line = bufferedReader.readLine()) != null) {
-                out.println(line);
-            }
-        } finally {
-            IOUtils.closeQuietly(bufferedReader);
+            String content = FileUtil.readToEnd(new BOMInputStream(new FileInputStream(file)));
```
ketan: I looked at the implementation of `FileUtil.readToEnd` and it does not seem like we're closing the `InputStream`, so we may be leaking an inputstream.

The rest looks OK.

ketan:
@jyotisingh — except for the comment about leaking a file handle, is this OK to merge given that you're the gatekeeper in charge for the release?

jyotisingh:
To me it looks like a small change to test so I guess it should be fine. BTW @zabil is the gatekeeper not me :)

PR #1699, opened 2015-12-14
Pass the installed version of Go while fetching the latest version

maheshp:


jyotisingh:
@maheshp - you can use `jQuery.ajax.calls.mostRecent().args[0]['url']` to assert the url in the jasmine tests

maheshp:
@jyotisingh done as part of https://github.com/gocd/gocd/pull/1702, thanks

PR #1713, opened 2015-12-15
Fixing build

juhijariwala:


server/webapp/WEB-INF/rails.new/app/controllers/admin/package_definitions_controller.rb
```diff
@@ -76,7 +76,7 @@ def subject(cruise_config)
       end
 
       def update(cruise_config)
-        packageRepository = cruise_config.getPackageRepositories().findPackageRepositoryHaving(params[:package_id])
+        packageRepository = cruise_config.getPackageRepositories().findPackageRepositoryWithPackageIdOrBomb(params[:package_id])
```
ketan: Why did this need to change?

PR #1716, opened 2015-12-16
Rename `exec` tasks `command` attribute to `executable`.

ketan:
* and some other formatting changes.


jyotisingh:
Changes look good to me. Can we rename this in `ExecTask` and xml too for consistency's sake?

ketan:
> Changes look good to me. Can we rename this in ExecTask and xml too for consistency's sake?

Can we do this for 16.1?

jyotisingh:
@ketan - will you be submitting another PR to take care of renaming the attribute in xml and ExecTask class or updating this one? Also, you need to resolve some conflicts on this one.

ketan:
I've resolved conflicts, however have to create an ApiV2 for this change, which I'll do later.

Also, opened #1755 to rename the `<exec/>` task attribute, if you're OK with that change, I'll update both the PRs to use ApiV2

jyotisingh:
Check comment on #1755

PR #1718, opened 2015-12-16
Gracefully handle cases where ambiguous scm password definition is flagged as an error (#1630, #1417)

ketan:
When users submit both `password` and `encrypted_password` for an SCM
material, the input will be considered as ambiguous and flagged as such

```json
{
  "type": "svn",
  "attributes": {
    "url": "http://funk.com/blank",
    "username": "bob",
    "password": "p@ssw0rd",
    "encrypted_password": "c!iph3rt3xt"
  }
}
```

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/perforce_material_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/perforce_material_representer.rb
@@ -22,8 +22,11 @@ module ApiV1
         property :server_and_port, as: :port
         property :user_name, as: :username
         property :password,
-                 skip_nil:    true,
-                 skip_render: true
+                 skip_render: true,
+                 skip_nil: true,
+                 setter:      lambda { |value, options|
+                   self.setCleartextPassword(value)
+                 }
         property :encrypted_password, skip_nil: true
         property :use_tickets
         property :view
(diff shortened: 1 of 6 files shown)


ketan:
## NOT TO BE MERGED YET!

This one is based off the SPA fixes, and the implementation itself is hacky for lack of java apis to clearly set the password and encrypted password on scm config objects.

I'm submitting this PR merely to see if there's a better alternative that does not involve the hack here.

ketan:
Can someone provide any insights or comments?

arvindsv:
Makes sense to me. I'd still prefer not to send the encrypted value back and forth at all. But, from what this PR claims to want to do, it makes sense to me.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/perforce_material_representer.rb
```diff
@@ -18,12 +18,16 @@ module ApiV1
   module Config
     module Materials
       class PerforceMaterialRepresenter < ScmMaterialRepresenter
+        P4MaterialConfig.send(:field_accessor, 'password')
```
ketan: @jyotisingh can you review this PR, this line (and similar lines in other classes) feels like a really bad hack. But I don't know how to cleanly get away with something similar without resorting to breaking the old pages.

jyotisingh: @ketan - the alternative would be introduce a new method in the java class which sets the password field directly without touching encryptedPassword, not sure what to call it though.

ketan: I'll rebase this PR, if you're fundamentally ok with this fix.

PR #1736, opened 2015-12-21
fix multiple agents setup recommendation

tomzo:
Same content as in https://github.com/gocd/documentation/pull/205

### DIFF
--- a/installers/agent/release/README.md
+++ b/installers/agent/release/README.md
@@ -1,10 +1,26 @@
 ### Run multiple agents on the same host
 
 ## Setup new agent
-```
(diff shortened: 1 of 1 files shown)


ketan:
I’m just going to verify it on a couple of versions of debian and ubuntu running ~~upstart~~, ~~systemd~~, ~~sysinit~~ systemd and then merge.

ketan:
I've not gotten time to test this with `insserv` and `update-rc.d` scripts, but fundamentally it looks OK, so I'm going to merge it in anyways.

PR #1739, opened 2015-12-22
#000 - Don't send 'Server:' header from Jetty

arvindsv:
It's a bit of a security hole to provide more information than needed.

Before:
```
$ curl -I http://localhost:8153
HTTP/1.1 301 Moved Permanently
Date: Tue, 22 Dec 2015 16:31:21 GMT
Location: /go/home
Content-Type: text/html; charset=ISO-8859-1
Content-Length: 13
Server: Jetty(9.2.z-SNAPSHOT)
```

After:
```
$ curl -I http://localhost:8153
HTTP/1.1 301 Moved Permanently
Date: Tue, 22 Dec 2015 16:31:21 GMT
Location: /go/home
Content-Type: text/html; charset=ISO-8859-1
Content-Length: 13
```


ketan:
It's no safer than reading out the version on the login page and checking what jetty version corresponds to the go version.

I'm fine with this change, but IMO it's security theatre :D

arvindsv:
Even more so for an open-source project, where code is available easily. I agree. It just makes some drive-by reconnaissance harder, that's all.

underyx:
This seems completely pointless, and I have no idea why someone would include this in the 'Important features and bug fixes in this release' list for 16.1.

PR #1740, opened 2015-12-23
Throwing BadCredentialsException when user is not found in configured…

jyotisingh:
… ldap so a potential attacker would not be able to differentiate between non-existent user or a wrong password.


server/test/unit/com/thoughtworks/go/server/security/LdapUserSearchTest.java
```diff
@@ -102,8 +103,8 @@ public void shouldThrowUserNameNotFoundExceptionWhenNoUserFound_WithMultipleSear
         try {
             spy.searchForUser("username");
             fail("Should have thrown up");
-        } catch (UsernameNotFoundException e) {
-            assertThat(e.getMessage(), is("User username not found in directory."));
+        } catch (BadCredentialsException e) {
+            assertThat(e.getMessage(), is("Bad credentials"));
         }
```
zabil: We must start using http://junit.org/apidocs/org/junit/rules/ExpectedException.html
For asserting exceptions.

Issue #1743, opened 2015-12-23
File handle leak that causes windows agents to not cleanup properly


ThomasSkelgaard:
After upgrading to 15.3 the test artifacts from working dir on agent is locker after stage is finish. The next stage will then fail as it is not able to clean working directory.

Sometime the files are unlocked after a long time or the Go Agent needs to be re-started to free the filelock.

Is this a known issue?

Stage that complete but seems to lock file
![gostagethatlock](https://cloud.githubusercontent.com/assets/6838005/11975183/644638c6-a96b-11e5-873b-9d287f4bc916.jpg)

The stage right after that cannot do a clean
![go](https://cloud.githubusercontent.com/assets/6838005/11975137/0d942f74-a96b-11e5-8dc9-38e407d38f47.jpg)

ketan:
@ThomasSkelgaard - we seem to think this issue is fixed by #1744. We are treating this as high priority enough to roll out a 15.3.1 release.

jyotisingh:
Was able to reproduce it on 15.3.0 with the below config and a single agent running both stages:
``` xml
<pipelines group="defaultGroup">
    <pipeline name="tesy">
      <materials>
        <git url="C:\test3" />
      </materials>
      <stage name="defaultStage">
        <jobs>
          <job name="defaultJob">
            <tasks>
              <exec command="copy">
                <arg>c:\a.xml</arg>
                <arg>a.xml</arg>
                <runif status="passed" />
              </exec>
              <exec command="ls">
                <arg>-l</arg>
                <runif status="passed" />
              </exec>
            </tasks>
            <artifacts>
              <test src="a.xml" />
            </artifacts>
          </job>
        </jobs>
      </stage>
      <stage name="second">
        <jobs>
          <job name="job">
            <tasks>
              <fetchartifact pipeline="" stage="defaultStage" job="defaultJob" srcfile="a.xml">
                <runif status="passed" />
              </fetchartifact>
              <exec command="ls">
                <arg>-l</arg>
                <runif status="passed" />
              </exec>
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </pipelines>
```
The second stage fails with the following error while cleaning up git repo:
```
05:57:59.101 [GIT] Cleaning all unversioned files in working copy
05:57:59.226 Error performing command: --- Command ---
git clean -dff
--- Environment ---
{}
--- INPUT ----


--OUTPUT ---
Removing cruise-output/
--- ERROR ---
[...]

PR #1744, opened 2015-12-23
Fix a file handle leak (#1743)

ketan:
This was caused by a combination of #1707 (which added a InputStream.close() at the wrong method)
and #1709 (which removed the InputStream.close()) with the assumption that #1707 did the right thing.


ketan:
Closing in favor of #1745

xli:
@ketan Sorry for confusion. But if you just look at changes in #1707 (don't combine it with any other changes), so it was not fixing wrong method, and it was fixing the problem I described.

PR #1751, opened 2016-01-04
Fixing tests that fail across JDK versions

zabil:
* Assert json on contents
* Fix setup for plugin notification service as the second plugin is not mocked on tests related to plugin1


zabil:
#1086

Issue #1646, opened 2015-11-25
Do not allow blank users in `<authorization/>`


ketan:
This should not be valid -

```xml
          <authorization>
            <user />
          </authorization>
```

rajiesh:
verified as part of upgrade testing on 16.2.0-2996 (d4db1b5d626eee6065152dd980bb919d8a1cfd10). Also verified new config saves with empty users/roles working as expected

PR #1754, opened 2016-01-05
Added validation and corresponding tests for user cannot be null (#1646)

varshavaradarajan:


varshavaradarajan:
This pertains to issue #1646

config/config-api/src/com/thoughtworks/go/config/AdminUser.java
```diff
@@ -35,6 +36,9 @@ public AdminUser(final CaseInsensitiveString name) {
     }
 
     public void validate(ValidationContext validationContext) {
+        if(name == null || StringUtil.isBlank(CaseInsensitiveString.str(name)))
```
ketan: There's already a `CaseInsensitiveString#isBlank` that you could use instead of pulling out the raw string.

ketan:
@jyotisingh — how do we handle existing configs with blank `<user/>` in them? I guess we could do a migration to drop such users from the config, so it does not error out.

Question is — does the first time migration happen before config validation or after.

How else can(should?) we be handling this?

jyotisingh:
@ketan - XSL migration is the best way to handle this. Its invoked during the [server startup](https://github.com/gocd/gocd/blob/3e0538d574664b4829f31e45aac118042281d170/server/src/com/thoughtworks/go/server/initializers/ApplicationInitializer.java#L98) and would be the first thing to be run on the config before deserialisation, validation etc.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1115,6 +1116,73 @@ public void shouldTrimLeadingAndTrailingWhitespaceFromCommandsInTemplates_asPart
...
+                +"    </authorization>"
+                +"    <pipeline name='Test' template='test_template'>"
+                +"      <materials>"
+                +"        <hg url='../manual-testing/ant_hg/dummy' />"
+                +"      </materials>"
+                +"     </pipeline>"
+                +"  </pipelines>"
+                +"</cruise>";
+        String migratedXml = migrateXmlString(configXml, 76);
+        assertThat(StringUtils.countMatches(migratedXml, "user"), is(2));
```
ketan: Should this not be `1` user? I'd think `<user>    </user>` should be treated the same as `<user`>`

varshavaradarajan:
Updated PR with newly added configuration migration.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1115,6 +1116,73 @@ public void shouldTrimLeadingAndTrailingWhitespaceFromCommandsInTemplates_asPart
...
+                +"      <materials>"
+                +"        <hg url='../manual-testing/ant_hg/dummy' />"
+                +"      </materials>"
+                +"     </pipeline>"
+                +"  </pipelines>"
+                +"</cruise>";
+        String migratedXml = migrateXmlString(configXml, 77);
+        assertThat(migratedXml, not(containsString("user")));
+        assertThat(migratedXml, not(containsString("view")));
+        assertThat(migratedXml, not(containsString("authorization")));
```
ketan: Can you change these strings to xml tags `<user>` ?

varshavaradarajan:
Updated PR with the changes.

config/config-server/resources/upgrades/78.xsl
```diff
@@ -0,0 +1,30 @@
...
+        <xsl:attribute name="schemaVersion">78</xsl:attribute>
+    </xsl:template>
+    <!-- Copy everything -->
+    <xsl:template match="@*|node()">
+        <xsl:copy>
+            <xsl:apply-templates select="@*|node()"/>
+        </xsl:copy>
+    </xsl:template>
+    <!-- Remove empty nodes recursively -->
+    <xsl:template match="*[not(.//@*) and not(normalize-space())]" />
```
ketan: Hey @jyotisingh - I missed this a bit — but should we be really stripping out all nodes or just a specific node matched by a specific xpath?

ketan:
@varshavaradarajan — I've had to revert your change. See https://build.go.cd/go/pipelines/build-linux/717/build-non-server/1 and my comment about for why.

PR #1755, opened 2016-01-05
Rename the `command` attribute of `<exec/>` element to `executable` to reduce ambiguity

ketan:


jyotisingh:
@ketan - These changes look fine to me. Apart from these, [this](https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/config/config-api/src/com/thoughtworks/go/config/ExecTask.java#L43) and [this](https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/server/webapp/localize.xml#L311) along with their usages needs to be changed as well which will take care calling it out as "Executable" instead of "Command" in the UI.

The only thing that would be left out is [command-repo](https://github.com/gocd/go-command-repo) which would continue to use the older-terminology. While we can fix the scripts in the repo pointed out earlier, the forks will break. I am fine leaving all the command-repo scripts as is, unless others think otherwise. 

If you end up renaming the html elements, just make sure that selecting the element from command-repo dropdown populates the `executable` box correctly.

ketan:
> @ketan - These changes look fine to me. Apart from these, [this](https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/config/config-api/src/com/thoughtworks/go/config/ExecTask.java#L43) and [this](https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/server/webapp/localize.xml#L311) along with their usages needs to be changed as well which will take care calling it out as "Executable" instead of "Command" in the UI.

That constant was exactly what I'm afraid of changing — that is used all over the rails codebase to name fields  and I fear that it'll break a lot of things that we don't have bandwidth to test.

PR #1756, opened 2016-01-05
Maintain the order of group creation while merging pipeline configs.

zabil:
@tomzo 
This is for fixing failures 
https://build.go.cd/go/tab/build/detail/build-linux/713/build-non-server/1/common-split-runInstance-8

I see that you've added these tests and they behave differently across jdk's. Are you sure this is the behavior you were looking for? 

coz 

https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java#L832

gets fudged here

https://github.com/gocd/gocd/blob/fd8a70ff3bdbb9e7e9c5f9ffec05d3ed2686bf75/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java#L312


tomzo:
This totally makes sense. I only tested on openjdk 7 so I would not know. Thanks for the fix.

zabil:
#1086

Issue #1724, opened 2015-12-17
[PipelineConfig] Materials


maheshp:
Acceptance Criteria
- [x] List existing materials
- [ ] Add different types of material
  - [x]  TFS
    - [x] Test Connection #1493
  - [x] Git
    - [x] Test Connection
    - [x] Set default branch to master
    - [x] Support shallow clone
  - [x]  Svn
    - [x] Test Connection
  - [x]  Mercurial
    - [x] Test Connection
  - [x]  Perforce
    - [ ] Make view field to multiline.
    - [x] Test Connection
  - [x]  Pipeline
    - [ ] Visualise pipeline dependency #1494 : Part of Fetch Artifact task
  - [ ] Package
    - [ ] Create API end point for Package Material #1507
  - [x] PluggableSCM
    - [x] Create API end point for Pluggable SCM Material #1508
    - [x] Build view to support Pluggable SCm #2488
- [x] Delete Material
- [ ] Validation and error messages
  - [x] Validate branch (#1734)
- [x] UX and UI
- [ ] Tool tip for ignore pattern
- [x] Whitelist of SCM materials refer https://github.com/gocd/gocd/pull/2144

zabil:
For test connection check if #1734 is still an issue

juhijariwala:
#1734 is still an issue.

arvindsv:
Closing. Will have to open new issues for a rewrite.

Issue #1493, opened 2015-09-16
Test connection functionality for materials


zabil:
A sub-task of #1404 

Design and implement a way to test validating of a material configuration before adding and saving it to a pipeline configuration.

ketan:
I looked at the source, and this seems a lot simpler to achieve once #1491 is merged.

I'm shelving this for now.

arvindsv:
This has been there for a while now.

PR #1763, opened 2016-01-07
Can test-connection for a material #1724 #1493

maheshp:


server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/materials.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-define(['mithril', 'lodash', 'string-plus', './model_mixins', './encrypted_value'], function (m, _, s, Mixins, EncryptedValue) {
+define(['mithril', 'lodash', 'string-plus', 'jquery',  './model_mixins', './encrypted_value', 'js-routes'], function (m, _, s, $, Mixins, EncryptedValue, Routes) {
```
zabil: I think we should use m.request
Here's pointers on how to make it testable
http://stackoverflow.com/questions/26373734/how-can-i-unit-test-a-model-using-m-request-in-mithril

server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/materials_config_widget_spec.js
```diff
@@ -66,7 +108,20 @@ define(["jquery", "mithril", "pipeline_configs/models/materials", "pipeline_conf
...
+      mount(materials);
+      viewMaterial();
+
+      verify(perforceMaterial);
+      expect($root.find("input[data-prop-name='port']").val()).toBe(perforceMaterial.port());
+      expect($root.find("input[data-prop-name='username']").val()).toBe(perforceMaterial.username());
+      expect($root.find("input[data-prop-name='view']").val()).toBe(perforceMaterial.view());
+      expect($root.find("input[data-prop-name='passwordValue']").val()).toBe('p@ssw0rd');
+      expect($root.find("input[data-prop-name='useTickets']").val()).toBe('on');
+    });
```
zabil: I think we should avoid clubbing all the expectations together. Failures become notoriously hard to locate without good descriptions.
You can nest the describes

```
describe("mercurial component", function () {
 beforeEach() {
  mount(materials);
 }
 describe("model binding" , function(){
    it("should bind url" , function(){
       expect($root.find("input[data-prop-name='url']").val()).toBe(mercurialMaterial.url());
     });
   });
}
```

http://jasmine.github.io/2.0/introduction.html#section-Nesting_<code>describe</code>_Blocks

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/materials.js
```diff
@@ -69,12 +69,11 @@ define(['mithril', 'lodash', 'string-plus', './model_mixins', './encrypted_value
...
-    this.parent = Mixins.GetterSetter();
-
-    this.type       = m.prop(type);
-    this.name       = m.prop(s.defaultToIfBlank(data.name, ''));
-    this.autoUpdate = m.prop(data.autoUpdate);
+    this.parent           = Mixins.GetterSetter();
+    this.type             = m.prop(type);
+    this.name             = m.prop(s.defaultToIfBlank(data.name, ''));
+    this.autoUpdate       = m.prop(data.autoUpdate);
+    this.testConnectionVM = m.prop(new Materials.Material.vm.testConnection());
```
zabil: After a long discussion, we've concluded that this view model should be pulled out of materials because it's better for testability and re-usability to render check connection as a separate component. The state must not be maintained in the actual material as testing a connection is a view only concern.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/materials.js
```diff
@@ -107,6 +105,52 @@ define(['mithril', 'lodash', 'string-plus', './model_mixins', './encrypted_value
...
     };
 
+    this._passwordHash = function() {
+      if (this.isPlainPasswordValue() || this.isDirtyPasswordValue()) {
+        return { password: this.passwordValue() };
+      }
+      return { encryptedPassword: this.passwordValue() };
+    };
+
+    this.testConnection = function (vm, pipelineName) {
```
zabil: This must be pulled out of here as 
* model should not be handing anything related to the view
* this need not be an instance method. Rather can take in all the parameters and return a promise (if done via m.request)

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/helpers/form_helper.js.msx
```diff
@@ -438,6 +438,17 @@ define(['mithril', 'lodash', 'string-plus', '../models/model_mixins', 'foundatio
...
       }
+    },
+
+    alert: {
+      view: function (ctrl, args, children) {
+        return (
+          <div>
+            <span data-alert class="alert-box secondary">{args.message}</span>
+            {children}
+          </div>
```
ketan: I'd suggested this earlier

```
return (
<div>
 <span data-alert class="alert-box secondary">{children}</span>
</div>
)
```

maheshp: Ok, we got it wrong. We will change it.

PR #1775, opened 2016-01-08
 Added validation for users cannot be blank (#1646)

varshavaradarajan:
* Added a new migration to remove existing blank users

### DIFF
--- a/base/src/com/thoughtworks/go/util/GoConstants.java
+++ b/base/src/com/thoughtworks/go/util/GoConstants.java
@@ -58,7 +58,7 @@ public class GoConstants {
 
     public static final String PRODUCT_NAME = "go";
 
-    public static final int CONFIG_SCHEMA_VERSION = 77;
+    public static final int CONFIG_SCHEMA_VERSION = 78;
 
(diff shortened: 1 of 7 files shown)


varshavaradarajan:
Updated PR adding all cases for user to be blank.

varshavaradarajan:
Have to update copyright in one of the files. Will do so after the build-linux-PR finishes building.

ketan:
@varshavaradarajan — is this ready to review?

varshavaradarajan:
Yes.

ketan:
@arvindsv @jyotisingh — I've merged this so we can run the extensive set of smokes and regression. But I'm left to wonder if we're breaking compatibility [here](https://github.com/gocd/gocd/pull/1775/files#diff-dea6f2dfc75a0ab60ebc7dbb6c7f51bcR1166).

arvindsv:
I don't think I have full context on this. Why would removing empty tags break compatibility? I mean, we don't really remove any information, right?

varshavaradarajan:
It removes only the empty tags within <authorization>.

PR #1789, opened 2016-01-12
Fixed test_connection_widget specs

maheshp:
* innerText does not work on FF, changed to text()

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/test_connection_widget_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/test_connection_widget_spec.js
@@ -36,7 +36,7 @@ define(["mithril", "lodash", "pipeline_configs/models/materials", "pipeline_conf
       it("should render test connection button", function () {
         mount(material);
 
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/test_connection_widget_spec.js
```diff
@@ -46,7 +46,7 @@ define(["mithril", "lodash", "pipeline_configs/models/materials", "pipeline_conf
 
         mount(material, {connectionState: state});
 
-        expect($root.find(".alert-box")[0].innerText).toBe('Test Connection Failed');
+        expect($($root.find(".alert-box")[0]).text()).toBe('Test Connection Failed');
```
ketan: You can just do `$root.find().text()` instead

ketan: See https://github.com/velesin/jasmine-jquery for how to write improved assertions.

maheshp: tried $root.find().text(), but that does not work

PR #1791, opened 2016-01-12
#000 - Add missing logging for config save

arvindsv:
Perf problems around the area always end up requiring more logging.

### DIFF
--- a/server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
+++ b/server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
@@ -16,9 +16,6 @@
 
 package com.thoughtworks.go.config;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import com.thoughtworks.go.config.validation.GoConfigValidity;
 import com.thoughtworks.go.listener.ConfigChangedListener;
 import com.thoughtworks.go.listener.PipelineConfigChangedListener;
@@ -30,6 +27,9 @@ import com.thoughtworks.go.serverhealth.ServerHealthState;
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
 
 import static com.thoughtworks.go.server.service.GoConfigService.INVALID_CRUISE_CONFIG_XML;
 import static com.thoughtworks.go.util.ExceptionUtils.bomb;
@@ -94,7 +94,9 @@ public class CachedFileGoConfig implements CachedGoConfig {
 
     private void loadFromDisk() {
         try {
+            LOGGER.debug("Config file (on disk) update check is in queue");
             synchronized (GoConfigWriteLock.class) {
+                LOGGER.debug("Config file (on disk) update check is in progress");
                 GoConfigHolder configHolder = dataSource.load();
                 if (configHolder != null) {
                     saveValidConfigToCacheAndNotifyConfigChangeListeners(configHolder);
@@ -230,9 +232,11 @@ public class CachedFileGoConfig implements CachedGoConfig {
         LOGGER.info("About to notify config listeners");
         for (ConfigChangedListener listener : listeners) {
             try {
+                long startTime = System.currentTimeMillis();
                 listener.onConfigChange(newCruiseConfig);
+                LOGGER.debug("Notifying " + listener.getClass() + " took (in ms): " + (System.currentTimeMillis() - startTime));
             } catch (Exception e) {
-                LOGGER.error("failed to fire config changed event for listener: " + listener, e);
+                LOGGER.error("Failed to fire config changed event for listener: " + listener, e);
             }
         }
         LOGGER.info("Finished notifying all listeners");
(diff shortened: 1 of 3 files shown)


server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
```diff
@@ -230,9 +232,11 @@ private synchronized void notifyListeners(CruiseConfig newCruiseConfig) {
         LOGGER.info("About to notify config listeners");
         for (ConfigChangedListener listener : listeners) {
             try {
+                long startTime = System.currentTimeMillis();
                 listener.onConfigChange(newCruiseConfig);
+                LOGGER.debug("Notifying " + listener.getClass() + " took (in ms): " + (System.currentTimeMillis() - startTime));
```
zabil: Wrap this up in a ```isDebugEnabled()``` ? 
Also ```String.format```?

Or better use slf4j instead like it's done on ```GoConfigDao```

arvindsv:
@zabil: Switched to slf4j there.

server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
```diff
@@ -230,9 +235,11 @@ private synchronized void notifyListeners(CruiseConfig newCruiseConfig) {
         LOGGER.info("About to notify config listeners");
         for (ConfigChangedListener listener : listeners) {
             try {
+                long startTime = System.currentTimeMillis();
                 listener.onConfigChange(newCruiseConfig);
+                LOGGER.debug("Notifying {} took (in ms): {}", listener.getClass(), (System.currentTimeMillis() - startTime));
             } catch (Exception e) {
-                LOGGER.error("failed to fire config changed event for listener: " + listener, e);
+                LOGGER.error("Failed to fire config changed event for listener: " + listener, e);
```
arvindsv: Since this is logging an exception, not using the `{}` syntax. Also, don't think this happens often enough for `String.format` to make any real difference.

server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
```diff
@@ -160,7 +163,9 @@ private void saveValidConfigToCacheAndNotifyPipelineConfigChangeListeners(Cached
         for (ConfigChangedListener listener : listeners) {
             if(listener instanceof PipelineConfigChangedListener){
                 try {
+                    long startTime = System.currentTimeMillis();
                     ((PipelineConfigChangedListener) listener).onPipelineConfigChange(saveResult.getPipelineConfig(), saveResult.getGroup());
+                    LOGGER.debug("Notifying {} took (in ms): {}", listener.getClass(), (System.currentTimeMillis() - startTime));
                 } catch (Exception e) {
                     LOGGER.error("failed to fire config changed event for listener: " + listener, e);
```
zabil: failed -> Failed 

for consistency sake.

server/src/com/thoughtworks/go/config/MergedGoConfig.java
```diff
@@ -101,7 +102,7 @@ public void tryAssembleMergedConfig(GoConfigHolder cruiseConfigHolder,List<Parti
             // save to cache and fire event
             this.saveValidConfigToCacheAndNotifyConfigChangeListeners(newConfigHolder);
         } catch (Exception e) {
-            LOGGER.error(String.format("Failed validation of merged configuration: %s", e));
+            LOGGER.error("Failed validation of merged configuration: {}", e);
```
zabil: There is no parameter to be substituted here. Don't if it bombs when that happens.
This will use the overridden method
```
 public void error(String msg, Throwable t);
```
Right?

arvindsv: It won't fail. But, it's not ideal. I'm changing it. Here's what happens:

```java
@Test
public void shouldLogException() throws Exception {
    Logger LOGGER = LoggerFactory.getLogger(ApprovalTest.class);
    RuntimeException e = new RuntimeException("Hello");
    LOGGER.error("=== Message 1: {} - {} _-_", e.getClass(), e.getMessage());
    LOGGER.error("--- Message 2: {} +++", e);
}
```

```
<=== Message 1: class java.lang.RuntimeException - Hello _-_>

<--- Message 2: {} +++>
java.lang.RuntimeException: Hello
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:606)
  at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
  ...
```

As you can see, slf4j will just print extra `{}` arguments, if there's nothing which matches and the exception shows fully. The earlier message seemed to just want to call `e.toString()` and not show the stack trace. I'll just do that too.

server/src/com/thoughtworks/go/config/CachedFileGoConfig.java
arvindsv: Ok, will change.

arvindsv:
Done.

Issue #1125, opened 2015-05-12
Delay between stages/jobs


hilverd:
Our teams are trying to reduce the overall time taken by pipelines. One of the things we have noticed is that there is a delay of approximately 20--30 seconds between stages. Some of our pipelines have up to 11 stages, so that adds up.

For example, consider this pipeline (cut from config XML):

```
<pipeline name="BetweenStagesDelay">
    <materials>
        <git url="https://github.com/gocd/gocd.git"/>
    </materials>
    <stage name="first">
        <jobs>
            <job name="j">
                <tasks>
                    <exec command="date"/>
                </tasks>
            </job>
        </jobs>
    </stage>
    <stage name="second">
        <jobs>
            <job name="j">
                <tasks>
                    <exec command="date"/>
                </tasks>
            </job>
        </jobs>
    </stage>
    <stage name="third">
        <jobs>
            <job name="j">
                <tasks>
                    <exec command="date"/>
                </tasks>
            </job>
        </jobs>
    </stage>
</pipeline>
```

If you trigger this pipeline, you'll see that there is approximately 20--30 seconds delay between stages "first" and "second" and between stages "second" and "third", even if there are no other pipelines defined and there are agents available.

From our run of this pipeline just now, the output from the job in each of these stages was:

```
Mon May 11 09:28:22 UTC 2015
Mon May 11 09:28:42 UTC 2015
Mon May 11 09:29:02 UTC 2015
```

Can this delay be reduced?

This was originally raised as request #23913 in the ThoughtWorks Knowledge Base, but I was advised to raise it here instead.

FredrikWendt:
+1

pmcgrath:
+1

lyptt:
+1

dsenecalb:
+1

dsenecalb:
What's the status of for the ticket. We are looking forward that functionality that allows admin to increase/reduce the delay timing

scottmuc:
+1

richard-hulm:
+1

mgreensmith:
+1

xli:
I'm looking into this.

xli:
Summary of why it happens:
* Overview of how agent and server works on scheduled job assignment:
  * Agent calls go-server every 10 sec to get new work when it is idling or completes a job.
  * Go server triggers job assignment only when an agent called and no work is assigned to agent.
  * Go server's work assignment is an asynchronous process.
* Why 20 seconds delay:
[...]

Issue #1098, opened 2015-05-05
[performance] Agent improvements


srinivasupadhya:
Improve Agent Server communication to reduce the problem of too many agents pinging the server at same time during start up & too many requests (1 ping + 1 work request = 2 requests / sec / agent) from agents.
- Thundering herd - implement exponential backoff while registering so that Server is not swamped with requests.
- Websocket - move from polling to server push model

greenmoss:
+1 on this! Rationale: every pipeline stage that is pending, must wait for an agent to check in. It seems the polling time is about one minute (not sure about this). So for a pipeline with 5 stages, the polling overhead can add up to 5 excess minutes to the pipeline's completion time. It would be great to get rid of that execution time overhead!

kmugrage:
+1 on the Websocket ( or similar ) change in model. The delay between a material change whether code or another pipeline is quickly becoming my biggest user experience issue. It's quite painful when modelling new stuff.

lenucksi:
Especially together with the configuration saving/validation improvements mentioned in #1498 this could really help to mitigate load and stability issues on large Go servers.

xli:
Spiking on websocket solution, talk to me if you did something related :)

ketan:
If you are :+1

xli:
My plan
--------------------

* Upgrade Spring to 4.3.3 (latest):  not full upgrade for spike, just do enough work for spike.
* Use Spring 4's WebSocket support to implement server side communication.
* Use Jetty's WebSocket client for Agent implementation.

Other options in consideration:
* Bypass Spring to use Jetty's WebSocket. I feel there maybe headache of how to make it work with existing stuff that is integrated with Spring. So I think it is better to impl in Spring framework if we impl it in Java, at least try using Spring framework's websocket first.
* Rails websocket: Rails 5 will have websocket support. For Rails 4, there are some gems support it (em-websocket, websocket-rails, tubesock), but no active maintainer. So I don't feel strong confidence to use any existing Websocket impl in Rails. Plus I think most of people will choose to support Rails' ActionCable for websocket. Hence most of other websocket impl on rails will die eventually. 
* Ruby websocket server through rack: https://github.com/faye/faye-websocket-ruby (I found this today, seems good, will do more research on it later)

My spike work is pushed to: https://github.com/xli/gocd/tree/websocketspike

Progress
---------------
[...]

Issue #1595, opened 2015-11-04
Make the error notification more visible


iliasbartolini:
From various colleagues I collected feedback that the important error notifications are not visible enough.
![screenshot from 2015-11-04 16 26 31](https://cloud.githubusercontent.com/assets/186468/10942630/354209a4-8311-11e5-9592-98b69897b510.png)

1) One first and very simple change would be to make this more visible.

2) The second proposal is that actually many of this error could be contextual to a specific pipeline/stage/job and would make more sense to see that one failing or having the error visible in that context. 
This speciffic example could make the first stage in the ABC pipeline to fail:
```Error while scheduling pipeline: Deploy-Dashing as no compatible revisions was identified. [Nov-04 15:31:14]
Failed resolution of pipeline ABC : Cause : No valid revisions found for the upstream dependency: DependencyMaterialConfig{pipelineName='DEF', stageName='SomeStage'}```

xli:
Looking into this issue. Thinking about group errors/warnings by pipeline and the show little icon to indicate the errors/warnings:

![pipelines_-_go](https://cloud.githubusercontent.com/assets/17474/12527465/a2355fe0-c130-11e5-9683-963b86c33254.jpg)

After clicked the icon, it should show a popup like current right bottom errors did, but only show errors/warnings related to the pipeline.

Current right bottom errors will still be there for global errors and list all errors

arvindsv:
I'm proposing a full rewrite of the dashboard with much better performance. Interested? It should include this and more (single click to reach the console of a failing job and more obvious access to the VSM, etc).

Ignoring what I said above for a moment and focusing on what you brought up: Yes, that makes sense. Of course, because a material can be shared by multiple pipelines, the same error can show up on different pipelines. Might be irritating for someone who is trying to fix all errors.

Second: I don't think the information sent back from the server is rich enough to figure out the pipelines affected by it. It needs us to move to the dashboard JSON API. That's what I want to replace with a more performant one as I mentioned above.

xli:
As long as you're ok with the user behaviors described above, I'll check out how to implement it. It's a chance to learn more about this part code.

arvindsv:
[...]

PR #1793, opened 2016-01-12
Websocket communication (#1098)

xli:
Implements websocket communication between agent and server (#1098 and #1125).

1. Use Jetty websocket in both client and server side
2. Server side push:
    1. new job
    2. set cookie
    3. cancel job
3. Agent will still ping server with agent runtime info periodically for:
    1. keep connection alive.
    2. robustness
4. A toggle (system environment property) in Agent side for switching to use websocket communication.
This toggle is designed for testing and continuous improvement before we can get ride of current spring RPC implementation.
5. More details and some design decisions see the following discussions
6. Message format is using Java serialization due to the complexity of BuildWork object.

---

**Edit:** (copying the design decisions here in case @xli ever removes his fork in the future) 

---

### Overview

When I started working on this PR, I did some analysis about how existing PRC agent works and how I'd like websocket version works, and captured here: https://github.com/xli/asset/blob/master/go-cd-agent-design-sequence-diagrams.JPG

It should give you an overview of how the system works.
Comparing with my final implementation, it only has a little bit different in details (several events order, and missing how reschedule job works).

### How server push job to agent?

* BuildAssignmentServer#onTimer triggers job plans refresh, then start to find matching job for idling agents that is registered by websocket. 
* Assign work to agent message will be sent asynchronously, thus it won't block #onTimer method.
* Server won't push blank work (NoWork) to agent (PRC behavior).

### How unregistered and disabled agents are handled on server?

1. Server always accepts agent websocket session no matter what agent status is when agent pings server. Hence an unregistered agent will be accepted. This is same logic with RPC version that will accept "getWork" call.
2. Server will send reregister message to agent if it found agent was not registered during the process of matching agents with jobs (similar with current RPC behavior). We don't handle it when websocket session is opened, because agent may become unregistered status if user deleted agent from agent list. So handle it in one place is robust enough and simpler.
3. However, we won't do anything to disabled agents. (agent will get NoWork in current RPC version)
4. Do we really need the complicated SSL certificate management we do today?
  * I try to avoid this discussion by implementing almost exactly same behavior, because of my lag knowledge of these code

### Why Server sends message to Agent in asynchronous mode? And how we'll handle asynchronous sending error?

This is designed for speed up all of processing with agent, especially we don't want to block and fail the process of matching agents with jobs. 

We rely on the way we handle websocket connection and rescheduling hanging jobs to handle sending message IO failures, which may need more test to find out how likely it will happen. 

The sending message failure will be logged by WebSocket onError method.

See question of ["What we do for handling websocket error?"](#what-we-do-for-handling-websocket-error-on-server) for more details of why we can ignore asynch sending message failure.

### When server found an agent lost connection

Server will mark this agent as lost contact and remove it from the agent list for matching agents and jobs.
This may cause server shows agent status "lost contact" temporary if the agent reconnects to server immediately.

### What happened when agent lost websocket connection and is building something?

* Server(AgentRemoteSocket) will receive connection close event and remove websocket session, then mark agent lost contact
* ScheduleService#rescheduleHungJobs (about 5 mins) will reschedule job if agent didn't reconnect.
* Agent reconnect to server building status, and it is still building the original job, if agent didn't restart. The running job should be rescheduled by ScheduleService#rescheduleHungJobs
* If Agent restarted, agent will connect to server with idle status, which will update server side agent status. The running job should be rescheduled when reconnected agent is assigned with a new job.

### What we do for handling websocket error on Server?

We log error and do nothing else. Our other parts of system will handle cases caused by these errors.
We'll go through all possible errors and verify it.

* There are 3 cases for sending messages:
   1. send "cancel", "setCookie": these 2 actions are triggered when agent pings server. Agent pings server periodically, hence these action will be triggered again and safe to ignore error.
   2. send "reregister": server does match agent with jobs periodically, hence this action will be triggered again.
   3. send "assignWork": when this message failed, agent status on server is changed to building and agent status on client is still idle. Hence when agent pings server, agent will be updated to idle status, which is similar case with "when agent is restarted (reconnected to server) just after assigned a new work" (see question [What happened when agent lost websocket connection and is building something?](#what-happened-when-agent-lost-websocket-connection-and-is-building-something) for details).
* Connection idle timeout:
   1. This error will cause connection closed, and mark agent lost contact. And it is same case with Agent stopped if this agent never connected again, or restarted if this agent reconnected later.
   2. Default connection timeout is 1 minute, and it can be changed by system property "go.websocket.max.idle.time".
   3. Agent should ping server within 1 minute. It is 10 seconds right now due to RPC getWork periodical interval setting, we can discuss it later to find a better interval.
* Receive message error: this is agent/client problem, it should not cause any problem as ping failed. And ping is a periodical task, hence it will be recovered if it was temp problem, and it's fine to ignore this error.
* Decode message error: this is server side bug, or mismatch encoding problem between agent and server. It causes agent is never added into waiting new job list (because ping failed), hence nothing we can do to recover, log error.
* Process message error: server runtime error, nothing we can do to recover, log error.

### What we do for handling websocket error on Client (Agent)?

* We do synch sending message, hence any error when sending message will cause error and stop current thread.
* It basically should be similar behavior with RPC version
* log error
* when ping failed, it's fine to ignore it, because ping itself is a periodical task
* when reportCurrentStatus, reportCompleted, or reportCompleting failed during running job, it will fail the job running thread and agent will be set to idle status

### How we convert BuildRepositoryRemote#isIgnored(JobIdentifier jobIdentifier)?

* Is ignored will check given job status, if it is canceled or rescheduled then return true.
* So, new implementation will change to monitor JobRunner isCancelled status.
* Then server will push Cancel message to agent if job is cancelled or rescheduled.
* The worst case is agent run job finished because cancel failed or before agent received cancel message, which in theory also happens in RPC mode.

### Multiple agents at same machine

* No extra configuration needed, tested

### Toggle for turn on agent websocket communication

There is no toggle on Go server, always available.
Go Agent: add GO_AGENT_SYSTEM_PROPERTIES: -Dgo.agent.websocket.enabled=Y

### Other system environment variables:

Server and Agent should use same configuration:
* go.websocket.max.idle.time, default 60 seconds
* go.websocket.max.text.message.size, default 128K

# Test

* ✓ multiple agents on same machine
* ✓ upgrade from RPC agent to websocket agent (enabled by default)
* ✓ Server behind a http proxy: agent will fetch ssl port from server, which means, there is no way to configure agent connect to proxy server port, so if there is proxy server between agent and server, it need act exactly same with server including open same ssl port. Then we need export Go server's ssl certificates, I was blocked here because I could not figure out what's keystore's password :(. Anyway, it seems we don't expect anyone has proxy between server and agent, so I'll stop here
* ✓ No test failure if we turned on go.agent.websocket.enabled by default
* AC/smoke test suite


ketan:
A quick heads up - suggest using slf4j instead of log4j for all logging, since we are planning on moving away from a log4j.

xli:
@ketan sure, will do it

ketan:
@xli — I've copied you design decisions page here for sake of future archeologists who may come to this page. You should be able to edit the first comment, if not let me know.

arvindsv:
One thing about logging errors: Whenever these errors are tied to retries from the agent side, it can cause the server logs to fill up, if an error is not something which will be fixed by itself. Just a general point to keep in mind.

arvindsv:
One more thought: If you look at jetty.xml, there are a max of 300 qtp threads configured. What happens to long-lived connections with websockets? If I have 100 agents, will 100 qtp threads be constantly active, or are those 100 tcp connections and the qtp threads are only on demand? Might be relevant for performance.

Currently, it's like the latter, I believe.

xli:
@arvindsv,

Error logging
==============
Throwing away an error without log seems a problem. Talked with @wpc, he suggested install a circuit breaker for agent ping loop. I think that's a good idea, because fundamentally agent should not keep retry so frequently when there is error.

Jetty Thread Pool
================
I confirmed threads will be shared for processing agent messages.

xli:
Not sure why build go-sdk failed, failed at "cp -v target/go-plugin-api-16.2.0-${GO_DEPENDENCY_LABEL_CREATE_MAVEN_RELEASE}.jar target/go-plugin-api-16.2.0.jar'", it looks like an old problem. anyone know what's wrong? I can see the jar is generated successfully in the log (although no log of file created)

ketan:
> Not sure why build go-sdk failed, failed at "cp -v target/go-plugin-api-16.2.0-${GO_DEPENDENCY_LABEL_CREATE_MAVEN_RELEASE}.jar target/go-plugin-api-16.2.0.jar'", it looks like an old problem. anyone know what's wrong? I can see the jar is generated successfully in the log (although no log of file created)

@xli — this is because the various `pom.xml` files in your fork still have 16.1.0 in it. Can you rebase your PR on top of current master?

arvindsv:
@xli:

> About circuit breaker:

Some retrying might be helpful (slowed down, maybe). Logging without stacktrace after some time might be helpful too. Whatever approach is chosen, a user needs to be able to figure out that there's a problem. Even if the circuit breaker decides to stop any further calls, I guess it should log the fact that it did that. :)

> I confirmed threads will be shared for processing agent messages.

Thank you! That's good.

xli:
@arvindsv circuit breaker will be at client side with log, and if you'd like, you can even have fancy monitoring dashboard (netflix Hystrix). 
Is there anything in Go server handles log like you said (a way to not keep logging errors when agent and server communication has problem) ?

arvindsv:
Not yet. We need that for MDU too. If a material has a problem, the logs get written for it every 60 to 90 seconds.

xli:
I wish Go could display an error tag at the pipeline (or the stage) on dashboard, so that I know there is something wrong :)

arvindsv:
[Yes!](https://github.com/gocd/gocd/issues/1595)

server/src/com/thoughtworks/go/server/websocket/JobInstanceStatusMonitor.java
```diff
@@ -0,0 +1,40 @@
...
+@Component
+public class JobInstanceStatusMonitor {
+
+    @Autowired
+    public JobInstanceStatusMonitor(final AgentRemoteHandler agentRemoteHandler, JobInstanceService jobInstanceService) {
+        jobInstanceService.registerJobStateChangeListener(new JobStatusListener() {
+            @Override
+            public void jobStatusChanged(JobInstance job) {
+                if (job.isRescheduled() || JobResult.Cancelled.equals(job.getResult())) {
+                    agentRemoteHandler.sendCancelMessage(job.getAgentUuid());
```
jyotisingh: Any specific reason for not sending the cancel message as a part of existing [AgentService#notifyJobCancelledEvent](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/src/com/thoughtworks/go/server/service/AgentService.java#L364)?

agent/src/com/thoughtworks/go/agent/service/AgentWebsocketService.java
```diff
@@ -0,0 +1,212 @@
...
+    @OnWebSocketClose
+    public void onClose(int closeCode, String closeReason) {
+        LOGGER.info(sessionName() + " closed.");
+    }
+
+    @OnWebSocketError
+    public void onError(Throwable error) {
+        LOGGER.error(sessionName() + " error", error);
+        if (error instanceof MessageTooLargeException) {
+            LOGGER.error("You can set Java system property 'go.websocket.max.text.message.size' to increase limit");
```
jyotisingh: The property name does not match the one in SystemEnvironment, may be you could use `SystemEnvironment.GO_WEBSOCKET_MAX_MESSAGE_SIZE.propertyName()`

server/src/com/thoughtworks/go/server/websocket/JobInstanceStatusMonitor.java
xli: The only place calls this method is "JobResultListener" with comment "// TODO - #2511 - this only works for cancelling a job.", then JobResultListener is a GoMessageListener which is part of AMQ infrastructure, which is coming from https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/src/com/thoughtworks/go/server/service/JobInstanceService.java#L235.
On another hand, there is https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/src/com/thoughtworks/go/server/service/JobInstanceService.java#L234 and JobStatusListener infrastructure to notify any listener cancel status changed.

And notice we are not only listening to cancelled status, but also job.isRescheduled() status, hence a JobStatusListener is better choice. 

Then, if it is better choice, why I don't merge them and get rid of JobResultListener and related code, refactoring them to use a JobStatusListener?

The reason is I'm doing Parallel (Make the new design and run both new and old simultaneously.) strategy when I implement websocket communication. see https://pragprog.com/magazines/2009-09/responsive-design for more details.
I want to make sure we have lowest risk to switch to websocket communication in production.

agent/src/com/thoughtworks/go/agent/service/AgentWebsocketService.java
xli: updated

xli:
@jyotisingh @ketan I did rebase with master, and just realized it may cause problem for you to continue to review my changes. So want to let you know there is no changes due to rebase.
I appended 2 changes after rebase: 65ce465 and 14f150c

server/src/com/thoughtworks/go/server/websocket/JobInstanceStatusMonitor.java
jyotisingh: Well, that TODO comment is a bit mis-leading. JobResultListener comes into picture for stage cancellations too, provided the job getting cancelled as a part of the stage has an agent assigned to it. Moreover, the class name JobResultListener itself is mis-leading, but thats the right listener which gets invoked when a job/stage gets cancelled either by a user through UI/API or a job-timeout happens. 

Thanks for pointing it out, I did not pay much attention to the reschedule check earlier. A job reschedule happens when the agent responsible for building the job has gone off the radar. In such a case, `AgentRemoteHandler` would probably not find any matching agents in agentSessions, and so a notification would not be sent anyway.

>> I want to make sure we have lowest risk to switch to (or try out) websocket communication in production.

Fair enough point. May be we should register (ie. `jobInstanceService.registerJobStateChangeListener`) only if websocket communication is turned on.

common/src/com/thoughtworks/go/server/websocket/Message.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public static void setupPolicy(WebSocketPolicy policy, SystemEnvironment environment) {
+        policy.setMaxTextMessageBufferSize(environment.getWebsocketMaxMessageSize());
+        policy.setMaxTextMessageSize(environment.getWebsocketMaxMessageSize());
+        policy.setMaxBinaryMessageSize(environment.getWebsocketMaxMessageSize());
+        policy.setMaxBinaryMessageBufferSize(environment.getWebsocketMaxMessageSize());
+    }
+
+    public static byte[] encode(Message msg) {
+        String encode = JsonMessage.encode(msg);
```
ketan: If we're serializing/de-serializing as JSON, why do we encode things as bytes instead of just sending strings back and forth?

common/src/com/thoughtworks/go/server/websocket/Message.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public static Message decode(InputStream input) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            byte[] buffer = new byte[8192];
+            int len;
+            while((len = input.read(buffer))>0)
+                baos.write(buffer, 0, len);
+            return decode(baos.toByteArray());
```
ketan: If we really need to handle byte-arrays, `IOUtils.toByteArray(InputStream)` provides a convenience for this.

common/src/com/thoughtworks/go/server/websocket/JsonMessage.java
```diff
@@ -0,0 +1,106 @@
...
+    private static Gson getGson() {
+        GsonBuilder builder = new GsonBuilder()
+                .registerTypeAdapter(Builder.class, new ObjectConverter())
+                .registerTypeAdapter(JobPlan.class, new ObjectConverter())
+                .registerTypeAdapter(Material.class, new ObjectConverter())
+                .registerTypeAdapter(Message.class, new MessageConverter())
+                .registerTypeAdapter(FetchHandler.class, new ObjectConverter())
+                .registerTypeAdapter(MaterialInstance.class, new ObjectConverter())
+                .excludeFieldsWithModifiers(Modifier.TRANSIENT, Modifier.STATIC);
+        return builder.create();
```
ketan: Gson can be an expensive object to create — It is threadsafe and is safe to cache at a class level (aka singleton)

agent/src/com/thoughtworks/go/agent/service/AgentWebsocketService.java
```diff
@@ -0,0 +1,212 @@
...
+            session.close();
+        }
+        LOGGER.info("Connecting to websocket endpoint: " + urlService.getAgentRemoteWebSocketUrl());
+        session = client.connect(this, new URI(urlService.getAgentRemoteWebSocketUrl()), new ClientUpgradeRequest()).get();
+    }
+
+    public synchronized void stop() {
+        if (isRunning()) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("close " + sessionName());
```
ketan: With SLF4j — this can be rewritten as `LOGGER.debug("close {}", sessionName())`. The actual string interpolation only happens if the message needs to be logged. You may drop the `isDebugEnabled` as well.

common/src/com/thoughtworks/go/server/websocket/Message.java
ketan: If you're worried about encoding issues then if you look at the implementation of `org.eclipse.jetty.websocket.common.frames.TextFrame`, you'll notice that it is consistently serializing the text using utf8 —


```java
public TextFrame setPayload(String str)
    {
        setPayload(ByteBuffer.wrap(StringUtil.getUtf8Bytes(str)));
        return this;
    }
```

config/config-api/src/com/thoughtworks/go/config/preprocessor/ParamResolver.java
```diff
@@ -103,6 +103,18 @@ public ParamResolver override(ParamsConfig params) {
         }
     }
 
+    private Field getField(Field leaf, Class type) throws NoSuchFieldException {
```
ketan: Why did this need to change?

server/src/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -130,6 +138,36 @@ private void reloadJobPlans() {
...
+                agent.send(new Message(Action.reregister));
+                return;
+            }
+            if (agentInstance.isDisabled() || !agentInstance.isIdle()) {
+                if (LOGGER.isDebugEnabled()) {
+                    LOGGER.debug("Ignore agent that is " + agentInstance.getStatus());
+                }
+                return;
+            }
+            Work work = BuildAssignmentService.this.assignWorkToAgent(agentInstance);
```
ketan: Probably don't need the `BuildAssignmentService.this` in here?

common/src/com/thoughtworks/go/server/websocket/Message.java
arvindsv: I don't think we're sending over JSON yet. Are we?

arvindsv: Hmm. Maybe we are.

server/webapp/WEB-INF/web.xml
```diff
@@ -252,6 +260,11 @@
     </servlet-mapping>
 
     <servlet-mapping>
+        <servlet-name>agent-websocket-remote</servlet-name>
+        <url-pattern>/aws/*</url-pattern>
```
arvindsv: Using "/aws" might be confusing. :)

jyotisingh:
Merging as most of the changes make sense, @xli can send new PRs for the nitpicks

server/src/com/thoughtworks/go/server/websocket/JobInstanceStatusMonitor.java
xli: @jyotisingh 
1. We need handle reschedule case for robustness. When agent disconnect and then reconnect, server side may lost track its status, hence server tries to assign new work to it, when we assigning new work to agent, we will check agent status and reschedule job that was assigned to the agent for whatever reason, see [BuildAssignmentService#assignWorkToAgent](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/BuildAssignmentService.java#L107). So listening to reschedule and tell agent cancel current work is necessary, but we don't rely on it, because we will try to cancel running work before we start a new work. I think it's a simple choice to just do it, especially when I don't know how much cases will trigger reschedule, and later we may add new way (or reason) to reschedule a job.
2. Server side will always turn on websocket communication, it's agent's choice to not use websocket right now. When there is no agent connected with websocket, there is almost not cost to loop through JobStateChangeListener

common/src/com/thoughtworks/go/server/websocket/Message.java
xli: I don't have prefer to String or byte[].
The reason I choose byte[] instead of String is:
I tried to add GZip to JSON message communication, because there is message size limit in websocket communication, and I worried 1mb limit size may have problem when user configured a job has lots of commands or materials stuff. When we serialize BuildWork, it contains a lot of stuff, I can't predict how much data it may include. We may need long time to find out all unnecessary large data fields (like Modification#files @wpc found yesterday), and exclude them from websocket message. Since GZip will need convert String into byte[], I don't think we need convert them back to String again, because in lower level communication, String will be converted to byte[] again.
I pulled out GZip, because I feel it adds complexity and it won't solve root cause. 
Since I don't have prefer, so left byte[] here.

common/src/com/thoughtworks/go/server/websocket/Message.java
xli: will update it

common/src/com/thoughtworks/go/server/websocket/JsonMessage.java
xli: will update it

server/src/com/thoughtworks/go/server/service/BuildAssignmentService.java
xli: Yes, it is refactored code, will remove it

common/src/com/thoughtworks/go/server/websocket/Message.java
arvindsv: The idea, once the communication starts working and is ok, to move away from these objects completely, to a defined set of messages (a good JSON-based API for agent-server communication). This is what will allow non-Java agents. That's also why versioning of these JSON messages we have now (even though they're just GSON-serialized Java objects) is important.

arvindsv: Oh, of course non-java agents will have a much smaller footprint.

PR #1797, opened 2016-01-13
Improves the Material API notification error message

### LINKED ISSUE #1787: Vague message about material when material API used to notify of a build
I just spent a fair amount of time trying to work out why a GitHub Enterprise web hook wasn’t working. I ended up just trying the curl example from the API docs as I was confused by the message I was getting (a 404 with body: “Unable to find material.”). Unfortunately I could only see that with a bit of digging in GitHub… the message the pre-built web hook gave me was ‘Invalid Go server URL’.

I spotted this message in against the key `MATERIAL_NOT_FOUND` in https://github.com/gocd/gocd/blob/master/server/webapp/localize.xml#L64.

How would people feel about this message suggesting that perhaps the material needs to be set to `autoUpdate="false"` in the configuration XML (as gently suggested in the [docs](https://api.go.cd/current/#notify-git-materials))?

mrmanc:
Fixes #1787 by improving the error message when the API receives a call to notify a material which either does not exist or can not be notified, explaining that this might be because the material is set to autoUpdate. Changes the name of the message to better reflect it’s content.


arvindsv:
Thanks, @mrmanc!

PR #1807, opened 2016-01-15
Select a range of agents via shift-clicking checkboxes

grahamc:
1. If you tick or untick a checkbox, then press shift while checking
   a different checkbox, all the agents between the two will be ticked
   or unticked to match the previous tick.
2. A bit of optimisation on caching jquery selectors
3. I tried to use `jQuery().click...` but the handlers would eventually
   be removed from the elements, forcing me to use `live`, which may
   not be ideal.


grahamc:
Closing, found a regression.

PR #1808, opened 2016-01-15
Select a range of agents via shift-clicking checkboxes

grahamc:
screencast-uywrw

1. If you tick or untick a checkbox, then press shift while checking
   a different checkbox, all the agents between the two will be ticked
   or unticked to match the previous tick.
2. I tried to use `jQuery().click...` but the handlers would eventually
   be removed from the elements, forcing me to use `live`, which may
   not be ideal.
3. A small improvement in the check-all/none code.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/views/agents/_agents_table.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/agents/_agents_table.html.erb
@@ -10,9 +10,7 @@
 Util.on_load(function() {
     jQuery("#select_all_agents").live('click',function() {
         var val = this.checked;
-        jQuery(".agent_select").each(function() {
-            this.checked = val;
-        });
+        jQuery(".agent_select").prop("checked", val);
     });
 
     var lastChecked = null;
@@ -20,7 +18,7 @@ Util.on_load(function() {
     jQuery(document).ready(function() {
         jQuery('.agent_select').live('click', function(e) {
             if (!this.checked) {
-                jQuery("#select_all_agents").removeAttr("checked");
+                jQuery("#select_all_agents").prop("checked", false);
             }
 
             if (!lastChecked) {
--- a/server/webapp/WEB-INF/rails.new/app/views/agents/_agents_table.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/agents/_agents_table.html.erb
@@ -8,33 +8,33 @@
 
 <script type="text/javascript">
 Util.on_load(function() {
-    jQuery("#select_all_agents").live('click',function() {
-        var val = this.checked;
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/rails.new/app/views/agents/_agents_table.html.erb
```diff
@@ -10,16 +10,32 @@
...
-    jQuery("input.agent_select").live("click", function() {
-        var unchecked_check_boxes = jQuery("input.agent_select:not(:checked)");
-        if(unchecked_check_boxes.length > 0) {
-            jQuery("#select_all_agents").removeAttr("checked");
-        }
+    var lastChecked = null;
+
+    jQuery(document).ready(function() {
+        jQuery('.agent_select').live('click', function(e) {
+            if (!this.checked) {
```
zabil: Use of ```.live``` is have you tried this with ```on```?
http://api.jquery.com/live/

ketan: It'll have to be written as

```javascript
jquery('container-elem').on('click', '.agent_select', handler)

```

zabil: The problem with ```live``` is it checks your dom for current and future elements which ends up making this script slow. 
Should take this for a spin with a lot more agents.

ketan:
@rajiesh - can you quickly QA this PR. It seems to be OK from a quick check I performed.

grahamc:
@ketan Updated to use `.on()` instead. Note I seem to have to target `#body_content` otherwise when the table refreshes, the javascript no longer works.

@zabil Does the current state seem like it'll be bad on lots of agents? I can't test that easily, myself.

rajiesh:
verified it on 16.2.0(2867-2df65395774dd0ed4f3b2dc0df1bf3043d675925) - works as expected

JorritSalverda:
For disabling and deleting agents it only does that for the first and last clicked agent, not the ones in between although they're shown as selected. Running Chrome Version 48.0.2564.109 m on Windows 7. 

I have shift pressed for both the first and last item, although I might release shift in between.

alexouzounis:
same issue with @JorritSalverda here, using Chrome Version 48.0.2564.97 (64-bit) on OSX 10.11.3 (15D21)

rajiesh:
Could reproduce this issue on 16.2.0, it happens when user makes the selection and clicks on the action(enable/disable/delete) immediately. If there is a delay of, say 5 secs, between the selection and the action the action applies to all the agents selected. Need to investigate

JorritSalverda:
That wait a couple of seconds trick seems to work for disabling agents, but not so much for deleting agents.

PR #1810, opened 2016-01-18
1133 UI changes

tomzo:
Minimal UI for #1133
These are changes to make UI edits of remote pipelines and environments uneditable.
Mostly this removes or disables links on admin page and dashboard.

I left a "leak" on purpose to be able to see pipeline configuration. User can enter pipeline edit view by going directly to URL like `http://<server>/go/admin/pipelines/<pipline-name>/general`. Considering incoming SPA, I think this is irrelevant how I would handle that page since it will be gone soon.


tomzo:
@naveenbhaskar Could you grayscale these [icons](https://github.com/gocd/gocd/pull/1810/files#diff-018e5284410f80fceba6efb5918fa09bR110) ?

There are other icons there that have both graysclaed and normal variants. Now when pipeline is defined remotely, we need to show user that it cannot be moved or cloned. Until now pipeline could always be moved or cloned, which is why these icons are missing.

arvindsv:
Temporarily, while @naveenbhaskar is on vacation, maybe you can use this: https://gist.github.com/4dca05977b448d4a910e.git

tomzo:
@arvindsv I added your icons but they are more dark than the rest of them. 
It is quite visible when looking on admin panel. I won't push it now and just wait.

arvindsv:
Yeah. I see. Ok. I've updated them to be a little lighter, if it helps. Used:

```
convert icon_edit_16.png -fx '(r+g+b)/2.1' -colorspace Gray icon_edit_16_disabled.png
```

arvindsv:
Looks like this to me:

image

Not perfect but a little better.

tomzo:
I commited the last ones. They look OK to me too.

tomzo:
In order to test UI changes that are applied here.
1. Start development and add my example repository by editing XML config
```
<config-repos>
   <config-repo plugin="json.config.plugin">
     <git url="https://github.com/tomzo/gocd-json-config-example.git" />
   </config-repo>
</config-repos>
```
2. After up to 60 seconds, remote pipelines and environment will be added in server.

Things to notice:
 * pipelines on dashboard  don't have edit sign
 * Admin->Pipelines has disabled icons and links
 * you can still manage (add/edit/move) locally defined pipelines.

jyotisingh:
I was looking at the code as well as testing some of this. 

Few things that I observed:
1. If I introduce a new group say `group1` through one of the partials, <go-server:8153>/admin/pipelines starts showing up the group as expected (with this PR). When I look at the response of `http://localhost:8153/go/admin/config_xml`, I can find an entry for `group1` here too. Then I look at the response of `http://localhost:8153/go/api/admin/config/current.xml` and I don't see this entry. So, the version of config being returned through different calls shows a different version of the config which looks misleading. That is because configForEdit does not really represent the config xml object anymore. It is a decorated version of the actual config xml object. 

2. There are 4 types of ConfigOrigin now - `UIConfigOrigin`, `FileConfigOrigin`, `RepoConfigOrigin` and `MergeConfigOrigin`. If I understand it right, groups/environments which are not part of main config but needs to be shown for edit under the admin section would have the origin set to `UIConfigOrigin`. It took me a while to figure out the purpose of this origin. Also, I wonder if they would introduce more weird scenarios, like the one mentioned in point (1). 

3. You need to rebase with master as the PR pipeline currently [fails](https://build.go.cd/go/pipelines/build-linux-PR/273/build-server/1) as it cannot find the rake task that is used to run rspecs, it was updated a week or two ago on master. Also, there are a few other failings tests -> https://build.go.cd/go/pipelines/build-linux-PR/273/build-non-server/1

tomzo:
I rebased all commits on master just now.

Fixed:
 * `shouldReturnSelfForGetLocalPartsWhenOriginIsFile` and `shouldReturnSelfForGetLocalWhenOriginIsNull`
 * both `shouldgetAllPipelineNamesFromAllParts` - it was invalid order of pipelines. After rebasing on master tests started to fail because of fix in #1756

The tests:
 *  `shouldFailToUpdateEnvironmentVariables_WhenSourceIsNonEditable`. 
 * `shouldUpdateEnvironmentVariables_WhenSourceIsNonEditable_ButChangesAreCompatible`
 * `shouldFailUpdatePipelinesWhenRemovingFromNonEditableSource`
 * `shouldFailToRemoveAgentFromEnvironment_WhenSourceIsNonEditable`
This is just no longer desired. I have simplified how merge environment config should behave [here](https://github.com/gocd/gocd/pull/1810/files#diff-cf30b5278b92eb3a85ed706225bb49b9L164). This long (removed in this PR) implementation was added out of my misunderstanding about attributes sent from UI to `setConfigAttributes`. I removed these tests. They only validated my initial false assumptions.

Don't understand yet why this is failed
 *  `shouldWriteOnlyLocalPartOfConfigWhenSavingMergedConfig`

tomzo:
> If I understand it right, groups/environments which are not part of main config but needs to be shown for edit under the admin section would have the origin set to UIConfigOrigin

I don't think this is exactly like you say. UIConfigOrigin is assigned to some config object when it has been introduced via UI. This is to differentiate which parts of 'decorated config' are actually decorations from partials and which ones are added just now, by user. Why do we need to distinguish them? Because we strip all elements to be saved to xml out of the merged (decorated) config object. The one to be saved are with origins UIConfigOrigin and FileConfigOrigin, while `RepoConfigOrigin` should be discarded when saving xml.
This 'stripping' was done in `getLocal` methods.
Or at least this is what I intended initially. I don't know how your changes would affect this.

tomzo:
I have split this into unquestionable part #2265 with rails changes to respect remote elements.
And *slightly questionable* part #2266 where we start to return **merged** config for edit.

PR #1811, opened 2016-01-18
Let Subversion use the password out of the store

d-led:
The current command line causes the svn updates to fail, with or without using a password store (not plaintext).

Windows: `password-stores = windows-cryptoapi`
Ubuntu: `password-stores = gnome-keyring`

Windows:

```
16:17:00.108 [go] Start updating ...repo at revision #### from http://my.../trunk
16:17:01.342 Updating 'pipelines\My\first':
16:17:01.373 STDERR: svn: E170013: Unable to connect to a repository at URL 'http://my.../trunk'
16:17:01.373 STDERR: svn: E215004: No more credentials or we tried too many times.
16:17:01.373 STDERR: Authentication failed
16:17:01.404 Failed to run svn --username my_username --password= update --non-interactive -r #### c:\...\pipelines\My\first
                                                      ^^^^^^^^^^^
16:17:01.467 [go] Job completed ...job_name on DESKTOP-MY_D [...\agent1]
```

same happens on Ubuntu. Without supplying the empty password parameter, the password caching doesn't fail at least on Windows.

An integration test is highly dependent on the environment: setting up different credential stores, checking out with full credentials, then updating to different revisions, etc. Is there a strategy already to test that?

### DIFF
--- a/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
+++ b/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java
@@ -255,10 +255,7 @@ public class SvnCommand extends SCMCommand implements Subversion {
     private void addCredentials(CommandLine line, StringArgument svnUserName, PasswordArgument svnPassword) {
         if (!StringUtils.isBlank(svnUserName.forCommandline())) {
             line.withArgs("--username", svnUserName.forCommandline());
-            if (StringUtils.isBlank(svnPassword.forCommandline())) {
-                line.withArg(String.format("--password="));
-            }
-            else {
+            if (!StringUtils.isBlank(svnPassword.forCommandline())) {
                 line.withArg("--password");
                 line.withArg(svnPassword);
             }
--- a/common/test/unit/com/thoughtworks/go/domain/materials/svn/SvnCommandTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/materials/svn/SvnCommandTest.java
@@ -452,10 +452,10 @@ public class SvnCommandTest {
     }
 
     @Test
-    public void shouldAddEmptyPasswordWhenUsernameIsProvidedWithNoPassword() throws IOException {
+    public void shouldNotAddEmptyPasswordWhenUsernameIsProvidedWithNoPassword() throws IOException {
         SvnCommand command = new SvnCommand(null, "url", "shilpaIsGreat", null, false);
         CommandArgument argument = new StringArgument(String.format("--password="));
-        assertThat(command.buildSvnLogCommandForLatestOne().getArguments(), hasItem(argument));
+        assertThat(command.buildSvnLogCommandForLatestOne().getArguments(), not(hasItem(argument)));
     }
 
     @Test


zabil:
Thanks for this @d-led 

We don't have tests where we set up ```~/.subversion/config``` or ```/etc/subversion/config``` or it's windows/mac equivalent.

From their doc on ```pasword-stores```

> password-stores
This comma-delimited list specifies which (if any) system-provided password stores Subversion should attempt to use when saving and retrieving cached authentication credentials, and in what order Subversion should prefer them. The default value is gnome-keyring, kwallet, keychain, windows-crypto-api, representing the GNOME Keyring, KDE Wallet, Mac OS X Keychain, and Microsoft Windows cryptography API, respectively. Listed stores which are not available on the system are ignored.

it appears to be desktop environment specific. Which makes writing integration tests tricky. 
We're better of testing svn commands with username and no password (which this patch will support) by setting up a server instance asks for user name but now the password (not sure how but ssh?) 

We have infrastructure in place to spawn docker instances before test runs.

e.g 

https://github.com/gocd/gocd/blob/336c49eaa6775ada0cc6db9eecc3485bb577a3b7/common/test/integration/com/thoughtworks/go/config/materials/mercurial/HgMaterialDockerTest.java

https://github.com/gocd/gocd/blob/336c49eaa6775ada0cc6db9eecc3485bb577a3b7/common/test-resources/integration/hg-material/Dockerfile

And for existing tests you can run 

https://github.com/gocd/gocd/tree/336c49eaa6775ada0cc6db9eecc3485bb577a3b7/common/test/unit/com/thoughtworks/go/domain/materials/svn

d-led:
I'll try to run the existing tests when I find time, but unfortunately it won't be in the nearest future due to limited time.

Perhaps, someone with more subversion + password store configuration experience could test in the meanwhile, if there's interest.

zabil:
@d-led you can click on details against the checks above to figure out what tests have failed.
The username:password for build.go.cd is view:password

d-led:
@zabil thanks! The test indeed checks, whether the command line is in the form `--username <username> --password=`, which is a question of specification, whether `shouldAddEmptyPasswordWhenUsernameIsProvidedWithNoPassword`. Is there any pre-github doc, describing why that spec exists?

I'll try to fix the test when I have time to set up a local environment, but probably, there'll be more value in an independent test of subversion command lines by themselves as input for the `SvnCommand` wrapper

d-led:
short manual test:

this command line prompts for **username** (unexpected!) in interactive mode, and fails, of course, in non-interactive mode
```
go@ubuntu:~$ svn --username <...> --password= update -r <...> /var/lib/go-agent/pipelines/<...>
                                  ^^^^^^^^^^^
Updating '/var/lib/go-agent/pipelines/<...>':
Authentication realm: <...>
Username: 
```
leaving out `--password=` succeeds, as password is taken from cache

zabil:
I think you can configure svn to use blank passwords, which is passed using ```--password```
There'll be someone relying on that functionality.

d-led:
I'll take a look, thanks

d-led:
@zabil is there any documentation concerning that? An initial search didn't yield any usable manual pages or discussions

zabil:
Give me a day to play the detective here. Will have to dig into some logs before we went open source.

jyotisingh:
@d-led - meanwhile could you sign the [CLA](https://www.go.cd/contribute/cla.html) :)

d-led:
@jyotisingh done

d-led:
here's a [reverse problem](https://www.redmine.org/issues/7300), where empty password is a configuration and the command line is required. In my case, the password is non-empty, and due to caching, providing an empty one results in authentication error

zabil:
@d-led 
Did some checking. The only history (for issues from before open source) is
> Console log of a svn material when failed shows password in plain text

The tests assumed a password field to be added (as blank) since the start.

I think it's fine to make this change and modify the unit tests to support the behavior.

Will also have to tweak the display on failures here
https://github.com/gocd/gocd/blob/6dc633f83d9e614bac9591e9ca2b03555546bb3c/common/src/com/thoughtworks/go/domain/materials/svn/SvnCommand.java#L238

To exclude the ---password when left blank.

d-led:
ok! how do we go from there? I haven't been able to find time to set up an environment yet

zabil:
Modify this pull request with a fix for the unit tests and we'll figure out the rest.

d-led:
ok!

ketan:
@zabil — can you review this once again?

d-led:
superb! thanks

Issue #871, opened 2015-02-04
Output environment variables once per job instead of per task


srinivasupadhya:
```
[go] Start to execute task: <exec command="./bn" workingdir="gocd" >
<arg>clean</arg>
<arg>cruise:prepare</arg>
</exec>. 
[go] setting environment variable 'GO_ENVIRONMENT_NAME' to value 'gocd'
[go] setting environment variable 'TZ' to value 'Asia/Calcutta'
[go] setting environment variable 'TLB_BASE_URL' to value ''
[go] setting environment variable 'MAVEN_OPTS' to value '-Xms256m -Xmx1024m'
[go] setting environment variable 'GO_SERVER_URL' to value ''
[go] setting environment variable 'GO_TRIGGER_USER' to value 'changes'
[go] setting environment variable 'GO_PIPELINE_NAME' to value 'build-linux'
[go] setting environment variable 'GO_PIPELINE_COUNTER' to value '348'
[go] setting environment variable 'GO_PIPELINE_LABEL' to value '15.1.0.348-4840f33d6c612951f60789cce6eca1458cbb01c3'
[go] setting environment variable 'GO_STAGE_NAME' to value 'build-non-server'
[go] setting environment variable 'GO_STAGE_COUNTER' to value '1'
[go] setting environment variable 'GO_JOB_NAME' to value 'agent'
[go] setting environment variable 'GO_REVISION_GOCD' to value '4840f33d6c612951f60789cce6eca1458cbb01c3'
[go] setting environment variable 'GO_TO_REVISION_GOCD' to value '4840f33d6c612951f60789cce6eca1458cbb01c3'
[go] setting environment variable 'GO_FROM_REVISION_GOCD' to value '71cb03e270059208b4fb6af86e07412208199bcf'
[go] setting environment variable 'GO_VERSION' to value '15.1.0'
[go] setting environment variable 'SKIP_WAR' to value 'Y'
[go] setting environment variable 'TLB_SPLITTER' to value 'tlb.splitter.DefaultingTestSplitter'
[go] setting environment variable 'TLB_PREFERRED_SPLITTERS' to value 'tlb.splitter.TimeBasedTestSplitter:tlb.splitter.CountBasedTestSplitter'
[go] setting environment variable 'TLB_ORDERER' to value 'tlb.orderer.FailedFirstOrderer'
[go] setting environment variable 'TLB_TMP_DIR' to value '/tmp'
[go] setting environment variable 'load_balance' to value 'yes'
[...]

PR #1820, opened 2016-01-21
for #871 output env variable once instead of per task

wpc:
As described in the issue #871, this PR dump environment variables at job level once.

### DIFF
--- a/common/src/com/thoughtworks/go/util/command/CommandLine.java
+++ b/common/src/com/thoughtworks/go/util/command/CommandLine.java
@@ -368,15 +368,6 @@ public class CommandLine {
                 errorPrefix);
     }
 
-    public static void setEnvironmentVariables(ProcessBuilder pb, EnvironmentVariableContext environmentVariableContext,
-                                               ConsoleOutputStreamConsumer consumer) {
-        Map<String, String> env = pb.environment();
-        for (String line : environmentVariableContext.report()) {
-            consumer.stdOutput(line);
-        }
-        env.putAll(environmentVariableContext.getProperties());
-    }
-
     public void waitForSuccess(int timeout) {
         ConsoleResult lastResult = ConsoleResult.unknownResult();
         long start = System.currentTimeMillis();
--- a/common/src/com/thoughtworks/go/util/command/CommandLine.java
+++ b/common/src/com/thoughtworks/go/util/command/CommandLine.java
@@ -371,8 +371,10 @@ public class CommandLine {
     public static void setEnvironmentVariables(ProcessBuilder pb, EnvironmentVariableContext environmentVariableContext,
                                                ConsoleOutputStreamConsumer consumer) {
         Map<String, String> env = pb.environment();
-
-        environmentVariableContext.setupRuntimeEnvironment(env, consumer);
+        for (String line : environmentVariableContext.report()) {
+            consumer.stdOutput(line);
+        }
+        env.putAll(environmentVariableContext.getProperties());
     }
 
     public void waitForSuccess(int timeout) {
--- a/config/config-api/src/com/thoughtworks/go/util/command/EnvironmentVariableContext.java
+++ b/config/config-api/src/com/thoughtworks/go/util/command/EnvironmentVariableContext.java
@@ -17,10 +17,7 @@
 package com.thoughtworks.go.util.command;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import com.thoughtworks.go.util.GoConstants;
 
@@ -227,25 +224,21 @@ public class EnvironmentVariableContext implements Serializable {
         return properties != null ? properties.hashCode() : 0;
     }
 
-    public String reportText() {
-        throw new RuntimeException("IMPLEMENT ME");
-    }
-
-    public void setupRuntimeEnvironment(Map<String, String> env, ConsoleOutputStreamConsumer consumer) {
+    public List<String> report() {
+        ArrayList<String> lines = new ArrayList<>(properties.size());
+        HashSet<String> names = new HashSet<>();
         for (EnvironmentVariable property : properties) {
             String name = property.name;
             String value = property.value;
             if (value != null) {
-                String line;
-                if (env.containsKey(name)) {
-                    line = format("[%s] overriding environment variable '%s' with value '%s'", GoConstants.PRODUCT_NAME, name, property.valueForDisplay());
+                if (names.contains(name)) {
+                    lines.add(format("[%s] overriding environment variable '%s' with value '%s'", GoConstants.PRODUCT_NAME, name, property.valueForDisplay()));
                 } else {
-                    line = format("[%s] setting environment variable '%s' to value '%s'", GoConstants.PRODUCT_NAME, name, property.valueForDisplay());
+                    lines.add(format("[%s] setting environment variable '%s' to value '%s'", GoConstants.PRODUCT_NAME, name, property.valueForDisplay()));
                 }
-
-                consumer.stdOutput(line);
-                env.put(name, value);
+                names.add(name);
             }
         }
+        return lines;
     }
 }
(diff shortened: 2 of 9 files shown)


arvindsv:
At a high level, the change makes sense to me. The printing moved out of the process (ProcessManager) and into the place where the job starts.

common/src/com/thoughtworks/go/util/command/CommandLine.java
```diff
@@ -368,13 +368,6 @@ private ProcessWrapper createProcess(EnvironmentVariableContext environmentVaria
                 errorPrefix);
     }
 
-    public static void setEnvironmentVariables(ProcessBuilder pb, EnvironmentVariableContext environmentVariableContext,
```
arvindsv: Love it when code goes away! Hate that it was around, though.

config/config-api/src/com/thoughtworks/go/util/command/EnvironmentVariableContext.java
```diff
@@ -227,25 +224,21 @@ public int hashCode() {
         return properties != null ? properties.hashCode() : 0;
     }
 
-    public String reportText() {
-        throw new RuntimeException("IMPLEMENT ME");
-    }
-
-    public void setupRuntimeEnvironment(Map<String, String> env, ConsoleOutputStreamConsumer consumer) {
+    public List<String> report() {
```
arvindsv: There's a bit of a behavior change here. The old `env` map was pre-filled. This new `names` set is not. Earlier call to `setupRuntimeEnvironment` was passing in the existing environment (which would be a combination of environment-level, pipeline-level and stage-level variables and these could be overridden by a job. With this change, I suspect that is broken. Not completely sure.

arvindsv: I mean, there's a difference between `env.containsKey` and `names.contains`. Particularly because of [this line](https://github.com/gocd/gocd/blob/2df65395774dd0ed4f3b2dc0df1bf3043d675925/common/src/com/thoughtworks/go/util/ProcessManager.java#L60).

arvindsv: I think there's definitely a missing `ProcessManager` test which should have caught this scenario.

arvindsv: We can actually have a discussion about whether `processBuilder.environment()` should even be considered for this overriding behavior. But, it used to be, before this change.

wpc: Good eyes :+1:  I am taking a look.

server/test/unit/com/thoughtworks/go/remote/work/BuildWorkTest.java
```diff
@@ -629,4 +613,30 @@ private void createBuildWorkWithJobPlan(JobPlan jobPlan) throws Exception {
...
+    public void shouldReportEnvironmentVariables() throws Exception {
+        buildWork = (BuildWork) getWork(WITH_ENV_VAR, PIPELINE_NAME);
+
+        buildWork.doWork(agentIdentifier, buildRepository, artifactManipulator, environmentVariableContext, AgentRuntimeInfo.fromAgent(agentIdentifier, "cookie", null), packageAsRepositoryExtension, scmExtension, taskExtension);
+
+        String consoleOut = artifactManipulator.consoleOut();
+
+        assertThat(trimTimeStamp(consoleOut), containsString("[go] setting environment variable 'GO_SERVER_URL' to value 'somewhere-does-not-matter'\n"
+                + "[go] setting environment variable 'GO_TRIGGER_USER' to value ''\n"
+                + "[go] setting environment variable 'GO_PIPELINE_NAME' to value 'pipeline1'\n"
```
arvindsv: I wonder why we don't use:

```
export GO_PIPELINE_NAME='pipeline1'
```

I think it'll read better. And maybe:

```
export GO_PIPELINE_NAME='pipeline1' # Overridden
```

Easier to copy too (of course people will need to get rid of the "[go] " part.

config/config-api/src/com/thoughtworks/go/util/command/EnvironmentVariableContext.java
wpc: @arvindsv  Pre-filled variables are system environment variables of agent process. So the behavior difference would just be saying "set 'PATH' to '/tmp/bin'" instead of "override 'PATH' with '/tmp/bin'" in the console, when user trying to override a system environment variable such as 'PATH', 'HOME'. Do you think that's worth paying extra code? If you do I can load the system envs from ProcessManager while reporting.

server/test/unit/com/thoughtworks/go/remote/work/BuildWorkTest.java
wpc: I guess windows user will have some difficulties. They will say I don't see anything "export" related in this dialog -- 
image

config/config-api/src/com/thoughtworks/go/util/command/EnvironmentVariableContext.java
wpc: I will just keep the old behavior. We can easily switch to not consider process level environment for overriding if we decide later.

server/test/unit/com/thoughtworks/go/remote/work/BuildWorkTest.java
arvindsv: :) Ok.

wpc:
Hi, @arvindsv 

I am not sure I follow. Isn't this line [ProcessManager.java#L60](https://github.com/wpc/gocd/blob/35beaf17bdd0b6c38a61b106b3acb47035f40733/common/src/com/thoughtworks/go/util/ProcessManager.java#L60) in createProcess suppose to do that?

-- wpc

ketan:
@wpc — Pro tip — you can trigger with options the "go-cd-PR" with the current SHA of your PR so that it builds and reports success here.

server/test/unit/com/thoughtworks/go/remote/work/BuildWorkTest.java
```diff
@@ -629,4 +616,31 @@ private void createBuildWorkWithJobPlan(JobPlan jobPlan) throws Exception {
...
+                + "[go] setting environment variable 'GO_PIPELINE_COUNTER' to value '-2'\n"
+                + "[go] setting environment variable 'GO_PIPELINE_LABEL' to value '100'\n"
+                + "[go] setting environment variable 'GO_STAGE_NAME' to value 'mingle'\n"
+                + "[go] setting environment variable 'GO_STAGE_COUNTER' to value '100'\n"
+                + "[go] setting environment variable 'GO_JOB_NAME' to value 'run-ant'\n"
+                + "[go] setting environment variable 'JOB_ENV' to value 'foobar'\n"
+                + "[go] overriding environment variable 'PATH' with value '/tmp'\n\n"));
+
```
ketan: Can we make it read like -

```
Setting environment variables -
  - 'FOO'='BAR'
  - 'X'='Y' (overriding)
```

Makes it easy to read and parse and conveys the same information as above paras of text.

Bonus - makes it easy to copy paste into a terminal with minimal edits.

arvindsv:
> If you want def.sh be found you actually need change PATH of who spawning, which in our case is agent process itself. Not possible in java.

Yeah. The exec done from Java will use the agent's PATH to search. From a user perspective, they'd expect PATH to work like I said. That's why I consider it a bug, but I know that it's technically how it is supposed to work.

wpc:
> From a user perspective, they'd expect PATH to work like I said. That's why I consider it a bug, but I know that it's technically how it is supposed to work.

I think for intuitiveness, maybe there should be "shell" task type. For so many times I need write awkward  "bash -c cp xxx" to make environment variables in the command get eval. 

But this is clearly out of scope of this PR.

-- wpc

arvindsv:
Yes, it's out of scope for this. We usually recommend [this](https://github.com/gocd-contrib/script-executor-task) for those who ask. It'll work on Windows too.

ketan:
Closing PR because of merge conflict — Will open another one with conflicts resolved

wpc:
@ketan thanks for the merging

Issue #1727, opened 2015-12-17
[PipelineConfig] Tasks Configuration


maheshp:
Acceptance Criteria

Should be able to ADD/Modify the below types of tasks
- [x] Ant 
- [x] Rake
- [ ] Fetch artifact 
- [x] Pluggable Tasks
- [x] Custom Command
- [x] Lookup Commands
- [x] Run if conditions
- [x] Advance options
- [ ] Fix UI & UX
- [ ] Enable validation

arvindsv:
Closing. Will have to open new issues for a rewrite.

Issue #1596, opened 2015-11-04
PWD environment variable is not updated to be the same as the pipeline workspace


scottmuc:
I'm not even sure if this is a bug, but I was urged to post this here so others can take part in the dialogue.

I have a Makefile that looks like the following:

```
make-pwd:
    @echo $(PWD)

shell-pwd:
    @echo $(shell pwd)

make-curdir:
    @echo $(CURDIR)

env:
    @env

.PHONY: make-pwd make-curdir shell-pwd env
```

When I have a pipeline that has a task as follows:

```
          <exec command="make">
            <arg>make-pwd</arg>
            <arg>shell-pwd</arg>
            <arg>make-curdir</arg>
            <arg>env</arg>
            <runif status="passed" />
          </exec>
```

The results look like:

```
14:12:31.951 /data/go-agent/workspace
14:12:31.953 /data/go-agent/workspace/pipelines/make-pwd-test
14:12:31.953 /data/go-agent/workspace/pipelines/make-pwd-test
14:12:31.954 GO_SERVER_URL=https://<redacted>:8154/go/
14:12:31.954 SHELL=/bin/sh
14:12:31.954 _=/usr/lib/jvm/jdk1.8.0_25/bin/java
14:12:31.954 XFILESEARCHPATH=/usr/dt/app-defaults/%L/Dt
14:12:31.954 GO_FROM_REVISION_GIT=195438d2ee17be1bc730ea15101740870459782b
14:12:31.954 GO_STAGE_NAME=test
14:12:31.954 PATH=/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
14:12:31.954 LOG_FILE=/data/go-agent/workspace/go-agent-bootstrapper.log
14:12:31.954 AGENT_STARTUP_ARGS=-Dcruise.console.publish.interval=10 -Xms128m -Xmx256m   
14:12:31.954 GO_TO_REVISION_GIT=195438d2ee17be1bc730ea15101740870459782b
14:12:31.954 PWD=/data/go-agent/workspace
14:12:31.955 GO_SERVER_PORT=8153
14:12:31.955 HOME=/srv/go-agent
14:12:31.955 LOGNAME=ci
14:12:31.955 GO_PIPELINE_COUNTER=1
14:12:31.955 SHLVL=1
14:12:31.955 AGENT_WORK_DIR=/data/go-agent/workspace
14:12:31.955 USER=ci
[...]

PR #1823, opened 2016-01-22
Can lookup for custom commands on pipeline_config SPA #1727

maheshp:
* CommandSnippets V1 API
* Widget to lookup for commands
* Exec command Args are either shown in textbox or textarea based on the
* type of args

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/argument.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/argument.js
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+define(['mithril', 'lodash'], function (m, _) {
+
+  var Argument = function(data) {
+    this.data = m.prop(data);
+
+    this.isList = function () {
+      return _.isArray(this.data());
+    };
+
+    this.toJSON = function () {
+      if(_.isEmpty(this.data())) {
+        return {};
+      }
+      return this.isList() ? {arguments: this.data()} : {args: this.data()};
+    };
+  };
+
+  Argument.create = function(args, argsList) {
+    return args ? new Argument(args) : new Argument(argsList || []);
+  };
+
+  Argument.vm = function (model) {
+    var listVM = function () {
+      this.data = function (val) {
+        if (val === undefined) {
+          return model.data().join('\n');
+        }
+        var args = _.isEmpty(val) ? [] : val.split('\n');
+        model.data(args);
+      }
+    };
+
+    var stringVM = function () {
+      this.data = function (val) {
+        if (val === undefined) {
+          return model.data();
+        }
+        model.data(val);
+      };
+    };
+
+    return model.isList() ? new listVM() : new stringVM();
+  };
+
+  return Argument;
+});
--- a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/pipeline_configs/_pipeline.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/pipeline_configs/_pipeline.scss
@@ -308,13 +308,51 @@
                 margin-top:   -2px;
               }
             }
+            textarea {
+              min-height: 98px
+            }
+            .lookup {
+              background: #f5f5f5;
+              margin: 0 -15px -1px;
+              padding: 10px 20px;
+            }
 
+            .snippet {
+              padding: 10px 20px;
+              margin: 23px 6px 20px 10px;
+              background: #ffffff;
+
+              header {
+                overflow: hidden;
+                border-bottom: 1px solid #B7B7B7;
+                padding-bottom: 4px;
+                margin-bottom: 10px;
+              }
+              .snippet-title {
+                float: left;
+                margin: 0;
+                font-size: 1rem;
+              }
+              p {
+                font-size: 0.875rem;
+                margin-bottom: 0.25rem;
+                a {
+                  padding-left: 5px;
+                }
+              }
+              .author {
+                overflow: hidden;
+                font-size: 0.875rem;
+                float: right;
+                label {
+                  padding-right: 5px;
+                }
+              }
+            }
           }
-
         }
       }
     }
-
   }
 
   .tabs {
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippet_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippet_representer.rb
@@ -1,0 +1,41 @@
+##########################################################################
+# Copyright 2016 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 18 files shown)


arvindsv:
The text boxes and the rest of the box (apart from the title itself) are drag targets. I don't think they should be. Only the title should be drag-able.

arvindsv:
Also, can you change "LookUp" to "Look up" in the title of the text box?

arvindsv:
@maheshp and I talked about reducing the text boxes, or hiding them. Most of the time, the command and args text boxes are useful. The working directory and lookup text boxes should show up on demand (unless working directory has been overridden?). @naveenbhaskar should weigh in.

ketan:
Seems to auto complete multiple if you put in a comma (you may have to disable multiple autocompletes)

screen shot 2016-01-22 at 11 21 58 pm

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippets_representer.rb
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  class CommandSnippetsRepresenter < ApiV1::BaseRepresenter
+    link :self do |opts|
+      opts[:url_builder].apiv1_admin_command_snippets_url(':prefix')
```
ketan: This seems to render as —

```json
{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/admin/command_snippets"
    },
}
```
What was your intention with `:prefix`?

ketan:
Also - It appears that selecting all the text in the `args` textbox and hitting backspace to clear the textbox does not work (not sure if it's an existing bug)

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/pipeline_configs/_pipeline.scss
```diff
@@ -343,3 +345,45 @@
     }
   }
 }
+
+
+//tasks
+.lookup{
```
ketan: Please - no top level css. We're already in a lot of css mess because of things like this.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/argument.js
```diff
@@ -0,0 +1,39 @@
...
+    this.toJSON = function () {
+      if(_.isEmpty(this.data())) {
+        return {};
+      }
+      return this.isList() ? {arguments: this.data()} : {args: this.data()};
+    }
+  };
+
+  Argument.create = function(args, arguments) {
+    return arguments ? new Argument(arguments) : new Argument(args || '');
```
ketan: `arguments` could be ambiguous because of [this](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/arguments)

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippet_representer.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    property :getName, as: :name
+    property :getDescription, as: :description
+    property :getAuthor, as: :author
+    property :getAuthorInfo, as: :author_info
+    property :getMoreInfo, as: :more_info
+    property :getCommandName, as: :command
+    property :arguments, exec_context: :decorator
+    property :getRelativePath, as: :relative_path
```
ketan: What is `relative_path` useful for?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippets_representer.rb
maheshp: :prefix is a filter. When we lookup for a command we pass the search term which is the :prefix

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/pipeline_configs/_pipeline.scss
maheshp: Ok, lets discuss bout it when @naveenbhaskar is around.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/command_snippet_representer.rb
maheshp: Currenty the CommandRepositoryService has only one api to fetch the CommandSnippet which is getCommandSnippetByRelativePath, thats the reason for this endpoint

ketan: Not sure I follow, let's talk about it on Monday

maheshp: Sure, lets talk

arvindsv:
Just a thought (you know how I am): Dragging a task across jobs and dragging a job across stages.

ketan:
@maheshp — [this](https://github.com/gocd/gocd/pull/1823#issuecomment-173993870) is not yet fixed.

Additionally, it seems that tasks with multiple "args" are rendered in a text box instead of a text area, and that is currently buggy. For e.g. when autocompleting "Chef Solo" the command as rendered by the API is —


```json
        "command": "/bin/bash",
        "arguments": [
          "-c",
          "ssh target-node 'cd target-dir;sudo chef-solo -c config-file -j json-attribs-file -l debug'"
        ]
```

Whereas on saving it is saved as —

```xml
<tasks>
  <exec command="/bin/bash" args="-c ssh target-node 'cd target-dir;sudo chef-solo -c config-file -j json-attribs-file -l debug'" />
</tasks>
```

instead of —

```xml
<tasks>
  <exec command="/bin/bash" args="-c &quot;ssh target-node 'cd target-dir;sudo chef-solo -c config-file -j json-attribs-file -l debug'&quot;" />
</tasks>
```

OR 

```xml
<tasks>
  <exec command="/bin/bash">
    <arg>-c</arg>
    <arg>ssh target-node 'cd target-dir;sudo chef-solo -c config-file -j json-attribs-file -l debug'</arg>
  </exec>
</tasks>
```

arvindsv:
Tab order is a little weird. Goes from "Command" to "Working directory" instead of "Args".

arvindsv:
Switching from "Single" to "Multi" seems to be fine, but switching back, it converts double quotes to escaped spaces. The final result is semantically correct, but unexpected. Is it easy to change that?

arvindsv:
This is the opportunity to remove "Command" and "Args" and just make it one text box! Who's with me? Imagine being able to enter a command, just as in a shell. Oooh.

maheshp:
@arvindsv while switching from 'Multi' to 'Single' the args are escaped before joining so the escape. If not escaped the args would be invalid, I am unable to think of anything that can be done for the double quotes

maheshp:
@arvindsv, true the tab order is wrong its because of the way the input boxes are structured. I spoke to @naveenbhaskar about the UX changes required which are,
1. Hiding the working directory and look up command by default
2. Only the header should be dragable
Once he decides on the UX for this bit will do the necessary changes.

maheshp:
I am all for a single text box, but just wondering what would be complexity of migrating the existing commands.

arvindsv:
> @arvindsv while switching from 'Multi' to 'Single' the args are escaped before joining so the escape. If not escaped the args would be invalid, I am unable to think of anything that can be done for the double quotes

Is it a matter of taking that argument (`ls\ -l\ tmp` of `/bin/bash -c "ls -l tmp"`), and replacing all `\ ` with space and surrounding with double quotes? Would that take care of things like: `/bin/bash -c 'ls -l "/tmp"'`?

arvindsv:
> I am all for a single text box, but just wondering what would be complexity of migrating the existing commands.

Ok, how about providing a single text box in the UI. We will be using shellwords and so, we know which is the command and which are the arguments, right? How away saving that in the backend the same way it is saved today?

arvindsv:
Whenever this is considered "ready to merge", and before merging, take a look at how it will handle something like [this](https://github.com/gocd/gocd/issues/1596#issuecomment-175665285).

maheshp:
Since there were issues with toggling the args I have removed the feature for now, with this commit SPA should work as the current UI.
I have created a new [issue](https://github.com/gocd/gocd/issues/1902) which captures the feedback for the Exec task feature

PR #1825, opened 2016-01-22
1133 configrepo extension

tomzo:
@arvindsv we've been discissing this so many times that I will just summarize what I think is most important to review.

Documentation WIP is [here](https://github.com/tomzo/documentation/tree/1133-configrepo-extension/developer/writing_go_plugins/configrepo)

Some notes:
 * field names in contract classes are used by gson to deduct proper json "schema". So once this is released we should not do renames without adding migration transform with jolt.
 * using gson has downside that it does not support "strict" mode. So when plugin reponse has unknown fields then these are ignored silently. Which is bad in terms of user errors e.g. when response has `"nme : "bla"` instead of `"name" : "bla"`, but it is good when user wants to start migration of configuration repositories before upgrading server. E.g. user can add fields which will be supported by next Go version and current server will not complain.
 * equality and constructors in contract classes are just for tests
 * [working example of configuration repository in json](https://github.com/tomzo/gocd-json-config-example) - I use it to validate end-user functionality. I plan to add more pipelines and environments to cover all objects. 

Code that requires looking at:
 * [Handling response from plugin](https://github.com/gocd/gocd/compare/master...tomzo:1133-configrepo-extension-pr?expand=1#diff-8300e2ccc8a80eafcc4964b68e138d15R46)
 * I developed [base test class](https://github.com/gocd/gocd/compare/master...tomzo:1133-configrepo-extension-pr?expand=1#diff-8e4f9225fad62e452d9e40a80a75f203R22) to help me manage the amount of contract classes to test. It uses *bad* and *good* examples to test common expectations about serialization and validation. There are also a few extra test cases on specific types when I felt it is necessary to test more details.

### New health scope

I wanted to make it a separate PR but it was hard to cherry-pick. So please have a look at just [this one commit](https://github.com/gocd/gocd/commit/5ed9ee9d01f18321f29ab0fd5cb74e04755cfb92) which adds new health scope and error messages per configuration repository. 

### Jolt and migration

I added jolt 0.0.20 as dependency in `plugin-access` module.
Now there is no need to migrate anything, but I wanted to prototype how it could look like with jolts help. So there is a ["test"](https://github.com/gocd/gocd/compare/master...tomzo:1133-configrepo-extension-pr?expand=1#diff-1f44ce40dfc08bf1d327e1e7ba7f6fffR23) which declares some hypothetic "V2" contract classes. Then I used jolt to move some json objects to another field and also I also used it to add some default values.

### Commit history

Unfortunatelly, there is a mess in few moments and I would not recommend reviewing this PR commit-by-commit. Mostly because initially I implemented this with `go-plugin-config-repo` module and then removed it as you know already. So the end result is quite far from what original intent was.


arvindsv:
image

Only you can have more commits than files changed. :) I'll admit I have procrastinated on reviewing this. I am doing this now.

arvindsv:
Alright! I think this is good. I have only a few small comments:

- We can remove deprecated constructor of PluggableTask (easy refactor). I can do this if you want.

- Couple of typos in XmlPartialConfigProvider:
  - ProviderName  => providerName
  - defaultPatter => defaultPattern

- PluginSettingsMetadataLoaderTest.java
  - @Test attribute was removed for the test `shouldFetchPluginSettingsMetadataForPluginBasedOnPluginId`. Intentional?

- GsonCodec#parseDirectoryResponseMessageFromJson can be removed

- In JsonMessageHandler1_0#responseMessageForParseDirectory, I think it'll be useful to propagate the errors found, instead of converting them to text right there. We could show errors separately, if we want, etc. More flexible.

- I really like the good examples, bad examples way you've written those tests.

- A bunch of ConfigConverter methods can be private.


Once this is rebased and conflicts are fixed, I'll take another slightly closer look at the set of CR* classes, but from my perspective, even if there's a bug there, it cannot affect the config, since validations will apply at that time and anyway nothing should get persisted to the filesystem. I'd still like to *try* and verify that everything has been mapped, there.

tomzo:
I just rebased all commits on current master. @arvindsv I will fix these issues you mentioned today.

> We can remove deprecated constructor of PluggableTask (easy refactor). I can do this if you want.

You mean the one which has unused `name` argument?

```java
@Deprecated
    public PluggableTask(String name, PluginConfiguration pluginConfiguration, Configuration configuration) {
        this.pluginConfiguration = pluginConfiguration;
        this.configuration = configuration;
    }
```

tomzo:
> A bunch of ConfigConverter methods can be private.

Most of them are public for testing in `ConfigConverterTest`. Is that what you mean?

tomzo:
> but from my perspective, even if there's a bug there, it cannot affect the config, since validations will apply at that time and anyway nothing should get persisted to the filesystem. I'd still like to try and verify that everything has been mapped, there.

It is true, I had errors there already, in worst case some element of pipeline is "unconfigurable". 
I don't know what you mean by "get persisted to the filesystem" - all of that data is stored in memory only.

arvindsv:
> You mean the one which has unused name argument?

Yes.

> Most of them are public for testing in ConfigConverterTest. Is that what you mean?

I thought they weren't being used from tests. Leave it there. I'll take a look.

> I don't know what you mean by "get persisted to the filesystem" - all of that data is stored in memory only.

As you said, all of that is in memory. I was just commenting that the risk is low, since they never get to the filesystem (or stored in cruise-config.xml).

tomzo:
> Only you can have more commits than files changed. :) 

There it is - 85. I squashed some of it :P

I just rebased and all issues you mentioned are done. This is ready on my end.

arvindsv:
@tomzo: I merged this, with the un-checked assumption that there are already some slight differences between the config on the server side, and the config-repo side (the CR* classes). For instance, the new git shallow clone option, I think, is not a part of this. I did it thinking it's better to fix it later, than to keep this open for longer.

The rest looked fine.

tomzo:
Actually I added shallow clone some time ago https://github.com/gocd/gocd/pull/1825/commits/580d0902b4dd76866a9e65fcf96bfd0e26dae789

arvindsv:
Ah, cool.

PR #1827, opened 2016-01-23
1133 fix invalid merged config cache

tomzo:
This should be merged after #1810, it is a single commit then.

@jyotisingh described the problem clearly

> if config save passes (ie. the new config is checked in to config.git repo) whereas tryAssembleMergedConfig fails, then the config object in memory would not be updated. So, you would have a different version of config object in memory and a different one in the git-repo and file-system. 

This is true. I added tests to confirm first. Then I altered behaviour so now when changes are introduced **because main config has changed** then in case of failed merge with **latest** partials, the merge will fallback to using previous partials (which are used anyway as current config).

In other words:
We want to ensure that UI changes are always present in current, valid,
cached config. There is always some valid base config. When partials are
involved there is at least "last known valid merged configuration". So
regardless of the fact if latest main config and partials are mergable,
we can always fallback to using old partials with new modified main
configuration.


tomzo:
I am closing, because problem is addressed in other PRs by @jyotisingh and me.

PR #1831, opened 2016-01-23
#1596 explicitly set PWD when agent starts process

tomzo:
Proposed fix for #1596 

We just force setting PWD when agents spawns a process.

I tested this by debugging development agent running exec task `env`. 

Before the change, it is set to agent root workspace.
```
21:19:41.450 PWD=/ide/work/agent
```

After this change, it is set to pipeline workspace as one would expect
```
21:42:14.495 PWD=/ide/work/agent/pipelines/pwd-test
```


arvindsv:
Hi @tomzo: I'll close this till we try out the `bash -c` option mentioned in our conversation in #1596.

PR #1834, opened 2016-01-24
WIP: Make build reproducible

arvindsv:
This is needed so that we can easily re-release an older version of Go, with some security fixes, etc. To be able to truly verify that an new build is just an old build with the new changes, we need the ability to rebuild old revisions, given a gocd/gocd commit SHA.

* Strip time from JAR and ZIP files so that they remain the same across runs.
* Handle one-jar files, which package jars and zips inside them.
* Command repository zip is not full reproducible because .git/index changes
  across clones, even if nothing has changed.

This works for mostly everything. The only thing in go.jar which is not reproducible is the command repository. I think we will need to store the command repository zip packaged with the installer (for every commit in the repo) somewhere, and reuse it if we truly want a reproducible build.

Second, installers are not handled in this PR. We should probably move everything out of cruise-modules.rb into fpm and try and handle it there. It won't be easy if we don't control building though.

Third, plugins need to be made reproducible too. Since it is a different repo, it can be done there. This is very easy to do.

This PR needs JDK8 (the reproducible-build-maven-plugin needs it). So, it cannot be merged now.

My idea is to run our build inside some VM or container, so that the version of JDK and other dependencies are fixed.

PS:  I really hate buildr and maven.

### DIFF
--- a/addon-api/database/pom.xml
+++ b/addon-api/database/pom.xml
@@ -45,6 +45,11 @@
             </plugin>
 
             <plugin>
+              <groupId>io.github.zlika</groupId>
+              <artifactId>reproducible-build-maven-plugin</artifactId>
+            </plugin>
+
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>2.4</version>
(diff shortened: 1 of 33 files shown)


ketan:
So I don't forget the link — http://imperceptiblethoughts.com/2013/12/20/reproducible-multi-project-gradle-builds-part-2.html

There should be a way to just override the standard "zip" and "jar" tasks in gradle to be able to do what that blog says.

zabil:
Closing this as the out builds are moving to gradle.

PR #1842, opened 2016-01-26
Make pipeline related errors more visible on dashboard

xli:
* Show an error icon for pipelines that have errors/warnings related like the following screenshot:

pipelines_-_go

* When user click the icon, we'll show all errors/warnings related to the pipeline in a modal dialog, which is same with current all errors modal dialog behavior. 
* Existing showing all errors at right bottom corner is not changed.
* Only show pipeline related errors icon on dashboard, no other places added yet.  I think it is enough to show errors on dashboard, because showing errors everywhere won't really help. However, if we'd like it show up somewhere else (e.g. pipeline details page), it should be easy to just add it. 

see #1595 for more details

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_errors.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_errors.js
@@ -1,0 +1,67 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2014 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *************************GO-LICENSE-END**********************************/
+
+PipelineErrors = function ($) {
+    var refreshPipelineErrors = function() {
+        var pipelinesHaveError = [];
+        $("#cruise_message_body [data-pipelines]").each(function (i, ele) {
+            if ($(ele).data("pipelines")) {
+                var pipelines = $(ele).data("pipelines").split(",");
+                $.each(pipelines, function(i, name) {
+                   if (pipelinesHaveError.indexOf(name) == -1) {
+                       pipelinesHaveError.push(name);
+                   }
+                });
+            }
+        });
+
+        $(".pipeline").each(function (i, pipeline) {
+            var name = $(pipeline).find(".pipeline_name_link a").text();
+            if (pipelinesHaveError.include(name)) {
+                if ($(pipeline).find(".pipeline-error").length == 0) {
+                    var icon = $('<i class="pipeline-error"></i>');
+                    icon.data("pipeline", name);
+                    $(pipeline).find(".pipeline_operations").prepend(icon);
+                }
+            } else {
+                $(pipeline).find(".pipeline-error").remove();
+            }
+        });
+    };
+    var showPipelineErrors = function(errorIcon) {
+        var pipelineName = $(errorIcon).data("pipeline");
+        var errors = $('<div class="cruise_message_body"/>');
+        $('#cruise_message_body .error').each(function(i, ele) {
+            var pipelines = $(ele).data("pipelines").split(",");
+            if (pipelines.include(pipelineName)) {
+                errors.append($(ele).clone());
+            }
+        });
+        Modalbox.show(errors[0], { title: pipelineName + ' error and warning messages', overlayClose: false });
+    };
+    var initialize = function() {
+        $(document).bind("dashboard-refresh-completed", refreshPipelineErrors);
+        refreshPipelineErrors();
+        $(document).on("click", ".pipeline-error", function(e) {
+            showPipelineErrors(e.target);
+        });
+    };
+
+    return {
+        initialize: initialize,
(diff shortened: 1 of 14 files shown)


arvindsv:
I haven't really, fully seen what you've done (sorry). But, here's a thought:

Would it have been easier and/or cleaner to embellish the [pipeline HTML snippet](https://github.com/gocd/gocd/blob/19e35a412f00e431f0c70da3fd2e8805bf392866/server/webapp/WEB-INF/rails.new/app/views/shared/_pipeline.html.erb#L27) with information from [here](https://github.com/gocd/gocd/blob/19e35a412f00e431f0c70da3fd2e8805bf392866/common/src/com/thoughtworks/go/presentation/pipelinehistory/PipelineInstanceModel.java#L36), and passing along some kind of a proper material ID from the server health messages.json? The JSON telling which material it is makes sense. The rest can be done easily in Javascript.

I know you have some JS, but wondering whether `getPipelineNames` is really needed. Even if it is, I'd look at whether that can be moved into the config object.

xli:
HealthStateScope has list of factory methods to create HealthStateScope from pipeline, stage, job, material and other stuff, so it knows how to convert scope back.

Material related error is one of errors related to pipeline, for example: job hanging warning, fanin. Some errors directly point to pipeline name. So the match work in JS will use pipeline and material info. 
Material related errors have 2 type: MATERIAL and MATERIAL_UPDATE, we probably need to convert them into one thing like fingerprint. (I could not figure out why they are different)

I do like your idea of outputing data (materials with pipeline, scope info with errors), then having JS to match them.

However, in this case, we can't avoid some preprocess for HealthStateScope's scope:
* Convert a job name to pipeline name because no job names on pipeline dashboard, or we need add job names on dashboard (more stuff output).
* convert material type's scope to material config fingerprint (discussed above)
* Direct output scope causes JS code need to know what is scope data meaning, which should be hide inside HealthStateScope.

So, I think convert everything into pipeline names by HealthStateScope, then matching pipeline names on UI will be simpler.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_errors.js
```diff
@@ -0,0 +1,62 @@
...
+PipelineErrors = function ($) {
+    var refreshPipelineErrors = function() {
+        var pipelinesHaveError = [];
+        $("#cruise_message_body [data-pipelines]").each(function (i, ele) {
+            if ($(ele).data("pipelines")) {
+                var pipelines = $(ele).data("pipelines").split(",");
+                pipelinesHaveError = pipelinesHaveError + pipelines;
+            }
+        });
+        $(".pipeline").each(function (i, pipeline) {
```
zabil: @xli have to tried doing this on a config with more than 500 pipelines? This is going to run very 5 seconds according to  ```DashboardPeriodicalExecuter``` I suspect this will block the thread for too long.

xli: @zabil Good point, I didn't, will try it out.
I think I can make it better by look up pipeline panel by name.
But I think there is no improvement for the case that all of 500 pipelines have errors.
What you expect for 500 pipelines errors case?

xli:
@zabil I did performance test with 1000 pipelines (all pipelines have errors), here is result:

Note: test is done in my development environment.

Before change, fetching pipelines.json and messages.json are slow, took 20~30 seconds:

p1

After change, fetching pipelines.json and messages.json are similarly slow, and more likely the performance of fetching messages.json is depending on performance of fetching pipelines.json:
pipelines_-_go 3

So, adding error related pipeline names into messages.json won't cause performance worse. However, messages.json size increased from 1003b to 3.1kb, as 1000 pipeline names are included.

Then, after change, how is performance of looping through 1000 pipelines to update error icon:
pipelines_-_go 5

------------------

I also tried another solution to not loop through all pipelines, using a cache to only find diff and update. But no performance improvement. 
Another ajax call for refreshing pipelines by calling pipelines.json will update pipeline panels, so I think it's better we refresh errors by going through all pipelines.

zabil:
nice. 
@naveenbhaskar looks after the UX, he'll put in suggestions (if there are any).

xli:
@naveenbhaskar, any suggestions?

naveenbhaskar:
@xli , it looks good to me. One suggestion is to change the red color to the existing error red color.
we have font-awesome icons available in the code. so please use the warning icon from font-awesome and then give the red color to it. You can use this color value for red (#de3730)
image

xli:
@naveenbhaskar It seems font-awesome is only used in stylesheets/pipeline_configs. I don't want to change too much of existing old scss files loading structure. Could we use the same warning icon for the right bottom errors? It will be consistent with existing icons.

xli:
@naveenbhaskar what you think about my previous comment?

ketan:
@xli — Would you have the time to rebase the PR, and continue using the icon you're already using. We'll sort out font-awesome later if needed.

Thanks.

xli:
@ketan done

ketan:
> @ketan done

Thank you :)

arvindsv:
I've merged this in. The feature is useful and this PR has been around for a bit.

But, I am concerned about the performance of this: Because of [this line](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/server/_messages_body.html.erb#L6) leading to [this](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/serverhealth/HealthStateScope.java#L151), this code ends up going through every material in every pipeline and creates a few objects for each, for every request for every user (it's called from a before filter in app controller).

We'll probably need to put it behind a toggle till we can decide on a proper caching mechanism for this.

arvindsv:
@ketan and I had talked about putting in a toggle for this, while we figure out better caching (this call was not well cached, even before this PR was submitted). I've done [this](https://github.com/gocd/gocd/commit/a7a5717cbd60c30006314fb8dd529796c93adaf0). @ketan will take a look at caching this better sometime. It's not very urgent, but I'd like to do it, at least as a part of the dashboard rewrite/refresh.

PR #1844, opened 2016-01-27
#1086 Change to maintain order on Maps where expected.

zabil:
Fix specs that expect order of Maps


ketan:
I'm fine for merging this sort of stuff in — but we should be know better to not rely on order of things.

Issue #313, opened 2014-07-07
Shallow clone of Git materials


hackinjack:
We have some git repos which contain pretty large history - it can take about 20 minutes just to check this out so obviously we don't need ALL this cloned into our deployments. Is there a way of getting the 'depth' parameter added to the clone commands performed internally by Go? That way we could check out this repo in seconds.

paulclarkin:
See also #711

paulclarkin:
Any update on this?

mitchtbaum:
Users on slow and/or metered networks could benefit from this.  So, +1!! :-)

scottmuc:
+1

eelkram:
+1.  I would think the default should be a depth of 1 and a setting if you needed to have the history.

cholmes1111:
+1. Would be nice to pull just the current one by default but have the ability to overwrite if you want the full history.

dsmorse:
+1 - last year at the atlassian conference, they mentioned that performing shallow clones on their build servers was one of the most important changes they made on their infrastructure to support enterprise scale devops.  I wish Go allow use to set the depth.

agbodike:
+1 - This would speed up builds for many people

ketan:
So as not to need a feature toggle, can we —

* add a `shallow` boolean attribute (default it to `--depth 40`) to `<git />` material. This is an opinionated and non-configurable choice.

OR 

* add a `depth` (Integer) attribute to `<git />` material?

agbodike:
I prefer the second option. I assume that for all existing configs that do not have the depth attribute set, it would do a full repo clone, but for those that want control over the depth, they can get it as fine grained as they want.

agbodike:
I'm not sure if this is feasible, or desirable, but a third option would be to add an `args` (String) attribute to <git /> material, which would allow for arbitrary parameters to be passed to the git clone command.

holmesc1111:
If the goal is to not add a feature toggle, I'd prefer that the default depth integer be set to 1. If for any reason someone wants more depth than that, they would override the default of 1.

I'm still not seeing a viable use case where full depth is needed every time a material check is performed for every pipeline.

eelkram:
I prefer the 2nd option (depth integer attribute) as well.  I think you want to support granularity in the depth so if someone want / needs any specific depth that option is available.  A default of 0 should clone the entire repo causing no functional change.

kkennedy1:
[...]

PR #1846, opened 2016-01-28
Shallow cloning for git material

wpc:
This PR implement the [proposal](https://github.com/gocd/gocd/issues/313#issuecomment-174103478) on issue #313. 

What it does
---
* Added a configuration option for git material called `shallowClone`, default to false
* Shallow clone config only applies to agent. On server side flyweight clone (includes configuration repository) is always shallow.
* If shallowClone is specified, git repository on agents clone will be a shallow clone with depth 2.
* If build is triggered with a missing revision on a shallow cloned repo, use a 2 steps process to unshallow the repo. 
* Convert existing shallow cloned repos to full clone when user turning shallowClone off.

How to test it
---
Since UI part of the shallow clone configuration is not included, enabling the configuration option need be through editing cruise-config.xml. The following xml snippet setup a pipeline checkout linux-kernel repository with shallow clone enabled.

```xml
<pipeline name="linux-kernel">
  <materials>
    <git url="git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git" shallowClone="true" materialName="linux-kernel" />
  </materials>
  <stage name="ls-stage">
    <jobs>
      <job name="ls-job">
        <tasks>
          <exec command="ls" />
        </tasks>
      </job>
    </jobs>
  </stage>
</pipeline>
```

Benefits
---
For the linux kernel repo,  on my machine/network,  without shallow clone, it takes more than 15 minutes to complete first build. With shallow clone option enabled, it takes about two minutes to complete first build.

Other notes
---
* This PR works with old git versions,  tested back to 1.7.1.
* This PR includes configuration migration (bumped schema version to 79)
* Server side shallow clone can be turn off by setting system property `go.server.shallowClone` to `N`
* This PR changed the way how ScheduledPipelineLoader lookup `knownMaterials`. It scopes material config lookup under the current pipeline. Set system property `go.server.scheduledPipelineLoader.globalMaterialLookup` to `Y` can bring the old behavior back (lookup material through whole cruise config)


common/test/unit/com/thoughtworks/go/domain/materials/git/GitTestRepo.java
```diff
@@ -67,7 +76,7 @@ public GitTestRepo(File gitBundleFile) {
     }
 
     public String projectRepositoryUrl() {
-        return gitRepo.getAbsolutePath();
+        return FileUtil.toFileURI(gitRepo);
```
wpc: This change make projectRepositoryUrl method return a real url (`file:///foo/bar.git` instead of `/foo/bar.git`). The reason I switch it is because git clone depth option can not be used with a local file system clone.

config/config-server/resources/cruise-config.xsd
```diff
@@ -393,6 +393,7 @@
         <xsd:attribute name="url" type="xsd:string" use="required"/>
         <xsd:attribute name="dest" type="filePathType" use="optional"/>
         <xsd:attribute name="branch" type="xsd:string" use="optional"/>
+        <xsd:attribute name="shallowClone" type="xsd:boolean" use="optional"/>
```
wpc: This is a pure "addition" to the scheme. Do I need bump up config schema version?

arvindsv: Yes, you'll need to.

arvindsv:
I haven't checked this, but I assume changing the shallow flag from false to true and then back to false, in the config, is handled gracefully.

wpc:
@arvindsv 
> I haven't checked this, but I assume changing the shallow flag from false to true and then back to false, in the config, is handled gracefully.

Haha, You got me. Not yet.  Will do it this way if shallow flag change and there is an existing repo:

Case 1: shallowClone is changed from on to off, existing repo is shallow - delete working copy and redo clone
Case 2: shallowClone is changed from on to off, existing repo is full - do nothing
Case 3: shallowClone is changed from off to on, existing repo is shallow - do nothing
Case 4: shallowClone is changed from off to on, existing repo is full - do nothing to save work

What do you think?

arvindsv:
@wpc: Makes sense. Case 3 shouldn't ever happen, if Case 1 has worked, though.

wpc:
@arvindsv 
for case 3: I guess the repo type left on agents will be uncertain, because agents may not be picked up during configuration changing back and forth. But it does not matter because we do not need do anything for this case anyway.

common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
```diff
@@ -106,28 +127,32 @@ public MaterialConfig config() {
...
     public MaterialInstance createMaterialInstance() {
-        return new GitMaterialInstance(url.forCommandline(), branch, submoduleFolder, UUID.randomUUID().toString());
+        return new GitMaterialInstance(url.forCommandline(), branch, submoduleFolder, UUID.randomUUID().toString(), shallowClone);
     }
 
     @Override
     protected void appendCriteria(Map<String, Object> parameters) {
         parameters.put(ScmMaterialConfig.URL, url.forCommandline());
         parameters.put("branch", branch);
+        parameters.put("shallowClone", shallowClone);
```
wpc: @arvindsv and @ketan : I have added ShallowClone flag into the material criteria -- which makes material with different ShallowClone config stored separately in DB. I feel this is important so that different pipeline can use different clone settings, e.g using shallow clone in test pipeline and using full clone in heroku deploy pipeline. 

There is one implications to this change: this will change how git materials generate fingerPrint. One thing that worries me is old materials in the db will have a fingerPrint value that out of sync. I manually tested the upgrading process, seems like go server do not really cares about that.  But my knowledge on go is too limited, so want to ask you guys is this fine?

arvindsv: Umm, that might not be ok. Changing fingerprint is not something that should be done as far as possible, since it has quite big implications on fan-in, etc. Might not show easily during an upgrade. This needs @jyotisingh's expertise. CCing her.

Does it really need to part of the fingerprint? Can't it just be stored in the DB, without being part of the fingerprint? Anyway, @jyotisingh will know more/better.

wpc: Got it. Is the fingerprint used externally such as file system? If not maybe I can refresh them during db migration.

Something like this:
```sql
SET @delimiter = '<|>';

UPDATE materials SET
    shallowClone = false
    fingerprint = HASH('SHA256', STRINGTOUTF8(concat('type=', type, @delimiter, 'url=', url, @delimiter, 'branch=', branch, @delimiter, 'shallowClone=false')), 1)
    WHERE type = 'GitMaterial';
``` 
What do you guys think?

arvindsv: Yes, fingerprint is used all over. My point is, why can't this value be like `submoduleFolder` which is in GitMaterial? It's not part of the fingerprint, but [is persisted](https://github.com/gocd/gocd/blob/eb0ae575fbb0a30f59ed646cd45d16eb57d26aa5/server/resources/hibernate-mappings/MaterialInstance.hbm.xml#L46).

Also, as a part of pluggable SCM support, a new field called "additionalData" was introduced. It can be used in any material and can hold a JSON value. Is that something we should look into for this? I ask this because: As you know, the materials table is flattened. So, even a GitMaterial row in it has a null value for a column called `useTickets`, which is a Perforce (@jyotisingh's favorite SCM) field. So, the idea of "additionalData" as a column was to have anything which is specific to that SCM, but not necessarily common to all. "shallowClone" fits that definition according to me.

Whatever we choose, I wonder if there's a reason to include shallowClone in the fingerprint. Assume it is not part of the fingerprint, and so the Go Server will have only one material for it, and won't differentiate between the two. But, the Go Server is not the goal of this PR. The agents are the goal (where a shallow or non-shallow checkout really matters).

I know it kind of matters in Go Server as well, but I don't know if this is big enough of a reason to change fingerprint. Migrating fingerprint, as you say, might be an option. We'd need to ensure that it is correct, though.

arvindsv: The fingerprint is used internally and flyweight is used externally (for the filesystem). So, I'm not worried about a re-clone on the filesystem. I'm worried about existing materials and fan-in. Migration might mitigate that problem, if done correctly. But, I'll let others on the team think/worry about it. :)

wpc: hi @arvindsv:

> My point is, why can't this value be like submoduleFolder which is in GitMaterial? It's not part of the fingerprint, but is persisted.

If user have two pipeline point to same git repo and branch, and they turn on or off shallow clone in one of them, what is the behavior should we expect if we do not distinguish them?  When I was reading the code, I do wonder how 'submoduleFolder' handling avoids this confusing when they have inconsistent settings.

>Also, as a part of pluggable SCM support, a new field called "additionalData" was introduced. It can be used in any material and can hold a JSON value. Is that something we should look into for this?

I was looking into it. 'additionalData' currently has very specific usage for storing additional value coming along with Pluggable SCM plugin polling result. I am not sure I want to mess with it. Plus it does not help solving the material identity problem.

>So, even a GitMaterial row in it has a null value for a column called useTickets, which is a Perforce (@jyotisingh's favorite SCM) field

Yes, once in a STI hell, very hard to get out :-). But it is not that bad in GO's case because MaterialInstance model is almost logic free.

>  But, the Go Server is not the goal of this PR. The agents are the goal (where a shallow or non-shallow checkout really matters).

I totally agree. I feel in a ideal implementations, flyweight clone on go server should always be shallow (because historical revision is never used), and shallow flag is turn on and off for agents. But that would involve too much changes to the existing process.

wpc: I guess another option instead of migrating fingerprints is to write extra configuration validation to ensure consistency. "Validation Error: You can not have multiple git materials to same repository and branch with different shallowClone value". But this seem throwing more burdens to the user's end and limiting the use of the feature.

arvindsv: About material identity: My guess about submoduleFolder is that it's not even used. Maybe I'm wrong.

> I totally agree. I feel in an ideal implementations, flyweight clone on go server should always be shallow (because historical revision is never used), and shallow flag is turn on and off for agents. But that would involve too much changes to the existing process.

[This](https://github.com/gocd/gocd/blob/eb0ae575fbb0a30f59ed646cd45d16eb57d26aa5/server/src/com/thoughtworks/go/server/service/materials/GitPoller.java#L31) is the only place called from the MDU, for git material clone. Do you think it's worth trying to shallow that up? :) Behind a flag? If what I'm saying is right and it's only that place, it might not be too much of a change.

wpc: @arvindsv 
dude, don't you need sleep ? :-D

On material identity:
Maybe I am wrong, but submoduleFolder is not included in the identity is problematic. Given following material configs:

```xml
<pipeline name="p1">
      <materials>
        <git url="https://the-repo" materialName="the-repo" submoduleFolder="foo" />
     <materials>
</pipeline>

<pipeline name="p2">
      <materials>
        <git url="https://the-repo" materialName="the-repo"  submoduleFolder="bar" />
     <materials>
</pipeline>
```
How do we guarantee p1 build get submoduleFolder "foo" and p2 build get submoduleFolder "bar"?

arvindsv: > dude, don't you need sleep ? :-D

I'm closer to you than you think ... :)

> Maybe I am wrong, but submoduleFolder is not included in the identity is problematic ...
> How do we guarantee p1 build get submoduleFolder "foo" and p2 build get submoduleFolder "bar"?

Could be. My point was submoduleFolder doesn't seem to be something you can set in config. I couldn't find it in the [schema](https://github.com/gocd/gocd/blob/eb0ae575fbb0a30f59ed646cd45d16eb57d26aa5/config/config-server/resources/cruise-config.xsd). So, I don't know what it is for, really. Do you?

arvindsv: Anyway, even if it can be set somehow, it's possible that it doesn't matter for a material checkout or clone on the server, but it is sent to the agent and it doesn't the right thing while preparing for a job run. In that sense, it is one material on the server, but a slightly different one on the agent. As long as the final revision of the checkout is ok, the agent shouldn't care.

That's what I was wondering whether we could do for shallow clone. Ignoring the shallowClone option on the server side and doing whatever is needed to make polling work (user's don't care about that clone. It's internal to the Go Server). On the agent side, do what the users asked for.

wpc: >  Ignoring the shallowClone option on the server side and doing whatever is needed to make polling work (user's don't care about that clone. It's internal to the Go Server). On the agent side, do what the users asked for.

I like this. The trouble I have is: material is a not per pipeline model, even in cruise-config.xml it looks like. So it is currently not possible to have per pipeline attributes on materials except you persist separated material instances into DB.

This afternoon when I dig deeper into the code, I found ScheduledPipelineLoader#pipelineWithPasswordAwareBuildCauseByBuildId method is very interesting.
```java
    public Pipeline pipelineWithPasswordAwareBuildCauseByBuildId(final long buildId) {
          ...
            if (materialConfig instanceof PasswordAwareMaterial) {
                PasswordAwareMaterial configuredMaterial = (PasswordAwareMaterial) materialConfig;
                ((PasswordAwareMaterial) usedMaterial).setPassword(configuredMaterial.getPassword());
            }
       ...
   }
```
How about we load material config from pipeline config here and merge settings into the material?
```java
    public Pipeline pipelineWithPasswordAwareBuildCauseByBuildId(final long buildId) {
         ...
           usedMaterial.updateFrom(materialConfigFromPipelineConfig)
        ...
   }
```
This way we do not even need write shallowClone flag into the database.

arvindsv: > How about we load material config from pipeline config here and merge settings into the material?

+1. I like that. @jyotisingh? Will need to see what happens to a re-run of an older pipeline? If config doesn't have that material, I guess a re-run fails (currently)? There's a horrible if-clause right above the snippet you pasted, which I think comes into effect.

jyotisingh:
Late to the party. 
I am with @arvindsv that `ShallowClone` should not be part of fingerprint(ie. material criteria). Fingerprint is a unique identifier for a material. 
Talking about `If user have two pipeline point to same git repo and branch, and they turn on or off shallow clone in one of them, what is the behavior should we expect if we do not distinguish them?`. I think this is a valid configuration, and the reason why you need to have full clone for one and shallow clone for the other is because thats what the corresponding tasks running on the agents are going to expect. IMO, the server does not really care about this history. If these two pipelines together formed an upstream to a third pipeline, then fan-in should be honoured, however it wouldn't be if the top-level-scm for these pipelines had different fingerprints. 

I see the option to perform shallow-clone or not as a agent side requirement. I cannot think of a reason why the server would need this history. Moreover, even if had the history, it completely ignores it anyhow. You guys speak of shallowClone on server by default and I totally agree with that. From what I understand, there are 4 main interactions with git that are encapsulated in GitMaterial. `GitMaterial#updateTo` is called from agents, the other three ie. `latestModification`, `modificationSince` and `checkConnection` are called from server. I think it should be enough to pass along shallowClone=false from `GitMaterial#latestModification` and  `GitMaterial#modificationsSince` which should take care of always shallow-clone on server. 

For `updateTo` (agent side) on the other hand, we could use the shallowClone flag that is configured by the user. I think @wpc's suggestion of doing `usedMaterial.updateFrom(materialConfig)` during work creation is completely alright.

>>Will need to see what happens to a re-run of an older pipeline? If config doesn't have that material, I guess a re-run fails (currently)? There's a horrible if-clause right above the snippet you pasted, which I think comes into effect.

Correct.

wpc:
@jyotisingh @arvindsv  thanks for the information. I will pull out the db migration and fingerprint changes and go with loading extra attributes during build work creation.

wpc:
@jyotisingh and @arvindsv 
I have finished what we previous talked about: 
* reload shallowClone config from pipeline config during BuildWork create (same way with how SCM password get passed through)
* pull out the migration because we do not need persist the configuration into database
* Do not touch fingerPrint generation, anymore, server side git repo with same url and branch shares same flyweight regardless of clone configuration.
* Use shallow clone on the server regardless of material clone configuration

wpc:
This PR is ready for review. I have updated PR description to include new changes.

There is a minor conflict in SystemEnvironment.java. If you guys prefer I can setup a new PR with rebased changes.

wpc:
I setup a new PR for this with rebased changes. I need do that because otherwise PR build can not run. (The build expect a new task named "parallel-spec").  

I am closing this lets move on to #1887 
===

PR #1848, opened 2016-01-28
Separate out jetty websocket api dependency into a separate maven module.

ketan:
Tweak the packaging scripts to package the websocket dependencies correctly.


ketan:
Fixing some packaging issues wrt #1793

PR #1855, opened 2016-01-29
add missing dependency commons-configuration to common's pom.xml

xli:


ketan:
@xli — this dependency already exists in the pom

xli:
Ok, probably my IDEA problem

PR #1856, opened 2016-01-29
Improve Websocket message size limit

xli:
Change to use partial send to make sure we can send messages that are larger than websocket max message size. Also remove jetty9-websocket module, move go.websocket package back to common, no dependency on jetty websocket anymore.


ketan:
@xli — this feels like a stop-gap arrangement while we figure out a way to trim down the websocket payload — I don't feel we should need so high limits.

Do you have a list of things that we can do to fix the problem by reducing the payload?

xli:
Still working on how to find out all stuff needed by Agent, then only serializing them.
One thing I'm sure is Job task configured by user maybe large.
Also user can configure a lot tasks in one job.

ketan:
> Still working on how to find out all stuff needed by Agent, then only serializing them.

Do you need any additional info or help to figure this out?

> One thing I'm sure is Job task configured by user maybe large.
> Also user can configure a lot tasks in one job.

I see your point and see why this PR is needed. If I understand correctly you've already marked [fields in `BuildWork`](https://github.com/gocd/gocd/blob/1fb1759420d05d7700bf072126a086d29d5bd3b7/common/src/com/thoughtworks/go/remote/work/BuildWork.java#L60-L66) to be transient, so that takes care of a majority of unnecessary payload. Are there any more that you think are unnecessary? I can see for e.g. that `BuildWork` has `MaterialRevisions` via `BuildAssignment`'s `BuildCause` attribute, which can be huge.

A few questions — 
* our http connections are already gzipped, I don't know if this applies to a websocket connection or not. Can we compress websocket connections? Is there a possibility to turn on per-message compression, and is it worth the effort?
* Would we be serializing more JSON data via websocket than using the earlier binary serialization (which I assume would be more "compact") that could cause delays in job assignment? This could mean a zero-gain in terms of how quickly jobs get picked up in certain pathological cases.

common/src/com/thoughtworks/go/websocket/Socket.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2015 ThoughtWorks, Inc.
```
ketan: Update your plugin to say 2016?

arvindsv:
> Would we be serializing more JSON data via websocket than using the earlier binary serialization (which I assume would be more "compact") that could cause delays in job assignment? This could mean a zero-gain in terms of how quickly jobs get picked up in certain pathological cases.

Getting rid of the ping mechanism should blow any small delays like this out of the water, right? That is itself huge, and I don't see how it can be zero-gain using JSON, even if bigger.

> Still working on how to find out all stuff needed by Agent, then only serializing them.

One approach would be to say, whatever JSON is going back and forth is "Version 1" and leaving it the way it is. Then, "Version 2" can be designed by you, to be ideally how the JSON should be (a proper JSON). Then, we don't need to tie it to existing objects (and worry about transient, etc). We'll need a bit of a transformation on both sides, though.

About this whole message limit thing, isn't there a chunked encoding equivalent for websockets? Or, what am I missing?

common/src/com/thoughtworks/go/websocket/Socket.java
xli: will update it

xli:
> Do you need any additional info or help to figure this out?
I'll ask when I need, thank you.

>  If I understand correctly you've already marked fields in BuildWork to be transient, so that takes care of a majority of unnecessary payload. Are there any more that you think are unnecessary?
Yes, you're right. I marked some fields transient to avoid unnecessary serialization. 
I plan to handle MaterialRevisions as it carries too much stuff is not used at Agent.

> our http connections are already gzipped, I don't know if this applies to a websocket connection or not. Can we compress websocket connections? Is there a possibility to turn on per-message compression, and is it worth the effort?

WebSocket connection does not share anything from http connection. Gzip is simple to add. I'll add gzip then, to make it same with current communication.

> Would we be serializing more JSON data via websocket than using the earlier binary serialization (which I assume would be more "compact") that could cause delays in job assignment?

I'm not sure about size, but I think the performance impact can be ignored comparing to 5 seconds refresh interval we have on server side to refresh job plans (in BuildAssignmentService#onTimer)
I think websocket stuff should not be tied with JSON format, but we still need this change to relax message size limit.

> One approach would be to say, whatever JSON is going back and forth is "Version 1" and leaving it the way it is. Then, "Version 2" can be designed by you, to be ideally how the JSON should be (a proper JSON). Then, we don't need to tie it to existing objects (and worry about transient, etc). We'll need a bit of a transformation on both sides, though.

That's cool. I won't spend too much time on version 1, just want to make sure it works robust.

> About this whole message limit thing, isn't there a chunked encoding equivalent for websockets? Or, what am I missing?

Jetty websocket does not handle chunked encoding.
Websocket has hard limit, you can increase it (like I did before). No matter what the limit is, I feel user may hit it anyway, and I also worry about large size message in production may cause unknown problem.

xli:
I was wrong about jetty handling chunked encoding.
I will rework this part code, found Jetty websocket commons lib has extensions support what we need.

PR #1864, opened 2016-02-01
Websocket communication (#1098) #1793: gzip message

xli:
Gzip should reduce message size a lot. 
@ketan mentioned current http call from Agent has gzip enabled, so I think this should be added to make websocket communication same with current implementation.


xli:
Jetty websocket has compress extension implemented, but it is not stable and turned off by default. 
And compress extension itself is still in draft.
So we need impl it by ourselves.

PR #1866, opened 2016-02-02
Websocket communication (#1098) #1793: enable jetty fragment extension

xli:
Ok, here is jetty fragment extension. Reverted changes merged by #1856.
It simplified implementation of relaxing message size limit. 
But it seems not all websocket impl will have fragment extension.
So, it may have problem when we impl another language agent, which probably is fine, because we may not need fragment anymore (sending small messages)


xli:
This is ready for review

ketan:
I checked in the morning and there is a deflate extension too. Any reason to not use that instead of our custom gzip?

ketan:
Ignore my comment about custom gzip. I just recall you mentioning it is still in draft

ketan:
@xli — can you resolve conflicts?

xli:
@ketan fixed

Issue #1799, opened 2016-01-13
pipeline config API: cannot create pipeline with dependency on an upstream with a template


virtuald:
Pipeline "upstream" uses some template that has a stage called "run". 

Using the pipeline config API, try to create a pipeline "downstream" with a material that points to the upstream pipeline. 

```
"materials": [ some_git_repo, {"attributes": {"auto_update": true, "name": "upstream", "pipeline": "upstream", "stage": "run"}, "type": "dependency"}], 
```

I get the following error:

```
Validations failed for pipeline 'downstream'.... errors: pipeline: "Stage with name 'run' does not exist on pipeline 'upstream'"
```

magnus-lycka:
I get this error doing PUT. I GET the pipeline config, change something else, PUT it again, and it complains about these dependencies on templated pipelines I got on the GET call. They work as expected. I also get another, related error message. One of the upstream pipelines, has a parameterized upstream dependency, and there is also an error message about that: "Pipeline \"#{UPSTREAM_PIPELINE}\" does not exist. It is used from pipeline \"pipeline_x\"." Of course, parameter UPSTREAM_PIPELINE is defined in pipeline_x. That was also already there in the config.

arvindsv:
/cc @jyotisingh, @maheshp - Any thoughts?

@magnus-lycka and @virtuald: I assume that just getting the config and putting it back, without changing anything does the same? If you have any more information or can make this reproducible with a small config, it might help, unless @jyotisingh or @maheshp can already guess from the info you've provided.

virtuald:
@arvindsv I'm not sure. I was only trying to use it to create new pipelines, not modify existing ones.

ketan:
@juhijariwala — can you look into this when you've got time?

magnus-lycka:
Yes, I get the same result if I do GET, then PUT exactly what I got from GET for a pipeline which is fetching artifacts from a templated upstream pipeline. I'm testing on 16.1. Do you want me to provide a test case? It's pretty trivial... 

Use any pipeline P1, and extract to template if it's not already templated.

Make a downstream pipeline P2 with a dependency on P1 which fetches the artifact from P1. (Maybe you don't even need the fetch. It complains in the material section too.) P2 does not need to use a template.

GET http://localhost:8153/go/api/admin/pipelines/P2 > temp.txt as per https://api.go.cd/current/#get-pipeline-config. 

[...]

PR #1869, opened 2016-02-02
Fixed bug- Cannot create pipeline with dependency on an upstream pipeline with a template (#1799)

juhijariwala:


server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
```diff
@@ -184,7 +184,7 @@ public void onPipelineConfigChange(PipelineConfig pipelineConfig, String group)
 
     @Override
     public void onConfigChange(CruiseConfig newCruiseConfig) {
-        PipelineConfigurationCache.getInstance().onConfigChange(goConfigService.getConfigForEditing());
+        PipelineConfigurationCache.getInstance().onConfigChange(goConfigService.cruiseConfig());
```
ketan: @jyotisingh — `getConfigForEditing` has been deprecated, but is still being used in a lot of places. Should we be replacing all occurrences to avoid such bugs?

jyotisingh: Removing usages of `getConfigForEditing` might not have helped avoid this bug. `getConfigForEditing` was deprecated as we did not want to expose the the full cruise-config object. We would have gone with pulling out the required object from this and pass it along. However, in this case, the object to be passed along was the preprocessed version of the config, ie. with template & params inlined and we were returning the other one(without inlining).
However, I agree that we should remove usages of deprecated methods in general and delete them altogether.

rajiesh:
verified this issue on 16.2.0-3018(bf8cd734e38545871223b90c4a6c098d0de337ce). It has taken care of the scenario when upstream pipeline is created via UI and downstream via API. When both are created using API still the validation for the stage  fails.

Created upstream with this request json,

```
"label_template": "${COUNT}",
  "enable_pipeline_locking": false,
  "name": "upstream",
  "template": "TestTemplate",
  "parameters": [
    {
      "name": "tests",
      "value": "value"
    }
  ],
  "environment_variables": [
    {
      "secure": false,
      "name": "GO_VERSION",
      "value": "16.3.0"
    }
  ],
  "materials": [
    {
      "type": "hg",
      "attributes": {
        "url": "file:///Users/rajieshn/workspace/functional-tests/test-repos/real_connectfour.hgbundle",
        "destination": "dest",
        "filter": null,
        "name": null,
        "auto_update": true
      }
    }
  ],
  "stages": null,
  "tracking_tool": null,
  "timer": null
```

and create downstream with this request json,

```
{
  "group": "group1",
  "pipeline": {
    "label_template": "${COUNT}",
    "enable_pipeline_locking": false,
    "name": "downstream",
    "template": null,
    "parameters": [
      {
        "name": "tests",
        "value": "tests"
      }
    ],
    "environment_variables": null
    ],
    "materials": [
      {
        "type": "dependency",
        "attributes": {
          "name": "dependency",
          "pipeline": "upstream",
          "stage": "defaultStage",
          "auto_update": true
        }
      }
    ],
    "stages": [
      {
        "name": "downstream",
        "fetch_materials": true,
        "clean_working_directory": false,
        "never_cleanup_artifacts": false,
        "approval": {
          "type": "success",
          "authorization": null
        },
        "environment_variables": null
        ],
        "jobs": [
          {
            "name": "tests",
            "run_instance_count": 1,
            "timeout": null,
            "environment_variables": null
            ],
            "resources": [
              "maven",
              "linux"
            ],
            "tasks": [
              {
                "type": "exec",
                "attributes": {
                  "run_if": [
                    "passed"
                  ],
                  "command": "mvn",
                  "arguments": [
                    "--version"
                  ],
                  "working_directory": "#{tests}"
                }
              }
            ],
            "tabs": null,
            "artifacts": null,
            "properties": null
          }
        ]
      }
    ],
    "tracking_tool": null,
    "timer": null
  }
}
```

will fail for the validation error ,

```
"errors": {
          "pipeline": [
            "Stage with name 'defaultStage' does not exist on pipeline 'upstream'"
          ]
        }
```

varshavaradarajan:
Verified working as expected when both the upstream pipeline and the downstream pipeline are created via the API. Okay to close https://github.com/gocd/gocd/issues/1799

PR #1870, opened 2016-02-02
Jasmine specs take less time to run

maheshp:
* beforeEach runs before each spec, replaced it with beforeAll as
* mounting a component is not required before each spec

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/materials_config_widget_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/materials_config_widget_spec.js
@@ -17,19 +17,9 @@
 define(["jquery", "mithril", "pipeline_configs/models/materials", "pipeline_configs/views/materials_config_widget"], function ($, m, Materials, MaterialsConfigWidget) {
   describe("Material Widget", function () {
     var $root, root;
-    beforeEach(function () {
-      root = document.createElement("div");
-      document.body.appendChild(root);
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails.new/spec/new_javascripts/pipeline_configs/views/materials_config_widget_spec.js
```diff
@@ -17,19 +17,9 @@
 define(["jquery", "mithril", "pipeline_configs/models/materials", "pipeline_configs/views/materials_config_widget"], function ($, m, Materials, MaterialsConfigWidget) {
   describe("Material Widget", function () {
     var $root, root;
-    beforeEach(function () {
```
ketan: @maheshp — the reason this was done in a `beforeEach` instead of `beforeAll` is because mithril "remembers" and caches the dom element and the component bound to it. If you're doing *any* UI interactions with the dom in a specific test case, the problem becomes worse. Depending on what order tests run — I sometimes noticed weird bugs because of such re-rendering different components on the same DOM element.

https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/mithril-0.2.0.js#L550

Are you sure the problem is caused by `beforeEach` vs `beforeAll` — I mean DOM rendering shouldn't be *that* slow?

zabil:
@ketan there are nested blocks in this test (with single asserts) which were split to improve readability and they were meant to depend operate on the remembered dom element.

PR #1874, opened 2016-02-02
Allow adding some metadata for jobs that should be run via elastic agent plugins (#1082)

ketan:
```xml
<job>
  <agentConfig pluginId="...">
    <property>
      <key>Dockerfile</key>
      <value>Dockerfile.example</value>
    </property>
  </agentConfig>
</job>
```

### DIFF
--- a/base/src/com/thoughtworks/go/util/GoConstants.java
+++ b/base/src/com/thoughtworks/go/util/GoConstants.java
@@ -58,7 +58,7 @@ public class GoConstants {
 
     public static final String PRODUCT_NAME = "go";
 
-    public static final int CONFIG_SCHEMA_VERSION = 78;
+    public static final int CONFIG_SCHEMA_VERSION = 79;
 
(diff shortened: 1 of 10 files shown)


ketan:
@arvindsv — we've currently set this to allow either `<resources/>` or `<agentConfig/>` but not both.

That look ok to you?

arvindsv:
Ok. Ideally, it's only one. But I guess it'll have to be something like this (if only one):

```
<agentConfig pluginId="go-server">
  <key>resources</key>
  <value>resource1 resource2 resource3</value>
</agentConfig>
```

which is a little icky, since it has a go-server plugin ID and resources are not separated. So, your option is fine, I guess.

ketan:
> which is a little icky, since it has a go-server plugin ID and resources are not separated. So, your option is fine, I guess.

It's also icky because "resources" is going to be a reserved-words of sorts for other purposes (agents page, apis etc)

arvindsv:
Ok, I understand.

Issue #1872, opened 2016-02-02
Secure variables value not masked when part of the environment variable output


ketan:
/cc @wpc 

Steps to reproduce (revision 36112fef22a7af7f88ad3e69b1e78bf815dd974c)

```xml
<?xml version="1.0" encoding="utf-8"?>
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="79">
  <server artifactsdir="logs" commandRepositoryLocation="default" serverId="dev-id">
  </server>
  <pipelines group="first">
    <pipeline name="up42">
      <environmentvariables>
        <variable name="foo">
          <value>password_foo</value>
        </variable>
        <variable name="secure" secure="true">
          <value>password</value>
        </variable>
      </environmentvariables>
      <materials>
        <git url="https://github.com/gocd/api.go.cd" />
      </materials>
      <stage name="up42_stage">
        <jobs>
          <job name="up42_job">
            <tasks>
              <exec command="ls" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </pipelines>
</cruise>
```

expected output —

```
16:04:29.879 [go] setting environment variable 'foo' to value '********_foo'
16:04:29.879 [go] setting environment variable 'secure' to value '********'
```

actual output —

```
16:04:29.879 [go] setting environment variable 'foo' to value 'password_foo'
16:04:29.879 [go] setting environment variable 'secure' to value '********'
```

wpc:
I am on this

arvindsv:
@ketan: I guess this can be closed, since the acceptance tests passed?

PR #1876, opened 2016-02-02
fix issue #1872: Secure variables value not masked

wpc:
This PR fixed issue #1872, which introduced by PR #1820


wpc:
@ketan, how could you merge faster than I send a message in gitter? Do you have a bot doing that? :-D

happy hacking
--wpc

PR #1886, opened 2016-02-03
Implement a fuller subset of python style string slicing for material revisions in pipeline label templates.

Drakekin:
This pull request adds a fuller subset of python style string slicing for material revisions in pipeline label templates. The material revision regular expression has been expanded to the following:

```
$\{[a-zA-Z0-9_\-\.!~'#:]+(\[(\-?\d+)?:?(\-?\d+)?\])\}
```

The slicing is used as follows:
- `${git[10:20]}` Selects the 11th character to the 20th.
- `${git[10:]}` Selects the 11th character onwards.
- `${git[:20]}` Selects up to the 20th character.
- `${git[20]}` Selects the 21st character.
- `${git[-5:]}` Selects the last five characters.
- `${git[-10:-5]}` Selects the five characters preceding the last five characters.
- `${git[-5]}` Selects the character five positions from the end.
- `${git[]}` Selects the entire string (for consistencies sake).
- `${git[-10:20]}` Selects the character between the tenth character from the end and the 21st character from the start. Whilst possible, it is uncommon to use this form.

Indexes out of bounds are truncated to the length of the string. Slices entirely outside of the bounds of a string are returned as an empty string.

This PR also refactors the tests to be more consistent, adds new test cases for the new slicing syntax and refactors the internal behaviour of `PipelineLabel.java` to be conceptually cleaner. The regex for matching materials has been cleaned up to use named groups to make it easier to read and make the code interfacing with it easier to reason about.

The `PipelineLabel` class now no longer performs an in-place replacement of the label template, instead it maintains a list of materials it has been passed internally which are replaced into the label template when the object is cast to a string.


arvindsv:
Closing. CLA is not signed and it got missed and we might need to look at this again, since there might be changes needed.

PR #1887, opened 2016-02-03
Shallow cloning for git material (rebased)

wpc:
This PR is rebased version of PR #1846, please find the original comments and discussion in #1846 
====

This PR implement the [proposal](https://github.com/gocd/gocd/issues/313#issuecomment-174103478) on issue #313. 

What it does
---
* Added a configuration option for git material called `shallowClone`, default to false
* Shallow clone config only applies to agent. On server side flyweight clone (includes configuration repository) is always shallow.
* If shallowClone is specified, git repository on agents clone will be a shallow clone with depth 2.
* If build is triggered with a missing revision on a shallow cloned repo, use a 2 steps process to unshallow the repo. 
* Convert existing shallow cloned repos to full clone when user turning shallowClone off.

How to test it
---
Since UI part of the shallow clone configuration is not included, enabling the configuration option need be through editing cruise-config.xml. The following xml snippet setup a pipeline checkout linux-kernel repository with shallow clone enabled.

```xml
<pipeline name="linux-kernel">
  <materials>
    <git url="git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git" shallowClone="true" materialName="linux-kernel" />
  </materials>
  <stage name="ls-stage">
    <jobs>
      <job name="ls-job">
        <tasks>
          <exec command="ls" />
        </tasks>
      </job>
    </jobs>
  </stage>
</pipeline>
```

Benefits
---
For the linux kernel repo,  on my machine/network,  without shallow clone, it takes more than 15 minutes to complete first build. With shallow clone option enabled, it takes about two minutes to complete first build.

Other notes
---
* This PR works with old git versions,  tested back to 1.7.1.
* This PR includes configuration migration (bumped schema version to 79)
* Server side shallow clone can be turn off by setting system property `go.server.shallowClone` to `N`
* This PR changed the way how ScheduledPipelineLoader lookup `knownMaterials`. It scopes material config lookup under the current pipeline. Set system property `go.server.scheduledPipelineLoader.globalMaterialLookup` to `Y` can bring the old behavior back (lookup material through whole cruise config)

### DIFF
--- a/common/src/com/thoughtworks/go/config/materials/Materials.java
+++ b/common/src/com/thoughtworks/go/config/materials/Materials.java
@@ -105,14 +105,6 @@ public class Materials extends BaseCollection<Material> {
         return revisions;
     }
 
-    public void updateTo(Revision revision, File baseFolder, ProcessOutputStreamConsumer outputStreamConsumer, final SubprocessExecutionContext execCtx) {
-        cleanUp(baseFolder, outputStreamConsumer);
-
-        for (Material material : this) {
-            material.updateTo(outputStreamConsumer, revision, baseFolder, execCtx);
-        }
-    }
-
     public void cleanUp(File baseFolder, ConsoleOutputStreamConsumer consumer) {
         if (hasMaterialsWithNoDestinationFolder()) {
             return;
--- a/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
@@ -23,8 +23,8 @@ import com.thoughtworks.go.domain.MaterialInstance;
 import com.thoughtworks.go.domain.materials.*;
 import com.thoughtworks.go.domain.materials.git.GitCommand;
 import com.thoughtworks.go.domain.materials.git.GitMaterialInstance;
-import com.thoughtworks.go.server.transaction.TransactionSynchronizationManager;
 import com.thoughtworks.go.domain.materials.svn.MaterialUrl;
+import com.thoughtworks.go.server.transaction.TransactionSynchronizationManager;
 import com.thoughtworks.go.util.GoConstants;
 import com.thoughtworks.go.util.StringUtil;
 import com.thoughtworks.go.util.command.InMemoryStreamConsumer;
@@ -41,7 +41,8 @@ import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import static com.thoughtworks.go.util.ExceptionUtils.*;
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+import static com.thoughtworks.go.util.ExceptionUtils.bombIfFailedToRunCommandLine;
 import static com.thoughtworks.go.util.FileUtil.createParentFolderIfNotExist;
 import static com.thoughtworks.go.util.FileUtil.deleteDirectoryNoisily;
 import static com.thoughtworks.go.util.command.ProcessOutputStreamConsumer.inMemoryConsumer;
@@ -52,9 +53,13 @@ import static java.lang.String.format;
  */
 public class GitMaterial extends ScmMaterial {
     private static final Logger LOG = Logger.getLogger(GitMaterial.class);
+    private static final int UNSHALLOW_TRYOUT_STEP = 100;
+    public static final int DEFAULT_SHALLOW_CLONE_DEPTH = 2;
+
 
     private UrlArgument url;
     private String branch = GitMaterialConfig.DEFAULT_BRANCH;
+    private boolean shallowClone = false;
     private String submoduleFolder;
 
     //TODO: use iBatis to set the type for us, and we can get rid of this field.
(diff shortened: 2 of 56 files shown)


common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
```diff
@@ -96,17 +115,18 @@ public GitMaterial(GitMaterialConfig config) {
...
         ArrayList<Modification> mods = new ArrayList<Modification>();
-        mods.add(getGit(baseDir).latestModification());
+        mods.add(getGit(baseDir, DEFAULT_SHALLOW_CLONE_DEPTH).latestModification());
         return mods;
     }
 
     public List<Modification> modificationsSince(File baseDir, Revision revision, final SubprocessExecutionContext execCtx) {
-        GitCommand gitCommand = getGit(baseDir);
+        GitCommand gitCommand = getGit(baseDir, DEFAULT_SHALLOW_CLONE_DEPTH);
+        unshallowIfNeeded(gitCommand, ProcessOutputStreamConsumer.inMemoryConsumer(), revision, baseDir);
```
jyotisingh: `modificationsSince` is invoked only by the server. In what cases do think we would need a `unshallowIfNeeded` on the server? `shallowClone` has now become a agent-side property so `unshallowIfNeeded` for agents makes sense. Wondering if I am missing some case.

jyotisingh:
While we are at it, I was wondering if we could also do `--no-checkout` on *server*. Flyweights consume a huge amount of disk space on Go servers. It would save a whole lot of disk space. 
Stats for gocd repo:
with history and files: ~1GB in size
with depth=1 and files: 340Mb
with depth=1 and no-checkout: 96Mb

I can't think of any reason why we need the files checked out on the server. What do you think?

common/src/com/thoughtworks/go/config/materials/Materials.java
```diff
@@ -105,11 +105,12 @@ public MaterialRevisions latestModification(File baseDir, final SubprocessExecut
         return revisions;
     }
 
+    //todo: move this to test
```
jyotisingh: you forgot the todo :)

server/src/com/thoughtworks/go/server/service/materials/GitPoller.java
```diff
@@ -23,22 +23,29 @@
...
     public List<Modification> latestModification(GitMaterial material, File baseDir, SubprocessExecutionContext execCtx) {
+        material = material.withShallowClone(isServerSideShallowCloneOn());
         return material.latestModification(baseDir, execCtx);
     }
-
     @Override
     public List<Modification> modificationsSince(GitMaterial material, File baseDir, Revision revision, SubprocessExecutionContext execCtx) {
-        return material.modificationsSince(baseDir, revision, execCtx);
+        material = material.withShallowClone(isServerSideShallowCloneOn());
+        return material.withShallowClone(true).modificationsSince(baseDir, revision, execCtx);
```
jyotisingh: The flag is being set to true always, the previous statement does the right thing - Did you leave this by mistake?

wpc:
> While we are at it, I was wondering if we could also do --no-checkout on server. Flyweights consume a huge amount of disk space on Go servers

@jyotisingh: I was thinking about the exact same thing when working on this! But to keep this PR scoped I created a separated issue #1890. Once this PR merged I can sending another PR for it.

common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
jyotisingh: Does this come into picture for cases where shallow-clone feature is toggled?

jyotisingh:
ah, did not see the other issue earlier. makes sense.

common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
wpc: @jyotisingh 

The main reason is to keep behavior consistency throughout GitMaterial implementation. The implementation should not assume how to it get used.  Make it consistent helps reduce surprises when usage context changed: e.g someday a developer need call modificationsSince somewhere else would wonder "why the hell I can not find a revision?"

Another (more practical) reason is to make it robust against external changes.  In cases of flyweight repo corrupted, user can just rm it and let go clone it again. Without `unshallowIfNeeded ` here, they will lost abilities of triggering old build.

server/src/com/thoughtworks/go/server/service/materials/GitPoller.java
wpc: Oops, a rookie mistake :-P. sorry about.

common/src/com/thoughtworks/go/config/materials/Materials.java
wpc: done :-)

server/src/com/thoughtworks/go/server/service/materials/GitPoller.java
wpc: done in e1a5901.

jyotisingh:
Reviewed the PR, looks alright to me. Pending:
* https://github.com/gocd/gocd/issues/313#issuecomment-186542685

@wpc - could you also squash the commits

wpc:
@jyotisingh 
>  Pending: #313 (comment)

Done and squashed.

Thanks for reviewing it.

-- wpc

wpc:
@jyotisingh
I have updated the PR since a config schema version 79 been occupied with @ketan's recent change. No conflicts anymore.

Issue #1850, opened 2016-01-28
Problem with/for Centos installer (no which)


arvindsv:
On a plain vanilla Centos, the "which" command has been removed from being installed by default. So, this happens:

```
[root@6824fbe7c0d5 /]# cat /etc/system-release
CentOS Linux release 7.2.1511 (Core)
[root@6824fbe7c0d5 /]# # yum install wget java-1.6.0-openjdk # Has happened
[root@6824fbe7c0d5 /]# wget -q https://download.go.cd/gocd-rpm/go-agent-16.1.0-2855.noarch.rpm
[root@6824fbe7c0d5 /]# rpm -i go-agent-16.1.0-2855.noarch.rpm
/var/tmp/rpm-tmp.SAiZuH: line 153: which: command not found
Path to java not found in JAVA_HOME or PATH.
WARNING: Unable to determine JAVA_HOME. Not setting or updating it in "/etc/default/go-agent". You might need to set it manually, after this.
Installation of Go Agent completed.
Now please edit /etc/default/go-agent and set GO_SERVER to the IP address of your Go Server.
Once that is done start the Go Agent with '/etc/init.d/go-agent start'
[root@6824fbe7c0d5 /]#
```

After installing which:

```
[root@6824fbe7c0d5 /]# yum install -y which >/dev/null
[root@6824fbe7c0d5 /]# rpm -i go-agent-16.1.0-2855.noarch.rpm
'Go Agent' installation will now use group 'go'
'Go Agent' installation will now use user 'go'
Found Java /usr/lib/jvm/java-1.6.0-openjdk-1.6.0.38.x86_64/jre in PATH, using it.
Installation of Go Agent completed.
Now please edit /etc/default/go-agent and set GO_SERVER to the IP address of your Go Server.
Once that is done start the Go Agent with '/etc/init.d/go-agent start'
```

ketan:
Steps to reproduce —

```
$ docker run -it centos:6 bash
```

Inside the docker container —
```
[root@1726e944fc10 /]# curl -L -s https://download.go.cd/gocd-rpm/go-agent-16.1.0-2855.noarch.rpm -O
[root@1726e944fc10 /]# curl -L -s https://download.go.cd/local/16.2.0-3018/go-agent-16.2.0-3018.noarch.rpm -O
[root@1726e944fc10 /]# rpm -ivh go-agent-16.1.0-2855.noarch.rpm
Preparing...                ########################################### [100%]
[...]

PR #1893, opened 2016-02-04
Add which as a dependency (#1850)

ketan:


jyotisingh:
`which` is used to determine  value of JAVA_HOME if its not set already. If JAVA_HOME is set then we do not even need `which`. On the other hand if JAVA_HOME wasn't set and `which` not installed, the only thing that would happen is that `JAVA_HOME` wouldn't be set and so Go wouldn't startup automatically. However, user could install java post Go's install, set JAVA_HOME appropriately and things would start rolling again. With this PR, if `which` is not installed, Go's installation would fail. That doesn't sound right.

ketan:
> With this PR, if which is not installed, Go's installation would fail. That doesn't sound right.

If you're installing via a package manager (yum) that would not happen. In most cases, which is installed, and the package won't fail. However in cases of minimal CentOS installs like on docker, this will pull down which.

In either case, if a package is required by GoCD, it should be declared it in its dependency.

PS: Arguably, on debian packages, we define the jdk as a dependency, which will also cause a similar problem that you describe.

arvindsv:
Yeah. Since it fails in certain cases, as mentioned in #1850, I think we will need to add this dependency.

arvindsv:
Need to ensure that "which" is a package on other RPM distributions, though. Like RedHat/Fedora.

ketan:
I checked it is.

ketan:
> Need to ensure that "which" is a package on other RPM distributions, though. Like RedHat/Fedora.

To add to that, `Requires: foo` can be used to mean install package `foo` or install package that "provides" `foo`.

But just in case -

* http://rpm.pbone.net/index.php3?stat=3&search=which&srodzaj=3
* https://www.rpmfind.net/linux/rpm2html/search.php?query=which

PR #1895, opened 2016-02-04
Get rid of the empty licensing module

ketan:


jyotisingh:
Changes look alright, assuming you have compiled the code :) so merging it

PR #1898, opened 2016-02-05
Security bug: when editing svn material, plain password is printed in html

xli:
Change to output "" instead of svn material password.

p4 and tfs materials have similar problem, changed together.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/views/admin/materials/shared/_password.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/admin/materials/shared/_password.html.erb
@@ -1,7 +1,7 @@
 <div class="form_item_block">
     <label><%= l.string('PASSWORD') -%></label>
     <%= scope[:form].hidden_field(com.thoughtworks.go.config.materials.ScmMaterialConfig::ENCRYPTED_PASSWORD, {:include_blank => true, :class => "form_input encrypted_password"}) -%>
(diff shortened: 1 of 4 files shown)


arvindsv:
Looks good to me. I checked the functionality too and it seems to work, adding and editing materials (SVN, TFS, Perforce).

arvindsv:
Maybe not. I think the password is being set to ```******``` when it is saved.

arvindsv:
No. I'm wrong. Ignore that. Still looks good to me, at this point.

xli:
Build failed by test "KillAllChildProcessTaskBuilderTest.shouldKillAllChildProcessOnbuild", should not related to this PR changes

ketan:
That's caused by JDK8 -
https://github.com/gocd/javasysmon/commit/c53cb0301c321d86ce9565148068a28fcdfa8739

zabil:
Still working of fixing it as the commit has to check a lot of system
dependent stuff.

Meanwhile I'll set JAVA_HOME on PR pipelines to point to JDK7.

On Fri, Feb 5, 2016 at 11:26 AM, Ketan Padegaonkar <notifications@github.com
> wrote:

> That's caused by JDK8 -
>
> https://github.com/gocd/javasysmon/commit/c53cb0301c321d86ce9565148068a28fcdfa8739
>

Issue #1903, opened 2016-02-08
Lodash Upgrade


maheshp:
Upgrade lodash to latest version

PR #1904, opened 2016-02-08
lodash upgraded to 4.2.1 #1903

maheshp:
Build: `lodash -o ./dist/lodash.js`
https://raw.githubusercontent.com/lodash/lodash/4.2.1/dist/lodash.js


ketan:
Can you add a note about what specific lodash distro this is (and the URL where you fetched this). It's useful for archeologists excavating fossils.

PR #1906, opened 2016-02-09
Add system properties to disable gadget and gadget rendering server (#1908)

ketan:


maheshp:
> I did not want to make too many changes, will submit a PR to make this change post 16.2.

Ok, we can do it later.

PR #1922, opened 2016-02-11
Move to Foundation 6

maheshp:


ketan:
Issues — 

* Tooltip text inherits the style from `<a/>` (should be black instead)
    screen shot 2016-02-15 at 8 10 20 pm

* extra semicolon at end of page 
    screen shot 2016-02-15 at 8 28 43 pm

* on a page load, default tab is not shown as selected (something to do with `[aria-selected='true']` not being applied, see `_tabs.scss:90`)
    screen shot 2016-02-15 at 8 34 01 pm

PR #1937, opened 2016-02-15
Fixing problem with windows dll versions.

zabil:
32 bit dll's were compiled as AMD64 instead of i386 leading to problem
while loading it on 32 bit JVM's

The compiled jar is taken from build
https://build.go.cd/go/tab/build/detail/javasysmon/6/package/1/jar


zabil:
Closing as this is now branch.

PR #1951, opened 2016-02-18
Remove metrics service

ketan:


jyotisingh:
This seems fine. In addition to this `applicationContext-acegi-security.xml`(https://github.com/gocd/gocd/blob/6c65033cdc9569c002f0361e6396bc83744efb8b/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L298-L299) and `web.xml`(https://github.com/gocd/gocd/blob/6c65033cdc9569c002f0361e6396bc83744efb8b/server/webapp/WEB-INF/web.xml#L191-L200) needs to be updated too.

PR #1952, opened 2016-02-18
Removing unused end points

zabil:


server/src/com/thoughtworks/go/server/controller/GoConfigAdministrationController.java
```diff
@@ -105,18 +73,6 @@ public void getConfigRevision(@RequestParam(value = "version", required = true)
...
-    public void getGroupAsXmlPartial(@RequestParam("pipelineGroup")String groupName,
-                                     @RequestParam(value = "md5", required = false)String md5,
-                                     HttpServletResponse response) throws Exception {
-        if (isTemplate(groupName)) {
-            getXmlPartial(groupName, md5, goConfigService.templatesSaver()).respond(response);
-            return;
-        }
-        getXmlPartial(groupName, md5, goConfigService.groupSaver(groupName)).respond(response);
-    }
-
```
ketan: Can you check if the methods on GoConfigService are used? `buildSaver` `stageSaver` and the likes?

jyotisingh:
There are a whole lot of deprecated implementations of `XmlPartialSaver` which could be deleted as well. The only usages for those would have been from `GoConfigAdministrationController`

zabil:
@ketan @jyotisingh 
Now bigger and better. Running this through a pr build.

ketan:
:+1:

PR #1953, opened 2016-02-18
Fixing rspec specs

maheshp:


server/webapp/WEB-INF/rails.new/app/views/navigation_elements/_navigation.html.erb
```diff
@@ -12,24 +12,31 @@
...
-            <%- elsif is_user_a_group_admin? -%>
-              <%= render partial: 'navigation_elements/dropdown_pipeline_group_admin' -%>
-            <%- elsif is_user_a_template_admin? -%>
-              <li role="presentation">
-                <%= link_to 'Templates', templates_path -%>
-              </li>
-            <%- end -%>
-          </ul>
+          <%
+             admin_header_link_label = "Admin"
```
ketan: Do we really need a variable for a string that's used in one place :)

maheshp: Not required, it looks stupid

Issue #1954, opened 2016-02-18
Proposal: New BuildCommand based Server-Agent Communication 


wpc:
# What this is about

Currently interface between server and agent are based on serialization on a big graph of java objects, rooted by BuildWork. There is quite a lot cons on this approach, and the majority one IMHO there is no clear boundary between server and agent, making agent implementation tightly coupled with Java platform and complex domain models on go server side.

During @xli's work on #1793, we found we actually can give it a very clean abstraction cut. We propose a BuildCommand based API between server and agent.
# BuildCommand

Instead of sending BuildWork from server to agent. We convert BuildWork to a tree of composed commands and send to agent. The responsibility of agent becomes to just processing commands:

![buildcommand api](https://cloud.githubusercontent.com/assets/10283/13155703/c754283e-d633-11e5-8268-a0014c932859.png)

BuildCommand is made from 5 elements:
- Name: name of the command
- Arguments [a list of String or Map]: arguments for the command, each command knows how to read
  and use
- WorkingDirectory [String]: a relative path to agent root directory
  for setting up command's working directory
- Test [BuildCommand, expectation]: command precondition, run command only if test result is
  expected. Default to none.
- RunIfConfig [String: "passed", "failed", "any"]: same concept with
  Go's runIfConfig configuration, run command based on current build status (pass/fail). "passed" is default.
### Atomic BuildCommand List

Basing on our spike, we found the following command list is enough to
do all the work current Go support:
- start: prepare environment for running a new list of build commands
- compose: compose a list of commands. Agent expands it and process
  each sub-commands.
- echo: output a string to console log that end user will see.
- export: setup environment variables for the rest of following
  commands. If no argument provide, it dumps all environment variables to the console.
- exec: execute a external (bash/shell) command.
- test: test file/directory existence. Same with Unix test command.
- generateProperty: extract property and value, upload to server.
- downloadFile: download a file.
- downloadDir: download a directory from server.
- uploadArtifact: upload artifact to server.
- generateTestReport: generate test report and upload to server.
- callExtension: call JSON based plugin extension, including scm and task
  plugins.
- callAPIBasedTaskExtension: call API based task extension.
- end: end of a build, agent will clean up build environment and
  change its status back to idle.
- reportCurrentStatus: report current build status to server.
- reportCompleting: report completing status to server.
- reportCompleted: report build completed to server.

See
[...]

PR #1962, opened 2016-02-19
Websocket communication (#1793): change back to use Java serialization

### LINKED ISSUE #1958: Environment variable GO_FROM_REVISION always set to the latest revision
Issue observed on 16.3.0-3055(2dcbbfe1a7b44897bc026b845c6b0822a27858a8), not observed on 16.2.1-3027

Steps to reproduce:

- Create a pipeline with git material(or any SCM)
- Add/modify files in git repo to create multiple revisions
- Trigger pipeline
- Check the value of GO_FROM_REVISION on the console
- The value is same as the GO_TO_REVISION even though there are more than one revision
- Expect the GO_FROM_REVISION to reflect the exact revision number from which the material update starts

xli:
We will change to JSON format when we have better protocal designed for agent and server communication (proposed by #1954).
Also remove all JSON serialization related optimization. Serialize BuildWork same with before, reduce the chance having problem.

This change should fix #1958 when websocket is enabled.


common/src/com/thoughtworks/go/websocket/Message.java
```diff
@@ -16,27 +16,24 @@
...
 
     public static byte[] encode(Message msg) {
-        String encode = JsonMessage.encode(msg);
         ByteArrayOutputStream bytes = new ByteArrayOutputStream();
         try {
-            GZIPOutputStream out = new GZIPOutputStream(bytes);
-            out.write(encode.getBytes(StandardCharsets.UTF_8));
-            out.finish();
+            GZIPOutputStream zip = new GZIPOutputStream(bytes);
+            ObjectOutputStream output = new ObjectOutputStream(zip);
```
jyotisingh: Need to `close` these streams

ketan:
@xli and @wpc — next time you update a PR to address any feedback, please be sure to leave a note. GitHub does not send a notification when a PR is updated and it gets difficult for the original reviewer to know if the PR was updated without visiting the PR page.

PR #1963, opened 2016-02-19
Websocket communication: fix issue: server may assign multiple cookies to same agent

xli:
It is possible that server receives multiple same agent ping messages due to network issues or agent ping bug, no matter what we do at agent side.
For identifying an agent that uses duplicated uuid, we have cookie system in place. 
This change will make sure we assign only one cookie to an agent that is connected by websocket.

In websocket case, we don't need cookie, because websocket itself can be used to identify 2 same UUID agents.
But before we remove old server agent communication code, we need maintain it.
We can clean up cookie stuff after we switched over to websocket.


server/src/com/thoughtworks/go/server/websocket/AgentRemoteHandler.java
```diff
@@ -88,6 +102,10 @@ public void process(Agent agent, Message msg) {
     }
 
     public void remove(Agent agent) {
+        if (agent == null) {
```
ketan: We don't seem to be passing in a `null` anywhere, can we remove this?

ketan:
Addressed this [comment](https://github.com/gocd/gocd/pull/1963#discussion_r53739173) and submitted pr #1972 instead.

PR #1964, opened 2016-02-20
Fix working directory disallow hidden folder and a relative path start with "./" issue

### LINKED ISSUE #1900: Working Directory Pattern
We have build related files abstracted in a hidden folder `.folder-name`, but the working directory file pattern does not match preceding `.`. This is presumably to avoid `./`, may i suggest something like this instead?

`\.?(?!\/)(([^. ].+[^. ])|([^. ][^. ])|([^. ]))`

I haven't been able to test it other than using http://fiddle.re/5qmhx6 . I can definitely setup the development environment at some point, but no time at the moment so i'll just log the issue.

xli:
Fix #1900 about hidden directory as working directory issue.

Example:
  allow ".hidden" as working directory.
  allow "./.hidden" as working directory.
  allow "./hello-working-dir" as working directory.

### DIFF
--- a/base/src/com/thoughtworks/go/util/GoConstants.java
+++ b/base/src/com/thoughtworks/go/util/GoConstants.java
@@ -58,7 +58,7 @@ public class GoConstants {
 
     public static final String PRODUCT_NAME = "go";
 
-    public static final int CONFIG_SCHEMA_VERSION = 80;
+    public static final int CONFIG_SCHEMA_VERSION = 81;
 
     public static final String APPROVAL_SUCCESS = "success";
     public static final String APPROVAL_MANUAL = "manual";
--- a/common/test/unit/com/thoughtworks/go/domain/ArtifactPlanTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/ArtifactPlanTest.java
@@ -20,6 +20,7 @@ import com.rits.cloning.Cloner;
 import com.thoughtworks.go.config.ArtifactPlan;
 import com.thoughtworks.go.config.ConfigSaveValidationContext;
 import com.thoughtworks.go.config.JobConfig;
+import com.thoughtworks.go.config.validation.FilePathTypeValidator;
 import com.thoughtworks.go.util.ClassMockery;
 import com.thoughtworks.go.work.DefaultGoPublisher;
 import org.jmock.Expectations;
@@ -139,7 +140,7 @@ public class ArtifactPlanTest {
    public void validate_shouldFailIfDestDoesNotMatchAFilePattern() {
         ArtifactPlan artifactPlan = new ArtifactPlan("foo/bar", "..");
         artifactPlan.validate(null);
-        assertThat(artifactPlan.errors().on(ArtifactPlan.DEST), is("Invalid destination path. Destination path should match the pattern ([^. ].+[^. ])|([^. ][^. ])|([^. ])"));
+        assertThat(artifactPlan.errors().on(ArtifactPlan.DEST), is("Invalid destination path. Destination path should match the pattern (([.]\\/)?[.][^. ]+)|([^. ].+[^. ])|([^. ][^. ])|([^. ])"));
     }
 
     @Test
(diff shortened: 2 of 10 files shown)


config/config-server/resources/cruise-config.xsd
```diff
@@ -808,7 +808,7 @@
     </xsd:element>
     <xsd:simpleType name="filePathType">
         <xsd:restriction base="xsd:string" xml:space="default">
-            <xsd:pattern value="([^. ].+[^. ])|([^. ][^. ])|([^. ])"/>
+            <xsd:pattern value="(([.]\/)?[.][^. ]+)|([^. ].+[^. ])|([^. ][^. ])|([^. ])"/>
```
ketan: @xli I think this will need versioning — see [this commit](https://github.com/gocd/gocd/commit/6b26201e18a19b74d38a84d02533e062254f1dea) on how it's done.

Please also squash the PRs.

common/test/unit/com/thoughtworks/go/domain/ArtifactPlanTest.java
```diff
@@ -139,7 +140,7 @@ public void validate_shouldFailIfSourceIsEmpty() {
    public void validate_shouldFailIfDestDoesNotMatchAFilePattern() {
         ArtifactPlan artifactPlan = new ArtifactPlan("foo/bar", "..");
         artifactPlan.validate(null);
-        assertThat(artifactPlan.errors().on(ArtifactPlan.DEST), is("Invalid destination path. Destination path should match the pattern ([^. ].+[^. ])|([^. ][^. ])|([^. ])"));
+        assertThat(artifactPlan.errors().on(ArtifactPlan.DEST), is("Invalid destination path. Destination path should match the pattern " + FilePathTypeValidator.PATH_PATTERN));
```
ketan: Better error message please :) Something along the lines of "must be a relative path" reads better than a regular expression.

base/src/com/thoughtworks/go/util/GoConstants.java
```diff
@@ -58,7 +58,7 @@
 
     public static final String PRODUCT_NAME = "go";
 
-    public static final int CONFIG_SCHEMA_VERSION = 79;
+    public static final int CONFIG_SCHEMA_VERSION = 80;
```
ketan: Just an FYI — you might conflict with WPC here. So I suggest bumping this to 81 and using @wpc's [cruise-config.xsd](https://github.com/gocd/gocd/pull/1887/files#diff-bdfabb73e2af551ffd8e50a573a66338) instead.

We'll make sure to merge yours after @wpc's or the other way around if @wpc prefers that.

Just let us know :)

ketan:
@xli — there was an interesting bug that came in today.

Someone was attempting to use a file/folder `..foobar`, can you test for that too, or if there's a better way we're open to it. But if a better way of solving this takes longer we'll do it in another PR.

xli:
@ketan is "..foobar" valid or invalid?

xli:
@ketan do you know why we have this validation? why disallow user to use a path outside of pipeline directory? If they want it, they can do it anyway right (e.g. in command line, or build script) ?

xli:
For example, user can do it like the following:

edit_job_-_go

xli:
Existing validation is more of restriction in some format, but not really do the job.
I found ExecTask.java#133 does real validation with file system. But regex based config validation won't work well right now.

ketan:
The intent of that feature (and regex) is to prevent users from specifying a path that is outside the working directory for the pipeline so things like `../../../etc/passwd` are not allowed. I've no idea why — but your question is legit.

Even if we normalize a path to ensure that it's inside the working directory, there's nothing preventing someone from either moving files into the working directory before having them uploaded, or symlinking to files from outside the working directory.

@arvindsv or @jyotisingh would probably know better :)

arvindsv:
> @arvindsv ... would probably know better :)

He doesn't. :) This was before my time. Maybe @jyotisingh knows.

I'm guessing the reason is some assumption of "security". It could also be to force people to get into the habit of not relying on absolute paths -- which is a little bit more of a good reason. To keep it repeatable across agents and builds.

xli:
@ketan I've rebased and updated with versioning

xli:
@ketan is this good to merge?

PR #1975, opened 2016-02-23
Accept application specific headers to handle properties.

zabil:
Escape user defined properties.

### DIFF
--- a/common/src/com/thoughtworks/go/publishers/GoArtifactsManipulator.java
+++ b/common/src/com/thoughtworks/go/publishers/GoArtifactsManipulator.java
@@ -201,7 +201,7 @@ public class GoArtifactsManipulator {
     public void setProperty(JobIdentifier jobIdentifier, Property property) {
         try {
             String propertiesUrl = urlService.getPropertiesUrl(jobIdentifier, property.getKey());
-            httpService.postToUrl(propertiesUrl, property.getValue());
+            httpService.postProperty(propertiesUrl, property.getValue());
         } catch (Exception e) {
             throw new ArtifactPublishingException(format("Failed to set property %s with value %s", property.getKey(), property.getValue()), e);
         }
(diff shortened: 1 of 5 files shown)


common/src/com/thoughtworks/go/util/HttpService.java
```diff
@@ -112,14 +112,15 @@ public int download(String url, FetchHandler handler) throws IOException {
...
 
-    public void postToUrl(String url, String value) throws IOException {
+    public void postProperty(String url, String value) throws IOException {
         LOGGER.info("Posting property to the URL " + url + "Property Value =" + value);
-        PostMethod propertyPost = httpClientFactory.createPost(url);
+        PostMethod post = httpClientFactory.createPost(url);
         try {
-            propertyPost.setRequestBody(new NameValuePair[]{new NameValuePair("value", value)});
-            execute(propertyPost);
+            post.setRequestHeader("Accept","application/vnd.go.cd.v1+text");
```
ketan: Can you submit a PR to the api docs site?

common/src/com/thoughtworks/go/publishers/GoArtifactsManipulator.java
```diff
@@ -201,7 +201,7 @@ private String getDestPath(String file) {
     public void setProperty(JobIdentifier jobIdentifier, Property property) {
         try {
             String propertiesUrl = urlService.getPropertiesUrl(jobIdentifier, property.getKey());
```
ketan: There's probably another issue here where we're building the url by concatenating the keys. Do we need URL escaping? (probably not, given that nobody has complained about it so far)

zabil: isn't this call in the chain good enough?
https://github.com/gocd/gocd/blob/69a39af5c0d7202b505cdf5578cebcff85643103/common/src/com/thoughtworks/go/domain/JobIdentifier.java#L213

PR #1981, opened 2016-02-24
Adding a simple token generator that validates post requests

zabil:


server/webapp/WEB-INF/vm/admin/admin-tab.vm
```diff
@@ -1,130 +0,0 @@
-#*************************GO-LICENSE-START*********************************
```
jyotisingh: Since this file is deleted, we can also delete `admin-tab.xml` and  its reference from `spring-tabs-servlet.xml`

zabil: Modified the pull request.

Issue #1890, opened 2016-02-04
Make git material clone use -n (--no-checkout) option


wpc:
During working the last PR(#1887), I found on server side there is no place use git repo's working copy (except configuration repo, which does explicit check out anyway).  And on agent side, a working copy of a  specific revision need be checked out when a build work comes, so initial HEAD checkout is a waste as well. For this I propose we use -n (--no-checkout) option for cloning. There are two benefit we can get:

**1. Saving disk space on server side**

For large repo, the working copy occupies large amount of disk space. But size of the repo itself (.git folder) could be just a fraction of the working copy size, especially when repo is shallow cloned. Take linux-kernel repository as example (as of revision b37a05c083c85c2657dca9bbe1f5d79dccf756d5)

| clone options | size |
|---|---|
| --branch master  |1.8G |
| --branch master -n | 1.1G |
| --branch master --depth 2 | 863M |
| --branch master --depth 2 -n | 151M | 	

On another repo: Thoughtworks Mingle,  I got similar percentage of size differences with different clone option: full clone with working copy is 2.1G vs 259M using shallow no-checkout clone.

**2. Saving clone time both on server side and agent side**

Checking out large amount files take considerable time, even on my SSD disk. Same linux kernel repo:

| clone options | time (minutes:seconds) |
|---|---|
| --branch master  |  5 minutes 11 seconds  |
| --branch master -n | 5 minutes 3 seconds |
| --branch master --depth 2 | 31 seconds |
| --branch master --depth 2 -n | 22 seconds | 	

On agent side, even working copy need be checkout eventually, directly reset to a specific revision is obviously more efficient than checkout HEAD then reset to a revision. 

```bash
% time sh -c 'git clone --branch master --depth 2 file:///tmp/linux.git checkout; cd checkout; git reset --hard b37a05c083c85c2657dca9bbe1f5d79dccf756d5'
sh -c   46.47s user 12.27s system 161% cpu 36.331 total

% time sh -c 'git clone --branch master --depth 2 -n file:///tmp/linux.git checkout; cd checkout; git reset --hard b37a05c083c85c2657dca9bbe1f5d79dccf756d5'
sh -c   46.24s user 10.66s system 188% cpu 30.215 total
``` 
In the above example, HEAD actually is b37a05c083c85c2657dca9bbe1f5d79dccf756d5.

What do you guys think? If there is no bad implications, I will start working on this after #1887 is merged.

arvindsv:
[...]

PR #1983, opened 2016-02-24
Make git material checking out working copy only when necessary

wpc:
For #1890. This PR helps saving disk space on server and speedup checking out both on server and agent.

### DIFF
--- a/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
+++ b/common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
@@ -59,7 +59,11 @@ public class GitCommand extends SCMCommand {
     // Clone repository from url with specified depth.
     // Special depth 2147483647 (Integer.MAX_VALUE) are treated as full clone
     public int cloneFrom(ProcessOutputStreamConsumer outputStreamConsumer, String url, Integer depth) {
-        CommandLine gitClone = git().withArg("clone").withArg(String.format("--branch=%s", branch));
+        CommandLine gitClone = git()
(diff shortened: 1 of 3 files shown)


arvindsv:
Looks ok to me. This should affect only new clones. Is there a verification missing in the tests?

common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
```diff
@@ -59,7 +59,11 @@ public int cloneFrom(ProcessOutputStreamConsumer outputStreamConsumer, String ur
     // Clone repository from url with specified depth.
     // Special depth 2147483647 (Integer.MAX_VALUE) are treated as full clone
     public int cloneFrom(ProcessOutputStreamConsumer outputStreamConsumer, String url, Integer depth) {
-        CommandLine gitClone = git().withArg("clone").withArg(String.format("--branch=%s", branch));
+        CommandLine gitClone = git()
```
ketan: This feels a bit odd in that it claims to say clone, and does not behave like `git clone` can we make the `noCheckout` a boolean arg?

Also, don't we need a clone with checkout on the agent?

arvindsv: Yeah, you're right (about the agent).

wpc: @ketan 
On agent we always do a hard reset to a specific revision, so the initial working copy checkout is not needed and just wasting time.

PR #1985, opened 2016-02-25
Revert "Enable agent websocket communication by default"

arvindsv:
Reverts gocd/gocd#1982

Seems like the Smoke test failed. cc @wpc


wpc:
cc @xli

PR #1990, opened 2016-02-25
Fix agent sending websocket message blocking error and wrong order issue

xli:
Add synch sending message support, server will send back an ack message if a message had ackId.
All agent messages will be sent with ack id, hence agent report current job status will wait until server responsed message is processed. The reason we need this change is because we need save job state in right order (preparing, building, completing and completed) on server side when agent reported status.

Notice: this pull request has a change to enable agent websocket communication by default (a756b0a). Please only revert change a756b0a when AC builds continue to fail after merged.


xli:
@ketan updated as you suggested.

PR #1991, opened 2016-02-26
Clean up KeyPair generation

wpc:
Currently we do weird thing on keypair generation.  In https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/security/KeyPairCreator.java after generateKeyPair via KeyPairGenerator, we recreate public key and private key again using KeyFactory with exactly same algorithm (RSA) and same modulus, public/private exponent. 

I dig up the history a little bit and found out why it is coded in this way.
We used to to use sun.security.rsa.RSAKeyPairGenerator to generate a KeyPair with sun implementation (RSAPublicKeyImpl and RSAPrivateCrtKeyImpl), then we need use KeyFactory to convert them to BouncyCastle keys for later processing. Later when we working on moving away from sun.security packages, we changed to use KeyPairGenerator to generate BouncyCastle keys directly, but forget remove the later conversion part.

Here is the original commit:
```
commit da34fd6de7237715ab9cb2755f3f597369a55a4e
Author: duck & jyoti <jsingh@thoughtworks.com>
Date:   Tue Oct 9 14:44:53 2012 +0530

    #6541 - removing sun specific KeyPairGenerator

diff --git a/common/src/com/thoughtworks/cruise/security/KeyPairCreator.java b/common/src/com/thoughtworks/cruise/security/KeyPa
irCreator.java
index 3c1352d..b482d94 100644
--- a/common/src/com/thoughtworks/cruise/security/KeyPairCreator.java
+++ b/common/src/com/thoughtworks/cruise/security/KeyPairCreator.java
@@ -2,18 +2,18 @@ package com.thoughtworks.cruise.security;

 import java.security.KeyFactory;
 import java.security.KeyPair;
+import java.security.KeyPairGenerator;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
 import java.security.spec.RSAPrivateKeySpec;
 import java.security.spec.RSAPublicKeySpec;

-import sun.security.rsa.RSAKeyPairGenerator;
 import static com.thoughtworks.cruise.util.ExceptionUtils.bomb;

 public class KeyPairCreator {
     public static KeyPair createKeyPair() {
         try {
-            KeyPair seed = new RSAKeyPairGenerator().generateKeyPair();
+            KeyPair seed = KeyPairGenerator.getInstance("RSA", "BC").generateKeyPair();
             RSAPrivateKey privateSeed = (RSAPrivateKey) seed.getPrivate();
             RSAPublicKey publicSeed = (RSAPublicKey) seed.getPublic();
             KeyFactory fact = KeyFactory.getInstance("RSA", "BC");
```

The key conversion process has some bad side effect. Private key’s CRT factors get lost during the process. They are suppose to be optional according the standard, but we need keep them because:

1. Leaving CRT factors there helps accelerating RSA decrypting and encrypting algorithm. (They are intermediate calculation result, keep them avoid calculate them again and again)
2. Some platform not fully PKCS#1 complaint assume they are always there. (I just submitted a issue for golang https://github.com/golang/go/issues/14512).  Without them alternative go-agent implementation on other platform could have problem. 

cc: @arvindsv  and @jyotisingh,  need your confirmation on my guess for the historical reason.

-- wpc


arvindsv:
I think I see what you're saying. I put in a couple of print statements to see the values:

```
public static KeyPair createKeyPair() {
    try {
        KeyPair seed = KeyPairGenerator.getInstance("RSA", "BC").generateKeyPair();
        RSAPrivateKey privateSeed = (RSAPrivateKey) seed.getPrivate();
        RSAPublicKey publicSeed = (RSAPublicKey) seed.getPublic();
        KeyFactory fact = KeyFactory.getInstance("RSA", "BC");
        RSAPrivateKeySpec privateKeySpec = new RSAPrivateKeySpec(
                privateSeed.getModulus(),
                privateSeed.getPrivateExponent()
        );
        RSAPublicKeySpec publicKeySpec = new RSAPublicKeySpec(
                publicSeed.getModulus(),
                publicSeed.getPublicExponent()
        );
        KeyPair pair = new KeyPair(fact.generatePublic(publicKeySpec), fact.generatePrivate(privateKeySpec));

        System.err.println("\nFirst seed key");
        System.err.println(ToStringBuilder.reflectionToString(seed) + "\n");
        System.err.println(ToStringBuilder.reflectionToString(seed.getPrivate()));

        System.err.println("\nFinal generated pair");
        System.err.println(ToStringBuilder.reflectionToString(pair) + "\n");
        System.err.println(ToStringBuilder.reflectionToString(pair.getPrivate()));

        return pair;
    } catch (Exception e) {
        throw bomb("Couldn't create public-private key pair", e);
    }
}
```

and I see this:

##### First seed key

```
java.security.KeyPair@6871729c[privateKey=RSA Private CRT Key
            modulus: af40d37186a8db2d3877f527cce0d70e9a623cdd58c447247461f8b82d497efc35d39705988a8d7bd4f9487f1cf8b190de58f8e758808f9eedd1a899570ef004a2aab95e272d8cb57ff02973e9615beecddaeeca5af5f303f95d64260977ed9755f38b35f642dd6799d619ab1fedbcd6e3ed35701e36d28f3012d90791c5cf8978c16ab1ac51822bba617ef169f08c231264c2c39aefc7f5cd0cae734a5671a8005686be24358d5c45f1751623be7293db69c91c21bdfacd2aa494ff9d72ce740907981e31ba9bd9ee1ca180d523c35199bd8871a87d09dd517a44d6f2bb57a8c763e697c7c297571207993f9ffbced140bc21d6aa6e129cd071aa0bf19f474f
    public exponent: 10001
   private exponent: 7e4afd1506b190a119dc7872e215f564b4a4aeafe4542de07b6e4eca8af837e690f396159643e289a684ae55cc0efc0725f4800a57a0eddede0b04b0dc280b978b91f45b816835ca6614300b33d7d1645a3cb77063f471d81f4fc9e80f52b989218b8c7aaaa8acdd07e35e440f92474fa49940b53750d2941554bfc9d0a6f8532bc57601ea4674f9570cd30399c62b6f4554158ae35027063f04182bebe6ec75c7cc25b906648fa6cb2e869a96486eae26ef3b157ccd9d8c2562e40cc506a4fc56fc3bc773ef74694116a1c13a5481fab4dc2f2f854801c43d8fc0d1f919af580e06f757b9a1e6456751feb41468984de45d1bca2c2515a6ef10b325229e6bc1
             primeP: d4847e2cc30c23d8d8e16e964e20cc9bb4ef676548f0ead06c0cc5294905b5a1c3218c969a95d2e0514490dd753fa71d0b99ce582d69dc3654be73c09e12e1138d02dc79b059617ed0d9c9426234a22416840401e137416ef4dcbc831453febdd5e5f4271e4f09f58c845002664990c3d32f02b2ba0c5594d5aaf6c2055197e7
             primeQ: d31c74cb553ede2f156ac2cc00b895ce0045bfce26d4f094f195a6e6075ef4b818779381581264ccdb9a8f7b5eed6ede23c2015fe470c53a587ec384336a31928c6831fc2552df5fca1c6e2d14c7fde7c4897b01aca64b7f6ddfd2c640bbd175e56b9e4706ef8253ae04a6e4f669c6f0801dd3ab475a3f94267220dc2323c859
     primeExponentP: b0d5b2bc026ce6a3390c945da3932d4e5f475a6cf03defd61022ac288767a5d47daf3bf1d19ddbc3e63bbca35a3de60b882caa83bafe3c4c4545d8c1ad871e1c8b9896e6f733d8a99295b61f73e2420cef6358864be6139f8546ebe1d297760076ad0af5460614ce6ae76a4024e70ffbb28d8a9c37adadc4d3ea97bba87b557f
     primeExponentQ: d2d169f8dbca0a4aee924cb1384f1c9a393f6e094a51ff665dd25a7700095d4cb2724705359438c132d42850c53d4a27df818fa9d9109c846577417c3573e8c936199f40f36cccd0759eea8f6e5f41393fd522d74b384dd095429cd38a358538bcc039b12845e9d9f2bf8682c36862582d389bf6412ce8ace3afb90cdcb29111
     crtCoefficient: 297b7b781a09d8b72c3ce5ed92864326fed3c3b2788977fbcbd99d86b960f2ebf25e62fd895cfa34396eeef3b70f22511d3a29c9f9bd76a86b4df18827da45210bd311af635738fba1aaa22cc015f80b2c60c21b5832d13990b4fbfba8e1f4dece552b4e1cde025fda4b4c0f2bb3ab3a6358835c93aac82d7609ffab7966d7f7
,publicKey=RSA Public Key
            modulus: af40d37186a8db2d3877f527cce0d70e9a623cdd58c447247461f8b82d497efc35d39705988a8d7bd4f9487f1cf8b190de58f8e758808f9eedd1a899570ef004a2aab95e272d8cb57ff02973e9615beecddaeeca5af5f303f95d64260977ed9755f38b35f642dd6799d619ab1fedbcd6e3ed35701e36d28f3012d90791c5cf8978c16ab1ac51822bba617ef169f08c231264c2c39aefc7f5cd0cae734a5671a8005686be24358d5c45f1751623be7293db69c91c21bdfacd2aa494ff9d72ce740907981e31ba9bd9ee1ca180d523c35199bd8871a87d09dd517a44d6f2bb57a8c763e697c7c297571207993f9ffbced140bc21d6aa6e129cd071aa0bf19f474f
    public exponent: 10001
]

org.bouncycastle.jcajce.provider.asymmetric.rsa.BCRSAPrivateCrtKey@7d99eb6c[publicExponent=65537,primeP=149234898893716497758826871704114636759678845056946167204954104074869832432987821552579544933793982586311189541476737115266605469568215268695776969137196555640928352815357910824312227830922908738329787978619992069673014794014610837970442808322504756722101332651655030240526252048788246777196705872370344237031,primeQ=148247296047593387236012636210100127780547417821816258203151943547093643381098762918202457401160370123525267863455706310619998636329199092440636583491724505759616279383555968631007687387871778865992243885839910218455526088622943764144231073762450207407162296375865211806310451871365291924605536772950092662873,primeExponentP=124177590389719422193297004663632505095069448670071033709015703457815837955172267157892798325748619136965501918997679251519504952499557246254586042357122537936840957994414040175507251175728504237718209985780307727474228539023738847170944011264646717096088499221702301968454737064201435242700580761065840399743,primeExponentQ=148041450434575839923220211321943527514620844816921537976463403838794735564938731372273342189546997621255698011683814565891579272063121043133953367459593847810303294635644047225611930799192157707306229967082392605048146244045610534326919859826621851381191443743455981382666186019704485943783419436997295640849,crtCoefficient=29129898725800794232315947413604097130194922712068839804629797939099817246698403874580172617624168731533951031757760066130659310268061770611424979399955694425280877285440766185802709139871658142385885040233891571204978955857437258390902075149847214007047312610865971479288080528926082220594344488628046452727,
  modulus=22123670236929456515511986910848172744293600927799941216186489363421027795069019530971221538266334973677462896386866630400894954845722127666037186149799754798674476112558703989697234592420947285216427478649713575056195457654957155434972413122101775377844551012734423306015591088744466263443835953647213909785678748608065177836758478083293835971537218709204691210936471658636112005645895466247337892865874920928922319297450254179959613699385398943205504475754291704545269505864516791542997861181165844257881455979775063400259593991784062794626490906067682259638421146419878130765296698328826891868651436528452285450063,
  privateExponent=15943004683609325607131850982277759164555872020662154565482941234045627671475954871426962644143651191492427448167583765210086926887922313279698555434071483583804540303094778095204556133647504438503462730391514300666096999773079581562825230433657668912901757102543927001487775393124855466224048772736784084278254824757907960932387706526664765710423943104735965404015039612386532901519701388048631789645457018109632529494176284637584266858129766938400583275361475269457544079428095493111934541280663823282028540856586423379216707756515728350640830285314497836524505107517986456023245321206842319962219108728085675928513]
```

##### Final generated pair

```
java.security.KeyPair@ac0cac8[privateKey=org.bouncycastle.jcajce.provider.asymmetric.rsa.BCRSAPrivateKey@b1c54c37,publicKey=RSA Public Key
            modulus: af40d37186a8db2d3877f527cce0d70e9a623cdd58c447247461f8b82d497efc35d39705988a8d7bd4f9487f1cf8b190de58f8e758808f9eedd1a899570ef004a2aab95e272d8cb57ff02973e9615beecddaeeca5af5f303f95d64260977ed9755f38b35f642dd6799d619ab1fedbcd6e3ed35701e36d28f3012d90791c5cf8978c16ab1ac51822bba617ef169f08c231264c2c39aefc7f5cd0cae734a5671a8005686be24358d5c45f1751623be7293db69c91c21bdfacd2aa494ff9d72ce740907981e31ba9bd9ee1ca180d523c35199bd8871a87d09dd517a44d6f2bb57a8c763e697c7c297571207993f9ffbced140bc21d6aa6e129cd071aa0bf19f474f
    public exponent: 10001
]

org.bouncycastle.jcajce.provider.asymmetric.rsa.BCRSAPrivateKey@774115fb[
            modulus=22123670236929456515511986910848172744293600927799941216186489363421027795069019530971221538266334973677462896386866630400894954845722127666037186149799754798674476112558703989697234592420947285216427478649713575056195457654957155434972413122101775377844551012734423306015591088744466263443835953647213909785678748608065177836758478083293835971537218709204691210936471658636112005645895466247337892865874920928922319297450254179959613699385398943205504475754291704545269505864516791542997861181165844257881455979775063400259593991784062794626490906067682259638421146419878130765296698328826891868651436528452285450063,
    privateExponent=15943004683609325607131850982277759164555872020662154565482941234045627671475954871426962644143651191492427448167583765210086926887922313279698555434071483583804540303094778095204556133647504438503462730391514300666096999773079581562825230433657668912901757102543927001487775393124855466224048772736784084278254824757907960932387706526664765710423943104735965404015039612386532901519701388048631789645457018109632529494176284637584266858129766938400583275361475269457544079428095493111934541280663823282028540856586423379216707756515728350640830285314497836524505107517986456023245321206842319962219108728085675928513]
```

The values of the exponents and moduli are the same between the two keys, but the second has lost some (supposedly optional) information - crts, prime exponents, etc.

So, that makes sense to me, that the first generated "seed" key itself can be used in place of the final key.

I also think what you said earlier makes sense (about it being missed in the conversion). As you said, earlier the conversion was needed to convert the keypair to a bcprov keypair. I guess, even then, the crts would have got lost.

Let's see what @jyotisingh says.

wpc:
> The values of the exponents and moduli are the same between the two keys, but the second has lost some (supposedly optional) information - crts, prime exponents, etc.

@arvindsv you are right.

And regarding the performance difference with/out CRT factors, I am really not kiding. Setup a test to show that. 
```java
 @Test
    public void shouldGenerateValidRSAKeyPair() throws Exception {
        String text = "this is a secret message";
        X509CertificateGenerator generator = new X509CertificateGenerator();
        Registration registration = generator.createAgentCertificate(keystore, "agentHostName");
        PrivateKey privateKey = registration.getPrivateKey();
        PublicKey publicKey = registration.getPublicKey();

        final Cipher cipher = Cipher.getInstance("RSA");
        cipher.init(Cipher.ENCRYPT_MODE, publicKey);
        byte[] encrypted = cipher.doFinal(text.getBytes());

        cipher.init(Cipher.DECRYPT_MODE, privateKey);
        byte[] decrypted = null;
        long start = System.currentTimeMillis();
        for(int i=0; i<1000; i++ ) {
            decrypted = cipher.doFinal(encrypted);
        }

        System.out.println("use " + (System.currentTimeMillis() - start) + "ms");

        assertThat(decrypted, is(text.getBytes()));
    }
```

Above test code encrypt a piece of text with public key and decrypt with private key for 1000 times. 
It takes `6766ms` for private key with CRTs vs takes `21577ms` for key without CRTs.  `219%` difference. Because we send big chuck of serialized data from server to agent, we are likely wasting a lot of CPU cycles on agent.

-- wpc

arvindsv:
Nice! (About the performance).

Just for my peace of mind, I did go and check whether those values were present when using sun.security.rsa.RSAKeyPairGenerator. They weren't still in the generated key. But the seed itself didn't have the crts at that time. So, it looks like the bouncy castle implementation has it. Anyway, I think this change makes sense. Once @jyotisingh is back on Monday, and takes a look, I think this can get merged.

One verification is to make sure that existing server-agent communication for existing keys and agents continues to work. Looking at the change, it should. But, I haven't tried it out myself.

jyotisingh:
Don't really remember if we left the code the way it is currently because we forgot about it or if it was intentional. But looking at the changes and debugging the code(new and old) confirms what both of you say. The change seems fine to me too. Regarding the communication between servers with existing certificates and registered agents, nothing changes as regeneration of certificates kick in only when `keystore` is missing on the server - verified it locally too. Just one rather small thing though, [this](https://github.com/gocd/gocd/blob/2feb8d861d6038ec1a4c5c0e9e316f1fe09594fd/cruise-modules.rb#L215) line needs to be removed. `./bn dist` currently fails because of this.

wpc:
@jyotisingh thx for reviewing this. I have removed the line you mentioned

PR #1999, opened 2016-02-29
Convert pipeline actions controller end-points to be AJAX only.

zabil:


ketan:
Merge https://github.com/gocd/api.go.cd/pull/74 after this is merged in.

PR #2020, opened 2016-03-07
Created backup filter to handle the urls when backup in progress.

juhijariwala:


server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        this.filterConfig = filterConfig;
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        SpringBeanAutowiringSupport.processInjectionBasedOnCurrentContext(this);
```
zabil: Can we put a condition on this? Otherwise it will be injected on every request.

have a private method autoWireService

Check if backupService == null and call this method

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+
+        String url = ((HttpServletRequest) request).getRequestURI();
+        if (isBackupFinishJsonUrl(url)) {
+            response.setContentType("application/json");
+            JsonObject json = new JsonObject();
+            json.addProperty("is_backup_finished", backupService.isBackingUp());
+            response.getWriter().print(json);
+            return;
```
ketan: You'd need the cache control headers here too. Plus better to extract the top level `if` block statements into methods.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+            ((HttpServletResponse) response).setDateHeader("Expires", 0);
+
+            if (isAPIUrl(url) && !isMessagesJson(url)) {
+                generateAPIResponse(request, response);
+            } else {
+                try {
+                    generateHTMLResponse(response);
+
+                } catch (Exception e) {
+                    e.printStackTrace();
```
zabil: Handle this in the generate html response method and do not print stack trace, log this as error using a logger.

server/webapp/static/backup_in_progress.html
```diff
@@ -118,15 +106,16 @@
...
 
         function _onComplete(xhr) {
-            var forcePageReloadUrl = xhr.getResponseHeader("X-GO-FORCE-LOAD-PAGE");
-            if (forcePageReloadUrl) {
-                window.location = forcePageReloadUrl;
+            var jsonResponse = JSON.parse(xhr.responseText);
+            var is_backup_finished = jsonResponse["is_backup_finished"];
+
+            if (!is_backup_finished) {
+                location.reload(true);
```
zabil: isn't this supposed to be

```
if (is_backup_finished) {
```
?

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+            json.addProperty("is_backup_finished", backupService.isBackingUp());
+            response.getWriter().print(json);
+            return;
+        }
+        if (backupService.isBackingUp()) {
+            ((HttpServletResponse) response).setHeader("Cache-Control", "private, max-age=0, no-cache");
+            ((HttpServletResponse) response).setDateHeader("Expires", 0);
+
+            if (isAPIUrl(url) && !isMessagesJson(url)) {
+                generateAPIResponse(request, response);
```
ketan: The api response is always JSON even when the user requested something else?

server/src/com/thoughtworks/go/server/web/BackupFilter.java
juhijariwala: I will do that.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+        response.getWriter().print(jsonObject);
+        ((HttpServletResponse) response).setStatus(503);
+    }
+
+    private boolean isBackupFinishJsonUrl(String url) {
+        return "/go/is_backup_finished.json".equals(url);
+    }
+
+    private boolean isAPIUrl(String url) {
+        Pattern pattern = Pattern.compile("^((.*/api/.*)|(.*[^/]+\\.(xml|json)(\\?.*)?))$");
```
ketan: Compiling patterns may be an expensive operation to perform on each request. I'd consider making this a constant.

server/webapp/static/backup_in_progress.html
```diff
@@ -1,6 +1,6 @@
 ﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
```
ketan: Consider moving this file out of `/static` because it'll be reachable if users connect to that URL directly.

server/webapp/static/backup_in_progress.html
juhijariwala: Bad name. I will correct it.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+        content = content.replaceAll("%backup_started_by%", urlEncode((backupService.backupStartedBy())));
+        response.getWriter().print(content);
+    }
+
+    private void generateAPIResponse(ServletRequest request, ServletResponse response) throws IOException {
+        response.setContentType(request.getContentType());
+        JsonObject jsonObject = new JsonObject();
+        jsonObject.addProperty("message", "Server is under maintenance mode, please try later.");
+        response.getWriter().print(jsonObject);
+        ((HttpServletResponse) response).setStatus(503);
```
ketan: :+1: on `503`

server/webapp/static/backup_in_progress.html
```diff
@@ -82,7 +82,7 @@
             <a href="#" id="application_logo">&nbsp;</a>
 
             <div class="message">Go backup in progress. Backup initiated by '<span
-                    id="backup_started_by">&nbsp;</span>' on '<span id="backup_started_at">&nbsp;</span>'
+                    id="backup_started_by">&nbsp;%backup_started_by%</span>' on '<span id="backup_started_at">&nbsp;%backup_initiated_by%</span>'
```
ketan: There's no need for the IDs around these elements, or the non-breaking-space.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,131 @@
...
+    @Autowired
+    private BackupService backupService;
+
+    // For the Test
+    protected BackupFilter(BackupService backupService) {
+        this.backupService = backupService;
+    }
+
+    private String urlEncode(String string) throws UnsupportedEncodingException {
+        return URLEncoder.encode(string, "UTF-8");
```
ketan: Did you mean html escape, or url encode?

server/webapp/static/backup_in_progress.html
zabil: or remove this entry?
https://github.com/gocd/gocd/blob/41cb67ad26a6c09dfeb26566ffa775b6f4c0b19b/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L246

server/resources/backup_in_progress.html
```diff
@@ -81,8 +81,7 @@
         <div class="header clear_float">
             <a href="#" id="application_logo">&nbsp;</a>
 
-            <div class="message">Go backup in progress. Backup initiated by '<span
-                    id="backup_started_by">&nbsp;</span>' on '<span id="backup_started_at">&nbsp;</span>'
+            <div class="message">Go backup in progress. Backup initiated by '<span>&nbsp;%backup_started_by%</span>' on '<span>&nbsp;%backup_initiated_by%</span>'
```
ketan: Still an unwanted non-breaking space there.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,173 @@
...
+ * @understands redirecting all requests to a service unavailable page when the server is being backed up.
+ */
+public class BackupFilter implements Filter {
+
+    public static final String JSON = "json";
+    public static final String XML = "xml";
+
+    @Autowired
+    private BackupService backupService;
+    private final static org.apache.log4j.Logger LOGGER = org.apache.log4j.Logger.getLogger(BackupFilter.class);    // For the Test
```
ketan: Prefer using slf4j instead

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,173 @@
...
+    private void generateHTMLResponse(ServletResponse response) {
+        String path = "backup_in_progress.html";
+        InputStream resourceAsStream = getClass().getClassLoader().getResourceAsStream(path);
+        response.setContentType("text/html");
+        StringWriter writer = new StringWriter();
+        try {
+            IOUtils.copy(resourceAsStream, writer);
+            String content = writer.toString();
+            content = replaceStringLiterals(content);
+            response.getWriter().print(content);
```
ketan: Prefer `FileUtils.readToEnd(InputStream)`? I think you may want to close the inputstream inside `FileUtils` as well.

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,173 @@
...
+        String url = request.getRequestURI().toString();
+        return header != null && header.contains(type) || url != null && url.contains(type) || contentType != null && contentType.contains(type);
+    }
+
+    private boolean isBackupFinishJsonUrl(String url) {
+        return "/go/is_backup_finished.json".equals(url);
+    }
+
+    private void generateResponseForIsBackupFinishedAPI(ServletRequest request, ServletResponse response) {
+        response.setContentType(request.getContentType());
```
ketan: Did you really mean to copy the content type here?

server/resources/backup_in_progress.html
```diff
@@ -118,15 +105,16 @@
...
         }
 
         function _onComplete(xhr) {
-            var forcePageReloadUrl = xhr.getResponseHeader("X-GO-FORCE-LOAD-PAGE");
-            if (forcePageReloadUrl) {
-                window.location = forcePageReloadUrl;
+            var jsonResponse = JSON.parse(xhr.responseText);
+            var is_backing_up = jsonResponse["is_backing_up"];
+
+            if (!is_backing_up) {
```
ketan: camelCase and snake_case in one file?

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,173 @@
...
+            LOGGER.error(String.format("General IOException", e.getMessage()));
+        }
+        ((HttpServletResponse) response).setStatus(503);
+    }
+
+    private boolean requestIsOfType(String type, HttpServletRequest request) {
+        String header = request.getHeader("Accept");
+        String contentType = request.getContentType();
+        String url = request.getRequestURI().toString();
+        return header != null && header.contains(type) || url != null && url.contains(type) || contentType != null && contentType.contains(type);
```
ketan: did you mean `url.endsWith`? This will return unexpected response for a url that looks like `/go/tab/pipeline/history/build-xml-document`

server/src/com/thoughtworks/go/server/web/BackupFilter.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        if (backupService == null) {
+            SpringBeanAutowiringSupport.processInjectionBasedOnCurrentContext(this);
+        }
+
+        String url = ((HttpServletRequest) request).getRequestURI();
+        ((HttpServletResponse) response).setHeader("Cache-Control", "private, max-age=0, no-cache");
+        ((HttpServletResponse) response).setDateHeader("Expires", 0);
```
ketan: The cache control headers are to be inserted ONLY IF the filter is rendering the response, not if the filter is forwarding along the chain. You may end up setting cache headers on assets among other things significantly killing performance both for clients and the server.

PR #2045, opened 2016-03-10
Using unique string for cipher generation

jyotisingh:
re-generation of cipher and re-encryption of secure values in config when unsafe cipher is available

### DIFF
--- a/server/src/com/thoughtworks/go/config/ConfigCipherUpdater.java
+++ b/server/src/com/thoughtworks/go/config/ConfigCipherUpdater.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.thoughtworks.go.config;
+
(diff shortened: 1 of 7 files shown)


server/src/com/thoughtworks/go/config/ConfigCipherUpdater.java
```diff
@@ -0,0 +1,130 @@
...
+                } catch (IOException e1) {
+                    LOGGER.error("Could not replace the cipher file [{}] with original one [{}], please do so manually. Error: [{}]", cipherFile.getAbsolutePath(), originalCipherFile.getAbsolutePath(), e.getMessage(), e);
+                    bomb(e1);
+                }
+            }
+        }
+    }
+
+    private String reEncryptUsingNewKey(byte[] oldCipher, byte[] newCipher, String encryptedValue) throws InvalidCipherTextException {
+        GoCipher cipher = new GoCipher();
```
arvindsv: Can make this a field, if needed. Ok to leave it in too. I thought that every call to `CipherProvider#primeKeyCache` reads the cipher file. But, it is `private static volatile`. So, it doesn't.

server/src/com/thoughtworks/go/config/ConfigCipherUpdater.java
```diff
@@ -0,0 +1,130 @@
...
+            byte[] oldCipher = FileUtils.readFileToByteArray(originalCipherFile);
+            new CipherProvider(systemEnvironment).resetCipher();
+
+            byte[] newCipher = FileUtils.readFileToByteArray(cipherFile);
+
+            if (new String(newCipher).equals(new String(oldCipher))) {
+                LOGGER.warn("Unable to generate a new safe cipher. Your cipher is unsafe.");
+                FileUtils.deleteQuietly(originalCipherFile);
+                FileUtils.deleteQuietly(originalConfigFile);
+                return;
```
ketan: We return without restoring the old cipher and config xml?

arvindsv: They were never moved, right?

jyotisingh: point noted, renamed variable

PR #2050, opened 2016-03-11
Upgrading xalan dependency of tfs

zabil:


zabil:
Will be done as a part of #2046

PR #2052, opened 2016-03-11
Build command API main process implementation

wpc:
About this PR
==========
This PR is first part of of implementation of proposal: **https://github.com/gocd/gocd/issues/1954 New BuildCommand based Server-Agent Communication**. The goal of this PR is implement only the main follow of protocol in the proposal, so there will be a few places stubbed by UnsupportedOperationException exception.

Toggle
=====
New API will be toggled by a agent side configuration. Most code for current mechanism will be untouched for now.

Two toggle need be turn on on ***agent side*** for testing the new protocol:
```
-Dgo.agent.websocket.enabled=Y 
-Dgo.agent.enableBuildCommandProtocol=Y
```

Scope
=====

- [x] Re-enable JSON WebSocket message serialization, except for BuildWork message. We left transfering BuildWork still via Java serialization because it is not in the blue print and we have less changes to test.
- [x] Clean up the JSON structure make it less java server domain dependent.
- [x] Change agent registration api to json format (with PEM format certs and private key, for the connivence of none-java agent implementation, e.g. https://github.com/gocd-contrib/gocd-golang-agent)
- [x] Toggle
- [x] Run a build with stubbed out commands implementation but correct status report.
- [x] BuildCommand-lize GitMaterial#update_to
- [x] BuildCommand-lize CommandBuilder
- [x] Environment variables
- [x] Mask secret variables
- [x] Cancel build
- [x] Upload Artifacts
- [x] Fetch artifacts
- [x] Git submodules
- [x] Generate test reports
- [x] Generate property from xml property file
- [x] Support elastic agent ping info

What's left
========
* SVN
* HG 
* P4
* TFS 
*  Pluggable SCM
* Task extensions

You can preview other changes on this POC fork: https://github.com/wpc/gocd/tree/lightweight-agent-spike

### DIFF
--- a/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
+++ b/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
@@ -1,0 +1,122 @@
+/*************************** GO-LICENSE-START*********************************
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ************************GO-LICENSE-END***********************************/
+package com.thoughtworks.go.buildsession;
+
+import com.jezhumble.javasysmon.JavaSysMon;
+import com.thoughtworks.go.domain.BuildCommand;
+import com.thoughtworks.go.util.SystemUtil;
+import com.thoughtworks.go.util.command.CommandLine;
+import com.thoughtworks.go.util.command.CommandLineException;
+import com.thoughtworks.go.util.command.StringArgument;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.lang.String.format;
+
+public class ExecCommandExecutor implements BuildCommandExecutor {
+    private static final Logger LOG = LoggerFactory.getLogger(ExecCommandExecutor.class);
+
+    @Override
+    public boolean execute(BuildCommand command, BuildSession buildSession) {
+        File workingDir = buildSession.resolveRelativeDir(command.getWorkingDirectory());
+        if (!workingDir.isDirectory()) {
+            String message = "Working directory \"" + workingDir.getAbsolutePath() + "\" is not a directory!";
+            LOG.error(message);
+            buildSession.println(message);
+            return false;
+        }
+
+        String cmd = command.getStringArg("command");
+        String[] args = command.getArrayArg("args");
+        CommandLine commandLine;
+
+        if (SystemUtil.isWindows()) {
+            commandLine = CommandLine.createCommandLine("cmd");
+            commandLine.withArg("/c");
+            commandLine.withArg(StringUtils.replace(cmd, "/", "\\"));
+        } else {
+            commandLine = CommandLine.createCommandLine(cmd);
+        }
+        commandLine.withWorkingDir(workingDir);
+        commandLine.withEnv(buildSession.getEnvs());
+        commandLine.withArgs(args);
+        for (SecretSubstitution secretSubstitution : buildSession.getSecretSubstitutions()) {
+            commandLine.withNonArgSecret(secretSubstitution);
+        }
+        return executeCommandLine(buildSession, commandLine) == 0;
+    }
+
+    private int executeCommandLine(final BuildSession buildSession, final CommandLine commandLine) {
+        ExecutorService executorService = Executors.newFixedThreadPool(2);
+        final AtomicInteger exitCode = new AtomicInteger(-1);
+        final CountDownLatch canceledOrDone = new CountDownLatch(1);
+
+        try {
+            final Future<?> executing = executorService.submit(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        exitCode.set(commandLine.run(buildSession.processOutputStreamConsumer(), null));
+                    } catch (CommandLineException e) {
+                        LOG.error("Command failed", e);
+                        String message = format("Error happened while attempting to execute '%s'. \nPlease make sure [%s] can be executed on this agent.\n", commandLine.toStringForDisplay(), commandLine.getExecutable());
+                        String path = System.getenv("PATH");
+                        buildSession.println(message);
+                        buildSession.println(format("[Debug Information] Environment variable PATH: %s", path));
+                        LOG.error(format("[Command Line] %s. Path: %s", message, path));
+                    } finally {
+                        canceledOrDone.countDown();
+                    }
+                }
+            });
+
+            Future<?> cancelMonitor = executorService.submit(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        buildSession.waitUntilCanceled();
+                        executing.cancel(true);
+                        new JavaSysMon().infanticide();
+                    } catch (InterruptedException e) {
+                        //ignored
+                    } finally {
+                        canceledOrDone.countDown();
+                    }
+                }
+            });
+
+            try {
+                canceledOrDone.await();
+            } catch (InterruptedException e) {
+                LOG.error("Building thread interrupted", e);
+            }
+            executing.cancel(true);
+            cancelMonitor.cancel(true);
+            return exitCode.get();
+        } finally {
+            executorService.shutdown();
+        }
+
+    }
+}
--- a/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
+++ b/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
@@ -28,8 +28,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -88,54 +86,45 @@ public class ExecCommandExecutor implements BuildCommandExecutor {
     }
 
     private int executeCommandLine(final BuildSession buildSession, final CommandLine commandLine) {
-        ExecutorService executorService = Executors.newFixedThreadPool(2);
         final AtomicInteger exitCode = new AtomicInteger(-1);
         final CountDownLatch canceledOrDone = new CountDownLatch(1);
+        buildSession.submitRunnable(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    exitCode.set(commandLine.run(buildSession.processOutputStreamConsumer(), null));
+                } catch (CommandLineException e) {
+                    LOG.error("Command failed", e);
+                    String message = format("Error happened while attempting to execute '%s'. \nPlease make sure [%s] can be executed on this agent.\n", commandLine.toStringForDisplay(), commandLine.getExecutable());
+                    String path = System.getenv("PATH");
+                    buildSession.println(message);
+                    buildSession.println(format("[Debug Information] Environment variable PATH: %s", path));
+                    LOG.error(format("[Command Line] %s. Path: %s", message, path));
+                } finally {
+                    canceledOrDone.countDown();
+                }
+            }
+        });
+
+        Future<?> cancelMonitor = buildSession.submitRunnable(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    buildSession.waitUntilCanceled();
+                } catch (InterruptedException e) {
+                    // ignore
+                } finally {
+                    canceledOrDone.countDown();
+                }
+            }
+        });
 
         try {
-            final Future<?> executing = executorService.submit(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        exitCode.set(commandLine.run(buildSession.processOutputStreamConsumer(), null));
-                    } catch (CommandLineException e) {
-                        LOG.error("Command failed", e);
-                        String message = format("Error happened while attempting to execute '%s'. \nPlease make sure [%s] can be executed on this agent.\n", commandLine.toStringForDisplay(), commandLine.getExecutable());
-                        String path = System.getenv("PATH");
-                        buildSession.println(message);
-                        buildSession.println(format("[Debug Information] Environment variable PATH: %s", path));
-                        LOG.error(format("[Command Line] %s. Path: %s", message, path));
-                    } finally {
-                        canceledOrDone.countDown();
-                    }
-                }
-            });
-
-            Future<?> cancelMonitor = executorService.submit(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        buildSession.waitUntilCanceled();
-                        executing.cancel(true);
-                    } catch (InterruptedException e) {
-                        //ignored
-                    } finally {
-                        canceledOrDone.countDown();
-                    }
-                }
-            });
-
-            try {
-                canceledOrDone.await();
-            } catch (InterruptedException e) {
-                LOG.error("Building thread interrupted", e);
-            }
-            executing.cancel(true);
-            cancelMonitor.cancel(true);
-            return exitCode.get();
-        } finally {
-            executorService.shutdown();
+            canceledOrDone.await();
+        } catch (InterruptedException e) {
+            LOG.error("Building thread interrupted", e);
         }
-
+        cancelMonitor.cancel(true);
+        return exitCode.get();
     }
 }
--- a/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
+++ b/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
@@ -15,17 +15,18 @@
  * ************************GO-LICENSE-END***********************************/
 package com.thoughtworks.go.buildsession;
 
-import com.jezhumble.javasysmon.JavaSysMon;
 import com.thoughtworks.go.domain.BuildCommand;
 import com.thoughtworks.go.util.SystemUtil;
 import com.thoughtworks.go.util.command.CommandLine;
 import com.thoughtworks.go.util.command.CommandLineException;
-import com.thoughtworks.go.util.command.StringArgument;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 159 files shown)


wpc:
@arvindsv: this is ready for review

arvindsv:
Ok, will take a look.

wpc:
hi, @arvindsv: sorry for the long diff. If you have trouble with it, let me know, I can unsquash it a little bit.

arvindsv:
:) Ok. I'll let you know.

common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+            });
+
+            Future<?> cancelMonitor = executorService.submit(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        buildSession.waitUntilCanceled();
+                        executing.cancel(true);
+                        new JavaSysMon().infanticide();
```
wpc: There is behavior difference here:

On trunk implementation on canceling a build it kills all subprocesses then wait for builder thread return. This is assuming subprocess can close the stdout & stderr cleanly. It is not always the case, base on my test there is a 1/10 chance it hangs at  ```processOutputStream.readToEnd();``` in ProcessWrapper.java

``` 
public int waitForExit() {
        int returnValue = -1;
        try {
            returnValue = process.waitFor();
            processOutputStream.readToEnd();
            processErrorStream.readToEnd();
        } catch (InterruptedException ignored) {
            LOGGER.warn(ignored.getMessage(), ignored);
        } finally {
            close();
        }
        return returnValue;
    }
```

So here I make it interrupt the execution thread first, then kill all subprocess.

wpc: There is another bonus for the new canceling implementation:  because it is implemented on exec command level, it can cleanly cancel build hangs at material updating stage. In current trunks implementation when `git fetch` hangs, user need kill process manually to get agent back.

arvindsv:
@wpc: I still have to take a look at changes in UrlBasedArtifactsRepository and BuildSession. But, meanwhile, I have some thoughts about the rest of the changes. They're not always well formed thoughts. We can have a small discussion :) - Here they are:

* Should we be thinking about versioning?
  - Maybe not because the life of a version of this protocol is just till the next server upgrade. All agents will have to upgrade immediately anyway.
  - However, with versioning, it will be possible to  allow agents to complete their jobs (while communicating with the server over the older protocol and then upgrade)
  - But, I don't think it's worth the complication, just to allow that.

* Earlier the status was updated immediately [after work](https://github.com/gocd/gocd/blob/aa5a08467523ff8875f3366bead899d4dd193003/agent/src/com/thoughtworks/go/agent/AgentController.java#L284). Now it [is not](https://github.com/gocd/gocd/pull/2052/files#diff-4aa6ab956d8b03c9c3fd4a4f958d21a6R356). Do we need to?

* [All these changes](https://github.com/gocd/gocd/pull/2052/files#diff-de55c3e027b0e5a2bc6927e540d15982R70) are used just in tests.  
  - I feel they can be moved out of there. They'll complicate what is a simple class and pollute its interface, don't you think?
  - Do we need a TestInMemoryConsumer, which has this extra test-specific behavior?

* AgentInstance#supportsBuildCommandProtocol can be removed.

* About [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/server/src/com/thoughtworks/go/server/websocket/AgentRemoteHandler.java#L143-L147)
  - I wonder if it is possible for the agent to not have pinged and so the AgentRuntimeInfo is not fully updated.
    If that happens, then supportsBuildCommandProtocol will not be set properly and it can get cancelJob instead of
    cancelBuild. Possible? Not possible?


* Are usages in MessageEncoding safe? Do some of those streams need to be closed? Missing finally blocks? Especially GZIPOutputStream in encodeMessage and GZIPInputStream in decodeMessage.

* Why doesn't ElasticAgentRuntimeInfo support build command protocol? I mean, why is it hard-coded to false? Hmm, why does ElasticAgentRuntimeInfo need to inherit from AgentRuntimeInfo?

* Question to @ketan, I guess: Should [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java#L115) be called only after a successful registration (somewhere around [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java#L125))?

* Is there a change in behavior around "on-cancel" tasks? Looks like earlier, if an "on cancel task" was specified, it wouldn't kill children, but would do whatever it was asked. Now, it looks like it will kill all its children anyway. I don't know yet. Haven't verified the old behavior fully.

* Should there be a reportCurrentStatus [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/server/src/com/thoughtworks/go/server/domain/BuildComposer.java#L47)?
  - Or, is it not needed because the server would have cancelled it anyway?

* I know [this is](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java#L44) how it was done before, but I think an easier way would be to use  [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/config/config-api/src/com/thoughtworks/go/util/command/UrlArgument.java#L93) to get the password and mark that a secret. But, what you have is fine.

* Instead of  [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/BuildSession.java#L206-L208), this [part](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/domain/BuildCommand.java#L97) could have been replaced with this: ```return secret(secretValue, "******")```

* Is this a change in behavior? [This](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/CleandirCommandExecutor.java#L33-L43) vs [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/config/materials/Materials.java#L115-L117).

* Small fixes:
  - Typo in method name: BuildSession#buildVariableSubstitude
  - BuildComposer: `echoWithPrefix("Job started ${date}")` => `echoWithPrefix("Job started at ${date}")`

wpc:
Hi, @arvindsv 
Wow this is quick around time 👍 

> Should we be thinking about versioning?

About versioning, I thought about it, but feel not worth pay the cost for now.  The java agent does not have much problem because autoupdating. For golang agent, ideally we should maintain protocol backward compatibility from agent side. golang agent will need protocol version information to make compatible with old server easier, but I would rather think about it when golang agent hit a public release. Anyway absent version information is a version information as well.

For the last build problem, when a build run, agent already get all information it needed to run a build, so there is not much risk unless we change transportation layer they communicate, or how to send cancel instruction from server. 

> Earlier the status was updated immediately after work. Now it is not. Do we need to?

We need do that. The line get lost when I moving code around.

> All these changes are used just in tests.

Will extract a TestInMemoryConsumer class

> AgentInstance#supportsBuildCommandProtocol can be removed.

Will remove that

> I wonder if it is possible for the agent to not have pinged and so the AgentRuntimeInfo is not fully updated. If that happens, then supportsBuildCommandProtocol will not be set properly and it can get cancelJob instead of cancelBuild. Possible? Not possible?

It is not possible.  The code checks agent session exists, which implicates agent at least pinged or report status once, in both case AgentRuntimeInfo will get updated. 
Even saying that, after recheck the code, I feel unnecessary behavior variance base an a remote config is nasty. I will change it use same websocket message, and move the if clause to agent side.

wpc:
> Are usages in MessageEncoding safe? Do some of those streams need to be closed? Missing finally blocks? Especially GZIPOutputStream in encodeMessage and GZIPInputStream in decodeMessage.

I prefer use try-with-resource to close resource, so there is no finally block there. But I do missed quite a lot of streams when I copy code from old json serialize code. Rookie mistake, fixing it now, sorry for that.

> Why doesn't ElasticAgentRuntimeInfo support build command protocol? I mean, why is it hard-coded to false?

It's the default constructor behavior. The buildcomand flag does get set in the ElasticAgentRuntimeInfo deserialization. [MessageEncoding::AgentRuntimeInfoTypeAdapter](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/websocket/MessageEncoding.java#L123).

> Hmm, why does ElasticAgentRuntimeInfo need to inherit from AgentRuntimeInfo?

It is what trunk looks like now, and I do not know the reason why. It does make lost a lot hair when trying to encoding AgentRuntimeInfo into JSON. I want to merge them then realized it will conflict with PR #2058. So I left a note there and work around it with a GSON type adapter.

> Should there be a reportCurrentStatus here?

It's not need, server reject any further status reports if job already canceled

> Instead of this, this part could have been replaced with this: return secret(secretValue, "******")

BuildCommand#secret is on server side just for compose the json structure. Default mask value in build session (agent side) is to make the protocol more tolerant. I need unit testing the case when substitution is missing from the json structure anyway.

> Is this a change in behavior? [This](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/CleandirCommandExecutor.java#L33-L43) vs [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/config/materials/Materials.java#L115-L117).

No behavior change, because allowedFolders is never empty. It at least has "cruise-output" dir

> Small fixes:

Will fix soon.

wpc:
> Is there a change in behavior around "on-cancel" tasks? Looks like earlier, if an "on cancel task" was specified, it wouldn't kill children, but would do whatever it was asked. Now, it looks like it will kill all its children anyway. I don't know yet. Haven't verified the old behavior fully.

I forget to bring this to discussion upfront. I was wondering whether worth to mimic the whole old behavior. Does user really know that the cancel task they specified need kill all child process?  Failing to do that will leak processes for the next build. To me it seems like unreasonable burden for the user.

The current implementation in BuildSession do have a problems: All child processes are killed before cancel task run. This make user have no chance to stop the running command gracefully. I can make change to the order, make sure the cancel task runs first, then at last kill all child process, make sure no leaking processes.

Or we can introduce a new command "killChildProcesses" to mimic the old behavior. What do you think?

arvindsv:
> I prefer use try-with-resource to close resource, so there is no finally block there.

Either is fine. I saw that the other two didn't have that.

Continuing from the earlier comment:

There's a small change in behavior [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/remote/work/BuildWork.java#L97-L101) vs. [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/BuildSession.java#L126). Earlier, if it failed to report completion, it would report failure. I've actually seen it fail to report final completion and fail the build (mainly because the server was very busy). Now, it kind of leaves that job hanging, since it just logs failures around [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/agent/src/com/thoughtworks/go/agent/service/AgentWebsocketService.java#L191). I know that the server will see that the same agent is now asking for work and will reschedule the job. But, was that the idea?

I think that there's a behavior change around how commands are executed and how secrets are (not) printed. It looks like [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/util/command/CommandLine.java#L453-L457) is missing somewhere around [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java#L80). The equivalent of [this test](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/test/unit/com/thoughtworks/go/util/command/CommandLineTest.java#L143-L153) actually (and those like shouldLogPasswordsOnOutputAsStarsUnderLinux, shouldNotLogPasswordsOnExceptionThrown, etc). Something like:

```java
@Test
public void shouldNotPrintSecureVariableValuesInConsole() throws Exception {
    runBuild(compose(
            export("SECURE", "abc", true),
            exec("echo", "Should not see: abc")), Passed);
    assertThat(console.output(), not(containsString("Should not see: abc")));
    assertThat(console.output(), containsString("Should not see: ******"));
}

```

`newSafeConsole()` should probably be used more than it is, in BuildSession.

That makes me think: Are there other missing tests? It's hard to know for sure. I understand that not every executor has a test, and whatever doesn't have its own test is tested using the BuildSessionTest and in some cases, they're probably so simple that there is no test necessary. But, in cases like CommandLineTest, there are some very useful (and sometimes OS-specific) tests, which I don't see migrated over (or replaced). I'm not able to do a great comparison across the two for tests. I feel that tests like `RegistrationTest.java` can get some more love. :) Apart from that, the other tests look comprehensive enough (quick look).

arvindsv:
> The current implementation in BuildSession do have a problems: All child processes are killed before cancel task run. This make user have no chance to stop the running command gracefully. I can make change to the order, make sure the cancel task runs first, then at last kill all child process, make sure no leaking processes.

I think that makes sense. It'll still be a change in behavior, but I'd like to think it's better than what it is right now (to not leave processes around after cancel task has run).

wpc:
> Earlier, if it failed to report completion, it would report failure. I've actually seen it fail to report final completion and fail the build (mainly because the server was very busy). Now, it kind of leaves that job hanging, since it just logs failures around here. I know that the server will see that the same agent is now asking for work and will reschedule the job. But, was that the idea?

I think the original catch clause in BuildWork is for Exception thrown from build, not for reportingCompleted call.  Say agent was reporting "success" result, and failed, likely report "failed" will be fail again. Even in the small chance that second report passed, it should send "success" instead of "failed", because the build is succeeded. So the old behavior (if implemented rightly) is a "retry" instead of report failure.

With our current websocket implementation, all message are sent asynchronously. The only exception we can got is "Connection is closed" (jetty doing that check before insert a frame into asyn send buffer). I think in that situation there  is no use for retry it.

Anyway server should not assume result always get reported, need handle the case anyway.

> I think that there's a behavior change around how commands are executed and how secrets are (not) printed. It looks like [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/util/command/CommandLine.java#L453-L457) is missing somewhere around [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java#L80). 

It is handled [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java#L64) in the same file 


> The equivalent of this test actually (and those like shouldLogPasswordsOnOutputAsStarsUnderLinux, shouldNotLogPasswordsOnExceptionThrown, etc) ... That makes me think: Are there other missing tests?

Most tests for those concerns are in BuildSessionTest. I generally use exec("echo") to test exec output because, believe it or not, echo command has same syntax on both linux and windows :-). So I can avoid a lot of OS specific test. Equivalent of shouldLogPasswordsOnOutputAsStarsUnderLinux is [here](https://github.com/wpc/gocd/blob/build_command_protocol/common/test/unit/com/thoughtworks/go/buildsession/BuildSessionTest.java#L377), but I did forget shouldNotLogPasswordsOnExceptionThrown. I will go through CommandLineTest.java make sure all logic tested get covered.

```
@Test
public void shouldNotPrintSecureVariableValuesInConsole() throws Exception {
    runBuild(compose(
            export("SECURE", "abc", true),
            exec("echo", "Should not see: abc")), Passed);
    assertThat(console.output(), not(containsString("Should not see: abc")));
    assertThat(console.output(), containsString("Should not see: ******"));
}
```
In old BuildWork implementation there is actually only two place use SafeOutputStreamConsumer:  CommandLine output and dumping environment variables info. So the echo actually should not mask the secrets. My guess is it do not want to generate message like "[go] Job is ******" and user immediately  know secret is "completed" :-)

> I feel that tests like RegistrationTest.java can get some more love. :)

Will make sure they get enough :-)

arvindsv:
> It is handled [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/common/src/com/thoughtworks/go/buildsession/ExecCommandExecutor.java#L64) in the same file

Oh, I see. Sorry about that. I thought it was only about not showing it in the command-line's display. But, I see that the secrets are properly passed on to the ProcessWrapper. I also see what was wrong with the test I showed there. Using "export" does not make the value a secret. I remember that `BuildComposer#setupSecrets` is the one that sets it up. So, your test makes more sense.

ketan:
> * Question to @ketan, I guess: Should [this](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java#L115) be called only after a successful registration (somewhere around [here](https://github.com/wpc/gocd/blob/cac818fe63ef5d1854ca2972ee40ccf5a110a548/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java#L125))?

Yes, I'd like it to be so, if at all possible. I've seen initial connection failures cause registration to fail (esp in cases of elastic agent cases) where the chances of something like this happening is higher due to the frequent agent creation/termination.

wpc:
@arvindsv:  latest protocol document: https://github.com/gocd-contrib/gocd-golang-agent/wiki/BuildCommand-Protocol, may help you review the code

wpc:
@arvindsv:

In last couple of commits I made the changes we discussed about:

* Extract test helper TestStreamConsumer for avoid polluting InMemoryConsumer (1108b1a)
* Remove unused AgentInstance#supportsBuildCommandProtocol (1108b1a)
* Make server send same cancel build message despite of support build command protocol supported or not (77235e4)
* Split more executor tests from BuildSessionTest (42dc7d6)
* Make cancel task run before kill all subprocess (42dc7d6)
* Harden ExecCommandExecutor, make sure secret not leak into either console or log, and port tests from CommandLineTest. (42dc7d6)
* Add more assertions for RegistrationTest (b9336ef)

Please take a look and let me know if anything is missed.

wpc:
For the concern of migrating old tests, the relationship from new/old test is sort of like following:

|New test                  |  partly ported from and should covers                                                           |
|----------------------|----------------------------------------------------------------------|
|BuildComposerTest               | BuildWorkTest.java + BuildWorkArtifactFetchingTest + BuildWorkArtifactUploadingTest |
|GitMaterialUpdaterTest| GitMaterialTest + GitMaterialShallowCloneTest + GitCommandLineTest|
|FetchArtifactBuilderBuildCommandTest| FetchArtifactBuilderTest|
|UploadArtifactCommandExecutorTest| BuildWorkArtifactUploadingTest + ArtifactPlanTest |
|UrlBasedArtifactsRepositoryTest| GoArtifactsManipulatorTest|
----------------

arvindsv:
Nice. All new commits look good. I really like the new test structure in 42dc7d6fbf005e2b95230120ed6fd61cf6aff294.

The only thought/concern I have is this:

On my box, at least, even if I comment out the line containing "new JavaSysMon().infanticide();", all tests in BuildSessionCancelingTest pass. That suggests that a test is missing. Maybe we should have a test with a second-level process or something, to see if it is left behind? Or, maybe some fix to the test? It looks like the `executorService.shutdownNow()` would kill the running sleep in the test anyway.

arvindsv:
Forgot to mention: Not that it matters too much, but [this](https://github.com/wpc/gocd/blob/build_command_protocol/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L228) should probably change too, right?

wpc:
>  Maybe we should have a test with a second-level process or something, to see if it is left behind? Or, maybe some fix to the test? It looks like the executorService.shutdownNow() would kill the running sleep in the test anyway.

I have tried hard and cannot find a way to write that test. Logically `executorService.shutdownNow()` should interrupt and kill the subprocess, and all descendant process should be gone with it. So the second level process will not do trick. The only way I figured to make executorService.shutdownNow() leak a process is spwan a process on trap signal:
```bash
#!/bin/bash
function finish {
    echo yawn
    sleep 50
}
trap finish SIGINT SIGTERM
sleep 50
````
But in this case even JavaSysMon.infanticide can not kill it.

So I think we leave there JavaSysMon.infanticide is just for incase default process killing not work on some platform. (Does it really happens before?)

arvindsv:
Yeah, I thought it would be like that. I think, earlier, the killing was dependent on the infanticide. Right now, it is it not. So, I agree with leaving it in as a last resort.

arvindsv:
Looks like it's all good, @wpc! Thank you!

PR #2057, opened 2016-03-14
Fix broken page when displaying unescaped errors

zabil:


maksat-software:
No description provided.

zabil:
@maksat-software description provided.

PR #2062, opened 2016-03-15
Enforce version attribute of `<go-plugin/>` element.

ketan:


jyotisingh:
@ketan - Changing the type to `int` obviously makes sense, but cannot recollect why have a fixed value of `1` for version number. Did you really mean `fixed` value or a `default` value? Please update the PR with the reason, I remember us speaking about it but have no recollection of the reasons.

ketan:
@jyotisingh — the `version` attribute is the equivalent of `schemaVersion` in the `cruise-config.xml` file. As such needs to start with 1, until we figure out some migration/upgrade story for the plugin.xml

PR #2076, opened 2016-03-25
fix: lost last line log when put console log has multiple lines and no "\n"

xli:
Current Java agent always put console log with whole lines.
So this change will not effect current Java agent. 
However, if we decided to change to put half line log so that user can see a long time junit test suite running progress, then this change makes server side ready to process those kind of log.

I need this change because new golang agent will not base on new line to put console log, it will just put all console log buffer when tried to flush local console log buffer to server.
It is better because we can see long running junit test suite running progress. here is example:
output_dultlp

### DIFF
--- a/server/src/com/thoughtworks/go/server/controller/ArtifactsController.java
+++ b/server/src/com/thoughtworks/go/server/controller/ArtifactsController.java
@@ -304,7 +304,7 @@ public class ArtifactsController {
 
     private ModelAndView putConsoleOutput(final JobIdentifier jobIdentifier, final InputStream inputStream) throws Exception {
         File consoleLogFile = consoleService.consoleLogFile(jobIdentifier);
-        boolean updated = consoleService.updateConsoleLog(consoleLogFile, inputStream, ConsoleService.LineListener.NO_OP_LINE_LISTENER);
+        boolean updated = consoleService.updateConsoleLog(consoleLogFile, inputStream);
         if (updated) {
             consoleActivityMonitor.consoleUpdatedFor(jobIdentifier);
             return FileModelAndView.fileAppended(consoleLogFile.getPath());
--- a/server/src/com/thoughtworks/go/server/service/ConsoleService.java
+++ b/server/src/com/thoughtworks/go/server/service/ConsoleService.java
@@ -25,6 +25,7 @@ import com.thoughtworks.go.server.view.artifacts.BuildIdArtifactLocator;
 import com.thoughtworks.go.server.view.artifacts.PathBasedArtifactsLocator;
 import com.thoughtworks.go.util.FileUtil;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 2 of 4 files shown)


ketan:
:heart:

xli:
The  build-windows-PR/build-non-server-windows failure is a  SvnCommandTest#shouldSupportUTF8CheckInMessageAndFilename, it failed when I rerun the job. Don't understand why it keeps fail.

jyotisingh:
@xli - I was looking at the reason behind the current implementation. Looks like the original version of the method was :

``` java
public boolean updateConsoleLog(File dest, String content) throws IOException {
        File parentFile = dest.getParentFile();
        parentFile.mkdirs();

        FileOutputStream out;
        if (LOGGER.isTraceEnabled()) {
            LOGGER.trace("Updating console log [" + dest.getAbsolutePath() + "]");
        }
        if (dest.exists()) {
            out = new FileOutputStream(dest, true);
        } else {
            out = new FileOutputStream(dest);
        }
        try {
            IOUtils.write(content, out);
        } catch (IOException e) {
            LOGGER.error("Failed to update console log at : [" + dest.getAbsolutePath() + "]", e);
            return false;
        } finally {
            out.close();
        }
        if (LOGGER.isTraceEnabled()) {
            LOGGER.trace("Console log [" + dest.getAbsolutePath() + "] saved.");
        }
        return true;
    }
```
This was later changed to the current version in order to fix some performance issues (commit message from the old repo)
```
commit e45098bdbae1ddf990e98f36cb4189acbc5f31f8
Author: <removed>
Date:   Wed Sep 21 20:59:28 2011 +0530

     #5488 - moved to fixed size, expandable char array based console-log post, so regex runs on preallocated char-array instead of a new string everytime(or wasting huge string by loading the entire thing in mem)
```

It looks like the original version had issues because of strings, but instead of using inputstream  directly and `IOUtils.copy(in, writer)`, they chose the current implementation. Wondering why?
I am unable to point out issues with this PR, but before I merge this can you have a look at that commit just to see if you find anything?

xli:
There was another feature removed later, and now it is meaningless to do the whole work.
The following was the place I believe had the issue (not 100%, but I think it was related):

``````java
-    public boolean updateBuildEvent(String content) {
-        String matchedBuildEvent = matcher.match(content);
+    public boolean updateBuildEvent(CharSequence consoleLine) {
+        String matchedBuildEvent = matcher.match(consoleLine);
``````

Here `matcher.match(consoleLine);`  is BuildOutputMatcher which is still in codebase but no place is using it. (probably should be removed too).
The regex in BuildOutputMatcher matches output line by line.
The whole change you mentioned is trying to split console log into lines, then add `ArtifactsService.LineListener` to call `scheduleService.updateBuildEvent` line by line.

``````java
+        boolean updated = artifactsService.updateConsoleLog(artifact, inputStream, new ArtifactsService.LineListener() {
+            public void copyLine(CharSequence line) {
+                scheduleService.updateBuildEvent(jobIdentifier, line);
+            }
+        });
``````

xli:
Just checked BuildOutputMatcher. There is ConfigElementImplementationRegistrar register it. Is it used in a plugin?

jyotisingh:
> The whole change you mentioned is trying to split console log into lines, then add ArtifactsService.LineListener to call scheduleService.updateBuildEvent line by line.

Yes. That was apparently done to fix a performance issue which had to do with handling of large statements I think. 

You are right, none of them seem to be used anymore. `BuildOutputMatcher` is indeed dead code that could be removed. `ConfigElementImplementationRegistrar ` is not plugin code, it is used for deserialisation of `CruiseConfig`, but we could remove the registration of `BuildOutputMatcher` from there. Found `BuildMatcherHandlerCallback` too which is unused. Both `BuildOutputMatcher` and `BuildMatcherHandlerCallback` are being registered in `sql-map-config.xml` - all of that can go away too.

xli:
@jyotisingh is there anything else you think I should do for this PR?

jyotisingh:
No, nothing more for this PR. I was waiting for 16.3 to be out in order to merge this PR. 
Its not necessary, but it would be nice if you could send another PR removing `BuildOutputMatcher` and its usages too. Merging this one.

xli:
@jyotisingh I'm still busy on finishing first working version of golang agent. I tried simply delete BuildOutputMatcher and related stuff, but there are failures after I fixed all obvious compiling errors. 
So, I probably won't have time to completely remove it.

jyotisingh:
@xli - ok sure.

PR #2080, opened 2016-03-29
forcing utf-8 encoding while writing config xml to file

jyotisingh:


server/test/integration/com/thoughtworks/go/config/GoFileConfigDataSourceIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+        ReflectionUtil.setStaticField(Charset.class, DEFAULT_CHARSET, Charset.forName("windows-1252"));
+        BasicCruiseConfig config = new BasicCruiseConfig();
+        GoFileConfigDataSource.GoConfigSaveResult result = dataSource.writeWithLock(new UpdateConfigCommand() {
+            @Override
+            public CruiseConfig update(CruiseConfig cruiseConfig) throws Exception {
+                PipelineConfig pipelineConfig = PipelineConfigMother.pipelineConfig(UUID.randomUUID().toString());
+                JobConfig job = new JobConfig("job");
+                ExecTask task = new ExecTask();
+                task.setCommand("powershell");
+                task.setArgs("Get-ChildItem -Path . â€“Recurse");
```
arvindsv: What happens to content such as this? I suspect it will be gibberish, but I wonder what the effect of this on an existing job is. Does it fail?

jyotisingh: Yes, it would be gibberish and yes the job will fail stating whatever is wrong. 
Would you prefer not saving the config itself if such content was present?

arvindsv: No, I think it's better to do what you've done. Otherwise, it'll cause issues during an upgrade. Now, if a job has this, and it was working, it'll fail in 16.4.0, right? If so, I think that's ok.

arvindsv:
@jyotisingh: Shouldn't this be merged sometime now?

server/test/integration/com/thoughtworks/go/config/GoFileConfigDataSourceIntegrationTest.java
jyotisingh: > if a job has this, and it was working, it'll fail in 16.4.0, right? If so, I think that's ok.

Yes it will fail

jyotisingh:
@arvindsv  - I was hoping for it to get merged, but no sure anyone reviewed it yet, so..

PR #2083, opened 2016-03-30
Plugin upload toggle + API custom header

maheshp:


matt-richardson:
Can you explain the reasoning why this random custom header was added? I didn't see any proposal or anything other than this commit? 

Wont this break backwards compatibility?

zabil:
Hey @matt-richardson 

They've been added to force a pre-flight (OPTIONS) on Cross origin requests.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests

We'll putting blog post about this in a week or two.

matt-richardson:
Okay, so now it forces a pre-flight check. Why is that worth breaking lots of apps/scripts?

zabil:
Security. It prevents CSRF attacks.

zabil:
There's a GoCD server system property you can use to toggle it

```
-Dgo.api.with.safe.mode=N
```

Use it with great caution as this will open up the server to api exploits.

matt-richardson:
Ahh, cool. That makes sense now - I hadn't seen anything about it.

Issue #1849, opened 2016-01-28
Console page seems slow sometimes, when it has lots of data (20MB page)


arvindsv:
[slow_job_console_load.har.txt](https://github.com/gocd/gocd/files/108416/slow_job_console_load.har.txt)

Reference: https://gitter.im/gocd/gocd?at=56aa33fb8fbaf4220af9e000

BrewRunner:
This is painfully slow for many of our pipelines.  The previous versions that did not colorize the log loaded much much faster.  Is there a way to turn off the white on black theme and receive console log output "the old way"?

zabil:
Profiled this for a log of 223kb with 2662 lines without ansi colors

![screen shot 2016-04-01 at 5 52 15 pm](https://cloud.githubusercontent.com/assets/54427/14206712/a4348a8a-f832-11e5-83f9-ddef62511364.png)

It takes close to 7 seconds to perform [this](https://github.com/gocd/gocd/blob/aa5a08467523ff8875f3366bead899d4dd193003/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js#L145) operation.

Just toggling off the ansi up coloring will not improve the performance. Gets exponentially bad as the size of the file goes up throwing the dreaded unresponsive page error.

Rendering the raw output as an iframe on the same page loads the page solves the issue. (The page loads without delay). However the console log won't be colored. 

![screen shot 2016-04-01 at 6 03 17 pm](https://cloud.githubusercontent.com/assets/54427/14206900/1c4d82aa-f834-11e5-9605-71b85a55c0f3.png)

BrewRunner:
Yes!  That's it exactly. The console log has become largely unusable for the majority of our pipelines, most of which, produce logs that are larger than what you are testing with. An option to just turn off the formatted console output and use raw would be ideal for users with "lengthy" console logs.

ketan:
Can you share the raw logs via a gist so we can use that data to run some browser profilers?

BrewRunner:
Here's a somewhat lengthy example: https://gist.github.com/BrewRunner/b980afbc3caab727ff4de6c49c594419

wpc:
I got some time today and had quick look of the problem. The code appending log content to the DOM in BuildOutputObserver looks like problematic:
```javascript
while(lines.length){
    var slice = lines.splice(0, 1000);
    var htmlContents = this.ansi_up.ansi_to_html(slice.join("").escapeHTML(), {use_classes: true});
[...]

PR #2094, opened 2016-04-08
Issue #1849 and some other page performance improvements

wpc:
* Fix issue #1849: A simple workaround that incrementally insert log lines into the DOM to avoid page freezing.  The fix also handles the case when user switch to console tab from other tabs.

* In ViewportPredicate module, void calling jQuery(window).scrollTop in onScroll handler. Because window.pageYOffset triggers DOM reflow, it is super expensive when there are new contents inserted into the DOM.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_output_observer.js
@@ -40,6 +40,7 @@ BuildOutputObserver.prototype = {
         this.consoleTabElement.on('click', function(){
             self.initializeScroll();
         });
+        this.outputBuffer = [];
         this.initializeScroll();
     },
     initializeScroll: function(){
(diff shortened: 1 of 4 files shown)


mrmanc:
Thanks for this PR @wpc - this has affected us recently very heavily with at least one of our pipeline jobs so this will help us out.

wpc:
@ketan and @zabil: this PR is ready for review

ketan:
Naïve question (I may be wrong here):

On the ajax poll that happens every 5 seconds, is there a possibility that the async `_renderOutput` that is called from one call of `_update_live_output_color` overlaps with the next call of `_update_live_output_color`, causing console log mangling?

Implying that only when `_renderOutput`'s `done()` arg is invoked, should the next ajax call be done (by bubbling the `done` call right upto the caller.

Happy to explain it on IM, if what I'm saying makes no sense whatsoever.

wpc:
@ketan:

It is handled [here](https://github.com/gocd/gocd/pull/2094/files#diff-d08449c7cf568e81489394835c957f35R142) in _update_live_output_color
```js
this.outputBuffer = this.outputBuffer.concat(build_output.match(/^.*([\n\r]+|$)/gm));
if(this.renderOutputTimeout) {
   // rendering is going on, new lines appended to buffer
   // will be picked up anyway.
   return;
}
```

if there are renderOutputTimeout, the function will just append the fetched output into the unrendered output buffer. The current running rendering process will take care of render them at last.

There would be an order issue because in JS everything runs in one single thread.

zabil:
@wpc 

This change is deployed on build.go.cd
My environment.
Mac OSX El Capitan
Chrome Version 49.0.2623.110 (64-bit)

Noticed a few issues 
For build https://build.go.cd/go/tab/build/detail/build-linux/868/build-server/1/server-split-runInstance-1

Try scrolling up or down. You get a frozen browser with a screen like this

screen shot 2016-04-12 at 9 41 11 pm

The scroll to end does not scroll to end, this happens sometimes when console tab is not the default tab loaded. For e.g

Open https://build.go.cd/go/tab/build/detail/build-linux/868/build-non-server/1/new_jasmine_tests

Cick on the "artifacts" tab.
Click on "new_jasmine_tests" from the stage page
https://build.go.cd/go/pipelines/build-linux/868/build-non-server/1

This happens. (notice scroll to bottom is green but it's not at the end of the console output)

screen shot 2016-04-12 at 9 58 06 pm

wpc:
@zabil:

I tried it myself on build.go.cd. With console output a little bit bigger, the user experience is not acceptable. The big black screen is because the repainting is blocked.  Could you help me revert 557b07995ce4c41bf9e78a1a401550a47ae298db. We need figure out another way.

zabil:
I'll revert it in a bit.

It works on smaller log files. 

Instead of dealing with the DOM overhead of large log files should we send a HEAD request to figure out the content-length and handle differently for large log file (again like before render without coloring in an iframe) and give out a warning?

wpc:
I feel the performance would always be bad if we insert so much things into the DOM.  The only thing can work is some sort of lazy rendering. That can be implemented pure client side without much work: render fix height place holder for each line, and fill visible placeholders and clear invisible placeholder during onscroll. The trouble would be search. Replicate browser in-page search functionality seems a lot work.

arvindsv:
I like that idea (render X lines, starting from the end). If someone scrolls up, then we can provide a link to the full log (preferably opening in a different tab). I don't see the need to replicate in-page search. A responsive page which shows a subset of the content quickly seems preferable to an unresponsive page which tries to load all content.

@zabil has said he'll try it. But, if you want to take a stab at it, that's fine too I guess. You'll just need to co-ordinate with @zabil to make sure you don't both end up working on it.

zabil:
Yeah. I am picking this up. 

Assuming that to quickly debug errors it's enough to display the last 1000-2000 lines. 
The raw output is available for search or complete output.

wpc:
Sounds fine to me.

matt-richardson:
Please don't just display a partial log... I've had to log bugs against some atlassian products about that. People use and rely on the in-browser search functionality.

All that said, I dont know how to handle super-large logs...

zabil:
@matt-richardson you can view entire log using the "raw output" link. Isn't that good enough?

d-led:
there's a similar discussion on [travis-ci#927](https://github.com/travis-ci/travis-ci/issues/927). Their solution is limiting the log. I'd also prefer not to have a partial log, as it will feel like a disruption of a smooth log viewing experience

mrmanc:
I have to agree that a partial log would be very frustrating. It is already frustrating having to wait for the page to load fully before being able to use ⌘-f to search for stuff on the pipelines, environments and admin pages. I'd rather see a raw but complete log than a pretty partial one.

zabil:
@mrmanc this is what we have at the moment.

screen shot 2016-04-14 at 1 48 22 pm

The "Raw output" button will display the entire log. 
Also, the log will be truncated only if it's larger than 1000 lines.

We are trying to make it less frustrating by displaying the "head" part of the console, which displays materials and environment variables, plus the tail part to quickly debug failures.

There's simply no way to display a log larger than 5 mb without freezing up the UI, it's a browser limitation.

Again this is for builds that have completed.
While building we'll show the entire log with live tailing, however depending on what part of the build you are at it will get sluggish.

zabil:
Interesting discovery made here https://github.com/gocd/gocd/issues/1849#issuecomment-209810151

PR #2096, opened 2016-04-08
A proposed change to how authentication is enabled

danielsomerfield:
This is one possible way that Go CD could decide whether to enable auth is by asking the auth plugins. Here is an example implementation of that. There are a few things I'm unsure of:
- I'm handling versioning very naively, and I expect it has to be handled better. I'm covering cases where the plugin doesn't know it has to respond to by handling 404s, but my guess is there are better ways but I don't quite understand the API versioning model yet
- I'm hitting the plugins on every auth request to the IsSecurityEnabledVoter and I would guess it should be cached in some ways
- I like the idea of ultimately removing the hard coded checks to the two built-in providers, but that probably has to wait until the file-based one goes away and the LDAP one is broken out into a plugin.

Before making those changes, I wanted to propose the general approach and see what you thought.

Let me know if you think it's valid, or if you have another way.

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRequestHelper.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authentication/AuthenticationPluginRequestHelper.java
@@ -1,0 +1,59 @@
+package com.thoughtworks.go.plugin.access.authentication;
+
+import com.google.gson.Gson;
+import com.thoughtworks.go.plugin.api.request.DefaultGoPluginApiRequest;
+import com.thoughtworks.go.plugin.api.response.DefaultGoApiResponse;
+import com.thoughtworks.go.plugin.api.response.GoPluginApiResponse;
+import com.thoughtworks.go.plugin.infra.PluginManager;
+import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+
+@Component
+public class AuthenticationPluginRequestHelper {
+
+    private static final Logger LOG = Logger.getLogger(AuthenticationPluginRequestHelper.class);
+    private PluginManager pluginManager;
+
+    private final static String REQUEST_NAME = "go.authentication.auth_enabled";
+    private final static String EXTENSION_NAME = "authentication";
+    private static final List<String> GO_SUPPORTED_VERSIONS = asList("1.0");
+
+    @Autowired
+    public AuthenticationPluginRequestHelper(PluginManager pluginManager) {
+        this.pluginManager = pluginManager;
+    }
+
+    public boolean isEnabled(String pluginId) {
+        if (!pluginManager.isPluginOfType(EXTENSION_NAME, pluginId)) {
+            throw new RuntimeException(format("Did not find '%s' plugin with id '%s'. Looks like plugin is missing", EXTENSION_NAME, pluginId));
+        }
+        try {
+            final String resolvedExtensionVersion = pluginManager.resolveExtensionVersion(pluginId, GO_SUPPORTED_VERSIONS);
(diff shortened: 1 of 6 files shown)


arvindsv:
I was thinking about this for a bit. I am willing to work on an LDAP plugin which removes the existing functionality from GoCD. And with your current file plugin, we can bundle both and keep the functionality backward compatible.

However, my thought around the auth enabled check is: What would happen if we made this happen through configuration instead? I mean that today, the security config looks like this:

```xml
<security>
  <ldap uri="ldap://xxx.yourcompany.com"
        managerDn="managerDn"
        managerPassword="password"
        searchBase="ou=Employees,ou=Enterprise,ou=Principal,dc=xxxx,dc=yyyy,dc=com"
        searchFilter="(sAMAccountName={0})" />
  <passwordFile path="/home/go/admins.properties"/>
  <roles>
  </roles>
  <admins>
  </admins>
</security>
```

If we time it right, and have both plugins bundled together, we can do an automatic config migration which converts that into something like this:

```xml
<security>
  <providers>

    <provider enabled="true">
      <plugin id="ldap-auth-plugin" version="1.0" />
      <configuration>
        <property><key>uri</key><value>ldap://xxx.yourcompany.com</value></property>
        <property><key>managerDn</key><value>managerDn</value></property>
        <property><key>managerPassword</key><value>password</value></property>
        <property><key>searchBase</key><value>ou=Employees,ou=Enterprise,ou=Principal,dc=xxxx,dc=yyyy,dc=com</value></property>
        <property><key>searchFilter</key><value>(sAMAccountName={0})</value></property>
      </configuration>
    </provider>

    <provider enabled="true">
      <plugin id="file-auth-plugin" version="1.0" />
      <configuration>
        <property><key>file</key><value>/home/go/admins.properties</value></property>
      </configuration>
    </provider>

  </providers>
  <roles>
  </roles>
  <admins>
  </admins>
</security>
```

The order of the provider config, along with the "enabled" attribute should be enough information to decide whether auth is enabled, right?

At the server configuration level, we could allow the user to change the order of the available plugins and also enable or disable them. Everything else (configuration properties included) are controlled by the plugin.

arvindsv:
Of course, the other auth plugins should continue to work as they do. We might have to move their stored configuration over into this place as well.

arvindsv:
We will have to think about how to save the config, though. It's not currently stored in the file, unfortunately. It's in the DB.

danielsomerfield:
Sorry for the delay. In inception.

- the plan do move it to config would work, but that would not take into
account whether a plugin was actually configured correctly, which is why I
was of the mind to push it to the plugins. But I assume with your approach,
the plugin would simply return a response without a authenticated user, and
the extension would move on to the next plugin to see if it had an opinion.
I think that would be ok.
- not sure what to say about the config in a DB. That's kind of a nuisance.
Not sure how to handle that.

I also think you need to decide when to deprecate the old plugin and when
to disable it entirely. It's simply not a safe way to keep it entirely
backwards compatible. If you don't rip off the bandaid at some point,
people just won't change. Forcing them to change with an upgrade path is
one path, but then we still need to force password changes eventually
because they should not be using the ones that were stored without a salt.

Finally, I'm still a little fuzzy how to handle the user management. I know
you were talking about a UI, but I wasn't clear on whether that was just a
UI in the plugin itself, or something at a higher level. UI is a lot of
work, so I don't want to build that in a vacuum, particularly now that I'm
staffed. I can continue to work on it on evenings and weekends, but I want
to make sure we know more or less what we're building, so it doesn't become
a time suck.

arvindsv:
I understand. I've been swamped too. I would like to take a step back and think about this. I don't want it to seem like I'm taking advantage of your goodwill in helping. So, I'd like to have a proper plan and a proper migration plan before I get back to you on what I think we should do.

Nagging people and providing them a reasonable approach to changing their password makes sense to me. If we don't provide them a good and easy way to migrate (while changing their password), then we won't be able to "force" them to change.

I will think about this some more.

@gocd/committers: If you have any thoughts, feel free to intervene.

ketan:
Closing in favor of #2956

Issue #1411, opened 2015-08-03
Spaces are not trimmed from environment variables


JorritSalverda:
I was under the impression I was overriding an environment variable, but judging from this log it turns out that one of the environment variables has a space at the end. I would expect that to be trimmed, as it's unlikely to be valid on any operating system.

14:30:09.878 [go] setting environment variable 'CACHE_CHECK_ERROR_ACTION' to value 'Stop'
14:30:09.893 [go] setting environment variable 'CACHE_CHECK_ERROR_ACTION ' to value 'Continue'

rajiesh:
verified in GoCD version 16.10.0 (4129-e51ab2ff4bc66d88a93254c209ee60e42a1112bc). Working as expected

Issue #2089, opened 2016-04-05
On build.go.cd view user can see the Pause pipeline button.


Junaidshah:
##### Issue Type

- Bug Report

##### Summary

On build.go.cd ```view``` user can see the Pause pipeline button. Clicking on the Pause button logs the user out.

Attached a screen shot from build.go.cd below. The user is logged in as a view user who has only read access. 

<img width="543" alt="screenshot 2016-04-05 11 04 32" src="https://cloud.githubusercontent.com/assets/5434680/14272187/3ec75670-fb1e-11e5-938c-02f47d4f2469.png">

Issue #63, opened 2014-04-16
Disabled users still get notifications on pipeline events


mdaliejaz:
Description
-----------

Disabled users still get notifications on pipeline events

### Steps to reproduce

- Log in as USER\_A

- Go to Preferences page and set notifications filter for pass event of pipeline X

- Log in as ADMIN and disable USER\_A

- Run pipeline X and let it pass

#### Expected result

Disabled users shouldn't receive notifications

#### **Current result**

Even thought USER\_A is disabled, user still gets notifications

curvedental:
We've run into this a bunch of times when disabling users since we forward emails from exiting employees to their respective manager, it would be nice if we didn't have to log in as the user to disable notifications after the fact.

PR #2097, opened 2016-04-09
fixed issues : #1411  #63  #2089

GaneshSPatil:


ketan:
Ganesh, we never fix multiple issues in one PR. Reason being that if we have to revert the PR for one fix, it'll revert all fixes. Please submit one PR for each issue you fix

GaneshSPatil:
Can you tell me how to create separate PRs

PR #2105, opened 2016-04-13
fixed issue: #2089 On build.go.cd view user can see the Pause pipeline button

GaneshSPatil:


arvindsv:
This changes functionality a bit (by removing those buttons for users who can't pause it). Hmm. That's fine, I think.

However, you should probably write a test for this, like [this](https://github.com/gocd/gocd/blob/master/server/test/unit/com/thoughtworks/go/server/view/velocity/MyCruisePreferencePageVelocityTemplateTest.java) or [this](https://github.com/gocd/gocd/blob/master/server/test/unit/com/thoughtworks/go/server/view/velocity/BuildDetailPageVelocityTemplateTest.java).

ketan:
Functionality should not change. IMO everyone should see the button, albeit disabled, with a tooltip explaining why "you don't have sufficient permissions"

GaneshSPatil:
Functionality should not change.. agree..
But the view user is also not able to see the admin privileges such as pipeline setting button, etc

zabil:
@ketan, @GaneshSPatil is right about consistency.

Anything you had in mind to display it nevertheless?

PR #2106, opened 2016-04-13
Handling eviction/expiry of composite-keys from GoCache

jyotisingh:
When a composite key is saved to cache, there are two entries created in the cache. One for the composite-key itself containing the actual value. Another entry to keep track of all sub-keys for a given parent. When a parent cache entry is explicitly removed, the corresponding caches for subkeys are removed as well. When a subkey is explicitly removed, the sub-key list with the parent cache entry is updated as well. However, when a sub-key entry gets evicted from the cache when it gets full or when it expires, the parent cache entry wasn't updated. This commit fixes that.

### DIFF
--- a/server/src/com/thoughtworks/go/server/cache/GoCache.java
+++ b/server/src/com/thoughtworks/go/server/cache/GoCache.java
@@ -1,18 +1,18 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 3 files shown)


server/src/com/thoughtworks/go/server/cache/GoCache.java
```diff
@@ -194,6 +196,23 @@ public void removeAll(List<String> keys) {
         }
     }
 
+    public void removeCompositeKeyFromParentCache(String key) {
+        if (key.contains(SUB_KEY_DELIMITER)) {
+            String[] parts = StringUtils.splitByWholeSeparator(key, SUB_KEY_DELIMITER);
```
zabil: Are these safe from null checks and index checks?

jyotisingh: `parentKey` cannot be null as that is one of the keys in the map maintained by ehcache, so it would fail at the time of insertion itself. `childKey` would be null if the key used in the cache is "parent!_#$#_!", but then it would lead to other cache related errors as all such composite caches will have the same key. Moreover `StringUtils.splitByWholeSeparator("parent!_#$#_!", "!_#$#_!")` returns 
```
result = {java.lang.String[2]@1305} 
 0 = "parent"
 1 = ""
```
so we should be ok I think.

PR #2109, opened 2016-04-15
fixed issue: #1411 Spaces are not trimmed from environment variables

GaneshSPatil:
added a validation for leading and trailing spaces of environmentVariable name. 
Added a xsl transformation to trim the existing environment variable name from cruise-config.xml file

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
@@ -172,6 +172,10 @@ public class EnvironmentVariableConfig extends PersistentObject implements Seria
         String currentVariableName = name.toLowerCase();
         String parentDisplayName = validationContext.getParentDisplayName();
         CaseInsensitiveString parentName = getParentNameFrom(validationContext);
+        if(currentVariableName.startsWith(" ") || currentVariableName.endsWith(" ")){
+            configErrors.add(NAME, String.format("Environment Variable cannot start or end with spaces for %s '%s'.", parentDisplayName, parentName));
+            return;
+        }
         if (StringUtil.isBlank(currentVariableName)) {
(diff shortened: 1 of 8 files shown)


ketan:
Should't we also be trimming values too? Anyone have any opinions on that trimming or not trimming ALL form fields?

I feel it's easy for users to add a whitespace that is not noticeable, but for cases where you really NEED the whitespace, this sort of forced trimming may in the way, with no way out for the user.

ketan:
Should we also need something similar to [this](https://github.com/gocd/gocd/commit/2b55b7cd7ba20c9d2fa06efb20d7fa5a8d557cbd#diff-bdfabb73e2af551ffd8e50a573a66338R715) to prevent direct XML changes that'd cause a whitespace to be added?

GaneshSPatil:
leading and trailing spaces are allowed in the VALUE of the environment variable. But not in the name of the environment variable.
will it be right to prevent the user from user from setting it?

arvindsv:
Please don't forget to close #1411, once this is merged.

config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
```diff
@@ -172,6 +172,10 @@ public void validateName(Map<String, EnvironmentVariableConfig> variableConfigMa
         String currentVariableName = name.toLowerCase();
         String parentDisplayName = validationContext.getParentDisplayName();
         CaseInsensitiveString parentName = getParentNameFrom(validationContext);
+        if(currentVariableName.startsWith(" ") || currentVariableName.endsWith(" ")){
```
ketan: @zabil — I'd not merged this because this is a really bad test.

Ideally we'd want to test for all whitespace, not just a space character. Basically test whether `currentVariableName.trim().equals(currentVariableName)`

zabil: Damn, I thought all feedback was worked on, when i was pinged. My bad.
@GaneshSPatil can you use trim instead of this?

PR #2113, opened 2016-04-16
Handling removal of subkeys from cache when a parent entry gets evict…

jyotisingh:
…ed/expired from cache

### DIFF
--- a/server/src/com/thoughtworks/go/server/cache/GoCache.java
+++ b/server/src/com/thoughtworks/go/server/cache/GoCache.java
@@ -16,11 +16,6 @@
 
 package com.thoughtworks.go.server.cache;
 
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-
 import com.thoughtworks.go.domain.NullUser;
 import com.thoughtworks.go.domain.PersistentObject;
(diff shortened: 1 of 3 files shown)


server/src/com/thoughtworks/go/server/cache/GoCache.java
```diff
@@ -200,14 +200,20 @@ public void removeAll(List<String> keys) {
         }
     }
 
-    public void removeCompositeKeyFromParentCache(String key) {
-        if (key.contains(SUB_KEY_DELIMITER)) {
+    public void removeAssociations(String key, Element element) {
+        if (element.getValue() instanceof KeyList) {
+            synchronized (key.intern()) {
+                for (String subkey : (KeyList) element.getValue()) {
+                    remove(compositeKey(key, subkey));
```
arvindsv: Do you want to do remove here, or ehcache.remove?

arvindsv: I'll probably merge this in now since it makes sense and is needed, but we could look later at [pulling this part out](https://github.com/jyotisingh/gocd/blob/69747ef2bdbca794a4fad15fdc292723e47352fe/server/src/com/thoughtworks/go/server/cache/GoCache.java#L171-L175) into its own method and using it here.

Issue #2119, opened 2016-04-19
Pipeline Config API does not allow Pipeline Group Admin to create/edit a pipeline.


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

A Pipeline Group Admin should be allowed to use the Pipeline Config API to create/edit/get a pipeline fr

###### Basic environment details

* Go Version: 16.3.0
* JAVA Version: 1.7

##### Steps to Reproduce

1. Add user 'operate'. 
2. Make 'operate' user a Pipeline Group Admin for an existing pipeline group, which would have a pipeline within, say 'foo'.
3. GET /go/api/admin/pipelines/foo

##### Expected Results
Pipeline Group Admin, that is the operate user must be allowed to get/edit/create pipeline within an existing group. If the group does not exist, it implies that the operate user is not a Pipeline group admin for a non-existent group and thus, should not be allowed to create a pipeline.

##### Actual Results
When using the Pipeline Config API as a Pipeline Group Admin of an existing group, access is denied to get/edit/create a pipeline within that group. As a Go Admin, when the Pipeline Config API is used to create a pipeline within a non-existent group, the pipeline group and the pipeline gets created. 
  
##### Possible Fix
2. Make Pipeline Config API accessible to Pipeline Group Admins and Go Admins. Disallow creation of a group from this API. Have a separate Group API, accessible only to Go Admins, to handle the creation of pipeline groups and to assign the user and role permissions. 

Notes: 
1.Disallow Pipeline Group Admin of another group should not have access to create/edit/get pipelines from a group that the user is not an admin of.
2. Allow users with view permissions to a pipeline group to use the GET request for Pipeline Group API. Pipeline group Admins can view, create or edit a pipeline.

Aaronov:
I am running 16.4 and having similar issues. I can't GET a pipeline config (https://api.go.cd/current/#get-pipeline-config) and instead just get a 404 no matter what user I use. I don't know if this is the same problem or not but this is a roadblock for our team.

zabil:
@Aaronov this issue is not related.

I suggest you mail the user groups with the problem you are facing and sharing more info about your configuration.

It will also help if you send a sample curl request (with sensitive info removed)

Aaronov:
Sorry I may have jumped the gun. It appears to have started to work now so I'm not really sure what was happening before. I'll report back in a new ticket if I see the problem again and provide further details.

GaneshSPatil:
[...]

PR #2125, opened 2016-04-20
#2119 Pipeline Config API does not allow Pipeline Group Admin to create/edit/get a pipeline

GaneshSPatil:
made the pipeline config api accessible to the pipeline group admin users and Go admin users

### DIFF
--- a/server/webapp/WEB-INF/applicationContext-acegi-security.xml
+++ b/server/webapp/WEB-INF/applicationContext-acegi-security.xml
@@ -264,6 +264,8 @@
                 /admin/package_definitions/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /api/admin/material_test=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /api/admin/internal/pipelines=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/pipelines=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/pipelines/*=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /api/admin/**=ROLE_SUPERVISOR
                 /api/config-repository.git/**=ROLE_SUPERVISOR
                 /api/jobs/scheduled.xml=ROLE_SUPERVISOR
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
@@ -54,7 +54,7 @@ module ApiV1
         end
       end
 
-      def handle_config_save_or_update_result(result, pipeline_name = params[:name])
+      def handle_config_save_or_update_result(result, pipeline_name = params[:pipeline_name])
         if result.isSuccessful
           load_pipeline(pipeline_name)
           json = ApiV1::Config::PipelineConfigRepresenter.new(@pipeline_config).to_hash(url_builder: self)
@@ -68,7 +68,7 @@ module ApiV1
 
 
       def check_for_attempted_pipeline_rename
-        unless CaseInsensitiveString.new(params[:pipeline][:name]) == CaseInsensitiveString.new(params[:name])
+        unless CaseInsensitiveString.new(params[:pipeline][:name]) == CaseInsensitiveString.new(params[:pipeline_name])
           result = HttpLocalizedOperationResult.new
           result.notAcceptable(LocalizedMessage::string("PIPELINE_RENAMING_NOT_ALLOWED"))
           render_http_operation_result(result)
@@ -76,14 +76,14 @@ module ApiV1
       end
 
       def check_for_stale_request
-        if (request.env["HTTP_IF_MATCH"] != "\"#{Digest::MD5.hexdigest(get_etag_for_pipeline_from_cache(params[:name]))}\"")
+        if (request.env["HTTP_IF_MATCH"] != "\"#{Digest::MD5.hexdigest(get_etag_for_pipeline_from_cache(params[:pipeline_name]))}\"")
           result = HttpLocalizedOperationResult.new
-          result.stale(LocalizedMessage::string("STALE_PIPELINE_CONFIG", params[:name]))
+          result.stale(LocalizedMessage::string("STALE_PIPELINE_CONFIG", params[:pipeline_name]))
           render_http_operation_result(result)
         end
       end
 
-      def load_pipeline(pipeline_name = params[:name])
+      def load_pipeline(pipeline_name = params[:pipeline_name])
         @pipeline_config = pipeline_config_service.getPipelineConfig(pipeline_name)
         raise RecordNotFound if @pipeline_config.nil?
       end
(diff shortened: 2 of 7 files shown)


server/webapp/WEB-INF/rails.new/app/helpers/api_v1/authentication_helper.rb
```diff
@@ -48,6 +48,15 @@ def check_admin_user_and_401
       end
     end
 
+    def check_pipeline_group_admin_user_and_401
+      groupName = params[:group] || go_config_service.findGroupNameByPipeline(com.thoughtworks.go.config.CaseInsensitiveString.new(params[:name]))
+      return unless security_service.isSecurityEnabled()
+      unless security_service.isUserAdminOfGroup(current_user.getUsername, groupName)
```
varshavaradarajan: Wouldn't this raise a null pointer exception if the group does not exist? Check out GoConfigService's function isUserAdminOfGroup()

GaneshSPatil: @varshavaradarajan 
Yes..for an invalid pipeline name
Are there any other cases will it throw null pointer error?

server/webapp/WEB-INF/rails.new/app/helpers/api_v1/authentication_helper.rb
```diff
@@ -74,5 +83,15 @@ def render_unauthorized_error
       render :json_hal_v1 => { :message => 'You are not authorized to perform this action.' }, :status => 401
     end
 
+    private
+    def is_user_a_group_admin?(current_user, group_name)
```
jyotisingh: Most of this logic including the check for `security_service.isSecurityEnabled` is already taken care of as part of [SecurityService.isUserAdminOfGroup](https://github.com/gocd/gocd/blob/824f940a5163d9c90e6e34747f87baf3e6f7f84c/server/src/com/thoughtworks/go/server/service/SecurityService.java#L69). 

`goConfigService.groups().hasGroup(groupName) && !securityService.isUserAdminOfGroup(currentUser.getUsername(), groupName)` should be good enough check

GaneshSPatil:
@jyotisingh 
I can not use goConfigService.isUserAdminOfGroup because it will throw `group not found exception` when group is not present.But in such case we want to check whether the user is Admin and if so, then create the group.

jyotisingh:
thats why I asked you to check for `goConfigService.groups().hasGroup(groupName)` first

server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -260,6 +260,8 @@
                 /admin/materials/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /admin/package_repositories/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /admin/package_definitions/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/pipelines=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/pipelines/*=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
```
ketan: @jyotisingh — I don't understand this DSL very much, so I'm not sure what the role would be in these cases —

```
GET .../pipelines - list pipelines, requires ROLE_USER (for future, we don't have a list pipelines call yet)
POST .../pipelines - create pipelines, requires ROLE_GROUP_SUPERVISOR, if the group already exists. requires ROLE_SUPERVISOR if the group does not exist
DELETE|POST|PUT .../pipelines/foo - requires atleast ROLE_GROUP_SUPERVISOR
```

ketan:
@GaneshSPatil — can you rebase and resolve conflicts?

server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -261,7 +261,7 @@
                 /admin/package_repositories/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /admin/package_definitions/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /api/admin/pipelines=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
-                /api/admin/pipelines/*=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/pipelines/*=ROLE_USER
```
jyotisingh: Thats incorrect. This will allow all users to be able to access the pipeline config. It needs to be `ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
```diff
@@ -17,7 +17,8 @@
 module ApiV1
   module Admin
     class PipelinesController < ApiV1::BaseController
-      before_action :check_pipeline_group_admin_user_and_401
+      before_action :check_pipeline_group_admin_user_and_401, except: [:show]
```
jyotisingh: I think you are confused between pipeline instance that we display for view users and pipeline config. No one apart from pipeline-group-admin and super-admin are allowed to view/edit/create pipelines. However, based on the configuration, some underprivileged users may have view/operate access to pipeline instances. They are two totally different things. What you originally had ie. `before_action :check_pipeline_group_admin_user_and_401` was the right thing to do. The same goes for the other changes you have made for this to work. Please fix.

server/webapp/WEB-INF/applicationContext-acegi-security.xml
jyotisingh: All the above methods are applicable to ROLE_SUPERVISOR and ROLE_GROUP_SUPERVISOR alone. ROLE_USER does not have any permissions as far as any config in Go is concerned. They may/may-not get view/operate permissions for pipeline-runs on dashboard, but never an access to the pipeline config itself. 

Having said that, the weird part of this config that comes into picture is for `POST .../pipelines` which ends up creating a group when it does not exist. Creation of a new group should be allowed for `ROLE_SUPERVISOR` only. I don't know if there is a decent way to configure that in `applicationContext-acegi-security.xml`. There is `PreAuthorize` tag which seem to do something that we would want in this case, but I haven't given it a try myself.

server/webapp/WEB-INF/rails.new/app/helpers/api_v1/authentication_helper.rb
```diff
@@ -56,22 +56,41 @@ def check_admin_user_or_group_admin_user_and_401
...
+      unless is_user_a_group_admin?(current_user, groupName)
+        Rails.logger.info("User '#{current_user.getUsername}' attempted to perform an unauthorized action!")
+        render_unauthorized_error
+      end
+    end
+
     def verify_content_type_on_post
       if [:put, :post, :patch].include?(request.request_method_symbol) && !request.raw_post.blank? && request.content_mime_type != :json
-        render ApiV1::BaseController::DEFAULT_FORMAT => { message: "You must specify a 'Content-Type' of 'application/json'" }, status: :unsupported_media_type
+        render ApiV1::BaseController::DEFAULT_FORMAT => {message: "You must specify a 'Content-Type' of 'application/json'"}, status: :unsupported_media_type
```
jyotisingh: this line should probably go away, right?

Issue #1924, opened 2016-02-11
Weak Fan-In Issues


FredrikWendt:
GoCD has great support for proper fan-out and fan-in - it's a unique selling point compared to other build automation tools. The Value Stream Map is a second one.

Both these features become flawed with manually triggered pipeline runs.

The actual scheduling of work, when caused by manually triggering a pipeline, will not honor the fan-in principles/framework built into the product. It will instead violate all such rules and pick whatever is latest from the configured materials.
There's no way to manually trigger scheduling of a pipeline which will respect the fan-in feature.

The Value Stream Map is limited to showing only one changeset from a source repo, although an UpstreamPipeline might've run with changeset r1, and a DownstramPipeline run with changeset r2. When the ValueStreamMap displays with focus set to the DownstreamPipeline run, instead of showing that UpstreamPipline pointing to r1 and DownstreamPipeline to r2, both runs point to r2 (which is incorrect for UpstreamPipeline).
The Value Stream Map feature also doesn't differentiate, visually, between pipeline runs which were scheduled honoring the fan-in rules, from those which were run with arbitrarily picked material (ie ignore fan-in).

Combined, this makes the two biggest selling points for GoCD voided.

FredrikWendt:
Some solutions are outlined in https://www.go.cd/documentation/developer/5/5.2.html#5-2-5-a-name-extensions-a-extensions where manual triggers could be checked to respect fan-in.

chrislovsund:
+1

xli:
+1 about manual trigger. Could cause big mistake when deploying to production.

arvindsv:
@maheshp is going to take a look at showing manual triggered pipelines on VSM. Making manual trigger respect fan-in is going to be a larger conversation and needs more thought.

xli:
@arvindsv, @wpc and I had discussion about manual triggering expectation with fan-in, we think it biggest problem is confusion and misunderstanding, maybe disable, or with a big warning (like delete repo in github) may help. However, I can't find out any reason I want to trigger a build with the following secanio:
   
      Codebase A is material of pipeline p1 and p2, p1 is also material of p2.
      A's v1 triggered p1 and then following p2.
      A's v2 triggered p1, but failed at p1, hence p2 won't trigger.

      Manual trigger p2, this will cause p2 pick up p1's v1 result, and A's v2.

So we think maybe the best option is disallow manual trigger without specifying version.

FredrikWendt:
[...]

PR #2136, opened 2016-04-26
Warnings on VSM if pipeline built with incompatible revisions

maheshp:
* Partial fix for #1924
* VSM for a pipeline shows warning message if built with incompatible revisions
* SCM Nodes list MaterialRevisions instead modifications.

### DIFF
--- a/server/src/com/thoughtworks/go/domain/valuestreammap/SCMDependencyNode.java
+++ b/server/src/com/thoughtworks/go/domain/valuestreammap/SCMDependencyNode.java
@@ -16,13 +16,23 @@
 
 package com.thoughtworks.go.domain.valuestreammap;
 
+import com.thoughtworks.go.domain.MaterialRevision;
+import com.thoughtworks.go.domain.materials.Material;
+import com.thoughtworks.go.domain.materials.Modification;
+import com.thoughtworks.go.domain.materials.Modifications;
+
 import java.util.*;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
 
 
 public class SCMDependencyNode extends Node {
     private Set<Revision> revisions = new HashSet<>();
     private String materialType;
     private HashSet<String> materialNames = new LinkedHashSet<>();
+    private Set<MaterialRevisionWrapper> materialRevisionWrappers = new HashSet<>();
+    private List<MaterialRevision> materialRevisions = new ArrayList<>();
 
     public SCMDependencyNode(String nodeId, String nodeName, String materialType) {
         super(DependencyNodeType.MATERIAL, nodeId, nodeName);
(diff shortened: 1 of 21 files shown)


srinivasupadhya:
Will this work in the below case:
```
    /-> p1 -- \
git            -> p3 -> p4
  \ \-> p2 -- /      /
   \----------------/
```

current pipeline is `p3`. `p1` built off of `g1`, `p2` built off of `g1`. `p4` built off of `g2`.

jyotisingh:
@srinivasupadhya - what's the build cause for P3 here? Did you mean to say that current pipeline was p4? If current pipeline is p3, then the build-cause for p4 will not affect P3 in anyway!

srinivasupadhya:
@jyotisingh yes. it wont. but if im not wrong we will show the warning on `p3` since material revisions for git will be `g1` & `g2`

jyotisingh:
if the current pipeline is p3, then g1 will be the only revision for git. P4 being a downstream, will not cause revisions to get added to the scm nodes, ie. git in this case. Thats how it has been all this while and I do not see any of that changed in this PR. 
Is there some code in specific that you think could potentially cause issue? Can you point it out.

srinivasupadhya:
ok. i forgot we populate SCM data only for upstream pipelines. ignore :+1:

arvindsv:
> ok. i forgot we populate SCM data only for upstream pipelines. ignore :+1:

*shocked* :)

arvindsv:
@maheshp: Adding a little screenshot to this might help everyone get an idea about how it looks.

arvindsv:
Also, this is related to https://github.com/gocd/gocd/issues/1924. Commenting so that GitHub links the two together.

maheshp:
screen shot 2016-04-27 at 9 58 29 am

In case of a manual trigger of a pipeline with incompatible revisions a warning on the current pipeline is shown and the SCM material lists all the revisions responsible for the build

maheshp:
screen shot 2016-04-27 at 9 58 29 am

If the current pipeline is built from incompatible revisions a warning is shown and the SCM node lists all the revisions responsible for the build.

jyotisingh:
@FredrikWendt - thought you might be interested in this

FredrikWendt:
YAY!  \o/

I assume this only looks at SCM revisions, and not the case where two different run instances of any pipeline?
```
    /-> p1 -- \
p0            -> p3 -> p4
  \ \-> p2 -- /      /
   \----------------/
```
If p1 was run with p0:1 and p2 with p0:2, would p3 be red flagged?

maheshp:
@FredrikWendt 
> If p1 was run with p0:1 and p2 with p0:2, would p3 be red flagged?

Yes p3 would be red flagged if p3 is triggered with p1(p0:1) and p2(p0:2)

maheshp:
@FredrikWendt 
> I assume this only looks at SCM revisions, and not the case where two different run instances of any pipeline?

Adding some context about the implementation. 

- The build cause for any pipeline run boils down to a set of SCM modifications(revisions), this can be represented as MaterialRevision(basically a Material with one or more modifications). 

- While building a VSM for a pipeline we traverse through all the upstream pipelines and list all the MaterialRevisions responsible for the run. If the MaterialRevisions differ we flag the pipeline and list the MaterialRevisions as part of the SCM node.

- In case of manual trigger there could be instances where the MaterialRevisions differ.

PR #2137, opened 2016-04-26
1133 config repo caching and two step validation

jyotisingh:
* Changes in main.xml should not break config-repo partials. handling cache issues etc.

* Adding groups and environments added by partials to the main xml so they could be edited to add authorisation, removing unused methods, and some minor cleanup

* Validating merged config before caching partials.

* Introducing two step validation and save when config-repos are configured

* Handling two step config validations when a config-repo partial changes

* There is just one CachedGoConfig, removed the interface and inlined all calls to cachedFileGoConfig from mergedGoConfig(now called CachedGoConfig). 

**Please merge https://github.com/gocd/functional-tests/pull/74 when you merge this otherwise regression tests would fail.**


config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -358,12 +360,14 @@ public void updateGroup(PipelineConfigs pipelineConfigs, String groupName) {
...
         @Override
         public ConfigOrigin getOrigin() {
-            MergeConfigOrigin origins = new MergeConfigOrigin(this.main.getOrigin());
-            for(PartialConfig part : this.parts)
-            {
-                origins.add(part.getOrigin());
-            }
-            return origins;
+            return new MergeConfigOrigin();
+            //TODO: jyoti, check if we still need this
```
tomzo: Probably we don't need this. IIRC there is no code that actually reads those partial origins from MergeConfigOrigin instances. 
However I added and implemented `MergeConfigOrigin` to help with debugging, after all it does give you some idea of what are the sources of parts in merged instance.
You can remove it if you like, `return new MergeConfigOrigin();` should work fine. There is a failing test because of this `shouldReturnOriginAsASumOfAllOrigins`

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -1367,6 +1366,16 @@ public boolean canDeletePluggableSCMMaterial(SCM scmConfig) {
     }
 
     @Override
+    public void setPartials(List<PartialConfig> partials) {
+        this.partials = new Cloner().deepClone(partials);
```
tomzo: Partials are immutable, why clone?

server/src/com/thoughtworks/go/config/GoPartialConfig.java
```diff
@@ -91,32 +76,89 @@ public void onFailedPartialConfig(ConfigRepoConfig repoConfig, Exception ex) {
...
+                @Override
+                public CruiseConfig update(CruiseConfig cruiseConfig) throws Exception {
+                    if (newPart != null && fingerprint != null) {
+                        cruiseConfig.getPartials().remove(findMatchingPartial(cruiseConfig, fingerprint));
+                        cruiseConfig.getPartials().add(new Cloner().deepClone(newPart));
+
+                        for (PartialConfig partial : cruiseConfig.getPartials()) {
+                            for (EnvironmentConfig environmentConfig : partial.getEnvironments()) {
+                                if (!cruiseConfig.getEnvironments().hasEnvironmentNamed(environmentConfig.name())) {
+                                    cruiseConfig.addEnvironment(new BasicEnvironmentConfig(environmentConfig.name()));
```
tomzo: Just to be clear - why are you adding this named-only environment config to main configuration? 
Is it to ensure that it will show up on config xml tab and on API endpoint? (I recall you mentioned it once, but cannot find where that was.)

server/src/com/thoughtworks/go/config/GoPartialConfig.java
```diff
@@ -91,32 +76,89 @@ public void onFailedPartialConfig(ConfigRepoConfig repoConfig, Exception ex) {
...
+                                }
+                            }
+                            for (PipelineConfigs pipelineConfigs : partial.getGroups()) {
+                                if (!cruiseConfig.getGroups().hasGroup(pipelineConfigs.getGroup())) {
+                                    cruiseConfig.getGroups().add(new BasicPipelineConfigs(pipelineConfigs.getGroup(), new Authorization()));
```
tomzo: I can see this is the same as above

tomzo:
I tried executing `shouldValidateConfigRepoLastKnownPartialsWithMainConfigAndUpdateConfigToIncludePipelinesFromPartials` but it fails. I found out that [here](https://github.com/jyotisingh/gocd/blob/1133_config_repo_caching_and_two_step_validation/server/src/com/thoughtworks/go/config/CachedGoPartials.java#L82) `fingerprintToLatestKnownConfigMap.get(fingerprint)` returns null because fingerprint is does not match to one stored in `fingerprintToLatestKnownConfigMap`. I did not debug further.

server/test/unit/com/thoughtworks/go/config/CachedGoConfigTest.java
```diff
@@ -19,62 +19,235 @@
...
+        when(dataSource.writePipelineWithLock(pipelineConfig, holderBeforeUpdate, saveCommand, user)).thenReturn(saveResult);
+
+        cachedGoConfig.writePipelineWithLock(pipelineConfig, saveCommand, user);
+        assertThat(cachedGoConfig.loadConfigHolder(), is(savedConfig));
+        assertThat(cachedGoConfig.currentConfig(), is(savedConfig.config));
+        assertThat(cachedGoConfig.loadForEditing(), is(savedConfig.configForEdit));
+        verify(dataSource).writePipelineWithLock(pipelineConfig, holderBeforeUpdate, saveCommand, user);
+    }
+
     protected CachedGoConfig cachedGoConfig;
```
tomzo: `cachedGoConfig` is always null here. 
Also it would be nice to put these fields higher.

server/src/com/thoughtworks/go/config/CachedGoPartials.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+
+    public void addOrUpdate(String fingerprint, PartialConfig newPart) {
+        fingerprintToLatestKnownConfigMap.put(fingerprint, new UpdatedPartial(newPart, new DateTime()));
+    }
+
+    public void markAsValid(String fingerprint, PartialConfig newPart) {
+        DateTime lastUpdated = fingerprintToLatestKnownConfigMap.get(fingerprint).lastUpdated;
```
tomzo: Considering that this is currently throwing NullPointerException maybe it should throw some nicer message when `markAsValid` is called with a fingerprint that does not exist in map

tomzo:
@jyotisingh I've read the whole thing. Except for few tests which are failing, I think this is good and I understand most of the changes.
The jvm crash you had is indeed not caused by your changes. It is something I had messed up earlier and apparently it is on master. Here is a short fix https://github.com/tomzo/gocd/commit/063fbbad853ea6822f680b73052ac96ff1dfe7f4
I have it fixed in #1810 which I will rebase on top of this PR, once this is ready.
I hope you'll find my comments useful.

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
jyotisingh: Partials contain other config objects such as  pipelineConfig internally which are merged with main config and validated. Validation errors (if any) will get added to the pipelineConfig object, which is later displayed in server-health messages. Need to clone so that the validation errors are not added to the cached-known-partials.

server/src/com/thoughtworks/go/config/GoPartialConfig.java
jyotisingh: Yes, the idea was to add such environments and groups to main xml and save so that it will show up on configXml tab as well as pipelines tab for edit, thus avoiding us to have something like `UIConfigOrigin`.

tomzo:
@jyotisingh  I've been looking at your code closely as I am rebasing UI changes.
I am having trouble to understand this:
In `BasicCruiseConfig` there is field `partials`
```java
    @IgnoreTraversal
    private List<PartialConfig> partials = new ArrayList<>();
```
Which can be assigned only by calling `setPartials`. But setting this `partials` does not actually change behaviour of `BasicCruiseConfig` instance. There is only `getPartials`.
There is also `merge` method, which effectively changes what is considered to be current pipelines and environments. It does that by replacing `group` and `environments`, which are created by `MergeStrategy` when `merge` is called.
I really don't get what `partials` is for? Especially that  `MergeStrategy` also has a field `parts`. Obviously these do not have the same content. Could elaborate why is it like that?

Consider such example - I can call `merge` to make a merged config, but then when I call `getPartials` that will return empty list. Was that intended?

jyotisingh:
You are right that `setPartials` is not the same as `merge`. `setPartials` started out as a temporary way to pass along the partials which would eventually be *merged* with the main config. If you look at the usages of `setPartials`, you will see that the value is set to `cachedGoPartials.lastKnownPartials()` or `cachedGoPartials.lastValidPartials()` when saving pipeline in main xml. When a partial config is modified, during the full validation invoked thereafter, the partials passed along to `setPartials` includes the changed. So, the `partials` on BasicCruiseConfig is a list of partials that would be merged at a later point by the preprocessor. I should probably use a better mechanism to set these partials, but this is what I have as of now. 
Does that make any sense?

tomzo:
Thank you, that makes sense. I would only suggest to change how they are named. I mean  `getPartials` is confusing, because does not say which partials we are obtaining. I may need to have `getMergedPartials` method, then we could rename `getPartials` to `getCandidatePartials` perhaps?
I would do it in #1810 if you agree.

jyotisingh:
`getPartials` would be set only for a brief period. The only usage of that method would be from the preprocessor. But sure, feel free to rename it to something more appropriate. However, I wonder what `getMergedPartials ` is going to be. The pipelines on BasicCruiseConfig would contain all pipelines (main + remote). Would this be a filtered list?

tomzo:
I had some bugs in #1810 which I had to fix, while doing so I found ignored test that you planned to do:
```
  // TODO: jyoti
    @Ignore("rewrite this test as integration test")
    @Test
    public void shouldFailWhenTryingToAddPipelineDefinedRemotely() throws Exception {
```
I implemented it in #1810 - https://github.com/gocd/gocd/pull/1810/commits/baa8d095b56865e6857a1c7c4295b1e0c7419601 together with a bunch of similar cases.

jyotisingh:
@tomzo - you are awesome! Thanks :)

PR #2147, opened 2016-04-29
Git materials can be shallow cloned on SPA #1724

maheshp:


ketan:
What about the controller and API?

Issue #2110, opened 2016-04-15
Automatic build in loop after rebase of a GIT branch


jguilloux71:
##### Issue Type
- Bug Report

##### Summary
This issue is appeared after an update from 15.2 to 16.3.0.

On a pipeline, with a GIT material, for a dedicated branch, if a rebase has been executed (git rebase -i other-branch) on this branch, then the Go server build one time on the new commit and then one time on the old commit (before rebase). And the server does that in loop!

It detects modifications for material but there is no modification!

##### Environment
Go Server is installed on a Debian Jessie, in version 16.3.0.
With the version 15.2, no issue.
Builds are triggered manually.

###### Basic environment details
* Go Version: 16.3.0(3183-60d6fd31f00527da00f2affdd8f7f7732e7be103)
* JVM version: 1.8.0_77
* OS Information: Linux 3.2.0-4-amd64*
Browser vendor and version (if relevant): all browser (IE, Chrome, Firefox, Safari)

jguilloux71:
Up. Go CD is unavailable with this issue :-(

zabil:
A git bisect tracked the problem to this commit (3981000187df0d7693226e7efbfdafc3663e726f)

zabil:
This line
https://github.com/gocd/gocd/commit/3981000187df0d7693226e7efbfdafc3663e726f#diff-97dd991971202dc0de417dbb4271aa82R101

Forms a wrong git log command 
For e.g 
Let's say you make a commit with the sha 1
Our database registers that as the latest commit executes the build and everything's good.
The git log to check for modification ends up as follows
```
git log origin/master 1..
```

Which return nothing.
However when you rebase you get a new sha 2 which executes the build. 
Now when it checks for modifications it executes

```
git log origin/master 2..
```
which returns 1 and then the loop starts

Next time it does

```
git log origin/master 1..
```
which returns 2

The fix is to use the right log command i.e

```
git log 2..origin/master
```

PR #2149, opened 2016-04-29
#2110 Changing the git log command to fix the build loop after rebase.

varshavaradarajan:


varshavaradarajan:
#2110

arvindsv:
Umm. Still ... no test? :) No, no, no. Reproduce this in a test, please.

arvindsv:
I'm sure you probably tried and decided not to. :) Must be some reason not to.

zabil:
This is to show progress, test will follow.

arvindsv:
Yay!

varshavaradarajan:
Wrote the test. @zabil Please review.

common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
```diff
@@ -99,9 +101,18 @@ public Modification latestModification() {
             throw new RuntimeException(String.format("Working directory: %s\n%s", workingDir, outputStreamConsumer.getStdError()), e);
         }
 
-        CommandLine gitCmd = git(environment).withArg("log").withArg(remoteBranch()).withArgs(args).withWorkingDir(workingDir);
+        CommandLine gitCmd = git(environment).withArg("log").withArgs(args).withWorkingDir(workingDir);
+        ConsoleResult result;
+        try {
+            result = runOrBomb(gitCmd);
+        } catch (CommandLineException exception) {
+            if (exception.getMessage().contains("Invalid revision range")) {
```
ketan: I'd not assume this message, it's prone to change in the future versions of git (and may be different in previous versions).

This message may also be different based on the locale and language settings that's being used and git may print this message in French or Russion for all you know :)

common/src/com/thoughtworks/go/config/materials/SubprocessExecutionContext.java
```diff
@@ -24,4 +24,5 @@
 public interface SubprocessExecutionContext {
     String getProcessNamespace(String fingerprint);
     Map<String, String> getDefaultEnvironmentVariables();
+    Boolean getGitShallowCloneFlag();
```
zabil: rename this to ```isShallowClone```

common/src/com/thoughtworks/go/domain/materials/AgentSubprocessExecutionContext.java
```diff
@@ -40,4 +40,9 @@ public String getProcessNamespace(String fingerprint) {
     public Map<String, String> getDefaultEnvironmentVariables() {
         return new HashMap<>();
     }
+
+    @Override
+    public Boolean getGitShallowCloneFlag() {
+        throw new UnsupportedOperationException();
```
zabil: Why is this returning unsupported?

ketan: Because agents SHOULD NOT be calling this method, and if they are, there's something really-really wrong. We perhaps need a better exception class, or a better error message `throw new RuntimeException("Really useful message")`?

zabil: ```UnsupportedOperationException(String message)``` is there.

PR #2151, opened 2016-05-02
Parse `variables` and `secure_variables` from request so it would be possible to schedule pipeline with custom variables.

grundic:
Parameters in request are plain text, so they must be parsed to hash in
order instance of `server/src/com/thoughtworks/go/server/scheduling/ScheduleOptions.java` would accept them. I took a look at tests at `server/webapp/WEB-INF/rails.new/spec/controllers/api/pipelines_controller_spec.rb` -- they work because you pass Hashes to tests.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api/pipelines_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api/pipelines_controller.rb
@@ -14,6 +14,8 @@
 # limitations under the License.
 ##########################GO-LICENSE-END##################################
 
+require 'json'
+
 class Api::PipelinesController < Api::ApiController
   include ComparisonHelper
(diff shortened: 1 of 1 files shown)


ketan:
I like being able to submit environment variables, because this represents trigger with options in a manner that's much closer. However having most of the payload being form post, and the rest being JSON seems like an oddball. Do you feel it's better to model this properly using a nicer (and well thought out) API?

ketan:
Having said that, since it's all rails, you can submit the entire payload as JSON without this patch —

```bash
curl https://ci.example.com/go/api/pipelines/:pipeline_name/schedule \
    -u user:password \
    -XPOST \
    -H 'Confirm: true' \
    -d '{
           "variables": ...,
           "secure_variables": ...,
        }'
```

grundic:
@ketan, you are right.
Was able to schedule pipeline -- I was missing that content-type header.
Sorry for your time and thank you for making this clear.

Issue #2152, opened 2016-05-03
Checkout working copy only on the agent side.


varshavaradarajan:
##### Issue Type

- Feature proposal

##### Summary

When a material is listed, on the server side, the flyweight directory is created for the new material. The clone command for the same is as follows
```
git clone --branch=<branchName> --no-checkout --depth=<depth> <url> <workingDirectory>
```

Although the --no-checkout and shallow clone flag improve the performance of the go-server greatly, having the flyweight folder as a bare repository can reduce the disk space and make the clone of large repositories faster than it is right now.

On the server side, the clone command could be 

```
git clone --bare --depth=<depth> <url> <workingDirectory>
```

##### Notes

The bare repository must only be on the server-side because the agent needs a working copy to perform the build operations.

FredrikWendt:
+1

varshavaradarajan:
Bare repository is not a feasible option for Go right now because of the difference in behaviour for the git commands. It involves a breaking change for the non-bare repositories on the server side. Needs to be analyzed further.

PR #2156, opened 2016-05-04
[WIP] #2152: Cloning a bare repository on the server side

varshavaradarajan:
This change clone's the flyweight folder with a ```--bare``` flag instead of a ```--no-checkout```

### DIFF
--- a/common/src/com/thoughtworks/go/config/materials/SubprocessExecutionContext.java
+++ b/common/src/com/thoughtworks/go/config/materials/SubprocessExecutionContext.java
@@ -25,4 +25,5 @@ public interface SubprocessExecutionContext {
     String getProcessNamespace(String fingerprint);
     Map<String, String> getDefaultEnvironmentVariables();
     Boolean isGitShallowClone();
+    boolean isServer();
 }
--- a/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
@@ -23,7 +23,6 @@ import com.thoughtworks.go.domain.MaterialInstance;
 import com.thoughtworks.go.domain.materials.*;
 import com.thoughtworks.go.domain.materials.git.GitCommand;
 import com.thoughtworks.go.domain.materials.git.GitMaterialInstance;
-import com.thoughtworks.go.domain.materials.svn.MaterialUrl;
 import com.thoughtworks.go.server.transaction.TransactionSynchronizationManager;
 import com.thoughtworks.go.util.GoConstants;
 import com.thoughtworks.go.util.StringUtil;
@@ -37,6 +36,7 @@ import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationAdapter;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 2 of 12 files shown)


common/src/com/thoughtworks/go/config/materials/SubprocessExecutionContext.java
```diff
@@ -24,4 +24,5 @@
 public interface SubprocessExecutionContext {
     String getProcessNamespace(String fingerprint);
     Map<String, String> getDefaultEnvironmentVariables();
+    boolean bare();
```
zabil: Rename this to ```isBare```

varshavaradarajan:
Issues to be fixed: 
1. isGitRepository will always return false as the .git folder does not exist in the bare repository. So, the flyweight is deleted and recreated with the same fingerprint each time a MDU happens (shallow clone or not) 

2. isShallow will always return false because .git/shallow does not exist.

ketan:
Is this still priority? Can we have the conflicts resolved please?

common/src/com/thoughtworks/go/config/materials/SubprocessExecutionContext.java
zabil: I think we let this slip in the last commit by adding the ```isGItShallowClone``` but ```SubprocessExecutionContext ``` is something used by all materials. We must not pollute it with git related properties. 

For the bare implementation i think we must check if the execution context is on the server and checkout a bare repository on the git material instead of having this specific flag.

And this is a forward only change, the git log commands must work with non bare repositories, that way we don't have to mess with migrations.

common/test/unit/com/thoughtworks/go/config/materials/git/GitMaterialTest.java
```diff
@@ -161,11 +161,11 @@ public void shouldRetrieveLatestModificationIfRevisionIsNotFound() throws IOExce
...
-    @Test
-    public void shouldNotCheckingOutWorkingCopyUponCallingModificationsSinceARevision() throws Exception {
-        git.modificationsSince(workingDir, REVISION_0, new TestSubprocessExecutionContext());
-        assertWorkingCopyNotCheckedOut(workingDir);
-    }
+//    @Test
+//    public void shouldNotCheckingOutWorkingCopyUponCallingModificationsSinceARevision() throws Exception {
+//        git.modificationsSince(workingDir, REVISION_0, new TestSubprocessExecutionContext(true));
+//        assertWorkingCopyNotCheckedOut(workingDir);
+//    }
```
zabil: Remove this bit of code if it's gonna remain commented

common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
```diff
@@ -54,17 +54,29 @@ public GitCommand(String materialFingerprint, File workingDir, String branch, bo
         this.environment = environment;
     }
 
-    public int cloneFrom(ProcessOutputStreamConsumer outputStreamConsumer, String url) {
-        return cloneFrom(outputStreamConsumer, url, Integer.MAX_VALUE);
+    public int bareClone(ProcessOutputStreamConsumer outputStreamConsumer, String url, int cloneDepth) {
+         CommandLine bareClone = cloneCommand().withArg("--bare");
+        if(cloneDepth < Integer.MAX_VALUE) {
```
zabil: Don't support depths for bare clone. Remove the if condition

common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
```diff
@@ -209,18 +209,24 @@ private GitCommand getGit(File workingdir, int preferredCloneDepth, SubprocessEx
         }
     }
 
+    private boolean isRepository(GitCommand gitCommand, GitRepository gitRepository) {
```
zabil: Do not migrate non bare repositories as this will be dead code soon and be unpredictable for large repos.

PR #2157, opened 2016-05-05
Tasks on SPA can have on_cancel_task #1727

maheshp:
* Added on_cancel attribute to Task model
* Added onCancel widget
* TaskConfigWidget is split into multiple sub_components

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/helpers/dragula_config.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/helpers/dragula_config.js
@@ -14,21 +14,17 @@
  * limitations under the License.
  */
 
-define(['dragula'], function (dragula) {
-  var dragulaConfig = function (elem, isInitialized, onDropCallback) {
-    if (isInitialized) {
-      return;
-    }
+define(['dragula', 'lodash'], function (dragula, _) {
+  var dragulaConfig = function (elem, options) {
 
-    var drake = dragula([elem], {
-      revertOnSpill:   true,
-      mirrorContainer: elem
-    });
+    var opts = _.extend({revertOnSpill: true, mirrorContainer: elem}, options.dragulaOptions);
+
+    var drake = dragula([elem], opts);
 
     drake.on('drop', function () {
       m.startComputation();
       try {
-        onDropCallback();
+        options.onDropCallback();
       } finally {
         m.endComputation();
       }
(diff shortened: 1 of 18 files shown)


server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/helpers/dragula_config.js
```diff
@@ -22,7 +22,10 @@ define(['dragula'], function (dragula) {
 
     var drake = dragula([elem], {
       revertOnSpill:   true,
-      mirrorContainer: elem
+      mirrorContainer: elem,
+      invalid: function(el, h) {
```
ketan: Can we pass this as an option to `dragulaConfig`? Seems like the implementation is very specific to tasks

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/tasks.js
```diff
@@ -47,6 +59,14 @@ define(['mithril', 'lodash', 'string-plus', './model_mixins', './argument', './r
     this._attributesToJSON = function () {
       throw new Error("Subclass responsibility!");
     };
+
+    this.hasOnCancelTask = function () {
+      return ! _.isNil(this.onCancelTask);
```
ketan: Did not know about `isNil`, can you get rid of `isUndefinedOrNull` in `string-plus.js` (or change impl to `isUndefinedOrNull: _.isNil`)

ketan:
Few general observations and peeves I have with the runif options (mostly labelling changes to clarify intent) —

Run if:
* the job has passed (probably even check this by default, since it is the default)
* the job has failed
* always (don't know what's a better label)

Missing lookup for oncancel?

Can we improve the "On Cancel Task" label to "If the job is cancelled when this stage is running", or something similar.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/helpers/dragula_config.js
maheshp: You are right, I should have passed it, will do.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/pipeline_configs/models/tasks.js
maheshp: will get rid of ```isUndefinedOrNull``` for now

maheshp:
> Missing lookup for oncancel?

I know bout it, there are some issues with the UI, will do once Naveen is around

maheshp:
> Run if:
>
> the job has passed (probably even check this by default, since it is the default)
>the job has failed
>always (don't know what's a better label)

`the job has passed` seems a little misleading since the job is in progress. How bout this,


`Run If Job Status` or `Run If Current Job Status`,

- Passed
- Failed
- Any

ketan:
Sounds a lot better, lets go with what you have.

Issue #2145, opened 2016-04-28
Agent goes out of memory when trying to build job with a lot of revisions


zabil:
##### Issue Type
- Bug Report

##### Summary

Having a big [changeset](https://github.com/ketan/gocd/commit/60e320b030f6c0a783843f92f65073160ad35e1f) makes the agent go out of memory because we pass serialize all the revisions.

This is because [BuildWork](https://github.com/gocd/gocd/blob/2c1c83c7946963a81657d8b28b9a3a894d001ee8/common/src/com/thoughtworks/go/remote/work/BuildWork.java#L82) is serialized along with revisions causing the agent to go out of memory when it deserializes the object.

###### Basic environment details

* Go Version: 15.1+
* JAVA Version: 1.7+
* OS: Cent OS 6.7

##### Steps to Reproduce

1. Make a [huge](https://github.com/ketan/gocd/commit/60e320b030f6c0a783843f92f65073160ad35e1f) commit 
2. Trigger a build with that commit
3. Start an agent with defaults to execute this job

##### Expected Results
Agent should execute the job without going out of memory

##### Actual Results
Agent goes out of memory and the jobs gets continuously reassigned.

##### Possible Fix
Modify [BuildWork](https://github.com/gocd/gocd/blob/2c1c83c7946963a81657d8b28b9a3a894d001ee8/common/src/com/thoughtworks/go/remote/work/BuildWork.java#L82)
To not pass the revisions and pass only whats required by the agent

##### Log snippets
```
org.springframework.remoting.RemoteAccessException: Could not access HTTP invoker remote service at [https://192.168.117.26:8154/go/remoting/remoteBuildRepository]; nested exception is java.lang.OutOfMemoryError: Java heap space
    at org.springframework.remoting.httpinvoker.HttpInvokerClientInterceptor.convertHttpInvokerAccessException(HttpInvokerClientInterceptor.java:212)
    at org.springframework.remoting.httpinvoker.HttpInvokerClientInterceptor.invoke(HttpInvokerClientInterceptor.java:145)
    at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
    at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:202)
    at com.sun.proxy.$Proxy5.getWork(Unknown Source)
    at com.thoughtworks.go.agent.AgentController.retrieveWork(AgentController.java:216)
[...]

PR #2158, opened 2016-05-05
#2145 - BuildCause in not kept in memory as a part of workAssignments, moreover it is not serialized over to the agents

jyotisingh:
BuildCause can be a huge object when there are too many modifications involved or even when a small number of modifications have a huge number of modifiedFiles in them. We seem to be loading up this object in memory for each job. When a given pipeline instance has a huge number of modifications/modifiedFiles, and the scheduled stage has too many jobs, this causes an out of memory of the agent. As the size of these modifiedFiles increase, it also tends to bring down the server. We do not need modifiedFiles to be passed along to the agents. We certainly don't need all of buildcause on the agent or in server memory as part of work-assignments, so removing that off.

### DIFF
--- a/common/src/com/thoughtworks/go/remote/work/BuildAssignment.java
+++ b/common/src/com/thoughtworks/go/remote/work/BuildAssignment.java
@@ -1,47 +1,63 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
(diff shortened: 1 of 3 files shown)


common/src/com/thoughtworks/go/remote/work/BuildAssignment.java
```diff
@@ -1,47 +1,65 @@
...
-        this.buildCause = buildCause;
         this.buildWorkingDirectory = buildWorkingDirectory;
         this.builders = builder;
         this.plan = plan;
+        for (MaterialRevision materialRevision : buildCause.getMaterialRevisions()) {
+            ArrayList<Modification> modifications = new ArrayList<>();
+            for (Modification modification : materialRevision.getModifications()) {
+                Modification modificationForAgent = new Modification(modification);
+                modificationForAgent.unsetModifiedFiles();
+                modifications.add(modificationForAgent);
```
zabil: Move this to a copy constructor with a flag that constructs Modification object without copying the files so that you can get rid of the ```unsetModifiedFiles``` method.

common/src/com/thoughtworks/go/remote/work/BuildAssignment.java
```diff
@@ -1,47 +1,65 @@
...
     private final EnvironmentVariableContext initialContext = new EnvironmentVariableContext();
+    private final MaterialRevisions materialRevisions = new MaterialRevisions();
+    private final String approver;
 
     private BuildAssignment(BuildCause buildCause, File buildWorkingDirectory, List<Builder> builder, JobPlan plan) {
-        this.buildCause = buildCause;
         this.buildWorkingDirectory = buildWorkingDirectory;
         this.builders = builder;
         this.plan = plan;
+        for (MaterialRevision materialRevision : buildCause.getMaterialRevisions()) {
```
zabil: Any reason why we are not using the ```Modifications``` class?

PR #2159, opened 2016-05-06
Preventing duplicate modifications from getting inserted to database

jyotisingh:
When the scm history is re-written or a commit is merged without rebase, the scm-pollers do return duplicate revisions. These revisions are blindly inserted to db as new modification, even though these revisions were already part of db. This causes pipelines to trigger with the revisions that they had already triggered with and a whole lot of other confusion. This commit causes the duplicates to be ignored while inserting the modifications to db. If all revisions in a given MDU cycle for a material happen to be duplicate, then an exception is thrown to indicate the same - This would normally happen when the poller incorrectly keeps returning the same set of modifications all the time.

Adding a db constraint for materialid + revision makes sense, but it would require data correction which could get very complicated.

This prevents a bug in any given material-poller from screwing up the db with numerous duplicate modifications, and thereafter causing issues for everything else that relies on modifications data from db, ie. scheduling etc

### DIFF
--- a/common/test/unit/com/thoughtworks/go/helper/ModificationsMother.java
+++ b/common/test/unit/com/thoughtworks/go/helper/ModificationsMother.java
@@ -16,11 +16,7 @@
 
 package com.thoughtworks.go.helper;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.HashMap;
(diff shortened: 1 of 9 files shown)


arvindsv:
Since you mentioned looking at it, @zabil, I've assigned this to you.

PR #2175, opened 2016-05-11
solved the flash of unstyled content problem on single page app

GaneshSPatil:


server/webapp/WEB-INF/rails.new/app/views/layouts/pipeline_configs.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html class="no-js">
```
zabil: Is this supposed to be added here?

This is not valid HTML according to 
https://www.w3.org/TR/html4/struct/global.html#h-7.3

ketan: Although strictly not valid HTML, it's a very common pattern. The idea is to mark the element as `no-js` to ensure that elements that needs JS to behave and render (in this case the top nav bar), when the JS is [evaluated](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/foundation-rails-6.1.2.0/vendor/assets/js/foundation.core.js#L252-L260), it typically removes the `no-js`  class.

PR #2177, opened 2016-05-12
Add indices to improve performance of queries that run in the background.

ketan:
* the index on BST is needed when listing builds that an agent has run.
* the index on artifactplans is needed because it is frequently called via
  the "scheduledPlan" from `BuildAssignmentService` on every timer tick.

### DIFF
--- a/server/db/migrate/h2deltas/1606001_add_index_to_buildstatetransitions.sql
+++ b/server/db/migrate/h2deltas/1606001_add_index_to_buildstatetransitions.sql
@@ -1,0 +1,21 @@
+--
+-- Copyright 2016 ThoughtWorks, Inc.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--    http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


arvindsv:
Be *very, very* careful about this one. buildstatetransitions is a huge table usually, on big instances. This will need to be perf tested almost immediately.

zabil:
Hmm, did you mean while applying the constraint?

arvindsv:
Yes, just during the server upgrade (dbdeploy). Two things I'm worried about:

1. Big DB with lots of instances: Amount of time taken to upgrade, because of the indexing.
    Typically, when indexing takes a lot of time, dbdeploy does not give you any indication what it's doing. So, if someone stops or kills the server during this time, it'll kill that and the index will not be done.

2. Size of the DB before and after the index: Not too worried. Just curious about it.

arvindsv:
If it takes a lot of time during an upgrade, people tend to get a little nervous.

arvindsv:
@ketan @zabil: Did you get anywhere with this? Also, is this related to #8? If so, we can close that with this.

ketan:
> @ketan @zabil: Did you get anywhere with this?

No haven't started on it.

> Also, is this related to #8? If so, we can close that with this.

Yes, closing #8.

varshavaradarajan:
@arvindsv 
Size of the db before and after the index is the same. It probably depends on the size of the db. 
The time take for the indexing is ```30s``` for the ```BuildStateTransition``` table and ```8s``` for the ```ArtifactPlans``` table for a H2 db of size 6.2G

arvindsv:
Hmm. Ok. Can you try it with one more? Also, how many rows were in that table?

PR #2180, opened 2016-05-12
adding retries for test data setup as it fails randomly on widows wit…

jyotisingh:
…h file lock exception


server/test/integration/com/thoughtworks/go/server/service/ConfigSaveDeadlockDetectionIntegrationTest.java
```diff
@@ -151,6 +149,36 @@ public void run() {
...
+        if (!SystemUtil.isWindows()) {
+            update(configFile);
+            return;
+        }
+        int retries = 1;
+        while (retries <= 5) {
+            try {
+                update(configFile);
+                return;
+            } catch (Exception e) {
```
zabil: Can we catch a specific exception here?

PR #2189, opened 2016-05-17
Remove duplication of symbol to make it easy to "port" APIs over.

ketan:
Bring APIv2 base stuff in line with apiv1.


ketan:
This is a copy-paste job to bring utility code over from ApiV1 into ApiV2, along with de-duplication so as to make it easy to copy over code from Apiv1 into ApiV2 with minimal string replacements.

PR #2193, opened 2016-05-18
#1133 config repos integration with post commit hooks

tomzo:
As a user of [configuration repositories](#1133) I may find it frustrating to wait up to 1 minute for newly pushed Go configuration to be parsed and validated in Go server.
There are 2 solutions for that -
 1. Make the dedicated config repo poller's interval smaller. (We can always do that anyway.)
 2. Use post commit hooks, just like on other materials. This is the fastest feedback, obviously.

## Problem

There is one problem with how current post commit hooks works. It requires `autoUpdate=false`. Which is in conflict with what config repo material requires -`autoUpdate=true`.
There is a **good reason why config repos must be polled anyway**:
1. At server start we must ensure latest remote configurations are fetched.
2. We cannot rely on post commit hooks to be **the only source of MDU requests**, because:
  * There would be no event to fetch **newly added repository**. E.g. user just adds new git repository where he has new pipelines, but in order to *see* any pipelines configured remotely, he would have to trigger the hook first, at least once.
  * If any API call to `api/material/notify` fails for whatever reason, there would be no mechanism to ensure that latest config finally gets parsed by server.

## Implementation

 * We keep the requirement that `autoUpdate=true` for all configuration repositories. We keep polling all configuration repositories. (No changes here)
 * If material is used in configuration repository then it is allowed to be triggered via post commit hook (API call to `api/material/notify`).

The effect is that any configuration repository is both 
 * polled once a minute
 * and it can be additionaly triggred by API call to `api/material/notify`

### DIFF
--- a/common/test/unit/com/thoughtworks/go/domain/PipelineGroupsTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/PipelineGroupsTest.java
@@ -17,6 +17,7 @@
 package com.thoughtworks.go.domain;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -48,6 +49,7 @@ import org.junit.Test;
 
 import static com.thoughtworks.go.helper.PipelineConfigMother.createGroup;
 import static com.thoughtworks.go.helper.PipelineConfigMother.createPipelineConfig;
+import static com.thoughtworks.go.helper.PipelineConfigMother.pipelineConfig;
 import static java.util.Arrays.asList;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsCollectionContaining.hasItem;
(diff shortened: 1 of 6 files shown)


tpidor:
Hi,
I've been trying to poke gocd api to apply the yaml pipeline immediately after I pushed to config-repo, but always getting the error below. Polling works, it's just that I can't keep waiting for the polling to update the pipeline.

curl -u 'tpidor:*******' 'http://***********:8153/go/api/material/notify/git' -H 'Confirm: true' -X POST -d "repository_url=git@********/tpidor/gocd_pipeline_****.git" 

log -
Unable to find material. Materials must be configured not to poll for new changes before they can be used with the notification mechanism.

This is in config xml -
    <config-repo plugin="yaml.config.plugin">
      <git url="git@**********:tpidor/gocd_pipeline_****.git" branch="test_branch_poc" />
    </config-repo>

Issue #2194, opened 2016-05-18
A problem should be reported to the user when a triggered build encounters a broken (Git) material (invalid branch)


MartinAhrer:
##### Issue Type
- Bug Report
##### Summary

Go should report an error when trying to build a (Git) branch that does not exist (has been deleted)
###### Basic environment details

```
Go Server Version:  16.3.0(3183-60d6fd31f00527da00f2affdd8f7f7732e7be103)
JVM version:    1.7.0_95
OS Information: Linux 4.4.8-boot2docker
Usable space in artifacts repository:   13 GB
Database schema version:    1503002
```
##### Steps to Reproduce
1. Setup a Git material with a branch 
2. Initial build works
3. Delete the branch from the repository
4. Trigger Build
5. Go will not schedule a build but will also not report the broken material configuration
##### Expected Results

Go should report a problem to the user
##### Actual Results

Nothing happens

zabil:
Isn't it reported on this box in the right bottom corner?

![screen shot 2016-05-19 at 8 51 37 am](https://cloud.githubusercontent.com/assets/54427/15381649/0e40097c-1d9f-11e6-8194-c9e1e41eaf12.png)

varshavaradarajan:
Hey,
Verified on 16.3 and 16.4 that once a branch is deleted, modification check for that material fails with the following error and comes up in the errors box as @zabil pointed out.
![modification_fail](https://cloud.githubusercontent.com/assets/4715748/15382341/21d064ca-1da7-11e6-8c60-38f967dabf29.png)

However, on 16.5 this bug exists. Probable cause could be a regression failure from https://github.com/gocd/gocd/pull/2149

varshavaradarajan:
The error message shown above is the result of a failed git fetch command - 

The command ```git fetch origin``` fails because the refspec for **shallowed** repositories is like this:
```
fetch = +refs/heads/<branch name>:refs/remotes/origin/<branch name>
```
For a **non-shallow repository**, the refspec is like this: 
```
fetch = +refs/heads/*:refs/remotes/origin/*
```
The reason that the git fetch command does not fail on 16.5 was because the shallow clone system property was turned off by default. So, any **new** material would have the latter refspec, which would not result in any error on running the command ```git fetch origin``` even if a particular branch did not exist. 
The bug exists on earlier versions of Go, before the shallow clone feature was introduced.

MartinAhrer:
[...]

PR #2198, opened 2016-05-20
Git fetch should remove the remote tracking refs from working copy if the ref no longer exists on remote. #2194

varshavaradarajan:


common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
```diff
@@ -299,7 +299,7 @@ private String remoteBranch() {
 
     public void fetch(ProcessOutputStreamConsumer outputStreamConsumer) {
         outputStreamConsumer.stdOutput("[GIT] Fetching changes");
-        CommandLine gitFetch = git(environment).withArgs("fetch", "origin").withWorkingDir(workingDir);
+        CommandLine gitFetch = git(environment).withArgs("fetch", "origin", "--prune").withWorkingDir(workingDir);
```
ketan: @varshavaradarajan — we only track one remote branch, why is this required then?

varshavaradarajan:
We do a fit fetch origin. For a non-shallow repo, the refspec is 
```
fetch = +refs/heads/*:refs/remotes/origin/*
```
So even if we have only one branch, ```git fetch origin``` will update all the remote refs kept in the working copy. I thought of changing the refspec while fetching to fetch only the branch specified in the config. But it will break the functionality of submodules.  So, prune will remove the remote ref from the working copy if the remote branch no longer exists.

Issue #2200, opened 2016-05-21
Make shine report generation thread safe


zabil:
##### Issue Type
 Bug Report

##### Summary

After a stage/job completion a sub component called shine extracts RDF using [GRDDL](https://www.w3.org/TR/2006/WD-grddl-primer-20061002/l). 

This is done by applying an xslt transformer that is cached but not thread safe leading to loss of data when stages/jobs complete at the same time

###### Basic environment details

Go Version: 16.5 and below

##### Expected Results

Generation of the shine stage/job RDF data

##### Actual Results
Depending on the transformer used
For Xalan it's
```
java.util.EmptyStackException
	at org.apache.xml.utils.ObjectStack.peek(ObjectStack.java:130)
	at org.apache.xpath.XPathContext.getNamespaceContext(XPathContext.java:922)
	at org.apache.xalan.transformer.TransformerImpl.transform(TransformerImpl.java:641)
	at org.apache.xalan.transformer.TransformerImpl.transform(TransformerImpl.java:1275)
	at org.apache.xalan.transformer.TransformerImpl.transform(TransformerImpl.java:1253)
	at com.thoughtworks.studios.shine.semweb.grddl.GRDDLTransformer.transform(GRDDLTransformer.java:49)
	at com.thoughtworks.studios.shine.cruise.GoGRDDLResourceRDFizer.importFile(GoGRDDLResourceRDFizer.java:52)
	at com.thoughtworks.studios.shine.cruise.GoGRDDLResourceRDFizer.importDocumentUsingGRDDL(GoGRDDLResourceRDFizer.java:70)
	at com.thoughtworks.studios.shine.cruise.GoGRDDLResourceRDFizer.importURIUsingGRDDL(GoGRDDLResourceRDFizer.java:61)
	at com.thoughtworks.studios.shine.cruise.stage.details.StageResourceImporter.loadIsolatedStageGraph(StageResourceImporter.java:96)
	at com.thoughtworks.studios.shine.cruise.stage.details.StageResourceImporter.load(StageResourceImporter.java:82)
	at com.thoughtworks.studios.shine.cruise.BackgroundStageLoader.handle(BackgroundStageLoader.java:71)
	at com.thoughtworks.studios.shine.cruise.stage.feeds.StageAtomFeedsReader.readFromLatest(StageAtomFeedsReader.java:51)
	at com.thoughtworks.studios.shine.cruise.BackgroundStageLoader.load(BackgroundStageLoader.java:61)
[...]

PR #2201, opened 2016-05-21
Fix multiple threading issues with `XSLTTransformerRegistry`

ketan:
* `XSLTTransformerRegistry` was fundamentally unsafe because of incorrect
  usage of the `Transformer`. Instead of storing instances of `Transformer`
  store instances of `Templates` that can create transformers on demand.
* Since creating instances of `XSLTTransformerRegistry` was expensive,
  `LazyStageGraphLoader` was holding an object pool containing instances
  of `XSLTTransformerRegistry` objects, which is no longer required.
  This class was also potentially leaking file handles before #2187


zabil:
Related to #2200

Issue #2203, opened 2016-05-22
Broken integration with script-executor plugin


tomzo:
##### Issue Type

- Bug Report

##### Summary

Started server with script executor plugin `0.2`. The following error is in the logs
```
java.lang.RuntimeException: Interaction with plugin with id 'script-executor' implementing 'task' extension failed while requesting for 'go.plugin-settings.get-configuration'. Reason: [null]
```
And plugin is not operational since start.
NullPointer suggests a bug IMHO.

###### Basic environment details

I ran this on master code + my changes, but I doubt that I broke it.

* Go Version: `16.5.0`
* JAVA Version: `7`
* OS: Linux

##### Log snippets

Complete message in log

[...]

PR #2204, opened 2016-05-22
#2203 properly handle null response from plugin

tomzo:
#2203


ketan:
LGTM, setting for 16.6.

ketan:
Can we just have a test going so as to not regress?

Issue #2100, opened 2016-04-11
Feature enhancement : API : Add environment endpoint to add a pipeline


patrick92100:
##### Issue Type

- Feature enhancement

##### Summary

API : Add environment endpoint to add a pipeline

##### Environment

I have got an environment ENV1(a GO-CD one), all my agents are under this environment ENV1

I create a pipeline with the API, I unpause the Pipeline with the API
but I can't add the pipeline to the ENV1environment with the API
I can trigger it with the API but as far as the pipeline is not in any environment, stage/jobs are scheduled but never assigned to Agent...

###### Basic environment details

Go Server Version: 	16.3.0(3183-60d6fd31f00527da00f2affdd8f7f7732e7be103)
JVM version: 	1.7.0_40
OS Information: 	Linux 2.6.18-194.el5

##### Steps to Reproduce

1. create a Pipeline with the API
2. unpause the Pipeline with the API
3. trigger the createdPipeline with the API 
4. stage/jobs are scheduled but never assigned...(not set in any environment)

##### Expected Results
add a pipeline to an environment with the API

##### Actual Results
stage/jobs are scheduled but never assigned...(not set in any environment)

##### Possible Fix
add environment API or 'environment' parameter in create pipeline API endpoint

herby:
+1000.
Pls, API for environment <-> pipeline relation is badly needed.

FredrikWendt:
+1

agustincastanio:
+1

arvindsv:
Coming up in [16.7.0](https://api.go.cd/16.7.0/#update-an-environment)? @GaneshSPatil, can this be closed once 16.7.0 is released?

GaneshSPatil:
@arvindsv -- yes

arvindsv:
Closing this. [Will be](https://github.com/gocd/gocd/pull/2207) in 16.7.0. Documentation [here](https://api.go.cd/16.7.0/#update-an-environment).

PR #2207, opened 2016-05-24
Added environments api (#2100)

### LINKED ISSUE #966: Api method for managing environments required
There are no API methods available for managing environments (add/remove/edit) or for assigning agents or pipelines to environments.

Adding environment variables would be a bonus.

GaneshSPatil:
Fixes #966.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/exceptions/NoSuchEnvironmentException.java
+++ b/config/config-api/src/com/thoughtworks/go/config/exceptions/NoSuchEnvironmentException.java
@@ -1,5 +1,5 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -12,7 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.config.exceptions;
 
--- a/server/src/com/thoughtworks/go/config/GoConfigDao.java
+++ b/server/src/com/thoughtworks/go/config/GoConfigDao.java
@@ -19,11 +19,13 @@ package com.thoughtworks.go.config;
 import com.rits.cloning.Cloner;
 import com.thoughtworks.go.config.commands.CheckedUpdateCommand;
 import com.thoughtworks.go.config.commands.EntityConfigUpdateCommand;
+import com.thoughtworks.go.config.update.AddEnvironmentCommand;
 import com.thoughtworks.go.config.update.ConfigUpdateCheckFailedException;
 import com.thoughtworks.go.config.validation.GoConfigValidity;
 import com.thoughtworks.go.listener.ConfigChangedListener;
 import com.thoughtworks.go.presentation.TriStateSelection;
 import com.thoughtworks.go.server.domain.Username;
+import com.thoughtworks.go.server.service.EnvironmentConfigService;
 import com.thoughtworks.go.server.util.UserHelper;
 import com.thoughtworks.go.util.ExceptionUtils;
 import org.slf4j.Logger;
@@ -36,7 +38,6 @@ import java.util.Arrays;
 import java.util.List;
 
 import static com.thoughtworks.go.util.ExceptionUtils.bomb;
-import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
 
 /**
  * @understands how to modify the cruise config sources
@@ -64,8 +65,8 @@ public class GoConfigDao {
         updateConfig(pipelineAdder(pipelineConfig, groupName));
     }
 
-    public void addEnvironment(BasicEnvironmentConfig environmentConfig) {
-        updateConfig(environmentAdder(environmentConfig));
+    public void addEnvironment(BasicEnvironmentConfig environmentConfig, Username user) {
+        updateConfig(new AddEnvironmentCommand(environmentConfig, user));
     }
 
     public CruiseConfig loadForEditing() {
@@ -211,7 +212,6 @@ public class GoConfigDao {
     }
 
 
-
     private UpdateConfigCommand pipelineAdder(final PipelineConfig pipelineConfig, final String groupName) {
         return new UpdateConfigCommand() {
             public CruiseConfig update(CruiseConfig cruiseConfig) {
@@ -220,16 +220,6 @@ public class GoConfigDao {
             }
         };
     }
-
-    private UpdateConfigCommand environmentAdder(final BasicEnvironmentConfig environmentConfig) {
-        return new UpdateConfigCommand() {
-            public CruiseConfig update(CruiseConfig cruiseConfig) {
-                cruiseConfig.addEnvironment(environmentConfig);
-                return cruiseConfig;
-            }
-        };
-    }
-
 
     public UpdateConfigCommand mailHostUpdater(final MailHost mailHost) {
         return new UpdateConfigCommand() {
@@ -308,39 +298,8 @@ public class GoConfigDao {
         }
     }
 
-    public static class ModifyEnvironmentCommand implements UpdateConfigCommand {
-        private final String uuid;
-        private final String environmentName;
-        private final TriStateSelection.Action action;
-
-        public ModifyEnvironmentCommand(String uuid, String environmentName, TriStateSelection.Action action) {
-            this.uuid = uuid;
-            this.environmentName = environmentName;
-            this.action = action;
-        }
-
-        public CruiseConfig update(CruiseConfig cruiseConfig) throws Exception {
-            AgentConfig agentConfig = cruiseConfig.agents().getAgentByUuid(uuid);
-            bombIfNull(agentConfig, "Unable to set agent resources; Agent [" + uuid + "] not found.");
-            EnvironmentConfig environmentConfig = cruiseConfig.getEnvironments().find(new CaseInsensitiveString(environmentName));
-            if (environmentConfig == null) {
-                agentConfig.addError("environments", "Environment [" + environmentName + "] does not exist.");
-            } else {
-                if (action.equals(TriStateSelection.Action.add)) {
-                    environmentConfig.addAgentIfNew(uuid);
-                } else if (action.equals(TriStateSelection.Action.remove)) {
-                    environmentConfig.removeAgent(uuid);
-                } else if (action.equals(TriStateSelection.Action.nochange)) {
-                    //do nothing
-                } else {
-                    bomb(String.format("unsupported action '%s'", action));
-                }
-            }
-            return cruiseConfig;
-        }
-    }
-
     public GoConfigHolder loadConfigHolder() {
         return cachedConfigService.loadConfigHolder();
     }
+
 }
--- a/server/src/com/thoughtworks/go/config/update/ModifyEnvironmentCommand.java
+++ b/server/src/com/thoughtworks/go/config/update/ModifyEnvironmentCommand.java
@@ -1,0 +1,40 @@
+package com.thoughtworks.go.config.update;
+
+import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.presentation.TriStateSelection;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
+
+@Deprecated //needed for old pages. Delete once API and SPA is ready
+public class ModifyEnvironmentCommand implements UpdateConfigCommand {
+    private final String uuid;
+    private final String environmentName;
+    private final TriStateSelection.Action action;
+
+    public ModifyEnvironmentCommand(String uuid, String environmentName, TriStateSelection.Action action) {
+        this.uuid = uuid;
+        this.environmentName = environmentName;
+        this.action = action;
+    }
+
+    public CruiseConfig update(CruiseConfig cruiseConfig) throws Exception {
+        AgentConfig agentConfig = cruiseConfig.agents().getAgentByUuid(uuid);
+        bombIfNull(agentConfig, "Unable to set agent resources; Agent [" + uuid + "] not found.");
+        EnvironmentConfig environmentConfig = cruiseConfig.getEnvironments().find(new CaseInsensitiveString(environmentName));
+        if (environmentConfig == null) {
+            agentConfig.addError("environments", "Environment [" + environmentName + "] does not exist.");
+        } else {
+            if (action.equals(TriStateSelection.Action.add)) {
+                environmentConfig.addAgentIfNew(uuid);
+            } else if (action.equals(TriStateSelection.Action.remove)) {
+                environmentConfig.removeAgent(uuid);
+            } else if (action.equals(TriStateSelection.Action.nochange)) {
+                //do nothing
+            } else {
+                bomb(String.format("unsupported action '%s'", action));
+            }
+        }
+        return cruiseConfig;
+    }
+}
--- a/server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
@@ -16,20 +16,13 @@
 
 package com.thoughtworks.go.server.service;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-
 import com.rits.cloning.Cloner;
 import com.thoughtworks.go.config.*;
 import com.thoughtworks.go.config.exceptions.NoSuchEnvironmentException;
-import com.thoughtworks.go.domain.AgentInstance;
-import com.thoughtworks.go.domain.ConfigElementForEdit;
-import com.thoughtworks.go.domain.EnvironmentPipelineMatcher;
-import com.thoughtworks.go.domain.EnvironmentPipelineMatchers;
-import com.thoughtworks.go.domain.JobPlan;
+import com.thoughtworks.go.config.update.AddEnvironmentCommand;
+import com.thoughtworks.go.config.update.DeleteEnvironmentCommand;
+import com.thoughtworks.go.config.update.UpdateEnvironmentCommand;
+import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.i18n.Localizable;
 import com.thoughtworks.go.i18n.LocalizedMessage;
 import com.thoughtworks.go.listener.ConfigChangedListener;
@@ -40,15 +33,22 @@ import com.thoughtworks.go.remote.work.BuildAssignment;
 import com.thoughtworks.go.server.domain.Username;
 import com.thoughtworks.go.server.service.result.HttpLocalizedOperationResult;
 import com.thoughtworks.go.server.service.result.LocalizedOperationResult;
+import com.thoughtworks.go.serverhealth.HealthStateScope;
 import com.thoughtworks.go.serverhealth.HealthStateType;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+
+import java.util.*;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
 
 /**
  * @understands grouping of agents and pipelines within an environment
  */
 @Service
 public class EnvironmentConfigService implements ConfigChangedListener {
+
     public final GoConfigService goConfigService;
     private final SecurityService securityService;
 
@@ -64,7 +64,7 @@ public class EnvironmentConfigService implements ConfigChangedListener {
 
     public void initialize() {
         goConfigService.register(this);
-        goConfigService.register(new EntityConfigChangedListener<Agents>(){
+        goConfigService.register(new EntityConfigChangedListener<Agents>() {
             @Override
             public void onEntityConfigChange(Agents entity) {
                 sync(goConfigService.getEnvironments());
@@ -151,6 +151,10 @@ public class EnvironmentConfigService implements ConfigChangedListener {
         return environments.named(new CaseInsensitiveString(environmentName));
     }
 
+    public EnvironmentConfig getEnvironmentConfig(String environmentName) throws NoSuchEnvironmentException {
+        return environments.named(new CaseInsensitiveString(environmentName));
+    }
+
     public ConfigElementForEdit<EnvironmentConfig> forEdit(String environmentName, HttpLocalizedOperationResult result) {
         ConfigElementForEdit<EnvironmentConfig> edit = null;
         try {
@@ -163,7 +167,7 @@ public class EnvironmentConfigService implements ConfigChangedListener {
         return edit;
     }
 
-    public void createEnvironment(final BasicEnvironmentConfig config, Username user, final LocalizedOperationResult result) {
+    public void createEnvironment(final BasicEnvironmentConfig config, final Username user, final LocalizedOperationResult result) {
         Localizable noPermission = LocalizedMessage.string("NO_PERMISSION_TO_ADD_ENVIRONMENT", user.getDisplayName());
         Localizable.CurryableLocalizable actionFailed = LocalizedMessage.string("ENV_ADD_FAILED");
         editEnvironments(new EditEnvironments() {
@@ -172,12 +176,12 @@ public class EnvironmentConfigService implements ConfigChangedListener {
                     result.conflict(LocalizedMessage.string("CANNOT_ADD_ENV_ALREADY_EXISTS", config.name()));
                     return;
                 }
-                goConfigService.addEnvironment(config);
+                goConfigService.updateConfig(new AddEnvironmentCommand(config, user));
             }
-        }, config, user, result, noPermission, actionFailed);
+        }, user, result, noPermission, actionFailed);
     }
 
-    private void editEnvironments(EditEnvironments edit, EnvironmentConfig config, Username user, LocalizedOperationResult result,
+    private void editEnvironments(EditEnvironments edit, Username user, LocalizedOperationResult result,
                                   Localizable noPermissionMessage, Localizable.CurryableLocalizable actionFailedMessage) {
         if (!securityService.isUserAdmin(user)) {
             result.unauthorized(noPermissionMessage, HealthStateType.unauthorised());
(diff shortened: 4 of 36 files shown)


server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -145,6 +136,24 @@ public EnvironmentConfig named(String environmentName) throws NoSuchEnvironmentE
         return environments.named(new CaseInsensitiveString(environmentName));
     }
 
+    public EnvironmentConfig getEnvironmentConfig(String environmentName) throws NoSuchEnvironmentException {
+        if(environments.hasEnvironmentNamed(new CaseInsensitiveString(environmentName)))
+            return environments.named(new CaseInsensitiveString(environmentName));
+        return null;
```
ketan: Does not really throw `NoSuchEnvironmentException` is it. There's also a race condition here. Between the time you check `hasEnvironmentNamed` and call `environments.named`, some other thread could have added/removed the environment.

IMO, it's probably best to delegate straight to `environments.named` instead.

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -145,6 +136,24 @@ public EnvironmentConfig named(String environmentName) throws NoSuchEnvironmentE
...
+    }
+
+    public void deleteEnvironment(String environmentName, HttpLocalizedOperationResult result) throws NoSuchEnvironmentException {
+        CruiseConfig config = goConfigService.getConfigForEditing();
+        EnvironmentsConfig environments = config.getEnvironments();
+        if (environments.remove(getEnvironmentConfig(environmentName))) {
+            config.setEnvironments(environments);
```
ketan: Race condition here —

If another thread were to do add/remove an environment after you store the original environments using `config.getEnvironments` this call to `config.setEnvironments()` will get rid of those environments that were deleted, or add back environments that were removed (essentially rolling back any changes by the other thread)

ketan: Also any edits to an existing environment by another thread may be lost here because you remember the "old state" of ALL environments.

ketan: @jyotisingh (and @tomzo) would know best how to deal with these changes given cruise config partials now :)

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,66 @@
+##########################################################################
+# Copyright 2015 ThoughtWorks, Inc.
```
ketan: We're well into 2016 now, I think :)

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+##########################################################################
+
+module ApiV1
+  module Admin
+    class EnvironmentsController < ApiV1::BaseController
+      before_action :check_admin_user_and_401
+      before_action :load_environment, only: [:show, :update, :destroy]
+
+      def index
+        environments = environment_config_service.getEnvironments().to_a
```
ketan: Should not need the `to_a` since the object being returned is already an enumerable (`java.util.Set`).

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      def destroy
+        result = HttpLocalizedOperationResult.new
+        environment_config_service.deleteEnvironment(params[:name], result)
+        render_http_operation_result(result)
+      end
+
+      private
+      def load_environment(environment_name = params[:name])
+        @environment_config = environment_config_service.getEnvironmentConfig(environment_name)
+        raise RecordNotFound if @environment_config.nil?
```
ketan: While this code is correct, the correct ruby idiom would be `raise RecordNotFound unless @environment_config`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      end
+
+      def destroy
+        result = HttpLocalizedOperationResult.new
+        environment_config_service.deleteEnvironment(params[:name], result)
+        render_http_operation_result(result)
+      end
+
+      private
+      def load_environment(environment_name = params[:name])
```
ketan: Is there ever a case where you pass a different environment name to have to justify the optional argument?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/environment_config_representer.rb
```diff
@@ -0,0 +1,71 @@
...
+        'http://api.go.cd/#environment_config'
+      end
+
+      link :find do |opts|
+        opts[:url_builder].apiv1_admin_environment_url(name: '__environment_name__').gsub(/__environment_name__/, ':environment_name')
+      end
+
+      property :name, case_insensitive_string: true
+
+      delegate :name, :name=, to: :environment
```
ketan: Please move this delegate to after all `property` and `collection` declarations, to make it easy to follow the code (and the convention)

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/environments_config_representer.rb
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Config
+    class EnvironmentsConfigRepresenter < ApiV1::BaseRepresenter
+
+      link :self do |opts|
+        p opts[:url_builder]
```
ketan: Remove this print statement?

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,245 @@
...
+      it 'should disallow anonymous users, with security enabled' do
+        enable_security
+        login_as_anonymous
+        expect(controller).to disallow_action(:get, :index).with(401, 'You are not authorized to perform this action.')
+      end
+
+      it 'should disallow normal users, with security enabled' do
+        login_as_user
+        expect(controller).to disallow_action(:get, :index).with(401, 'You are not authorized to perform this action.')
+      end
```
ketan: @jyotisingh @arvindsv — as of now read-only users can see a list of environments and also view the pages for environment variables. Intentional or not — do we want to replicate the same behaviour for API?

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -145,6 +136,24 @@ public EnvironmentConfig named(String environmentName) throws NoSuchEnvironmentE
...
+            this.environments.remove(getEnvironmentConfig(environmentName));
+            result.setMessage(LocalizedMessage.string("ENVIRONMENT_DELETE_SUCCESSFUL", environmentName));
+        } else {
+            result.notFound(LocalizedMessage.string("ENVIRONMENT_NOT_FOUND", environmentName), HealthStateType.general(HealthStateScope.GLOBAL));
```
ketan: This feels misleading/confusing given what you're doing here https://github.com/gocd/gocd/pull/2207/files#diff-c13835adf2f105a5355767fdee02e76eR61

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
tomzo: I'd use `EnvironmentsConfig.add` for adding environments.
For removal you can add a `remove` method in `EnvironmentsConfig`, which will search for environment with given name and remove it from list  then removing like this should be safe
```
EnvironmentsConfig environments = config.getEnvironments();
environments.remove(environmentName)
```

Partials won't change much here, because this operation is above all the merging magic. 

Don't call `setEnvironments`, IIRC that is used only from tests, not actual config update operations.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        environment_config_service.deleteEnvironment(params[:name], result)
+        render_http_operation_result(result)
+      end
+
+      private
+      def load_environment
+        begin
+          @environment_config = environment_config_service.getEnvironmentConfig(params[:name])
+        rescue
+          raise RecordNotFound
```
ketan: Are all errors record not found, or just `NoSuchEnvironmentException`?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+      end
+
+      def update
+        environment_config = BasicEnvironmentConfig.new
+        ApiV1::Config::EnvironmentConfigRepresenter.new(environment_config).from_hash(params[:environment])
+        result = environment_config_service.updateEnvironment(params[:name], environment_config, current_user, params[:md5])
+        json = ApiV1::Config::EnvironmentConfigRepresenter.new(environment_config).to_hash(url_builder: self)
+        render_http_operation_result(result, {data: json})
+      end
```
ketan: You're not handling any `NoSuchEnvironmentException` here?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def destroy
+        result = HttpLocalizedOperationResult.new
+        environment_config_service.deleteEnvironment(params[:name], result)
+        render_http_operation_result(result)
+      end
```
ketan: You're not handling any `NoSuchEnvironmentException` here?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Admin
+    class EnvironmentsController < ApiV1::BaseController
```
ketan: See `BaseController` on how to handle generic exceptions without actually `catch`ing them in the controller.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
GaneshSPatil: `load_environment` is called before "show / update / destroy" so if the environment must be present

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
GaneshSPatil: may be `load_environment` will handle it as it is called before `update`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
GaneshSPatil: may be `load_environment` will handle it as it is called before `destroy`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
ketan: Do you think the following tests will pass with your implementation of `load_environment`?


```ruby
environment_config_service.should_receive(:getEnvironments)
       .with('some-environment')
       .and_raise(NullPointerException.new(""))

get :show, name: 'some-environment'
expect(response.code).to eq(500)
```

```ruby
environment_config_service.should_receive(:getEnvironments)
        .with('some-environment')
        .and_raise(NoSuchEnvironmentException.new("foo"))

get :show, name: 'some-environment'
expect(response.code).to eq(404)
```

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -145,6 +136,21 @@ public EnvironmentConfig named(String environmentName) throws NoSuchEnvironmentE
         return environments.named(new CaseInsensitiveString(environmentName));
     }
 
+    public EnvironmentConfig getEnvironmentConfig(String environmentName) throws NoSuchEnvironmentException {
+        return environments.named(new CaseInsensitiveString(environmentName));
+    }
+
+    public void deleteEnvironment(String environmentName, HttpLocalizedOperationResult result) throws NoSuchEnvironmentException {
+        if (environments.hasEnvironmentNamed(new CaseInsensitiveString(environmentName))) {
+            CruiseConfig config = goConfigService.getConfigForEditing();
```
jyotisingh: You should never edit `goConfigService.getConfigForEditing()` directly. You would normally make all relevant changes to a copy of config and then send it for save. Once all the validations pass and save is completed, `getConfigForEditing` on the service would start returning the updated config. Look at `PipelineConfigService` ([this](https://github.com/gocd/gocd/blob/5f54cf4338eb2ff5c9f486ede6ae97e9aae51e82/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java#L97) and [this](https://github.com/gocd/gocd/blob/5f54cf4338eb2ff5c9f486ede6ae97e9aae51e82/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java#L134)) to know more about what I mean

server/src/com/thoughtworks/go/config/GoConfigDao.java
```diff
@@ -624,7 +624,42 @@ public CruiseConfig update(CruiseConfig cruiseConfig) throws Exception {
         }
     }
 
+    public static class UpdateEnvironmentCommand implements UpdateConfigCommand {
+
+        private final String environmentToUpdate;
+        private final EnvironmentConfig newEnvironmentConfig;
+
+        public UpdateEnvironmentCommand(String environmentToUpdate, EnvironmentConfig newEnvironmentConfig) {
```
ketan: Isn't `environmentToUpdate` supposed to be the same as `newEnvironmentConfig.name()`?

GaneshSPatil: We can modify the name of the existing environment,
So if the name has modified, the old name will be used to get the existing copy of the environment and in this case `environmentToUpdate != newEnvironmentConfig.name`

ketan: Do you have a test for this behaviour, so that the next person who decides to "refactor" this code in the future knows not to?

server/test/integration/com/thoughtworks/go/server/service/EnvironmentConfigServiceIntegrationTest.java
```diff
@@ -160,7 +163,101 @@ public void shouldReturnBadRequestForUpdateWhenUsingInvalidEnvName() {
...
+        HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
+        service.updateEnvironment("bar-env", env("foo-env", new ArrayList<String>(), new ArrayList<Map<String, String>>(), new ArrayList<String>()), new Username(new CaseInsensitiveString("any")), result);
+        assertThat(result.message(localizer), is("Failed to update environment 'bar-env'. Duplicate unique value [foo-env] declared for identity constraint \"uniqueEnvironmentName\" of element \"environments\"."));
+    }
+
+    @Test
+    public void shouldReturnBadRequestForUpdateWhenUsingInvalidEnvName_ForNewUpdateEnvironmentMethod_ForNewUpdateEnvironmentMethod() {
+        configHelper.addEnvironments("foo-env");
+        HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
+        service.updateEnvironment("foo-env", env("foo env", new ArrayList<String>(), new ArrayList<Map<String, String>>(), new ArrayList<String>()), new Username(new CaseInsensitiveString("any")), result);
```
ketan: I think just `env(String name)` should be good here?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+        render_http_operation_result(result)
+      end
+
+      private
+
+      def load_environment(environment_name = params[:name])
+        begin
+          @environment_config = environment_config_service.getEnvironmentConfig(environment_name)
+        rescue
+          raise RecordNotFound
```
ketan: This is still not fixed — see my comment here https://github.com/gocd/gocd/pull/2207#discussion_r64517551

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      def get_environment_from_request
+        @environment_config_from_request ||= BasicEnvironmentConfig.new.tap do |config|
+          ApiV1::Config::EnvironmentConfigRepresenter.new(config).from_hash(params[:environment])
+        end
+      end
+
+      def get_etag_for_environment(environment_name, json)
+        cache_key = environment_name.downcase
+        etag      = go_cache.get("GO_ENVIRONMENT_CONFIGS_ETAGS_CACHE", cache_key)
```
ketan: You seem to be using a cache key — but this is never being invalidated. You might want to add something to `EnvironmentConfigService#onConfigChange` to invalidate the cache key. See `PipelineConfigService#onConfigChange`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+        end
+        etag
+      end
+
+      def handle_config_save_or_update_result(result, environment_name = params[:name])
+        if result.isSuccessful
+          load_environment(environment_name)
+          json = ApiV1::Config::EnvironmentConfigRepresenter.new(@environment_config).to_hash(url_builder: self)
+          response.etag = [get_etag_for_environment(@environment_config.name.to_s, json)]
+          render json_hal_v1: json
```
ketan: Please use `DEFAULT_FORMAT` instead of `json_hal_v1`

server/src/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -929,6 +930,15 @@ public String unmodifiedMd5() {
         });
     }
```
jyotisingh: Would prefer if we move both these methods to `EnvironmentConfigService` instead. From there, you could invoke `public ConfigSaveState updateConfig(UpdateConfigCommand command)` (which internally invokes `goConfigDao.updateConfig`)

arvindsv:
Fixes #966. Let's see if this comment sets it up to close #966 automatically. If not, I'll change the description.

boranseref:
Great! We automated the creation of new pipelines, but we still need to add them to a specific environment manually. Any update on this PR? We're waiting for this one to upgrade our server.

config/config-api/src/com/thoughtworks/go/config/exceptions/NoSuchEnvironmentException.java
```diff
@@ -25,4 +25,8 @@
     public NoSuchEnvironmentException(CaseInsensitiveString environmentName) {
         super("Environment [" + environmentName + "] does not exist.");
     }
+
+    public NoSuchEnvironmentException(String environmentName) {
```
varshavaradarajan: I do not see usages of this method in your code. Please check and remove this.

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -172,12 +174,12 @@ public void performEdit() {
                     result.conflict(LocalizedMessage.string("CANNOT_ADD_ENV_ALREADY_EXISTS", config.name()));
                     return;
                 }
-                goConfigService.addEnvironment(config);
+                goConfigService.addEnvironment(config, user);
```
varshavaradarajan: Consider using the AddEnvironmentCommand for consistency as you use UpdateEnvironmentCommand and DeleteEnvrironmentCommand in the respective functions

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -223,25 +225,157 @@ private void editEnvironments(EditEnvironments edit, EnvironmentConfig config, U
...
+        public CruiseConfig update(CruiseConfig cruiseConfig) {
+            cruiseConfig.addEnvironment(environmentConfig);
+            return cruiseConfig;
+        }
+
+        @Override
+        public ConfigModifyingUser user() {
+            return new ConfigModifyingUser(user);
+        }
+    }
```
varshavaradarajan: Consider moving AddEnvironmentCommand, DeleteEnvironmentCommand, UpdateEnvironmentCommand, ModifyEnvironmentCommand out of the service for readability

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -223,25 +225,157 @@ private void editEnvironments(EditEnvironments edit, EnvironmentConfig config, U
...
+        }
+
+        @Override
+        public ConfigModifyingUser user() {
+            return new ConfigModifyingUser(username);
+        }
+    }
+
+    public static class ModifyEnvironmentCommand implements UpdateConfigCommand {
+        private final String uuid;
```
varshavaradarajan: Can you think of another name for this class? UpdateEnvironmentCommand and ModifyEnvironmentCommand seem a little confusing :)

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+module ApiV1
+  module Admin
+    class EnvironmentsController < ApiV1::BaseController
+      before_action :check_admin_user_and_401
+      before_action :load_environment, only: [:show, :update, :destroy]
+      before_action :check_for_stale_request, only: [:update]
+
+      def index
+        render :json_hal_v1 => ApiV1::Config::EnvironmentsConfigRepresenter.new(environment_config_service.getEnvironments()).to_hash(url_builder: self)
```
varshavaradarajan: Please use ```render DEFAULT_FORMAT``` instead of ```render json_hal_v1:``` makes it easy to upgrade to api v2 if needed

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+        @environment_name = SecureRandom.hex
+        @environment_config_service.stub(:getEnvironmentConfig).and_raise(com.thoughtworks.go.config.exceptions.NoSuchEnvironmentException.new('foo-env'))
+        patch_with_api_header :update, name: @environment_name
+        expect(response).to have_api_message_response(404, 'Either the resource you requested was not found, or you are not authorized to perform this action.')
+      end
+    end
+  end
+
+  describe :destroy do
+    before(:each) do
```
varshavaradarajan: security tests for delete action

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+
+      it 'should disallow normal users, with security enabled' do
+        login_as_user
+        expect(controller).to disallow_action(:get, :show, name: @environment_name).with(401, 'You are not authorized to perform this action.')
+      end
+    end
+  end
+
+  describe :update do
+
```
varshavaradarajan: security tests for update action

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+        expect(actual_response).to eq({ :message => "Someone has modified the configuration for environment 'foo-environment'. Please update your copy of the config with the changes." })
+      end
+
+      it 'should not update environment config if no etag is passed' do
+        login_as_admin
+        patch_with_api_header :update,name: @environment_name, :environment => { name: @environment_name, pipelines: [], agents: [], environment_variables: []}
+
+        expect(response.status).to eq(412)
+        expect(actual_response).to eq({ :message => "Someone has modified the configuration for environment 'foo-environment'. Please update your copy of the config with the changes." })
+      end
```
varshavaradarajan: change to use have_api_message_response

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+      it 'should allow deleting environments' do
+        login_as_admin
+
+        @environment_config_service.should_receive(:deleteEnvironment).with(@environment_name,an_instance_of(Username), an_instance_of(HttpLocalizedOperationResult)) do |name, user, result|
+          result.setMessage(LocalizedMessage.string('ENVIRONMENT_DELETE_SUCCESSFUL', [name].to_java(java.lang.String)))
+        end
+        controller.request.env['HTTP_IF_MATCH'] = "\"#{Digest::MD5.hexdigest('latest-etag')}\""
+
+        delete_with_api_header :destroy, name: @environment_name
+        expect(response.status).to eq(200)
```
varshavaradarajan: No need for this as you already check for it next.

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+        login_as_admin
+
+        @environment_config_service.should_receive(:createEnvironment)
+
+        post_with_api_header :create, :environment => { name: @environment_name, pipelines: [], agents: [], environment_variables: []}
+        expect(response.status).to be(200)
+        expect(actual_response).to eq(expected_response(@environment_config, ApiV1::Config::EnvironmentConfigRepresenter))
+      end
+    end
+
```
varshavaradarajan: Perhaps there are some scenarios when environment cannot be created? 
Example: If environment name already exists

config/config-api/src/com/thoughtworks/go/config/exceptions/NoSuchEnvironmentException.java
GaneshSPatil: used in environments_controller_spec

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
GaneshSPatil: those scenarios are tested in `EnvironmentConfigServiceIntegrationTest.java`

config/config-api/src/com/thoughtworks/go/config/exceptions/NoSuchEnvironmentException.java
varshavaradarajan: Why not just use the Exception with CaseInsensitiveString object? This is introducing the code only for the spec's sake.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/environments_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      def index
+        render DEFAULT_FORMAT => ApiV1::Config::EnvironmentsConfigRepresenter.new(environment_config_service.getEnvironments()).to_hash(url_builder: self)
+      end
+
+      def show
+        json = ApiV1::Config::EnvironmentConfigRepresenter.new(@environment_config).to_hash(url_builder: self)
+        if stale?(etag: JSON.generate(json))
+          render DEFAULT_FORMAT => json
+        end
```
varshavaradarajan: Can you inline this please? ```render DEFAULT_FORMAT => json if stale?(etag: JSON.generate(hash)```

server/src/com/thoughtworks/go/config/GoConfigDao.java
```diff
@@ -221,13 +221,8 @@ public CruiseConfig update(CruiseConfig cruiseConfig) {
...
     }
 
-    private UpdateConfigCommand environmentAdder(final BasicEnvironmentConfig environmentConfig) {
-        return new UpdateConfigCommand() {
-            public CruiseConfig update(CruiseConfig cruiseConfig) {
-                cruiseConfig.addEnvironment(environmentConfig);
-                return cruiseConfig;
-            }
-        };
+    private UpdateConfigCommand environmentAdder(final BasicEnvironmentConfig environmentConfig, Username user) {
```
ketan: Please remove and inline these one line "factory" methods, we're trying to get away from these.

server/src/com/thoughtworks/go/config/update/ModifyEnvironmentCommand.java
```diff
@@ -0,0 +1,39 @@
+package com.thoughtworks.go.config.update;
+
+import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.presentation.TriStateSelection;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
+
+public class ModifyEnvironmentCommand implements UpdateConfigCommand {
```
ketan: Please mark these old classes as `@Deprecate` so we can get rid of them when the API is ready and the page is replaced with a SPA.

Issue #2210, opened 2016-05-24
go server and agent deb installation does not allow ignoring of java installation


brettcave:
##### Issue Type
 - Bug Report

##### Summary

The go-server and go-agent deb installation packages force installation of java through a `Pre-Depends` and does not allow the use of a custom java installation.

##### Environment

Ubuntu (14.04 LTS, but expected across all dpkg / apt based package managed distros).

##### Additional Environment

An example use-case is when wanting to run GOCD with Oracle JDK 8 instead of openjdk 7 and the system is provisioned using Chef. Using the go-cookbook in conjunction with the Java cookbook, Java can be configured to install Oracle JDK8, but when go-server installs, it forces an installation of openjdk 7 because of `Pre-Depends` in the control file. 

If `Pre-Depends`  is changed to `Depends`, the options can be overridden through options, for example in a `package` resource in Chef:

```
package "go-server" do
    version [:go][:version]
    options "--force-yes -o Dpkg::Options::='--ignore-depends=java7-runtime-headless'"
   notifies :start, 'service[go-server]', :immediately
end
```

##### Possible Fix
https://github.com/gocd/gocd/blob/master/installers/server/deb/control.go-server
```
9  - Pre-Depends: java7-runtime-headless
 9 + Depends: java7-runtime-headless
```
and the same for https://github.com/gocd/gocd/blob/master/installers/agent/deb/control.go-agent

###### Current workaround
The deb file can be repackaged and go-server installed with Pre-Depends removed as follows for a deb-based system:
```
$ ar x go-server-xxx.deb
$ FILES=`tar zxvf control.tar.gz`
$ sed -r -i '/Pre-Depends: .+/d' control
$ tar zcf control.tar.gz $FILES
$ ar rcs go-server.deb debian-binary control.tar.gz data.tar.gz _gpgbuilder
$ sudo dpkg -i go-server.deb
```

##### More info

https://www.debian.org/doc/debian-policy/ch-relationships.html
```
Pre-Depends
    This field is like Depends, except that it also forces dpkg to complete installation of the packages named before even starting the installation of the package which declares the pre-dependency
```
The ignore-depends flag is not applied to pre-depends dependencies.

brettcave:
Have not investigated RPM packaging, not sure if a similar issue is relevant to it.

ketan:
> Have not investigated RPM packaging, not sure if a similar issue is relevant to it.

[...]

PR #2010, opened 2016-03-03
Move to using gradle

ketan:
Pending stuff -

- [x] `NOT_IN_PRODUCTION ` flag
- [ ] Solaris installers
- [x] Windows installers
- [x] Debian installers
- [x] RPM installers
- [x] Zip installers
- [x] OSX installers
- [x] old jasmine tests
- [x] new jasmine tests
- [ ] ~~inline partials~~ (Not needed anymore)
- [x] rspec tests
 
Don't merge this yet — still need to sort out deb, rpm windows and OSX packaging (but I think that's simple to do, now that most of this works fine)

Some gradle targets for you to run, and their equivalent from maven —

buildr |gradle
----|----
`./bn clean cruise:prepare`     | `./gradle clean fatJar`
`SKIP_WAR=Y ./bn ...`           | `./gradle ... -PfastBuild`
`./bn clean cruise:prepare dist` | `./gradle clean agentGenericZip serverGenericZip`

Some build timings —

**Full build, no asset precompilation**

```
./gradlew --profile clean agentGenericZip serverGenericZip -PfastBuild -q 
62.82s user 4.48s system 179% cpu 37.462 total
```

**Full build, no asset precompilation, parallel with some tweaks**

```
./gradlew clean agentGenericZip serverGenericZip -PfastBuild -q
0.97s user 0.12s system 4% cpu 23.623 total
```

**The gradle tweaks**

```
$ cat ~/.gradle/gradle.properties
org.gradle.daemon=true
org.gradle.parallel=true
org.gradle.workers.max=4
```

**What goes where**

* various jars — `target/libs` (gradle convention)
* final distributions — `target/distributions` (inside the "new" installers project)

This is now ready to review (and merge)

PR #2211, opened 2016-05-24
#2210 - go-server and go-agent Depends on java instead of Pre-Depends.

brettcave:
#2210 - The ability to use `--ignore-depends` for the Java dependency in `go-server` and `go-agent` deb packages.

### DIFF
--- a/installers/agent/deb/control.go-agent
+++ b/installers/agent/deb/control.go-agent
@@ -3,7 +3,7 @@ Section: base
 Priority: optional
 Version: @VERSION@-@RELEASE@
 Architecture: all
-Pre-Depends: java7-runtime-headless
+Depends: java7-runtime-headless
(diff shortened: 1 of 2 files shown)


ketan:
@brettcave Could you please sign the [CLA](https://go.cd/cla) it is a one time requirement.

ketan:
I checked on another PR. I've already (accidentally) fixed this here https://github.com/ketan/gocd/blob/move-to-gradle/buildSrc/src/main/groovy/com/thoughtworks/go/build/DebPackagingTask.groovy#L36.

You may track progress on https://github.com/gocd/gocd/pull/2010 which is due to be merged in 16.6 in a few days from now.

brettcave:
Great, thanks @ketan, will keep an eye out for the release.

brettcave:
CLA signed.

Issue #1873, opened 2016-02-02
Create API to list installed plugins for SCM, Repository and Task


juhijariwala:
**Usage:**
Add ability to list installed plugins of specific type for the following reasons:
- When you create/edit pluggable scm or package material from SPA, there is no way to list down installed plugins.
- When you create PluggbleSCM/PackageRepository/PluggableTask via API you need to know the plugin ID and configuration keys.

**Structure:**
URL : ```https://ci.example.com/go/api/admin/plugins/[:type] ```

JSON response for SCM plugin list:

```json
{
  "_links": {
    "self": {
      "href": "https://ci.example.com/go/api/admin/plugins/scm"
    },
    "doc": {
      "href": "http://api.go.cd/#plugins"
    }
  },
  "_embedded": {
    "plugins": [
      {
        "id": "github.pr",
        "version": "1.2.3",
        "configuration_property_keys": [
          {"property": "url", "required": true, "secure": false},
          {"property": "username", "required": false, "secure": false},
          {"property": "password", "required": false, "secure": true}
        ]
      }
    ]
  }
}
```

JSON response for Repository plugin list:

```json
{
  "_links": {
    "self": {
      "href": "https://ci.example.com/go/api/admin/plugins/package_repository"
    },
    "doc": {
      "href": "http://api.go.cd/#plugins"
    }
  },
  "_embedded": {
    "plugins": [
      {
        "id": "deb",
        "version": "1",
        "repo_configuration_property_keys": [
          {"property": "REPO_URL", "required": true, "secure": false},
        ],
        "package_configuration_property_keys": [
          {"property": "PACKAGE_NAME", "required": true, "secure": false},
          {"property": "VERSION_SPEC", "required": true, "secure": false},
          {"property": "ARCHITECTURE", "required": true, "secure": false},
        ]
      }
    ]
  }
}
```

JSON response for Task plugin list:

```
{
  "_links": {
    "self": {
      "href": "https://ci.example.com/go/api/admin/plugins/task"
    },
    "doc": {
      "href": "http://api.go.cd/#plugins"
    }
  },
  "_embedded": {
    "plugins": [
      {
        "id": "xunit.converter.task.plugin",
        "version": "1",
        "configuration_property_keys": [
          "ConverterType",
          "InputDirectory",
          "OutputDirectory"
        ]
      }
    ]
  }
}
```
Note: In the api documentation mention that all the configuration keys are of type String.

maheshp:
[...]

PR #2228, opened 2016-05-27
Admins/Group Admins PluginInfos API to list and view #1873

maheshp:
*    PluginInfo api to fetch plugin details, this will be used to build views on SPA
* PluginService exposes methods to list all plugins and fetch a plugin_info by id
* Added PluginInfoViewModelBuilders for each plugin type
* PluginInfo represents all types of plugins
* plugin_infos_controller exposes api's to list and show

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/plugins/builder/PluginInfoBuilder.java
+++ b/server/src/com/thoughtworks/go/server/service/plugins/builder/PluginInfoBuilder.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.service.plugins.builder;
+
+import com.thoughtworks.go.plugin.access.authentication.AuthenticationExtension;
+import com.thoughtworks.go.plugin.access.authentication.AuthenticationPluginRegistry;
+import com.thoughtworks.go.plugin.access.notification.NotificationExtension;
+import com.thoughtworks.go.plugin.access.notification.NotificationPluginRegistry;
+import com.thoughtworks.go.plugin.access.packagematerial.JsonBasedPackageRepositoryExtension;
+import com.thoughtworks.go.plugin.access.pluggabletask.JsonBasedTaskExtension;
+import com.thoughtworks.go.plugin.access.scm.SCMExtension;
+import com.thoughtworks.go.plugin.infra.PluginManager;
+import com.thoughtworks.go.server.service.plugins.InvalidPluginTypeException;
+import com.thoughtworks.go.server.ui.plugins.PluginInfo;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Component
+public class PluginInfoBuilder {
+
+    private Map<String, ViewModelBuilder> pluginExtensionToBuilder = new HashMap<>();
+
+    @Autowired
+    public PluginInfoBuilder(AuthenticationPluginRegistry authenticationPluginRegistry, NotificationPluginRegistry notificationPluginRegistry, PluginManager pluginManager) {
+        pluginExtensionToBuilder.put(AuthenticationExtension.EXTENSION_NAME, new AuthenticationViewModelBuilder(pluginManager, authenticationPluginRegistry));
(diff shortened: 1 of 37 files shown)


server/src/com/thoughtworks/go/server/ui/PluginViewModel.java
```diff
@@ -0,0 +1,100 @@
...
+
+        this.configurations = pluginConfigurationViewModels;
+    }
+
+    public PluginViewModel(String id, String name, String version, String type, String viewTemplate, List<PluginConfigurationViewModel> pluginConfigurationViewModels) {
+        this(id, name, version, type);
+
+        this.viewTemplate = viewTemplate;
+        this.configurations = pluginConfigurationViewModels;
+    }
```
ketan: These sort of constructors are a huge problem, can we ensure that the constructor with less number of args, calls the constructor with more number of args? This allows for all variables to be marked final. Notice you you are not setting `configurations` but it cannot be final because of how your constructors are wired.

So for e.g. —

```

public PluginViewModel(String id, String name, String version, String type, List<PluginConfigurationViewModel> pluginConfigurationViewModels) {
  this(id, name, version, type, null, pluginConfigurationViewModels);
}

public PluginViewModel(String id, String name, String version, String type) {
  this(id, name, version, type, null, null);
}
```

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -19,7 +19,7 @@
   mount JasmineRails::Engine => '/jasmine-specs-new', as: :jasmine_new if defined?(JasmineRails)
 
   unless defined?(CONSTANTS)
-    USER_NAME_FORMAT = GROUP_NAME_FORMAT = TEMPLATE_NAME_FORMAT = PIPELINE_NAME_FORMAT = STAGE_NAME_FORMAT = ENVIRONMENT_NAME_FORMAT = /[\w\-][\w\-.]*/
+    USER_NAME_FORMAT = GROUP_NAME_FORMAT = TEMPLATE_NAME_FORMAT = PIPELINE_NAME_FORMAT = STAGE_NAME_FORMAT = ENVIRONMENT_NAME_FORMAT = PLUGIN_ID_FORMAT = /[\w\-][\w\-.]*/
```
ketan: Plugin IDs have this format — https://github.com/gocd/gocd/blob/242d113bb7724911ee0c7166b200d56a86f3ada3/plugin-infra/go-plugin-api/resources/plugin-descriptor.xsd#L30-L37

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/plugins_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      before_action :check_admin_user_or_group_admin_user_and_401
+
+      def index
+        render json_hal_v1: ApiV1::PluginsRepresenter.new(plugin_service.plugins(params[:type])).to_hash(url_builder: self)
+      end
+
+      def show
+        plugin = plugin_service.plugin(params[:id])
+
+        raise RecordNotFound if plugin.nil?
```
ketan: `unless plugin`?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/plugin_configuration_representer.rb
```diff
@@ -0,0 +1,30 @@
...
+##########################################################################
+
+module ApiV1
+  class PluginConfigurationRepresenter < BaseRepresenter
+    property :key
+    property :type, skip_nil: true
+    property :metadata, exec_context: :decorator
+
+    def metadata
+      represented.metadata.inject({}) do |hash, (k,v)|
```
ketan: You seem to be transforming a `java.util.Map` returned by `PluginConfigurationViewModel#getMetadata`, any reason to have to do this? If you really needed a ruby hash (i presume for testing) you could have run `represented.metadata.to_hash` as described here https://github.com/jruby/jruby/wiki/CallingJavaFromJRuby#javautilmap-to-hash

maheshp: Reason for the transformation was the ```java.util.Map``` was converted to Array of arrays. I will change it to ```represented.metadata.to_hash```

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/plugins_controller.rb
maheshp: Ok

server/webapp/WEB-INF/rails.new/config/routes.rb
maheshp: Dint know bout it, will change.

server/src/com/thoughtworks/go/server/service/plugins/builder/PluginInfoBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+        pluginExtensionToBuilder.put(JsonBasedPackageRepositoryExtension.EXTENSION_NAME, new PackageViewModelBuilder(pluginManager));
+        pluginExtensionToBuilder.put(JsonBasedTaskExtension.TASK_EXTENSION, new PluggableTaskViewModelBuilder(pluginManager));
+        pluginExtensionToBuilder.put(SCMExtension.EXTENSION_NAME, new SCMViewModelBuilder(pluginManager));
+    }
+
+    public List<PluginInfo> allPluginInfos(String type) {
+        List<PluginInfo> plugins = new ArrayList<>();
+
+        if (type != null) {
+            return pluginExtensionToBuilder.containsKey(type) ? pluginExtensionToBuilder.get(type).allPluginInfos() : plugins;
```
jyotisingh: If the key is invalid(meaning not a known extension type), could we return an error message instead of returning an empty plugin list?

PR #2238, opened 2016-05-30
Added 'pipeline' material support for SPA #1724

maheshp:
* Can fetch Pipelines for a given user from PipelineConfigService
* Added internal API to fetch Pipelines with stages
* Added Dependecy Material model in js
* Added PipelineStage field with autocomplete

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
@@ -34,6 +34,7 @@ import com.thoughtworks.go.util.Node;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.List;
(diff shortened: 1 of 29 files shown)


server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
```diff
@@ -132,6 +136,17 @@ public void updatePipelineConfig(final Username currentUser, final PipelineConfi
         update(currentUser, pipelineConfig, result, updatePipelineConfigCommand);
     }
 
+    public List<Pipeline> pipelinesFor(Username username) {
```
zabil: I don't think we need this method or the objects ```Pipeline```, ```Stage``` etc

There a method for getting pipeline groups of the user.
https://github.com/gocd/gocd/blob/e43774e9f991c2a0e2c82a3e41feeaa01a8efcbe/server/src/com/thoughtworks/go/server/service/SecurityService.java#L168

Can't we just get this from the controller and transform using the representer?

zabil: Check 
https://github.com/gocd/gocd/blob/e43774e9f991c2a0e2c82a3e41feeaa01a8efcbe/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/dashboard_controller.rb

for an example.

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -233,6 +233,10 @@
         resources :command_snippets, only: [:index]
         get 'command_snippets/show', controller: :command_snippets, action: :show, as: :command_snippet
         post :material_test, controller: :material_test, action: :test, as: :material_test
+
+        namespace :internal do
+          resources :pipelines, param: :name, only: [:index]
+        end
```
ketan: Could you move the pluggable tasks into this URL and code namespace too?

ketan: It's defined around line 366.

Issue #2202, opened 2016-05-22
Add REST API to get GoCD version number


grundic:
##### Summary

I would like to get version number of running GoCD server. I found [this](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/models/api_v1/go_latest_version.rb), but looks like it used for automatic updates or something similar. Could you add additional endpoint to get version number?

##### Expected Results
It should return version number (maybe some additional info) of running GoCD server.

arvindsv:
Closing, since #2239 is merged. Documentation PR is https://github.com/gocd/api.go.cd/pull/84.

PR #2239, opened 2016-05-30
Added an endpoint to get gocd server version details (#2202)

GaneshSPatil:


server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -269,6 +269,7 @@
                 /api/feeds/**=ROLE_SUPERVISOR
                 /api/support=ROLE_SUPERVISOR
                 /api/pipelines.xml=ROLE_USER
+                /api/server_version=ROLE_USER
```
ketan: You probably mean to say `IS_AUTHENTICATED_ANONYMOUSLY` so that anonymous users can call this too?

ketan: Also how about simply calling this `/api/version`?

server/webapp/WEB-INF/vm/admin/server_version.txt.vm
```diff
@@ -0,0 +1 @@
+{}
```
ketan: You probably want to populate a default file with actual keys and values set to `unknown` or something similar.

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/server_version_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+require 'spec_helper'
+
+describe ApiV1::ServerVersionController do
+  describe :show do
+    it 'should render the current gocd server version for admins' do
+      version_file = Rails.root.join('..', 'vm', 'admin', 'server_version.txt.vm')
+      file_content = '{"go_version":"16.6.0", "go_build_number":"235", "git_sha":"69ef4921709a84831913d9fa7e750fbf840f213c"}'
+      expected_json = { :_links => {
+                          :self=>{:href=> 'http://test.host/api/server_version' },
```
ketan: Please indent this code properly. The consistent way to test this is like [this](https://github.com/gocd/gocd/blob/1236015adabbfdf88a786f9e0d66aabd071c1c12/server/webapp/WEB-INF/rails.new/spec/presenters/api_v1/user_representer_spec.rb)

ketan: Sorry, I mean this https://github.com/gocd/gocd/blob/1236015adabbfdf88a786f9e0d66aabd071c1c12/server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/users_controller_spec.rb#L33

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/server_version_representer.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    link :doc do
+      'http://api.go.cd/#server_version'
+    end
+
+    property :version, getter: lambda { |options| go_version }
+    property :build_number, getter: lambda {| options| go_build_number }
+    property :git_sha, getter: lambda { |options| git_sha }
+    property :go_full_version, getter: lambda { |options| "#{go_version} (#{go_build_number}-#{git_sha})" }
```
ketan: simply `full_version` instead of `go_full_version`?

buildfile
```diff
@@ -74,6 +74,10 @@ define "cruise" do |project|
   ENV["VERSION_NUMBER"] = VERSION_NUMBER
   ENV["RELEASE_COMMIT"] = RELEASE_COMMIT
 
+  ENV["GO_VERSION"] = GO_VERSION
+  ENV["RELEASE_REVISION"] = "#{RELEASE_REVISION}"
+  ENV["GIT_SHA"] = GIT_SHA
```
zabil: We have a variable called ```RELEASE_VCS_REV``` what's the difference between that and ```GIT_SHA```?

GaneshSPatil: @zabil 
can you tell me where we have used `RELEASE_VCS_REV `?

server/webapp/WEB-INF/vm/admin/server_version.txt.vm
```diff
@@ -0,0 +1 @@
+{"go_version":"16.6.0","go_build_number":"3341", "git_sha":"1236015adabbfdf88a786f9e0d66aabd071c1c12"}
```
zabil: This must be generated only while packaging.
Check-in this file with an empty json

ketan: We're adding a json with required fields, but values set to `unknown`, so the reader knows what fields are needed in the file.

timoa:
Thanks @ketan! Good news for the next releases to call this endpoint rather than the login page and search in the HTML the version label in the bottom ;)

PR #2248, opened 2016-06-01
Improve page load performance across the board.

ketan:
This PR solves 2 problems —

Problem 1
---------

The `ApplicationController#populate_health_messages` filter was called
on every page load, and was re-computing all health state messages.
This computation had n^2 complexity for every health state message that
had to be rendered, this health state was rendered at the bottom of
every page inside go.

We now don't render the errors on page load, but rely on the ajax poll
to eventually render the errors in a few seconds after the page loads

Problem 2
---------

To further reduce the cost of computation, we now pre-compute the health
state messages, and the pipelines that each server health state message
belongs to, via a background thread.

### DIFF
--- a/common/src/com/thoughtworks/go/serverhealth/ServerHealthService.java
+++ b/common/src/com/thoughtworks/go/serverhealth/ServerHealthService.java
@@ -1,39 +1,40 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.serverhealth;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-
 import com.thoughtworks.go.config.CruiseConfig;
+import com.thoughtworks.go.config.CruiseConfigProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 @Service
 public class ServerHealthService {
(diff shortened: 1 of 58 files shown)


common/src/com/thoughtworks/go/serverhealth/ServerHealthService.java
```diff
@@ -72,6 +73,25 @@ public HealthStateType update(ServerHealthState serverHealthState) {
         }
     }
 
+    // called from spring timer
+    public synchronized void onTimer(CruiseConfigProvider provider) {
+        CruiseConfig currentConfig = provider.getCurrentConfig();
+        getAllValidLogs(currentConfig);
+        LOG.info("Recomputing material to pipeline mappings.");
```
arvindsv: Should this be INFO level or DEBUG? I guess it's ok because our default log level for the appender is WARN.

common/src/com/thoughtworks/go/serverhealth/ServerHealthService.java
```diff
@@ -72,6 +73,25 @@ public HealthStateType update(ServerHealthState serverHealthState) {
...
+
+        for (Map.Entry<HealthStateType, ServerHealthState> entry : serverHealth.entrySet()) {
+            erroredPipelines.put(entry.getValue(), entry.getValue().getPipelineNames(currentConfig));
+        }
+        pipelinesWithErrors = erroredPipelines;
+        LOG.info("Done recomputing material to pipeline mappings.");
+    }
+
+    public Set<String> getPipelinesWithErrors(ServerHealthState serverHealthState) {
+        return pipelinesWithErrors.get(serverHealthState);
```
arvindsv: Since this is not synced and not got in one shot, it's possible that stale data will be mixed with new data (while it is being called multiple times from the ERB file). I suppose that's ok?

server/resources/available.toggles
```diff
@@ -10,11 +10,6 @@
          "key": "pipeline_config_single_page_app_key",
          "description": "Enables the pipeline config single page application. Default: off.",
          "value": false
-      },
-      {
-         "key": "better_dashboard_errors_key",
-         "description": "Makes pipeline related errors more visible in the dashboard (https://github.com/gocd/gocd/pull/1842). Default: off.",
```
arvindsv: So much confidence in your performance-problem fixing abilities. :P

server/webapp/WEB-INF/spring-cruise-remoting-servlet.xml
```diff
@@ -82,6 +88,17 @@
         </property>
     </bean>
 
+    <bean name="recomputeMaterialPipelineMappingsThread" class="org.springframework.scheduling.timer.TimerFactoryBean">
+        <property name="scheduledTimerTasks">
+            <list>
+                <bean class="org.springframework.scheduling.timer.ScheduledTimerTask"
+                      p:delay="10000"
```
arvindsv: Is there any reason to make this a property? Part of cruise.properties or whatever that file is.

arvindsv:
Looks good to me.

common/src/com/thoughtworks/go/serverhealth/ServerHealthService.java
ketan: Yes, that's OK.

server/webapp/WEB-INF/spring-cruise-remoting-servlet.xml
ketan: Making it configurable felt like an overkill. @jyotisingh agreed with me on this :)

arvindsv: Fiiine. :)

Issue #2249, opened 2016-06-02
Websocket based communication with Shine enabled does not play well


jyotisingh:
##### Issue Type
- Bug Report

##### Summary

* Enable websockets for agent-server communication
* Enable Shine
* When a job completion causes a stage to get completed, it sends a callback to shine(check [this](https://github.com/gocd/gocd/blob/23aad62d61ebdbb2eeb57cd7d11a9fc7dac462f1/server/src/com/thoughtworks/go/server/service/ScheduleService.java#L527) and follow through [this](https://github.com/gocd/gocd/blob/23aad62d61ebdbb2eeb57cd7d11a9fc7dac462f1/server/src/com/thoughtworks/go/server/service/StageService.java#L284), then [this](https://github.com/gocd/gocd/blob/23aad62d61ebdbb2eeb57cd7d11a9fc7dac462f1/server/src/com/thoughtworks/go/server/service/StageService.java#L233), then [this](https://github.com/gocd/gocd/blob/23aad62d61ebdbb2eeb57cd7d11a9fc7dac462f1/server/src/com/thoughtworks/studios/shine/cruise/BackgroundStageLoader.java#L103))
* For some reason, shine fails to find `GoDateTime` class which is used for the xslt transformation and fails with the following stacktrace:
```
java.lang.RuntimeException: Cannot call method public void com.thoughtworks.go.server.websocket.AgentRemoteSocket#onMessage(java.io.InputStream) with args: [org.eclipse.jetty.websocket.common.message.MessageInputStream]
    at org.eclipse.jetty.websocket.common.events.annotated.CallableMethod.unwrapRuntimeException(CallableMethod.java:93)
    at org.eclipse.jetty.websocket.common.events.annotated.CallableMethod.call(CallableMethod.java:75)
    at org.eclipse.jetty.websocket.common.events.annotated.OptionalSessionCallableMethod.call(OptionalSessionCallableMethod.java:72)
    at org.eclipse.jetty.websocket.common.events.JettyAnnotatedEventDriver$1.run(JettyAnnotatedEventDriver.java:91)
    at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635)
    at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555)
    at java.lang.Thread.run(Thread.java:745)
[...]

PR #2254, opened 2016-06-03
Set a class loader for looking up java extension methods in xslt

zabil:
Checking if this solved #2249


arvindsv:
This constant seems to have been introduced [in 7u79](http://atetric.com/atetric/javadoc/openjdk/jdk/7u79/com/sun/org/apache/xalan/internal/XalanConstants.html). Wasn't [in 7u67](http://atetric.com/atetric/javadoc/openjdk/jdk/7u67/com/sun/org/apache/xalan/internal/XalanConstants.html), which is what I was using. Just FYI.

ketan:
I'm reverting this change, will need to dig deeper. @zabil — would you like to do that, or may I?

Issue #1773, opened 2016-01-07
Pipeline API: no way to delete a pipeline?


eugeneventer:
https://api.go.cd/current/#pipeline-config - there does not seem to be a way to delete a pipeline...

Is this correct or am I missing something?

Thanks!
Eugene.

jyotisingh:
Pipeline deletion is not really supported in Go at the moment - through the UI as well as the API. Well you could go ahead to the config xml and remove the pipeline from there but it works with caveat. Look at the [documentation](https://www.go.cd/documentation/user/current/configuration/managing_pipelines.html#delete-an-existing-pipeline) and [this](https://github.com/gocd/gocd/issues/879) issue that provide more details about what I mean. 
Until #879 is addressed, it doesnot seem right to provide a way to delete the pipeline through the UI/API.

eugeneventer:
@jyotisingh but I'm able to easily delete a pipeline via the UI, via /go/admin/pipelines - there is a 'Delete this pipeline' action icon and link?

jyotisingh:
Sorry, I have been day-dreaming :disappointed: 

Renaming pipeline is not allowed for the reasons I mentioned earlier. And pipeline deletion (even though is allowed) suffers from the same issue. ie  it doesnot delete data from db or filesystem. So, if you create a pipeline with same name in future, then it picks historical data from older pipeline which could be mis-leading for the user. 

But sure I get your point, if it is allowed from UI then why not from API.

jasonolmstead33:
Any movement on this? my team and I would love to star using Go, and the Go API, but would really need a way to delete pipelines, or at least move them to a new group.. through the API

remiego:
We are looking at using go for pipeline management via the api and seems wrong that a pipeline can be created bit not removed. We are generating transient pipelines and need to remove these after they have been created and invoked, obviously this needs to be aware of downstream/upstream pipelines...

yunlzheng:
Is there have any way to delete go pipeline through the API?

zabil:
@yunlzheng not yet

varshavaradarajan:
Delete pipeline API is available at: https://api.go.cd/current/#delete-a-pipeline

arvindsv:
@varshavaradarajan: The doc page just says: "Deletes a pipeline". Do you think it's worth it to mention where it deletes it from? Meaning, that is deletes it only from config and not from DB.

ByteFlinger:
[...]

PR #2255, opened 2016-06-03
Added Pipeline-config DELETE api (#1773)

GaneshSPatil:
Pipelines can only be deleted if they don't have any dependencies

Pipelines CAN NOT be deleted
-- If it is used as a material in some another pipeline (if it has downstream pipelines)
--If is belongs to any environment

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
@@ -21,6 +21,7 @@ import com.thoughtworks.go.config.commands.EntityConfigUpdateCommand;
 import com.thoughtworks.go.config.exceptions.GoConfigInvalidException;
 import com.thoughtworks.go.config.update.ConfigUpdateCheckFailedException;
 import com.thoughtworks.go.config.update.CreatePipelineConfigCommand;
+import com.thoughtworks.go.config.update.DeletePipelineConfigCommand;
 import com.thoughtworks.go.config.update.UpdatePipelineConfigCommand;
 import com.thoughtworks.go.config.remote.ConfigOrigin;
 import com.thoughtworks.go.i18n.LocalizedMessage;
@@ -171,6 +172,14 @@ public class PipelineConfigService implements ConfigChangedListener, Initializer
     public void createPipelineConfig(final Username currentUser, final PipelineConfig pipelineConfig, final LocalizedOperationResult result, final String groupName) {
         CreatePipelineConfigCommand createPipelineConfigCommand = new CreatePipelineConfigCommand(goConfigService, pipelineConfig, currentUser, result, groupName);
         update(currentUser, pipelineConfig, result, createPipelineConfigCommand);
+    }
+
+    public void deletePipelineConfig(final Username currentUser, final PipelineConfig pipelineConfig, final LocalizedOperationResult result) {
+        DeletePipelineConfigCommand deletePipelineConfigCommand = new DeletePipelineConfigCommand(goConfigService, pipelineConfig, currentUser, result);
+        update(currentUser, pipelineConfig, result, deletePipelineConfigCommand);
+        if(result.isSuccessful()) {
+            result.setMessage(LocalizedMessage.string("PIPELINE_DELETE_SUCCESSFUL", pipelineConfig.name()));
+        }
     }
 
     @Override
--- a/server/test/integration/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
+++ b/server/test/integration/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
@@ -45,6 +45,7 @@ import com.thoughtworks.go.util.GoConstants;
 import com.thoughtworks.go.util.SystemEnvironment;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -142,7 +143,7 @@ public class PipelineConfigServiceIntegrationTest {
     }
 
     @Test
-    public void shouldCreatePipelineConfigWhenPipelineGroupDoesExist() throws GitAPIException {
+    public void shouldCreatePipelineConfigWhenPipelineGroupDoesNotExist() throws GitAPIException {
         GoConfigHolder goConfigHolderBeforeUpdate = goConfigDao.loadConfigHolder();
         PipelineConfig downstream = GoConfigMother.createPipelineConfigWithMaterialConfig(UUID.randomUUID().toString(), new DependencyMaterialConfig(pipelineConfig.name(), pipelineConfig.first().name()));
         pipelineConfigService.createPipelineConfig(user, downstream, result, "does-not-exist");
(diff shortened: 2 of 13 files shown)


server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
```diff
@@ -137,6 +138,13 @@ public void createPipelineConfig(final Username currentUser, final PipelineConfi
...
     }
 
+    public void deletePipelineConfig(final Username currentUser, final String pipelineName, final LocalizedOperationResult result) {
+        PipelineConfig pipelineConfig = getPipelineConfig(pipelineName);
+        DeletePipelineConfigCommand deletePipelineConfigCommand = new DeletePipelineConfigCommand(goConfigService, pipelineConfig, currentUser, result);
+        update(currentUser, pipelineConfig, result, deletePipelineConfigCommand);
+        if(result.isSuccessful()) result.setMessage(LocalizedMessage.string("PIPELINE_DELETE_SUCCESSFUL", pipelineName));
+    }
+
     @Override
```
varshavaradarajan: We usually follow the convention of surrounding if by brackets.

server/test/integration/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
```diff
@@ -360,6 +361,32 @@ public void shouldMapErrorsBackToPackageMaterials() throws Exception {
...
+
+        pipelineConfigService.deletePipelineConfig(user, pipelineConfig.name().toString(), result);
+
+        assertTrue(result.isSuccessful());
+        assertThat(goConfigService.getAllPipelineConfigs().size(), is(0));
+        assertFalse(goConfigService.hasPipelineNamed(pipelineConfig.name()));
+    }
+
+    @Test
+    public void shouldNotDeleteThePipelineForUnauthorizedUsers() throws Exception {
```
varshavaradarajan: This test should be where the check is made. In DeletePipelineConfigCommand

server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
```diff
@@ -137,6 +138,13 @@ public void createPipelineConfig(final Username currentUser, final PipelineConfi
         update(currentUser, pipelineConfig, result, createPipelineConfigCommand);
     }
 
+    public void deletePipelineConfig(final Username currentUser, final String pipelineName, final LocalizedOperationResult result) {
+        PipelineConfig pipelineConfig = getPipelineConfig(pipelineName);
+        DeletePipelineConfigCommand deletePipelineConfigCommand = new DeletePipelineConfigCommand(goConfigService, pipelineConfig, currentUser, result);
+        update(currentUser, pipelineConfig, result, deletePipelineConfigCommand);
+        if(result.isSuccessful()) result.setMessage(LocalizedMessage.string("PIPELINE_DELETE_SUCCESSFUL", pipelineName));
```
varshavaradarajan: We follow the convention of surrounding if by brackets so that additional statements made later are not confusing.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
```diff
@@ -18,9 +18,10 @@ module ApiV1
   module Admin
     class PipelinesController < ApiV1::BaseController
       before_action :check_admin_user_and_401
-      before_action :load_pipeline, only: [:show]
+      before_action :load_pipeline, only: [:show, :destroy]
```
varshavaradarajan: Do you need to load the pipeline when in the destroy method you don't need to pass the pipeline to pipeline_config_service?

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/pipelines_controller_spec.rb
```diff
@@ -365,6 +363,65 @@
       end
     end
 
+    describe :destroy do
```
varshavaradarajan: Please add the specs for security. It is already there for show, create and update in pipelines_controller_spec.rb

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/pipelines_controller_spec.rb
```diff
@@ -365,6 +363,65 @@
...
+        can_not_delete_message = com.thoughtworks.go.server.presentation.CanDeleteResult.new(true, LocalizedMessage.string("CANNOT_DELETE_PIPELINE_IN_ENVIRONMENT", @pipeline_name, "foo-environment"))
+        pipelines_can_be_deleted.put(CaseInsensitiveString.new(@pipeline_name), can_not_delete_message)
+        @pipeline_config_service.should_receive(:canDeletePipelines).and_return(pipelines_can_be_deleted)
+
+        put_with_api_header :destroy, name: @pipeline_name
+
+        expect(response.code).to eq("422")
+        expect(actual_response).to eq({ :message => "Cannot delete pipeline 'pipeline1' as it is present in environment 'foo-environment'" })
+      end
+
```
varshavaradarajan: the statement ```@pipeline_config_service.should_receive(:getPipelineConfig).with(@pipeline_name).and_return(@pipeline)``` is present in all specs but one. Consider moving it to a ```before``` block.

varshavaradarajan:
Please send a pull request for the documentation.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
GaneshSPatil: it will raise RecordNotFound Exception if the user tried to delete a pipeline which does not exists

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/pipelines_controller_spec.rb
```diff
@@ -365,6 +363,65 @@
...
+        @pipeline_name = "pipeline1"
+        @pipeline      = PipelineConfigMother.pipelineConfig(@pipeline_name)
+      end
+
+      it "should delete pipeline config for an admin" do
+        @pipeline_config_service.should_receive(:getPipelineConfig).with(@pipeline_name).and_return(@pipeline)
+        pipelines_can_be_deleted = java.util.HashMap.new
+        can_delete_message = com.thoughtworks.go.server.presentation.CanDeleteResult.new(true, LocalizedMessage.string("CAN_DELETE_PIPELINE"))
+        pipelines_can_be_deleted.put(CaseInsensitiveString.new(@pipeline_name), can_delete_message)
+        @pipeline_config_service.should_receive(:canDeletePipelines).and_return(pipelines_can_be_deleted)
```
ketan: How about —

```ruby
pipeline_config_service.should_receive(:canDeletePipelines).and_return({'key' => value})
```

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/pipelines_controller_spec.rb
```diff
@@ -365,6 +363,65 @@
...
+
+        put_with_api_header :destroy, name: @pipeline_name
+
+        expect(response.code).to eq("200")
+        expect(actual_response).to eq({ :message => "Pipeline 'pipeline1' was deleted successfully." })
+      end
+
+      it "should not delete the pipeline config if the pipeline is used as a material in some another pipeline" do
+        @pipeline_config_service.should_receive(:getPipelineConfig).with(@pipeline_name).and_return(@pipeline)
+        pipelines_can_be_deleted = java.util.HashMap.new
```
ketan: How about —

```ruby
pipeline_config_service.should_receive(:canDeletePipelines).and_return({'key' => value})
```

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
varshavaradarajan: Then, isn't it better to remove the getPipelineConfig(pipelineName) from deletePipelineConfig() in PipelineConfigService and directly pass the pipeline config object so that you don't do this twice?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pipelines_controller.rb
```diff
@@ -119,6 +126,18 @@ def check_if_pipeline_by_same_name_already_exists
         end
       end
 
+      def check_for_pipeline_dependency
+        can_be_deleted_message = LocalizedMessage.string("CAN_DELETE_PIPELINE")
+        pipelines_can_be_deleted = pipeline_config_service.canDeletePipelines()
+        actual_message = pipelines_can_be_deleted[CaseInsensitiveString.new(params[:name])].message
+
+        if(!can_be_deleted_message.equals(actual_message))
```
ketan: Ouch, this seems like a dangerous thing to do! relying on a string message to imply deletion is not a great idea.

Consider using `CanDeleteResult#canDelete` instead?

PR #2256, opened 2016-06-03
WIP - 1133 config repo caching and two step validation

jyotisingh:
* Changes in main.xml should not break config-repo partials. handling cache issues etc.

* Adding groups and environments added by partials to the main xml so they could be edited to add authorisation, removing unused methods, and some minor cleanup

* Validating merged config before caching partials.

* Introducing two step validation and save when config-repos are configured

* Handling two step config validations when a config-repo partial changes

* There is just one CachedGoConfig, removed the interface and inlined all calls to cachedFileGoConfig from mergedGoConfig(now called CachedGoConfig). 

Pending:([link](https://github.com/gocd/gocd/pull/2256/files#r66037657))

* Deletion of config-repos. As of now this will cause problems if end-users remove a partial which was being referred to from another partial.

* Surviving a Go server restart: A server restart causes Go to forget all known and valid partials. After the next MDU, the server finds all the known partials. However, if the latest set of known partials are invalid, the valid partial list will be empty. We do not persist this info in db as of now.

* Various config APIs and their call backs: We need to ensure the same two step validation happen for the config APIs too. Also, revisit the listeners to check if we need to add any additional logic to handle remote environments/groups.

**Please merge https://github.com/gocd/functional-tests/pull/74 when you merge this otherwise regression tests would fail.**


config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -287,14 +224,15 @@ private PipelineGroups mergePipelineConfigs() {
...
+            for (PipelineConfigs partPipesConf : this.main.getGroups()) {
                 allPipelineConfigs.add(partPipesConf);
             }
             // then add from each part
             for (PartialConfig part : this.parts) {
-                for(PipelineConfigs partPipesConf : part.getGroups())
-                {
+                for (PipelineConfigs partPipesConf : part.getGroups()) {
+                    for (PipelineConfig pipelineConfig : partPipesConf) {
+                        main.getAllPipelineConfigs().add(pipelineConfig);
```
arvindsv: Because of this line, we have to be sure that this method is called only once, and we have to make sure that it is called on a config which is bare/new (doesn't have pipeline configs from partials). Can this be done elsewhere, or be more explicit? One way is to return a merge result object around here, with both groups and these configs, which can then be explicitly added to the allPipelineConfigs list. Just a thought.

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -1356,6 +1216,16 @@ public boolean canDeletePluggableSCMMaterial(SCM scmConfig) {
     }
 
     @Override
+    public void setPartials(List<PartialConfig> partials) {
+        this.partials = new Cloner().deepClone(partials);
```
arvindsv: I'd consider doing this outside and forcing the caller to make this decision. For instance, maybe this is already cloned (it's not, today). I'd think that [this](https://github.com/gocd/gocd/pull/2256/files#diff-e236b0af07a149fc225404e31015c033R256) is a better place to clone. It's explicit there that it's an attempt at saving. So, you're saying you need it to throw away the clone if it does not work.

config/config-api/src/com/thoughtworks/go/config/merge/MergeEnvironmentConfig.java
```diff
@@ -414,7 +414,7 @@ public boolean hasVariable(String variableName) {
 
     @Override
     public boolean hasSamePipelinesAs(EnvironmentConfig other) {
-        return false;
+        return false; //TODO: jyoti - is this correct?
```
arvindsv: Someone make a decision. :)

config/config-api/src/com/thoughtworks/go/config/preprocessor/ConfigRepoPartialPreprocessor.java
```diff
@@ -0,0 +1,42 @@
...
+    public ConfigRepoPartialPreprocessor() {
+    }
+
+    public void init(PartialsProvider partialsProvider) {
+        this.partialsProvider = partialsProvider;
+    }
+
+    public void process(CruiseConfig cruiseConfig) {
+        if (cruiseConfig.getPartials().isEmpty() && partialsProvider != null) {
+            cruiseConfig.merge(new Cloner().deepClone(partialsProvider.lastPartials()));
```
arvindsv: Shouldn't `setPartials` be called here, somewhere?

config/config-api/test/com/thoughtworks/go/config/MergeCruiseConfigTest.java
```diff
@@ -299,56 +298,22 @@ public void addPipelineWithoutValidation_shouldAddPipelineToMain()
         assertThat(cruiseConfig.hasPipelineNamed(new CaseInsensitiveString("pipe3")), is(true));
 
     }
+
     @Test
-    public void addPipelineWithoutValidation_shouldFailToAddPipelineWhenItExistsInPartialConfig()
```
arvindsv: What does it mean that these tests are removed now? I don't see replacements for all. Does it mean it was wrong, or not needed or that functionality is gone?

server/src/com/thoughtworks/go/config/CachedGoConfig.java
```diff
@@ -18,36 +18,207 @@
...
+    }
+
+    //NOTE: This method is called on a thread from Spring
+    public void onTimer() {
+        this.forceReload();
+    }
+
+    public void forceReload() {
+        LOGGER.debug("Config file (on disk) update check is in queue");
+        synchronized (GoConfigWriteLock.class) {
```
arvindsv: Though it might not be a part of this PR, I was thinking about this. Taking this lock here means that any time the hard disk is checked to see if the config file has changed, all config saves will be blocked. A more ideal approach will be to do the check outside of this sync block and only get into it if a reload is needed. Right? Could help in situations where the file operation is slow.

server/src/com/thoughtworks/go/config/CachedGoConfig.java
```diff
@@ -18,36 +18,207 @@
...
+    }
+
+    private synchronized void saveConfigError(Exception e) {
+        this.lastException = e;
+        ServerHealthState state = ServerHealthState.error(INVALID_CRUISE_CONFIG_XML, GoConfigValidity.invalid(e).errorMessage(), HealthStateType.invalidConfig());
+        serverHealthService.update(state);
+    }
+
+    public CruiseConfig loadForEditing() {
+        return currentConfigForEdit;
```
arvindsv: I guess there's never a time that this is null. But, we do sometimes check for null (like in `currentConfig` below). Why not here? I know `loadConfigIfNull` called from ApplicationInitializer, but if so, why check in `currentConfig`?

server/src/com/thoughtworks/go/config/CachedGoConfig.java
```diff
@@ -18,36 +18,207 @@
...
+        return currentConfig;
+    }
+
+    public void loadConfigIfNull() {
+        if (currentConfig == null || currentConfigForEdit == null || configHolder == null) {
+            forceReload();
+        }
+    }
+
+    public synchronized ConfigSaveState writeWithLock(UpdateConfigCommand updateConfigCommand) {
```
arvindsv: At some point, can we consider replacing all of the synchronized methods in this class with a `synchronized(this)` block and logging before and after it to see how long it was stuck trying to get the log?

arvindsv: Same comment for GoFileConfigDataSource.

server/src/com/thoughtworks/go/config/CachedGoConfig.java
```diff
@@ -18,36 +18,207 @@
...
+            notifyListeners(currentConfig);
+        }
+    }
+
+    public String getFileLocation() {
+        return dataSource.getFileLocation();
+    }
+
+    public synchronized void save(String configFileContent, boolean shouldMigrate) throws Exception {
+        GoConfigHolder newConfigHolder = dataSource.write(configFileContent, shouldMigrate);
```
arvindsv: Why the deprecated `write` method?

server/src/com/thoughtworks/go/server/initializers/ApplicationInitializer.java
```diff
@@ -97,10 +101,17 @@ public void onApplicationEvent(ContextRefreshedEvent contextRefreshedEvent) {
             //config
 
             configCipherUpdater.migrate(); // Should be done before configs get loaded
+            ConfigRepoPartialPreprocessor preprocessor = (ConfigRepoPartialPreprocessor) ListUtil.find(MagicalGoConfigXmlLoader.PREPROCESSORS, new ListUtil.Condition() {
```
arvindsv: Why not delegate this to MagicalGoConfigXmlLoader itself?

server/src/com/thoughtworks/go/config/GoPartialConfig.java
```diff
@@ -12,108 +12,166 @@
...
-        if(removedAny)
-            this.notifyListeners();
+        for (String fingerprint : cachedGoPartials.getFingerprintToLatestValidConfigMap().keySet()) {
+            if (!newConfigRepos.hasMaterialWithFingerprint(fingerprint)) {
+                removed = true;
+                cachedGoPartials.removeValid(fingerprint);
+            }
+        }
+        if (removed) {
+            updateConfig(null, null, null);
```
arvindsv: Umm, this seems a little dangerous. Have you thought through all that this can go into and made sure this is handled properly?

I'm sure you two have. :) But, just checking.

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
jyotisingh: I see. let me revisit this.

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
jyotisingh: makes sense

config/config-api/test/com/thoughtworks/go/config/MergeCruiseConfigTest.java
jyotisingh: These were testing the behaviour of methods such as `addPipeline`, `addPipelineWithoutValidation` etc in `MergeStrategy`, which are deleted as a part of this PR, hence deleting the corresponding tests.

server/src/com/thoughtworks/go/config/CachedGoConfig.java
jyotisingh: Merely moving all methods from `CachedFileGoConfig` to this class "as is". Did not want to change too much as a part of this PR.

server/src/com/thoughtworks/go/config/GoPartialConfig.java
jyotisingh: No, I haven't thought through this yet :)
May be I should have mentioned this in the PR description. There are few things which do not work as one would expect as far as config-repo goes
* Deletion of config-repos. As of now this will cause problems if end-users remove a partial which was being referred to from another partial. 
* Surviving a Go server restart: A server restart causes Go to forget all known and valid partials. After the next MDU, the server finds all the known partials. However, if the latest set of known partials are invalid, the valid partial list will be empty. We do not persist this info in db as of now.
* various config APIs and their call backs: We need to ensure the same two step validation happen for the config APIs too. Also, revisit the listeners to check if we need to add any additional logic to handle remote environments/groups.

jyotisingh: As far as the NullPointerException is concerned, that might not happen with the current code since we check for all those three parameters. But as I mentioned earlier, the implementation is not complete or clean.

server/src/com/thoughtworks/go/config/CachedGoConfig.java
jyotisingh: yes, its not a part of this PR, intentionally else you would never merge my PR :). I can make that change and send another PR for it. I think if [this test](https://github.com/gocd/gocd/blob/master/server/test/integration/com/thoughtworks/go/server/service/ConfigSaveDeadlockDetectionIntegrationTest.java) passes, we are good. What do you think?

server/src/com/thoughtworks/go/config/CachedGoConfig.java
jyotisingh: Good catch, I was just copying over these methods from `CachedFileGoConfig` but missed copying this one

server/src/com/thoughtworks/go/config/CachedGoConfig.java
jyotisingh: Can club this with the PR for https://github.com/gocd/gocd/pull/2256/files#r66040671

server/src/com/thoughtworks/go/config/CachedGoConfig.java
jyotisingh: There are a whole lot of tests which use this method. Ideally I would like to delete this instead of changing the implementation.

config/config-api/src/com/thoughtworks/go/config/merge/MergeEnvironmentConfig.java
jyotisingh: Spoke to @tomzo. He would be updating the implementation of this method as a part of #2266

server/src/com/thoughtworks/go/config/CachedGoConfig.java
arvindsv: Sounds good. I'm not merging this. :P @tomzo can, now. :)

server/src/com/thoughtworks/go/config/GoPartialConfig.java
arvindsv: Ok, then we should think about those soon. I think we've talked about surviving a restart and decided that it's impossible to do (what if the repo itself is gone?). The other two, I'd expect, can be solved in some way.

tomzo: > Deletion of config-repos. As of now this will cause problems if end-users remove a partial which was being referred to from another partial. 

@jyotisingh If by deletion you mean physically removing config repo and making it impossible to checkout, then there is nothing on Go side that we can do. But I think you mean removal of the `<config-repo>` element from XML. I had some thoughts on this:

If user is trying to remove partial which is referenced from other partials, then we can forbit this operation, just like we have forbidden renaming local stage which is remotely referenced.
The only operation that should pass is when user removes a bunch of dependent partials together.

config/config-api/src/com/thoughtworks/go/config/preprocessor/ConfigRepoPartialPreprocessor.java
jyotisingh: So, GoFileConfigDataSource#internalLoad (which is called from the other config save workflows such as editing config file directly) invokes validation of config, however it does not do the two step validation. So, `cruiseConfig.getPartials()` would be empty in those workflows. For now, the merged config would contain main+valid-partials alone for these flows. This needs to be changed though. I will be doing it as a part of the next PR which takes care of the other scenarios that I mentioned [here](https://github.com/gocd/gocd/pull/2256/files#r66037657)

tomzo:
One thing that is broken here is that config repo error messages are displayed twice.
```
Invalid Merged Configuration [Jun-13 10:13:16]
Stage with name 'up42_stage' does not exist on pipeline 'up42', it is being referred to from pipeline 'pipe-ref-main' (http://gitlab.ai-traders.com/tomzo/gocd-example-xmlconfig.git at 4085d1db8b18debcdb326998ab8dae3aa13e1c10)
Invalid Merged Configuration [Jun-13 10:13:16]
Stage with name 'up42_stage' does not exist on pipeline 'up42', it is being referred to from pipeline 'pipe-ref-main' (http://gitlab.ai-traders.com/tomzo/gocd-example-xmlconfig.git at 4085d1db8b18debcdb326998ab8dae3aa13e1c10) - Config-Repo: http://gitlab.ai-traders.com/tomzo/gocd-example-xmlconfig.git at 4085d1db8b18debcdb326998ab8dae3aa13e1c10
```

PR #2258, opened 2016-06-03
#000 - Add debug information for flaky test (processCancelBuildCommandBuild)

arvindsv:
Trying to find out why [this failure](https://build.go.cd/go/tab/build/detail/build-linux/944/build-non-server/1/agent#tab-tests) keeps happening intermittently. This extra logging information might give a clue.

It looks like, sometimes, the test build command fails, rather than running the "sleep" and getting cancelled. So, instead of [this](https://github.com/gocd/gocd/blob/5feec49c4413035f71e8c0646502ee85287ad5d5/common/src/com/thoughtworks/go/buildsession/BuildSession.java#L170-L173) happening, [this](https://github.com/gocd/gocd/blob/5feec49c4413035f71e8c0646502ee85287ad5d5/common/src/com/thoughtworks/go/buildsession/BuildSession.java#L161-L166) or [this](https://github.com/gocd/gocd/blob/5feec49c4413035f71e8c0646502ee85287ad5d5/common/src/com/thoughtworks/go/buildsession/BuildSession.java#L170-L173) seems to happen. The extra logging *might* help pinpoint it, but not sure.

### DIFF
--- a/agent/test/unit/com/thoughtworks/go/agent/AgentControllerTest.java
+++ b/agent/test/unit/com/thoughtworks/go/agent/AgentControllerTest.java
@@ -35,6 +35,7 @@ import com.thoughtworks.go.remote.work.Work;
 import com.thoughtworks.go.server.service.AgentBuildingInfo;
 import com.thoughtworks.go.server.service.AgentRuntimeInfo;
 import com.thoughtworks.go.util.HttpService;
+import com.thoughtworks.go.util.LogFixture;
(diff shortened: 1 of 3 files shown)


ketan:
Does this need to be merged in,  or are you going to trigger this commit on the PR  pipeline?

arvindsv:
Merged in, please. Since it doesn't always fail regularly, I'd like to have it in there, just in case it fails sometimes.

Issue #2259, opened 2016-06-04
Pipeline Config API - Environment variable 


varshavaradarajan:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: `16.6`
* JAVA Version: `1.8`
* OS: `OS independent`
* Browser vendor and version (if relevant): `N/A`

##### Steps to Reproduce

1. Use the pipeline config API to Specify the environment variable like below:
```
"environment_variables": [
    {
      "name": "USERNAME",
      "encrypted_value": "cipherText",
      "secure": false
    }
]
```
##### Expected Results

A validation error like, "Encrypted value can only be specified when secure is true."

##### Actual Results

Pipeline config gets saved. The config xml -
```
<environmentvariables>
    <variable name="USERNAME">
        <value />
    </variable>
</environmentvariables>
```

varshavaradarajan:
Fixed as part of 16.6.

PR #2260, opened 2016-06-04
Specified validation error when encrypted value is given when secure is false. #2259

varshavaradarajan:


varshavaradarajan:
#2259

config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
```diff
@@ -204,9 +205,16 @@ public boolean hasName(String variableName) {
 
     public boolean validateTree(ValidationContext validationContext) {
         validate(validationContext);
+        validateEncryptedValue(validationContext);
         return errors().isEmpty();
     }
 
+    public void validateEncryptedValue(ValidationContext validationContext) {
```
ketan: Does this need to be public?

ketan:
LGTM, can you trigger the PR build?

ketan:
@varshavaradarajan — I'm merging this in, if you see it being problematic we can pair on fixing it, or roll it back.

PR #2265, opened 2016-06-05
1133 rails ui

tomzo:
"Rails only" changes to display only local elements of configurations.
This is non-functional, mostly updates in views. A few changes in java to split remote from local config.

I have extracted these commits from #1810, there will be a second PR on top of #2256

@jyotisingh I think you saw these all changes in this PR already. If you are fine with it, then I'd like it to be merged first. Then we can discuss #2256 and my second PR #2266


config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -414,8 +421,17 @@ private void verifyUniqueNameInParts(PipelineConfig pipelineConfig) {
 
         @Override
         public void update(String groupName, String pipelineName, PipelineConfig pipeline) {
-            // this was called only from tests
-            throw bomb("Cannot update pipeline group in merged configuration");
+            if (groups.isEmpty()) {
```
jyotisingh: Why do we need this to be implemented for `MergeStrategy`? As a part of https://github.com/gocd/gocd/pull/2256, I am getting rid of the update method on MergeStrategy. Am I missing any scenario?

config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
```diff
@@ -339,4 +348,18 @@ public void setEntityId(Long entityId) {
...
     }
+
+    @Override
+    public ConfigOrigin getOrigin() {
+        return origin;
+    }
+
+    public boolean isRemote()
+    {
+        return origin != null && !origin.isLocal();
```
jyotisingh: There are similar checks implemented at few other places. Does it make sense to implement a null object pattern here like `NullOrigin` instead of checking for not-null and then for `isLocal()` or `isRemote` everywhere?

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
tomzo: Right, this should be removed. I checked that this is not needed even in #2266

config/config-api/src/com/thoughtworks/go/config/merge/MergeEnvironmentConfig.java
```diff
@@ -526,6 +335,56 @@ public EnvironmentVariablesConfig getSecureVariables() {
...
+                return  part;
+        }
+        return  null;
+    }
+
+    @Override
+    public boolean isLocal() {
+        for(EnvironmentConfig part : this)
+        {
+            if(part.isLocal())
```
jyotisingh: should it be `!part.isLocal()` instead of `part.isLocal()`?

config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
tomzo: So you mean we could have something like
```
private ConfigOrigin origin = new NullOrigin();
```
in all classes that have `origin` field?
I think it might not work in all cases, e.g. magical loader instantiates elements by reflection, it would be null then. But we can try. Do you want a commit added in this PR or should we refactor with your changes too?

config/config-api/src/com/thoughtworks/go/config/merge/MergeEnvironmentConfig.java
tomzo: It should. I will fix that.

config/config-api/src/com/thoughtworks/go/config/EnvironmentVariableConfig.java
tomzo: How about we make static methods which accept null?
```
boolean isLocal(ConfigOrigin origin)
boolean isRemote(ConfigOrigin origin)
```

jyotisingh: > So you mean we could have something like
> ```
> private ConfigOrigin origin = new NullOrigin();
> ```

Yes, thats what I meant. Since the initialisation would be at the field level, magical loader will not make it null, unless there is an explicit call to make it null or change the value. 

>Do you want a commit added in this PR or should we refactor with your changes too?

Fine with either. If you do not want to make the change now, its fine. I will make the change in my PR if you are ok with it.

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -191,9 +191,29 @@ private void editEnvironments(EditEnvironments edit, EnvironmentConfig config, U
...
+    public List<EnvironmentPipelineModel> getAllLocalPipelinesForUser(Username user) {
+        List<PipelineConfig> pipelineConfigs = goConfigService.getAllLocalPipelineConfigs();
+        return getAllPipelinesForUser(user, pipelineConfigs);
+    }
+    public List<EnvironmentPipelineModel> getAllRemotePipelinesForUserInEnvironment(Username user,EnvironmentConfig environment) {
+        List<EnvironmentPipelineModel> pipelines = new ArrayList<EnvironmentPipelineModel>();
+        for (EnvironmentPipelineConfig pipelineConfig : environment.getRemotePipelines()) {
+            String pipelineName = CaseInsensitiveString.str(pipelineConfig.getName());
+            if (securityService.hasViewPermissionForPipeline(user, pipelineName)) {
+                if (environment != null) {
```
jyotisingh: Why do we need this null check here? If `environment` was `null`, wouldn't [this line ](https://github.com/gocd/gocd/pull/2265/files#diff-cf566ce8a9594182b030044a7a8398dbR200) throw an exception already?

server/webapp/WEB-INF/rails.new/app/views/admin/pipeline_groups/index.html.erb
```diff
@@ -46,20 +46,33 @@
                     <tbody>
                     <% group.each do |pipeline| %>
                         <tr class="pipeline">
-                            <td class="name"><%= link_to pipeline.name(), pipeline_edit_path(:pipeline_name => pipeline.name(), :current_tab => "general"), :class => 'wrapped_word' %></td>
+                            <td class="name">
+															<% if pipeline.isLocal() %>
+															  <%= link_to pipeline.name(), pipeline_edit_path(:pipeline_name => pipeline.name(), :current_tab => "general"), :class => 'wrapped_word' %>
+															<% else %>
+																<%= pipeline.name() %>
```
jyotisingh: Please add `:class => 'wrapped_word'` for this too just in case pipeline name is too long. Also, the formatting a bit off :)

jyotisingh:
Few issues that we found while testing this out:
1) Added a config repo with a few groups and pipelines in them. The pipelines show up as expected on dashboard, and admin pages. The actions on pipeline are disabled as expected. The `edit` link for the pipeline group is enabled. I click on the `edit` link for one of the groups defined locally and try adding a user to `User Permissions` section and hit on save, it throws up the following error:
```
GoFileConfigDataSource:259 - Configuration file is invalid: Authorization can be defined only in configuration file
com.thoughtworks.go.config.exceptions.GoConfigInvalidException: Authorization can be defined only in configuration file
	at com.thoughtworks.go.config.MagicalGoConfigXmlLoader.validateCruiseConfig(MagicalGoConfigXmlLoader.java:121)
	at com.thoughtworks.go.config.MagicalGoConfigXmlLoader.preprocessAndValidate(MagicalGoConfigXmlLoader.java:99)
	at com.thoughtworks.go.config.MagicalGoConfigXmlWriter.write(MagicalGoConfigXmlWriter.java:75)
	at com.thoughtworks.go.config.GoFileConfigDataSource.configAsXml(GoFileConfigDataSource.java:352)
	at com.thoughtworks.go.config.GoFileConfigDataSource.getModifiedConfig(GoFileConfigDataSource.java:283)
	at com.thoughtworks.go.config.GoFileConfigDataSource.writeWithLock(GoFileConfigDataSource.java:249)
	at com.thoughtworks.go.config.CachedFileGoConfig.writeWithLock(CachedFileGoConfig.java:143)
	at com.thoughtworks.go.config.MergedGoConfig.writeWithLock(MergedGoConfig.java:155)
	at com.thoughtworks.go.config.GoConfigDao.updateConfig(GoConfigDao.java:111)
	at com.thoughtworks.go.server.service.GoConfigService.updateConfig(GoConfigService.java:235)
	at com.thoughtworks.go.server.service.GoConfigService.updateConfigFromUI(GoConfigService.java:283)
```

2) I had a group named 'third' defined remotely. Added a group by the same name in mail xml through config xml, and defined authorization there. Now if I head over to the pipeline group edit page, the following exception gets thrown:
```
2016-06-07 11:08:35,728 FATAL [qtp34800122-22] Rails:? -
ActionView::Template::Error (Not implemented):
     8:
     9:             <div class="fieldset">
    10:                 <%= md5_field %>
    11:                 <%== required_label(f, com.thoughtworks.go.config.PipelineConfigs::GROUP, l.string("PIPELINE_GROUP_NAME")) -%>
    12:                 <%= f.text_field com.thoughtworks.go.config.PipelineConfigs::GROUP, {:class => "form_input", :readonly => @group.hasRemoteParts()} -%>
    13:                 <%= error_message_on(@group, com.thoughtworks.go.config.PipelineConfigs::GROUP, :css_class => "form_error") %>
    14:                 <div class="clear"></div>
  com.thoughtworks.go.config.merge.MergePipelineConfigs.getOrigin(com/thoughtworks/go/config/merge/MergePipelineConfigs.java:150)
  com.thoughtworks.go.config.merge.MergePipelineConfigs.hasRemoteParts(com/thoughtworks/go/config/merge/MergePipelineConfigs.java:186)
  java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:606)
  RUBY._app_views_admin_pipeline_groups_edit_html_erb___1850727316_10514(app/views/admin/pipeline_groups/edit.html.erb:11)
```

3) Move pipeline to another group feature is not working as expected. Define a pipeline in main xml, head to pipelines admin page, click on "move to" from the dropdown, select a remote group name. The user is displayed a success message, but the pipeline is no where in the config xml - it got deleted. Verified that this wasn't one of those caching issues.

4) Unable to add agents or environment variables to an environment defined locally or remotely. A success message show up but the changes are lost. The reason for this seems to be [this](https://github.com/gocd/gocd/blob/c1ec6a072223aff93fb58dd9937542ad108902ea/config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java#L77) line 
```
configForEdit = configForEdit.getLocal();
```
which replaces the config with a stale version returned by `getLocal()`.

arvindsv:
@jyotisingh: About [this](https://github.com/gocd/gocd/pull/2265#issuecomment-224174912), I'm fine with the icon. I'd just want to make sure that upon mouse-over, it shows a meaningful message. I think it already has a title attribute and does that.

jyotisingh:
Looks like these issues have been addressed as part of https://github.com/gocd/gocd/pull/2266#issuecomment-224256031. So, merging this one.

jyotisingh:
@arvindsv regarding [this comment](https://github.com/gocd/gocd/pull/2265#issuecomment-224267099) - yes it shows a a proper error message.

PR #2267, opened 2016-06-06
Plugins API used to list and show pluggable tasks #1727

maheshp:
* Added plugins model, which is populated using the Plugin API
 * Pluggable Tasks are built from plugins


maheshp:
Changing the services to fetch plugins, closing for now.

Issue #695, opened 2014-11-03
Length of Admin Email for GO Server


dave-ingham-tlr:
I am establishing the "Email Notification" section of the Server Configuration for the GO server via the web user interface. 

I have found that the Admin Email address appears to be truncated at 65 characters. The address that I need to supply is 80 characters long and is therefore being rejected as the truncated address is not valid. 

Is this a hard limit to the length or can it be increased?

arvindsv:
Cleanup: Was done. Closing.

PR #2269, opened 2016-06-06
#695 Fixing admin email length to fit RFC 3696, Errata 1003

tedops:
According to the IETF, [RFC 3696](https://tools.ietf.org/html/rfc3696) in [Errata 1003](https://www.rfc-editor.org/errata_search.php?rfc=3696&eid=1003) states that email address lengths are to be 254 octets. This is a fix to both follow that standard and fix the root of [issue #695](https://github.com/gocd/gocd/issues/695).


ketan:
Would we not "fix" this in the XSD as well?

PR #2274, opened 2016-06-07
No checkout flag is specified only for server side clone.

varshavaradarajan:
#2152

### DIFF
--- a/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
+++ b/common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
@@ -235,7 +235,13 @@ public class GitMaterial extends ScmMaterial {
                 });
             }
             int cloneDepth = shallowClone ? preferredCloneDepth : Integer.MAX_VALUE;
-            int returnValue = gitCommand.cloneFrom(outputStreamConsumer, url.forCommandline(), cloneDepth);
+            int returnValue;
+            if(executionContext.isServer()) {
+                returnValue = gitCommand.cloneWithNoCheckout(outputStreamConsumer, url.forCommandline());
+            }
+            else {
+                returnValue = gitCommand.clone(outputStreamConsumer, url.forCommandline(), cloneDepth);
+            }
             bombIfFailedToRunCommandLine(returnValue, "Failed to run git clone command");
         }
         return gitCommand;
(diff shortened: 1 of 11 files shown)


common/src/com/thoughtworks/go/config/materials/git/GitMaterial.java
```diff
@@ -235,7 +235,13 @@ public void afterCompletion(int status) {
                 });
             }
             int cloneDepth = shallowClone ? preferredCloneDepth : Integer.MAX_VALUE;
-            int returnValue = gitCommand.cloneFrom(outputStreamConsumer, url.forCommandline(), cloneDepth);
+            int returnValue;
+            if(executionContext.isServer()) {
+                returnValue = gitCommand.serverSideClone(outputStreamConsumer, url.forCommandline());
```
tomzo: Hi @varshavaradarajan 
I am wondering if there is any way that checkout on server can be still executed? There is `ConfigMaterialUpdater` which relies on `MaterialPoller.checkout` to actually make a clean checkout of git repository. The path starts around [here](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/materials/ConfigMaterialUpdater.java#L114) and IIRC it eventually calls `updateTo` which it seems to me, now would never execute a full checkout if running on server side.

varshavaradarajan: Hi @tomzo, 

Previously, the ```--no-checkout``` flag was there on both the server and the agent side. When ```updateTo``` was called from the agent side, the following are executed:
```
git fetch origin --prune
GitCommand.resetWorkingDir()
```
The above would update the working copy to the given revision, which is the behaviour of ```git checkout``` 

So, when you call ```MaterialPoller.checkout``` from ```ConfigMaterialUpdater```, the working copy on the server side does get updated to the revision that you specify.

zabil: > I am wondering if there is any way that checkout on server can be still executed?

Yes, If the material is a configuration repository. So this feature will only work for materials that don't store configuration.

common/src/com/thoughtworks/go/domain/materials/git/GitCommand.java
```diff
@@ -54,17 +54,22 @@ public GitCommand(String materialFingerprint, File workingDir, String branch, bo
         this.environment = environment;
     }
 
-    public int cloneFrom(ProcessOutputStreamConsumer outputStreamConsumer, String url) {
-        return cloneFrom(outputStreamConsumer, url, Integer.MAX_VALUE);
+    public int serverSideClone(ProcessOutputStreamConsumer outputStreamConsumer, String url) {
```
zabil: call this method ```cloneWithNoCheckout```

Issue #1913, opened 2016-02-10
Pipeline Config API: Validation errors missing


rajiesh:
Validation errors are missing for below scenarios,

1. When the config xml do not have any pipeline group authorization specified and when user create or update a pipeline with stage approval defined as,

```
"approval": {
          "type": "manual",
          "authorization": {
            "users": ["user"],
          }
        }
```

The API return 422(validation error) but the response body do not have any specific error on the invalid role which is specified

2. Create or update a pipeline with this request json which has a fetch task with non existing stage or job name, will return 422 (valdiation error) as expected but specific error on the fetch task is missing,

```
{
  "group": "group1",
  "pipeline": {
    "label_template": "${COUNT}",
    "enable_pipeline_locking": false,
    "name": "DownStream",
    "template": null,
    "parameters": [
      {
        "name": "tests",
        "value": "tests"
      }
    ],
    "environment_variables": null
    ],
    "materials": [
      {
        "type": "dependency",
        "attributes": {
          "name": "dependency",
          "pipeline": "upstream",
          "stage": "defaultStage",
          "auto_update": true
        }
      }
    ],
    "stages": [
      {
        "name": "downstream",
        "fetch_materials": true,
        "clean_working_directory": false,
        "never_cleanup_artifacts": false,
        "approval": {
          "type": "success",
          "authorization": null
        },
        "environment_variables": [
          {
          "secure": false,
          "name": "GO_VERSION",
          "value": "16.1.0"
          }
        ],
        "jobs": [
          {
            "name": "tests",
            "run_instance_count": 1,
            "timeout": null,
            "environment_variables": null
            ],
            "resources": [
              "maven",
              "linux"
            ],
            "tasks": [
              {
                "type": "exec",
                "attributes": {
                  "run_if": [
                    "passed"
                  ],
                  "command": "mvn",
                  "arguments": [
                    "--version"
                  ],
                  "working_directory": "#{tests}"
                }
              },
              {
                "type": "fetch",
                "attributes": {
                  "run_if": [
                    "passed"
                  ],
                  "pipeline": "upstream",
                  "stage": "non-existing-stage",
                  "job": "non-existing-job",
                  "is_source_a_file": false,
                  "source": "result",
                  "destination": "test"
                }
              }
            ],
            "tabs": null,
            "artifacts": null,
            "properties": null
          }
        ]
      }
    ],
    "tracking_tool": null,
    "timer": null
  }
}
```

PR #2278, opened 2016-06-08
Added PipelineConfigErrorCopier (#1913)

GaneshSPatil:
Added PipelineConfigErrorCopier which copies the error from preprocessed pipeline-config to the user provided pipeline-config


arvindsv:
@GaneshSPatil Just a quick thought: One of the differences between a preprocessed config with a "non-preprocessed" one is that templates are inlined. Currently, I think templates are not supported in the pipeline config API (right?) and so, it might be ok. But, if templates are supported at some point, this copier will probably not do the right thing, I think. I mean, it might copy over job and stage errors which might not make sense at the pipeline level (since they're defined in the template).

PR #2282, opened 2016-06-08
Allow turning on web request logging by tweaking `log4j.properties`

ketan:
RequestLogHandler is added into the handler chain by default, however
since request logging is a CPU intensive operation, we short circuit it
by disabling logging if the log level is set to `WARN` or higher level.

### DIFF
--- a/cruise-modules.rb
+++ b/cruise-modules.rb
@@ -202,6 +202,7 @@ define "cruise:server", :layout => server_layout("server") do
     include_fileset_from_target(jar, 'jetty9', "**/Jetty9ServletHelper*.class")
     include_fileset_from_target(jar, 'jetty9', "**/Jetty9Request.class")
     include_fileset_from_target(jar, 'jetty9', "**/Jetty9Response.class")
+    include_fileset_from_target(jar, 'jetty9', "**/Slf4jRequestLogger.class")
     # # ---- Jetty 9 end ---
 
     include_fileset_from_target(jar, 'common', "**/SubprocessLogger*.class")
--- a/jetty9/src/com/thoughtworks/go/server/logging/Slf4jRequestLogger.java
+++ b/jetty9/src/com/thoughtworks/go/server/logging/Slf4jRequestLogger.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.logging;
+
+import org.eclipse.jetty.server.AbstractNCSARequestLog;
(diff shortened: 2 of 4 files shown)


GaneshSPatil:
```java
package com.thoughtworks.go.server.logging;

import org.eclipse.jetty.server.HttpChannelState;
import org.eclipse.jetty.server.Request;
import org.eclipse.jetty.server.Response;
import org.eclipse.jetty.server.handler.RequestLogHandler;
import org.junit.Test;
import org.springframework.mock.web.MockHttpServletRequest;

import javax.servlet.DispatcherType;

public class Slf4jRequestLoggerTest {

    @Test
    public void name() throws Exception {
        RequestLogHandler requestLogHandler = new RequestLogHandler();
        Slf4jRequestLogger logger = new Slf4jRequestLogger();
        logger.setExtended(false);
        logger.setLogCookies(false);
        logger.setLogLatency(true);

        String[] paths = new String[]{"/go/assets/*", "/go/server/messages.json"};
        logger.setIgnorePaths(paths);
        requestLogHandler.setRequestLog(logger);


        long startTime = System.currentTimeMillis();
        int no_of_req = 1000000;
        for (int i = 0; i < no_of_req; i++) {
            Request baseRequest = new Request(null, null){
                @Override
                public HttpChannelState getHttpChannelState() {
                    return new HttpChannelState(null){
                        @Override
                        public boolean isAsync() {
                            return false;
                        }
                    };
                }
            };
            baseRequest.setDispatcherType(DispatcherType.REQUEST);
            MockHttpServletRequest request = new MockHttpServletRequest(null);
            Response response = new Response(null, null);
            requestLogHandler.handle("foo", baseRequest, request, response);
        }
        long endTime = System.currentTimeMillis();

        double total = endTime - startTime;
        System.out.println("difference " + total +"ms");
        System.out.println("per 1000 request "+ total/no_of_req * 1000);
    }
}
```

GaneshSPatil:
Output :

```
difference 2786.0ms
per 1000 request 2.786
```

GaneshSPatil:
The code above is a test case to demonstrate the overhead cost of the new handler when logging is turned off.

It seems like the overhead cost is 2.786ms per 1000 requests, which is NOT difference that is noticeable for the benefit that this PR brings — to turn on logging at runtime, when the server is actually slow (instead of rebooting the server and waiting a few more hours for the server to slow down)

maheshp:
Did a apache benchmark with a 1,000,000 requests and results below.
**With Request Logging off**
```
Document Path:          /go/assets/cruise-1592088ba651470e554a54371b6be6b1336462c2186e74fb24f54f177377b538.ico
Document Length:        1150 bytes

Concurrency Level:      4
Time taken for tests:   51.755 seconds
Complete requests:      1000000
Failed requests:        0
Keep-Alive requests:    1000000
Total transferred:      1320000000 bytes
HTML transferred:       1150000000 bytes
Requests per second:    19321.62 [#/sec] (mean)
Time per request:       0.207 [ms] (mean)
Time per request:       0.052 [ms] (mean, across all concurrent requests)
Transfer rate:          24906.78 [Kbytes/sec] received
```
**With Request Logging on**
```
Document Path:          /go/assets/cruise-1592088ba651470e554a54371b6be6b1336462c2186e74fb24f54f177377b538.ico
Document Length:        1150 bytes

Concurrency Level:      4
Time taken for tests:   68.305 seconds
Complete requests:      1000000
Failed requests:        0
Keep-Alive requests:    1000000
Total transferred:      1320000000 bytes
HTML transferred:       1150000000 bytes
Requests per second:    14640.27 [#/sec] (mean)
Time per request:       0.273 [ms] (mean)
Time per request:       0.068 [ms] (mean, across all concurrent requests)
Transfer rate:          18872.23 [Kbytes/sec] received
```

Issue #2140, opened 2016-04-27
/go/api/support should be available as JSON and with configurable scope


lenucksi:
##### Issue Type
- Feature enhancement

##### Summary
The /go/api/support endpoint outputs a lot of very valuable information. 
As a user operating a Go server I would like to be able to regularly ingest this information into my monitoring system.
For this I would like to be able to directly extract the desired information, ideally as JSON, without having to parse the natural language format it is delivered in now.
To be able to reduce the load of the server I would like to be able to specify which sections are included in the output using a parameter.

ketan:
Related PR #2283

ketan:
@GaneshSPatil — see https://github.com/gocd/gocd/issues/2285#issuecomment-225477870

PR #2283, opened 2016-06-09
Converted api/support page server-information into JSON format (#2140)

GaneshSPatil:
`/api/support`      => gives information in `JSON` format
`/api/support.txt` => gives information in `PLAIN TEXT` format


ketan:
@rajiesh — when you have time, can you do some random sampling of the old and new inputs to make sure we've not missed anything?

PR #2298, opened 2016-06-14
Adding validation of config repo partials as a part of all config save flows

jyotisingh:
As of this commit, new Entity config APIs will validate against the valid-partials alone.
Throwing away PipelineConfigurationCache as the based on the partials configured, and changes done to it the cache will have to keep changing. Instead of caching this info, the required info is computed by PipelineConfigSaveValidationContext itself which would always have the right version of the cruiseConfig


server/src/com/thoughtworks/go/config/GoFileConfigDataSource.java
```diff
@@ -332,14 +349,21 @@ private String getMergedConfig(NoOverwriteUpdateConfigCommand noOverwriteCommand
...
 
         String mergedConfigXml = configRepository.getConfigMergedWithLatestRevision(configRevision, oldMd5);
-        validateMergedXML(mergedConfigXml, latestMd5);
+        validateMergedXML(mergedConfigXml, latestMd5, partials);
+        cachedGoPartials.markAsValid(modifiedConfig.getPartials());
         return mergedConfigXml;
     }
 
-    private CruiseConfig validateMergedXML(String mergedConfigXml, String latestMd5) throws Exception {
+    //TODO: jyoti - write a test for a case with config-repo post-merge validation failing
```
tomzo: I think that is in `shouldNotAllowAGitMergeOfConcurrentChangesIfTheChangeCausesMergedPartialsToBecomeInvalid`?

jyotisingh: No, that test takes care of pre-merge validation. There is a second round of validation that is performed after a git-merge is done on the config (post-merge validation).

arvindsv:
* Why all the change from CruiseConfig to BasicCruiseConfig?

* I guess the PipelineConfigurationCache didn't work out so well because every config which is being changed is different, and it cannot be cached? I see that [the commit message](https://github.com/gocd/gocd/pull/2298/commits/fb8402bf90c71bfedb964cec47cf5a04472c95dc) says as much.
  
  However, I hope the PipelineConfigSaveValidationContext is long-lived enough in that flow that the cache is not calculated multiple times. Otherwise, it will be expensive.

* The method getPipelinesWithDependencyMaterials doesn't seem to be doing the right thing any more, right? Currently, the set it returns is just the full list of all pipelines in the config. The earlier code used to do something using PipelineConfigMap, which *might* have done what its name suggests.


* [This](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R263) seems wrong: Isn't this marking last known partials as valid, without them having gone through the check in the line after this? What happens if they fail that validation?

  I formatted that method like this, to help me understand it better:

  ```java
  public synchronized EntityConfigSaveResult writeEntityWithLock(EntityConfigUpdateCommand updatingCommand, GoConfigHolder configHolder, Username currentUser) {
      CruiseConfig modifiedConfig = cloner.deepClone(configHolder.configForEdit);

      try {
          updatingCommand.update(modifiedConfig);
      } catch (Exception e) {
          bomb(e);
      }

      List<PartialConfig> lastValidPartials = cachedGoPartials.lastValidPartials();
      List<PartialConfig> lastKnownPartials = cachedGoPartials.lastKnownPartials();

      if (lastKnownPartials.isEmpty() || areKnownPartialsSameAsValidPartials(lastKnownPartials, lastValidPartials)) {
          return trySavingEntity(updatingCommand, currentUser, modifiedConfig, lastValidPartials);
      }

      try {
          return trySavingEntity(updatingCommand, currentUser, modifiedConfig, lastValidPartials);
      } catch (GoConfigInvalidException e) {
          StringBuilder errorMessageBuilder = new StringBuilder();
          try {
              String message = String.format(
                      "Merged update operation failed on VALID %s partials. Falling back to using LAST KNOWN %s partials. Exception message was: [%s %s]",
                      lastValidPartials.size(), lastKnownPartials.size(), e.getMessage(), e.getAllErrorMessages());
              errorMessageBuilder.append(message);
              LOGGER.warn(message, e);

              updatingCommand.clearErrors();

              List<PartialConfig> partials = lastKnownPartials;
              modifiedConfig.setPartials(partials);
              String configAsXml = configAsXml(modifiedConfig, false);
              cachedGoPartials.markAsValid(partials);

              GoConfigHolder holder = internalLoad(configAsXml, new ConfigModifyingUser(currentUser.getUsername().toString()), partials);
              LOGGER.info(String.format("Update operation on merged configuration succeeded with %s KNOWN partials. Now there are %s LAST KNOWN partials",
                      lastKnownPartials.size(), cachedGoPartials.lastKnownPartials().size()));
              return new EntityConfigSaveResult(holder.config, holder);
          } catch (Exception exceptionDuringFallbackValidation) {
              String message = String.format(
                      "Merged config update operation failed using fallback LAST KNOWN %s partials. Exception message was: %s",
                      lastKnownPartials.size(), exceptionDuringFallbackValidation.getMessage());
              LOGGER.warn(message, exceptionDuringFallbackValidation);
              errorMessageBuilder.append(System.lineSeparator());
              errorMessageBuilder.append(message);
              throw new GoConfigInvalidException(e.getCruiseConfig(), errorMessageBuilder.toString());
          }
      }
  }
  ```


* In [this line](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R264): How can there be no call to trySavingEntity here? What happens if it succeeds? That change won't be committed to the config.git repo, right? That seems wrong. What am I missing?

* Is there a need for [this clone](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R283)?

* I know we talked about this and decided it was this way, but can you please refresh my memory? - Question is: Why are we checking (new_requested_main_config + valid_partials) before checking (new_requested_main_config + known_partials)?

  If we do this, then a new change in a failing partial which would have been fixed by a change to the main might not be seen (since latest is not used). I can see that it is be a convoluted case. I am not too worried about that happening.

  However, this is unlike what is happening in both forceLoad and writeWithLock, both of which check with lastKnown before lastValid. Any reason that writeEntityWithLock is different in its order?

* How can the lastValidPartials be different [here](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R328)?

* Is [this server health message](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033L272) (which was removed) taken care of somewhere else?

* [This](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R371) is what you were asking about earlier, I guess: About the equals and rewriting. I think it's fine.

* I see you have written tests for the git merge flow. I think we should remove that flow. :) - Also, that class (GoFileConfigDataSource) is ready for a "throw-away-and-rewrite" session. The whole of config save, in my opinion.

* About the shouldSaveConfigChangesWhenFullConfigIsBeingSavedFromConfigXmlTabAndAllKnownConfigRepoPartialsAreInvalid test: I think it makes sense that a full XML save should be allowed, but I'm wondering which flow is allowing this. All of them seemed to check with partials.

* Is [this TODO](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R441) needed any more?

jyotisingh:
>* Why all the change from CruiseConfig to BasicCruiseConfig?

As far as I know, when that interface came up, the intent was to have two implementations for CruiseConfig interface - BasicCruiseConfig and MergedCruiseConfig, but then `MergedCruiseConfig` went on to become something else altogether and did not implement that interface anymore. I dont see a need to have an interface with just one implementer, and I think it should be ok to get rid of that interface but I haven't done that yet. But I can revert my changes so that we continue using CruiseConfig.

>* I guess the PipelineConfigurationCache didn't work out so well because every config which is being changed is different, and it cannot be cached? I see that [the commit message](https://github.com/gocd/gocd/pull/2298/commits/fb8402bf90c71bfedb964cec47cf5a04472c95dc) says as much.
  
Yes thats correct. 

> However, I hope the PipelineConfigSaveValidationContext is long-lived enough in that flow that the cache is not calculated multiple times. Otherwise, it will be expensive.

PipelineConfigSaveValidationContext is constructed and used during a config save validation flow. However, during a given config save flow, if we decide to fallback to different list(valid/known) of partials then a new PipelineConfigSaveValidationContext object would be created and those caches will be recomputed. 

> * The method getPipelinesWithDependencyMaterials doesn't seem to be doing the right thing any more, right? Currently, the set it returns is just the full list of all pipelines in the config. The earlier code used to do something using PipelineConfigMap, which *might* have done what its name suggests.

I dont see much difference in the implementation, its just moving getPipelinesWithDependencyMaterials from PipelineConfigurationCache to PipelineConfigSaveValidationContext. PipelineConfigMap was a map of pipelinename and PipelineCOnfig, but that information is already available in the form of a list in CruiseConfig class(`allPipelineConfigs`), so I am making use of that to construct the cache.

> * [This](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R263) seems wrong: Isn't this marking last known partials as valid, without them having gone through the check in the line after this? What happens if they fail that validation?

No, its not since the merged config has already been validated as part of `String configAsXml = configAsXml(modifiedConfig, false);`, but I get your point that the config has not been checked-in yet. Updated PR.

> * In [this line](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R264): How can there be no call to trySavingEntity here? What happens if it succeeds? That change won't be committed to the config.git repo, right? That seems wrong. What am I missing?

Its basically going through a full config validation flow and not just entity-specific validation, hence you dont see `trySavingEntity`. I have explained it in detail for the question `Why are we checking (new_requested_main_config + valid_partials) before checking (new_requested_main_config + known_partials)`

> * Is there a need for [this clone](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R283)?

Yes. Thats the place where we construct the preprocessed-config by cloning the config-for-edit. This preprocessed config will have parameters and templates inlined. Since we need to cache both config-for-edit and preprocessConfig if the validations pass, we need both those objects to be different objects.

> * I know we talked about this and decided it was this way, but can you please refresh my memory? - Question is: Why are we checking (new_requested_main_config + valid_partials) before checking (new_requested_main_config + known_partials)?
If we do this, then a new change in a failing partial which would have been fixed by a change to the main might not be seen (since latest is not used). I can see that it is be a convoluted case. I am not too worried about that happening.
However, this is unlike what is happening in both forceLoad and writeWithLock, both of which check with lastKnown before lastValid. Any reason that writeEntityWithLock is different in its order?

Yes, it is different from whats happening in the rest of the config save flows. Thing is Entity-save flows perform just entity-specific validation, ie. only the part that is changing or the ones that might get affected are validated, nothing else. In all other cases we merge with known-partials, validate full config and then if all is OK then we save. If we merge with valid partials, we can do away with just the entity-specific validations. But if we merge with known-partials, we cannot perform just entity-specific validation, it has to be full config validation since some other aspects of one of the known partials might be invalid(say a missing mandatory field). The only point of this was to perform the full validation as the last resort.

> * How can the lastValidPartials be different [here](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R328)?

It cannot be since that can change only inside a GoConfigWriteLock which the current thread has already acquired, updated the PR.

> * Is [this server health message](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033L272) (which was removed) taken care of somewhere else?

Yes, the server health messages are taken care of at individual config-repo level, instead of a gloabl message consisting of errors from all config-repos. The health messages are updated as a part of CachedGoPartials#markAsValid and GoPartialConfig#onSuccessPartialConfig.

> * [This](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R371) is what you were asking about earlier, I guess: About the equals and rewriting. I think it's fine.

Cool

> * I see you have written tests for the git merge flow. I think we should remove that flow. :) - Also, that class (GoFileConfigDataSource) is ready for a "throw-away-and-rewrite" session. The whole of config save, in my opinion.

I agree, but getting rid of git merge flow cannot happen unless we implement a flow similar to that of pipeline-config-api for those too, else we will be going back to the time when *concurrent* config save operations in the system were not allowed.

> * About the shouldSaveConfigChangesWhenFullConfigIsBeingSavedFromConfigXmlTabAndAllKnownConfigRepoPartialsAreInvalid test: I think it makes sense that a full XML save should be allowed, but I'm wondering which flow is allowing this. All of them seemed to check with partials.

Now that periodic load of config from filesystem also goes through the same config save flow ie. merge_with_config_repo+validate, it is important that this flow works, hence the test.

> * Is [this TODO](https://github.com/gocd/gocd/pull/2298/files#diff-e236b0af07a149fc225404e31015c033R441) needed any more?

Probably not. I will get rid of it.

We also spoke about handling throwable for plugin calls, I see that its already handled as a part of `FelixGoPluginOSGiFramework`. All the plugin interactions catch a `Throwable` and throw a runtime exception instead. I think thats the right place to do this, so I have left that bit untouched.

tomzo:
> Why all the change from CruiseConfig to BasicCruiseConfig?

@jyotisingh you are right about the origin of this interface. However I left it intentionally for other reasons. In future we don't want to have single `CruiseConfig` interface which is referenced in 1000+ other places like now. Instead we want to break it into parts, e.g. AgentsCruiseConfig, PipelinesCruiseConfig, AuthCruiseConfig, etc. Even then BasicCruiseConfig can still exist and implement them all. But ideally, second step is to remove `BasicCruiseConfig` and replace it by many other smaller implementations.

jyotisingh:
@tomzo - I am not sure if I follow what you mean by that. 
* What would we achieve by making `BasicCruiseConfig` implement interfaces like AgentsCruiseConfig, PipelinesCruiseConfig, AuthCruiseConfig, etc? 
* `second step is to remove BasicCruiseConfig and replace it by many other smaller implementations`
BasicCruiseConfig composes all the other config entities. Why do you want to get rid of BasicCruiseConfig? 
Are you saying this in context of config-repo partials or about the system in general?

arvindsv:
@jyotisingh: I looked at the two commits you made. Looks ok to me. One question I have is: Why do we need the clones on the partials which come from the CachedGoPartials? The lastValidPartials and lastKnownPartials methods both create a new list. The components of those lists (UpdatedPartial objects) are not mutable and any change to the cache which happens externally during the config save should not affect this list anyway.

tomzo:
I am talking about system at high scope, config repos are slightly relevant.

> What would we achieve by making BasicCruiseConfig implement interfaces like AgentsCruiseConfig, PipelinesCruiseConfig, AuthCruiseConfig, etc?

Nothing, that is not the point. At point when BasicCruiseConfig implements all those interfaces this is just *transition* state. What matters is that we acheive less coupling with `CruiseConfig` interface when we split it into parts.
It is often pointless to obtain a full configuration instance only to get a small independent part of it.

> BasicCruiseConfig composes all the other config entities. Why do you want to get rid of BasicCruiseConfig?

Some reasons to remove BasicCruiseConfig:

1. Because it is too big and hard to manage. There are parts of configuration that could be managed independently of others, for example agents config. This composition is creating unnecesarily a *bottle neck*, because every, (even small) config update must flow through this big global object.
2. Instead of having one config with everything, there could be smaller services providing only necessary information that particular service consumer needs. `GoConfigService` is currently the closest to such architecture but only when looking from outside. Internally we know that is actually [calls](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/GoConfigService.java#L195) `goConfigDao.load()` which requires a full global, valid config. So below that service is where currently *bottleneck design* starts.
Take for example case when new agent is registered, there is no need to validate pipelines when adding agent. That operation could bypass all that is outside of agents config scope. But we need to change dependencies of agent-related services. Add something like `AgentsDao` service, `AgentConfigService`, `AgentConfigDataSource`, etc. 
3. Moreover big, global, valid config was an option when there is only one source of configuration. But that is no longer true. When services reflect this new distributed nature of config sources then we will be able to solve such issues like you mentioned last time (with 2 or more errors in remote pipelines).
4. BasicCruiseConfig and its children are also coupled to xml because of how magical xml loader works. So currently any service using a config is actually coupled to xml config store. If we make services depend on smaller interfaces instead then we have removed indirect dependency on xml format.

jyotisingh:
@arvindsv - regarding [this](https://github.com/gocd/gocd/pull/2298#issuecomment-229158022) comment, cloning is required since validation process will add ConfigErrors to the config entities such as pipelineConfig etc and we wouldn't want the cached partial objects(which internally refer to PipelineConfig) to have those errors. However, I looked at the code again after you pointed that out, and found that I was cloning the partials twice, once (explicitly) in the line that you pointed out and second time (implicitly) during creation of preprocessed-config, so I have removed the first clone(in [a93bec8](https://github.com/gocd/gocd/pull/2298/commits/a93bec8447ec260bdd3a63d0715a2d393a895962))

By the way, I had to rebase since there were conflicts with some new changes on master.

arvindsv:
Alright.

Issue #2299, opened 2016-06-14
git commit author email does not get displayed properly in build-cause popup on pipeline history page


jyotisingh:
##### Issue Type
- Bug Report

##### Summary

Configure a pipeline with a git material, make a commit, and then trigger the build. The build-cause popup displays the author details properly, but the build-cause popup on pipeline history does not:

Dashboard (email shows up correctly):
<img width="669" alt="screen shot 2016-06-14 at 19 41 59" src="https://cloud.githubusercontent.com/assets/1046342/16045909/21c4259a-3268-11e6-8c75-ff11bd012f07.png">

Pipeline History (incorrect => `<` and `>` are escaped):
<img width="639" alt="screen shot 2016-06-14 at 19 41 42" src="https://cloud.githubusercontent.com/assets/1046342/16047884/7ad86ef0-326f-11e6-933b-71c994d0ecfe.png">

rajiesh:
Verified on 16.6.0-3590(af0f8b8ae8a4342ab748ce80224971887f018ccc) and works as expected

PR #2301, opened 2016-06-15
#2299 Removed HTML escaping of build-cause-author-name 

GaneshSPatil:
Removed HTML escaping of build-cause-author-name to display the name without escaping

author details:

previous result : `ganeshpl &lt;ganespl@INganespl.local&gt;`
Changed to     : `ganeshpl <ganespl@INganespl.local>`


ketan:
#2299

PR #2305, opened 2016-06-15
 fix #993 by skipping empty svn log entries, take 2

### LINKED ISSUE #993: "Modification check failed for material" with restrictive subversion permissions

norcnorc:
I don't see a way to reopen  #1265, so I'm creating a new PR.

### DIFF
--- a/common/src/com/thoughtworks/go/util/SvnLogXmlParser.java
+++ b/common/src/com/thoughtworks/go/util/SvnLogXmlParser.java
@@ -56,7 +56,7 @@ public class SvnLogXmlParser {
         for (Iterator iterator = logEntries.iterator(); iterator.hasNext();) {
             Element logEntry = (Element) iterator.next();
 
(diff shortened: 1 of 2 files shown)


norcnorc:
@ketan - signed.

PR #2309, opened 2016-06-17
Revert "Allow turning on web request logging by tweaking `log4j.properties`"

zabil:
Reverts gocd/gocd#2282


zabil:
Opening another one with slight modifications.

PR #2318, opened 2016-06-17
Ignore flaky test

arvindsv:
I am working on this. I've brought up a VM, and limited it to 512MB and 2 CPUs, to try and reproduce this. Till that point, this test will be ignored. I considered removing it, but decided to ignore it instead.


arvindsv:
Has been running on the VM for a while (`while mvn -Dtest=AgentControllerTest test; do sleep 3; done; say Failed`). Has not failed. :(

I took over agent03, to try it on that itself. It failed as soon as I did `mvn test --batch-mode`. Trying that on local VM, and it doesn't fail like that. Sigh.

PR #2328, opened 2016-06-21
Re-org all new SPA assets to make room for other SPAs

ketan:


arvindsv:
Don't let these kind of changes rot. Get @maheshp to merge it in, quick. :) Also, it would be ideal to be able to run Jasmine tests against each SPA separately, if possible. Will make development easier, I assume. I mean, without having to bring up the app and rails. Just a thought. Not necessarily a part of this PR.

ketan:
Already spoken to @maheshp about this, he'll review it soon, I've intentionally kept this change small(ish) to help review.

PR #2330, opened 2016-06-21
Ignore specs of turned off feature

arvindsv:
Related to #2326. Since the feature is off, the specs are failing.


arvindsv:
@ketan - Can you take a look at this later, to see what you want to do? I considered changing the way the flag is decided (so that it can be controlled from the test). Decided not to do it for now. You decide.

arvindsv:
I will merge it, so that the tests pass, but you can take a look at how to revert this later.

PR #2333, opened 2016-06-22
There's no `config-repository` URL for cloning.

ketan:


arvindsv:
Shouldn't it be removed from acegi, and shouldn't that filter class be removed too?

ketan:
It's not there in acegi. I'd put in 2 url in both places (acegi, web.xml) and later decided to remove one from acegi, but forgot to remove it from web.xml

PR #2334, opened 2016-06-22
Delete pipeline check

GaneshSPatil:
Moved the dependency check for the pipeline from controller to deletePipelineConfigCommand

### DIFF
--- a/server/src/com/thoughtworks/go/config/update/DeletePipelineConfigCommand.java
+++ b/server/src/com/thoughtworks/go/config/update/DeletePipelineConfigCommand.java
@@ -16,11 +16,11 @@
 
 package com.thoughtworks.go.config.update;
 
-import com.thoughtworks.go.config.BasicCruiseConfig;
 import com.thoughtworks.go.config.CruiseConfig;
+import com.thoughtworks.go.config.EnvironmentConfig;
 import com.thoughtworks.go.config.PipelineConfig;
-import com.thoughtworks.go.config.PipelineConfigSaveValidationContext;
 import com.thoughtworks.go.config.commands.EntityConfigUpdateCommand;
+import com.thoughtworks.go.i18n.Localizable;
 import com.thoughtworks.go.i18n.LocalizedMessage;
 import com.thoughtworks.go.server.domain.Username;
 import com.thoughtworks.go.server.service.GoConfigService;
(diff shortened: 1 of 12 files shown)


server/src/com/thoughtworks/go/config/update/DeletePipelineConfigCommand.java
```diff
@@ -50,6 +50,22 @@ public void update(CruiseConfig cruiseConfig) throws Exception {
...
     public boolean isValid(CruiseConfig preprocessedConfig) {
+        for (PipelineConfig pipeline : preprocessedConfig.getAllPipelineConfigs()) {
+            if(pipeline.materialConfigs().hasDependencyMaterial(pipelineConfig)){
+                Localizable.CurryableLocalizable message = LocalizedMessage.string("CANNOT_DELETE_PIPELINE_USED_AS_MATERIALS", pipelineConfig.name(), pipeline.name());
+                this.result.unprocessableEntity(message);
+                return false;
+            }
+        }
+
+        for (EnvironmentConfig environment : goConfigService.getEnvironments()) {
```
jyotisingh: Please change this to `preprocessedConfig.getEnvironments()` here instead of asking the same from service

server/src/com/thoughtworks/go/config/update/DeletePipelineConfigCommand.java
```diff
@@ -50,6 +50,22 @@ public void update(CruiseConfig cruiseConfig) throws Exception {
 
     @Override
     public boolean isValid(CruiseConfig preprocessedConfig) {
+        for (PipelineConfig pipeline : preprocessedConfig.getAllPipelineConfigs()) {
+            if(pipeline.materialConfigs().hasDependencyMaterial(pipelineConfig)){
+                Localizable.CurryableLocalizable message = LocalizedMessage.string("CANNOT_DELETE_PIPELINE_USED_AS_MATERIALS", pipelineConfig.name(), pipeline.name());
```
jyotisingh: replace `pipeline.name()` with `String.format("%s (%s)", pipeline.name(), pipeline.getOriginDisplayName())`
`pipelineConfig.name()` can remain as is since we do not allow deletion of remote(config-repo) pipelines. The check for that is currently missing at the controller side, the CRUD operations on remote pipelines will eventually fail deep down in the models, but we can fix that later.

PR #2357, opened 2016-06-28
speeding up jasmine specs #000

maheshp:
* Mounting multiple components against the same root element would
  unmount the previous components, so there is no need to explicitly
  create and remove the root element after each spec


ketan:
Improving on this PR by opening #2358

PR #2358, opened 2016-06-28
Improve jasmine test speeds

ketan:
Mithril seems to keep track of DOM elements (https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/mithril-0.2.0.js#L550)
even when the element is removed from the document. It keeps re-painting
the elements on a 60Hz cycle.

The "fix" is to re-use the same DOM element and mount mithril components
onto the same DOM element so that we don't "leak" DOM elements and slow
down mithril re-draws as more and more tests run.


zabil:
On another note,
 We should be using https://github.com/StephanHoyer/mithril-query

 For testing the virtual dom and not queries on the real dom.

Issue #1508, opened 2015-09-21
API to create pluggable scm material config


jyotisingh:
... as edit pipeline config api allows only reference to an existing pluggable scm

varshavaradarajan:
URL: ```curl https://ci.example.com/go/api/admin/scm/[:material-name]```
GET Response: 
``` 
{
  "id": "uuid",
  "name": "github",
  "auto_update": true,
  "plugin_configuration": {
    "id": "github.pr",
    "version": "1"
  },
  "configuration": [
    {
      "key": "username",
      "value": "admin"
    },
    {
      "key": "password",
      "encrypted_value": "1frsf33h4t0fjds"
    }
  ]
}

```
CREATE
URL: ```curl https://ci.example.com/go/api/admin/scm```
Data:       
```  
{
  "name": "github",
  "auto_update": true,
  "plugin_configuration": {
    "id": "github.pr",
    "version": "1"
  },
  "configuration": [
    {
      "key": "username",
      "value": "admin",
      "secure": false
    },
    {
      "key": "url",
      "value": "http://git-url",
      "secure": true
    },
    {
      "key": "password",
      "encrypted_value": "1f3rrsnuhn63hd",
      "secure": true
    }
  ]
}
```
EDIT
URL: ```curl https://ci.example.com/go/api/admin/scm/[:material-name]```
Data
```
{
  "id": "uuid",
  "name": "github",
  "auto_update": true,
  "plugin_configuration": {
    "id": "github.pr",
    "version": "1"
  },
  "configuration": [
    {
      "key": "username",
      "value": "admin",
      "secure": false
    },
    {
      "key": "url",
      "value": "http://git-url",
      "secure": true
    },
    {
      "key": "password",
      "encrypted_value": "1f3rrsnuhn63hd",
      "secure": true
    }
  ]
}        
```
Note: 
1. Do not allow rename of material name while editing. 
2. Throw an error if the SCM id is passed while *creating* a new SCM 
3. Check all the existing domain validations for SCM

PR #2360, opened 2016-06-29
Pluggable SCM - get, create, update (#1508)

varshavaradarajan:
* Changed Configuration Property Builder to accomodate for all types of plugins.

### DIFF
--- a/server/src/com/thoughtworks/go/config/update/CreateSCMConfigCommand.java
+++ b/server/src/com/thoughtworks/go/config/update/CreateSCMConfigCommand.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.config.update;
+
+import com.thoughtworks.go.config.CruiseConfig;
+import com.thoughtworks.go.domain.scm.SCM;
+import com.thoughtworks.go.domain.scm.SCMs;
+import com.thoughtworks.go.server.domain.Username;
+import com.thoughtworks.go.server.service.GoConfigService;
+import com.thoughtworks.go.server.service.materials.PluggableScmService;
+import com.thoughtworks.go.server.service.result.LocalizedOperationResult;
+
+public class CreateSCMConfigCommand extends SCMConfigCommand {
+
+    public CreateSCMConfigCommand(SCM globalScmConfig, PluggableScmService pluggableScmService, LocalizedOperationResult result, Username currentUser, GoConfigService goConfigService) {
+        super(globalScmConfig, pluggableScmService, goConfigService, currentUser, result);
+    }
+
+    @Override
+    public void update(CruiseConfig modifiedConfig) throws Exception {
+        SCMs scms = modifiedConfig.getSCMs();
+        scms.add(globalScmConfig);
+        modifiedConfig.setSCMs(scms);
+    }
+
+}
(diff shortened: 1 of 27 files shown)


server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pluggable_scms_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      def update
+        result = HttpLocalizedOperationResult.new
+        @scm = ApiV1::Scms::PluggableScmRepresenter.new(SCM.new).from_hash(params)
+        pluggable_scm_service.updatePluggableScmMaterial(current_user, @scm, result)
+
+        json = ApiV1::Scms::PluggableScmRepresenter.new(@scm).to_hash(url_builder: self)
+        result.isSuccessful ? (render DEFAULT_FORMAT => json) : (render_http_operation_result(result, {data: json}))
+      end
+
+      def check_for_stale_request
```
ketan: please mark these methods as `private` as a matter of hygine, it only takes a bad route to be able to reach this method.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/scms/pluggable_scm_representer.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+      link :doc do
+        'http://api.go.cd/#pluggable-scms'
+      end
+      property :errors, exec_context: :decorator, decorator: ApiV1::Config::ErrorRepresenter, skip_parse: true, skip_render: lambda { |object, options| object.empty? }
+      property :id
+      property :name
+      property :auto_update
+      property :plugin_configuration,
```
ketan: I know I'm deviating, but should we call this `plugin_metadata`, `configuration` and `plugin_configuration` are quite ambiguous.

ketan:
I know that `PluginConfigurationPropertyRepresenter` was not part of this PR, but what happens when you submit both `value` and `encrypted` value to it?

varshavaradarajan:
For ```PluginConfigurationPropertyRepresenter```, a configuration property object will be created with both the value and encrypted_value set. But this is not the property object that will be added to the configuration be it scm or pluggable task but rather the property object created out of this deserialized property object and the configuration preference or metadata which provides the options for each of the properties

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/scms/pluggable_scm_representer.rb
varshavaradarajan: makes sense. But in the model it is known as PluginConfiguration and Configuration anyway.

server/src/com/thoughtworks/go/config/update/CreateSCMConfigCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public void update(CruiseConfig modifiedConfig) throws Exception {
+        SCMs scms = modifiedConfig.getSCMs();
+        scms.add(globalScmConfig);
+        modifiedConfig.setSCMs(scms);
+    }
+
+    @Override
+    public boolean isValid(CruiseConfig preprocessedConfig) {
```
ketan: Is valid is duplicated between create and update commands, is there a way to remove that duplication?

server/src/com/thoughtworks/go/config/update/CreateSCMConfigCommand.java
```diff
@@ -0,0 +1,85 @@
...
+        return false;
+    }
+
+    @Override
+    public SCM getPreprocessedEntityConfig() {
+        return preprocessedGlobalScmConfig;
+    }
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
```
ketan: This too seems to be duplicated

server/src/com/thoughtworks/go/config/update/CreateSCMConfigCommand.java
```diff
@@ -56,30 +41,16 @@ public void update(CruiseConfig modifiedConfig) throws Exception {
...
     }
 
     @Override
     public boolean canContinue(CruiseConfig cruiseConfig) {
-        if (!(goConfigService.isUserAdmin(currentUser)) || goConfigService.isGroupAdministrator(currentUser.getUsername())) {
-            result.unauthorized(LocalizedMessage.string("UNAUTHORIZED_TO_EDIT"), HealthStateType.unauthorised());
-            return false;
-        }
-        return true;
+        return super.canContinue(cruiseConfig);
```
ketan: no need for a super call, goes to super by default :)

Issue #2218, opened 2016-05-25
pipeline iteration comment "null"


fabriciocduarte:
##### Issue Type

- Bug Report

##### Summary

When looking at the iterations of a pipeline / pipeline history, it is displaying "null" on all entries that don't have a comment added to.

http://goserver/go/tab/pipeline/history/myPipeline

###### Basic environment details

* Go Version: Go Version: 16.3.0(3183-60d6fd31f00527da00f2affdd8f7f7732e7be103)
* OS: Windows Server 2008 R2
* Browser vendor and version (if relevant): Google Chrome v. 50.0.2661.102 m

##### Steps to Reproduce

1. 
2. 
3. 
4.

PR #2361, opened 2016-06-29
Added a null check before displaying the pipeline instance comment. (#2218)

varshavaradarajan:


ketan:
Is this a candidate for `StringUtils.isBlank`?

zabil:
It's a js template

PR #2366, opened 2016-06-30
Added an internal endpoint to get all the resources and environments

GaneshSPatil:
needed for agents-SPA

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/internal/resources_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/internal/resources_controller.rb
@@ -1,0 +1,32 @@
+##########################################################################
+# Copyright 2016 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/internal/resources_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+##########################################################################
+
+module ApiV1
+  module Admin
+    module Internal
+      class ResourcesController < ApiV1::BaseController
+        before_action :check_admin_user_and_401
+
+        def index
+          render json_hal_v1: { :allResources => go_config_service.getResourceList()}
```
ketan: Just a simple array should be sufficient, I think. No need for keys in there.

`['foo', 'bar']`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/internal/resources_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Admin
+    module Internal
+      class ResourcesController < ApiV1::BaseController
+        before_action :check_admin_user_and_401
```
ketan: Anyone who can edit pipelines should be able to access this, so I suppose it should be ok to allow group admins too?

ketan:
Can you add the relevant acegi filters and if needed urlrewrite filters?

PR #2367, opened 2016-06-30
Remove old controllers that have been superseeded

ketan:


server/webapp/WEB-INF/rails.new/spec/controllers/api/server_controller_spec.rb
```diff
@@ -29,32 +29,4 @@
     controller.stub(:populate_config_validity)
```
varshavaradarajan: Why is the before(:each) block required when all the specs are removed?

server/properties/test/cruise.properties
```diff
@@ -1,19 +1,19 @@
...
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-##########################GO-LICENSE-END##################################
-
+##########################GO-LICENSE-START################################
+# Copyright 2014 ThoughtWorks, Inc.
```
varshavaradarajan: 2016. :)

ketan: Leaving it because it'll cause conflicts with the elastic agent :)

server/webapp/WEB-INF/rails.new/spec/controllers/api/server_controller_spec.rb
```diff
@@ -29,32 +29,4 @@
...
   end
```
varshavaradarajan: Also, noticed that there were no specs for capture_support_info. Intentional?

ketan:
> Also, noticed that there were no specs for capture_support_info. Intentional?

The old one did not have it. And it was hard to stub and mock stuff that was originally not intended to be, so we let it be.

PR #2368, opened 2016-06-30
#1133 add missing hasSamePipelinesAs implementation

tomzo:
This fixes missing implementation of `MergeEnvironmentConfig.hasSamePipelinesAs`.

In `EnvironmentsConfig` any `EnvironmentConfig` instance must have unique pipelines which are final members of that environment. Regardless of config sources of that environment.
So in case when `EnvironmentsConfig.validateNotADuplicate(EnvironmentConfig environmentConfig)` is called on environment with many sources, we should still validate that `MergeEnvironmentConfig` is the only definition for that environment.

### DIFF
--- a/config/config-api/test/com/thoughtworks/go/config/MergeCruiseConfigTest.java
+++ b/config/config-api/test/com/thoughtworks/go/config/MergeCruiseConfigTest.java
@@ -8,9 +8,7 @@ import com.thoughtworks.go.config.merge.MergePipelineConfigs;
 import com.thoughtworks.go.config.remote.*;
 import com.thoughtworks.go.domain.ConfigErrors;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
-import com.thoughtworks.go.helper.GoConfigMother;
-import com.thoughtworks.go.helper.PartialConfigMother;
-import com.thoughtworks.go.helper.PipelineConfigMother;
+import com.thoughtworks.go.helper.*;
 import org.hamcrest.core.Is;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


config/config-api/test/com/thoughtworks/go/config/MergeCruiseConfigTest.java
```diff
@@ -49,6 +48,21 @@ protected BasicCruiseConfig createCruiseConfig() {
...
+        ConfigRepoConfig repoConfig1 = new ConfigRepoConfig(MaterialConfigsMother.gitMaterialConfig("url1"), "plugin");
+        ConfigRepoConfig repoConfig2 = new ConfigRepoConfig(MaterialConfigsMother.gitMaterialConfig("url2"), "plugin");
+        cruiseConfig.setConfigRepos(new ConfigReposConfig(repoConfig1, repoConfig2));
+        PartialConfig partialConfigInRepo1 = PartialConfigMother.withEnvironment("environment", new RepoConfigOrigin(repoConfig1, "repo1_r1"));
+        RepoConfigOrigin configOrigin = new RepoConfigOrigin(repoConfig2, "repo2_r1");
+        PartialConfig partialConfigInRepo2 = PartialConfigMother.withEnvironment("environment", configOrigin);
+        BasicEnvironmentConfig environment2InRepo2 = EnvironmentConfigMother.environment("environment2_in_repo2");
+        environment2InRepo2.setOrigins(configOrigin);
+        partialConfigInRepo2.getEnvironments().add(environment2InRepo2);
+        cruiseConfig.merge(asList(partialConfigInRepo2, partialConfigInRepo1), false);
```
jyotisingh: Can u please add an assertion at the end of this test to check that the environments are all part of cruiseConfig

Issue #2348, opened 2016-06-24
Invalid entries in cruise-config.xml break the login mechanism


lenucksi:
##### Issue Type
- Bug Report

##### Summary
If the cruise-config.xml is edited manually and a user introduces a role definition within a pipeline, all authentication mechanisms break and a login into the server is not possible anymore.
As soon as the erroneous XML segment is removed, login works again.

Details:
 - The role with its members is not defined in the correct global position but within the pipeline. Thus the pipeline does not reference a role for a given authorization but attempts to define it in-place.
- There is no useful logging using the standard logging configuration that would hint to this issue.

##### Environment
- Go 16.1
- CentOS 7
- OpenJDK 7

##### Steps to Reproduce
See above. 
Additional details will be provided soon.

##### Expected Results
- The server should complain about invalid config XML upon starting / config reload in the log.
- It would also be nice to see the offending line.
- Login should either work aside from this issue or the login screen should present a warning that hints to the logs. 

##### Actual Results
The logs only show
```
2016-06-22 11:27:30,474 ERROR [qtp57730642-30] BasicAuthenticationFilter:56 -
```
The login stops working with all authentication backends.

##### Log snippets
```
2016-06-22 11:27:30,282 DEBUG [qtp57730642-25] AnonymousProcessingFilter:93 - Populated SecurityContextHolder with anonymous token: 'org.springframework.security.providers.anonymous.AnonymousAuthenticationToken@6fa8940c: Principal: anonymousUser; Password: [PROTECTED]; Authenticated: true; Details: org.springframework.security.ui.WebAuthenticationDetails@fffde5d4: RemoteIpAddress: 1.2.3.4; SessionId: xxx; Granted Authorities: ROLE_ANONYMOUS'
2016-06-22 11:27:30,283 DEBUG [qtp57730642-25] GoExceptionTranslationFilter:104 - Chain processed normally
2016-06-22 11:27:30,474 ERROR [qtp57730642-30] BasicAuthenticationFilter:56 -
2016-06-22 11:27:30,474 DEBUG [qtp57730642-30] BasicAuthenticationFilter:57 -
java.lang.NullPointerException
    at com.thoughtworks.go.config.AdminRole.isSameAs(AdminRole.java:45)
    at com.thoughtworks.go.config.AdminsConfig.isAdmin(AdminsConfig.java:42)
    at com.thoughtworks.go.config.TemplatesConfig.canViewAndEditTemplate(TemplatesConfig.java:102)
[...]

PR #2377, opened 2016-07-01
NPE thrown by AdminRole

sadovnikov:
This PR is related to #2348, which can be broken into three parts:

- ```AdminRole``` throws NPE;
- ```BasicAuthenticationFilter``` does not report error, when message is not present in caught exception;
- problems in ```cruise-config.xml``` are not reported

I was not able to reproduce the exact reported problem and sample of ```cruise-config.xml``` is not provided (yet). Therefore this PR fixes the first two problems in the code


zabil:
Thanks!

PR #2378, opened 2016-07-01
Fix links in APIs (mostly)

arvindsv:
* Change http to https
* Also, add /current/ where needed. Redirect won't work properly with that fragment otherwise.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/pipeline_config_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/pipeline_config_representer.rb
@@ -31,7 +31,7 @@ module ApiV1
       end
 
       link :doc do |opts|
-        'http://api.go.cd/#pipeline-config'
+        'https://api.go.cd/current/#pipeline-config'
       end
 
       link :find do |opts|
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 # Contributing
 
-We encourage you to contribute to Go. For information on contributing to this project, please see our <a href="http://www.go.cd/contribute/">contributor's guide</a>.
-A lot of useful information like links to user documentation, design documentation, mailing lists etc can be found in the <a href="http://www.go.cd/community/resources.html">resources</a> section.
+We encourage you to contribute to GoCD. For information on contributing to this project, please see our <a href="http://www.go.cd/contribute/">contributor's guide</a>.
+A lot of useful information like links to user documentation, design documentation, mailing lists etc can be found in the <a href="https://www.go.cd/resources/">resources</a> section.
 
 ## Contributing Code?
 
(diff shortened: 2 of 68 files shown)


arvindsv:
If there's no problem with this and it gets merged, I assume I'll need to change it in all the documentation as well. Will do if it's merged.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/pipeline_config_representer.rb
```diff
@@ -31,7 +31,7 @@ class PipelineConfigRepresenter < ApiV1::BaseRepresenter
       end
 
       link :doc do |opts|
-        'http://api.go.cd/#pipeline-config'
+        'https://api.go.cd/current/#pipeline_config'
```
varshavaradarajan: @arvindsv, in the current api docs, it is pipeline-config. Did you change the api docs as well?

ketan:
Would it not be easier to setup redirects at the index.html page over at api.go.cd?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/pipeline_config_representer.rb
arvindsv: No. This is just a search-replace in the codebase.

arvindsv:
Maybe it is easier, but I feel that correcting this is ... more correct.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/pipeline_config_representer.rb
arvindsv: But, I see what you mean. Maybe I need to rebase and re-do it. I will.

varshavaradarajan: Yeah, There was one PR today morning to replace pipeline_config with pipeline-config. :)

arvindsv: I remember. The one where @ketan kept changing his mind. :)

I've now corrected this.

arvindsv:
@ketan thinks we should just redirect in the docs. I think otherwise. The next person who has an opinion gets to break the tie. I won't begrudge you if the PR gets closed because of your decision, as long as there is a decision. :)

ketan:
Can you make this 2 PRs? One for URL change in representers and another for docs and header/footer?

On another note, u was considering replacing the footer on all pages to be the same as/similar to the one in the SPA. Thoughts?

arvindsv:
> Can you make this 2 PRs? One for URL change in representers and another for docs and header/footer?

@ketan: Ok, closing it.

> On another note, u was considering replacing the footer on all pages to be the same as/similar to the one in the SPA. Thoughts?

As long as it is consistent and looks good, I'm fine with it. @naveenbhaskar might be able to help, to say if it looks and feels ok. It shouldn't feel like a franken-app where going from an SPA page to a non-SPA page is a jarring experience.

PR #2380, opened 2016-07-02
#1133 whitelist support for config repos

tomzo:
Until now, config-repo extension point accepted only
```json
"filter": {
    "ignore": [
      "externals",
      "tools"
    ]
  }
```
Now it can also accepts filters with whitelist:
```json
"filter": {
    "whitelist": [
      "src",
      "tests"
    ]
  }
```

 * Added a few tests to cover git case fully. 
 * Also if user specifies both non-empty `ignore` and `whitelist` then it is considered as error in config repo.

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRFilter.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRFilter.java
@@ -9,15 +9,20 @@ import java.util.List;
 
 public class CRFilter extends CRBase {
     private List<String> ignore  = new ArrayList<String>();
-    // room for whitelist
+    private List<String> whitelist  = new ArrayList<String>();
 
-    public CRFilter(List<String> ignore) {
-        this.ignore = ignore;
+    public CRFilter(List<String> list,boolean whitelist) {
+        if(whitelist)
+            this.whitelist = list;
+        else
+            this.ignore = list;
     }
 
     @Override
     public void getErrors(ErrorCollection errors, String parentLocation) {
-
+        if (this.isBlacklist() && this.isWhitelist()) {
+            errors.addError(getLocation(parentLocation), "Material filter cannot contain both ignores and whitelist");
+        }
     }
 
     @Override
@@ -26,14 +31,34 @@ public class CRFilter extends CRBase {
         return String.format("%s; Filter",myLocation);
     }
 
-    public List<String> getIgnore() {
-        return ignore;
+    public boolean isWhitelist() {
+        return whitelist != null && whitelist.size() > 0;
+    }
+
+    public List<String> getList() {
+        if(isBlacklist())
+            return ignore;
+        else
+            return whitelist;
+    }
+
+    private boolean isBlacklist() {
+        return ignore != null && ignore.size() > 0;
     }
 
     public void setIgnore(List<String> ignore) {
         this.ignore = ignore;
+        this.whitelist = null;
     }
 
+    public void setWhitelist(List<String> whitelist) {
+        this.ignore = null;
+        this.whitelist = whitelist;
+    }
+
+    public void setWhitelistNoCheck(List<String> list) {
+        this.whitelist = list;
+    }
 
     @Override
     public boolean equals(Object o) {
@@ -48,6 +73,9 @@ public class CRFilter extends CRBase {
         if (ignore != null ? !CollectionUtils.isEqualCollection(this.ignore, that.ignore) : that.ignore != null) {
             return false;
         }
+        if (whitelist != null ? !CollectionUtils.isEqualCollection(this.whitelist, that.whitelist) : that.whitelist != null) {
+            return false;
+        }
 
         return true;
     }
@@ -56,6 +84,7 @@ public class CRFilter extends CRBase {
     public int hashCode() {
         int result = 32;
         result = 31 * result + (ignore != null ? ignore.size() : 0);
+        result = 31 * result + (whitelist != null ? whitelist.size() : 0);
         return result;
     }
 
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRPluggableScmMaterial.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRPluggableScmMaterial.java
@@ -22,13 +22,13 @@ public class CRPluggableScmMaterial extends CRMaterial implements SourceCodeMate
         super(TYPE_NAME,materialName);
         this.scm_id = scmId;
         this.destination = folder;
-        this.filter = new CRFilter(Arrays.asList(filters));
+        this.filter = new CRFilter(Arrays.asList(filters),false/*not supported in 16.6*/);
     }
     public CRPluggableScmMaterial(String name, String scmId, String directory, List<String> filter) {
         super(TYPE_NAME,name);
         this.scm_id = scmId;
         this.destination = directory;
-        this.filter = new CRFilter(filter);
+        this.filter = new CRFilter(filter,false);
     }
 
     @Override
@@ -79,10 +79,10 @@ public class CRPluggableScmMaterial extends CRMaterial implements SourceCodeMate
         this.scm_id = scmId;
     }
 
-    public List<String> getFilterIgnores() {
+    public List<String> getFilterList() {
         if(filter == null)
             return null;
-        return filter.getIgnore();
+        return filter.getList();
     }
 
     public void setFilterIgnore(List<String> filter) {
--- a/config/config-api/src/com/thoughtworks/go/config/materials/git/GitMaterialConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/materials/git/GitMaterialConfig.java
@@ -41,8 +41,6 @@ public class GitMaterialConfig extends ScmMaterialConfig {
 
     private String submoduleFolder;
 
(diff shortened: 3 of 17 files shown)


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRFilter.java
```diff
@@ -9,15 +9,20 @@
 
 public class CRFilter extends CRBase {
     private List<String> ignore  = new ArrayList<String>();
-    // room for whitelist
+    private List<String> whitelist  = new ArrayList<String>();
```
ketan: Seems odd that we're storing both and toggling on presence of one of the objects. Can't we just remember the `whitelist` boolean value instead?

ketan: Please ignore my comment above, I just saw the setters.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/material/CRPluggableScmMaterial.java
```diff
@@ -22,13 +22,13 @@ public CRPluggableScmMaterial(String materialName,String scmId,String folder,Str
         super(TYPE_NAME,materialName);
         this.scm_id = scmId;
         this.destination = folder;
-        this.filter = new CRFilter(Arrays.asList(filters));
+        this.filter = new CRFilter(Arrays.asList(filters),false/*not supported in 16.6*/);
```
ketan: What's that comment supposed to mean?

tomzo: That there is no whitelist support in pluggable scm materials yet. Isn't that right?
I found that `invertFilter` field exists only in `ScmMaterialConfig`, there is no such field in `PluggableSCMMaterialConfig`.

ketan: You're right about that. Might need to spend some time thinking about how we can go about adding whitelist support in pluggable SCMs.

arvindsv:
@tomzo: I was wondering whether we'd write the first jolt upgrade script for this (but I guess it's not needed yet). Also, now that this is starting to come together and the pipeline config API is too, what are your thoughts about making this whole CR set of classes mimic the pipeline config API? I know we've talked about it before and we had decided not to tie the two together and let them evolve separately.

My current thought is: *If* we do that, we could save on having to document this API is two places (and in two different styles), etc. Might end up being less confusing to users. But, I realize it's some work and they can still drift apart (for instance, adding a feature to the config API won't automatically change the CR set of classes - and it shouldn't).

tomzo:
> I was wondering whether we'd write the first jolt upgrade script for this (but I guess it's not needed yet).

No it's not needed, I merely added an optional "whitelist" to filter element. If it is not present, then it is the same as if user specified null.

> what are your thoughts about making this whole CR set of classes mimic the pipeline config API? 
 
I did review and update CR classes not so long ago to match with API. It got merged in #1825. In context of making the both APIs more user friendly I think we are good:
 * field names and their meaning are the same in both APIs
 * I switched to using underscores in all names
But these are not exactly the same "schemas". Biggest difference is that API is following such pattern for objects:
```json
{
   "type" : "git"
   "attributes" : {
      "url" : ...
   }
}
```
And extension point does not have `attributes` anywhere:
```json
{
   "type" : "git"
   "url" : ...
}
```
Other differences are lack of templates and parameters in extension point. But that will stay, at least now. So if you are talking about having the **same document** to for both, I don't think that is good idea.
I would do some copy pasting from `api.go.cd` and fill in the rest as I have already documented on [json config plugin](https://github.com/tomzo/gocd-json-config-plugin#format). And BTW the right place to put the docs would be `plugin-api.go.cd`, that would take of having the same style, right? I have an old PR in https://github.com/gocd/documentation/pull/212

tomzo:
One more thing I am lately thinking is that we should start maintaining actual JSON schemas for elements of config repo extension. Primarily to be able to develop out of Go tools for editing configuration. Once we have such schema I can easily validate it before committing and sending it to Go server. The way I see it is that config repo plugin can have complementary tool which does this:
1. Parse users repository in *whatever* format
2. Prepare JSON representation as would be sent to Go
2. Use JSON schema to validate it and show user the errors

Also similar idea, there could be a *Go domain validation library*, that would allow to catch errors which are not possible to validate with JSON schema. This library would be used outside of Go as dependency in the *editing* tools.

And lastly - forget all above and instead we could make config repo validation API endpoint which would accept *anything* and return configuration errors to the user. Editing tool would need a configured access to particular Go server. But it has the benefits of being always up to date with the exact server that user has.

arvindsv:
> Other differences are lack of templates and parameters in extension point. But that will stay, at least now. So if you are talking about having the same document to for both, I don't think that is good idea.

I didn't mean the CR* interface and the JSON format between the plugin and the server. I was thinking more about the external facing format of the JSON plugin (that users see). But, it's not critical.

> I would do some copy pasting from api.go.cd and fill in the rest as I have already documented on json config plugin. And BTW the right place to put the docs would be plugin-api.go.cd, that would take of having the same style, right? I have an old PR in gocd/documentation#212

Correct. For the plugin to server communication, it should be in plugin-api.go.cd. @ketan knows that best.

arvindsv:
> One more thing I am lately thinking is that we should start maintaining actual JSON schemas for elements of config repo extension. Primarily to be able to develop out of Go tools for editing configuration. Once we have such schema I can easily validate it before committing and sending it to Go server. The way I see it is that config repo plugin can have complementary tool which does this:
1. Parse users repository in *whatever* format
2. Prepare JSON representation as would be sent to Go
2. Use JSON schema to validate it and show user the errors

Yes, I agree. This is very important. Otherwise, the feedback loop to see errors is too large (check-in, wait, see in server messages).

> Also similar idea, there could be a *Go domain validation library*, that would allow to catch errors which are not possible to validate with JSON schema. This library would be used outside of Go as dependency in the *editing* tools.

> And lastly - forget all above and instead we could make config repo validation API endpoint which would accept *anything* and return configuration errors to the user. Editing tool would need a configured access to particular Go server. But it has the benefits of being always up to date with the exact server that user has.

Yes, I like the second option (config repo validation API), rather than a domain validation library. But, would it validate against inter-pipeline dependencies? I guess it would. I mean, if there's a fetch artifact task from a pipeline in the main config, this validation API would be able to catch problems in that, I guess?

tomzo:
> external facing format of the JSON plugin (that users see).

Currently they are identical if I recall correctly. And I slightly dislike the current JSON format, because
 * it is quite bloated, I have some `gopipeline.json` files that have 600 lines when there are 2-3 stages and many jobs. Hard to grasp all at once. 
 * all quotations `"` and brackets `{}` are just noise and those extra commas after members and items in arrays are really annoying
 * format does not enforce correctness where it could. E.g. we require that stage names are unique. But then we have an array of stages where each stage has `"name":  "something"`, there is nothing at format level to prevent me from creating 2 elements with same name. A map would fit better here.
 * some form of `$ref` support would be nice to reduce duplication in files within one config repo. (Across repos too, but that is another bigger feature)
 * some keywords are too long to remember. E.g. `clean_working_directory`... I have to check it each time.
 * plugin relies on filename to determine what is inside the file (a pipe or an environment). So I am limited to file per pipeline, which is not ideal sometimes.
 * **JSON lacks comments** - putting comment next to config element explaining why it is configured like this would blow your mind, wouldn't it?

I have an early draft of YAML format, where I try to address above. I haven't started implementing it yet.
```yaml
#ci.yaml - any filename will do
pipeline: # tells plugin that here is one pipeline
  name: example-pipe
  group: g2
  label_template: "${mygit[:8]}"
  materials:
    mygit:  # this is the name of material
      type: git
      url: http://my.example.org/mygit.git
      branch: ci
    upstream:
      # type is optional here, material type is implied based on pipeline and stage fields
      type: dependency 
      pipeline: pipe2
      stage: test
  stages:
    build: # name of stage
      clean: true
      jobs:
        csharp: # name of the job
          resources: 
           - net45
          artifacts:
           - build:
               source: bin/
               destination: build
           - test:
               source: tests/
               destination: test-reports/
          tabs:
            report: test-reports/index.html
          tasks:
           - fetch:
               pipeline: pipe2
               stage: build
               job: test
               source: test-bin/
               destination: bin/
           - exec: # indicates type of task
               command: make
               args:
                - "VERBOSE=true"
          # shorthand for script-executor plugin
           - script: ./build.sh ci
```

> config repo validation API... But, would it validate against inter-pipeline dependencies? 

I think it could do exactly the same level of validation that we do in config merge operations. So yes, everything including the pipeline deps and even the current state of other repos and main XML. 
I think this is primary benefit to have API rather than domain validation library.

arvindsv:
Having it as a hash instead of a stage would make order not be strictly enforced in certain cases, I think.

Yes, a yaml plugin, and having the ability to have comments would be really nice.

> plugin relies on filename to determine what is inside the file (a pipe or an environment). So I am limited to file per pipeline, which is not ideal sometimes.

That's just a plugin-enforced limitation, which is not necessary, right? I, for one, like the one pipeline per file apporach. :) I like that they're separated.

> I think it could do exactly the same level of validation that we do in config merge operations. So yes, everything including the pipeline deps and even the current state of other repos and main XML. 
I think this is primary benefit to have API rather than domain validation library.

I agree.

PR #2382, opened 2016-07-03
#1133 fix server-side checkout destination in config repos

tomzo:
This is a critical bug fix for config repos in cases when many scm materials are used in a pipeline.

## What's broken

In case when we have a pipeline with 2+ scm materials, eg.
```json
  "materials" : [
    {
      "type" : "git",
      "name" : "git1",
      "destination" : "dest1",
      "url"  : "repo1.git"
    },
    {
      "type" : "git",
      "name" : "git2",
      "destination" : "dest2",
      "url"  : "url2"
    }
  ]
```
And one of above materials is also a config repo. Let's assume it's `git1` repository. 

Then expected behaviour is to have `git1` repository checkout **in flyweight directory**. Meaning something like 
```
/var/lib/go-server/pipelines/flyweight/3bdb04b2-44d0-4c15-9a5e-4159ffcec626
```
and **NOT in**
```
/var/lib/go-server/pipelines/flyweight/3bdb04b2-44d0-4c15-9a5e-4159ffcec626/dest1
```
I have observed checkouts done in both folders above on my prod server.

## Cause

As part of work on #1133 I added this `checkout` method on `MaterialPoller` interface
```
void checkout(T material, File baseDir, Revision revision, final SubprocessExecutionContext execCtx);
```
which internally calls `updateTo` in particular material implementations. The problem is that `updateTo`  was designed to be used on agents, so it performs checkout to `flyweight/destination` directory when `destination` is specified.

## Changes

I have added extra parameter `ignoreDestination` to `updateTo` method and few tests to assert that final directory is correct.
Then`MaterialPoller.checkout` is calling `updateTo` with `ignoreDestination=true`

### DIFF
--- a/common/src/com/thoughtworks/go/config/materials/perforce/P4Material.java
+++ b/common/src/com/thoughtworks/go/config/materials/perforce/P4Material.java
@@ -114,12 +114,12 @@ public class P4Material extends ScmMaterial implements PasswordEncrypter, Passwo
     }
 
     public List<Modification> latestModification(File baseDir, final SubprocessExecutionContext execCtx) {
-        P4Client p4 = getP4(baseDir);
+        P4Client p4 = getP4(execCtx.isServer() ? baseDir : workingdir(baseDir));
         return p4.latestChange();
     }
 
     public List<Modification> modificationsSince(File baseDir, Revision revision, final SubprocessExecutionContext execCtx) {
-        P4Client p4 = getP4(baseDir);
+        P4Client p4 = getP4(execCtx.isServer() ? baseDir : workingdir(baseDir));
         return p4.changesSince(revision);
     }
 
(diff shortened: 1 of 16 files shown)


common/src/com/thoughtworks/go/config/materials/PackageMaterial.java
```diff
@@ -118,7 +118,7 @@ public String getFolder() {
     }
 
     @Override
-    public void updateTo(ProcessOutputStreamConsumer outputStreamConsumer, File baseDir, RevisionContext revisionContext, SubprocessExecutionContext execCtx) {
+    public void updateTo(ProcessOutputStreamConsumer outputStreamConsumer, File baseDir, boolean ignoreDestination, RevisionContext revisionContext, SubprocessExecutionContext execCtx) {
```
zabil: With this change in https://github.com/gocd/gocd/commit/11e36ea7c1a9ee1cf5bcfcc4eedbdf6bf076c8af

You don't need the extra parameter the ```SubprocessExecutionContext``` has the ```isServer``` method that can be used to add this behaviour.

common/src/com/thoughtworks/go/config/materials/perforce/P4Material.java
```diff
@@ -139,23 +139,24 @@ protected void appendAttributes(Map<String, Object> parameters) {
...
         } catch (Exception e) {
             bomb(e.getMessage() + " " + outputConsumer.getStdError(), e);
         }
         return p4;
     }
 
-    public void updateTo(ProcessOutputStreamConsumer outputConsumer, File baseDir, RevisionContext revisionContext, final SubprocessExecutionContext execCtx) {
-        boolean cleaned = cleanDirectoryIfRepoChanged(workingdir(baseDir), outputConsumer);
+    public void updateTo(ProcessOutputStreamConsumer outputConsumer, File baseDir, boolean ignoreDestination, RevisionContext revisionContext, final SubprocessExecutionContext execCtx) {
+        File workingDir = ignoreDestination ? baseDir : workingdir(baseDir);
```
zabil: Can be changed to 
```execCtx.isServer() ? basedir: workingdir(baseDir)```

**Comment by tomzo** [comment]
@zabil thanks for suggestion, it is much shorter now. If you are OK with this then please merge.

File: common/src/com/thoughtworks/go/config/materials/perforce/P4Material.java [hunk 3]
```diff
@@ -114,12 +114,12 @@ public MaterialConfig config() {
     }
 
     public List<Modification> latestModification(File baseDir, final SubprocessExecutionContext execCtx) {
-        P4Client p4 = getP4(baseDir);
```
zabil: We can do the same thing here too right?

```
    public List<Modification> latestModification(File baseDir, final SubprocessExecutionContext execCtx) {
        P4Client p4 = getP4(execCtx.isServer() ? baseDir: workingdir(baseDir));
        return p4.latestChange();
    }
```

PR #2385, opened 2016-07-04
Remove empty javascript tag.

tootedom:
Hi there,

We are suffering performance issues when gocd is repainting all the pipelines on: /go/pipelines
One of the reasons for this is the number of pipelines we have which is about ~700 (give or take).  

- The number of elements on the page is huge:
-- document.getElementsByTagName("*").length : 46647

However, we noticed the number of script elements is also very large:
-- document.getElementsByTagName("script").length : 695

Looking through it, there's this empty javascript element.  Is there any reason for this?  Can we remove it?

thanks
/dom


ketan:
PR looks good, do you mind signing the [CLA](https://go.cd/cla)

tootedom:
Done

Cheers
/dom

ketan:
Thank you for your contribution :)

PR #2388, opened 2016-07-05
Moving command_snippets api to internal #000

maheshp:


arvindsv:
@maheshp: I've removed the specs related to the `show` action (in PR #2393) to fix the build. Please take a look and revert that if it's not right.

This PR doesn't say *why* the action was removed.

maheshp:
@arvindsv thanks for fixing it, the ```show``` endpoint is not used so I have removed for now.

Issue #599, opened 2014-09-29
for reporting, configure agent network interface or IP address


greenmoss:
I am running Docker on my GoCD agents, which makes them report their Docker IP address as the agent address to the GoCD server. However I want my agents to instead show their routable IP address on the server agent list.

Please update the [agent self-reporting code](https://github.com/gocd/gocd/blob/master/base/src/com/thoughtworks/go/util/SystemUtil.java#L69) to allow admins to configure either a network interface or an IP address. This could be done alongside other daemon settings such as GO_SERVER, etc.

nicolaasuni:
I am experiencing the same issue.

In my case the [getFirstLocalNonLoopbackIpAddress()](https://github.com/gocd/gocd/blob/master/base/src/com/thoughtworks/go/util/SystemUtil.java#L69) method is returning the IP address of the *docker0* interface, that is the same for all agents and is not the external routable IP address of the go-agent.

An host could have multiple network interfaces, and an interface could be bound to more than one IP address. Additionally you may have a load balancer or router between the go-agent and go-server.

This means that the strategy to obtain the go-agent IP address from the list of its network interfaces works only in specific conditions and fails in multiple scenarios.

My advice is to add a parameter in the go-agent configuration file to explicitly define the go-agent IP address to be used by the go-server. Something like:

GO_AGENT_IP=123.123.123.123

Some fallback mechanism can be put in place in case this parameter is not defined. For example trying to get the IP from the hostname if this is set in the DNS server (i.e. hostname --ip-address).

greenmoss:
Another strategy I have seen for determining the "primary" IP address is to use the IP address associated with the default route. Just an idea.

steven-collins-omega:
Me too. If I could vote this up I would.

The funny thing, I think, is that the agent IP address is reported correctly _until_ the admin actually enables that agent. Would it not be OK to interpret the admin enabling the agent as also confirming that the IP address is as expected, and therefore correct?

nicolaasuni:
As a collateral bug of this one, I noticed that sometimes the go-server raise the following exception that seems due to the change of IP between registering and enabling phase:
com.thoughtworks.go.server.service.AgentWithDuplicateUUIDException

nicolaasuni:
Why this issue has not been marked as a bug?

mattcl:
[...]

PR #2390, opened 2016-07-05
Connect to go server to determine the agent ip address

aj-jaswanth:


arvindsv:
@aj-jaswanth and @ketan: Wondering if the changes around this area can lead to or help with #599.

ketan:
This would make the reported ip address look correct. Later we can make it
configurable if its really needed (in cases where there is more than 1
network interface that can connect to the server, and you'd like to "bind"
which iPad dress the agent uses to connect to the server)

ketan:
There are merge conflicts and broken tests could you please rebase? Also which IDE are you on. The default intellij imports wshould not have caused you to submit a second PR to rejig imports.

PR #2391, opened 2016-07-05
Filter for pipeline history view

Mdevlin4:
Added an input box to the pipeline history view to filter pipelines by their label, comment, build cause, or revision.


zabil:
@Mdevlin4 thanks for putting this together!

To make the search box look and behave like how it is on the dashboard page we think it should look like this

By default
screen shot 2016-07-07 at 6 02 43 pm

With a filter (notice the x button to cancel)

screen shot 2016-07-07 at 6 02 43 pm withfilter

Tell us if you need help in getting it look like that.

Mdevlin4:
Thanks for the feedback! I made those changes to make it look and function more like the search box on the dashboard page. 

Here is the default look:
image

And the 'x' button to remove the filter:
image

Thanks again for the suggestions!

server/src/com/thoughtworks/go/server/controller/PipelineHistoryController.java
```diff
@@ -106,7 +108,10 @@ public ModelAndView list(@RequestParam("pipelineName")String pipelineName,
 
         PipelinePauseInfo pauseInfo = pipelinePauseService.pipelinePauseInfo(pipelineName);
         boolean hasBuildCauseInBuffer = pipelineScheduleQueue.hasBuildCause(CaseInsensitiveString.str(pipelineConfig.name()));
-        PipelineInstanceModels pipelineHistory = pipelineHistoryService.load(pipelineName, pagination, username, true);
+        PipelineInstanceModels pipelineHistory = (labelFilter == null || labelFilter.trim().isEmpty()) ?
```
zabil: Sorry, missed this earlier.
We have a utility method https://github.com/gocd/gocd/blob/master/base/src/com/thoughtworks/go/util/StringUtil.java#L161 for checking blank strings can you use that here?

Mdevlin4:
Thanks for the tip about the utility methods, I switched it to use the StringUtil class for the blank string check.

Issue #903, opened 2015-02-23
task plugin & secure fields might not be working


srinivasupadhya:
They might not be getting encrypted. Also, check [1](https://github.com/gocd/gocd/blob/2275d7e8d2e976231f99550f2c9eda9b01e02b41/config/config-api/src/com/thoughtworks/go/config/pluggabletask/PluggableTask.java#L130) & [2](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/service/tasks/PluggableTaskService.java#L70).

arvindsv:
@varshavaradarajan: Were you working on something similar to this, recently?

PR #2395, opened 2016-07-06
Pluggable tasks secure configurations are encrypted before saving in …

maheshp:
…config.xml #903

* Pluggable tasks with secure configurations were saved as plain text in config.xml, this fix is to store them as encrypted value
* Not migrating existing secure configurations saved as plain text,
  these configurations will be encrypted on update of the PluggableTask


arvindsv:
After the change to add `applyPluginMetadata`, is this line in the description correct?

> Not migrating existing secure configurations saved as plain text, these configurations will be encrypted on update of the PluggableTask

maheshp:
@arvindsv updated the description, after the ```applyPluginMetadata``` fix the existing secure configs which are un-encrypted would be encrypted.

PR #2400, opened 2016-07-07
Remove dead code and improve code where possible

aj-jaswanth:


agent/test/unit/com/thoughtworks/go/agent/AgentProcessParentImplTest.java
```diff
@@ -278,14 +275,10 @@ private File randomFile(final String pathname) {
...
-        FileOutputStream output = null;
-        try {
-            output = new FileOutputStream(agentJar);
+        try (FileOutputStream output = new FileOutputStream(agentJar)) {
             IOUtils.write(data, output);
         } catch (IOException e) {
-            throw new RuntimeException(e);
-        } finally {
-            IOUtils.closeQuietly(output);
+            e.printStackTrace();
```
arvindsv: I'm not reviewing this. Someone else will. But, I have a question about this line. Why? Why does it not throw any more? Shouldn't matter that much in this test, but still.

The only other part that seems a little risky here is the change to applicationContext, which needs to be verified (to make sure the timers are triggering as expected, and there is no difference between the old and the new). For instance, the behavior when one of the scheduled task runs takes more than the interval time - will it get queued vs. will it get run anyway.

agent/test/unit/com/thoughtworks/go/agent/AgentProcessParentImplTest.java
```diff
@@ -246,7 +244,7 @@ public void shouldNotDownloadPluginsZipIfPresent() throws Exception {
                 bootstrapper.run("launcher_version", "bar", getURLGenerator(), m(AgentProcessParentImpl.AGENT_STARTUP_ARGS, "foo bar  baz with%20some%20space"), context());
                 assertThat(new File(Downloader.AGENT_PLUGINS).lastModified(), is(expectedModifiedDate));
             } finally {
-                delete(pluginZip, agentLauncher);
+                delete(pluginZip, null);
```
zabil: change to ```delete(pluginZip)``` 
Same everywhere else this change is made

agent/test/unit/com/thoughtworks/go/agent/AgentProcessParentImplTest.java
zabil: I think it's ok to not wrap it around a runtime exception as the junit test where it's used has ```Exception``` on the signature.

arvindsv: Sure. That's fine too. This code swallows the exception, though.

zabil: Ahh I see. You are right 

@aj-jaswanth can you make it throw the exception?

aj-jaswanth:
@zabil @arvindsv Sure. Will make the necessary changes.

PR #2408, opened 2016-07-08
Upgrade logging jars

ketan:
* slf4j from 1.7.12 to 1.7.21
* commons logging from 1.1.1 to 1.2


Tpbrown:
... I'm pretty sure if you replace log4j with logback GoCD server will run in the JRE compact3 profile ;-)

(The compact3  ref impl is about half the size of the full JRE)

ketan:
> ... I'm pretty sure if you replace log4j with logback GoCD server will run in the JRE compact3 profile ;-)

One tiny step at a time, the idea is to move towards using slf4j and allow users to choose a logging backend of their choice, with some carrot to help steer them towards logback :)

PR #2409, opened 2016-07-08
Upgrade ant version 1.7.1 to 1.9.7

ketan:


ketan:
Merging myself, since this is purely a test dependency

PR #2410, opened 2016-07-09
Make FileUtil delegate to FileUtils instead.

ketan:


varshavaradarajan:
These changes look okay to me. Will trigger the PR build.

Issue #2284, opened 2016-06-09
Use unique hashcode of object as etag for APIs


varshavaradarajan:
##### Issue Type

- Feature proposal

##### Summary
We use etags to avoid the conflicts that arise while making simultaneous config change requests pertaining to specific objects like Pipeline Config.

Currently, we compute the etag for a specific object like this:
```
Digest::MD5.hexdigest(JSON.generate(object))
```

The above is done twice,  once when a ```GET``` is performed, And once during ```UPDATE``` to check if the header ```If-Match``` has the correct etag. So, the computation of etag incurs an overhead and degrades the performance of the API for large objects. 

##### Possible Fix

Since, each object has a unique hashcode associated with it, we can potentially use the hashcode as etag. 

Requires further analysis

arvindsv:
Hashcode is never guaranteed to be unique. Be careful.

ketan:
`object.hashcode()`s uniqueness is(should?) not going to be any different than `Digest::MD5.hexdigest(JSON.generate(object))`

arvindsv:
We are talking about Java hashCode, right?

ketan:
yes, java hashcode

arvindsv:
The contract of hashCode is that it is (paraphrased): "Try and keep it different between objects, but if it's the same object, return the same hashCode ... *please*". Right?

So, given objects obj1 and obj2, this is what happens (and is possible and valid):

```
obj1.hashCode().equals(obj2.hashCode()) == true
obj1.equals(obj2) == false
```

This should never happen:

```
Given: obj1.equals(obj2) == true
Then: obj1.hashCode().equals(obj2.hashCode()) == false  [Should not happen]
```

arvindsv:
The point is that hashCode can be generated without seeing all the parts of the object. "return 1" is a very valid hashCode. Not recommended.

ketan:
I'm suggesting we build hashcode by looking at all parts of the object, in order to save on performance. Unless you have a better idea.

arvindsv:
If you're sure that you're using all parts, then sure, there is no real difference between equals and hashCode. Then, you'll be ok. But, you need to ensure that all hashCode, all the way down the chain follows that. They need not. [Here's a test](https://github.com/gocd/gocd/blob/97905f733eec520fa25585d20cb31262a35240f8/common/test/unit/com/thoughtworks/go/domain/MaterialRevisionTest.java#L224) which is written to ensure that it doesn't, for instance.

[...]

PR #2416, opened 2016-07-11
 Etag computation (#2284)

varshavaradarajan:
#2284


server/src/com/thoughtworks/go/config/update/UpdateSCMConfigCommand.java
```diff
@@ -18,33 +18,62 @@
...
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
+        return isUserAuthorized() && isRequestFresh(cruiseConfig);
+    }
+
+    private boolean isRequestFresh(CruiseConfig cruiseConfig) {
+        SCM existingSCM = findSCM(cruiseConfig);
+        return entityHashingService.md5ForEntity(existingSCM, existingSCM.getName()).equals(md5);
+    }
```
GaneshSPatil: can we add a message to the result if it is stale request
something like ` result.stale()`

varshavaradarajan:
To be discussed and done: 

The rails method ```stale?``` ends up computing the md5 on any object passed. Currently, we pass a string computed by the ```EntityHashingService``` to this method and rails computes the md5 on the string passed. This results in md5 computation twice: once on the domain object and once on the consequent md5 string. But, this is not a performance constraint.

Possible Fix: To have a our own implementation of the stale method on the rails side and set the etag header on the the response manually.

server/test/unit/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
```diff
@@ -1,58 +1,66 @@
...
     @Test
-    public void shouldGenerateMD5ForAPipelineConfig() {
-        PipelineConfig pipelineConfig = mock(PipelineConfig.class);
-
-        when(goConfigService.pipelineConfigNamed(any(CaseInsensitiveString.class))).thenReturn(pipelineConfig);
-        when(entityDigest.md5ForPipeline(pipelineConfig)).thenReturn("pipeline_config_md5");
+    public void shouldThrowAnExceptionWhenObjectIsNull() {
+        thrown.expect(NullPointerException.class);
+        entityHashingService.md5ForEntity(null, null);
+    }
```
GaneshSPatil: is there any scenario where `domainObject` can be null?

varshavaradarajan: Yes. Suppose there is no pipeline config named "foo". That was the intention of this test. But of course, the respective entity command would have already checked for such a case. Hopefully. :)

GaneshSPatil:
can we write some integration tests for `PluggableScmService`?
as we might miss some cases such as the exception handling part in the `PluggableScmService.update` method which is not covered in the unit tests written for specific commands.

server/test/unit/com/thoughtworks/go/config/update/UpdateSCMConfigCommandTest.java
```diff
@@ -80,9 +85,46 @@ public void shouldUpdateAnExistingSCMWithNewValues() throws Exception {
...
+    @Test
+    public void shouldNotContinueWithConfigSaveIfUserIsUnauthorized() {
+        when(goConfigService.isUserAdmin(currentUser)).thenReturn(false);
+        when(goConfigService.isGroupAdministrator(currentUser.getUsername())).thenReturn(false);
+
+        SCM updatedScm = new SCM("id", new PluginConfiguration("plugin-id", "1"), new Configuration(new ConfigurationProperty(new ConfigurationKey("key1"),new ConfigurationValue("value1"))));
+        UpdateSCMConfigCommand command = new UpdateSCMConfigCommand(updatedScm, pluggableScmService, goConfigService, currentUser, result, "md5", entityHashingService);
+
+        assertThat(command.canContinue(cruiseConfig), is(false));
+    }
```
GaneshSPatil: assert on the result object --
```
assertFalse(result.isSuccessful())
assertThat(result.toString(), containsString("UNAUTHORIZED_TO_EDIT"));
```

server/test/unit/com/thoughtworks/go/config/update/UpdateSCMConfigCommandTest.java
```diff
@@ -80,9 +85,46 @@ public void shouldUpdateAnExistingSCMWithNewValues() throws Exception {
...
+    public void shouldNotContinueWithConfigSaveIfRequestIsNotFresh() {
+        when(goConfigService.isUserAdmin(currentUser)).thenReturn(true);
+        when(goConfigService.isGroupAdministrator(currentUser.getUsername())).thenReturn(false);
+
+        SCM updatedScm = new SCM("id", new PluginConfiguration("plugin-id", "1"), new Configuration(new ConfigurationProperty(new ConfigurationKey("key1"),new ConfigurationValue("value1"))));
+        updatedScm.setName("material");
+        when(entityHashingService.md5ForEntity(cruiseConfig.getSCMs().find("id"), "material")).thenReturn("another-md5");
+        UpdateSCMConfigCommand command = new UpdateSCMConfigCommand(updatedScm, pluggableScmService, goConfigService, currentUser, result, "md5", entityHashingService);
+
+        assertThat(command.canContinue(cruiseConfig), is(false));
```
GaneshSPatil: assert on the result object --
```
assertFalse(result.isSuccessful())
assertThat(result.toString(), containsString("STALE_RESOURCE_CONFIG"));
assertThat(result.toString(), containsString(scm_material_config_name)));
```

server/src/com/thoughtworks/go/config/update/UpdateSCMConfigCommand.java
```diff
@@ -18,33 +18,67 @@
...
         }
-        else {
+        return true;
+    }
+
+    private SCM findSCM(CruiseConfig modifiedConfig) {
+        SCMs scms = modifiedConfig.getSCMs();
+        SCM existingSCM = scms.find(globalScmConfig.getSCMId());
+        if (existingSCM == null) {
             result.notFound(LocalizedMessage.string("SCM_MATERIAL_NOT_FOUND"), HealthStateType.notFound());
```
ketan: Use `RESOURCE_NOT_FOUND` instead, and remove `SCM_MATERIAL_NOT_FOUND` if it's not used elsewhere.

server/test/unit/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
```diff
@@ -1,58 +1,66 @@
...
 
-        assertThat(md5ForPipeline, is("pipeline_config_md5"));
-        verify(goCache).put("GO_PIPELINE_CONFIGS_ETAGS_CACHE", "p1", "pipeline_config_md5");
+        assertThat(entityHashingService.md5ForEntity(pipelineConfig, "P1"), is(CachedDigestUtils.md5Hex(xml)));
     }
 
     @Test
     public void shouldReturnCachedMD5IfPresent() {
-        when(goCache.get("GO_PIPELINE_CONFIGS_ETAGS_CACHE", "p1")).thenReturn("pipeline_config_md5");
+        when(goCache.get("GO_ETAG_CACHE", "p1")).thenReturn("pipeline_config_md5");
```
ketan: Use the constant in `EntityHashingService`?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pluggable_scms_controller.rb
```diff
@@ -28,42 +28,44 @@ def index
...
-      def check_for_stale_request
-        if (request.env["HTTP_IF_MATCH"] != "\"#{get_etag_for_material(params[:material_name])}\"")
-          render_message("Someone has modified the global SCM '#{params[:material_name]}'. Please update your copy of the config with the changes.", :precondition_failed)
+      def handle_create_or_update_response(result, updated_scm)
+        if result.isSuccessful
+          json = ApiV1::Scms::PluggableScmRepresenter.new(updated_scm).to_hash(url_builder: self)
+          response.etag = [get_etag_for_scm_object(updated_scm.getName)]
+          render DEFAULT_FORMAT => json
+        else
+          json = ApiV1::Scms::PluggableScmRepresenter.new(updated_scm).to_hash(url_builder: self)
```
ketan: Extract this out of the loop?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pluggable_scms_controller.rb
```diff
@@ -28,42 +28,44 @@ def index
...
+          render_http_operation_result(result, {data: json})
         end
       end
 
-      def get_etag_for_material(material_name)
-        hash = get_scm_hash(material_name)
-        get_etag_for_scm_json(hash)
+      def check_for_stale_request
```
ketan: Why is this implementation different from the one in `environments_controller`?

server/webapp/WEB-INF/rails.new/lib/spring.rb
```diff
@@ -38,7 +38,7 @@ def self.context
           @context =  load_context
           begin
             self.bean("viewRenderingService").registerRenderer(com.thoughtworks.go.plugins.presentation.Renderer.ERB, ErbRenderer.new) if @context
-            self.bean("entityHashingService").initializeWith(PipelineDigest.new) if @context
+            # self.bean("entityHashingService").initialize if @context
```
ketan: Delete this?

server/test/unit/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
varshavaradarajan: This would involve changing the access specifier from private to protected. Is that necessary for the test's sake?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/pluggable_scms_controller.rb
varshavaradarajan: Not really, just a change in how the message and status code is rendered, which was a refactoring change you had asked me to do. Will refactor to reflect the same in environments_controller, ok?

ketan: Since we're doing `result.stale` using a generic enough message defined via `STALE_RESOURCE_CONFIG`, lets use `render_result` at both places.

My recommendation earlier was so as to avoid having to declare another internationalization string for something that was used in one place.

PR #2421, opened 2016-07-12
Detect linux distributions properly

aj-jaswanth:


ketan:
Can you explain the reason for doing this? I can see that you're shelling out to external programs to perform detection, some of these may not be available in constrained environments (docker containers for e.g.)

aj-jaswanth:
The main reason I am using python is that, it already has a lot of code in platform module that is required to detect a distribution name. Any OS using YUM will have python installed. In Ubuntu containers python is not installed. Have to use /etc/os-release or lsb-release as a fallback. What do you think about it?

ketan:
> Can you explain the reason for doing this?

aj-jaswanth:
The simplest reason would be Linux is not an operating system. Its just a kernel. It would be helpful to see what exactly the agent is running. In our project we have around 15 agents. Most of them are CentOS and a few Ubuntu. Agents are using different versions of CentOS between 6.2 to 6.7. We have a hard time finding the version unless we manually do SSH. It helps to have it on the agents page, so it will be available for anyone to see.

ketan:
> In our project we have around 15 agents. Most of them are CentOS and a few Ubuntu. Agents are using different versions of CentOS between 6.2 to 6.7. We have a hard time finding the version unless we manually do SSH. It helps to have it on the agents page, so it will be available for anyone to see.

That's a legit use case, and I see value in having improved OS/distro detection. However what's a no-go is to have to rely on shelling out, because it leaves out the users on docker (plus GoCD is going to be doing a push for elastic agents soon, so docker is a real problem even now.)

Reading from `/etc/*-release` or `/etc/lsb-release` is an option, but I'm not sure how many distributions that'll leave behind.

Would you consider using a 3rd party library to perform OS detection?  A quick google search shows up https://github.com/aurbroszniowski/os-platform-finder or https://github.com/redpois0n/oslib.

aj-jaswanth:
Sure. I will try to find a better solution.

aj-jaswanth:
@ketan There are no proper 3rd party libraries available for this task. Even oslib you mentioned is not a library. Added fallback to use /etc/os-release file to get PRETTY_NAME property.

aj-jaswanth:
@ketan Added fallback to /etc/os-release file. It is present in Ubuntu as well as Alpine containers also. Using PRETTY_NAME property to get os name. Using python covers fedora/centos operating systems as YUM is written in python :smile:. Similar bet is that all variants of Debian also have python installed except if they are meant for containers. In which case /etc/os-release is guaranteed to be present.

common/test/unit/com/thoughtworks/go/util/OperatingSystemTest.java
```diff
@@ -12,40 +12,20 @@
...
 import java.util.Collections;
 import java.util.List;
 
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
 
-import com.thoughtworks.go.util.OperatingSystem;
-import org.junit.Test;
-
 public class OperatingSystemTest {
```
varshavaradarajan: This file is not necessary to be kept, is it? Now that there are no tests in it.

aj-jaswanth: Yes correct. Removing it.

base/src/com/thoughtworks/go/util/OperatingSystem.java
```diff
@@ -12,44 +12,65 @@
...
+
+public class OperatingSystem {
+    private static final String OS_FAMILY_NAME = System.getProperty("os.name");
+    public static final String WINDOWS = "Windows";
+    private static String OS_COMPLETE_NAME = detectCompleteName();
+
+    private static String detectCompleteName() {
+        String[] command = {"python", "-c", "import platform;print(platform.linux_distribution())"};
+        try {
+            Process process = Runtime.getRuntime().exec(command);
```
varshavaradarajan: So, we have something called ```CommandLine#run``` and ```CommandLine#runOrBomb``` to execute commands. May be we could use that? @ketan, what do you think?

ketan: `CommandLine` belongs to the common module, so can't be used inside base :(

varshavaradarajan:
Looks good to me. 👍

Tpbrown:
Or /etc/issue ?

aj-jaswanth:
/etc/issue is certainly useful but, it is used as welcome message for user login. It will be difficult to parse, as for example alpine gives 
"Welcome to Alpine Linux 3.4
Kernel \r on an \m (\l)"

zabil:
@aj-jaswanth we'd like to get this merged for 16.9. Can you rebase this when you get time?

aj-jaswanth:
Sure. I am currently on vacation. Will do it on Monday.

aj-jaswanth:
@zabil Fixed conflicts. I had to remove a test that depends on the environment in which it runs. https://github.com/aj-jaswanth/gocd/commit/eabebb7d352374068b52372c3c2dceda8a238b7c#diff-3fa5d4639f1444263e6d5b370ea04f18L145

PR #2422, opened 2016-07-12
Added partial representation for pluggable scm index api.

varshavaradarajan:
This is done keeping the config SPA in mind. An index call need not have the full SCM object, but just the details necessary to make a Get SCM API call and subsequently the update SCM call. 

The index call now returns the following for all SCMs:
1. id
2. name
3. plugin_metadata

The show action returns the above along with ```auto_update``` and ```configuration``` 

@maheshp, Please review the changes

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/scms/pluggable_scm_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/scms/pluggable_scm_representer.rb
@@ -16,26 +16,10 @@
 
 module ApiV1
   module Scms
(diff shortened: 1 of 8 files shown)


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/scms/pluggable_scm_representer.rb
```diff
@@ -16,26 +16,10 @@
 
 module ApiV1
   module Scms
-    class PluggableScmRepresenter < BaseRepresenter
```
zabil: The convention is 'summary',
Can you rename this to PluggableScmSummaryRepresenter?

Issue #2354, opened 2016-06-27
Pipeline Config API: support for material whitelist


rajiesh:
##### Issue Type

- Feature enhancement

##### Summary

Material whitelist is now allowed after the this [feature](https://github.com/gocd/gocd/pull/2144) was added in 16.6 release. Pipeline config API need to support it

PR #2426, opened 2016-07-12
Added pipeline-config API material whitelist support

GaneshSPatil:
added new api_v2 for pipeline config

old Material filter : 
```
{
  filter: {
    ignore: []
  }
}
```

Material filter with whitelist support : 
```
{
  filter: {
    ignore: []
  },
  invert_filter: true
}
```

not providing `invert_filter` will set it to `false` by default


GaneshSPatil:
fixes -- #2354

PR #2427, opened 2016-07-12
Revision Search on Dashboard Screen

Mdevlin4:
I added a search box to the dashboard screen that lets users search for the value stream map associated with a particular revision. The revision search box is positioned in the same place as the pipeline search box, and I added a toggle to switch between the two. 

Here is a picture of the toggle to select between the pipeline search and the revision search:
image

And here is a picture of how the revision search results are displayed. Selecting one of the options takes you to the material value stream map associated with that revision:
image

I also added a link to the value stream map from the material revisions popup dialog:

image

Any feedback on the positioning/appearance or the functionality of the revision search would be greatly appreciated, and I hope these features are useful!


zabil:
@Mdevlin4 thanks for sending this, but we won't be merging this as we plan to deprecate the existing dashboard page.

We will consider the functionality that you've added in this PR when we come up with a new design.
You can keep track of that here at #2819

PR #2431, opened 2016-07-13
Display the correct footer on the login page. Render the page without footer for modals.

zabil:
The login page displayed the old footer. 
screen shot 2016-07-13 at 8 15 56 pm

There was also a problem with the footer getting rendered on modal windows
screen shot 2016-07-13 at 8 16 23 pm

This fix updates the login page to
screen shot 2016-07-13 at 8 18 17 pm

And displays the modal without including the footer (improves the modal load time as it now fetched and renders only the fragment).
screen shot 2016-07-13 at 8 19 57 pm

### DIFF
--- a/server/webapp/WEB-INF/vm/auth/login.vm
+++ b/server/webapp/WEB-INF/vm/auth/login.vm
@@ -1,5 +1,5 @@
 #*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


server/webapp/WEB-INF/vm/auth/login.vm
```diff
@@ -15,177 +15,164 @@
...
+            Licensed under <a href="http://www.apache.org/licenses/LICENSE-2.0" target="_blank">Apache License, Version
+              2.0</a>.<br>
+            Go includes <a href="/go/NOTICE/cruise_notice_file.pdf" target="_blank">third-party software</a>.
+            Go Version: #include("admin/admin_version.txt.vm").
+          </p>
+
+        </div>
+        <div class="small-12 medium-6 large-4 columns">
+                    <span class="inline-list social">
+                        <a href="http://twitter.com/goforcd" title="twitter" class="twitter"></a>
```
aj-jaswanth: Can we use https instead of http?

zabil: Makes sense. Will modify this.

PR #2433, opened 2016-07-14
Environment-Update Commands Implements EntityConfigUpdateCommand

GaneshSPatil:
Moved Environment-operation commands to implement EntityConfigUpdateCommand from UpdateConfigCommand

### DIFF
--- a/server/src/com/thoughtworks/go/config/update/AddEnvironmentCommand.java
+++ b/server/src/com/thoughtworks/go/config/update/AddEnvironmentCommand.java
@@ -1,24 +1,95 @@
+/*************************GO-LICENSE-START*********************************
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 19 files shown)


server/src/com/thoughtworks/go/config/update/AddEnvironmentCommand.java
```diff
@@ -1,24 +1,82 @@
...
+    @Override
+    public void update(CruiseConfig preprocessedConfig) throws Exception {
+        preprocessedConfig.addEnvironment(environmentConfig);
+    }
+
+    @Override
+    public boolean isValid(CruiseConfig preprocessedConfig) {
+        for (GoConfigValidator validator : VALIDATORS) {
+            try {
+                validator.validate(preprocessedConfig);
```
varshavaradarajan: The EnvironmentPipelineValidator does not seem to check if pipeline is a part of another environment. A pipeline can only belong to one environment. Can you please confirm or point out to where this validation happens?

server/src/com/thoughtworks/go/config/update/UpdateEnvironmentCommand.java
```diff
@@ -1,33 +1,83 @@
...
+        BasicCruiseConfig.clearErrors(oldEnvironmentConfig);
+    }
+
+    @Override
+    public EnvironmentConfig getPreprocessedEntityConfig() {
+        return newEnvironmentConfig;
+    }
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
```
varshavaradarajan: Just adding this so we won't forget. :) After #2416 is merged, we can make the change to check for a stale request as well.

PR #2443, opened 2016-07-15
Remove code that's needed for upgrade from really ancient gocd versions (12.3.1)

ketan:


ketan:
This was used for migration of log4j.properties file by adding specific appenders for shine and rails. I don't see the need for this code anymore since we've had the right log files for quite some time now.

varshavaradarajan:
@ketan, Done reviewing it. Seems alright. Just have to update the copyright on LoggingValidatorTest. I will trigger the PR build for this.

PR #2446, opened 2016-07-18
Minor debian package fixes

ketan:


installers/linux-shared/after-remove.sh.erb
```diff
@@ -17,9 +17,9 @@
 -%>
 
 <% if type == 'rpm' -%>
-    rm -rf /var/lib/<%= name %> /var/log/<%= name %> /etc/defaults/<%= name %>
+    rm -rf /var/lib/<%= name %> /var/log/<%= name %>
 <% elsif type == 'deb' -%>
     update-rc.d <%= name %> remove
-    rm -rf /var/lib/<%= name %> /var/log/<%= name %> /etc/defaults/<%= name %>
+    rm -rf /var/lib/<%= name %> /var/log/<%= name %>
```
varshavaradarajan: Whether it is rpm or deb, ```rm -rf /var/lib/<%= name %> /var/log/<%= name %>``` is the same for both. Perhaps you could refactor this?

installers/linux-shared/after-upgrade.sh.erb
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#########################################################################
+-%>
+
+<% if type == 'rpm' -%>
+    # do nothing
```
varshavaradarajan: This if check is not required. It improves readability. But can't we add a comment instead to show that if the type is rpm, we do nothing?

varshavaradarajan: Ah, never mind. :) We have to throw the unsupported package exception. So we end up making this if check anyway.

installers/linux-shared/after-remove.sh.erb
ketan: Intentional, makes it easy to read code though the crazy if-else blocks.

arvindsv: Especially since `/var/lib/go-server/db` exists, should after-remove remove `/var/lib/go-server`? Maybe yes on a purge. Is this called only on purge?

varshavaradarajan: It is there in the postuninstall script. So, it is called everytime an uninstall happens and not just on purge. @ketan told me "dirs are only removed by the package manager, if they are empty
so it’ll attempt to remove files first, then dirs if they are empty
if they are non-empty, it’s left to the discretion of the post uninstall script to do whatever it feels like"

arvindsv: Compared to how the old installers were, this is dangerous (unless update-remove.sh is called properly elsewhere - haven't checked). The earlier script (in 16.6 deb installer) was:

```
if [ "$1" = "purge" ] ; then
    update-rc.d go-server remove >/dev/null || exit $?

    rm -fr /var/lib/go-server
    rm -fr /etc/go
    rm -fr /usr/share/go-server
    rm -fr /var/run/go-server
    rm -fr /var/log/go-server
fi
```

PR #2452, opened 2016-07-19
Log error on the agent logs when artifact publishing fails.

zabil:
This change makes the artifact publisher log all exception to the agent logs.

### DIFF
--- a/common/src/com/thoughtworks/go/domain/DefaultJobPlan.java
+++ b/common/src/com/thoughtworks/go/domain/DefaultJobPlan.java
@@ -32,6 +32,8 @@ import com.thoughtworks.go.config.TestArtifactPlan;
 import com.thoughtworks.go.util.GoConstants;
 import com.thoughtworks.go.util.command.EnvironmentVariableContext;
 import com.thoughtworks.go.work.DefaultGoPublisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DefaultJobPlan implements JobPlan {
 
@@ -48,6 +50,7 @@ public class DefaultJobPlan implements JobPlan {
     private boolean fetchMaterials = StageConfig.DEFAULT_FETCH_MATERIALS;
     private boolean cleanWorkingDir = StageConfig.DEFAULT_CLEAN_WORKING_DIR;
 
+    private static final Logger LOG = LoggerFactory.getLogger(DefaultJobPlan.class);
 
     //FOR IBATIS
     protected DefaultJobPlan() {
@@ -104,6 +107,7 @@ public class DefaultJobPlan implements JobPlan {
             try {
                 artifactPlan.publish(goPublisher, workingDirectory);
             } catch (Exception e) {
+                LOG.error(e.getMessage(), e);
                 failedArtifact.add(artifactPlan);
             }
         }
--- a/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 ThoughtWorks, Inc.
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


common/src/com/thoughtworks/go/domain/DefaultJobPlan.java
```diff
@@ -32,6 +32,8 @@
 import com.thoughtworks.go.util.GoConstants;
 import com.thoughtworks.go.util.command.EnvironmentVariableContext;
 import com.thoughtworks.go.work.DefaultGoPublisher;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
```
arvindsv: Use slf instead?

arvindsv:
One thing is, I *think* this will log duplicate errors. Some are logged at a lower level too (ones inside the publish which does it three time). Might be worth a quick look.

zabil:
@arvindsv common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java I had meant to use Log4j some organise imports made intellij use commons.≈

You can't unit test slf4j because it uses an underlying logger. So the tests will have to be on log4j. 

Will check the duplication. I see errors written to the publisher. The problem is not all exceptions are caught. Might have scan and check for specific exception or figure out what path is missing.

arvindsv:
It's possible, but not easy. But, I think `LogFixture.startListening()` should be used here. Once we decide to move away from log4j, it'll help. What do you think? Something like this?

```diff
diff --git a/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java b/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java
index 0356534..478a97b 100644
--- a/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java
+++ b/common/test/unit/com/thoughtworks/go/domain/DefaultJobPlanTest.java
@@ -16,42 +16,38 @@
 
 package com.thoughtworks.go.domain;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
-
 import com.thoughtworks.go.config.*;
 import com.thoughtworks.go.util.FileUtil;
+import com.thoughtworks.go.util.LogFixture;
 import com.thoughtworks.go.util.TestFileUtil;
 import com.thoughtworks.go.util.command.EnvironmentVariableContext;
 import com.thoughtworks.go.work.GoPublisher;
 import org.apache.commons.io.FileUtils;
-import org.apache.log4j.Appender;
 import org.apache.log4j.Level;
-import org.apache.log4j.spi.LoggingEvent;
-import org.hamcrest.core.Is;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
-import org.mockito.ArgumentCaptor;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 import static com.thoughtworks.go.helper.EnvironmentVariablesConfigMother.env;
 import static com.thoughtworks.go.utils.SerializationTester.serializeAndDeserialize;
-import static org.apache.log4j.LogManager.getRootLogger;
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsNull.nullValue;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.spy;
 
 public class DefaultJobPlanTest {
 
     private File workingFolder;
     private File toClean;
+    private LogFixture logFixture;
 
     @Before
     public void setUp() throws IOException {
@@ -59,10 +55,14 @@ public class DefaultJobPlanTest {
         File file = new File(workingFolder, "cruise-output/log.xml");
         file.getParentFile().mkdirs();
         file.createNewFile();
+
+        logFixture = LogFixture.startListening(Level.ERROR);
     }
 
     @After
     public void tearDown() {
+        logFixture.stopListening();
+
         FileUtil.deleteFolder(workingFolder);
         FileUtils.deleteQuietly(toClean);
     }
@@ -242,11 +242,6 @@ public class DefaultJobPlanTest {
 
     @Test
     public void shouldLogExceptionWhenUploadFails() throws Exception {
-        Appender appender = mock(Appender.class);
-        getRootLogger().addAppender(appender);
-
-        ArgumentCaptor<LoggingEvent> argumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
-
         ArtifactPlans artifactPlans = new ArtifactPlans();
         DefaultJobPlan plan = new DefaultJobPlan(new Resources(), artifactPlans, new ArtifactPropertiesGenerators(), -1,
                 null);
@@ -264,14 +259,6 @@ public class DefaultJobPlanTest {
         } catch (Exception ignore) {
         }
 
-        verify(appender).doAppend(argumentCaptor.capture());
-        LoggingEvent event = argumentCaptor.getValue();
-
-        getRootLogger().removeAppender(appender);
-
-        assertThat(event.getRenderedMessage(), is("Error while trying to upload"));
-        assertThat(event.getThrowableInformation().getThrowable(), Is.<Throwable>is(toBeThrown));
-        assertThat(event.getLoggerName(), is(DefaultJobPlan.class.getName()));
-        assertThat(event.getLevel(), is(Level.ERROR));
+        assertThat(logFixture.allLogs(), is("Error while trying to upload"));
     }
 }
```

arvindsv:
Of course, that will still be on log4j, as you said. But, not directly.

arvindsv:
About the duplicate logs, I understand. I'd rather have duplicate logs than no logs.

zabil:
I am closing this as we do log the failures here.
https://github.com/gocd/gocd/blob/27511b3d02e0ef40b73d7140b47589811a0dd937/common/src/com/thoughtworks/go/publishers/GoArtifactsManipulator.java#L111

Issue #2455, opened 2016-07-19
Change the entries in server's Info.plist to reference to new location of main class


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

Currently, the Info.plist has the following

```
<key>ClassPath</key>
 <string>$APP_PACKAGE/go.jar</string>
<key>MainClass</key>
<string>com.thoughtworks.go.server.util.GoMacLauncher</string>
```
GoMacLauncher has been moved from ```com.thoughtworks.go.server.util.GoMacLauncher``` to ```com.thoughtworks.go.server.launcher.GoMacLauncher``` 

Additionally, the built package no longer has ```go.jar``` in ```Go\ Server.app/Contents/Resources```. But, it has a ```server-launcher.jar```. Appropriate corrections need to be made in ```Info.plist``` and ```GoMacLauncher```

ketan:
@varshavaradarajan I've pushed a fix, mind verifying?

varshavaradarajan:
@ketan, Server starts now without any issues

PR #2456, opened 2016-07-19
Changing the server's Info.plist to reflect the correct entries.

varshavaradarajan:
#2455 , @ketan, please take a look at it.

### DIFF
--- a/installers/go-server/osx/Info.plist
+++ b/installers/go-server/osx/Info.plist
@@ -49,7 +49,7 @@
             <key>ClassPath</key>
             <string>$APP_PACKAGE/go.jar</string>
             <key>MainClass</key>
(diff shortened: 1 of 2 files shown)


varshavaradarajan:
Well, its go.jar in the generic zip. I suppose we have to change it to go.jar here as well.

arvindsv:
Yes, please. I guess [this](https://github.com/gocd/gocd/blob/280b26c4a89023e25ac06b7e517c0819f041418a/installers/osx.gradle#L44) gets changed: `$1.$2` => `go.$2`?

ketan:
I've fixed it so as to change the jar to `go.jar`, and partially rolled back @varshavaradarajan's change.

Issue #2450, opened 2016-07-19
Remove /var/lib/go-server only on apt-get --purge remove and not just apt-get remove


arvindsv:
See comment: https://github.com/gocd/gocd/pull/2446/files#r71135322

The idea is that in 16.6, an `apt-get remove go-server` would not have removed the DB. Now, it does. `apt-get --purge remove go-server` in 16.6 and below would have removed DB - and that’s what I would expect.

arvindsv:
Looking [this](https://github.com/jordansissel/fpm/blob/398eeafde92fc2b62a93d3b92b3d8c9348fee23e/lib/fpm/package/deb.rb#L412-L414) and then [this](https://github.com/jordansissel/fpm/blob/398eeafde92fc2b62a93d3b92b3d8c9348fee23e/templates/deb/postrm_upgrade.sh.erb#L23), I'd remove the whole of after-remove.sh.erb. As you [can see](https://github.com/jordansissel/fpm/blob/398eeafde92fc2b62a93d3b92b3d8c9348fee23e/templates/deb/postrm_upgrade.sh.erb#L26-L31), it does not do anything on purge, but just on remove. So, with this script, it'll remove the DB, etc. when you do an `apt-get remove go-server`.

moritz:
@arvindsv according to the apt-get man page, puring deletes *configuration* files, but data files (like the artifact repository, pipeline history) shouldn't be deleted.

As a comparison,  if you `apt-get remove --purge mysql-server-5.5` on Debian Jessie, it doesn't delete `/var/lib/mysql`, which contains the acutal data files.

So I think package uninstallation shouldn't remove ` /var/lib/go-server` completely. In fact one could argue that the XML config isn't so much configuration of the application but rather contents, and should remain as well.

arvindsv:
I agree.

arvindsv:
Needs to be changed:

1. Remove the whole of the purge part.

2. Change: `invoke-rc.d go-agent stop` to `invoke-rc.d go-agent stop || true` to keep it consistent with 16.6. Same with server. Otherwise, analyze why it was done before.

3. Make debian installer file names follow the 16.6 naming format (rather than the one which has _all.deb). Maybe consider changing it next release to the correct Debian convention (ie, _all.deb).

4. Next release: Remove: "Conflicts: cruise-server" and "Replaces: cruise-server".

PR #2471, opened 2016-07-21
Remove DEB/RPM installer purge script #2450

arvindsv:
* Also, change debian installer filename to be consistent with 16.6.

Relates to #2450.


installers/linux-shared/before-remove.sh.erb
```diff
@@ -20,6 +20,6 @@
     /etc/init.d/<%= name %> stop 2>&1
     /sbin/chkconfig --del <%= name %>
 <% elsif type == 'deb' -%>
-    invoke-rc.d <%= name %> stop
+    invoke-rc.d <%= name %> stop || true
```
ketan: @arvindsv — Should we be doing a `update-rc.d remove` — the equivalent of `chkconfig --del`?

I remember not adding it here, because it was handled in the `after-remove` script for debian packages.

arvindsv: Yeah, you're right! Fixed. Thanks!

PR #2474, opened 2016-07-22
Leave partials AS-IS when building on windows.

ketan:


clearly:
How far away is 16.7? Would love to integrate it this weekend.

zabil:
Early next week.

PR #2481, opened 2016-07-25
Modify the git command to list contributors

zabil:
Scrape the ```git shortlog``` command to get correct list of contributors

### DIFF
--- a/installers/release-notes.gradle
+++ b/installers/release-notes.gradle
@@ -42,11 +42,12 @@ task releaseNotes {
 
     ByteArrayOutputStream gitContributorsOut = new ByteArrayOutputStream()
     exec {
-      commandLine "git", "log", "--no-merges", "--format=%cN", "${project.previousVersion}..${rootProject.gitRevision}"
+
+      commandLine "git", 'shortlog', "-s", "--no-merges", "${project.previousVersion}..${rootProject.gitRevision}"
       standardOutput = gitContributorsOut
     }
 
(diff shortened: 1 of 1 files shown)


installers/release-notes.gradle
```diff
@@ -42,11 +42,11 @@ task releaseNotes {
 
     ByteArrayOutputStream gitContributorsOut = new ByteArrayOutputStream()
     exec {
-      commandLine "git", "log", "--no-merges", "--format=%cN", "16.6.0..${rootProject.gitRevision}"
+      commandLine "bash", '-c', "git shortlog -s --no-merges 16.6.0..${rootProject.gitRevision} | awk '!(\$1=\"\")' | tr '\n' ',' | sed s'/.\$//' | sed s'/^.//'"
```
ketan: Instead of piping output, is it better to process `gitContributorsOut.toString()` using groovy instead? Lot more portable when running on OSes that don't have `awk` and `sed`.

```groovy
# remove the leading commit count
gitContributorsOut.toString().readLines().collect { line -> line.trim().split(/\s+/, 2).last() }
```

ketan: Another reason being that errors in piping are not quite captured and bubbled out properly. The following returns a `0` exit code 

`true | false | true`

whereas this returns a non-zero —

`true | true | false`

zabil: That makes sense, will modify the script to use groovy for manipulations.

PR #2486, opened 2016-07-26
Added a friendlier error message on failing to send test email.

varshavaradarajan:


varshavaradarajan:
To do: Change the error message on failure to send email in functional tests.

PR #2513, opened 2016-08-02
#1133 fix CRTimer mapping to TimerConfig

tomzo:
Minor fix to config repos. If user (plugin) does not return `only_on_changes` in timer specification null pointer would be thrown. This is a fix + tests to cover all `CRTimer` conversion cases.

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/CRTimer.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract/CRTimer.java
@@ -12,7 +12,7 @@ public class CRTimer extends CRBase {
     {
         this.spec = timerSpec;
     }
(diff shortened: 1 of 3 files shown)


arvindsv:
Makes sense to me.

zabil:
Small nitpick.
Can we expand CR? I am assuming is stands for ConfigRepo
Much easier to use the same term while navigating code.

tomzo:
@zabil are you suggesting to rename all [these contract](https://github.com/tomzo/gocd/tree/81a0fb8bf888a5ef241ecd32a04b66bd19ff0e88/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/configrepo/contract) classes then? I am not sure it will be more readable, it will result in class names like `ConfigRepoConfigurationProperty` or `ConfigRepoPluginConfiguration` and such suggests different meaning at first glance. `CR`stands more for *contract* than *config repos*

PR #2514, opened 2016-08-03
Encoded cruise config md5 for display.

varshavaradarajan:


server/webapp/WEB-INF/rails.new/app/helpers/application_helper.rb
```diff
@@ -393,7 +393,10 @@ def make_https url
   end
 
   def config_md5_field
-    "<input type=\"hidden\" name=\"cruise_config_md5\" value=\"#{cruise_config_md5}\"/>".html_safe
+    input_field = "<input type=\"hidden\" name=\"cruise_config_md5\" value=\"".html_safe
```
ketan: Please no concatenation this is messy and unsafe. Use [`hidden_field_tag`](http://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-hidden_field_tag) instead.

PR #2515, opened 2016-08-03
Disallow directory listing by the web server.

varshavaradarajan:


server/webapp/WEB-INF/web.xml
```diff
@@ -207,6 +207,15 @@
     <load-on-startup>1</load-on-startup>
   </servlet>
 
+  <servlet>
+    <servlet-name>default</servlet-name>
+    <servlet-class>org.eclipse.jetty.servlet.DefaultServlet</servlet-class>
```
ketan: This servlet is already defined in `webdefault.xml` with a bunch of defaults. This change replaces those defaults with those here.

This will kill everything from cache headers to some perf optimizations that are applied for serving static assets.

Please move this to webdefault.xml

PR #2516, opened 2016-08-03
Added custom header to create artifact end point.

varshavaradarajan:


arvindsv:
Hmm. Doesn't some agent code need to change to include this header? Or, is this not used from the agent side?

PR #2526, opened 2016-08-04
Backups header fix

varshavaradarajan:
Since, we will anyway break the backups API with the custom header, thought we could also change the url by moving it to admin namespace. 

Old URL: ```http://localhost:8153/go/api/backups```
New URL: ```http://localhost:8153/go/api/admin/backups```

Custom Header: ```'Confirm:true'```

Is this okay or not?

### DIFF
--- a/server/webapp/WEB-INF/rails.new/config/routes.rb
+++ b/server/webapp/WEB-INF/rails.new/config/routes.rb
@@ -218,7 +218,7 @@ Go::Application.routes.draw do
 
   scope :api, as: :apiv1, format: false do
     api_version(:module => 'ApiV1', header: {name: 'Accept', value: 'application/vnd.go.cd.v1+json'}) do
-      resources :backups, only: [:create]
+      resources :backups, only: [:create], constraints: HeaderConstraint.new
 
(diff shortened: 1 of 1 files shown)


ketan:
> Custom Header: 'Confirm:true'

The `confirm` header was a stop gap for the old APIs. WDYT about expecting users to submit `Content-Type` on all requests that change state (POST, PUT, PATCH, DELETE ...), even if there's no content being submitted?

ketan:
Assigning to myself :)

arvindsv:
@rajiesh @varshavaradarajan @ketan: There was a flag to make the new header not be required, right? We should ensure that it gets used for this case. Of course, if the API is going to be fully changed, then it might not make sense.

varshavaradarajan:
Yes, ```go.api.with.safe.mode``` can be used. But if we are going to change the url, what's the point? Its going to be broken anyway. Unless you are suggesting that we don't change the url.

arvindsv:
No. If we're changing it, then we have to. What we could do is not change it this release. Just add the header. Then, put out a deprecation notice and change it in a month or so. I'm sure people have that backup API as a part of cron.

zabil:
> Just add the header. Then, put out a deprecation notice and change it in a month or so

@varshavaradarajan that's the thumb rule for all API's

Issue #2532, opened 2016-08-04
Add API route tests


maheshp:
Currently the API routes lacks tests, this is especially needed since we use constraints in some of the routes. To avoid issues like https://github.com/gocd/gocd/issues/2529, add route tests for API which would check if a given URL maps to appropriate controller:action

PR #2534, opened 2016-08-05
 Testing routes and constraints without making actual requests.

varshavaradarajan:
https://github.com/gocd/gocd/issues/2532

### DIFF
--- a/server/webapp/WEB-INF/rails.new/config/routes.rb
+++ b/server/webapp/WEB-INF/rails.new/config/routes.rb
@@ -317,7 +317,7 @@ Go::Application.routes.draw do
 
       # stage api's
       post 'stages/:id/cancel' => 'stages#cancel', constraints: HeaderConstraint.new, as: :cancel_stage
-      constraints pipeline_name: PIPELINE_NAME_FORMAT do
+      constraints pipeline_name: PIPELINE_NAME_FORMAT, stage_name: STAGE_NAME_FORMAT do
         post 'stages/:pipeline_name/:stage_name/cancel' => 'stages#cancel_stage_using_pipeline_stage_name', constraints: HeaderConstraint.new, as: :cancel_stage_using_pipeline_stage_name
       end
 
--- a/server/webapp/WEB-INF/rails.new/config/routes.rb
+++ b/server/webapp/WEB-INF/rails.new/config/routes.rb
@@ -26,7 +26,7 @@ Go::Application.routes.draw do
     PIPELINE_LOCATOR_CONSTRAINTS = {:pipeline_name => PIPELINE_NAME_FORMAT, :pipeline_counter => PIPELINE_COUNTER_FORMAT}
     STAGE_LOCATOR_CONSTRAINTS = {:stage_name => STAGE_NAME_FORMAT, :stage_counter => STAGE_COUNTER_FORMAT}.merge(PIPELINE_LOCATOR_CONSTRAINTS)
     ENVIRONMENT_NAME_CONSTRAINT = {:name => ENVIRONMENT_NAME_FORMAT}
-    PLUGIN_ID_FORMAT = /[a-zA-Z0-9\-_.]+/
+    PLUGIN_ID_FORMAT = /[\w\-.]+/
     ALLOW_DOTS = /[^\/]+/
     CONSTANTS = true
   end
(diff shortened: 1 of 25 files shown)


ketan:
The approach looks good to me, can we move these into specific controllers instead?

varshavaradarajan:
Well, I know the specs aren't exactly routing but is there any reason not to have this structure?

```
└── routing
    ├── api
    ├── api_v1
    └── api_v2
```

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -101,6 +121,42 @@
...
+
+    describe :route do
+      describe :with_header do
+        before :each do
+          Rack::MockRequest::DEFAULT_ENV["HTTP_ACCEPT"] = "application/vnd.go.cd.v1+json"
+        end
+        after :each do
+          Rack::MockRequest::DEFAULT_ENV = {}
+        end
+        it 'should route to show action of environments controller for alphanumeric environment name' do
```
ketan: You can reduce a lot of dup if you change most of this code to —

```ruby
%w(foo123 foo.123 foo-123).each do |name|
  it "should route ... to ... with param name #{name}" do
    expect(get: ".../#{name}").to route_to(action: ..., controller: ..., name: ...)
  end
end

```

maheshp: The only drawback I see with this approach is that the testing scenario  is not explicit,

eg: ```it('should route to ... for pipeline name with dot')``` somebody will have to see the test data to understand the scenarios being tested.

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -26,7 +26,7 @@
     PIPELINE_LOCATOR_CONSTRAINTS = {:pipeline_name => PIPELINE_NAME_FORMAT, :pipeline_counter => PIPELINE_COUNTER_FORMAT}
     STAGE_LOCATOR_CONSTRAINTS = {:stage_name => STAGE_NAME_FORMAT, :stage_counter => STAGE_COUNTER_FORMAT}.merge(PIPELINE_LOCATOR_CONSTRAINTS)
     ENVIRONMENT_NAME_CONSTRAINT = {:name => ENVIRONMENT_NAME_FORMAT}
-    PLUGIN_ID_FORMAT = /[a-zA-Z0-9\-_.]+/
+    PLUGIN_ID_FORMAT = /[\w\-.]+/
```
ketan: See https://github.com/gocd/gocd/blob/242d113bb7724911ee0c7166b200d56a86f3ada3/plugin-infra/go-plugin-api/resources/plugin-descriptor.xsd#L33-L33

server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/environments_controller_spec.rb
```diff
@@ -54,6 +54,26 @@
...
     end
+
+    describe :route do
+      describe :with_header do
+        before :each do
+          Rack::MockRequest::DEFAULT_ENV["HTTP_ACCEPT"] = "application/vnd.go.cd.v1+json"
+        end
+        after :each do
+          Rack::MockRequest::DEFAULT_ENV = {}
+        end
```
ketan: `DEFAULT_ENV` is supposed to contain other keys as well — https://github.com/gocd/gocd/blob/f53cc516f5ef04295fd03dfa9dd8160b0f0745cd/server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/rack-1.5.5/lib/rack/mock.rb#L43-L50

Might as well include a helper module —

```ruby
module ApiHeaderSetupTeardown
  def self.included(klass)
    klass.class_eval do
      before :each do
        @default_env = Rack::MockRequest::DEFAULT_ENV.dup
        Rack::MockRequest::DEFAULT_ENV.merge!({'HTTP_ACCEPT' => "application/vnd.go.cd.v#{supported_version}+json"})
      end
      after :each do
        Rack::MockRequest::DEFAULT_ENV = @default_env
      end
    end
  end
end

describe FooController do
  include ApiHeaderSetupTeardown
  def supported_version
    1
  end

  it 'should foo' do
    ...
  end
end
```

PR #2537, opened 2016-08-09
Use 'command' instead of 'which' to auto detect java

zabil:
There's a lot of info about why here
http://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then

Also, base images on centos or redhat don't have ```which``` command.


ketan:
Can you make the same change in server.sh too?

PR #2548, opened 2016-08-11
Add custom error page

iamsumanth:


jetty9/resources/error.html
```diff
@@ -0,0 +1,61 @@
+<!--
```
ketan: You're not supposed to add copyright in HTML pages. This is because the `<meta charset` needs to be in the first 512 characters in a web page, for most browsers to determine the charset properly.

PR #2550, opened 2016-08-12
Changed the package material check-connection request from GET to POST

GaneshSPatil:


server/webapp/WEB-INF/rails.new/app/assets/javascripts/package_material_check_connection.js
```diff
@@ -15,18 +15,22 @@
  *************************GO-LICENSE-END**********************************/
 
 PackageMaterialCheckConnection = function (url) {
-
     _bind = function (parent_container_selector, check_connection_selector, connection_message_selector) {
         jQuery(parent_container_selector).on('click', check_connection_selector, function (e) {
             var connection_message_element = jQuery(connection_message_selector);
-            connection_message_element.removeClass("error_message").removeClass("ok_message").text('Checking connection...');
+            var formData = jQuery(parent_container_selector + " :input").serialize();
+            formData = formData.replace("_method=PUT", "_method=POST");
```
zabil: Do we need to do this?

server/webapp/WEB-INF/rails.new/app/assets/javascripts/package_material_check_connection.js
```diff
@@ -15,18 +15,22 @@
...
+
             jQuery.ajax({
-                type: "GET",
+                type: "POST",
```
zabil: Also add header constraint of ```Confirm: true``` like the other POST methods

server/webapp/WEB-INF/rails.new/app/assets/javascripts/package_material_check_connection.js
GaneshSPatil: yes.. otherwise the request will be "PUT" not post..
the same case in pluggableSCM
https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/pluggable_scm_check_connection.js#L24

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -145,7 +145,7 @@
   get "admin/package_definitions/check_connection" => "admin/package_definitions#check_connection", as: :package_definition_check_connection
 
   get "admin/package_repositories/new" => "admin/package_repositories#new", as: :package_repositories_new
-  get "admin/package_repositories/check_connection" => "admin/package_repositories#check_connection", as: :package_repositories_check_connection
+  post "admin/package_repositories/check_connection" => "admin/package_repositories#check_connection", as: :package_repositories_check_connection
```
zabil: You need to add the header constraint here also see
https://github.com/gocd/gocd/blob/1190b342edf23b8ce8e5dabce41a9ecc29ec9638/server/webapp/WEB-INF/rails.new/config/routes.rb#L294

server/webapp/WEB-INF/rails.new/app/assets/javascripts/package_material_check_connection.js
ketan: That's not true. the `_method` hack is in rails because of the fact that old browsers did not support some of the verbs like PUT, PATCH, OPTIONS, so everything was submitted as a `POST` with an additional `_method` parameter that rails could use to distinguish between different HTTP verbs.

You should not have to do any `formData.replace` in the javascript, because the relevant form fields should have the correct method on them, and submitting using `POST` should work just fine.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/package_material_check_connection.js
zabil: You can write a spec to test it 
https://github.com/gocd/gocd/blob/4a8fbd842355b4a1b82f85211c95001e7b215c46/server/webapp/WEB-INF/rails.new/spec/controllers/api/server_state_controller_spec.rb#L35

PR #2554, opened 2016-08-12
Add error page

iamsumanth:


jetty9/test/com/thoughtworks/go/server/JettyCustomErrorPageHandlerTest.java
```diff
@@ -44,8 +44,8 @@ public void setUp() throws Exception {
     }
 
     @Test
-    public void shouldWriteErrorPage() throws Exception {
-        errorHandler.writeErrorPage(request, writer, 404, "Not Found", false);
+    public void shouldWriteErrorPageFor404WithMessage() throws Exception {
+        errorHandler.writeErrorPage(request, writer, 404, null, false);
```
ketan: Could we add a test that the message passed in is not rendered —

```
errorHandler.writeErrorPage(request, writer, 404, "this should not be rendered", false);


assertThat(page, not(containsString("this should not be rendered")));
```

Issue #2464, opened 2016-07-20
Pending agent shows a misleading stack trace


arvindsv:
Mentioned here: https://groups.google.com/d/topic/go-cd/zLqeECge1t4/discussion

* Server gets an agent [registration request](https://github.com/gocd/gocd/blob/d87f6242bb52541cf20b01226c0a5fc994474a2e/server/src/com/thoughtworks/go/server/service/AgentService.java#L325).
* Since the agent should be pending, it returns a [null private key](https://github.com/gocd/gocd/blob/d87f6242bb52541cf20b01226c0a5fc994474a2e/common/src/com/thoughtworks/go/domain/AgentInstance.java#L220) to [here](https://github.com/gocd/gocd/blob/d87f6242bb52541cf20b01226c0a5fc994474a2e/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L233).
* This gets [serialized](https://github.com/gocd/gocd/blob/d87f6242bb52541cf20b01226c0a5fc994474a2e/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L249) and goes over with no key.
* Agent tries to deserialize it and fails.

Log with more context than the one in the mailing list, courtesy @dudadornelles:

```
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - >> "GET /go/admin/latest-agent.status HTTP/1.1[\r][\n]"
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - >> "User-Agent: Jakarta Commons-HttpClient/3.0.1[\r][\n]"
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - >> "Host: delivery.sonicdrivein.digital:8154[\r][\n]"
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - >> "Cookie: $Version=0; JSESSIONID=1lhi0tydld2weamt53xuxwzle; $Path=/go[\r][\n]"
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - >> "[\r][\n]"
2016-07-20 22:05:37,112 [loopThread] DEBUG httpclient.wire.header:69 - << "HTTP/1.1 200 OK[\r][\n]"
[...]

PR #2558, opened 2016-08-16
Tweak the agent registration protocol a bit (#2464)

ketan:
* The server responds with a 202 (Accepted) and an empty JSON object `{}`
  as long as the registration is pending.
* The server responds with a 200 (OK) and the old JSON object containing
  the private key and the cert chain.

### DIFF
--- a/common/test/unit/com/thoughtworks/go/security/RegistrationTest.java
+++ b/common/test/unit/com/thoughtworks/go/security/RegistrationTest.java
@@ -16,20 +16,23 @@
 
 package com.thoughtworks.go.security;
 
-import com.thoughtworks.go.util.TestFileUtil;
 import org.apache.commons.lang.builder.EqualsBuilder;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 
 import java.io.File;
+import java.io.IOException;
 
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.*;
 
 public class RegistrationTest {
-    private static String authorityKeystorePath = "tempAuthorityKeystore";
+    @Rule
+    public TemporaryFolder tmpFolder = new TemporaryFolder();
 
     @Test
-    public void decodeFromJson() {
+    public void decodeFromJson() throws IOException {
         Registration origin = createRegistration();
         Registration reg = RegistrationJSONizer.fromJson(RegistrationJSONizer.toJson(origin));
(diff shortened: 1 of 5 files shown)


common/test/unit/com/thoughtworks/go/security/RegistrationTest.java
```diff
@@ -39,7 +40,25 @@ public void decodeFromJson() {
...
+    @Test
+    public void shouldEncodeDecodeEmptyRegistration() throws Exception {
+        Registration toSerialize = Registration.createNullPrivateKeyEntry();
+        Registration deserialized = Registration.fromJson(toSerialize.toJson());
+
+        assertTrue(EqualsBuilder.reflectionEquals(toSerialize, deserialized));
+    }
+
+    private static Registration createRegistration() {
         File tempKeystoreFile = TestFileUtil.createUniqueTempFile(authorityKeystorePath);
```
zabil: Can we use junit temporary folder rule to manage temporary file creation?

common/test/unit/com/thoughtworks/go/security/RegistrationTest.java
```diff
@@ -39,7 +40,25 @@ public void decodeFromJson() {
         assertThat(reg.getChain().length, is(3));
     }
 
-    public static Registration createRegistration() {
+
+    @Test
+    public void shouldBeValidWhenPrivateKeyAndChainIsPresent() throws Exception {
+        assertFalse(Registration.createNullPrivateKeyEntry().isValid());
```
tomzo: I would assert something like this too:
```
 assertFalse(Registration.fromJson("{}").isValid());
```

arvindsv:
Does it fix [this](https://github.com/gocd/gocd/issues/2464#issuecomment-244779458) as well?

ketan:
> Does it fix #2464 as well?

Yes, see title.

arvindsv:
@ketan: I had linked to @tomzo's recent commit on #2464 - not just to the issue. I guess that is fixed by this as well.

ketan:
@arvindsv — the problem looks identical, not sure why the `guid.txt` needed to be removed, something that I'd need to investigate, at first thought it looks like the culprit is [this line](https://github.com/gocd-contrib/go-cookbook/blob/master/providers/agent.rb#L66) — that file is no longer used by new agents.

It'll remain when upgrading the agents, just in case someone decides to downgrade.

@tomzo — would you be able to investigate further why the `guid.txt` needs to be removed?

PR #2563, opened 2016-08-16
Fixed update environment failing tests

GaneshSPatil:


server/webapp/WEB-INF/rails.new/spec/controllers/environments_controller_spec.rb
```diff
@@ -268,46 +281,23 @@ def md5
...
-    end
-
-    it "should return error when invalid agent uuid is added" do
-      put :update, :no_layout => true, :environment => {'agents' => [{'uuid' => "invalid-one"}]}, :name => @environment_name, :cruise_config_md5 => md5
-
-      expect(response.status).to eq(400)
-      expect(response.body).to eq("Failed to update environment 'foo-environment'. Environment 'foo-environment' has an invalid agent uuid 'invalid-one'\n")
-    end
-
-    it "should return error when updating a non-exitant environment" do
```
maheshp: We need this scenario to be tested

PR #2564, opened 2016-08-16
Re-use SSL connections by specifying a user principal as part of all remoting connections.

ketan:


ketan:
### Some perf numbers

**Instance spec** c3.2xlarge - 8 core/15.0GB — Using -Xmx=10G
**Installed JREs** Java 7u111 & Java 8u101
**Number of agents** 400
**OS** CentOS 6 (ami-1c221e76) with latest patches of everything. `yum clean all && yum install -y epel-release && yum update -y && yum install -y java-1.7.0-openjdk java-1.8.0-openjdk htop curl unzip`

Config xml had no pipelines or materials, since we were only interested in Agent SSL numbers, with the server processing nothing else.

There was no observable performance difference between JRE 7 and JRE 8.

However with the patch applied, the baseline CPU usage which was around 6% went down by 50% to about 3% CPU. 

To put that into context — 6% CPU usage on an 8 core machine is approximately 1 CPU permanently pegged at 50%.

Screenshots from cloudwatch attached

screen shot 2016-08-17 at 3 26 46 pm

PR #2569, opened 2016-08-18
Added validation for encrypted password attribute of svn material.

varshavaradarajan:


ketan:
Additionally, where we're doing a `SvnMaterialConfig#currentPassword`, we're only handling a `InvalidCipherTextException`, ideally it should catch `Exception` instead.

PR #2584, opened 2016-08-23
Removed the url from the not found page

GaneshSPatil:


server/webapp/localize.xml
```diff
@@ -119,7 +119,7 @@
     <entry key="FAILED_JOBS">Failed: {0}</entry>
     <entry key="IN_PROGRESS_JOBS">In Progress: {0}</entry>
     <entry key="UNABLE_TO_RETRIEVE_FAILURE_RESULTS">Unable to retrieve failure results.</entry>
-    <entry key="URL_NOT_KNOWN">The url [ {0} ] you are trying to reach doesn''t appear to be correct.</entry>
+    <entry key="URL_NOT_KNOWN">The url you are trying to reach doesn't appear to be correct.</entry>
```
zabil: Change to

The url you are trying to reach appears to be incorrect

Issue #2233, opened 2016-05-29
Create / Modify / List Templates over Rest API


unixunion:
##### Issue Type

- Feature enhancement

##### Summary

As a Continuous Delivery and Integration automation system developer, I would like to be able to create / modify and list templates that are available within the Go Server through the REST Api, similarly to the how pipelines are created. 

I want to create templates and define their stages
list the templates I have created ( plus a description of the template )
I want to modify templates 

/K

varshavaradarajan:
INDEX: 

URL: ```https://ci.example.com/go/api/admin/templates```
Header ```Accept: application/vnd.go.cd.v1+json```

```
{
  "templates": [
    {
      "name": "template",
      "pipelines": [
       {
         "name": "up42"
       }
      ]
    }
  ]
}
```

GET
URL: ```https://ci.example.com/go/api/admin/templates/foo```
Header ```Accept: application/vnd.go.cd.v1+json```

```
{
  "name": "foo",
  "stages": [
    {
      "name": "up42_stage",
      "fetch_materials": true,
      "clean_working_directory": false,
      "never_cleanup_artifacts": false,
      "approval": {
        "type": "success",
        "authorization": {
          "roles": [

          ],
          "users": [

          ]
        }
      },
      "environment_variables": [

      ],
      "jobs": [
        {
          "name": "up42_job",
          "run_instance_count": null,
          "timeout": null,
          "environment_variables": [

          ],
          "resources": [

          ],
          "tasks": [
            {
              "type": "exec",
              "attributes": {
                "run_if": [

                ],
                "on_cancel": null,
                "command": "ls",
                "working_directory": null
              }
            }
          ],
          "tabs": [

          ],
          "artifacts": [

          ],
          "properties": null
        }
      ]
    }
  ]
}
```

CREATE

URL: ```https://ci.example.com/go/api/admin/templates```
Header ```Accept: application/vnd.go.cd.v1+json```
Header ```Content-Type: application/json```

```
"name": "template",
 "stages": [
    {
      "name": "my_stage",
      "fetch_materials": true,
      "clean_working_directory": false,
      "never_cleanup_artifacts": false,
      "approval": {
        "type": "success",
        "authorization": {
          "roles": [],
          "users": []
        }
      },
      "environment_variables": [],
      "jobs": [
        {
          "name": "my_job",
          "run_instance_count": null,
          "timeout": 0,
          "environment_variables": [],
          "resources": [
                "Linux",
                "Java"
          ],
[...]

PR #2591, opened 2016-08-25
Templates API - index, show, create, update. (#2233)

varshavaradarajan:


server/src/com/thoughtworks/go/config/update/DeleteTemplateConfigCommand.java
```diff
@@ -0,0 +1,81 @@
...
+            result.unprocessableEntity(LocalizedMessage.string("CANNOT_DELETE_TEMPLATE"));
+            throw new GoConfigInvalidException(preprocessedConfig, String.format("The template '%s' is being reference by pipelines: %s", templateConfig.name(), pipelinesAssociatedWithTemplate));
+        }
+        return true;
+    }
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
+        return isUserAuthorized() && doesTemplateExist(cruiseConfig);
+    }
```
GaneshSPatil: do we need to check `doesTemplateExist(cruiseConfig)` as at the controller we are doing a `load_template` before `destroy`
Same check doing twice

server/src/com/thoughtworks/go/config/update/TemplateConfigCommand.java
```diff
@@ -0,0 +1,78 @@
...
+
+    protected PipelineTemplateConfig findAddedTemplate(CruiseConfig cruiseConfig) {
+        if (templateConfig == null || templateConfig.name() == null) {
+            result.unprocessableEntity(LocalizedMessage.string("TEMPLATE_INVALID"));
+            throw new NullPointerException("Template name cannot be null.");
+        } else {
+            PipelineTemplateConfig pipelineTemplateConfig = cruiseConfig.findTemplate(templateConfig.name());
+            if (pipelineTemplateConfig == null) {
+                result.notFound(LocalizedMessage.string("RESOURCE_NOT_FOUND"), HealthStateType.notFound());
+                throw new NullPointerException(String.format("The template with name '%s' is not found.", templateConfig.name()));
```
GaneshSPatil: Can we throw a specific exception `throw new NoSuchTemplateException(templateName);`
as we do for `invalid agent` 
https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java#L84

server/src/com/thoughtworks/go/config/update/DeleteTemplateConfigCommand.java
varshavaradarajan: Yes, this check is in the synchronized block while the ```load_template``` is not. The template may not exist while the config save request from the api is being processed.  

Same reason why we check the md5 twice on an update call, remember?

GaneshSPatil: Agree, can you add the same check for UpdateTemplateConfigCommand
https://github.com/gocd/gocd/pull/2591/files#diff-340d04c19bfc07d0a9360303aa562cedR54

GaneshSPatil:
Create Template returns `403 Forbidden` for `non-admin` users, it should return 401

arvindsv:
Linking to: #2233.

config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
```diff
@@ -96,6 +97,18 @@ public StageConfig getStage(final CaseInsensitiveString stageName) {
         return findBy(stageName);
     }
 
+    public List<StageConfig> getStages() {
+        return this;
+    }
+
+    public void setName(String name) {
+        this.name = new CaseInsensitiveString(name);
```
ketan: Preferably have one call the other?

server/src/com/thoughtworks/go/config/update/CreateTemplateConfigCommand.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void update(CruiseConfig modifiedConfig) throws Exception {
+        modifiedConfig.addTemplate(templateConfig);
+    }
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
+        if (!(goConfigService.isUserAdmin(currentUser))) {
```
ketan: Maybe too many parens? Kill the innermost?

server/webapp/localize_ja.xml
```diff
@@ -528,6 +528,8 @@
     <entry key="MORE_INFORMATION">More Information</entry>
     <entry key="NO_ENVIRONMENTS_CONFIGURED">No environments are displayed because either no environments have been set up or you are not authorized to view the pipelines within any of the environments.</entry>
     <entry key="CANNOT_DELETE_TEMPLATE">Cannot delete this template as it is used by at least one pipeline</entry>
+    <entry key="TEMPLATE_DELETED_SUCCESSFUL">The template ''{0}'' was deleted successfully.</entry>
+    <entry key="TEMPLATE_INVALID">The template config is invalid. Template name cannot be null.</entry>
```
ketan: Seems misleading, call it `TEMPLATE_NAME_CANNOT_BE_NULL`?

server/src/com/thoughtworks/go/config/update/TemplateConfigCommand.java
```diff
@@ -0,0 +1,79 @@
...
+            templatesConfig.validate(null);
+            BasicCruiseConfig.copyErrors(preprocessedTemplateConfig, templateConfig);
+            return preprocessedTemplateConfig.getAllErrors().isEmpty() && templatesConfig.errors().isEmpty();
+        }
+        BasicCruiseConfig.copyErrors(preprocessedTemplateConfig, templateConfig);
+        return false;
+    }
+
+    protected PipelineTemplateConfig findAddedTemplate(CruiseConfig cruiseConfig) {
+        if (templateConfig == null || templateConfig.name() == null) {
```
ketan: What about blank name?

server/src/com/thoughtworks/go/config/update/CreateTemplateConfigCommand.java
```diff
@@ -0,0 +1,51 @@
...
+        this.goConfigService = goConfigService;
+    }
+
+    @Override
+    public void update(CruiseConfig modifiedConfig) throws Exception {
+        modifiedConfig.addTemplate(templateConfig);
+    }
+
+    @Override
+    public boolean canContinue(CruiseConfig cruiseConfig) {
```
ketan: If you move this to the super class, this implementation can be removed. Other implementations can become —


```
class DeleteTemplateConfigCommand extends ... {
  public boolean canContinue() {
    super.canContinue() && doesTemplateExist();
  }
}
```

server/src/com/thoughtworks/go/config/update/TemplateConfigCommand.java
```diff
@@ -0,0 +1,79 @@
...
+            result.unprocessableEntity(LocalizedMessage.string("TEMPLATE_INVALID"));
+            throw new NullPointerException("Template name cannot be null.");
```
ketan: Why NPE, and not `IllegalArgumentException`?

server/src/com/thoughtworks/go/server/service/EntityHashingService.java
```diff
@@ -109,6 +110,14 @@ public void onEntityConfigChange(SCM scm) {
         }
     }
 
+    class TemplateConfigChangedListner extends EntityConfigChangedListener<PipelineTemplateConfig> {
+        @Override
+        public void onEntityConfigChange(PipelineTemplateConfig pipelineTemplateConfig) {
+            String cacheKey =  pipelineTemplateConfig.getClass().getName() + pipelineTemplateConfig.name().toLower();
+            goCache.remove(ETAG_CACHE_KEY, cacheKey.toLowerCase());
```
ketan: This could be problematic — if there is an environment, pipeline, scm and template which share the same name, you'd see cache key collision.

Probably change the cache key to `ETAG_CACHE_KEY + configObject.getClass() + configObject.name().toLower()`

Could you fix them in a separate PR, and fix this instance here?

server/src/com/thoughtworks/go/server/service/TemplateConfigService.java
```diff
@@ -61,6 +70,36 @@ public void removeTemplate(String templateName, CruiseConfig cruiseConfig, Strin
...
+
+    private void update(Username currentUser, LocalizedOperationResult result, EntityConfigUpdateCommand command) {
+        try {
+            goConfigService.updateConfig(command, currentUser);
+        } catch (Exception e) {
+            if (e instanceof ConfigUpdateCheckFailedException) {
+                return;
+            }
+            LOGGER.error(e.getMessage(), e);
+            result.unprocessableEntity(LocalizedMessage.string("SAVE_FAILED_WITH_REASON", e.getMessage()));
```
ketan: That feels a bit dangerous to let out an internal error to end user.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/templates_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+        if result.isSuccessful
+          response.etag = [get_etag_for_template(updated_template)]
+          render DEFAULT_FORMAT => json
+        else
+          render_http_operation_result(result, {data: json})
+        end
+      end
+
+      def get_etag_for_template(template)
+        entity_hashing_service.md5ForEntity(template, template.name.to_s)
```
ketan: To lower on this? Better yet, have it accept `CaseInsensitiveString` as a second arg to prevent this issue?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/template_config_representer.rb
```diff
@@ -0,0 +1,56 @@
...
+
+module ApiV1
+  module Config
+    class TemplateConfigRepresenter < ApiV1::BaseRepresenter
+      alias_method :template, :represented
+
+      error_representer
+
+      link :self do |opts|
+        opts[:url_builder].apiv1_admin_template_url(template_name: template.name.to_s) unless template.name.nil? || template.name.to_s.blank?
```
ketan: Can be simply written as `unless template.name.blank?` Since `CaseInsensitiveString#isBlank` exists and `nil` responds to `.blank?` :)

See http://api.rubyonrails.org/v4.0.4/classes/NilClass.html#method-i-blank-3F

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/template_config_representer.rb
```diff
@@ -0,0 +1,56 @@
...
+      property :name, case_insensitive_string: true
+
+      collection :stages,
+                 exec_context: :decorator,
+                 decorator:    ApiV1::Config::StageRepresenter,
+                 expect_hash:  true,
+                 class:        com.thoughtworks.go.config.StageConfig
+
+      def stages
+        template.getStages() if !template.getStages().isEmpty
```
ketan: use unless?

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/template_config_representer.rb
varshavaradarajan: But if ```template.name``` is nil, ```nil.blank?``` will bomb, right?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/templates_controller.rb
varshavaradarajan: Handled in ```EntityHashingService```.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/templates_config_representer.rb
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Config
+    class TemplatesConfigRepresenter < ApiV1::BaseRepresenter
```
ketan: No links like in [here](https://github.com/gocd/gocd/blob/f18d7930b9213a5a46b37ffc1c24b9365fd15ac9/server/webapp/WEB-INF/rails.new/app/presenters/api_v4/agents_representer.rb#L26-L32)?

Issue #2586, opened 2016-08-23
Make it clearer that the Task plugin extension does not support plugin settings


alethenorio:
##### Issue Type
- Documentation Report
##### Summary

The documentation structure seems to imply that a task extension plugin supports retrieving plugin settings using the `GoApplicationAccessor` however retrieving plugin settings through the task extension is not actually supported and the documentation should reflect that.
##### Environment

GOCD Server and agent running in separate docker containers
###### Basic environment details
- Go Version: `16.8.0`
- JAVA Version: `1.7`
- OS: `Debian`
##### Details

If one looks at the documentation for writing a GO Plugin

https://developer.go.cd/current/writing_go_plugins/go_application_accessor/go_application_accessor.html

Notice how it falls directly under `Writing Go Plugins` section implying it is supported by all types of plugins. This is further supported by the fact that the task will request and response to the followinf JSON API messages

`go.plugin-settings.get-configuration`
`go.plugin-settings.get-view`
`go.plugin-settings.validate-configuration`

And it will be possible to configure and save such configuration in the UI.

When you try to send a message for retreiving plugin configuration when running a task extension plugin however you get the following error

<pre>
[...]

PR #2593, opened 2016-08-25
Dropped GET settings request for Task Plugin

GaneshSPatil:
Dropped following requests for task-plugins: 
- GET-plugin-settings.get-configuration 
- GET-plugin-settings.get-view

GoCD does not supports retrieving plugin settings for task plugins.


varshavaradarajan:
Relates to https://github.com/gocd/gocd/issues/2586

PR #2601, opened 2016-08-27
Revert "Detect linux distributions properly"

zabil:
Reverts gocd/gocd#2421


zabil:
@aj-jaswanth 
Reverting this as it's breaking builds on windows.
Refer https://build.go.cd/go/pipelines/build-windows/1097/build-non-server/1

Also detecting using python needs to be handled when python is installed in non linux distros 
refer https://github.com/gocd/gocd/commit/dea3654e4c8f68ce1832846b68d0552de84d12e0

It's better to modify this pull request to only have pretty name detection while keeping everything else as it is.

aj-jaswanth:
@zabil Fixing the tests.

PR #2608, opened 2016-08-27
Detect linux distributions properly

aj-jaswanth:
This reverts commit 75828d45cc4566212dcb00d2b6fbbf261050b48a.

@zabil Fixed the issue with Windows systems. Please run the PR build for windows.


ketan:
@aj-jaswanth — if I understand, the only thing you changed is `FileUtils#isAbsolutePath`. Is that correct?

Reading from the documentation of `org.apache.tools.ant.util.File#isAbsolutePath` (from where this code originated), I see this javadoc —

```java
    /**
     * Verifies that the specified filename represents an absolute path.
     * Differs from new java.io.File("filename").isAbsolute() in that a path
     * beginning with a double file separator--signifying a Windows UNC--must
     * at minimum match "\\a\b" to be considered an absolute path.
     * @param filename the filename to be checked.
     * @return true if the filename represents an absolute path.
     * @throws java.lang.NullPointerException if filename is null.
     * @since Ant 1.6.3
     */
    public static boolean isAbsolutePath(String filename) {
```

It feels like this change might just break a lot more things than we expect.

Could you explain why this change was needed?

```diff
$ diff <(curl  -Ls https://github.com/gocd/gocd/pull/2421.diff) <(curl -Ls https://github.com/gocd/gocd/pull/2608.diff)
2c2
< index 430246c..85e9836 100644
---
> index 430246c..d9c3282 100644
21a22,53
> @@ -237,29 +231,8 @@ public static boolean isSubdirectoryOf(File parent, File subdirectory) throws IO
>      //CopiedFromAnt
>
>      public static boolean isAbsolutePath(String filename) {
> -        int len = filename.length();
> -        if (len == 0) {
> -            return false;
> -        }
> -        char sep = File.separatorChar;
> -        filename = filename.replace('/', sep).replace('\\', sep);
> -        char c = filename.charAt(0);
> -        if (!(ON_DOS || ON_NETWARE)) {
> -            return (c == sep);
> -        }
> -        if (c == sep) {
> -            // CheckStyle:MagicNumber OFF
> -            if (!(ON_DOS && len > 4 && filename.charAt(1) == sep)) {
> -                return false;
> -            }
> -            // CheckStyle:MagicNumber ON
> -            int nextsep = filename.indexOf(sep, 2);
> -            return nextsep > 2 && nextsep + 1 < len;
> -        }
> -        int colon = filename.indexOf(':');
> -        return (Character.isLetter(c) && colon == 1
> -                && filename.length() > 2 && filename.charAt(2) == sep)
> -                || (ON_NETWARE && colon > 0);
> +        File file = new File(filename);
> +        return file.isAbsolute();
>      }
>
>      public static String[] dissect(String path) {
```

aj-jaswanth:
We don't have dos and other operating systems from the previous changes. Missed that documentation detail. Thanks for pointing out.

It doesn't make sense to have `\\host\`. share point is necessary like `\\host\dir`. Modified the code to have a check for it.

base/src/com/thoughtworks/go/util/FileUtil.java
```diff
@@ -237,29 +242,14 @@ public static boolean isSubdirectoryOf(File parent, File subdirectory) throws IO
...
-        }
-        if (c == sep) {
-            // CheckStyle:MagicNumber OFF
-            if (!(ON_DOS && len > 4 && filename.charAt(1) == sep)) {
-                return false;
+        File file = new File(filename);
+        boolean absolute = file.isAbsolute();
+        if (absolute && OperatingSystem.isFamily(OperatingSystem.WINDOWS)) {
+            if (filename.startsWith("\\\\") && !filename.matches("\\\\\\\\.*\\\\.+")) {
+                absolute = false;
```
varshavaradarajan: Can you please add some tests for this?

aj-jaswanth: Sure. Will add them.

varshavaradarajan:
@aj-jaswanth, can you please separate the cleanup into another PR?

 Will check the installers once the PR is updated.

common/test/unit/com/thoughtworks/go/util/FileUtilTest.java
```diff
@@ -271,6 +271,20 @@ public void shouldReturnFalseIfGivenFolderIsAbsolute() {
...
+        assertThat(FileUtil.isAbsolutePath("\\\\host"), is(false));
+    }
+
+    @Test
+    @RunIf(value = EnhancedOSChecker.class, arguments = {EnhancedOSChecker.WINDOWS})
+    public void shouldReturnTrueForValidWindowsUNCFilePath() {
+        assertThat(FileUtil.isAbsolutePath("\\\\host\\share"), is(true));
+        assertThat(FileUtil.isAbsolutePath("\\\\host\\share\\dir"), is(true));
+    }
+
```
aj-jaswanth: @varshavaradarajan Added tests for Windows UNC pathnames.

aj-jaswanth:
@varshavaradarajan @ketan Removed clean up code from the current PR.

common/test/unit/com/thoughtworks/go/util/FileUtilTest.java
varshavaradarajan: @aj-jaswanth, will trigger the PR build.

aj-jaswanth: @varshavaradarajan Is this PR looking good without any build breaks?

varshavaradarajan:
@aj-jaswanth, we'll merge it soon. Thanks for the PR.

base/src/com/thoughtworks/go/util/OperatingSystem.java
```diff
@@ -12,44 +12,65 @@
...
+            }
+        }
+        return OS_COMPLETE_NAME;
     }
 
-    public static OperatingSystem fromProperty() {
-        String osName = new SystemEnvironment().getPropertyImpl("os.name");
-        return parseOperatingSystem(osName);
+    private static String readFromOsRelease() throws Exception {
+        FileReader fileReader;
```
zabil: @aj-jaswanth 
Please this commit here
https://github.com/gocd/gocd/commit/4f7db5908a5163ef4895003dc62be8a63a7e0c6b

I had reverted it

We cannot use StringUtils or anything from apache commons as they are not loaded into the class path when     the agent or server is packaged.

aj-jaswanth:
@zabil I tested both server and client. They are booting up successfully and the agent is registering fine with the server.

Made one new change here https://github.com/gocd/gocd/pull/2608/files#diff-24c880b793ce940fb08ac11072cd64aeR166 to send complete OS name when registering the agent.

Issue #2567, opened 2016-08-18
[Elastic agents] Autogenerate `agentAutoRegisterKey` on `<server/>`


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Currently, the go server, when it starts up the first time, will ensure each server has a unique [`serverId`](https://github.com/gocd/gocd/blob/008816d6dbe217fd3e61547d2463fce8efe1762d/config/config-api/src/com/thoughtworks/go/config/ServerConfig.java#L68-L73). With elastic agents being pushed out as beta, users are required to assign an `agentAutoRegisterKey` to the `<server/>` tag.

I'm left wondering if we should just auto-generate the key if it's missing, just like with `serverId`

PR #2611, opened 2016-08-29
Autogenerate the agent autoregister key to ease setting up of elastic agents (#2567)

ketan:


arvindsv:
Link: #2567 

GitHub seems to link issues only in description or comments. Not in title.

PR #2614, opened 2016-08-29
Delete killableprocess.exe

ketan:


ketan:
This is an unused binary

Issue #1946, opened 2016-02-17
[RFC] Deprecating solaris support


ketan:
In light of downward trend in solaris installers, we are now considering a timeline to stop providing solaris installers starting from 1st June 2016. This does not mean that we've taken a firm decision, it just means that we're considering taking one.
##### What does this mean for users installing GoCD on Solaris

Users can continue to use the generic zip files to run GoCD on Solaris. However the support for solaris will be on a best-effort basis.

If you're someone who's using GoCD with Solaris, please provide your comments on this issue to make sure your voice is heard.

fire:
I'm using gocd on SmartOS. However, it's not the installer.

sriramnrn:
Hi:

We have commercial support in place (and this has been extended recently for a few years more). We use GoCD on Solaris 10 and on Solaris 11. Since I also use GoCD on other opensolaris distress, I know that just the zip file is enough to get started.

However, the startup scripts, the packaging recipe, etc are useful for those of us who want to install via packages.

Would it be possible to move these Solaris scripts into a contrib folder? That way, you don't necessarily need to supply signed installers, but those who want to build the installers can still do so.

I will reply via the commercial support channel to discuss what an EoL duration would mean when one has commercial support.

-- Ram

arvindsv:
This has been done.

PR #2615, opened 2016-08-29
Remove solaris installer code

ketan:


sriramnrn:
Wow.. why is this being removed !? We are using this on my current project !!!

ketan:
We've not provided solaris installers for the past couple of releases now.

https://github.com/gocd/gocd/issues/1946
https://groups.google.com/forum/#!topic/go-cd/sw9hLBX8iDY

You can continue to use

ketan:
Fat fingered the send button –

You may continue to use the zip installers, but solaris support would be on a best effort basis, since the team can no longer justify building and testing installers on solaris for the few downloads that the installers receive.

sriramnrn:
Thanks for the link to the forum topic. I have certainly not been keeping track of everything on GoCD. Also, we had installed the GoCD agent for Solaris last November and have been upgrading it since then.

My situation at my current client is going to get tricky, since we have commercial support for GoCD here, and didn't get a formal EoL notification.

PR #2621, opened 2016-08-30
Add switch link between old and new agent pages

iamsumanth:
Show spinner while loading agents page


server/webapp/WEB-INF/rails.new/app/controllers/agents_controller.rb
```diff
@@ -38,6 +38,13 @@ class AgentsController < ApplicationController
   layout "application"
   prepend_before_filter :set_tab_name
 
+  def check_feature_toggle
+    if Toggles.isToggleOn(Toggles.AGENTS_SINGLE_PAGE_APP)
```
ketan: This can just be `Toggles.isToggleOn(Toggles.AGENTS_SINGLE_PAGE_APP)`, also the convention for methods that return a boolean is to add a `?` — so `is_feature_on?`

server/webapp/WEB-INF/rails.new/app/views/agents/index.html.erb
```diff
@@ -1,8 +1,12 @@
 <%- @view_title = "Agents" -%>
 <div class="page_header" style='overflow:inherit;'>
     <h1 class="entity_title">Agents</h1>
+  <% if @agents_page_feature_toggle %>
+      <a href="/go/admin/agents" class="new_agents_page">Check out the new agents page</a>
+  <% end %>
```
ketan: The idiomatic way to write this is —

```ruby
<%= link_to(..., class: :new_agents_page) if is_feature_on? %>
```

and make `is_feature_on?` a [helper function](http://api.rubyonrails.org/v4.0.4/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper_method) instead of polluting with instance variables.

server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/agents.scss
```diff
@@ -183,25 +214,27 @@ $border-color:#cccccc;
   .button[disabled] {
     opacity: 1;
     cursor: not-allowed;
+    &:hover{
```
ketan: missing whitespace before `{`?

Issue #2553, opened 2016-08-12
Missing validation errors on pipeline_config API


maheshp:
##### Issue Type
- Bug Report

##### Summary

On pipeline_config update if validation fails in any of ```Tracking tool, Environment Variables, Parameters, Artifacts, Tabs, Properties``` the API returns a error response ```422``` and a generic error message. This does not give enough information bout the reasons for validation failure.

Return validation error messages for the above mentioned entities.

rajiesh:
Validation for `properties` do not have the specific error on the element, just generic error at the pipeline level, need to be fixed

varshavaradarajan:
Non-existent roles are validated but nor represented. @GaneshSPatil is looking into it

varshavaradarajan:
Timer spec validation. Refer to the ```TimerConfig#validate()``` Error added but copied. @GaneshSPatil: similar to the parameter validation. Check your branch ```arguments-validation```

rajiesh:
Verified on GoCD version 16.11.0 (4185-e7acc53aac8bb5ad2a7f306ca8f3e752855d844b), working as expected

varshavaradarajan:
Mask password field for svn, tfs and perforce materials.

PR #2627, opened 2016-08-31
Mapped Properties error back to the JSON

GaneshSPatil:
Errors were not represented on the specific fields of the properties, only the top level error message was shown,
Now the errors related to the specific fields of a property are mapped back to the JSON,
for eg: `null` value in `property name`

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/presenters/api_v2/config/property_config_representer_spec.rb
+++ b/server/webapp/WEB-INF/rails.new/spec/presenters/api_v2/config/property_config_representer_spec.rb
@@ -1,0 +1,68 @@
+##########################################################################
+# Copyright 2016 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
related to https://github.com/gocd/gocd/issues/2553#issuecomment-243954921

server/webapp/WEB-INF/rails.new/spec/presenters/api_v2/config/property_config_representer_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    actual    = com.thoughtworks.go.config.ArtifactPropertiesGenerator.new
+    presenter = ApiV2::Config::PropertyConfigRepresenter.new(actual)
+    presenter.from_hash(property_hash)
+    expect(actual.getSrc).to eq(expected.getSrc)
+    expect(actual.getName).to eq(expected.getName)
+    expect(actual.getXpath).to eq(expected.getXpath)
+    expect(actual).to eq(expected)
+  end
+
```
varshavaradarajan: Can you refactor the above two tests a bit, seems like you can have the ```property``` object in before each, remove the expected object in the second test and do,

```
expect(actual).to eq(@property)
```

PR #2631, opened 2016-09-01
Added Agents Api V4

varshavaradarajan:
* Added build details to be represented.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api_v4/agents_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api_v4/agents_controller.rb
@@ -1,0 +1,129 @@
+##########################################################################
+# Copyright 2016 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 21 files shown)


server/webapp/WEB-INF/rails.new/app/controllers/api_v4/agents_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    def load_agent
+      @agent_instance = agent_service.findAgent(params[:uuid])
+      raise RecordNotFound if @agent_instance.nil? || @agent_instance.isNullAgent()
+    end
+
+    def agent_presenter
+      ApiV4::AgentRepresenter.new([@agent_instance, environment_config_service.environmentsFor(@agent_instance.getUuid()), security_service, current_user])
+    end
```
maheshp: Should we consider using a hash instead of an array for this, because we rely on the order of the parameters in the representer. Changing the order might lead to inconsistent behaviour.

server/webapp/WEB-INF/rails.new/config/initializers/json_hal_renderer.rb
```diff
@@ -17,7 +17,7 @@
 
 #responsible for pretty printing hal+json responses for API clients
 
-%w(v1 v2 v3).each do |version|
+%w(v1 v2 v3 v4).each do |version|
```
ketan: Could you add this type to webdefault.xml `gzipFilter`?

PR #2637, opened 2016-09-02
Refactor CaseInsensitiveString to use StringUtils for comparison

mdlsr:


ketan:
Could you tell us why this fix was needed @mdlsr?

mdlsr:
@ketan Initially this code was prone to NPE. 
To reproduce you should:
CaseInsensitiveString npeTest = new CaseInsensitiveString(null);
CaseInsensitiveString npeTest1 = new CaseInsensitiveString("anyString");
npeTest.compareTo(npeTest1); //will throw NPE. 

This is against contract. As compareTo could throw NPE only in case of null parameter. 
Unit test for this case was also added.

Also readability was improved in several places. I think it's good idea to use Apache string utils and object utils, instead of implementing this functionality ourselves.

ketan:
Closing since there's no demonstrable bug in gocd that requires this particular refactoring.

PR #2639, opened 2016-09-04
Escape the jar URL to allow symbols such as #.

tanderegg:
I am attempting to run GoCD on Mesos (https://mesos.apache.org/) using persistent volumes which ensure that even if the Mesos task dies, GoCD will recognize the new task as the same Agent and not list the old task as "lost contact".  I ran into an issue, however, since Mesos stores its persistent volumes under a directory path that contains the '#' symbol.  The GoCD utility method objectFromJar could not handle the # symbol, and would omit everything in the string after that point, thus throwing an MalformedUrlException.  This change addresses this issue by properly escaping any symbols such as # that are valid directory paths but have special meaning with an URL.  I am a new contributor, and just signed the CLA.  Thanks!

### DIFF
--- a/agent-common/src/com/thoughtworks/go/agent/common/util/JarUtil.java
+++ b/agent-common/src/com/thoughtworks/go/agent/common/util/JarUtil.java
@@ -98,10 +98,11 @@ public class JarUtil {
                 depsDir.mkdirs();
                 String entryBaseName = entryName.replaceAll(".*/", "");
                 File extractedJar = new File(depsDir, entryBaseName);
+                String escapedJarURL = new File(absolutePath).toURI().toURL().toExternalForm();
                 InputStream jarStream = null;
                 FileOutputStream fileOutputStream = null;
                 try {
-                    URL jarUrl = new URL("jar:file:" + absolutePath + "!/" + entryName);
+                    URL jarUrl = new URL("jar:" + escapedJarURL + "!/" + entryName);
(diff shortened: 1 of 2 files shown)


ketan:
Would it be possible to write a simple unit test for this?

tanderegg:
Sure, I think that should be doable.

tanderegg:
@ketan I'm run into a few issues executing the tests, I'm using `./gradlew test`, and I've had to install the Perforce client and server, but I'm still getting errors about the client not successfully connecting to the server.  Is there any documentation about how to setup the environment for testing of GoCD?  I couldn't find any in the regular documentation.  Thanks!

ketan:
Hi, sorry to respond so late. Usually you would not run `./gradlew test` because that takes forever to run, and as you said quite some dependencies too. Usually, we just run the specific tests in the IDE.

Once you think the PR is ready, you can submit it, and there is a PR build at https://build.go.cd (user:view password: password) that someone on the team can trigger.

Let us know when the PR is ready, and someone will review it. 

Also, have you signed the [CLA](https://go.cd/cla)?

tanderegg:
@ketan Alright, this should be good to go, thanks!  FYI

tanderegg:
@ketan Also, I have signed the CLA.

ketan:
@tanderegg — LGTM, I've triggered a build to check if anything else bombs :)

tanderegg:
@ketan Thanks, looks like it's all good.

Issue #2647, opened 2016-09-06
Secure variables visible in plain if defined in a template.


dennisgranath:
##### Issue Type
- Bug Report

##### Summary

Secure variables visible in plain if defined in a template.
![plain_text_secret](https://cloud.githubusercontent.com/assets/9192644/18267078/cd7fcfa0-741d-11e6-8a24-86821f8eca1a.JPG)
![template_secret](https://cloud.githubusercontent.com/assets/9192644/18267084/d465e85e-741d-11e6-9d1b-9e65fd45f4a8.JPG)

###### Basic environment details

```
Go Server Version:	16.9.0 (4001-0af23a08078987314cb93012762688f1fd60680d)
JVM version:	1.7.0_111
OS Information:	Linux 4.4.16-boot2docker
Usable space in artifacts repository:	14.4 GB
Database schema version:	1607001
```

##### Steps to Reproduce

1. Create a template with a Job that has a secure variable defined 
2. Create a pipeline using template from step 1
3. Press "settings" button for pipeline on main dashboard
4.  Press view to see the details of the template (the magnifying glass button )

##### Expected Results
The secure variable should not be visible in plain

##### Actual Results
The secure variable is visible in plain

rajiesh:
verified on 16.10.0 (4116-b8f1c5283cdec60ef25fbaf1a3b93dedaa76f975), working as expected

PR #2652, opened 2016-09-07
Do not display the value of secure environment variable in templates view. (#2647)

varshavaradarajan:
#2647


zabil:
#2647 

Write an assertion for this at
./gocd/server/webapp/WEB-INF/rails.new/spec/views/config_view/templates/show_html_spec.rb

PR #2660, opened 2016-09-08
Add code style settings for js and css

ketan:


ketan:
There are a few specific rules for css, scss, sass and JS, the rest is just indent level changes to use 2 spaces for all languages.

PR #2667, opened 2016-09-09
Fix a typo with a variable name

ketan:


ketan:
@maheshp — I'm merging this in, but could you verify if it was indeed a typo, and if this fix is good?

PR #2692, opened 2016-09-15
Improvements to the `EntityHashingService`

ketan:
Let the `EntityHashingService` be in charge of generating the cache keys and
managing if the key is supposed to be case insensitive based on the type of
the entity being passed.


ketan:
@varshavaradarajan @maheshp — could you please review this on priority please. Happy to answer any questions you may have.

PR #2693, opened 2016-09-16
Silence the server output to remove the xerces warnings when running tests

ketan:


ketan:
FYI: @zabil, I solved the problem :P

PR #2697, opened 2016-09-19
Handled Timer-Validation-Error PipelineConfig API

GaneshSPatil:
related https://github.com/gocd/gocd/issues/2553#issuecomment-247315614

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/TimerConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/TimerConfig.java
@@ -78,6 +78,10 @@ public class TimerConfig implements Validatable {
     }
 
     public void validate(ValidationContext validationContext) {
+        if (timerSpec == null) {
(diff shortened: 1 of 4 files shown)


config/config-api/src/com/thoughtworks/go/config/TimerConfig.java
```diff
@@ -78,6 +78,10 @@ public boolean validateTree(ValidationContext validationContext) {
     }
 
     public void validate(ValidationContext validationContext) {
+        if (timerSpec == null) {
+            errors.add(TIMER_SPEC, "Timer Spec can not be null.");
+            return;
+        }
```
varshavaradarajan: timer spec can be null. check ```api.go.cd```

varshavaradarajan: Okay.

PR #2700, opened 2016-09-19
Gradle task to build agent and server docker images from trunk

zabil:
WIP: Do not merge yet

To build a docker agent run the task from the root folder as

```
$ gradle installers:agentDocker
```

To build a docker server run the task from the root folder as

```
$ gradle installers:serverDocker
```

This will use the debian installers to build and tag the current version of trunk as a docker image.


maheshp:
After talking to @zabil these changes are obsolete now since the plan is to move to alpine

PR #2712, opened 2016-09-21
Log improvements

varshavaradarajan:
* Improved logs to differenciate secure environment variable errors when cipher is invalid, encrypted value is invalid and unknown errors.
* Added logs in BackupService to provide info about how long each backup took.
* Logging the exception that is caught while trying to create artifact zip file.

@arvindsv, is this what you had in mind for backups and artifact?

### DIFF
--- a/server/src/com/thoughtworks/go/server/cache/ArtifactCache.java
+++ b/server/src/com/thoughtworks/go/server/cache/ArtifactCache.java
@@ -1,5 +1,5 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -12,7 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.server.cache;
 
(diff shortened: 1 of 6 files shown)


server/src/com/thoughtworks/go/server/service/BackupService.java
```diff
@@ -116,14 +117,26 @@ public ServerBackup startBackup(Username username, HttpLocalizedOperationResult
...
                 backupConfig(destDir);
+                LOGGER.info(String.format("Finished backing up config directory. Took {%s} ms.", System.currentTimeMillis() - start));
+                LOGGER.info(String.format("Backing up config.git repository"));
+                start = System.currentTimeMillis();
                 configRepository.doLocked(new VoidThrowingFn<IOException>() {
                     @Override public void run() throws IOException {
                         backupConfigRepository(destDir);
                     }
                 });
+                LOGGER.info(String.format("Finished backing up config directory. Took {%s} ms.", System.currentTimeMillis() - start));
```
zabil: Spring has this for measuring time http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StopWatch.html use that if spring classes are available in the classpath

server/src/com/thoughtworks/go/server/cache/ArtifactCache.java
```diff
@@ -24,11 +24,13 @@
 import java.util.concurrent.ConcurrentSkipListSet;
 
 import com.thoughtworks.go.server.service.ArtifactsDirHolder;
+import org.apache.log4j.Logger;
```
zabil: Use slf4j

server/src/com/thoughtworks/go/server/cache/ArtifactCache.java
```diff
@@ -74,6 +76,7 @@ public void run() {
                     try {
                         createCachedFile(artifactLocation);
                     } catch (Exception e) {
+                        LOG.error(String.format("An error occurred while trying to create the artifact zip file of directory '%s'", artifactLocation), e);
```
zabil: with slf4j use paramerized logging http://slf4j.org/faq.html#logging_performance

server/src/com/thoughtworks/go/server/cache/ArtifactCache.java
zabil: Also refer comments on https://github.com/gocd/gocd/pull/2452
For testing the log messages.

arvindsv:
@varshavaradarajan Yes, that looks ok. Not sure about the use of org.springframework.util.StopWatch. :) I see that you've not been scarred by Spring yet.

PR #2713, opened 2016-09-21
Fix lint errors

ketan:


ketan:
Assigning to @maheshp

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/pipeline_configs/scm_config_angular_widget.js.msx
```diff
@@ -36,7 +36,7 @@ define(['mithril', 'lodash', 'jquery', 'helpers/form_helper', 'string-plus', 'he
           self.templateHTML = template(plugin.viewTemplate());
           self.defaultConfig = plugin.configurations();
 
-          self.ngController = angular.module(self.appName, []).controller(self.ngControllerName, ['$scope', '$http', function ($scope, $http) {
+          self.ngController = angular.module(self.appName, []).controller(self.ngControllerName, ['$scope', '$http', function ($scope) {
```
ketan: Can remove the `$http` here as well, I think.

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/pipeline_configs/task_basic_views_widget.js.msx
```diff
@@ -309,8 +309,6 @@ define([
                     _.each(response.data, $scope.addError, $scope);
                   } else if (response.status === 520) {
                     $scope.pluggableTaskGenericError = response.data.error;
-                  } else {
-                    console.log('Something went wrong, we do not know what!');
```
ketan: Should we throw an alert here, instead of ignoring the error?

PR #2717, opened 2016-09-22
New theme for gocd

naveenbhaskar:
* New UI theme for old pages
* All changes does as part of new-theme.scss, which overrides the existing styles.
* remove new-theme.scss or do not include to get back to old styles

WIP - Do not merge


zabil:
Closed in favor of #2809

Issue #1722, opened 2015-12-17
[PipelineConfig] General Options


maheshp:
- Fix UI & UX
- Perform simple cron validation? Nothing comprehensive, just basic stuff like maybe if there are 
- #1495 Handle pipeline templates
- Automatic pipeline scheduling checkbox (map to stage)
- Do we want to use [pretty cron](https://github.com/azza-bazoo/prettycron) to help users figure out cron syntax? See [this page](https://snap-ci.com/gocd/api.go.cd/branch/master/configuration/cron) for an example.
  sufficient number of fields?

arvindsv:
Closing. Will have to open new issues for a rewrite.

PR #2724, opened 2016-09-23
Automatic pipeline scheduling info (#1722).

varshavaradarajan:
Pending: UI

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/new_javascripts/views/pipeline_configs/pipeline_config_widget_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/new_javascripts/views/pipeline_configs/pipeline_config_widget_spec.js
@@ -63,6 +63,14 @@ define([
       expect(pipeline.enablePipelineLocking()).toBe(true);
     });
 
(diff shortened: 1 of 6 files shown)


server/webapp/WEB-INF/rails.new/spec/new_javascripts/views/pipeline_configs/pipeline_config_widget_spec.js
```diff
@@ -63,6 +63,14 @@ define([
       expect(pipeline.enablePipelineLocking()).toBe(true);
     });
 
+    it("should render the pipeline scheduling type in the pipeline settings view", function() {
+      expect($root.html()).toContain('Automatically triggered');
```
iamsumanth: you can use `jasmine-jquery.js` then it becomes `.toContainHtml("Automatically triggered")`.

ketan: more specicially `toContainText`, since we're not verifying HTML here.

Issue #2523, opened 2016-08-03
Message at startup can be changed


arvindsv:
[This](https://github.com/gocd/gocd/blob/837149ac7c64a222d7890d9b792e9915574b6f3d/installers/go-server/release/go-server.init#L78) and [this](https://github.com/gocd/gocd/blob/837149ac7c64a222d7890d9b792e9915574b6f3d/installers/go-server/release/go-server.init#L63), to mention the HTTPS port as well, maybe?

arvindsv:
Not doing this. See https://github.com/gocd/gocd/pull/2736#discussion_r80496365.

PR #2736, opened 2016-09-26
Change startup messages #2523

arvindsv:
* Go => GoCD
* Print out the https URL as well

### DIFF
--- a/installers/go-server/release/go-server.init
+++ b/installers/go-server/release/go-server.init
@@ -16,14 +16,14 @@
 #*************************GO-LICENSE-END**********************************
 
 # chkconfig: 2345 90 90
-# description: Go Server
+# description: GoCD Server
 ### BEGIN INIT INFO
 # Provides: go-server
 # Required-Start: $network $remote_fs
 # Required-Stop: $network $remote_fs
 # Default-Start: 2 3 4 5
 # Default-Stop: 0 1 6
-# Description: Start the Go Server
+# Description: Start the GoCD Server
 ### END INIT INFO
 
 PID_FILE="/var/run/go-server/go-server.pid"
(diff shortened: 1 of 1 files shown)


arvindsv:
For #2523

installers/go-server/release/go-server.init
```diff
@@ -54,13 +54,13 @@ check_proc() {
...
     if [ "${CUR_USER}" != "root" ] && [ "${CUR_USER}" != "go" ]; then
-        log_failure_msg "Go Server can only be started as 'root' or 'go' user."
+        log_failure_msg "GoCD Server can only be started as 'root' or 'go' user."
         exit 4
     fi
 
     check_proc
     if [ $? -eq 0 ]; then
-        log_success_msg "Go Server already running on http://$HOSTNAME:$GO_SERVER_PORT/go"
+        log_success_msg "GoCD Server already running on http://$HOSTNAME:$GO_SERVER_PORT/go and https://$HOSTNAME:$GO_SERVER_SSL_PORT/go"
```
ketan: An assumption of `GO_SERVER_SSL_PORT` cannot be made in the init script. You'd have to have lines similar to [this](https://github.com/gocd/gocd/blob/730ff1867576754414cc632957f344d0263bc06d/installers/go-server/release/server.sh#L77-L78), since `/etc/defaults/go-server` can possibly be empty and the `server.sh` script assumes defaults.

arvindsv: Right. I saw that. However, if GO_SERVER_PORT can be assumed to exist, then GO_SERVER_SSL_PORT should be too. If not, both should change. Right? The only option I can see apart from this is to remove the whole message.

ketan: > Right. I saw that. However, if GO_SERVER_PORT can be assumed to exist, then GO_SERVER_SSL_PORT should be too. If not, both should change. Right?

Lets make sure the script is clean. I'd much rather we not make new assumptions, even if it means they are consistent with existing assumptions :)

arvindsv: Ok.

Issue #2735, opened 2016-09-26
Agents Bulk update API: Enabling a pending agent returns 400 bad request


rajiesh:
##### Issue Type

- Bug Report
##### Summary

Using agents bulk update api enabling a pending agent returns 400 bad request
###### Basic environment details

- Go Version: `16.10.0 (4129-e51ab2ff4bc66d88a93254c209ee60e42a1112bc)`
##### Steps to Reproduce

1. Start an agent without autoRegistration key setup
2. GET request to get the agent UUID
   `curl 'http://localhost:8153/go/api/agents' -u 'username:password' -H 'Accept: application/vnd.go.cd.v4+json'`
3. Identify the pending agent UUID and PATCH call to enable the agent
   `curl -X PATCH 'http://localhost:8153/go/api/agents' -u 'username:password' -H 'Accept: application/vnd.go.cd.v4+json' -d {"uuids": ["148587a1-dd8b-48f3-a514-8e5c709f5ce8"], "agent_config_state": "Enabled"}`
4. Request returns 400 bad request with response as 
   `{
   "message": "Agent(s) with uuid(s) [148587a1-dd8b-48f3-a514-8e5c709f5ce8] not found."
   }`

PR #2740, opened 2016-09-27
Added Agent-Update-Operation Validation for pending agents

GaneshSPatil:
related https://github.com/gocd/gocd/issues/2735 and https://github.com/gocd/gocd/issues/2762

Only `enable` option is allowed for `pending agents`.

Other operations apart from `enable`( such as `resources add`,  `resources remove`,  `environments add`, `environments remove`, `disable`) on pending agents are not allowed.

https://docs.go.cd/current/navigation/agents_page.html#agents


server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -141,6 +154,41 @@ public void update(CruiseConfig preprocessedConfig) throws Exception {
...
+    }
+
+    private void handlePendingAgentsUpdate(ArrayList<String> allPendingAgentUuids, LocalizedOperationResult result, AgentInstances agentInstances, CruiseConfig preprocessedConfig) throws InvalidPendingAgentOperationException {
+        if (isInvalidOperationRequested()) {
+            result.badRequest(LocalizedMessage.string("PENDING_AGENT_INVALID_OPERATION", allPendingAgentUuids));
+            throw new InvalidPendingAgentOperationException(allPendingAgentUuids);
+        }
+        for (String agentUuid : allPendingAgentUuids) {
+            AgentInstance agent = agentInstances.findAgent(agentUuid);
+            agent.enable();
```
iamsumanth: there is no way to disable the agent, it is directly enabling irrespective of the state being passed.

server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -141,6 +154,41 @@ public void update(CruiseConfig preprocessedConfig) throws Exception {
...
+            if (allPendingAgentUUIDs.contains(uuid)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public boolean isInvalidOperationRequested() {
+        return !(resourcesToAdd.isEmpty() && resourcesToRemove.isEmpty() && environmentsToAdd.isEmpty() && environmentsToRemove.isEmpty() && state.isTrue());
+    }
```
iamsumanth: Do we need ```state.isTrue()``` check. As a functionality we allow pending agents to be disabled.

config/config-api/src/com/thoughtworks/go/config/AgentConfig.java
```diff
@@ -145,6 +146,10 @@ public boolean isDisabled() {
         return Boolean.TRUE.equals(isDisabled);
     }
 
+    public boolean isEnabled() {
+        return Boolean.FALSE.equals(isDisabled);
```
varshavaradarajan: can be ```!isDisabled()```

config/config-api/src/com/thoughtworks/go/config/AgentConfig.java
```diff
@@ -165,6 +170,10 @@ public boolean isNull() {
         return false;
     }
 
+    public AgentConfig deepClone() {
+        return new Cloner().deepClone(this);
```
varshavaradarajan: I see that this method is used in method ```findPendingAgents```. Why does the agent config need to be cloned to add to the pending agent list?

server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -60,55 +65,38 @@ public AgentsEntityConfigUpdateCommand(Username username, LocalizedOperationResu
...
     public void update(CruiseConfig preprocessedConfig) throws Exception {
-        List<AgentConfig> goodAgents = getValidAgents(uuids, result, preprocessedConfig.agents());
-        checkElasticAgentsResourceUpdated(goodAgents, resourcesToAdd, resourcesToRemove);
-
         Set<CaseInsensitiveString> allEnvironmentNames = new HashSet<>(goConfigService.getEnvironments().names());
 
-        validateEnvironment(allEnvironmentNames, environmentsToAdd, result);
-        validateEnvironment(allEnvironmentNames, environmentsToRemove, result);
+        // validate all inputs
+        validateEnvironmentNames(allEnvironmentNames, environmentsToAdd);
```
varshavaradarajan: Consider renaming the method, as you are just checking for the presense of the environment in the config.

server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -60,55 +65,38 @@ public AgentsEntityConfigUpdateCommand(Username username, LocalizedOperationResu
...
-                result.badRequest(LocalizedMessage.string("RESOURCE_NOT_FOUND", "Environment", environmentName));
-                throw new NoSuchEnvironmentException(environmentName);
-            }
-        }
+        result.badRequest(LocalizedMessage.string("NO_OPERATION_PERFORMED_ON_AGENTS"));
+        return false;
     }
 
     @Override
     public void update(CruiseConfig preprocessedConfig) throws Exception {
```
varshavaradarajan: Its modifiedConfig. The preprocessing of the config happens only after update.

server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -60,55 +65,38 @@ public AgentsEntityConfigUpdateCommand(Username username, LocalizedOperationResu
...
         Set<CaseInsensitiveString> allEnvironmentNames = new HashSet<>(goConfigService.getEnvironments().names());
 
-        validateEnvironment(allEnvironmentNames, environmentsToAdd, result);
-        validateEnvironment(allEnvironmentNames, environmentsToRemove, result);
+        // validate all inputs
+        validateEnvironmentNames(allEnvironmentNames, environmentsToAdd);
+        validateEnvironmentNames(allEnvironmentNames, environmentsToRemove);
 
-        for (AgentConfig agentConfig : goodAgents) {
+        validateAgentUUIDs();
```
varshavaradarajan: Consider renaming the method, as you are only checking for the presence of the agent in the config.

server/src/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommand.java
```diff
@@ -141,26 +129,94 @@ public void update(CruiseConfig preprocessedConfig) throws Exception {
         }
     }
 
-    private void checkElasticAgentsResourceUpdated(List<AgentConfig> uuids, List<String> resourcesToAdd, List<String> resourcesToRemove) throws ElasticAgentsResourceUpdateException {
-        ArrayList<String> elasticAgentUUIDs = findAllElasticAgentUuids(uuids);
+    private void validateAgentUUIDs() throws NoSuchAgentException {
```
varshavaradarajan: We can have a generic exception ```NoSuchResourceException``` instead of creating an exception class for each resource.

server/src/com/thoughtworks/go/server/service/AgentConfigService.java
```diff
@@ -232,13 +233,13 @@ public ConfigModifyingUser user() {
...
+    public void bulkUpdateAgentAttributes(AgentInstances agentInstances, final Username username, final LocalizedOperationResult result, final List<String> uuids, final List<String> resourcesToAdd, final List<String> resourcesToRemove, final List<String> environmentsToAdd, final List<String> environmentsToRemove, final TriState enable) {
+        Localizable.CurryableLocalizable successMessage = LocalizedMessage.string("BULK_AGENT_UPDATE_SUCESSFUL", StringUtils.join(uuids, ", "));
+        EntityConfigUpdateCommand<Agents> agentsEntityConfigUpdateCommand = new AgentsEntityConfigUpdateCommand(agentInstances, username, result, uuids, environmentsToAdd, environmentsToRemove, enable, resourcesToAdd, resourcesToRemove, goConfigService);
         try {
             goConfigService.updateConfig(agentsEntityConfigUpdateCommand, username);
             if(result.isSuccessful()){
-                result.setMessage(LocalizedMessage.string("BULK_AGENT_UPDATE_SUCESSFUL", StringUtils.join(uuids, ", ")));
+                result.setMessage(successMessage);
             }
         } catch (Exception e) {
```
varshavaradarajan: Please check the messages for validation errors (post validations). Since we add errors on the agent config object and throw the GoConfigInvalidException, we must show 422 (unprocessable entity) instead of 500 (internal server error)

server/test/unit/com/thoughtworks/go/config/update/AgentsEntityConfigUpdateCommandTest.java
```diff
@@ -0,0 +1,261 @@
...
+    private Username currentUser;
+    private BasicCruiseConfig cruiseConfig;
+    private GoConfigService goConfigService;
+    private AgentInstances agentInstances;
+    List<String> uuids;
+    List<String> environmentsToAdd;
+    List<String> environmentsToRemove;
+    List<String> resourcesToAdd;
+    List<String> resourcesToRemove;
+    TriState triState;
```
varshavaradarajan: The above instance variables can be made private.

config/config-api/src/com/thoughtworks/go/config/AgentConfig.java
iamsumanth: We get the pending agent from AgentInstances object which is a live instance and we do not want to modify it. That is why we deep clone the object and perform modifications on top of that.

PR #2745, opened 2016-09-28
Refactor agent controller to switch between http and websockets

zabil:
Pulled up an AgentController
Instantiate the agent controller using a factory method.
Refactored the web socket controller and service to avoid  circular dependencies.
The agent service will only manage the connections and session, while the controller will handle the messages.

Changes in jetty 9.2.19

jetty-9.2.19.v20160908 - 08 September 2016
 + 817 NPE in jndi Resource
 + 830 Test webapp not properly copied to demo-base
 + 832 ServerWithJNDI example uses wrong webapp
 + 851 MBeanContainer no longer unregisters MBeans when "stopped"
 + 868 ClassLoader leak with Jetty and Karaf - static instances of
   java.lang.Throwable
 + 880 Refactor jetty-http's HostPortHttpField logic into new jetty-util class
 + 882 Add IPv6 support to IPAddressMap in jetty-util
 + 894 When adding servless class, preserve Class instead of going through
   String
 + 899 PathFinderTest fails in jetty-9.2.x


agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
```diff
@@ -0,0 +1,272 @@
...
+    public void execute() {
+        // Do nothing
+    }
+
+    @Override
+    public void work() throws Exception {
+        if (sslInfrastructureService.isRegistered()) {
+            if (webSocketSessionHandler.isNotRunning()) {
+                webSocketSessionHandler.clearCallBacks();
+                synchronized (this) {
```
jyotisingh: Why is the sync required? If this is done to synchronize on session, then shouldn't all the three calls to `webSocketSessionHandler` be in the same synchronized block, and may be we should be double checking on `webSocketSessionHandler.isNotRunning()` as well.

agent/src/com/thoughtworks/go/agent/AgentController.java
```diff
@@ -17,231 +17,106 @@
...
 
-    private void handleIfSecurityException(Exception e) {
-        if (!isCausedBySecurity(e)) {
-            return;
-        }
+    private void handleSecurityException(Exception e) {
         sslInfrastructureService.invalidateAgentCertificate();
         LOG.error("There has been a problem with one of Go's SSL certificates." +
-                        " This can be caused by a man-in-the-middle attack, or by pointing the agent to a new server, or by" +
+                        " This can be caused by a man-in-the-middle attack, or by pointing the agent to a new e, or by" +
```
jyotisingh: Typo : 'server' replaced by 'e' :)

agent/resources/applicationContext.xml
```diff
@@ -17,106 +17,89 @@
...
-        <property name="scheduledTimerTasks">
-            <list>
-                <bean class="org.springframework.scheduling.timer.ScheduledTimerTask"
-                      p:delay="${agent.get.work.delay}"
-                      p:period="${agent.get.work.interval}"
-                      p:timerTask-ref="agentControllerLooper"/>
-            </list>
-        </property>
-    </bean>
+  <bean id="agentController" class="com.thoughtworks.go.agent.AgentController" factory-method="createInstance">
```
jyotisingh: You could create a `AgentControllerFactory` with non-static factory-method, then you wouldn't have to pass on `constructor-arg` from the xml file. You could just rely on spring to autowire the constructor arguments, and then a parameter-less factory method to instantiate the desired controller. The xml would look something like this then:
``` xml
  <bean id="agentControllerFactory" class="com.thoughtworks.go.agent.AgentControllerFactory" />
  <bean id="agentController" class="com.thoughtworks.go.agent.AgentController" factory-bean="agentControllerFactory" factory-method="createInstance" />
```

AgentControllerFactory.java:
``` java
@Component
public class AgentControllerFactory {
    private final BuildRepositoryRemote server;
    private final GoArtifactsManipulator manipulator;
    private final PluginManager pluginManager;
    private final SslInfrastructureService sslInfrastructureService;
    private final AgentRegistry agentRegistry;
    private final SubprocessLogger subprocessLogger;
    private final AgentUpgradeService agentUpgradeService;
    private final PackageAsRepositoryExtension packageAsRepositoryExtension;
    private final SCMExtension scmExtension;
    private final TaskExtension taskExtension;
    private final HttpService httpService;
    private final WebSocketClientHandler webSocketClientHandler;
    private final WebSocketSessionHandler sessionHandler;

    @Autowired
    public AgentControllerFactory(
            BuildRepositoryRemote server,
            GoArtifactsManipulator manipulator,
            AgentUpgradeService agentUpgradeService,
            PluginManager pluginManager,
            SslInfrastructureService sslInfrastructureService,
            SystemEnvironment systemEnvironment,
            AgentRegistry agentRegistry,
            SubprocessLogger subprocessLogger,
            PackageAsRepositoryExtension packageAsRepositoryExtension,
            SCMExtension scmExtension,
            TaskExtension taskExtension,
            HttpService httpService,
            WebSocketClientHandler webSocketClientHandler, WebSocketSessionHandler sessionHandler) {
        this.server = server;
        this.manipulator = manipulator;
        this.pluginManager = pluginManager;
        this.sslInfrastructureService = sslInfrastructureService;
        this.systemEnvironment = systemEnvironment;
        this.agentRegistry = agentRegistry;
        this.subprocessLogger = subprocessLogger;
        this.agentUpgradeService = agentUpgradeService;
        this.packageAsRepositoryExtension = packageAsRepositoryExtension;
        this.scmExtension = scmExtension;
        this.taskExtension = taskExtension;
        this.httpService = httpService;
        this.webSocketClientHandler = webSocketClientHandler;
        this.sessionHandler = sessionHandler;
        PluginManagerReference.reference().setPluginManager(pluginManager);
    }

    public AgentController createInstance() {
        if (systemEnvironment.isWebsocketEnabled()) {
            return new AgentWebSocketClientController(
                    server,
                    manipulator,
                    sslInfrastructureService,
                    agentRegistry,
                    agentUpgradeService,
                    subprocessLogger,
                    systemEnvironment,
                    pluginManager,
                    packageAsRepositoryExtension,
                    scmExtension,
                    taskExtension,
                    httpService, webSocketClientHandler, sessionHandler);
        } else {
            return new AgentHTTPClientController(
                    server,
                    manipulator,
                    sslInfrastructureService,
                    agentRegistry,
                    agentUpgradeService,
                    subprocessLogger,
                    systemEnvironment,
                    pluginManager,
                    packageAsRepositoryExtension,
                    scmExtension,
                    taskExtension);
        }
    }
}
```

Technically, it doesn't make much difference, but it keeps us from updating the xml file everytime the signature changes.

agent/src/com/thoughtworks/go/agent/AgentController.java
```diff
@@ -17,231 +17,106 @@
...
+    protected AgentAutoRegistrationProperties getAgentAutoRegistrationProperties() {
+        return agentAutoRegistrationProperties;
+    }
+
+    protected SystemEnvironment getSystemEnvironment() {
+        return systemEnvironment;
+    }
+
+    protected AgentIdentifier agentIdentifier() {
+        return new AgentIdentifier(SystemUtil.getLocalhostNameOrRandomNameIfNotFound(), SystemUtil.getClientIp(systemEnvironment.getServiceUrl()), agentRegistry.uuid());
```
jyotisingh: Hostname and IP computations have been inlined here. Given that this method is called during every ping and loop call, does it make sense to determine this once and reuse thereafter?

agent/src/com/thoughtworks/go/agent/DefaultBuildStateReporter.java
```diff
@@ -15,7 +15,6 @@
  * ************************GO-LICENSE-END***********************************/
```
jyotisingh: license needs to be updated

agent/resources/applicationContext.xml
zabil: Yes I tried this as the first option.
However spring is not able to calculate the dependencies and fails in the object creation order.

zabil: Yes I tried this as the first option.
However spring is not able to calculate the dependencies and fails in the object creation order.

agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
zabil: You are right the synchronized is not required here.

agent/src/com/thoughtworks/go/agent/DefaultBuildStateReporter.java
zabil: Isn't it supposed to be 2016?
What update did you mean?

agent/resources/applicationContext.xml
jyotisingh: Hmm...it worked when I tried this out..

agent/src/com/thoughtworks/go/agent/DefaultBuildStateReporter.java
jyotisingh: I meant, its the old style copyright with `GO-LICENSE-END`. We have been moving the files we touch to the new one such as the one you have in AgentWebSocketClientController.java. Also, files like BuildRepositoryRemoteAdapter.java don't seem to have a copyright notice.

PR #2749, opened 2016-09-29
Update after-upgrade.sh.erb

netdevfighter:
After updating the 'go-agent' is not started again.
Or explain me what is this reason for this "if condition"?

### DIFF
--- a/installers/linux/deb/after-upgrade.sh.erb
+++ b/installers/linux/deb/after-upgrade.sh.erb
@@ -18,7 +18,4 @@
 
 <%= ERB.new(File.read(File.join(install_scripts_dir, 'shared', 'partials', '_set-debian-permissions.sh.erb')), nil, '-', "_#{SecureRandom.hex}").result(binding) %>
 
-<% if name == 'go-server' -%>
-    # Start the service
-    /etc/init.d/go-server start
-<% end -%>
+    /etc/init.d/<%= name %> start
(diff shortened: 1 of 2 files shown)


ketan:
@arvindsv — this was done to keep parity with the old installers, and I don't really see the need for the `if` condition. Do you have any concerns with merging this PR in?

@netdevfighter Would you mind signing the [CLA](https://www.go.cd/cla). It's a one time requirement for contributing to GoCD?

arvindsv:
@ketan I'm fine with removing it, though I'm not sure I understand how the name is not go-server. What is it then?

ketan:
It's a common script shared between the agent and server. So the mane can
be either of them.

arvindsv:
Ah, I see. The *agent* does not start up. Ok, makes sense. The reason the agent doesn't start up normally, is that the GO_SERVER_URL will need to be changed. But, that's not true after an upgrade. Maybe true during an install. Since this is after_upgrade.erb, it should be fine to remove the check. I remember us talking about this.

netdevfighter:
@ketan The CLA signing was done.

varshavaradarajan:
@netdevfighter Can you do the same for ```linux/deb/after-upgrade.sh.erb``` ?

ketan:
@varshavaradarajan Could we hold onto merging this? Not saying the fix is incorrect. But I'm considering taking some steps to bring parity between deb and rpm installers. Currently they're slightly different from each other in how they behave with respect to fresh installs and upgrades.

netdevfighter:
@varshavaradarajan done

ketan:
Will fix this as part of #3119.

Issue #484, opened 2014-09-09
Need ability to configure location of pipelines directory on agent


reidca:
Currently the agents use a directory below the path of the executable for the pipeline jobs.

This directory is used to store fetched materials and is the default working directory for the job execution.

On windows the default install path is "C:\Program Files (x86)\Go Agent" which means the pipeline location is "C:\Program Files (x86)\Go Agent\pipelines" .

There are THREE  problems with this location:
1. The path to the pipelines is quite long and since TFS and other .NET based programs have a bug where paths exceeding 260 characters in length do not work, this causes a number of problems when for example downloading materials from TFS.
2. The PROGRAM FILES location is a well known path which is used to store programs however the pipeline folder really contains data, temporary data in fact. If you run the GO Agent a user other than LOCAL SYSTEM or an Administrator then these users do not have write permission to the PROGRAM FILES location and this must be assigned manually.
3. The pipelines folder can take up a very large amount of disk space however it is quite hidden in its current location. System Administrators and users not familiar with GO who are looking to recover disk space are unlikely to consider deleting the contents of this folder even though it is a temporary data folder.

Adding a configuration property to the GO configuration files that allow this directory to be specified for each agent and defaulting it to the TEMP location would resolve these three issues.

pmcgrath:
+1 Having the same issue with running pipelines where file paths are long on Windows systems

arvindsv:
Does the core.longpaths option mentioned [here](https://groups.google.com/d/msg/go-cd/qPAMe9kgvlM/pCoZ4rLmCQAJ) help?

efaruk:
All windows user's have the same problem, please someone!? make that pull request...

manojjenago:
Hi ,

I am getting error "C:\RIT_Build_Process\eBayProject-build.xml:11: Could not create parameter file in default temporary-file directory" .
It seems that GOCD is trying to create files in Temp directory. I changed the System properties to point to the location where I have the access. (TEMP= C:\Temp) .Still it fails.

Is there any way to mention the temp dir for GOCD . Is there any file that can be set for Temp dir??
As "reidca" mentioned creating a configuration property to the GO should solve the issues. Please let me know when I should create it.

Thanks in advance,
Manoj

arvindsv:
Can you provide some more information about where you see this error? In the console log? If so, can you provide some more lines for context. I'm trying to see if it is happening from the GoCD side, or from a custom command.

reidca:
[...]

PR #2754, opened 2016-10-01
#484 - Allow flyweight and agent work dirs to be configurable

arvindsv:
Implements the work mentioned in #484.

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java
+++ b/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java
@@ -39,6 +39,7 @@ import org.apache.log4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
@@ -60,8 +61,8 @@ public class SslInfrastructureService {
     private transient boolean registered = false;
 
     @Autowired
-    public SslInfrastructureService(URLService urlService, GoAgentServerHttpClient httpClient, AgentRegistry agentRegistry) throws Exception {
-        this(new RemoteRegistrationRequester(urlService.getAgentRegistrationURL(), agentRegistry, httpClient), httpClient);
+    public SslInfrastructureService(URLService urlService, GoAgentServerHttpClient httpClient, AgentRegistry agentRegistry, SystemEnvironment systemEnvironment) throws Exception {
+        this(new RemoteRegistrationRequester(urlService.getAgentRegistrationURL(), agentRegistry, httpClient, systemEnvironment), httpClient);
     }
 
     // For mocking out remote call
@@ -151,13 +152,15 @@ public class SslInfrastructureService {
 
     public static class RemoteRegistrationRequester {
         private final AgentRegistry agentRegistry;
+        private SystemEnvironment systemEnvironment;
         private String serverUrl;
         private GoAgentServerHttpClient httpClient;
 
-        public RemoteRegistrationRequester(String serverUrl, AgentRegistry agentRegistry, GoAgentServerHttpClient httpClient) {
+        public RemoteRegistrationRequester(String serverUrl, AgentRegistry agentRegistry, GoAgentServerHttpClient httpClient, SystemEnvironment systemEnvironment) {
             this.serverUrl = serverUrl;
             this.httpClient = httpClient;
             this.agentRegistry = agentRegistry;
+            this.systemEnvironment = systemEnvironment;
         }
 
         protected Registration requestRegistration(String agentHostName, AgentAutoRegistrationProperties agentAutoRegisterProperties) throws IOException, ClassNotFoundException {
@@ -165,11 +168,12 @@ public class SslInfrastructureService {
                 LOGGER.debug(String.format("[Agent Registration] Using URL %s to register.", serverUrl));
             }
 
+            String agentWorkDir = systemEnvironment.resolveAgentWorkingDirectory(new File(".")).getAbsolutePath();
             HttpRequestBase postMethod = (HttpRequestBase) RequestBuilder.post(serverUrl)
                     .addParameter("hostname", agentHostName)
                     .addParameter("uuid", agentRegistry.uuid())
-                    .addParameter("location", SystemUtil.currentWorkingDirectory())
-                    .addParameter("usablespace", String.valueOf(AgentRuntimeInfo.usableSpace(SystemUtil.currentWorkingDirectory())))
+                    .addParameter("location", agentWorkDir)
+                    .addParameter("usablespace", String.valueOf(AgentRuntimeInfo.usableSpace(agentWorkDir)))
                     .addParameter("operatingSystem", new SystemEnvironment().getOperatingSystemCompleteName())
                     .addParameter("agentAutoRegisterKey", agentAutoRegisterProperties.agentAutoRegisterKey())
                     .addParameter("agentAutoRegisterResources", agentAutoRegisterProperties.agentAutoRegisterResources())
--- a/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java
+++ b/agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java
@@ -168,7 +168,7 @@ public class SslInfrastructureService {
                 LOGGER.debug(String.format("[Agent Registration] Using URL %s to register.", serverUrl));
             }
 
-            String agentWorkDir = systemEnvironment.resolveAgentWorkingDirectory(new File(".")).getAbsolutePath();
+            String agentWorkDir = systemEnvironment.resolveAgentWorkingDirectory().getAbsolutePath();
             HttpRequestBase postMethod = (HttpRequestBase) RequestBuilder.post(serverUrl)
                     .addParameter("hostname", agentHostName)
                     .addParameter("uuid", agentRegistry.uuid())
--- a/agent/test/unit/com/thoughtworks/go/agent/JobRunnerTest.java
+++ b/agent/test/unit/com/thoughtworks/go/agent/JobRunnerTest.java
@@ -61,6 +61,7 @@ public class JobRunnerTest {
     private BuildWork buildWork;
     private AgentIdentifier agentIdentifier;
     private UpstreamPipelineResolver resolver;
+    private SystemEnvironment systemEnvironment;
 
     public static String withJob(String jobXml) {
         return "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n"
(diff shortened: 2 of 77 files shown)


arvindsv:
This is what was done to test this:

1. <details>
        <summary>Setup Perforce server and a repo to work with.</summary>
    ```
    $ docker run -d -p 8080:8080 -p 1666:1666 -h perforce --name perforce ambakshi/perforce-server
    # Run "docker logs -f perforce" if you want, now.
    
    # Do this in an empty directory
    $ export P4PORT=localhost:1666 P4PASSWD='pass12349ers!' P4USER=p4admin P4CLIENT=myp4client
    $ p4 client
    Perforce password (P4PASSWD) invalid or unset.
    Attempting login for user 'p4admin' against server 'localhost:1666'
    Enter password:     <--- Enter the same password (once) here: pass12349ers!
    User p4admin logged in.
    
    # An editor opens. You can close it. It'll show a message "Client myp4client saved."
    
    $ touch a_file
    $ p4 add a_file
    //depot/a_file#1 - opened for add
    /your/path/a_file - empty, assuming text.
    
    $ p4 submit
    # An editor opens. Change the description part to something - it's a commit message
    # Once you exit, it shows some messages.
    Change 1 created with 1 open file(s).
    Submitting change 1.
    Locking 1 files ...
    add //depot/a_file#1
    Change 1 submitted.
    
    $ p4 sync
    File(s) up-to-date.
    ```
    </details>

2. Download 16.10.0 server and agent

3. Use the config and cipher from [here](https://gist.github.com/arvindsv/4e62da7b3e917bb234db7bea7321bf2c) and move it into the config/ directory on the server

4. Download the script executor plugin from [here](https://github.com/gocd-contrib/script-executor-task/releases) and move it into plugins/external/

5. Keep track of the values of:
   1. In "Test" pipeline: The material checkout directory (should be default).
   2. In "Test" pipeline: Output of first exec task: pwd
   3. In "Test" pipeline: Output of ant task which runs inside "innerdir".
   4. In "Test" pipeline: Output of exec task which runs inside "innerdir".
   5. In "Test" pipeline: Output of rake task which uses innerdir/Rakefile but runs outside.
   6. In "Test" pipeline: Output of rake task which runs inside "innerdir".
   7. In "Test" pipeline: Output of plugin task which runs in main dir.
   8. In "Test" pipeline: The message "Uploading artifacts from".
   9. In "Test" pipeline: The message "Start to create properties" should say: junit.tests = 3
   10. In "Test" pipeline: Test report should be created (from sample.junit.xml)
   11. In "Downstream" pipeline: Prints out working directory around "Start to prepare"
   12. In "Downstream" pipeline: Prints out pwd
   13. In "Downstream" pipeline: Fetches artifact into default work dir (verify on filesystem)
   14. In "WithPerforceMaterial" pipeline: Perforce clone/update prints out full work dir
   15. In "WithPerforceMaterial" pipeline: GO_P4_CLIENT_PERFORCE environment variable value
   16. On "Agents" page: The value of free space and the value of the sandbox directory

   For Windows, use "TestWindows" and "DownstreamWindows" instead of "Test" and "Downstream".

6. Stop 16.10.0 server and agent

7. Upgrade to new server and agent with this change, but the same config, cipher, plugins, etc.

8. Start server and agent with no overridden system properties. Then, force the "Test" and "WithPerforceMaterial" pipelines. The "Downstream" pipeline will trigger automatically.

9. All values noted in point 5 above should be the same now (almost - there are some differences where it was earlier saying /var/lib/go-agent and now it'll say /var/lib/go-agent/pipelines/Test and I *think* it's ok).

10. Especially, the `GO_P4_CLIENT_PERFORCE` environment variable value should be the same.

11. Stop server and agent. Rename the server flyweight directory (pipelines/flyweight) and the agent work directory (pipelines/Test, pipelines/Downstream and pipelines/WithPerforceMaterial), if you want, to make sure that they're not used at all.

12. Start server with: `GO_SERVER_SYSTEM_PROPERTIES='-Dserver.flyweight.dir=/tmp/new.server.flyweight.dir' ./server.sh`

13. Create a small disk and use that as the agent work directory (to check free space properly). On Mac:

    ```
    hdiutil create -size 10m -volname small_10m_disk small_10m_disk.dmg -fs HFS+J
    open small_10m_disk.dmg # This mounts /Volumes/small_10m_disk
    ```
    
    Start agent with: `GO_AGENT_SYSTEM_PROPERTIES="-Dagent.work.dir=/Volumes/small_10m_disk" ./agent.sh`

    For Windows, you can similarly create a VHD: https://technet.microsoft.com/en-us/magazine/ee872416.aspx and give it a size and assign it a drive letter (like E:).

14. All values in point 5 should be changed appropriately, when pipelines "Test" and "WithPerforceMaterial" are retriggered. However, GO_P4_CLIENT_PERFORCE should change, since the path has changed. Also, make sure that the free space shown is small (since the agent work dir has only 10MB of space).

15. If you had deleted the old work and flyweight directories, make sure they've not come back.

arvindsv:
There's a PR build for it [here](https://build.go.cd/go/tab/build/detail/installers-PR/89/dist/1/dist).

arvindsv:
To review this: It should be easiest to review it commit by commit - oldest to newest. Overall, all that has happened is any usage of current working directory is replaced by a call which will resolve the working directory based on a system property.

Also, from the server side, when a job was sent to the agent, the directory was sent as "pipelines/PIPELINE_NAME/working_dir". Now, it is sent as: "PIPELINE_NAME/working_dir". This allows the overridden working directory to not have "pipelines/" as a part of its name. If you override the working dir to: "/abc", then the working dir for the pipeline will be: "/abc/PIPELINE_NAME".

common/src/com/thoughtworks/go/domain/Pipeline.java
```diff
@@ -59,7 +59,7 @@ public Pipeline(String pipelineName, BuildCause buildCause, Stage... stages) {
     }
 
     public File defaultWorkingFolder() {
-        return new File(CruiseConfig.WORKING_BASE_DIR + pipelineName);
+        return new File("" + pipelineName);
```
ketan: It seems odd to be be calling `defaultWorkingFolder` on a pipeline with no name.

agent/src/com/thoughtworks/go/agent/service/SslInfrastructureService.java
```diff
@@ -148,25 +149,28 @@ private void deleteKeyStores() {
...
         protected Registration requestRegistration(String agentHostName, AgentAutoRegistrationProperties agentAutoRegisterProperties) throws IOException, ClassNotFoundException {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug(String.format("[Agent Registration] Using URL %s to register.", serverUrl));
             }
 
+            String agentWorkDir = systemEnvironment.resolveAgentWorkingDirectory(new File(".")).getAbsolutePath();
             HttpRequestBase postMethod = (HttpRequestBase) RequestBuilder.post(serverUrl)
                     .addParameter("hostname", agentHostName)
                     .addParameter("uuid", agentRegistry.uuid())
-                    .addParameter("location", SystemUtil.currentWorkingDirectory())
```
arvindsv: I am trying ... and there are 150 usages in tests. I could get it down to 20 usages by extracting a method in each of those tests and then I'd need to inline it. Do you think it's worth it, for what is essentially `new File(".").getAbsolutePath`?

common/src/com/thoughtworks/go/domain/Pipeline.java
arvindsv: Makes sense not to have it (we should never have null pipeline name). The reason for it was probably only because [some tests](https://build.go.cd/go/tab/build/detail/build-linux-PR/813/build-server/1/server-split-runInstance-6) failed. My guess is that the tests are wrong. So, I'll fix them instead.

arvindsv: Oh, I remember why I did this. It's because it's passed something called a `NullPipeline` in the test (it has `NullStage`s, etc). `NullPipeline` is also used in non-test code (around scheduler) and I couldn't rule out the possibility that it is not ever going to call `defaultWorkingFolder`. I suspect it is not. Do you have any idea if it is, @jyotisingh?

I can change the tests, but don't know if the tests are really representative, in this case.

ketan: Can we make the intention clear?

```
if this instanceof NullPipeline
  return ...
else
  return ...
end
```

arvindsv: There's only [one usage](https://github.com/gocd/gocd/blob/4509fe7c72bf919c33bf212752e88a45dc09bbbd/common/src/com/thoughtworks/go/domain/StageIdentifier.java#L32) of [this default constructor](https://github.com/gocd/gocd/blob/4509fe7c72bf919c33bf212752e88a45dc09bbbd/common/src/com/thoughtworks/go/domain/NullPipeline.java#L22-L24). What do you think of removing that constructor and replacing the call with

```
new NullPipeline("")
```

Then, I can remove the `"" + pipelineName` bit. If not, I'll do what you suggested. Though explicit, it makes it feel a little wrong.

arvindsv:
@ketan @rajiesh - I guess this is not going to go into 16.11? If so, can you please change the milestone? Thanks! I realize it's not the most trivial change and requires looking into. :)

ketan:
@arvindsv — Could you rebase the PR and trigger a PR build?

arvindsv:
@ketan Done. It's [running](https://build.go.cd/go/pipelines/value_stream_map/trigger/246). I'll take a look at it later or tomorrow.

agent/test/unit/com/thoughtworks/go/agent/JobRunnerTest.java
```diff
@@ -143,31 +145,31 @@ private BuildWork getWork(JobConfig jobConfig) {
 
         List<Builder> builder = BuilderMother.createBuildersAssumingAllExecTasks(config, pipelineName, stageName, JOB_PLAN_NAME);
 
-        BuildAssignment buildAssignment = BuildAssignment.create(jobPlan, BuildCause.createWithEmptyModifications(), builder, new File(CruiseConfig.WORKING_BASE_DIR + pipelineName));
```
arvindsv: This is done.

arvindsv:
Hmm. According to the functional tests, I've broken something around package repos. I'll take a look.

arvindsv:
@ketan False alarm. The tests weren't consistent. Something to do with a local package repo they have. [All green](https://build.go.cd/go/pipelines/value_stream_map/build-linux-PR/876).

arvindsv:
I'll need to rebase, though.

ketan:
@arvindsv — could you rebase one last time and retrigger the PR build?

arvindsv:
This will happen today.

arvindsv:
Found a small bug around checksum files. That's why this is not done yet. Anyway ...

arvindsv:
@ketan - Tests have passed on this PR. @rajiesh has paused the smoke and acceptance PR pipelines. So, it hasn't run. I guess this is again too close to a release that it won't go in anyway. I'll rebase again, in a few days.

arvindsv:
Closing for now. Will need to rebase.

johusman:
May I ask whatever happened to this? This doesn't seem to have made its way into the code base.

ryudice:
Was this released?

Issue #2753, opened 2016-10-01
Quick edit adds extra element to ant/rake tasks


arvindsv:
##### Issue Type

- Bug Report

##### Summary

Quick edit adds optional attributes of tasks upon save, and defaults them to empty. This causes some tasks to become invalid when run.

###### Basic environment details

* Go Version: 16.10.0
* JAVA Version: 1.7
* OS: Linux

##### Steps to Reproduce

1. Have a config with these tasks:

    ```
    <ant target="dist" workingdir="somedir">
      <runif status="passed" />
    </ant>

    <ant buildfile="abc" target="dist">
      <runif status="passed" />
    </ant>
    ```

2. Open the "Quick Edit" for the pipeline.

3. Click "Save".

4. Check the XML.
 
##### Expected Results

No change in the XML.

##### Actual Results

The XML would have changed to:

```
<ant buildfile="" target="dist" workingdir="somedir">
  <runif status="passed" />
</ant>

<ant buildfile="abc" target="dist" workingdir="">
  <runif status="passed" />
</ant>
```

The workingdir is harmless (though wrong), but the "buildfile" attribute makes the ant task incorrect (won't run).

##### Any other info

This happens for Rake as well, and I guess Nant too. Maybe for anything (or just any task) which has an optional attribute?

rajiesh:
Verified the fix on GoCD version 16.11.0 (4164-1d1fbc36cd04b463a8cf2142cb6eb1243023b412) and working as expected

PR #2755, opened 2016-10-03
Added a blank check for build tasks attributes. (#2753)

varshavaradarajan:
#2753

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/tasks/ant_task_representer.rb
+++ b/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/tasks/ant_task_representer.rb
@@ -1,5 +1,5 @@
 ##########################################################################
-# Copyright 2015 ThoughtWorks, Inc.
+# Copyright 2016 ThoughtWorks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


arvindsv:
In #2753, something is adding that blank attribute, though (don't know if it's sent over in the JSON request or happening during deserialization). Don't know if this fixes that.

Also, adding this check retroactively (without a config migration to remove the blank values in existing config) might cause issues upon editing one of those configs, right?

ketan:
IMO, we should not be calling the setter unless there's something to be set. Handle this in the representer, should we?

varshavaradarajan:
@arvindsv, In quick edit, we default to a blank string if the user input is not specified. I chose not to default to null on the javascript side because the using the api independently would still have this problem. 
@ketan , we call the setter from the pipeline config API representers. From the old UI, the ```setConfigAttributes``` method on the config entity objects are used which makes these checks anyway, hopefully for all config objects. Doing this check only at the representer would have meant that someone directly editing the xml would still face the issue.

Also, I don't think a config migration was needed because this is not called from the ```validateTree``` method, thus not breaking the existing xml. But only to set the future values to null upon encountering a blank string (while setting it using the api only). We should not be validating these fields as they are optionally provided.

It isn't harmful that the xml may have an empty string for these values as @arvindsv  pointed in the issue, except for the ```buildFile``` attribute. However, if we decide that a config migration is needed, I'll send another PR. :)

arvindsv:
Talked to @varshavaradarajan: Decided not to have a config migration for now. If this has affected anyone, we would have learnt about it.

Issue #2758, opened 2016-10-04
ETag value contain `--gzip` postfix when using Accept-Encoding:gzip


grundic:
##### Summary

Setting `Accept-Encoding` header to `gzip` value causes ETag value to contain `--gzip` postfix, which intended to be removed by `JettyWeakEtagMiddleware` class in `WEB-INF/rails.new/lib/jetty_weak_etag_middleware.rb`. 

###### Basic environment details

* Go Version: `16.9.0`
* JAVA Version: `1.7.0_111-b01`

##### Steps to Reproduce

Run curl command:
```
curl -s -D - -o /dev/null -u admin:12345 -H 'Accept:application/vnd.go.cd.v1+json' -H 'Accept-Encoding:gzip' 'http://localhost:8153/go/api/admin/scms/foobar' | grep ETag
ETag: "97718bb418984b13bc975106ee268e81--gzip"
```
##### Expected Results
Expected to get ETag value without `--gzip` postfix: `"97718bb418984b13bc975106ee268e81"`

##### Actual Results
The value of ETag contains `--gzip` postfix: `"97718bb418984b13bc975106ee268e81--gzip"`. With this postfix in place it's not possible to update any resources, as ETag value should match server's value: 
```
Note: The update pluggable SCM API requires that you submit the If-Match header with the latest ETag value that is representative of the current pipeline config resource. 
```

grundic:
Setting deflate causes similar problem:
```
curl -s -D - -o /dev/null -u admin:12345 -H 'Accept:application/vnd.go.cd.v1+json' -H 'Accept-Encoding:deflate' 'http://localhost:8153/go/api/admin/scms/foobar' | grep ETag
ETag: "dbb1d8e2c03143e02a5f099d3e6102ae--deflate"
```

It might worth to change regex or to take care about this case somehow as well..

ketan:
@grundic 

See https://en.wikipedia.org/wiki/HTTP_ETag#Strong_and_weak_validation for why the `--gzip` and `--deflate` exist — this is both intentional and by design.

There was a bug where deflate was not handled properly in the `JettyWeakEtagMiddleware` implementation, and I've fixed that.

We've identified another issue with authorization that @varshavaradarajan will fix, she will respond to this ticket soon.

PR #2759, opened 2016-10-05
Improve handling of weak etags with deflate (#2758)

ketan:
#2758


ketan:
@varshavaradarajan - See jetty's `org.eclipse.jetty.servlets.GzipFilter` and https://en.wikipedia.org/wiki/HTTP_ETag#Strong_and_weak_validation

This one allows for deflate encoding in addition to gzip, which was already supported

ketan:
@varshavaradarajan — I'm merging this but if you see any problems, I'll revert.

Issue #2688, opened 2016-09-15
[Elastic agents] Introduce elastic agent profile API


ketan:
##### Issue Type

- Feature proposal (#2687)

See https://github.com/gocd/api.go.cd/pull/119

rajiesh:
Verified on GoCD version 16.11.0 (4185-e7acc53aac8bb5ad2a7f306ca8f3e752855d844b) working as expected

PR #2761, opened 2016-10-05
Add API to perform CRUD operations on elastic profiles (#2688)

ketan:


ketan:
Feel free to review, don't merge yet.

server/src/com/thoughtworks/go/config/update/ElasticAgentProfileCreateCommand.java
```diff
@@ -0,0 +1,34 @@
...
+import com.thoughtworks.go.server.service.GoConfigService;
+import com.thoughtworks.go.server.service.result.LocalizedOperationResult;
+
+public class ElasticAgentProfileCreateCommand extends ElasticAgentProfileCommand {
+    public ElasticAgentProfileCreateCommand(ElasticProfile elasticProfile, GoConfigService goConfigService, Username currentUser, LocalizedOperationResult result) {
+        super(elasticProfile, goConfigService, currentUser, result);
+    }
+
+    @Override
+    public void update(CruiseConfig preprocessedConfig) throws Exception {
```
varshavaradarajan: Can you please change this to ```modifiedConfig```? I know it doesn't make a difference but the preprocess method is called after update.

server/src/com/thoughtworks/go/config/update/ElasticAgentProfileDeleteCommand.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class ElasticAgentProfileDeleteCommand extends ElasticAgentProfileCommand {
+
+    public ElasticAgentProfileDeleteCommand(ElasticProfile elasticProfile, GoConfigService goConfigService, Username currentUser, LocalizedOperationResult result) {
+        super(elasticProfile, goConfigService, currentUser, result);
+    }
+
+    @Override
+    public void update(CruiseConfig preprocessedConfig) throws Exception {
+        ElasticProfile existingProfile = findExistingProfile(preprocessedConfig);
```
varshavaradarajan: Can you throw ElasticProfileNotFoundException() if existing profile is null? the remove function below can be successfully run when null is passed to it, but the user won't know that nothing has changed.

server/test/unit/com/thoughtworks/go/config/update/ElasticAgentProfileCreateCommandTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        when(goConfigService.isUserAdmin(currentUser)).thenReturn(false);
+
+        HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
+        ElasticAgentProfileCreateCommand command = new ElasticAgentProfileCreateCommand(elasticProfile, goConfigService, currentUser, result);
+        assertThat(cruiseConfig.server().getElasticConfig().getProfiles().find("foo"), nullValue());
+
+        assertThat(command.canContinue(cruiseConfig), is(false));
+        assertThat(result.toString(), containsString("UNAUTHORIZED_TO_EDIT"));
+    }
```
varshavaradarajan: Can you also add a test for when user is group admin?

server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -268,6 +268,7 @@
                 /api/admin/scms/**=ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR
                 /api/admin/plugin_info/**=ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR
                 /api/admin/agents=ROLE_SUPERVISOR
+                /api/elastic/profiles/**=ROLE_SUPERVISOR
```
varshavaradarajan: ROLE_GROUP_SUPERVISOR as well

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/elastic/profile_representer.rb
```diff
@@ -0,0 +1,48 @@
...
+      property :plugin_id
+
+      collection :properties, exec_context: :decorator, decorator: ApiV1::Config::PluginConfigurationPropertyRepresenter, class: ConfigurationProperty
+
+      def properties
+        profile.to_a
+      end
+
+      def properties=(new_properties)
+        profile.addAll(new_properties)
```
varshavaradarajan: Is there a possibility of the properties having value and encrypted_value? I see you have used ```PluginConfigurationPropertyRepresenter``` which deals with value and encrypted_value. If so, we have to add error when both are specified? Checkout ```ConfigurationPropertyBuilder``` and ```SCM#addConfigurations```

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -19,7 +19,7 @@
   mount JasmineRails::Engine => '/jasmine-specs-new', as: :jasmine_new if defined?(JasmineRails)
 
   unless defined?(CONSTANTS)
-    USER_NAME_FORMAT = GROUP_NAME_FORMAT = TEMPLATE_NAME_FORMAT = PIPELINE_NAME_FORMAT = STAGE_NAME_FORMAT = ENVIRONMENT_NAME_FORMAT = /[\w\-][\w\-.]*/
+    ELASTIC_AGENT_PROFILE_ID_FORMAT = USER_NAME_FORMAT = GROUP_NAME_FORMAT = TEMPLATE_NAME_FORMAT = PIPELINE_NAME_FORMAT = STAGE_NAME_FORMAT = ENVIRONMENT_NAME_FORMAT = /[\w\-][\w\-.]*/
```
varshavaradarajan: Don't you have to make use of NameTypeValidator on the ElasticProfile object? I see that it only makes a blank check right now.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/elastic/profiles_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Elastic
+    class ProfilesController < ::ApiV1::BaseController
+      before_action :check_for_stale_request, only: [:update]
+      before_filter :check_admin_user_and_401
```
varshavaradarajan: check for group admin too? Your entity commands do.

server/src/com/thoughtworks/go/config/update/ElasticAgentProfileCommand.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public boolean isValid(CruiseConfig preprocessedConfig) {
+        preprocessedElasticProfile = findExistingProfile(preprocessedConfig);
+        preprocessedElasticProfile.validate(null);
+
+        if (preprocessedElasticProfile.errors().isEmpty()) {
+            ElasticProfiles profiles = preprocessedConfig.server().getElasticConfig().getProfiles();
+            profiles.validate(null);
```
varshavaradarajan: What about the plugin validation? That is:
1. Whether the plugin id is valid
2. The properties that the plugin requires, whether the property should be secure etc. I know you said you are working on this. Just adding it so that we won't forget later. :)

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/elastic/profiles_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        profile_from_request = ProfileRepresenter.new(ElasticProfile.new).from_hash(params[:profile])
+
+        result = HttpLocalizedOperationResult.new
+        elastic_profile_service.update(current_user, etag_for(profile), profile_from_request, result)
+        handle_create_or_update_response(result, profile)
+      end
+
+      def create
+        result = HttpLocalizedOperationResult.new
+        profile = ProfileRepresenter.new(ElasticProfile.new).from_hash(params[:profile])
```
iamsumanth: Do we provide `:profile` param as a part of create operation, if not `params[:profile]` becomes null and can't create `ProfileRepresenter`. Using https://github.com/ketan/api.go.cd/blob/5dda72801af7eda946f974de05356a631c6e3053/source/includes/elastic_profiles/_30-create.md.erb as a reference to make a request.

PR #2775, opened 2016-10-10
Add sticky header for agents spa page

iamsumanth:
Show dropdown for build details on building agent

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agent_row_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agent_row_widget.js.msx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-define(['mithril', 'lodash'], function (m, _) {
+define(['mithril', 'lodash', 'helpers/form_helper', 'views/agents/build_details_widget'], function (m, _, f, BuildDetailsWidget) {
 
   var cacheKey = function (things) {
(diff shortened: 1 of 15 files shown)


ketan:
@iamsumanth — can you look at the jslint errors?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agent_row_widget.js.msx
```diff
@@ -34,15 +34,44 @@ define(['mithril', 'lodash'], function (m, _) {
...
+  var createBuildDetailsDropDown = function (agent, args) {
+    return (
+      <BuildDetailsWidget agent={agent}
+                          dropDownReset={args.dropDownReset}/>
+    );
+  };
+
   var AgentRowWidget = {
+    controller: function (args) {
+      var ctrl = {
```
iamsumanth: Across all the controllers, we follow the same pattern of returning a ctrl instead of assigning a property. To maintain consistency I did the same here. When I looked at the mithril documentation also most of the examples returns an object instead of assigning a property.

zabil: @ketan @iamsumanth 
Controller can also be used as a [class constructor ](http://mithril.js.org/mithril.component.html#controller-as-a-class-constructor) but it's better to return a value as the functions and variables in it are neatly scoped.

PR #2785, opened 2016-10-14
Simplify how version number is determined across the board

ketan:


server/webapp/WEB-INF/vm/shared/_header.vm
```diff
@@ -19,7 +19,7 @@
 <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8"/>
 
-    <link href="$req.getContextPath()/$concatenatedCruiseIconFilePath?#include("admin/admin_version.txt.vm")" rel="shortcut icon" />
+    <link href="$req.getContextPath()/$concatenatedCruiseIconFilePath?${currentGoCDVersion.fullVersion()}" rel="shortcut icon" />
```
jyotisingh: The query param seems redundant here, might as well get rid of it.

server/webapp/WEB-INF/vm/auth/login.vm
```diff
@@ -20,7 +20,7 @@
 <head>
   <meta http-equiv="content-type" content="text/html; charset=UTF-8"></meta>
 
-  <link href="$req.getContextPath()/$concatenatedCruiseIconFilePath?#include("admin/admin_version.txt.vm")"
+  <link href="$req.getContextPath()/$concatenatedCruiseIconFilePath?${currentGoCDVersion.fullVersion()}"
```
jyotisingh: redundant query param

server/webapp/WEB-INF/vm/shared/_header.vm
ketan: Intentional to bust cache :) Rails does this nicely for us by adding a checksum of the resource in the URL.

server/webapp/WEB-INF/vm/auth/login.vm
ketan: Intentional to bust cache :) Rails does this nicely for us by adding a checksum of the resource in the URL.

ketan:
@jyotisingh — also deleted `server_version.json` that was used from rails.

PR #2788, opened 2016-10-17
Updated windows installers

jyotisingh:
* Upgraded Tanuki wrapper binaries to version 3.5.30
* Updated license
* Both 32 and 64 bit versions of tanuki wrappers are packaged with installer to handle JVM version switch post installation
* Separate installers containing 64bit JVM and 32bit JVM for both server and agent.


installers/windows-shared/windows-installer-base.nsi
```diff
@@ -23,7 +23,7 @@ MiscButtonText "< &Back" "&Next >" "&Cancel" "&Finish"
 Name "Go $%NAME% $%VERSION%"
 
 ; The file to write
-OutFile "$%OUTDIR%\go-$%MODULE%-$%VERSION%-setup.exe"
+OutFile "$%OUTDIR%\go-$%MODULE%-$%VERSION%-with$%FLAVOUR%jre-setup.exe"
```
ketan: The general convention seems to be to use `x86` and `x64` in the binary names.

buildSrc/src/main/groovy/com/thoughtworks/go/build/WindowsPackagingTask.groovy
```diff
@@ -105,13 +105,17 @@ class WindowsPackagingTask extends DefaultTask {
...
-    "http://download.oracle.com/otn-pub/java/jdk/7u79-b15/jre-7u79-windows-i586.tar.gz"
+    (System.getenv('FLAVOUR') == "32bit") ? "http://download.oracle.com/otn-pub/java/jdk/8u102-b14/jre-8u102-windows-i586.tar.gz" : "http://download.oracle.com/otn-pub/java/jdk/8u102-b14/jre-8u102-windows-x64.tar.gz"
   }
 
   static def jreLocationSpecified() {
     System.getenv('WINDOWS_JRE_URL') != null
   }
 
+  def flavour() {
+    System.getenv('FLAVOUR') ?: "64bit"
```
ketan: I'm glad you're using the `?:` operator :)

installers/go-agent/win/config/wrapper-agent.conf
```diff
@@ -33,10 +31,8 @@ wrapper.java.mainclass=org.tanukisoftware.wrapper.WrapperSimpleApp
 
 # Java Classpath (include wrapper.jar)  Add class path elements as
 #  needed starting from 1
-wrapper.java.classpath.1=lib/wrappertest.jar
-wrapper.java.classpath.2=lib/wrapper.jar
-wrapper.java.classpath.3=agent-bootstrapper.jar
-
+wrapper.java.classpath.1=lib/wrapper.jar
+wrapper.java.classpath.2=agent-bootstrapper.jar
```
ketan: While you're at it, could you also add `wrapper.ignore_sequence_gaps=true` so as to eliminate the "reservedForFuture" in wrapper-server.conf.

This also gives us the opportunity to index `config/wrapper-properties.conf` from say `100` instead of users having to know what properties to start indexing at.

installers/go-agent/win/config/wrapper-agent.conf
```diff
@@ -137,14 +133,17 @@ wrapper.ntservice.interactive=true
...
-wrapper.license.key.1=ae3e-a990-bff1-2fda
-wrapper.license.key.2=8657-f2d4-4fa2-9688
-wrapper.license.key.3=3c6e-1981-de9b-f0d6
-wrapper.license.key.4=c5f5-b5d4-bef8-0014
+wrapper.license.host_id=080027d3e5d0
+wrapper.license.features=pro, 64bit, trial
+wrapper.license.upgrade_term.begin_date=2016-10-13
+wrapper.license.upgrade_term.end_date=2016-11-13
+wrapper.license.lease_term.begin_date=2016-10-13
+wrapper.license.lease_term.end_date=2016-11-13
```
ketan: There's an option somewhere to not have to specify the term dates in here. Can we use that?

installers/windows.gradle
```diff
@@ -48,11 +48,10 @@ task agentWindowsExe(type: WindowsPackagingTask, dependsOn: [':agent-bootstrappe
     project.copy { from('go-agent/release/start-agent.bat') into(versionedDir()) }
     project.copy { from('go-agent/release/stop-agent.bat') into(versionedDir()) }
     project.copy { from('go-agent/release/README.md') into(versionedDir()) }
-    project.copy { from('go-agent/win/cruisewrapper.exe') into(versionedDir()) }
+    project.copy { from("windows-shared/tanuki/${flavour()}/cruisewrapper.exe") into(versionedDir()) }
```
ketan: Are we building a 64bit and 32 bit installers? Or just separate 32 bit installers with 32 bit and 64 bit bundled jre?

buildSrc/src/main/groovy/com/thoughtworks/go/build/WindowsPackagingTask.groovy
```diff
@@ -105,13 +105,17 @@ class WindowsPackagingTask extends DefaultTask {
   }
 
   static def defaultJreLocation() {
-    "http://download.oracle.com/otn-pub/java/jdk/7u79-b15/jre-7u79-windows-i586.tar.gz"
+    (System.getenv('FLAVOUR') == "32bit") ? "http://download.oracle.com/otn-pub/java/jdk/8u102-b14/jre-8u102-windows-i586.tar.gz" : "http://download.oracle.com/otn-pub/java/jdk/8u102-b14/jre-8u102-windows-x64.tar.gz"
```
jyotisingh: makes sense :)

installers/go-agent/win/config/wrapper-agent.conf
jyotisingh: This is a trial license that I have put up as is so we could test out the generated installers. The purchase of the actual one is in process, which might not have the dates hardcoded.

installers/windows.gradle
jyotisingh: Building a 64 bit installer meant taking care of migration of existing installation in an automatic fashion during upgrades/server startup or by providing some script to do the same or at the very least by giving a heads up to the existing users. Not in favour of doing it automatically as the server installation folder contains flyweights and artifacts which could be huge in size and might take forever to migrate. We should give an advance notice to the users regarding the same, but it did not make sense to stop this piece of work while we figure out a migration strategy. As of now, the 64bit installers package 64bit jre and 32bit one packages the 32 bit jre. 
As far as naming the installers is concerned, since the installers are not really 64 bit, it did not make sense to call them x64 or x86.

installers/windows-shared/windows-installer-base.nsi
jyotisingh: check https://github.com/gocd/gocd/pull/2788#discussion_r83781835

PR #2801, opened 2016-10-20
Upgrade to TFS SDK 14.0.3 (#234)

ketan:


common/src/com/thoughtworks/go/domain/materials/tfs/TfsSDKCommandBuilder.java
```diff
@@ -79,6 +83,14 @@ private TfsCommand instantitateAdapter(String materialFingerPrint, UrlArgument u
 
     }
 
+    private String tfsSdkCommandTCLAdapterClassName() {
+        String className = "com.thoughtworks.go.tfssdk14.TfsSDKCommandTCLAdapter";
+        if (new SystemEnvironment().get(SystemEnvironment.TFS_SDK_10)) {
```
jyotisingh: Cannot rely on the SystemEnvironment on the agent side to get this value.

ketan:
@arvindsv — TFS is set to default to 14.0.3 :) We hope that this should be enough of a reason for users to give us feedback in case things don't work, all the while for us to provide a fallback in case things go wrong.

arvindsv:
Yay!

Issue #2470, opened 2016-07-21
Placeholder issue for GitHub PR scenario


arvindsv:
This needs me to put in more details and results from further investigation. Scenario goes something like this:

1. Configure the PR plugin
2. Open a PR for a commit.

3. 

  a. Open another PR for the same commit (or maybe merge it into another branch and push that branch and submit a PR for that).

- OR -

  a. Submit a PR for a commit
  b. Let build finish
  c. Submit a commit on the same PR
  d. Let build finish
  e. `git reset HEAD^1` to go back to previous commit, force push
  f. warning in gocd.

The plugin responds with the old commit and GoCD server 16.6.0 and beyond has some code to reject duplicate commits. The PR build does not trigger.

arvindsv:
Don't know yet if this a PR plugin issue or something in core. That's why this issue is here.

stania1:
More detailed steps to help replicate the issue:

1. Configure a pipeline with the PR plugin (Github material)
2. Create a branch, push a commit, open a PR for that branch (PR 1). You may have to close this PR right away.
3. Wait for a while. Let Go pick up that PR and commit. Only proceed when the material in the DB already has PR 1 and that commit.
4. Open another PR (PR 2). Make sure that the PR ID is _actually_ 1 above the previous one (e.g. if the old one is 2006, this one should be 2007). I think this number is essential, see **(*)** below.
5. Let Go + the Github PR pick up this commit. Actually, it won't and will pretty much refuse to pick up subsequent commits. (I wrote "pretty much" because in the early hours of noticing this problem I saw some, but very few, commits still getting picked up, before it stopped completely. Not sure why it did that yet.) What I saw happening in the UI:
```
Modification check failed for material: [url=https://github.com/bla/repo.git] [Jul-20 12:16:10]
Cannot save modification: All modifications already exist in db: [commit-SHA]
```

In `go-server.log`:

```
2016-07-19 13:10:43,795  INFO [94@MessageListener for MaterialUpdateListener] ScmMaterialUpdater:63 - [Material Update] Found '1' modifications for material ''PluggableSCMMaterial{[url=https://github.com/bla/repo.git]}'' with flyweight '...' using working directory '/home/go-server/pipelines/flyweight/...'
2016-07-19 13:10:43,796 ERROR [94@MessageListener for MaterialUpdateListener] MaterialRepository:818 - Cannot save modification:
java.lang.RuntimeException: All modifications already exist in db: [commit-SHA]
[...]

PR #2803, opened 2016-10-21
Do not raise an exception when a duplicate modification is seen (#2470)

ketan:


arvindsv:
Won't this break tests like `MaterialRepositoryIntegrationTest#shouldReportErrorIfAnAttemptIsMadeToInsertOnlyDuplicateModificationsForAGivenMaterial`?

jyotisingh:
oops...yes it will. We were lazy to run the PR pipeline so decided to merge it and see the tests fail 😛

Issue #2789, opened 2016-10-17
NPE in Agent Launcher


aschrijver:
##### Issue Type
- Bug Report
##### Summary

Agent cannot start due to NullPointerException
###### Basic environment details
- Go 16.10.0 (4131-730ff1867576754414cc632957f344d0263bc06d)
- Ubuntu 15.10 (4.2.0-41-generic)
- JDK 1.8.0_101-b13

Standard configuration (no special changes) with single agent running.
##### Steps to Reproduce
1. Start a build. Any build fails, because agent does not launch
2. Started build hangs indefinitely at the first job
##### Expected Results

Agent should start. Build should work.
##### Actual Results

NullPointerException occurs in Agent Launcher
##### Log snippets

```
2016-10-17 15:02:27,661 [main     ] ERROR cruise.agent.launcher.AgentLauncherImpl:98 - Launch encountered an unknown exception
java.lang.NullPointerException
        at com.thoughtworks.cruise.agent.launcher.AgentLauncherImpl.getPort(AgentLauncherImpl.java:136)
        at com.thoughtworks.cruise.agent.launcher.AgentLauncherImpl.getUrlGenerator(AgentLauncherImpl.java:127)
        at com.thoughtworks.cruise.agent.launcher.AgentLauncherImpl.launch(AgentLauncherImpl.java:75)
        at com.thoughtworks.go.agent.bootstrapper.AgentBootstrapper.go(AgentBootstrapper.java:72)
        at com.thoughtworks.go.agent.bootstrapper.AgentBootstrapper.main(AgentBootstrapper.java:54)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at com.simontuffs.onejar.Boot.run(Boot.java:306)
        at com.simontuffs.onejar.Boot.main(Boot.java:159)

```
##### Any other info

I was not aware that Go.CD auto-upgrades when doing apt-get on Ubuntu. I had a working installation before the upgrade :(
Also it seems I'll have to upgrade plugins individually by downloading jar-files (did that for gradle plugin)

ketan:
[...]

PR #2820, opened 2016-10-27
#2789 cleaning up agent binaries from previous version upon upgrade

jyotisingh:


arvindsv:
Don't know if it's a problem, but if `/var/lib/go-agent/*agent-launcher.jar` does not exist, then the `rm` will fail, causing whatever called it to fail. Might want to consider rm ... || true`.

ketan:
From `rm(1)`:

```
       -f, --force
              ignore nonexistent files and arguments, never prompt
```

```
$ rm -rfv /asdfasdfasfd; echo $?
0
```

arvindsv:
Ah, didn't see the `-f`. Cool.

Issue #2825, opened 2016-10-28
Email validation message is broken


arvindsv:
##### Issue Type
- Bug Report
##### Summary

All email addresses are shown as "Not valid" when setting up the "From" address for email notifications.
###### Basic environment details
- Go Version: 16.11.0
- Browser vendor and version (if relevant): Google Chrome
##### Steps to Reproduce
1. Go to Admin -> Server Configuration.
2. Scroll down to the Email Notification section.
3. Enter a valid email address in one of the email address fields and tab out of it.
##### Expected Results

It should now show a message.
##### Actual Results

Shows an "Not a valid email address" message (see image).

![image](https://cloud.githubusercontent.com/assets/172235/19808981/31971a48-9cf5-11e6-8c60-3e8fbd647f8b.png)
##### Code snippets/Screenshots and Possible Fix

Seems to have happened around the time the GET was changed to POST for the validate, [here](https://github.com/gocd/gocd/blob/3e2890503713c158232c3ba749d265c83c598451/server/webapp/WEB-INF/rails.new/app/assets/javascripts/server_configuration.js#L83-L92). It seems to be encoded from the JS side and goes as "abc%40gmail.com", but is not decoded on the server side and so, the validation is called with that value (without the @ in it).

Possible fix: We should be using a JS object instead of encoding by hand and converting it to a query string. I mean something like:

```
new Ajax.Request(url, {
    method: 'POST',
    parameters: {
        'server_configuration_form[hostName]': $('server_configuration_form_hostName').value,
        ....
    }
```

Edit: The real problem, in this specific case is [here](https://github.com/gocd/gocd/blob/3e2890503713c158232c3ba749d265c83c598451/server/webapp/WEB-INF/rails.new/app/assets/javascripts/server_configuration.js#L21-L22).
##### Any other info

The saving grace is that this is just during validation. Clicking "Save" on it will save it properly (with the @ and notifications work).

arvindsv:
Assigned to you, @iamsumanth, since you said you were looking at it.

arvindsv:
You might need to move it to the project wall.

rajiesh:
Verified on GoCD version 16.12.0 (4305-25b6f3d9eef3f9b30acd65ac9965b8776a917e9d), working as expected

PR #2847, opened 2016-11-09
Removing unnecessary url encoding of data passed via jQuery POST. (#2825)

varshavaradarajan:
@arvindsv - Please review.


ketan:
Why was this needed?

varshavaradarajan:
@ketan, sometime ago, the validation for email address, hostname and port was changed from jQuery get to post. Since, during a post call, the default encoding is already UTF-8, we didn't have to encode it again. The change from get to post was made only for the validity of hostname, port, email address. But we didn't have to do it for sendTestEmail and validateLdapSettings also. Check #2825. 

@arvindsv , I verified for everything but ldap. Will check ldap tomorrow?

arvindsv:
Ok. It should work too. Let me know if anything doesn't or you need help.

varshavaradarajan:
@arvindsv , verified the check LDAP call. Works fine.

PR #2849, opened 2016-11-10
Create a unique temp folder for every test that uses a config folder

zabil:


server/test/integration/com/thoughtworks/go/server/security/LdapAuthenticationTest.java
```diff
@@ -54,7 +54,7 @@
 public class LdapAuthenticationTest {
     @Autowired private GoConfigDao goConfigDao;
     @Autowired private LdapAuthenticationProvider ldapAuthenticationProvider;
-    private static final GoConfigFileHelper CONFIG_HELPER = new GoConfigFileHelper();
+    private static GoConfigFileHelper CONFIG_HELPER;
```
ketan: There's no real need for it to be a `static` constant anymore, is it?

PR #2850, opened 2016-11-10
Do not extract agent, agent-bootstrapper and agent-launcher jars on the server

ketan:


server/src/com/thoughtworks/go/server/GoServer.java
```diff
@@ -67,6 +67,12 @@ public void go() throws Exception {
         }
     }
 
+    private void cleanupTempFiles() {
+        FileUtils.deleteQuietly(new File("agent-bootstrapper.jar"));
+        FileUtils.deleteQuietly(new File("agent.jar"));
+        FileUtils.deleteQuietly(new File("agent-launcher.jar"));
+    }
```
ketan: These files don't need to be extracted. See the `validate()` method below.

server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
```diff
@@ -88,11 +88,6 @@ public void checkAgentStatus(HttpServletResponse response) throws IOException {
         setOtherHeaders(response);
     }
 
-    @RequestMapping(value = "/admin/latest-agent.status", method = RequestMethod.GET)
```
ketan: This was doing the same thing as `checkAgentStatus` so merged the GET and HEAD calls.

PR #2851, opened 2016-11-10
Display pluggable scm name in the header for material.

varshavaradarajan:
@maheshp - Please review when you are free.


server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/pipeline_configs/materials_config_widget.js.msx
```diff
@@ -159,7 +159,7 @@ define([
           if(Materials.isBuiltInType(args.material.type())) {
             return [args.material.type(), '-', args.material.name()].join(' ');
           }
-          if(_.isEqual(args.material.type, 'scm')) {
+          if(_.isEqual(args.material.type(), 'plugin')) {
```
ketan: use `===` instead?

maheshp:
@varshavaradarajan this would not work for Package Material once you introduce it, you may want to keep that in mind.

PR #2858, opened 2016-11-14
Change tests that use LogFixture to use specific loggers

zabil:
Make the log fixtures use the right logger instead of the root logger to
avoid tests accidentally over writing the log levels.


util/src/com/thoughtworks/go/util/LogFixture.java
```diff
@@ -25,56 +25,38 @@
...
-    public static LogFixture startListening() {
-        return startListening(Level.DEBUG);
-    }
-
-    public static LogFixture startListening(Level level) {
-        LogFixture testing = new LogFixture();
-        testing.activateOptions();
-        setLevel(level);
-        Logger.getRootLogger().addAppender(testing);
-        return testing;
```
ketan: If I understand this "fix" properly, we were basically creating a new appender for each test, which you're now auto-closing. Plus, you're creating a new logger for each test.

Would it be possible to just have `LogFixture` take in a level, and auto-generate the logger name, so it does not need to be passed in, even if just to avoid copy-paste errors?

zabil: > Would it be possible to just have LogFixture take in a level, and auto-generate the logger name, so it does not need to be passed in, even if just to avoid copy-paste errors?

This is done to make the intent of the test clear. For e.g. when the code logs to a logger
```
Logger LOG = Logger.getLogger(MyClass.class);
public void doSomething() {
 LOG.info("Executing something");
 ...
}
```

The test needs to show what logger. message and level is being tested.

```
try(LogFixture fixture = new LogFixture(MyClass.class, Level.INFO){
 assert(fixture.getMessages, "Executing something");
}
```
The above test makes it easy to understand the right logger and level.

Issue #2855, opened 2016-11-11
Agents sticky header on new agents page hides few agents


iamsumanth:
##### Issue Type

- Bug Report

##### Summary
If there are only few agents which can be accommodated in a page, then scroll shouldn't happen. But now it scrolls and the agents sticky header hides few agents.

##### Possible Fix

Allow scrolling the page only if the agents can't be accommodated.

PR #2868, opened 2016-11-17
Add extra div to make agents table start from sticky header

iamsumanth:
#2855

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agents_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agents_widget.js.msx
@@ -233,6 +233,7 @@ define([
 
       return (
         <div onclick={ctrl.hideDropDownsAndMessage.bind(ctrl)}>
+          <div class="extra-div-to-replace-sticky-header-height" style={!s.isBlank(message) ? "height:284px;" : ''}></div>
           <div class="agent-header-panel">
             <ButtonRowWidget isAnyAgentSelected={ ctrl.isAnyAgentSelected.bind(ctrl) }
                              dropdown={args.vm.dropdown}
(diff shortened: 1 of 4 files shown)


server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/agents/agents_widget.js.msx
```diff
@@ -233,6 +233,7 @@ define([
 
       return (
         <div onclick={ctrl.hideDropDownsAndMessage.bind(ctrl)}>
+          <div class="extra-div-to-replace-sticky-header-height" style={!s.isBlank(message) ? "height:284px;" : ''}></div>
```
zabil: Do not use styles, add or remove classes, 
move this method to controller.

```extra-div-to-replace-sticky-header-height``` is too specific call it ```agent-header-spacer```

File: server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_common.scss [hunk 2]
```diff
@@ -29,6 +29,9 @@ body {
 
 .main-container {
   margin-top: 50px;
+  #agents-page & {
```
ketan: File says common.scss, please move this into the agents page.

server/webapp/WEB-INF/rails.new/app/views/layouts/single_page_app.html.erb
```diff
@@ -13,7 +13,7 @@
   <%= csrf_meta_tags %>
 </head>
 
-<body data-controller-name="<%= controller_name %>" data-action-name="<%= action_name %>">
+<body data-controller-name="<%= controller_name %>" data-action-name="<%= action_name %>" id="<%= controller_name %>-page">
```
ketan: This id is not needed, the data attributes are sufficient selectors.

iamsumanth:
This fix is asserted as a part of https://github.com/gocd/gocd/pull/2809 pull request.

Can someone please close it.

PR #2880, opened 2016-11-21
Add a link to the elastic profiles page from all pages

ketan:


ketan:
@arvindsv

The page is currently explicitly marked as experimental. Both on the page, and in the nav bar.

Do we want to remove the beta label for 16.12?

screen shot 2016-11-21 at 4 19 42 pm

/cc @jyotisingh

arvindsv:
My opinion: If it works well and we're happy with it, the beta label can go away. It's not a feature that's going to go away, as well. So, it can't be beta in that sense.

ketan:
Allright, will remove the beta label and merge it in. Yay!

Issue #2842, opened 2016-11-08
Generate sha256 webserver certs instead of sha1


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Sha1 certs are being deprecated in a few months from now. GoCD must now generate sha256 certs instead.

ketan:
The self-signed web-server certificate that GoCD generates is now a SHA-256 certificate instead of a SHA-1 certificate. SHA1 certificates have been [deprecated](https://blog.qualys.com/ssllabs/2014/09/09/sha1-deprecation-what-you-need-to-know) by most browser vendors and users will start to see a warning in their browsers.

    Any new installations of GoCD will generate a SHA-256 certificate. For existing installations, users should remove the file `config/keystore` to allow GoCD to regenerate a new server certificate. Depending on your [end-to-end transport security](https://docs.go.cd/current/installation/ssl_tls/end_to_end_transport_security.html) settings you may need to configure the `-rootCertFile` argument to the GoCD agent.

PR #2887, opened 2016-11-23
Generate SHA-256 webserver certs instead of SHA1 (#2842)

ketan:


jetty9/src/com/thoughtworks/go/server/util/GoSslSocketConnector.java
```diff
@@ -52,9 +52,7 @@ public GoSslSocketConnector(Jetty9Server server, String password, SystemEnvironm
     }
 
     private Connector sslConnector(Server server) {
-        if (!keystore.exists()) {
```
zabil: Why is this removed?

common/src/com/thoughtworks/go/security/X509CertificateGenerator.java
```diff
@@ -269,7 +273,7 @@ public V3X509CertificateGenerator(Date startDate, X509Principal issuerDn, X509Pr
                                           PublicKey publicKey, BigInteger serialNumber) {
             X509V3CertificateGenerator gen = new X509V3CertificateGenerator();
             gen.reset();
-            gen.setSignatureAlgorithm("SHA1WithRSAEncryption");
+            gen.setSignatureAlgorithm("SHA256WithRSAEncryption");
```
zabil: Can we put a toggle via a system property and default to the ```SHA256WithRSAEncryption```

jetty9/src/com/thoughtworks/go/server/util/GoSslSocketConnector.java
zabil: Ok i see it moved to, ```createAndStoreX509Certificates```

common/src/com/thoughtworks/go/security/X509CertificateGenerator.java
ketan: I've moved this to a feature toggle in a newer commit.

common/src/com/thoughtworks/go/security/X509CertificateGenerator.java
```diff
@@ -64,8 +66,13 @@ public X509CertificateGenerator() {
 
     public void createAndStoreX509Certificates(File keystore, File truststore, File agentKeystore,
                                       String password, String principalDn) {
-        storeX509Certificate(keystore, password, createCertificateWithDn(principalDn));
-        storeX509Certificate(truststore, password, createAndStoreCACertificates(agentKeystore));
+        if (!keystore.exists()) {
```
ketan: I tried deleting the keystore and restarted the server. The agent continued to ping without any issues.

base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -161,6 +161,8 @@
 
     public static GoSystemProperty<Boolean> FETCH_ARTIFACT_AUTO_SUGGEST = new GoBooleanSystemProperty("go.fetch-artifact.auto-suggest", true);
     public static GoSystemProperty<String> GO_SSL_TRANSPORT_PROTOCOL_TO_BE_USED_BY_AGENT = new GoStringSystemProperty("go.ssl.agent.protocol", "TLSv1.2");
+    public static GoSystemProperty<String> GO_SSL_CERTS_ALGORITHM = new GoStringSystemProperty("go.ssl.cert.algorithm", "SHA512WITHRSA");
```
ketan: Fixed.

ketan:
The self-signed web-server certificate that GoCD generates is now a SHA-256 certificate instead of a SHA-1 certificate. SHA1 certificates have been [deprecated](https://blog.qualys.com/ssllabs/2014/09/09/sha1-deprecation-what-you-need-to-know) by most browser vendors and users will start to see a warning in their browsers.

    Any new installations of GoCD will generate a SHA-256 certificate. For existing installations, users should remove the file `config/keystore` to allow GoCD to regenerate a new server certificate. Depending on your [end-to-end transport security](https://docs.go.cd/current/installation/ssl_tls/end_to_end_transport_security.html) settings you may need to configure the `-rootCertFile` argument to the GoCD agent.

PR #2895, opened 2016-11-24
Allow selecting the elastic profile id on a job from a dropdown

ketan:


bdpiparva:
Edit and profile --> change Plugin Id --> Auto select appears in middle of page. (Just ui bug not functional bug)

maheshp:
Edit profile -> Change Plugin Id -> Save 
On save the modal is closed and we get the below message.

screen shot 2016-11-24 at 12 22 57 pm

ketan:
Perform a delete, cancel, edit, cancel
Modal stays.

maheshp:
Add profile -> Save(On save modal closes) 
There are errors in console.

screen shot 2016-11-24 at 1 22 56 pm

maheshp:
Clicking outside modal closes the modal and unsaved changes could be lost. This can be controlled by specifying the attribute ```data-close-on-click='false'```

maheshp:
Do you think we should add a warning message while editing a profile? Something like ```Editing a global copy, which might affect the jobs using it.....```.

Listing out the jobs using the profile can be done later.

PR #2901, opened 2016-11-27
Set gc aggressive to false. Set gc expire time correctly.

varshavaradarajan:


config/config-server/src/com/thoughtworks/go/service/ConfigRepository.java
```diff
@@ -394,7 +394,7 @@ public void run() throws Exception {
                     LOGGER.info("Before GC: {}", git.gc().getStatistics());
                     long date = systemEnvironment.getConfigGitGCExpireTime();
                     git.gc().setAggressive(systemEnvironment.get(SystemEnvironment.GO_CONFIG_REPO_GC_AGGRESSIVE))
-                            .setExpire(new Date(date))
+                            .setExpire(new Date(System.currentTimeMillis() - date))
```
ketan: Rename the `date` variable to something more meaningful, perhaps `expireAge` or something better?

PR #2907, opened 2016-11-30
Added a property for auto suggestion of dependencies for templates fetch artifact.

varshavaradarajan:
Computing the dependencies for auto suggest for fetch artifact is not performant for templates. The GoCD server computes the dependencies for all pipelines in the config. Since the auto suggest feature is helpful for pipelines and not templates, a feature toggle to stop computing the dependencies for templates fetch artifact for now should help with the performance.

This toggle should be removed once we improve the logic and the performance to compute the dependencies. 

@arvindsv - Please review


arvindsv:
For consisteny's sake, I think it should be a system property. I think toggles should be used to toggle real features (slightly bigger scope). Other than that, it looks fine to me.

PR #2912, opened 2016-11-30
Config save optimizations

maheshp:
* Current config save handles multiple scenarios so there are repetitive preprocess, validation,
  cloning..., which is one of reason for slow config updates on servers
  with large config.
* This commit introduces an optimized config save which handles only full config
  saves through UI, API, config upgrade(during server restart) and config loads(timer every 5 secs)
* Entity updates through UI which do not have an API still follow the old flow.
* Config save can be toggled between new and old flow using the System
  Environment 'optimize.full.config.save', this is 'true' by default.

### DIFF
--- a/config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java
+++ b/config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java
@@ -29,6 +29,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Field;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -88,7 +90,19 @@ public class MagicalGoConfigXmlWriter {
         LOGGER.debug("[Serializing Config] Finished writing config partial.");
     }
 
-    private void verifyXsdValid(Document document) throws Exception {
+    public Document documentFrom(CruiseConfig config) {
+        Document document = createEmptyCruiseConfigDocument();
+        write(config, document.getRootElement(), configCache, registry);
+        return document;
+    }
+
+    public String toString(Document document) throws IOException {
+        org.apache.commons.io.output.ByteArrayOutputStream outputStream = new org.apache.commons.io.output.ByteArrayOutputStream();
+        XmlUtils.writeXml(document, outputStream);
+        return outputStream.toString(StandardCharsets.UTF_8);
+    }
+
+    public void verifyXsdValid(Document document) throws Exception {
         ByteArrayOutputStream buffer = new ByteArrayOutputStream();
         XmlUtils.writeXml(document, buffer);
         InputStream content = toInputStream(buffer.toString());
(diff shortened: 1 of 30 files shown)


config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java
```diff
@@ -88,7 +89,19 @@ public void write(CruiseConfig configForEdit, OutputStream output, boolean skipP
...
+    public Document documentFrom(CruiseConfig config) {
+        Document document = createEmptyCruiseConfigDocument();
+        write(config, document.getRootElement(), configCache, registry);
+        return document;
+    }
+
+    public String toXml(Document document) throws IOException {
+        org.apache.commons.io.output.ByteArrayOutputStream outputStream = new org.apache.commons.io.output.ByteArrayOutputStream();
+        XmlUtils.writeXml(document, outputStream);
+        return outputStream.toString(Charset.defaultCharset());
```
ketan: Default to `utf-8` via `StandardCharsets.UTF_8` instead of `Charset.defaultCharset()`

server/src/com/thoughtworks/go/config/GoFileConfigDataSource.java
```diff
@@ -338,6 +350,66 @@ public synchronized GoConfigSaveResult writeWithLock(UpdateConfigCommand updatin
...
+    public synchronized GoConfigSaveResult writeFullConfigWithLock(FullConfigUpdateCommand updatingCommand, GoConfigHolder configHolder) {
+        try {
+            GoConfigHolder validatedConfigHolder;
+            try {
+                validatedConfigHolder = trySavingConfigWithLastKnownPartials(updatingCommand, configHolder);
+            } catch (Exception e) {
+                if (cannotSaveWithLastValidPartials()) {
+                    throw e;
+                } else {
+                    LOGGER.warn(String.format(
```
ketan: Use slf4j to avoid need for `String.format`

server/src/com/thoughtworks/go/config/GoFileConfigDataSource.java
```diff
@@ -338,6 +350,66 @@ public synchronized GoConfigSaveResult writeWithLock(UpdateConfigCommand updatin
...
+                            cachedGoPartials.lastKnownPartials().size(), cachedGoPartials.lastValidPartials().size(), e.getMessage()), e);
+
+                    validatedConfigHolder = trySavingConfigWithValidPartials(updatingCommand, configHolder);
+                }
+            }
+
+            ConfigSaveState configSaveState = shouldMergeConfig(updatingCommand, configHolder) ? ConfigSaveState.MERGED : ConfigSaveState.UPDATED;
+            return new GoConfigSaveResult(validatedConfigHolder, configSaveState);
+        } catch (ConfigFileHasChangedException e) {
+            LOGGER.warn("Configuration file could not be merged successfully after a concurrent edit: " + e.getMessage(), e);
```
ketan: Avoid string concat?

zabil:
Can you post some ab test results with old and new logic to show the performance improvements?

jyotisingh:
@maheshp - you need to rebase your PR :)

server/src/com/thoughtworks/go/config/GoFileConfigDataSource.java
maheshp: Done

server/src/com/thoughtworks/go/config/GoFileConfigDataSource.java
maheshp: Done

Issue #2913, opened 2016-12-01
Handle job run history for elastic agents on the new agents page.


zabil:
##### Issue Type
- Bug Report

##### Summary

Linking to the job run history page on elastic agents won't work as the agents are ephemeral.

![agentspage](https://cloud.githubusercontent.com/assets/54427/20779548/31c3f282-b79b-11e6-8955-da6eaf54a9d9.png)

###### Basic environment details
* Go Version: 16.12.0 (4305-25b6f3d9eef3f9b30acd65ac9965b8776a917e9d). 

##### Expected Results
~Disable the link.~ Display the right information.

##### Actual Results
Leads to
![pagenotfound](https://cloud.githubusercontent.com/assets/54427/20779686/0b966026-b79c-11e6-9920-21f8f2b79f9d.png)

PR #2917, opened 2016-12-01
Disable link to job run history for elastic agents #2913

iamsumanth:
* Remove space between agents table header and body


ketan:
I feel this is a bit pre-mature. Just because our agents go away immediately after a build does not necessarily mean that all elastic agents go away immediately.

Can we punt on this for a while?

varshavaradarajan:
@ketan if its a long running job on an elastic agent, it is shown on the agents spa with the hostname link broken, right?

ketan:
@varshavaradarajan — no reason for the link to be broken. The link should be broken only when the agent disappears from the config, not otherwise.

zabil:
Modified the issue description. Any relevant information other than page not found.

PR #2924, opened 2016-12-01
Extend Elastic Agents API with Job and Agents related calls

vlfig:
The main purpose of these changes is to allow for plugins to get hold and keep track of the global agent and job states. This is necessary for plugins that aim to be a turnkey solution for existing installations which do not require or want a distinction between pipelines for elastic agents or even between agents themselves, aside from the already existing based on environments and resources, and just want to compute metrics about the number of jobs vs number of agents.

See https://groups.google.com/forum/#!topic/go-cd-dev/32pB0lA33XU for a discussion.

The chosen way to keep track of said states is through support for one additional call from the plugin to the server (`.list-all-agents`) and one additional request from the server to the plugin (`.job-status-change`). Some refactoring was applied around the usage of `ElasticAgentMetadata` and to avoid the class casts involving `ElasticAgentRuntimeInfo`.

This PR is being created now in order to garner early feedback about its overall choices and design. It is by no means complete. Please review it lightly, for now.

Thank you


ketan:
This PR in the current state cannot be accepted for a few reasons —

* An elastic plugin must not know about all/other agents as this can allow that plugin to disable and delete agents managed by a different plugin causing side-effects, security issues and instability.

* This PR makes all agents look identical. Elastic agents are expected to **NOT** have resources, and job assignment is left entirely to the plugin. This is by design — see the section "Lax matching of agents to jobs" in [this doc](https://plugin-api.go.cd/current/elastic-agents/#process-of-handling-elastic-agents) for some background.

* Except for the `.should-assign-work` and the settings related calls, all other messages sent from the server to the plugin are via a message-queue — this is done to avoid blocking the server from doing its thing while the plugin makes expensive network calls to figure out how to best respond to the message (create/terminate instances and periodically inventory checking/audits to ensure that agents known to the server are really in-line with virtual instances somewhere in the "cloud"). Early iterations of elastic plugins had an equivalent of the `.job-status` message. This caused a lot more confusion for plugin code than was worth, in that by the time the plugin received the message, the agent **may** already be assigned work. It was deemed reasonable to let the plugin ask the server to provide the list and state of its elastic agents (`list-agents`) when the plugin desired.

If you're really looking to scale out on AWS using an ASG as you mention. You should very well be able to do that by simply polling on the [get scheduled job](https://api.go.cd/current/#get-scheduled-jobs) API and the [agents API](https://api.go.cd/current/#the-agent-object) at a frequency you desire. Looking at this PR — the agents API has exactly all the information that you desire.

This was recommended by Ashwanth on the [email thread](https://groups.google.com/d/msg/go-cd-dev/32pB0lA33XU/eBEtEABhBwAJ) and is something that a lot of users have been doing until now.

PR #2939, opened 2016-12-06
Add a link to pipeline settings in the pipeline history page.

varshavaradarajan:
@ketan - Please review.

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
@@ -37,6 +37,7 @@ import org.springframework.stereotype.Service;
 
 import java.util.*;
 
+
 /**
  * @understands providing services around a pipeline configuration
  */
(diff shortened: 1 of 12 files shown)


server/src/com/thoughtworks/go/server/service/PipelineConfigService.java
```diff
@@ -109,6 +111,15 @@ public PipelineConfig getPipelineConfig(String pipelineName) {
         return goConfigService.getConfigForEditing().getPipelineConfigByName(new CaseInsensitiveString(pipelineName));
     }
 
+    public String getPipelineSettingsLink(String pipelineName, Username username) {
```
zabil: This logic must be in the service, Ideally the view should be checking if the user is admin before rendering the link.

Look at how it's being set in the view
Check usages of https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/server/web/GoVelocityView.java#L44

Issue #2953, opened 2016-12-07
Check Repository Connection Internal API


varshavaradarajan:
##### Issue Type

- Feature proposal

##### Summary

Request: POST
Headers: 

1.  ```Accept: application/vnd.go,cd.v1+json```

2. ```Content-Type: application/json```

URL:```https://ci.example.com/go/api/admin/internal/repository_check_connection```

Data
```
{
  "plugin": "deb",
  "configuration": [
    {
      "key": "REPO_URL",
      "value": "http://repo"
    }
  ]
}
``` 

Response
```
{
  "message": "Connection OK."
}
```

rajiesh:
verified on 17.1.0 (4511-c7c379ff500a5ae70c8cd7a49f936595c6bd2d69), working as expected

PR #2954, opened 2016-12-07
Check Repository Connection API. (#2953)

varshavaradarajan:


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/materials/repository_check_connection_representer.rb
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV1
+  module Config
+    module Materials
+      class RepositoryCheckConnectionRepresenter < ApiV1::BaseRepresenter
```
ketan: Why is this object different from `PackageRepositoryRepresenter`? Ideally I should be able to submit the same JSON as https://api.go.cd/current/#the-package-repository-config-object to a different endpoint to avoid having to write a slightly different JSONizer.

PR #2970, opened 2016-12-08
Add pipeline label to notification api response

symm:
Whilst developing a notification plugin I found a need to get the `pipeline label` value but it's currently not available as it was removed in [this commit](https://github.com/gocd/gocd/commit/7e13e5203112709ff9cc806d05e13ee93ca529f7#diff-4064d61cc0533d9678ddd8cdf0aefde7L88).


arvindsv:
@symm: I did this as a part of https://github.com/gocd/gocd/pull/3030. There was some versioning infrastructure that needed to be changed, so that this can be properly versioned. Instead of being added to v1. So, this is now added to v2 and it will be available soon. A pre-merge build is running for that PR.

arvindsv:
So, I'll close this PR. Thank you!

Issue #2975, opened 2016-12-09
Remove Support for Text based API Support log


GaneshSPatil:
Currently,

go/api/support => JSON based support logs
go/api/support.txt => TEXT based support logs

Remove support for text based api support logs

PR #2976, opened 2016-12-09
Remove support for text based API support log(#2975)

GaneshSPatil:
output format of go/api/support logs changed from `text` to `json` in 16.6 (https://github.com/gocd/gocd/pull/2283)

Removing the support for TEXT based support logs (go/api/support.txt).


server/webapp/WEB-INF/rails.new/app/controllers/api/server_controller.rb
```diff
@@ -19,28 +19,12 @@ class Api::ServerController < Api::ApiController
...
-      format.any do
-        information = server_status_service.captureServerInfo(current_user, result = HttpLocalizedOperationResult.new)
-
-        if result.isSuccessful()
-          render text: information, layout: false
-        else
-          render_localized_operation_result(result)
-        end
-      end
+    information = server_status_service.asJson(current_user, result = HttpLocalizedOperationResult.new)
```
ketan: Can we not assign variables as part of a method call?

Issue #1829, opened 2016-01-23
Encrypt secrets over API


tomzo:
I propose Go API support for encrypting arbitrary string on request.

With request payload being something like

``` json
{
  "value" : "secret"
}
```

and response

``` json
{
   "encrypted_value" : "@#$FSDsd"
}
```

The first reason is that now when user commits pipeline configuration in SCM then secure variables should not be plain text. 
There is no way now to obtain encrypted values prior to saving it in configuration. (Except for a very long walkaround: Save some variable via UI, lookup how it looks in `cruise-config.xml` and copy-paste that into SCM code.)

Either API or small part in UI should make it much easier.

Byron-TW:
GoCD seems to keep his DES cipher key in `/etc/go/cipher`, which can be used to encrypt and decrypt values.
When looking [at the code](https://github.com/gocd/gocd/blob/69a39af5c0d7202b505cdf5578cebcff85643103/config/config-api/src/com/thoughtworks/go/security/GoCipher.java), it appears there is nothing more to it.

varshavaradarajan:
@tomzo , for scm and package repository plugins, if the plugin specifies a property (say, password) in the configuration as ```secure```, the APIs for scm and package repository(yet to be announced) return an encrypted_value for property password. Is this what you intended? Or do you think further work is required?

tomzo:
@varshavaradarajan sorry for delayed response.
I don't think that it helps at all. A configuration repository can also have secure environment variables, not just scms and package passwords. IMO the best way to solve this is by adding an endpoint like
```
GET /go/api/secret?secret=<plain-text>
```
which responds with encrypted text. Then this encrypted text can be placed in a file in configuration repository.

Basically the point here is to be able to get properly encrypted value from Go server without altering current configuration. I don't see how  `APIs for scm and package repository`  would help with it. When I want to encrypt a secure variable, I don't want to add scm or package repository.

varshavaradarajan:
@tomzo, thank you. Makes sense. I didn't think about the environment variables for pipelines in config repo.

dudadornelles:
Wasnt this closed by https://github.com/gocd/gocd/pull/2977 ?

ketan:
Thanks for pointing out, I'm closing this issue.

PR #2977, opened 2016-12-09
API to encrypt plain text to use in other APIs and config repositories. (#1829)

varshavaradarajan:
* Rate limit accessing the API per user per minute so as to restrict attackers from trying to get the cipher file.


server/webapp/WEB-INF/rails.new/lib/encryption_api_rate_limiter.rb
```diff
@@ -0,0 +1,31 @@
...
+      super
+    else
+      true
+    end
+  end
+
+  def client_identifier(request)
+    com.thoughtworks.go.server.util.UserHelper.getUserName().getUsername().to_s
+  end
+end
```
ketan: The expected content type is JSON, while this limiter prints `text/plain` with the message `Rate Limit Exceeded`. Consider using overriding `http_error(code, msg, headers)`

server/webapp/WEB-INF/rails.new/config/application.rb
```diff
@@ -31,6 +31,10 @@ class Application < Rails::Application
     # Replacement for "helper :all", used to make all helper methods available to controllers.
     config.action_controller.include_all_helpers = true
 
+    #Set up rate limiting
+    config.require "encryption_api_rate_limiter"
+    config.middleware.use EncryptionApiRateLimiter, {max_per_minute: 30}
```
ketan: Magic number — consider making this configurable via `SystemEnvironment`

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/encryption_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module ApiV1
+  module Admin
+    class EncryptionController < ApiV1::BaseController
+     before_action :check_admin_user_or_group_admin_user_and_401
+
+     @@go_cipher = GoCipher.new
+
+      def encrypt_value
+        render DEFAULT_FORMAT => {encrypted_value: @@go_cipher.encrypt(params[:plain_text_value])}
```
ketan: Exception handling — is there a need, can we encrypt really-really long strings, if encryption blows up, can we give a generic error instead of a 500 html page?

server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/rack-throttle-0.4.2/UNLICENSE
```diff
@@ -0,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
```
ketan: Excellent choice of license :)

PR #2980, opened 2016-12-11
Multiple fixes -

ketan:
* Show the elastic agent profiles page dropdown link for admins
* Move the image over to apiv2. This was accidentally added to apiv1.


server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/pipeline_configs/plugin_infos.js
```diff
@@ -72,8 +72,8 @@ define(['mithril', 'lodash', 'string-plus', 'helpers/mrequest', 'models/shared/i
   PluginInfos.PluginInfo.get = function (id) {
     return m.request({
       method:      'GET',
-      url:         Routes.apiv1AdminPluginInfoPath({id: id}),
-      config:      mrequest.xhrConfig.v1,
+      url:         Routes.apiv2AdminPluginInfoPath({id: id}),
```
varshavaradarajan: I suspect there will be some spec asserting on the header.

PR #2981, opened 2016-12-11
Allow dots in agent elastic agent UUID contain dots

ketan:


server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -316,7 +316,7 @@
     get 'agents', to: redirect('/go/agents')
   end
 
-  resources :agents, only: [:index], controller: "admin/agents", as: :agents
+  resources :agents, only: [:index], controller: "admin/agents", as: :agents, constraints: {id: ALLOW_DOTS}
```
varshavaradarajan: index action does not take any parameter to put a constraint on..

PR #3000, opened 2016-12-13
WIP: Package repository single page app

varshavaradarajan:
Pending:
1. Mapping of errors to specific fields if possible.
2. Check Repository and Check Package connection on click of button.
3. Some minor refactoring


ketan:
This review is by no means complete :-/ This is too large for me to review. I'll continue over the next few days.

varshavaradarajan:
About point 2 in the review summary:
> It's not entirely obvious that after creating a package repo, one has to create a package.

After creating or selecting a repository with no packages in it, we display the message ```No packages available in this repository```. Isn't that sufficient? What do you think should be improved to highlight this more?

> So a question, is there any way to show the package, before the repository — not sure how we'd drive the user to do one before the other.

Keeping in mind that a package belongs to a repository and one definitely has to create or select a repository to which a package belongs to, the UX would be somewhat confusing if we did show the package before the repository. The old package material form too forces you to select a existing repository before proceeding further.

varshavaradarajan:
Point 3 in review summary:
> The package repo and package can't quite be identified, they both look similar (dropdown) + some bunch of keys in a rectangular box, can we add some marking as to what those things are 

Would having a tooltip help? That seems to be the consistent way on the quick edit page for explaining things. Of course, we can also have icons to differentiate the two along with providing information on hover.

ketan:
> After creating or selecting a repository with no packages in it, we display the message `No packages available in this repository`. Isn't that sufficient? What do you think should be improved to highlight this more?

Hmm. Now that you mention this, I did notice the message, perhaps make it the same red color that we use for errors on form fields.


> Would having a tooltip help? That seems to be the consistent way on the quick edit page for explaining things. Of course, we can also have icons to differentiate the two along with providing information on hover.

Experiment a couple of options, send screenshots of how it'd look with both of them? Having the tooltip to explain what the repo and package are would definitely help quite a bit.

varshavaradarajan:
Closing for now. Will reopen after talk with @naveenbhaskar.

PR #3002, opened 2016-12-13
Migrate to rspec 3

ketan:


jyotisingh:
@ketan - rebase this

ketan:
> @ketan - rebase this

I'm giving up on this — this requires a herculean effort.

Most of the migration is using an automated tool (transpec). But a lot of test fail (in most cases — bad tests and mocks/stubs for methods that don't exist *gasp*).

@gocd/committers Do you think we can sign up everyone to fix these tests. This should take everyone a couple of hours, but would mean us fixing a lot of tests that give a false positive, and getting rid of some tech-debt along the way.

arvindsv:
I tried something similar, to get rid of our [partial mocks](https://github.com/arvindsv/gocd/commit/f9c045980d1daa7f60a8ffd0387c3343801dff54#diff-c4335ab7e7915be496aa0ce651dc2e92R63) and have pure rspec tests. It's too much work for one person. I'm happy to help with anything.

PR #3003, opened 2016-12-13
Add timeline entries to the accumulator before adding to the timeline for rollback.

varshavaradarajan:


server/test/unit/com/thoughtworks/go/server/persistence/PipelineRepositoryTest.java
```diff
@@ -89,15 +116,89 @@ public void shouldCachePipelineSelectionForGivenId() throws Exception {
...
+        assertNotNull(timelineEntry1);
+        assertNotNull(timelineEntry2);
+        assertThat(tempEntries.size(), is(2));
+        assertThat(tempEntries, containsInAnyOrder(timelineEntry1, timelineEntry2));
+    }
+
+    @Test
+    public void shouldNotUpdateTimelineEntriesAndEntriesForRollbackUponFailureDuringRetrieval() throws Exception {
+        Object[] pipelineRow1 = {"p1", new BigInteger("1"), new BigInteger("1"), new Date(), "fingerprint", 1.0, "r1", null, new BigInteger("1"), new BigInteger("1")};
+        Object[] pipelineRow2 = {"p1", "2abcd", new BigInteger("2"), new Date(), "fingerprint", 2.0, "r2", null, new BigInteger("1"), new BigInteger("1")};
```
arvindsv: Can you change "2abcd" to something which describes what's happening? Something like "WILL-CAUSE-FAILURE-DURING-RETRIEVAL"?

arvindsv:
Just for the record, the issue here was that when called through [here](https://github.com/gocd/gocd/blob/c749c47a12df283c2e333b9cc918e161aa6c1803/server/src/com/thoughtworks/go/server/domain/PipelineTimeline.java#L135) and there is a failure somewhere [here](https://github.com/gocd/gocd/blob/c749c47a12df283c2e333b9cc918e161aa6c1803/server/src/com/thoughtworks/go/server/persistence/PipelineRepository.java#L91), [the rollback](https://github.com/gocd/gocd/blob/c749c47a12df283c2e333b9cc918e161aa6c1803/server/src/com/thoughtworks/go/server/domain/PipelineTimeline.java#L114) wouldn't happen properly. DB rollback happens, but the [in-memory entries](https://github.com/gocd/gocd/blob/c749c47a12df283c2e333b9cc918e161aa6c1803/server/src/com/thoughtworks/go/server/domain/PipelineTimeline.java#L129-L130) wouldn't get rolled back - because the exception would mean that `tempEntries[0]` is null. Now, there's an accumulating collection passed in, which is filled up as the entries are added. So, even if an exception is thrown somewhere, the partial entries are accumulated, so that they can be rolled back. The tests prove that.

arvindsv:
An example of the exception you'd see:

```
Caused by: java.lang.RuntimeException: Calculated natural ordering 37.0 is not the same as the existing naturalOrder 38.0, for pipeline PIPELINE-1, with id 12345
at com.thoughtworks.go.util.ExceptionUtils.bomb(ExceptionUtils.java:28)
at com.thoughtworks.go.domain.PipelineTimelineEntry.updateNaturalOrder(PipelineTimelineEntry.java:180)
at com.thoughtworks.go.server.domain.PipelineTimeline.add(PipelineTimeline.java:100)
at com.thoughtworks.go.server.persistence.PipelineRepository$1.populateFrom(PipelineRepository.java:176)
at com.thoughtworks.go.server.persistence.PipelineRepository$1.doInHibernate(PipelineRepository.java:90)
at org.springframework.orm.hibernate3.HibernateTemplate.doExecute(HibernateTemplate.java:406)
at org.springframework.orm.hibernate3.HibernateTemplate.execute(HibernateTemplate.java:339)
at com.thoughtworks.go.server.persistence.PipelineRepository.updatePipelineTimeline(PipelineRepository.java:75)
at com.thoughtworks.go.server.domain.PipelineTimeline.updateTimelineOnInit(PipelineTimeline.java:186)
at com.thoughtworks.go.server.initializers.ApplicationInitializer.onApplicationEvent(ApplicationInitializer.java:129)
... 37 more
2016-10-27 14:54:37,006 WARN [main] WebAppContext:514 - Failed startup of context o.e.j.w.WebAppContext@299a06ac{/go,file:/var/lib/go-ser
ver/work/jetty-0.0.0.0-8153-cruise.war-_go-any-/webapp/,STARTING}{cruise.war}
java.lang.RuntimeException: java.lang.RuntimeException: Calculated natural ordering 37.0 is not the same as the existing naturalOrder 38.0
```

PR #3006, opened 2016-12-14
Adding CPU time in api/support for the threads.

jyotisingh:
* CPU time for all threads
* CPU consumed by the current set of operations for daemon threads used for MDU/Scheduling etc
* AllocatedMemory for each thread
* Capturing of allocated-memory, thread-wait times and thread-blocked times are not enabled by default. Enabling capturing of these stats based on a diagnostics flag

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/support/ResourceMonitoring.java
+++ b/server/src/com/thoughtworks/go/server/service/support/ResourceMonitoring.java
@@ -45,7 +45,7 @@ public class ResourceMonitoring {
             if (threadMXBean.isThreadCpuTimeSupported()) {
                 threadMXBean.setThreadCpuTimeEnabled(true);
             }
-            if (((com.sun.management.ThreadMXBean) threadMXBean).isThreadAllocatedMemorySupported()) {
+            if ((threadMXBean instanceof com.sun.management.ThreadMXBean) && ((com.sun.management.ThreadMXBean) threadMXBean).isThreadAllocatedMemorySupported()) {
                 ((com.sun.management.ThreadMXBean) threadMXBean).setThreadAllocatedMemoryEnabled(true);
             }
         }
--- a/server/src/com/thoughtworks/go/server/service/support/ResourceMonitoring.java
+++ b/server/src/com/thoughtworks/go/server/service/support/ResourceMonitoring.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2016 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


server/src/com/thoughtworks/go/server/service/support/ThreadInformationProvider.java
```diff
@@ -68,15 +77,21 @@ public double priority() {
...
         for (ThreadInfo threadInfo : threadInfos) {
             LinkedHashMap<String, Object> threadStackTrace = new LinkedHashMap<>();
             threadStackTrace.put("Id", threadInfo.getThreadId());
             threadStackTrace.put("Name", threadInfo.getThreadName());
             threadStackTrace.put("State", threadInfo.getThreadState());
-
+            threadStackTrace.put("UserTime(nanoseconds)", threadMXBean.getThreadUserTime(threadInfo.getThreadId()));
+            threadStackTrace.put("CPUTime(nanoseconds)", threadMXBean.getThreadCpuTime(threadInfo.getThreadId()));
+            threadStackTrace.put("DaemonThreadInfo", daemonThreadStatsCollector.statsFor(threadInfo.getThreadId()));
+            if (threadMXBean instanceof com.sun.management.ThreadMXBean) {
```
ketan: Casting to this seems dangerous? Can we instead duck-type?

```
method = org.springframework.util.ReflectionUtils.findMethod(threadMXBean.getClass(), "getThreadAllocatedBytes")
if (method != null) {
  bytes = org.springframework.util.ReflectionUtils.invokeMethod(method, threadMXBean, threadInfo,getThreadId());
  threadStackTrace.put("AllocatedMemory(Bytes)", bytes);
}
```

server/src/com/thoughtworks/go/server/service/support/ResourceMonitoring.java
```diff
@@ -0,0 +1,53 @@
...
+        if (systemEnvironment.get(SystemEnvironment.GO_DIAGNOSTICS_MODE)) {
+            ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
+            if (threadMXBean.isThreadContentionMonitoringSupported()) {
+                threadMXBean.setThreadContentionMonitoringEnabled(true);
+            }
+            if (threadMXBean.isThreadCpuTimeSupported()) {
+                threadMXBean.setThreadCpuTimeEnabled(true);
+            }
+            if (((com.sun.management.ThreadMXBean) threadMXBean).isThreadAllocatedMemorySupported()) {
+                ((com.sun.management.ThreadMXBean) threadMXBean).setThreadAllocatedMemoryEnabled(true);
```
ketan: Same comment as above.

server/src/com/thoughtworks/go/server/service/support/ThreadInformationProvider.java
jyotisingh: We check for the type before casting so it is safe. But yes, we missed the type check in `enableIfDiagnosticsModeIsEnabled`. Will fix that.

ketan: Checking for type on a JVM that does not have the required class will throw a CNFE when loading `ThreadInformationProvider` class, which is why I suggested the duck-type approach.

PR #3015, opened 2016-12-16
Remove old agents page

iamsumanth:
* Environments page uses old agents table to edit agents, need to find alternate.


server/webapp/WEB-INF/rails.new/app/views/layouts/agent_detail.html.erb
```diff
@@ -19,7 +19,6 @@
       <div class="page_header">
         <div class="row">
           <ul class="entity_title">
-            <!--<li class="name"><%= link_to l.string("AGENTS_POPUP_TITLE"), url_for(:controller => "agents", :action => "index") -%> </li>-->
```
jyotisingh: why did you get rid of the link? wouldn't pointing to the new page suffice?

server/src/com/thoughtworks/go/server/service/AgentService.java
```diff
@@ -114,8 +114,8 @@ public void sync(Agents agents) {
     }
 
     @Deprecated // used by old code, stop using AgentViewModel
-    public AgentsViewModel agents() {
-        return toAgentViewModels(agentInstances.allAgents());
+    public AgentInstances agents() {
```
jyotisingh: No one seems to be using this method apart from the old controller which is being deleted as a part of this PR. Why not just get rid of the method?

server/test/integration/com/thoughtworks/go/server/service/AgentServiceIntegrationTest.java
```diff
@@ -255,22 +254,6 @@ public void shouldRemoveResourcesFromMultipleAgents() {
     }
 
     @Test
-    public void shouldFindAnAgentForAGivenUUID() {
```
jyotisingh: `findAgentAndRefreshStatus` still exists in AgentService. Why delete the test?

server/webapp/WEB-INF/rails.new/app/views/layouts/agent_detail.html.erb
iamsumanth: Link is already commented and we are not using it anymore to point to agents page. Thats why I removed it.

PR #3017, opened 2016-12-16
Remove freemarker, we do not use it anymore

ketan:


arvindsv:
Looks good to me. Here are some comments [from gitter](https://gitter.im/gocd/gocd?at=58629af1c02c1a3959ba9371):

@arvindsv 11:46: Is the erb renderer gone because the old config UI is gone?

@arvindsv 11:46: I remember that it was around, for task plugins.

@arvindsv 11:53: I see. It looks like you're using the file directly from Rails. So, the ERB renderer is not required. That makes sense.

@arvindsv 11:53: Once the old config UI is gone (when is that?), we can throw all of this away, I guess.

@arvindsv 11:53: I'm surprised that the whole factoryContext vs. localContext etc. did not cause any tests to fail. Not sure if there's any clashing names which could cause an issue somewhere. But, happy to take that risk.

arvindsv:
The part I haven't considered is: "What else can be removed as a part of this?". I'm fine with the part that is removed. We can remove more later, when we remove the whole old config UI part.

PR #3019, opened 2016-12-19
Templates spa

GaneshSPatil:


server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/template_configs/template.js
```diff
@@ -0,0 +1,154 @@
...
+      this.authorization = m.prop(s.defaultToIfBlank(data.authorization, new Template.Authorization({})));
+      this.stages        = s.collectionToJSON(m.prop(s.defaultToIfBlank(data.stages, new Stages())));
+
+      this.validatePresenceOf('name');
+      this.validateAssociated('stages');
+
+      this.update = function (etag, extract) {
+        var self = this;
+
+        var config = function (xhr) {
```
iamsumanth: All you wanted is to add etag to the xhr, you will get xhr object in config itself. Please look at https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/elastic_profiles/elastic_profiles.js#L72

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/template_configs/templates.js
```diff
@@ -0,0 +1,77 @@
...
+          config: mrequest.xhrConfig.v4
+        });
+      };
+    };
+
+    Templates.Template.fromJSON = function (data) {
+      return new Templates.Template({
+        name:      data.name,
+        url:       data._links.self.href,
+        pipelines: data._embedded.pipelines
```
iamsumanth: Can we pass the object which holds name and url from here instead of constructing it in Templates.Template

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/single_page_apps/new_templates.js
```diff
@@ -0,0 +1,46 @@
...
+  $(function () {
+    var templatesConfigElem = $('#templates');
+    var allUserNames        = JSON.parse(templatesConfigElem.attr('data-user-names'));
+    var allRoleNames        = JSON.parse(templatesConfigElem.attr('data-role-names'));
+    var isUserAdmin         = JSON.parse(templatesConfigElem.attr('is-current-user-an-admin'));
+
+    Users.initializeWith(allUserNames);
+    Roles.initializeWith(allRoleNames);
+
+    m.sync([ElasticProfiles.all()]).then(function (args) {
```
iamsumanth: Can it be ```ElasticProfiles.all().then(function (args){ ``` when we want to depend only on one response.

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_row_widget.js.msx
```diff
@@ -0,0 +1,87 @@
...
+          pipelinesList = (
+            <dl class="key-value-pair">
+              {_.map(args.template.pipelines(), function(pipeline) {
+                return [
+                  (<dt>
+                    <a href={Routes.editAdminPipelineConfigPath({pipeline_name: pipeline.name()})}>{pipeline.name()}</a>
+                  </dt>),
+                ];
+              })}
+            </dl>
```
iamsumanth: Can you please check with @naveenbhaskar if using ```<dt>``` alone without ```<dd>``` is a good way to write html. Generally ```<dl>``` holds both ```<dt>``` and ```<dd>```.
Does it need to be a class with "key-value-pair" because I see html doesn't represent any key value pair.

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_row_widget.js.msx
```diff
@@ -0,0 +1,87 @@
...
+          </div>
+        )
+      };
+
+      return (
+        <div class="template-link">
+          <div id="template-link-header" onclick={ctrl.toggleHide.bind(ctrl)}>
+            {args.template.name()}
+          </div>
+          {pipelines()}
```
iamsumanth: Can it be a variable as well instead of function ?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/templates_config_widget.js.msx
```diff
@@ -0,0 +1,148 @@
...
+          self.reloadPage();
+          self.message({type: 'success', message: response.message});
+        };
+
+        var onFailure = function (response) {
+          modal.destroy();
+          self.message({type: 'alert', message: response.message});
+        };
+
+        m.redraw(true);
```
iamsumanth: is there any reason behind this force redraw ?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_config_widget.js.msx
```diff
@@ -0,0 +1,148 @@
...
+        <form class='template-main'>
+          {header()}
+            <div class='template-body'>
+                <f.row>
+                    {errors}
+                </f.row>
+              {auth}
+          <f.row>
+            <f.column size={12}>
+              <PipelineFlowWidget class="stages"
```
iamsumanth: I don't see any use of ```class="stages"``` in ```PipelineFlowWidget```. Can we not pass it if we are not using ?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_row_widget.js.msx
GaneshSPatil: it contains a lot of logic, for all the logic related to the pipelines list at one place, its better to keep it as a function than a variable

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/approval_widget.js.msx
```diff
@@ -0,0 +1,85 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+define([
+    'mithril', 'string-plus', 'jquery', 'helpers/form_helper', 'helpers/pipeline_configs/tooltips',
+    'helpers/mithril_component_mixins', 'models/shared/users', 'models/shared/roles', 'jquery.textcomplete'
+], function (m, s, $, f, tt, ComponentMixins, Users, Roles) {
+
+  var autoComplete = function (values, model, attrName) {
```
iamsumanth: autoComplete method is already implemented as a part of ```views/pipeline_configs/stages_config_widget``` if possible can you please reuse the same method ?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/new_template_config_widget.js.msx
```diff
@@ -0,0 +1,138 @@
...
+        <div class="header-panel">
+          <header class="page-header">
+            <f.row>
+              <f.column size={10}>
+                <h1>
+                  Add Template
+                </h1>
+              </f.column>
+              <f.column size={1}>
+                <f.button onclick={ctrl.onBack.bind(ctrl)}>
```
iamsumanth: Does ```ctrl.onBack.bind(ctrl)``` need to bing with controller ?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_row_widget.js.msx
iamsumanth: Does refactoring pipelines method to 

```
        var pipelinesList = (
          <f.info>No pipelines are associated with this template</f.info>
        );
      
        if(args.template.pipelines().length !== 0) {
          pipelinesList = (
            <dl class="key-value-pair">
              {_.map(args.template.pipelines(), function(pipeline) {
                return [
                  (<dt>
                    <a href={Routes.editAdminPipelineConfigPath({pipeline_name: pipeline.name()})}>{pipeline.name()}</a>
                  </dt>),
                ];
              })}
            </dl>
          );
        }
```

and adding 
```
          <div class={'template-used-by-pipelines ' + ctrl.showState()}>
            {pipelinesList}
          </div>
```
directly in the view seems good ?

It doesn't matter much but just a suggestion. Finally its your call :)

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/views/template_configs/template_config_widget.js.msx
GaneshSPatil: used in tests

varshavaradarajan:
Authorization for the templates SPA should be based on https://github.com/gocd/gocd/issues/3081 and particularly https://github.com/gocd/gocd/issues/3081#issuecomment-277275692

- Index page - can be accessed by users who have view access to at least one template. 
- Index page should show the list of pipelines associated with each viewable template. Based on the `can_edit` attribute on each pipeline, pipeline names should be links to the pipeline edit page.
- Editing of permissions for a template should be enabled only for admins.
- The `create new template` button should be enabled for admins and pipeline group admins(who will become template admins of the template they create) 
- Deleting of template should be allowed for admins and template admins
- Template view users cannot create or delete a template.

GaneshSPatil:
Need to decide on the authorization bit.
Will continue once we have a way to refer to template authorization.

PR #3029, opened 2016-12-22
Move all agent-server communication over to jetty-client from httpclient

ketan:


ketan:
Closing PR — this was proving to be a bit too flaky.

PR #3031, opened 2016-12-23
Encryption API: Use a representer to render the response along with links.

varshavaradarajan:


server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -265,6 +265,7 @@
                 /api/admin/internal/repository_check_connection=ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR
                 /api/admin/pipelines=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /api/admin/pipelines/*=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
+                /api/admin/encrypt=ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR
```
ketan: Would template admins need access to this?

Issue #2989, opened 2016-12-12
Template admin role validation and providing template admin authorization to users within a role. 


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

```
<security>
      ...
      <roles>
        <role name="template-admin">
          <users>
            <user>jez</user>
          </users>
        </role>
      </roles>
...
  <templates>
    <pipeline name="template1">
      <authorization>
        <admins>
          <role>non-existent-role</role>
          <role>template-admin</role>
          <user>someUser</user>
        </admins>
      </authorization>
```
1. A non existent role configured at the template level is not validated.
2. When logged in as user ```jez```,   jez (who is a part of role ```template-admin```) is not a template admin.

###### Basic environment details

* Go Version: `16.12`

##### Steps to Reproduce

1. Setup configurations as shown above
2. Save the configuration 
3. Log in as user 'jez' (defined in the password properties file or LDAP)

##### Expected Results

1. Saving the configuration with non-existent-role should not be allowed.
2. User 'jez' should be a template admin 

##### Actual Results

1. Role configured for template admins should be validated.
2. Users with an existing role, which is configured to be template admin, should be giving the right permissions.

##### Possible Fix

1. For template level authorization, role should be validated. 
2. A role configured for template admin using a parameter should be ignored. (or not allowed)
3. A user configured for template admin using a parameter should be ignored. (or not allowed)
4. For existing configurations, since users within a role are not given the right authorization, having a role currently is meaningless. Specifying a non-existent role for a template admin in this case is a rare instance. Therefore, showing a validation error is better than having a config migration to remove such spurious entries.
5. Users within a role should be given the appropriate permission.

rajiesh:
verified on 17.1.0 (4511-c7c379ff500a5ae70c8cd7a49f936595c6bd2d69), working as expected

PR #3035, opened 2016-12-23
Provide template admin authorization to users within a role. (#2989)

varshavaradarajan:
* Validate whether role configured for template admin exists. If specified as a param, consider the param as a string and validate whether the string exists as role in the security config.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
@@ -24,6 +24,7 @@ import com.thoughtworks.go.config.preprocessor.SkipParameterResolution;
 import com.thoughtworks.go.config.validation.NameTypeValidator;
 import com.thoughtworks.go.domain.BaseCollection;
 import com.thoughtworks.go.domain.ConfigErrors;
+import com.thoughtworks.go.domain.Task;
 import com.thoughtworks.go.domain.config.Admin;
 
 import java.util.HashMap;
@@ -66,6 +67,47 @@ public class PipelineTemplateConfig extends BaseCollection<StageConfig> implemen
 
     public CaseInsensitiveString name() {
         return name;
+    }
+
+    public void validateTree(ValidationContext validationContext, CruiseConfig preprocessedConfig, boolean isTemplateBeingCreated) {
+        validate(validationContext);
+        if (!isTemplateBeingCreated) {
+            validateParamsAndFetchTasks(preprocessedConfig);
+        }
+    }
+
+    private void validateParamsAndFetchTasks(CruiseConfig preprocessedConfig) {
+        List<CaseInsensitiveString> pipelineNames = preprocessedConfig.pipelinesAssociatedWithTemplate(this.name());
+        ParamsConfig paramsConfig = this.referredParams();
+        for (CaseInsensitiveString pipelineName : pipelineNames) {
+            PipelineConfig pipelineConfig = preprocessedConfig.getPipelineConfigByName(pipelineName);
+            validateParams(pipelineConfig, paramsConfig);
+            validateFetchTasks(preprocessedConfig, pipelineConfig);
+        }
+    }
+
+    private void validateFetchTasks(CruiseConfig preprocessedConfig, PipelineConfig pipelineConfig) {
+        PipelineConfigSaveValidationContext contextForStages = PipelineConfigSaveValidationContext.forChain(false, "", preprocessedConfig, pipelineConfig);
+        for (StageConfig stageConfig : pipelineConfig.getStages()) {
+            PipelineConfigSaveValidationContext contextForJobs = contextForStages.withParent(stageConfig);
+            for (JobConfig jobConfig : stageConfig.getJobs()) {
+                PipelineConfigSaveValidationContext contextForTasks = contextForJobs.withParent(jobConfig);
+                for (Task task : jobConfig.getTasks()) {
+                    if (task instanceof FetchTask) {
+                        task.validate(contextForTasks);
+                        this.errors().addAll(task.errors());
+                    }
+                }
+            }
+        }
+    }
+
+    private void validateParams(PipelineConfig pipelineConfig, ParamsConfig paramsConfig) {
+        for (ParamConfig paramConfig : paramsConfig) {
+            if (!pipelineConfig.getParams().hasParamNamed(paramConfig.getName())) {
+                this.addError("params", String.format("The param '%s' is not defined in pipeline '%s'", paramConfig.getName(), pipelineConfig.getName()));
+            }
+        }
     }
 
     public void validate(ValidationContext validationContext) {
--- a/config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 ThoughtWorks, Inc.
+ * Copyright 2017 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -83,7 +83,7 @@ public class PipelineTemplateConfig extends BaseCollection<StageConfig> implemen
             PipelineConfig pipelineConfig = preprocessedConfig.getPipelineConfigByName(pipelineName);
             PipelineConfigSaveValidationContext contextForStages = PipelineConfigSaveValidationContext.forChain(false, "", preprocessedConfig, pipelineConfig);
             validateParams(pipelineConfig, paramsConfig);
-            validateFetchTasks(preprocessedConfig, pipelineConfig, contextForStages);
+            validateTaskAndElasticProfileId(pipelineConfig, contextForStages);
             validateDependencyMaterialAndFetchArtifactForDownstreams(pipelineConfig, contextForStages);
         }
     }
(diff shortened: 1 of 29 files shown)


config/config-api/src/com/thoughtworks/go/config/AbstractValidationContext.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+package com.thoughtworks.go.config;
+
+import com.thoughtworks.go.config.materials.MaterialConfigs;
+import com.thoughtworks.go.config.remote.ConfigReposConfig;
+import com.thoughtworks.go.domain.packagerepository.PackageRepository;
+import com.thoughtworks.go.domain.scm.SCM;
+
+public abstract class AbstractValidationContext implements ValidationContext{
```
arvindsv: Maybe call this DelegatingValidationContext?

config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
```diff
@@ -70,9 +71,21 @@ public CaseInsensitiveString name() {
...
+        validateTemplateAuth(new AbstractValidationContext(validationContext) {
+            @Override
+            public boolean shouldNotCheckRole() {
+                return false;
+            }
+        });
         validateStageConfig(validationContext);
     }
 
+    private void validateTemplateAuth(AbstractValidationContext abstractValidationContext) {
```
arvindsv: Rename this param to `validationContextWhichChecksRole`?

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
```diff
@@ -114,7 +115,41 @@ public void shouldReturnFalseIfUserCannotViewAndEditAtLeastOneTemplate() throws
...
         TemplatesConfig templates = new TemplatesConfig(templateList.toArray(new PipelineTemplateConfig[0]));
 
-        assertThat(templates.canViewAndEditTemplate(nonTemplateAdmin), is(false));
+        assertThat(templates.canViewAndEditTemplate(nonTemplateAdmin, null), is(false));
+    }
+
+    @Test
+    public void shouldReturnTrueIfUserWithinARoleCanViewAndEditTemplates() {
+        ArrayList<PipelineTemplateConfig> templateList = new ArrayList<PipelineTemplateConfig>();
+        for (int i = 0; i < 100; i++) {
```
arvindsv: What's the point of adding a 100 templates? Why not just 1?

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
```diff
@@ -114,7 +115,41 @@ public void shouldReturnFalseIfUserCannotViewAndEditAtLeastOneTemplate() throws
...
+        }
+        CaseInsensitiveString templateAdmin = new CaseInsensitiveString("template-admin");
+        Role securityConfigRole = new Role(new CaseInsensitiveString("role1"), new RoleUser(templateAdmin));
+        AdminRole templateAdminRole = new AdminRole(securityConfigRole);
+        templateList.add(PipelineTemplateConfigMother.createTemplate("templateName", new Authorization(new AdminsConfig(templateAdminRole)), StageConfigMother.manualStage("stage-name")));
+        TemplatesConfig templates = new TemplatesConfig(templateList.toArray(new PipelineTemplateConfig[0]));
+        List<Role> roles = new ArrayList<>();
+        roles.add(securityConfigRole);
+
+        assertThat(templates.canViewAndEditTemplate(templateAdmin, roles), is(true));
```
arvindsv: I found it a little hard to understand this test. 101 templates are added. It seems like only one (the last one) has some authorization added to to. The stage name param seems irrelevant and could be renamed to "dont-care" or something to show that.

Then, when you ask whether either of `templateAdmin` or `roles` can access the template, it says "yes". To make it more clear, you can do this, I feel:

```
assertThat(templates.canViewAndEditTemplate(someRandomNormalUser, roleWithTemplateAdmin), is(true));
```

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
```diff
@@ -114,7 +115,41 @@ public void shouldReturnFalseIfUserCannotViewAndEditAtLeastOneTemplate() throws
...
+        }
+        CaseInsensitiveString templateAdmin = new CaseInsensitiveString("template-admin");
+        Role securityConfigRole = new Role(new CaseInsensitiveString("role1"), new RoleUser(templateAdmin));
+        AdminRole templateAdminRole = new AdminRole(securityConfigRole);
+        templateList.add(PipelineTemplateConfigMother.createTemplate("templateName", new Authorization(new AdminsConfig(new AdminUser(new CaseInsensitiveString("random-user")))), StageConfigMother.manualStage("stage-name")));
+        TemplatesConfig templates = new TemplatesConfig(templateList.toArray(new PipelineTemplateConfig[0]));
+        List<Role> roles = new ArrayList<>();
+        roles.add(securityConfigRole);
+
+        assertThat(templates.canViewAndEditTemplate(templateAdmin, roles), is(false));
```
arvindsv: Similar to above.

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
```diff
@@ -135,7 +170,37 @@ public void shouldReturnFalseIfUserCannotEditTemplate() {
...
+        Role securityConfigRole = new Role(new CaseInsensitiveString("role1"), new RoleUser(templateAdmin));
+        AdminRole templateAdminRole = new AdminRole(securityConfigRole);
+        List<Role> roles = new ArrayList<>();
+        roles.add(securityConfigRole);
+        String templateName = "template1";
+        PipelineTemplateConfig template = PipelineTemplateConfigMother.createTemplate(templateName, new Authorization(new AdminsConfig(templateAdminRole)),
+                StageConfigMother.manualStage("stage-name"));
+        TemplatesConfig templates = new TemplatesConfig(template);
+
+        assertThat(templates.canUserEditTemplate(templateName, templateAdmin, roles), is(true));
```
arvindsv: Similar to above. Maybe pull out methods like `roleWithUser("template-admin")` so that the whole bit around wrapping the `CaseInsensitiveString` in a `Role` and then in an `AdminRole` which is not very relevant to this test can go away and make this test easier to read.

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
```diff
@@ -135,7 +170,37 @@ public void shouldReturnFalseIfUserCannotEditTemplate() {
...
+        Role securityConfigRole = new Role(new CaseInsensitiveString("role1"), new RoleUser(templateAdmin));
+        AdminRole templateAdminRole = new AdminRole(securityConfigRole);
+        List<Role> roles = new ArrayList<>();
+        roles.add(securityConfigRole);
+        String templateName = "template1";
+        PipelineTemplateConfig template = PipelineTemplateConfigMother.createTemplate(templateName, new Authorization(new AdminsConfig(new AdminRole(new CaseInsensitiveString("another-role")))),
+                StageConfigMother.manualStage("stage-name"));
+        TemplatesConfig templates = new TemplatesConfig(template);
+
+        assertThat(templates.canUserEditTemplate(templateName, templateAdmin, roles), is(false));
```
arvindsv: Same as above.

server/src/com/thoughtworks/go/config/update/TemplateConfigCommand.java
```diff
@@ -46,7 +46,7 @@ public TemplateConfigCommand(PipelineTemplateConfig templateConfig, LocalizedOpe
     public boolean isValid(CruiseConfig preprocessedConfig) {
         TemplatesConfig templatesConfig = preprocessedConfig.getTemplates();
         preprocessedTemplateConfig = findAddedTemplate(preprocessedConfig);
-        preprocessedTemplateConfig.validate(ConfigSaveValidationContext.forChain(preprocessedConfig));
+        preprocessedTemplateConfig.validate(ConfigSaveValidationContext.forChain(templatesConfig));
```
arvindsv: I know we talked about this. Since there is a much smaller part of the config being sent here, are you sure there is no other part of the config which was being validated here that isn't any more? Are you depending on the preprocessed config being validated later somewhere?

config/config-api/test/com/thoughtworks/go/config/TemplatesConfigTest.java
varshavaradarajan: The ``canViewAndEditTemplate``` method takes two arguments: user and the roles to which the user belongs to (not either user or roles). Then, the validation proceeds to check whether any one of the roles to which the user belongs to is specified as template admin.

arvindsv: Ah, ok. That makes sense.

server/src/com/thoughtworks/go/config/update/TemplateConfigCommand.java
varshavaradarajan: The preprocessed config is not being validated elsewhere later. We skip it because we only want to validate a specific entity(template) and related entities (eg. templates). 
> Since there is a much smaller part of the config being sent here, are you sure there is no other part of the config which was being validated here that isn't any more?

The template config validation only deals with its own object and nothing else as far as I know.

jyotisingh: Not really introduced as a part of this PR. But looks like we do not seem to be validating things like fetch artifacts task for the template config api. This could lead to invalid config being saved. Not sure what kind of issues would this lead to.

varshavaradarajan: Because of https://github.com/gocd/gocd/blob/master/config/config-api/src/com/thoughtworks/go/config/FetchTask.java#L234 ?
We do seem to be validating the task once the template is used in a pipeline and that's what needs to be done because a fetch artifact task needs a upstream pipeline material that can only be configured for a pipeline and not template?

config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
```diff
@@ -67,12 +69,62 @@ public CaseInsensitiveString name() {
         return name;
     }
 
+    public void validateTree(ValidationContext validationContext, CruiseConfig preprocessedConfig) {
+        validate(validationContext);
+        validateParamsAndFetchTasks(preprocessedConfig);
+    }
+
+    private void validateParamsAndFetchTasks(CruiseConfig preprocessedConfig) {
```
jyotisingh: Can you try and profile this with a large enough config to see what kind of impact it adds.

config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
```diff
@@ -67,12 +69,62 @@ public CaseInsensitiveString name() {
...
+                        task.validate(contextForTasks);
+                        this.errors().addAll(task.errors());
+                    }
+                }
+            }
+        }
+    }
+
+    private void validateParams(PipelineConfig pipelineConfig) {
+        for (ParamConfig paramConfig : this.referredParams()) {
```
jyotisingh: `this.referredParams()` does a deepclone of template-config. Since validateParams is being invoked for each dependent pipeline, the clone operation becomes a function that. You might want to pull it out of the pipeline loop.

config/config-api/src/com/thoughtworks/go/config/PipelineTemplateConfig.java
```diff
@@ -67,12 +69,62 @@ public CaseInsensitiveString name() {
         return name;
     }
 
+    public void validateTree(ValidationContext validationContext, CruiseConfig preprocessedConfig) {
+        validate(validationContext);
+        validateParamsAndFetchTasks(preprocessedConfig);
+    }
+
+    private void validateParamsAndFetchTasks(CruiseConfig preprocessedConfig) {
+        List<CaseInsensitiveString> pipelineNames = preprocessedConfig.pipelinesAssociatedWithTemplate(this.name());
```
jyotisingh: you don't need this while creating a template, could avoid config parsing

server/test/unit/com/thoughtworks/go/config/update/UpdateTemplateConfigCommandTest.java
```diff
@@ -95,17 +95,18 @@ public void shouldAllowSubmittingInvalidElasticProfileId() throws Exception {
 
     @Test
     public void shouldValidateElasticProfileIdWhenTemplateIsUsedInAMaterial() throws Exception {
```
varshavaradarajan: WhenTemplateIsUsedInAPipeline, you mean?

PR #3037, opened 2016-12-23
Remove unwanted mail jars, and upgrade the remaining

ketan:


ketan:
@jyotisingh — looking at [the code](https://github.com/gocd/gocd/blob/d05746b57a1b28593ddcf667cbad0cea4dd65e30/server/src/com/thoughtworks/go/config/GoMailSender.java), it seems that the only implementation is [this](https://github.com/ketan/gocd/blob/d05746b57a1b28593ddcf667cbad0cea4dd65e30/server/src/com/thoughtworks/go/config/GoSmtpMailSender.java#L35-L35).

I don't see the need to pull down imap, dsn and pop3 mail jars — not sure why they were added to begin with.

ketan:
@arvindsv @jyotisingh — merging myself but if you see an issue later, can always rollback.

PR #3041, opened 2016-12-27
Allow template admins to access templates APIs v4.

varshavaradarajan:
@arvindsv, @jyotisingh, @ketan - According to https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/admin/templates/_templates_listing.html.erb#L23 , only Go System Administrators can delete a template. Do you know why we should retain this behaviour? If so, I'll update the PR to remove access to delete template API for template admins.


ketan:
Given that we’re talking about mere "edit permissions" for template admins (oxymoron!) — I don't feel they should be allowed to delete templates.

We'd likely need a fine grained authorization for templates, if you ask me.

PR #3042, opened 2016-12-28
Remove objnesis

ketan:


jyotisingh:
@ketan - looks like there is a runtime dependency on objenesis. All functional tests seem to be failing for this PR.

PR #3057, opened 2017-01-03
Rewrite all URLs to gocd.io

ketan:


ketan:
@arvindsv — I suppose we do want to move to .io long term, so I'm not merging this myself.

PR #3059, opened 2017-01-03
Footer links should open in a new tab.

naveenbhaskar:


ketan:
can we hold merging this. there is another pr to change host names

Issue #2350, opened 2016-06-25
One resource being loaded over HTTP


aj-jaswanth:
##### Issue Type

- Bug Report
##### Summary

 The endpoint is being accessed over HTTP 'http://server-address/go/api/version_infos/stale' even when the site is being accessed over HTTPS. 

Chrome gives a warning 'Mixed Content: The page at 'https://server-address/go/agents' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://server-address/go/api/version_infos/stale'. This request has been blocked; the content must be served over HTTPS.send @ application-788070d….js:174'
###### Basic environment details

- Go Version: `16.6.0`
- JAVA Version: `1.7.0_79`
- OS: `Centos 6.7`
- Browser vendor and version (if relevant): `Chrome 51`
##### Steps to Reproduce

1. 
2. 
3. 
4.

ketan:
If you're fronting your server with a reverse proxy, have you set the `x-forwarded-proto` header?

ketan:
@aj-jaswanth any updates? I'm going to mark this issue as closed unless you can provide more information. Exporting requests from chrome using HAR would be great in terms of resolving this.

aj-jaswanth:
@ketan Didn't get time to look into it. We can close this issue as it is not directly related to it. Moreover, we can use Upgrade-Insecure HTTP header to solve it.

ketan:
The point I'm trying to make is that this should work fine. If it is not,
then we need to investigate and put in a proper fix, and not duct-tape :)

stackfull:
I'm seeing this on 16.12. It seems to be only for that 1 request - all the others are working and the UI seems to be fully functional.

arvindsv:
/cc @maheshp - Can you take a look at this?

Taking a quick look at the code around [here](https://github.com/gocd/gocd/blob/500e24c0db41bb9ca7538412134dffe7a16b023e/server/webapp/WEB-INF/rails.new/app/views/navigation_elements/_footer.html.erb#L29), I think this could be because `apiv1_stale_version_info_url` is used and not `apiv1_stale_version_info_path`.

[...]

PR #3061, opened 2017-01-03
Version updater fixes

maheshp:
* Version update check was enabled on new SPA pages.
* New Go version notifications are displayed on new SPA pages.
* Version check is skipped on DevelopmentServer
* Using route helper with _path rather than _url to fix https://github.com/gocd/gocd/issues/2350

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/new_javascripts/models/shared/version_updater_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/new_javascripts/models/shared/version_updater_spec.js
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2017 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+define(['mithril', 'jquery', 'models/shared/version_updater'], function (m, $, VersionUpdater) {
+  describe("VersionUpdater", function () {
+    describe('update', function() {
+      beforeEach(function () {
+        localStorage.clear();
+      });
+
+      it('should fetch the stale version info', function () {
+        var deferred = $.Deferred();
+        var xhr = jasmine.createSpyObj(xhr, ['setRequestHeader']);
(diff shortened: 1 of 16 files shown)


server/webapp/WEB-INF/rails.new/app/views/new_navigation_elements/_footer.html.erb
```diff
@@ -22,4 +22,9 @@
             </span>
       </div>
   </div>
+  <script type="application/javascript">
+    require(['models/shared/version_updater'], function(VersionUpdater) {
```
ketan: See [the production build](https://build.gocd.io/go/tab/build/detail/installers/241/dist/1/dist) does not package this file. So this will fail in production. Additionally even if you do package the file in production, you're likely not going to get the correct hash for `version_updater` in production. Post that, this will cause additional http requests to the server. Making for a less-than-ideal page load experience. Any possibility that we add this to each SPA?

We already require `foundation.dropdownMenu` and a few other friends for the navigation header on top, so don't mind putting that in the top level require itself

server/webapp/WEB-INF/rails.new/spec/new_javascripts/models/shared/version_updater_spec.js
```diff
@@ -0,0 +1,116 @@
...
+define(['mithril','models/shared/version_updater'], function (m, VersionUpdater) {
+  describe("VersionUpdater", function () {
+    describe('update', function() {
+      it('should fetch the stale version info', function () {
+        var deferred = $.Deferred();
+        var xhr = jasmine.createSpyObj(xhr, ['setRequestHeader']);
+        var thirtyOneMinutesBack = new Date(Date.now() - 31 * 60 * 1000);
+
+        spyOn(localStorage, 'getItem').and.returnValue(JSON.stringify({last_updated_at: thirtyOneMinutesBack}));
+        spyOn(m, 'request').and.returnValue(deferred.promise());
```
ketan: Instead of stubbing `m.request` any possibility of directly using mock-ajax — makes for cleaner behaviorial tests. Here you're effectively testing what attributes you set on the deferred, aka — the internal implementation of `m.request`. If that changes, we'd have to re-write a whole bunch of tests, in addition to changing the actual code.

See [this](https://github.com/gocd/gocd/blob/5de3e1bc8b8868fb9ad931501d6f27d6b6a6dd3d/server/webapp/WEB-INF/rails.new/spec/new_javascripts/models/elastic_profiles/elastic_profiles_spec.js#L90-L118) for reference.

server/webapp/WEB-INF/rails.new/app/views/new_navigation_elements/_footer.html.erb
ketan: Additionally the `check_go_updates?` guard exists on old pages, but not in this one?

maheshp: I was looking at avoiding adding this script in each SPA, so was trying this option. Let me think bout and it we can look at the right way to do this.

```check_go_updates?``` needs to be added, will add it.

PR #3070, opened 2017-01-04
Fix update call to copy over the existing authorization.

varshavaradarajan:
Leaving Templates API V4 as is for now, in case we need to have a PATCH call for stage and auth config.


server/webapp/WEB-INF/rails.new/app/controllers/api_v2/admin/templates_controller.rb
```diff
@@ -42,7 +42,7 @@ def create
 
       def update
         result = HttpLocalizedOperationResult.new
-        updated_template = ApiV2::Config::TemplateConfigRepresenter.new(PipelineTemplateConfig.new).from_hash(params[:template])
+        updated_template = ApiV2::Config::TemplateConfigRepresenter.new(@template).from_hash(params[:template])
```
ketan: This might actually be a very dangerous thing to do... If the update fails midway, you'd have still applied partial changes onto the actual object as part of the deserialization call.

PR #3071, opened 2017-01-04
PluginProfile Changes and Role Config Backend changes

bdpiparva:


server/src/com/thoughtworks/go/config/update/RoleConfigCommand.java
```diff
@@ -0,0 +1,141 @@
...
+            return t;
+        }
+    }
+
+    private String getTagName() {
+        return role.getClass().getAnnotation(ConfigTag.class).value();
+    }
+
+    protected final boolean isAuthorized() {
+        if (!(goConfigService.isUserAdmin(currentUser) || goConfigService.isGroupAdministrator(currentUser.getUsername()))) {
```
ketan: Only super-admins should be allowed this operations. Group admins is purely for managing pipeline groups.

ketan: Additionally, I just looked at `SecurityAuthConfigCommand` and we're inheriting `isAuthorized` from `PluginProfileCommand` which was originally intented for `ElasticAgentProfileCommand` and subclasses. This too allows group admins. Might want to mark `PluginProfileCommand#isAuthorized` abstract, and pad direct subclasses of `PluginProfileCommand` with some tests.

PR #3072, opened 2017-01-04
Change the migration number for the removal of gadget related tables

varshavaradarajan:
Change the sql syntax to include the 'IF EXISTS' clause so that the migration doesn't bomb if already applied previously.

### DIFF
--- a/server/db/migrate/h2deltas/1701001_remove_gadget_oauth_tables.sql
+++ b/server/db/migrate/h2deltas/1701001_remove_gadget_oauth_tables.sql
@@ -14,14 +14,9 @@
 -- limitations under the License.
 --
 
--- H2 automatically creates indices on primary keys and foreign keys,
--- while postgres only creates indices on primary keys.
--- A migration was added to create the missing index on a foreign key in postgres.
--- To maintain sanity, and avoid mismatch of the migrations we have on h2 and postgres, this migration is added.
-
-DROP TABLE gadgetOauthAuthorizationCodes;
-DROP TABLE gadgetOauthAccessTokens;
-DROP TABLE gadgetOauthClients;
+DROP TABLE IF EXISTS gadgetOauthAuthorizationCodes;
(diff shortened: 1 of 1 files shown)


ketan:
This PR was already applied, so I don't see why we'd want to apply it again, even if to just bump the version number.

varshavaradarajan:
So that we are aware in which release, a migration was applied? There was no corresponding pg migration and I kept thinking why the pg build didn't break for 16.10 release. 
Also, to follow the convention and be consistent. We could just as well have marked all migrations from 1-142 otherwise.

ketan:
The numbers are meant more as a convention, and are ideally supposed to be in increasing order (db migrate does not care, it just applies migrations it does not find in the DB in an increasing order)

The fact that we put version numbers in there is meant more for GoCD developers than end users. This migration has already run on a bunch of servers where experimental builds are pushed out. Also the next question someone asks looking at this migration is why `DROP TABLE IF EXISTS ...`?

Additionally — this will skew the view of the `changelog` table where there'll be a migration added in the table, but not to be found anywhere else — something that most people don't want be dealing with during a 3:00AM support call :)

maheshp:
Agree with @ketan, lets not change the migration since this would have already run on multiple servers. Changing this would lead to inconsistency and confusion.

For future migrationsI would still prefer the convention of having the migration number prepended with the release number.

jyotisingh:
You are right - dbdeploy doesn't really care about the migration/delta number. Its for the developers/support people alone. The whole point of using the existing convention for the db migrations was so we do not have to think twice as to which release was a certain migration added, the delta number tells you about the release and order. Your point about skewing up the view of changelog table is valid. I agree, it wouldn't make sense to rename deltas that have gone out in a release for the reasons that you mention. But since the original migration hasn't gone live/released yet, we won't see this issue on any production instance. It would be isolated to the dogfooding environments alone. As I understand, the whole point of this PR is so people do not have to sit and scratch their heads during a support call as to when was a certain migration added.

Issue #3063, opened 2017-01-03
Config repo based pipelines should not have edit option on stage details and job details page


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Edit pipeline option should be removed from job details and stage details page for pipelines configured using config repo

###### Basic environment details

* Go Version: `16.12.0`

##### Steps to Reproduce

1. Add a pipeline using a config repository
2. After a run visit the stage details or job details page

##### Expected Results

Edit pipeline settings gear on top should not be available

##### Actual Results

Edit pipeline settings gear on top is available

rajiesh:
verified on 17.1.0 (4495-d95a3882d7eadb84330b641744e9a937ef8ed834), it works fine

PR #3075, opened 2017-01-05
Do not show the settings link in stage, job details, pipeline history  pages, if pipeline is not editable via the UI. (#3063)

varshavaradarajan:


server/webapp/WEB-INF/rails.new/app/controllers/stages_controller.rb
```diff
@@ -245,8 +245,9 @@ def convert_to_scale duration_array
     stage_duration_array
   end
 
-  def has_permission_to_view_settings
+  def can_view_settings?
```
ketan: Make this a helper method using `helper_method :can_view_settings?` and access it directly from the view, instead of via `@can_user_view_settings`

If performance is a concern, you can always memoize the result —

```ruby
def can_view_settings?
  @can_view_settings ||= begin
    group_name = ...
    !group_name.blank? && ...
  end
end
```

ketan:
On second thought, scratch what I said. Old code.

Issue #2900, opened 2016-11-25
Cyrillic characters are not displayed correctly in the console


APshenkin:
##### Issue Type
- Bug Report

##### Summary

Cyrillic characters are not displayed correctly in the console. Looks like GO Server can't convert characters from Windows Agents because of difference in encoding
Here is an example:
String ```Проверка перехода на страницу о достижениях```
displays ```�?�?ове�?ка пе�?е�?ода на �?�?�?ани�?�? �? до�?�?ижени�?�?``` if English was setted for non-Unicode Programs (OEM encoding cp437)

displays ```?????????�?????� ???�???�???????� ???� ???????�???????? ?? ???????????�?�????????``` if Russian was setted for non-Unicode Programs (OEM encoding cp866)

###### Basic environment details
* Go Version: 16.7.0
* JAVA Version: 1.8.0_66
* OS: Ubuntu 14.04

###### Additional Environment Details
Windows 7 x64 and Windows 10 x64 agents

##### Steps to Reproduce
1.  Add any logs with Cyrillic symbols in your application
2. Set language for non-Unicode Programs (English or Russian)
3. Assign Job with your application to Windows agent
4. Check the console output on Go Server

##### Expected Results
Cyrillic characters are displayed in console with correct encoding

##### Actual Results
Cyrillic characters are displayed with broken chars

##### Possible Fix
Before sending the buffer with updated console log, convert encoding of it to UTF-8

carrchang:
Chinese characters also, any way to set encoding of the console?

arvindsv:
Unfortunately not. If you or someone else can dig into this a bit more and see where it is being forced to be UTF-8, then maybe we might be able to change it.

ketan:
Closing as dup of #3926.

PR #3079, opened 2017-01-06
Force encoding of the console log to utf-8

ketan:
* ensure that the output is read in utf-8
* ensure that it is transmitted in utf-8
* ensure that it is written AS-IS byte-by-byte
* ensure that it is read in utf-8 before being rendered to the browser

### DIFF
--- a/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
+++ b/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
@@ -23,6 +23,7 @@ import org.apache.http.entity.StringEntity;
 import org.apache.log4j.Logger;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 public class RemoteConsoleAppender implements ConsoleAppender {
 
@@ -43,7 +44,7 @@ public class RemoteConsoleAppender implements ConsoleAppender {
                 LOGGER.debug("Appending console to URL -> " + consoleUri);
             }
 
-            putMethod.setEntity(new StringEntity(content));
+            putMethod.setEntity(new StringEntity(content, StandardCharsets.UTF_8));
             HttpService.setSizeHeader(putMethod, content.getBytes().length);
             CloseableHttpResponse response = httpService.execute(putMethod);
             if (LOGGER.isDebugEnabled()) {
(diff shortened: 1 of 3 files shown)


arvindsv:
Looks good to me. I guess there's no easy way for us to enable #2900.

ketan:
> I guess there's no easy way for us to enable #2900.

Not sure I understand. Care to elaborate?

arvindsv:
#2900 is about non-UTF8 characters being coerced into UTF-8, essentially, as far as I understand. Of course, there will be a bit about the console log GET method forcibly returning UTF-8 and the page forced to UTF-8 (in the view).

This PR seems to make that a little more hard-coded. Especially the changes to ConsoleService and StreamPumper. Just based on my quick look.

I understand that encoding is more about how you interpret a string than make any changes to it and that the bytes/pairs of the Cyrillic and Chinese characters are not really lost, but shown assuming it's UTF-8. But, as I said, this assumes all console output to be UTF-8, right?

ketan:
> But, as I said, this assumes all console output to be UTF-8, right?

Correct, I was talking about this with @maheshp yesterday. And realized that this does not feel like it's the right approach to the problem. Here's what we discussed —

GoCD "reads" bytes from various streams. The encoding of these streams is not directly controlled by GoCD —

* password file typically containing usernames with multi-byte chars (server)
* SCM material process execution (`git log`, `svn log`, `hg log`, etc) (server). SCM updates on the agent can probably be ignored, since GoCD does not read the output of such commands.
* Build script execution (agent)

Git, SVN and HG support a `--encoding` argument to override the commit encoding. However all of them claim that the encoding defaults to utf-8. And encourage that you users use that encoding. Git specifically stores the encoding of each commit object. 

For `git` — see [this](https://git-scm.com/docs/git-commit/#_discussion) and [this](https://git-scm.com/docs/git-log/#_discussion).
For `hg` - see [hg(1)](https://linux.die.net/man/1/hg) (search for encoding)
For `svn` - see http://svnbook.red-bean.com/en/1.7/svn.ref.svn.html

As far as reading these console outputs is concerned — I'm not sure what's the best way to proceed, I'm sure `jGit` can probably help us read the `git log` output in a better manner. We might need to look at other libraries for svn, hg and p4.

As far as console output for builds is concerned, we might need to consider not reading the output as `utf-8`. However that'd mean we figure out(not sure how), or allow the user to specify what the encoding is(not sure where, i.e. on each agent, or per job) — and more importantly, save that encoding (on the server) along with the console log, so we can render it in the correct encoding when sending it to the browser (on the job detail page).

PR #3083, opened 2017-01-07
Allow plugins with API version 2.0 to access plugin settings

salliewalecka:
With the changes in https://github.com/gocd/gocd/pull/3022, if you set your plugin message version to 2.0 the server throws an exception when you try and access the plugin settings.  These changes are to prevent that exception from occurring.


arvindsv:
Looks good. Tests pass. Merging.

PR #3084, opened 2017-01-09
Remove templates API V4.

varshavaradarajan:


maheshp:
V4 was added to support template authorization. Since authorization would be handled as a separate resource there is no need for V4, hence removing it.

PR #3086, opened 2017-01-09
Pipeline selections not saved in cookie

jyotisingh:
Pipeline selections is not determined based on the cookie. There is just one selection saved for the anonymous user which is shared across sessions

* Background: When security is turned off, all users are allowed anyonymous access to GoCD. Such users could select pipelines to be displayed for themselves. This information was saved to db, but the id of the saved 'selections' object was saved in a cookie. Whenever the dashboard was refreshed for each users, it would look for the selections-id in the cookie and render the dashboard based on that.
If the cookie was deleted or expired, a new selections object was saved in the the db again.
This commit gets rid of using the cookie to determine the pipeline sections for anonymous users. When security is not turned on, all users share the pipeline selections. Edit made my one user would be reflected for all users. If users really want to save their own preferences, they need to bring in the concept of real users (ie.turn on authentication)

* Cleaning up data for pipeline selections.


maheshp:
What do you think about turning off this feature when security is disabled?

This feature is for personalisation, in absence of a User,  personalisation does not makes sense.

jyotisingh:
Closing this in favor of implementing a full fledged support for anonymous user access to Go.

Issue #2943, opened 2016-12-06
Environments created by config repos should be editable for sections defined in main config xml


rajiesh:
##### Issue Type

- Bug Report

##### Summary

As part of this fix #2835 , the environments created by config repo plugins are accessible. However they have to be made editable so that user should be able to add/remove pipelines/agents and environment variables in main config xml from the environments UI page. The config repository pipelines/agents and environment variables will continue to be non editable. Please refer to the comments here - https://github.com/gocd/gocd/pull/2867#issuecomment-262719664

Steps to reproduce:

1. Create a config repository with a pipeline(say `config-repo-pipe`) and an environment(say `config-repo-env`)
2. Add the config repository to the cruise_config.xml using the <config-repos> section
3. Create a pipeline say `main-pipeline` from the server UI
4. Now visit the Environments page and select the environment `config-repo-env`
5. Click `Edit` on the pipelines section
6. The Config repo based pipeline is shown in separate section(but in non-editable mode) and pipelines defined in cruise-config.xml is shown in separate section(editable mode)
7. Select the pipeline `main-pipeline` and click add

Actual result:

Save failed with no valid description

Excepted result:

Save should be successful and the environment section in cruise-config.xml should be merged successfully with the partials from config repository and UI should reflect the change

rajiesh:
Verified on GoCD version 17.3.0 (4701-42a55e643a6e6a39ca07ce6bc00df7b4be2a28df)

PR #3092, opened 2017-01-12
Allow editing config-repo environments(#2943)

GaneshSPatil:


jyotisingh:
The above PR will fix the bug in question, but might leave a few other scenarios unaddressed. Environment config save needs to be looked at in detail, there are few aspect which do not seem right at this point in the current implementation. @GaneshSPatil and I are going over it right now. Will reopen the PR post that.

PR #3120, opened 2017-01-25
Remove hooks for job completion and and scheduling

ketan:


server/src/com/thoughtworks/go/server/service/JobInstanceService.java
```diff
@@ -133,7 +126,7 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
                 internalUpdateJobStateAndResult(job);
                 /* POST JOB COMPLETION HOOK - BEGIN */
```
jyotisingh: I guess we could get rid of the `if` block and the comments around it. Looks good otherwise.

Issue #2956, opened 2016-12-07
Authorization Plugin Endpoint


ketan:
##### Issue Type

- Feature proposal

##### Summary

Here's a proposal to move the current LDAP and Password file auth into a file

LDAP and password is currently configured as follows —

```xml
<security>
  <ldap uri="ldap://ldap.example.com" managerDn="cn=Bob,ou=Users,dc=example,dc=com" password="s3cr3t" searchFilter="(sAMAccountName={0})">
    <bases>
      <base value="ou=skunkworks,dc=example,dc=com" />
      <base value="ou=area51,dc=example,dc=com" />
    </bases>
  </ldap>
  <passwordFile path="/etc/go/password.properties" />

  <roles>
    <role name="SpaceTiger">
      <user>bob</user>
      <user>alice</user>
    </role>
  </roles>
</security>


<pipeline>
  <authorization>
    <admins>
      <role>SpaceTiger</role>
    </admins>
  </authorization>
</pipeline>
```

This configuration will now become (along with a hypothetical github authorization plugin) —

```xml
<security>

  <authConfigs>
    <!-- The password file configuration -->
    <authConfig pluginId="cd.go.authorization.passwordFile" id="password">
      <property>
        <key>hashType</key>
        <value>md5|bcrypt|sha1</value>
      </property>
    </authConfig>

    <!-- The github configuration -->
    <authConfig pluginId="cd.go.authorization.github" id="github">
      <property>
        <key>apiKey</key>
        <value>....</value>
      </property>
    </authConfig>

    <!-- The ldap server and base configuration -->
    <authConfig pluginId="cd.go.authorization.ldap" id="ldap">
      <property>
        <key>uri</key>
        <value>ldap://ldap.example.com</value>
      </property>
      <property>
        <key>managerDn</key>
        <value>...</value>
      </property>
      <property>
        <key>bases</key>
        <value>
          ou=skunkworks,dc=example,dc=com
          ou=area51,dc=example,dc=com
        </value>
      </property>
    </authConfig>
  </authConfigs>

  <roles>
    <role name="blackbird" authConfigId="ldap">
      <property>
        <key>group</key>
        <value>ou=blackbird,ou=area51,dc=example,dc=com</value>
      </property>
    </role>

    <role name="oss" authConfigId="github">
      <property>
        <key>organization</key>
        <value>gocd</value>
      </property>
      <property>
        <key>team</key>
        <value>Committers,Owners</value>
      </property>
    </role>

[...]

PR #3121, opened 2017-01-25
Deriving roles for users on authentication using the Authorization plugins #2956

bdpiparva:
refer #2956
* PluginRoleService manages PluginRoles for a user
* PluginRoleUserStore used to store users for a PluginRole
* Added endpoint for invalidation of cache from authorization plugin
* Unloading an Authentication plugins forces re-authentication of all
users through the RemoveAdminPermissionFilter

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
+++ b/config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
@@ -25,17 +25,13 @@ import static com.google.common.collect.Multimaps.synchronizedSetMultimap;
 
 public class PluginRoleUsersStore {
     private final SetMultimap<PluginRoleConfig, RoleUser> roleToUsersMappings = synchronizedSetMultimap(HashMultimap.create());
-    private static PluginRoleUsersStore self;
 
     private PluginRoleUsersStore() {
 
     }
 
     public static PluginRoleUsersStore instance() {
-        if (self == null) {
-            self = new PluginRoleUsersStore();
-        }
-        return self;
+        return PluginRoleUsersStoreHolder.PLUGIN_ROLE_USERS_STORE;
     }
 
     public void assignRole(String user, PluginRoleConfig pluginRoleConfig) {
@@ -46,8 +42,12 @@ public class PluginRoleUsersStore {
         return new ArrayList(roleToUsersMappings.get(pluginRoleConfig));
     }
 
-    public Set<PluginRoleConfig> pluginRoles() {
-        return new HashSet<>(roleToUsersMappings.keySet());
+    public void removePluginRolesNotIn(List<PluginRoleConfig> pluginRoles) {
+        for (PluginRoleConfig pluginRole : pluginRoles()) {
+            if (!pluginRoles.contains(pluginRole)) {
+                remove(pluginRole);
+            }
+        }
     }
 
     public void remove(PluginRoleConfig pluginRole) {
@@ -70,7 +70,16 @@ public class PluginRoleUsersStore {
         }
     }
 
+    protected Set<PluginRoleConfig> pluginRoles() {
+        return new HashSet<>(roleToUsersMappings.keySet());
+    }
+
+//    Used only in tests
     public void clearAll() {
         roleToUsersMappings.clear();
+    }
+
+    private static class PluginRoleUsersStoreHolder {
+        static final PluginRoleUsersStore PLUGIN_ROLE_USERS_STORE = new PluginRoleUsersStore();
     }
 }
--- a/config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
+++ b/config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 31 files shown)


config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.SetMultimap;
+
+import java.util.*;
+
+import static com.google.common.collect.Multimaps.synchronizedSetMultimap;
+
+/**
+ * Singleton class to hold list of users against a PluginRole. This class will be the source of truth to find users for a given PluginRole
```
ketan: No comments, please :) These start lying when the next person changes the implementation on the class.

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
```diff
@@ -0,0 +1,80 @@
...
+            self = new PluginRoleUsersStore();
+        }
+        return self;
+    }
+
+    public void assignRole(String user, PluginRoleConfig pluginRoleConfig) {
+        roleToUsersMappings.put(pluginRoleConfig, new RoleUser(user));
+    }
+
+    public List<RoleUser> usersWithRole(PluginRoleConfig pluginRoleConfig) {
```
ketan: `usersInRole`?

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
```diff
@@ -0,0 +1,80 @@
...
+        return new ArrayList(roleToUsersMappings.get(pluginRoleConfig));
+    }
+
+    public Set<PluginRoleConfig> pluginRoles() {
+        return roleToUsersMappings.keySet();
```
ketan: This keyset is mutable and changes to the keyset will cause changes in the underlying map. See if you need to return a copy?

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
maheshp: There have been instances in past where a comment was insisted upon, we need to decide on the convention to follow.

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
```diff
@@ -0,0 +1,83 @@
...
+        roleToUsersMappings.remove(pluginRole);
+    }
+
+    public void remove(Collection<PluginRoleConfig> pluginRoles) {
+        for (PluginRoleConfig role : pluginRoles) {
+            remove(role);
+        }
+    }
+
+    public void revokeAllRolesFor(String username) {
```
maheshp: I thought ```revokeAllRolesFor``` read better, if you think its confusing or not conveying the intent we can change

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
maheshp: sounds better

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
maheshp: returning a copy makes sense

config/config-api/src/com/thoughtworks/go/config/PluginRoleUsersStore.java
```diff
@@ -0,0 +1,76 @@
...
+        final RoleUser roleUser = new RoleUser(username);
+        synchronized (roleToUsersMappings) {
```
jyotisingh: Why is this synchronized?

server/src/com/thoughtworks/go/server/security/RemoveAdminPermissionFilter.java
```diff
@@ -89,4 +94,9 @@ public void onConfigChange(CruiseConfig newCruiseConfig) {
     private SecurityConfig securityConfig(CruiseConfig newCruiseConfig) {
         return newCruiseConfig.server().security();
     }
+
+    @Override
+    public void onAuthorizationPluginUnload() {
```
jyotisingh: This callback happens in case of plugin unload as well as when plugin initiates a request to invalidate the roles cache. May be rename the interface and method to suit both workflows

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverter.java
```diff
@@ -48,15 +49,19 @@
...
-    String searchUsersRequestBody(String searchTerm);
+    String searchUsersRequestBody(String searchTerm, List<SecurityAuthConfig> authConfigs);
 
     String processGetRoleConfigsRequest(String requestBody);
 
     String getProcessRoleConfigsResponseBody(List<PluginRoleConfig> roles);
+
+    String getInvalidateCacheResponseBody();
+
+    String getProcessPluginConfigResponseBody(Map<String, Map<String,String>> authConfigProfiles);
```
jyotisingh: Couldn't find any usages for `getProcessPluginConfigResponseBody`. remove it?

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
```diff
@@ -84,13 +85,27 @@ public String verifyConnectionRequestBody(Map<String, String> configuration) {
     }
 
     @Override
-    public String authenticateUserRequestBody(String username, String password) {
+    public String authenticateUserRequestBody(String username, String password, List<SecurityAuthConfig> authConfigs) {
```
jyotisingh: We are dealing with a Map and string keys in this class. Could we may be use a DTO for representing the data instead? Something like [this](https://github.com/gocd/gocd/blob/5e0a5079f8cc06b1292c006ca40950ca8aea2177/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/v1/StageNotificationDTO.java) which is used by Notification extension.  Things like `securityAuthConfig.getConfigurationAsMap` are also prone to break as the domain implementation changes.

Issue #3055, opened 2017-01-02
Getting an uncategorized SQL exception while opening a job details page


kashishm:
##### Issue Type

- Bug Report

##### Summary

Getting an exception when trying to access job details page.

###### Basic environment details

* Go Version: `16.12.0 (4352-018c65856d46007240cc75577e9cc8535b44148d)`
* JAVA Version: `1.7.0_85`
* OS: `Linux 2.6.32-504.el6.x86_64`

##### Steps to Reproduce

1. Create a pipeline named `foo` with a stage and job named `bar`, run it a few times.
2. Delete the pipeline `foo`.
3. Create a pipeline with the same name i.e `foo` with a different stage and job name i.e `baz` and run it.
4. Try to access the `baz` job run details page (gives an exception).
> The build count for the `baz` job is starting from `1` but the pipeline history page is showing the details of deleted pipeline's stages and jobs.

##### Log snippets
```
org.springframework.jdbc.UncategorizedSQLException: SqlMapClient operation; uncategorized SQLException for SQL []; SQL state [null]; error code [0]; Error: executeQueryForObject returned too many results.; nested exception is java.sql.SQLException: Error: executeQueryForObject returned too many results.
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:83)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:80)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:80)
	at org.springframework.orm.ibatis.SqlMapClientTemplate.execute(SqlMapClientTemplate.java:203)
	at com.thoughtworks.go.server.transaction.SqlMapClientDaoSupport$SqlMapClientTemplate.execute(SqlMapClientDaoSupport.java:112)
	at org.springframework.orm.ibatis.SqlMapClientTemplate.queryForObject(SqlMapClientTemplate.java:268)
	at com.thoughtworks.go.server.transaction.SqlMapClientDaoSupport$SqlMapClientTemplate.access$501(SqlMapClientDaoSupport.java:39)
	at com.thoughtworks.go.server.transaction.SqlMapClientDaoSupport$SqlMapClientTemplate$4.execute(SqlMapClientDaoSupport.java:97)
[...]

Issue #1471, opened 2015-09-02
Go Server - Console tab not loading when new pipeline created with same name as deleted pipeline


sharathsatish:
The issue is that we had created a test pipeline with name lets say "ABC" and then deleted it. Later when another pipeline was created with the same name. We couldn't access any of the console logs for that build.

Here is the stack trace:

```
Caused by: java.sql.SQLException: Error: executeQueryForObject returned too many
 results.
        at com.ibatis.sqlmap.engine.mapping.statement.MappedStatement.executeQue
ryForObject(MappedStatement.java:124)
        at com.ibatis.sqlmap.engine.impl.SqlMapExecutorDelegate.queryForObject(S
qlMapExecutorDelegate.java:518)
        at com.ibatis.sqlmap.engine.impl.SqlMapExecutorDelegate.queryForObject(S
qlMapExecutorDelegate.java:493)
        at com.ibatis.sqlmap.engine.impl.SqlMapSessionImpl.queryForObject(SqlMap
SessionImpl.java:106)
        at org.springframework.orm.ibatis.SqlMapClientTemplate$1.doInSqlMapClien
t(SqlMapClientTemplate.java:270)
        at org.springframework.orm.ibatis.SqlMapClientTemplate.execute(SqlMapCli
entTemplate.java:200)
```

ketan:
What version of go are you on?

zabil:
might be related to this https://github.com/gocd/gocd/issues/971

sharathsatish:
Yes there was a change in case, but it was deleted first and then a new one created with a different case. We are on v 15.2.0

zabil:
More context here https://github.com/gocd/gocd/issues/879 and [here](https://groups.google.com/forum/#!topic/go-cd/LVnapi82dmg/discussion)

Unfortunately, you cannot delete a pipeline and pipeline names were changed to be case insensitive, however some queries were not fixed (there's a lot out there) Is that the full stack trace? If not can you post the rest of it?

jyotisingh:
Deletion of a pipeline from config does not get rid of the history from db, so Go would treat both of them as the same pipeline for most part - well there are a few *case-sensitivity bugs* and this looks like one of them to me.

ivanmoore:
I was going to raise this bug but see it has already been raised. Here's a "test" of sorts which shows one of the manifestations of changing case of a pipeline. This uses [Gomatic](https://github.com/SpringerSBM/gomatic) against GoCD running on localhost.

```python
[...]

PR #3124, opened 2017-01-25
1471 case sensitivity issues - #1471, #3055

jyotisingh:


varshavaradarajan:
Another bug that @jyotisingh came across and has to fix.

PR #3132, opened 2017-01-30
Sort resources and environments alphabetically on agents page dropdowns

iamsumanth:


server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/agents/environments.js
```diff
@@ -18,17 +18,24 @@ define(['mithril', 'lodash', 'helpers/mrequest', 'models/agents/tri_state_checkb
...
+  };
+
   Environments.init = function (selectedAgents) {
     m.request({
       method:        'GET',
       url:           Routes.apiv1AdminInternalEnvironmentsPath(),
       config:        mrequest.xhrConfig.v1,
       unwrapSuccess: function (responseBody) {
-        Environments.list = _.map(responseBody, function (value) {
-          return new TriStateCheckbox(value, _.map(selectedAgents, function (agent) {
```
ketan: This could have been written as?

```
Environments.list = _.map(responseBody.sort(), ...)
```

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/agents/resources.js
```diff
@@ -18,19 +18,25 @@ define(['mithril', 'lodash', 'helpers/mrequest', 'models/agents/tri_state_checkb
...
+  };
+
   Resources.init = function (selectedAgents) {
     m.request({
       method:        'GET',
       url:           Routes.apiv1AdminInternalResourcesPath(),
       config:        mrequest.xhrConfig.v1,
       unwrapSuccess: function (responseBody) {
-        var list = _.map(responseBody, function (value) {
-          return new TriStateCheckbox(value, _.map(selectedAgents, function (agent) {
```
ketan: Same as above?

server/webapp/WEB-INF/rails.new/app/assets/new_javascripts/models/agents/environments.js
iamsumanth: Yeah. I will do ```responseBody.sort()``` but will put it in another method because ```_.map(selectedAgents, function (agent) ``` part of the code iterates for every environment which is not required. And it will be difficult to read if I extract that variable in ```unwrapSuccess``` itself.

Issue #3119, opened 2017-01-25
Go server linux installers should not automatically start the Go server service


vivekprahlad:
##### Issue Type
- Feature proposal
##### Summary

Currently, according to https://github.com/gocd/gocd/blob/master/installers/linux/shared/partials/_after_install_go-server.sh.erb, the Go server is automatically started once it is installed via linux installers. 

This is problematic in situations where Go server AMIs or Docker containers are built using the linux installers (automatically starting the server leads to non-deterministic builds, as the server takes a while to do so, and makes a number of changes to the file system as it boots), and this leaves the zip file based install as the only option.

IMHO, it is trivial to explicitly start the service. However, the opposite is not true: it is currently not possible to prevent the service from automatically starting. Interestingly, the go agent installer does not have this behaviour.

Other open source packages (for example, ElasticSearch) used to have this behaviour as well, but have since disabled it. Please see [this issue](https://github.com/elastic/elasticsearch/issues/3722) as an example.

arvindsv:
I think @ketan had a thought similar to this. I remember talking about installation vs. upgrade.

ketan:
@vivekprahlad — this has been discussed a number of times before. I will get back to you with something in a couple of days.

PS: Hi!

ketan:
@vivekprahlad — could you take a look at https://github.com/gocd/gocd/pull/3133 and see if it's in line with what you expect?

vivekprahlad:
@ketan - Thanks, looks good!

rajiesh:
Verified on GoCD version 17.2.0 (4587-442a7d67087844e44a2dcf5824164e1dfe00cc45). In debian and rpm based installers after fresh install the go-server and go-agent do not start automatically. In upgrade scenario the state of the service before upgrade is maintained

PR #3133, opened 2017-01-30
Do not start service on a new install. Allow users to configure the agent and server and have them start it up. (#3119)

ketan:


ketan:
Here's the diff of the package scripts on debian —

https://gist.github.com/ketan/460271873845ad46d897015f15b48dbe


Here's the diff of the package scripts on RPMS —

https://gist.github.com/ketan/80cd68bc1eea627d80c0eea877fe8bfc/revisions

server/config/cruise-config.xml
```diff
@@ -1,28 +1,80 @@
 <?xml version="1.0" encoding="utf-8"?>
-<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="87">
+<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="88">
```
varshavaradarajan: This file shouldn't be checked in, right?

ketan: Probably not as part of this PR. But if we bump the version, we should just check this file in.

varshavaradarajan: Sure. Its already done though. https://github.com/gocd/gocd/blob/master/server/config/cruise-config.xml.

varshavaradarajan:
Pending: Show a post install message on fresh install of the go server.

PR #3137, opened 2017-01-30
Only search the mail field in LDAP for a valid email address

karelbemelmans:
Only search the mail field in LDAP for a valid email address for the user, not in the userPrincipleName field. The userPrincipleName field can contain a value that is not active as an email address, so using it is not recommended.

### DIFF
--- a/server/src/com/thoughtworks/go/server/security/LdapUserSearch.java
+++ b/server/src/com/thoughtworks/go/server/security/LdapUserSearch.java
@@ -53,7 +53,6 @@ public class LdapUserSearch implements org.springframework.security.ldap.LdapUse
     private static final String SAM_ACCOUNT_NAME = "sAMAccountName";
     private static final String UID = "uid";
     private static final String COMMON_NAME = "cn";
-    private static final String USER_PRINCIPLE_NAME = "userPrincipalName";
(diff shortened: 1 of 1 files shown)


bdpiparva:
@ketan: making sense to me. We already fixed that by https://github.com/gocd/gocd-ldap-authentication-plugin/blob/master/src/main/java/cd/go/authentication/ldap/UserMapper.java#L32 but this fix is part of plugin.

karelbemelmans:
So you are saying we should use the plugin? I can't see the use for that plugin as the Go server has this LDAP authentication built in already?

ketan:
@karelbemelmans — we're working building the plugin alongside a new authorization endpoint (#2956) that'll allow plugging in authentication and authorization and allow for more flexibility in how LDAP is configured and setup.

ketan:
@karelbemelmans — your PR looks good. I've triggered a PR build to check if this breaks something.

In the meanwhile, would you mind signing the [CLA](https://gocd.io/cla).

varshavaradarajan:
@karelbemelmans - Can you please sign the [CLA](https://gocd.io/cla) so that we can get this merged?

karelbemelmans:
Done.

Issue #3081, opened 2017-01-06
Templates API - Description of feature


arvindsv:
There seems to be some confusion about some edge cases in the templates API. I'll write what I feel is the ideal API behavior and it can be critiqued / discussed here.

1. [Template administrators](https://docs.gocd.io/current/configuration/dev_authorization.html#specifying-permissions-for-templates) can be composed of users and/or roles.

2. Template administrators of a template should be able to GET, PUT and PATCH (?) an existing template by name.

3. *Any* pipeline group administrator should be able to POST (create) a new template, for which they automatically become template administrator.

4. Template API GET should provide a link to authorization (not embedded, but a sub-resource of the template). Template API PUT and PATCH should not allow those to be changed by a template administrator who is not a super-administrator.

5. Only super administrators should be able to GET, PUT, PATCH template authorization.

This allows any administrator of pipelines to create a template. It allows multiple administrators (pipeline group administrators) to collaborate and *use* a template. But, to add a second pipeline group administrator to a template created by one pipeline group administrator will involve a super administrator.

I feel this is a good balance between self-service templates and involving super administrators only in authorization related cases.

/cc @varshavaradarajan

rajiesh:
@arvindsv isn't the template administrator permission one which a pipeline group admin should be granted by default. A pipeline administrator without template admin permissions and a template admin without pipeline admin permission is of less value.

IMO pipeline group admin should have permission to edit the templates in the group(irrespective of who created it) and an explicit authorization for templates can be avoided

arvindsv:
> IMO pipeline group admin should have permission to edit the templates in the group(irrespective of who created it) and an explicit authorization for templates can be avoided

The problem is that templates are a global concept. They're not tied to a group. That's what point 3 looks to help with. It says that any pipeline group admin can create a template, and they will be granted admin permission on it.

Given that it is global, we can't get rid of explicit auth, because you might want to restrict certain templates to certain people.

Does that make sense?

jyotisingh:
Would it make sense to allow a template admin to add more collaborators?

arvindsv:
> Would it make sense to allow a template admin to add more collaborators?

It would make sense, to remove the super admin from the picture, essentially. But then, can those new collaborators add others? If so, everyone is a template admin. It's like the GRANT option in a relational DB. We don't have a "collaborator" as an option in auth currently (without GRANT, that is).

[...]

PR #3145, opened 2017-01-31
Template Authorization UI and API

varshavaradarajan:
(#3081)


varshavaradarajan:
This PR does the following
- Have template view authorization and copy the existing pipeline group admins to the view config of every template in the config (the XSL migration).
- Allow template admins to delete templates via UI and API
- Allow template view users to access to get the template configuration via UI and API
- Allow template view users to access the template listing via UI and API - It is a filtered view of templates for a specific user.

This PR does not take care of -
- Introducing a new version of templates API to add the `can_edit` attribute in the pipeline summary for the templates index call
- Does not allow pipeline group admins to create a template and be made template admins by default.

server/src/com/thoughtworks/go/config/update/DeleteTemplateConfigCommand.java
```diff
@@ -54,7 +55,15 @@ public boolean isValid(CruiseConfig preprocessedConfig) {
 
     @Override
     public boolean canContinue(CruiseConfig cruiseConfig) {
-        return super.canContinue(cruiseConfig) && doesTemplateExist(cruiseConfig);
+        return isUserAuthorized() && doesTemplateExist(cruiseConfig);
```
arvindsv: Can you see if `TemplateConfigCommand`'s `canContinue` method is correct? Should it call `isAuthorizedToEditTemplate` too, instead of `isUserAdmin`? It is being used only in `CreateTemplateConfigCommand` since you removed the only other usage of `super.canContinue` in this context.

I guess it's fine, because only pipeline group admins and above can create the template. If so, I'd make the `canContinue` in `TemplateConfigCommand` fully abstract and move the implementation down to `CreateTemplateConfigCommand`, since it's the only one using it.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1462,6 +1462,292 @@ public void shouldCreateProfilesFromMultipleAgentConfigsAcrossPipelines_asPartOf
         assertThat(profiles.get(1), is(expectedAWSProfile));
     }
 
+    @Test
+    public void shouldCreateAViewElementInEachTemplateIfPipelineGroupAdminsExist_asPartOfMigration89() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
```
arvindsv: To be extremely technical and pedantic, only pipeline group admins who had pipelines depending on templates could see those templates. Right? Or at least, because of an issue where permissions were not properly checked, anyone who knew a template name could view it. Now, all group admins can view all templates by default, unless they're removed. Right?

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1462,6 +1462,292 @@ public void shouldCreateProfilesFromMultipleAgentConfigsAcrossPipelines_asPartOf
...
+                + "         <admins>"
+                + "             <user>user1</user>"
+                + "             <role>role1</role>"
+                + "         </admins>"
+                + "     </authorization>"
+                + " </pipelines>"
+                + " <pipelines group='second'>"
+                + "     <authorization>"
+                + "         <admins>"
+                + "             <user>user2</user>"
```
arvindsv: Can you change/add one of these users, to have a repeated user between the two groups and similarly a repeated role between the two groups, so that you can show that the final list has no duplicates?

Edit: I see you have a different test for it later (which doesn't test unique roles, though).

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1462,6 +1462,292 @@ public void shouldCreateProfilesFromMultipleAgentConfigsAcrossPipelines_asPartOf
...
+
+        String xmlString = migrateXmlString(configXml, 88);
+        String template1 = "<templates>\n"
+                +"<pipeline name=\"template1\">\n"
+                +"<authorization>\n"
+                +"<view>";
+        String template2 = "<pipeline name=\"template2\">\n"
+                + "<authorization>\n"
+                + "<view>";
+        assertThat(xmlString, not(containsString(template1)));
```
arvindsv: What's the purpose of this?

server/test/unit/com/thoughtworks/go/server/service/GoConfigServiceTest.java
```diff
@@ -1135,11 +1135,33 @@ public void shouldSayThatAUserIsAuthorizedToViewAndEditTemplatesWhenTheUserHasPe
 
         assertThat(goConfigService.isAuthorizedToViewAndEditTemplates(new Username(templateAdminName)), is(true));
         assertThat(goConfigService.isAuthorizedToViewAndEditTemplates(new Username(secondTemplateAdminName)), is(true));
-        assertThat(goConfigService.isAuthorizedToViewAndEditTemplates(new Username(new CaseInsensitiveString(theSuperAdmin))), is(false));
+        assertThat(goConfigService.isAuthorizedToViewAndEditTemplates(new Username(new CaseInsensitiveString(theSuperAdmin))), is(true));
```
arvindsv: Do you know why it was `false` earlier? That seems wrong. What were the consequences of that?

server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -251,7 +251,7 @@
                 /admin/restful/configuration/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /admin/pipelines/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
                 /admin/pipeline_group/**=ROLE_SUPERVISOR,ROLE_GROUP_SUPERVISOR
-                /admin/templates/**=ROLE_SUPERVISOR,ROLE_TEMPLATE_SUPERVISOR
+                /admin/templates/**=ROLE_USER
```
arvindsv: Shouldn't this be: `ROLE_SUPERVISOR,ROLE_TEMPLATE_SUPERVISOR,ROLE_TEMPLATE_VIEW_USER` or something like that?

server/webapp/WEB-INF/rails.new/app/controllers/admin/templates_controller.rb
```diff
@@ -134,10 +136,12 @@ def load_templates from
   end
 
   def load_cruise_config
-    result = HttpLocalizedOperationResult.new
-    assert_load :cruise_config, go_config_service.loadCruiseConfigForEdit(current_user, result)
-    unless result.isSuccessful()
-      render_localized_operation_result result
+    if security_service.isUserAdmin(current_user) || security_service.isAuthorizedToViewAndEditTemplates(current_user)
```
arvindsv: I think `isAuthorizedToViewAndEditTemplates` is already checking `isUserAdmin`.

arvindsv: Also, isn't this kind of a check taken care of in the `before_filter`?

server/webapp/WEB-INF/rails.new/app/controllers/config_view/templates_controller.rb
```diff
@@ -16,11 +16,18 @@
...
   class TemplatesController < ConfigView::ConfigViewController
+    before_action :is_user_authorized_to_view_template
 
     def show
       result = HttpLocalizedOperationResult.new
       @template_config = template_config_service.loadForView(params[:name], result)
       render_localized_operation_result(result) unless result.isSuccessful()
     end
+
+    def is_user_authorized_to_view_template
```
arvindsv: Mark this "private"?

server/webapp/WEB-INF/rails.new/app/views/admin/templates/_templates_listing.html.erb
```diff
@@ -2,77 +2,81 @@
...
-                        <%- elsif !is_pipeline_empty -%>
-                            <%= action_icon({:type => "delete", :text => l.string("DELETE"), :title => l.string("CANNOT_DELETE_TEMPLATE", template, pipelines), :disabled => true}) %>
-                        <%- else -%>
-                            <%= form_tag(delete_template_path(:pipeline_name => template), :method => :delete, :id => "delete_template_#{template}") %>
-                            <%= md5_field %>
-                            <span class="delete_icon action_icon delete_parent" id="trigger_delete_<%= template %>" title="<%= l.string("DELETE_TEMPLATE") -%>">
+        <div class="title_action_wrapper">
+          <ul>
+            <li>
+              <%- if is_user_an_admin? || is_user_a_template_admin? %>
```
arvindsv: Is the `is_user_an_admin?` really needed here?

server/webapp/WEB-INF/rails.new/lib/config_update/check_is_template_admin.rb
```diff
@@ -0,0 +1,29 @@
+##########################################################################
+# Copyright 2016 ThoughtWorks, Inc.
```
arvindsv: 2017?

server/webapp/WEB-INF/rails.new/spec/controllers/config_view/templates_controller_spec.rb
```diff
@@ -25,8 +25,46 @@
     end
   end
 
+  describe 'security' do
+
+    it 'should disallow anonymous users, with security enabled' do
```
arvindsv: What happens when security is disabled? Free for all?

server/webapp/WEB-INF/rails.new/spec/views/admin/templates/index_html_spec.rb
```diff
@@ -283,16 +351,38 @@
     end
   end
 
-  it "should disable the delete button next to the template name for template admin" do
+  it "should display the delete button next to the template name for template admin" do
```
arvindsv: Hmm. I don't remember discussing whether template admins can now, suddenly, delete a template. I suppose it makes sense, given that they can create it? The only thing they cannot do is edit its permissions? It seems like existing template admins are now a little more powerful, with no way to reduce that.

varshavaradarajan: https://github.com/gocd/gocd/issues/3081#issuecomment-274057578 - We discussed and we didn't know why template admins were not allowed to do this before. It made sense before because only those who can create a template should be able to delete it. Then, we discussed that pipeline group admins can create a template and be made template admins by default. So, if pipeline group admins can create a template, then they should be able to delete it as template admins, which means all template admins should be able to delete it.

server/webapp/WEB-INF/rails.new/spec/controllers/config_view/templates_controller_spec.rb
varshavaradarajan: Yes.

server/webapp/WEB-INF/applicationContext-acegi-security.xml
varshavaradarajan: Oh. Yeah. Eliminates the need for before filters in the controller. Thanks!

server/test/unit/com/thoughtworks/go/server/service/GoConfigServiceTest.java
varshavaradarajan: No impact at all. Because the method `isUserAuthorizedToViewAndEditTemplates` was only checking the template authorization and not checking if user is a super admin. So, wherever we used this method on the rails views, the `is_user_an_admin` method was called. Like this - 
`is_user_an_admin? || is_user_authorized_to_view_and_edit_templates?`

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
varshavaradarajan: Yes. Is that okay?

server/webapp/WEB-INF/applicationContext-acegi-security.xml
varshavaradarajan: Actually, no. Still need the before filters. But still. Makes sense to have this.

server/webapp/WEB-INF/rails.new/app/controllers/admin/templates_controller.rb
varshavaradarajan: Well, for index, we need to load the cruise_config for an admin or template admin, and not for template view users. But we need to allow the template view users to reach the action index. If we don't make this check, then it throws an error for template view user in the template listing page.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
varshavaradarajan: Correction: 
> only pipeline group admins who had pipelines depending on templates could see those templates. Right?

No. When a pipeline group admin wants to create a pipeline using a template, they'll need to view all templates. So, all pipeline group admins ended up seeing the configuration of all templates.

server/webapp/WEB-INF/rails.new/spec/controllers/config_view/templates_controller_spec.rb
jyotisingh: I guess that's a cue for writing a test with security disabled :)

config/config-api/src/com/thoughtworks/go/config/TemplatesConfig.java
```diff
@@ -100,15 +100,29 @@ public PipelineTemplateConfig templateByName(CaseInsensitiveString foo) {
...
+        for (PipelineTemplateConfig templateConfig : this) {
+            if (hasViewAccessToTemplate(templateConfig, username, roles, isGroupAdministrator)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public boolean canUserEditTemplate(PipelineTemplateConfig template, CaseInsensitiveString username, List<Role> roles) {
+        return template.getAuthorization().getAdminsConfig().isAdmin(new AdminUser(username), roles);
```
jyotisingh: Try pushing methods such as `getAuthorization().getViewConfig().isAdmin` and `getAuthorization().getAdminsConfig().isAdmin` to Authorization class

server/src/com/thoughtworks/go/config/update/DeleteTemplateConfigCommand.java
jyotisingh: The check should be inverted else it will throw a null pointer exception. The same for Update-template-config-command

server/webapp/WEB-INF/rails.new/spec/controllers/config_view/templates_controller_spec.rb
arvindsv: You were always the perceptive one. :P

server/webapp/WEB-INF/rails.new/app/controllers/admin/templates_controller.rb
```diff
@@ -16,6 +16,8 @@
 
 class Admin::TemplatesController < AdminController
 
+  before_filter :check_admin_user_and_401, only: [:edit_permissions, :update_permissions, :new, :create]
+  before_filter :check_admin_or_template_admin_and_401, only: [:destroy, :update]
```
jyotisingh: Missed seeing this before, but `edit` action is now available for template-view-users. This needs to be fixed.

server/webapp/WEB-INF/rails.new/app/views/admin/stages/index.html.erb
```diff
@@ -10,17 +10,15 @@
...
-                            <% template_name = (@pipeline.getTemplateName() || @template_list.first).to_s %>
+                          <%= f.label l.string("SELECT_TEMPLATE") %>
+                          <%= f.select(com.thoughtworks.go.config.PipelineConfig::TEMPLATE_NAME, @template_list, {:selected => @pipeline.getTemplateName()}, {:id => "select_template"}) %>
 
-                            <%= link_to l.string("VIEW"), "#", :class=> "view_template_link action_icon button preview_icon skip_dirty_stop"%>
-                            <% if is_user_an_admin? %>
-                                <%=link_to l.string("EDIT"), template_edit_path(:pipeline_name => template_name, :stage_parent => "templates", :current_tab => 'general'), :class=> "action_icon edit_icon edit_template_link"%>
-                            <% end %>
+                          <% template_name = (@pipeline.getTemplateName() || @template_list.first).to_s %>
+                              <%= link_to l.string("VIEW"), "#", :class => "view_template_link action_icon button preview_icon skip_dirty_stop" %>
```
jyotisingh: view link should be added/enabled based on user's permissions. If the allowgroup-admin flag on template is turned off, this link shouldn't be available

jyotisingh:
/cc @varshavaradarajan 
Pending: 
* This PR makes template view link on pipeline creation wizard available only for super-admin. To be fixed as a part of a separate PR. 
* Pipeline group admins cannot create a template as of this PR. That would be handled in a separate PR.

PR #3146, opened 2017-02-01
ReAuthenticate users after a time interval #2956

maheshp:
* Added a ReAuthenticationFilter to trigger re-authentication after an
  interval of time.
* The idea is re-authentication would enable refereshing the pluginRoles
  for a user derived from a plugin.
* ReAuthentication and the interval are configurable through
  SystemEnvironment, currently this feature is turned off.


server/test/integration/com/thoughtworks/go/server/security/ReAuthenticationFilterTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Test
+    public void shouldReAuthenticateIfReAuthTimeIntervalHasElapsed() throws IOException, ServletException {
+        long currentTimeMillis = DateTimeUtils.currentTimeMillis();
+        long minuteBack = DateTimeUtils.currentTimeMillis() - 60000;
+        Authentication authentication = setupAuthentication();
+
+        when(timeProvider.currentTimeMillis()).thenReturn(currentTimeMillis);
+        when(systemEnvironment.isReAuthenticationEnabled()).thenReturn(true);
+        when(systemEnvironment.reAuthenticationTimeInterval()).thenReturn(55000L);
+        when(session.getAttribute("last_reauthentication_check_time")).thenReturn(minuteBack);
```
GaneshSPatil: can we extract this into constant  `LAST_REAUTHENICATION_CHECK_TIME` or use the same constant as of the filter?

base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -811,6 +813,14 @@ public boolean isProductionMode() {
         return GO_SERVER_MODE.getValue().equalsIgnoreCase("production");
     }
 
+    public boolean isReAuthenticationEnabled() {
+        return REAUTHENTICATION_ENABLED.getValue();
+    }
+
+    public long reAuthenticationTimeInterval() {
```
GaneshSPatil: can we say `getReAuthenticationTimeInterval`?

Issue #3164, opened 2017-02-10
Using different charset for reading and writing to config file


maheshp:
##### Issue Type

- Bug Report

##### Summary

The config xml file is [written](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/config/GoConfigFileWriter.java#L55) using the charset ```UTF_8``` but a charset is not specified while [reading](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/config/GoConfigFileReader.java#L33) it. In absence of a charset specified while reading the reader relies on the default charset which is based on the locale of underlying OS.

This inconsistency in using the charset could lead to undesired config change upon reading and writing back to the config.

PR #3165, opened 2017-02-10
Config file reader uses UTF_8 charset while reading the file #3164

maheshp:
**WIP - Do not merge**
* With this change, both of ConfigReader and ConfigWriter would use the
  same charest

### DIFF
--- a/server/src/com/thoughtworks/go/config/GoConfigFileReader.java
+++ b/server/src/com/thoughtworks/go/config/GoConfigFileReader.java
@@ -22,6 +22,8 @@ import org.apache.commons.io.FileUtils;
 import java.io.File;
 import java.io.IOException;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 public class GoConfigFileReader {
     private final SystemEnvironment systemEnvironment;
 
@@ -30,7 +32,7 @@ public class GoConfigFileReader {
     }
 
     public String configXml() throws IOException {
-        return FileUtils.readFileToString(fileLocation());
+        return FileUtils.readFileToString(fileLocation(), UTF_8);
     }
 
     public File fileLocation() {
--- a/base/src/com/thoughtworks/go/util/FileUtil.java
+++ b/base/src/com/thoughtworks/go/util/FileUtil.java
@@ -30,6 +30,7 @@ import java.io.OutputStream;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.net.URI;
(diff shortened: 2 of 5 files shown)


server/src/com/thoughtworks/go/config/GoConfigFileReader.java
```diff
@@ -21,16 +21,19 @@
...
     private final SystemEnvironment systemEnvironment;
+    private final Charset UTF_8 = StandardCharsets.UTF_8;
 
     public GoConfigFileReader(SystemEnvironment systemEnvironment) {
         this.systemEnvironment = systemEnvironment;
     }
 
     public String configXml() throws IOException {
-        return FileUtils.readFileToString(fileLocation());
+        return FileUtils.readFileToString(fileLocation(), UTF_8);
```
ketan: Simply use static import instead of a constant that points to another one?

jyotisingh:
There are a few other places too which rely on the default-charset as far as the config xml is concerned. 

* During [checkin](https://github.com/gocd/gocd/blob/4ef23ba6cba377a4896562c60d8a31f680955f25/config/config-server/src/com/thoughtworks/go/service/ConfigRepository.java#L110) of the config xml to the config.git repo, default-charset is assumed there. 
* If the config file was found to be invalid during a server startup, it reverts itself from the last known good revision from config.git. Check [this](https://github.com/gocd/gocd/blob/4ef23ba6cba377a4896562c60d8a31f680955f25/config/config-server/src/com/thoughtworks/go/config/GoConfigMigration.java#L148) and [this](https://github.com/gocd/gocd/blob/4ef23ba6cba377a4896562c60d8a31f680955f25/config/config-server/src/com/thoughtworks/go/config/GoConfigMigration.java#L160). 
* Config upgrade (xsl migration) when a server upgrades - check [this](https://github.com/gocd/gocd/blob/4ef23ba6cba377a4896562c60d8a31f680955f25/config/config-server/src/com/thoughtworks/go/config/GoConfigMigration.java#L123)

maheshp:
@lijie7 any changes to the Materials through the UI are written to config xml by explicitly using ```UTF-8``` encoding but the read from the config xml relies on the Default encoding of the OS.

So our assumption here is that since a single encoding is not used for both reads and writes there could be distortion as in your case.

While we make the necessary code changes to fix this issue, will you be able to change the locale of your OS to use ```UTF-8``` encoding to see if this issue gets resolved.

maheshp:
This needs some more thought and testing, closing for now.

PR #3168, opened 2017-02-12
WIP - Spring upgrade

bdpiparva:
WIP: Do not merge.


ketan:
@bdpiparva — I'm going to close this since this since it's a herculean task to upgrade spring. This may involve upgrading just about every library that spring uses (hibernate, commons-*, quartz, ibatis, cglib, ...)

Can you instead open small tiny PRs upgrading individual libraries that are easier to merge?

Many of these may actually not require a review from someone if all it is is a change in `build.gradle` or some minor code changes to accommodate for change in API of the libraries.

bdpiparva:
@ketan — That make more sense to me.

PR #3177, opened 2017-02-14
Fix for editing config-repo environments, renamed method names according to the usage forEdit or forDisplay

GaneshSPatil:
contains a tiny fix as well for editing config-repo environments.

needs to be merged for release 17.2

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
@@ -161,11 +161,11 @@ public class EnvironmentConfigService implements ConfigChangedListener {
         return environments.named(new CaseInsensitiveString(environmentName));
     }
 
-    public EnvironmentConfig forEdit(String environmentName) {
-        return cloner.deepClone(environments.find(new CaseInsensitiveString(environmentName)));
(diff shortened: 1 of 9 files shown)


server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -161,11 +161,11 @@ public EnvironmentConfig getEnvironmentConfig(String environmentName) throws NoS
         return environments.named(new CaseInsensitiveString(environmentName));
     }
 
-    public EnvironmentConfig forEdit(String environmentName) {
-        return cloner.deepClone(environments.find(new CaseInsensitiveString(environmentName)));
+    public EnvironmentConfig getEnvironmentForEdit(String environmentName) {
```
jyotisingh: @GaneshSPatil - Can you write a test to reproduce the bug that you fixed. This specific change is the fix for the bug, but its not very evident as to what was the bug, so there is nothing stopping anyone from changing this back to `environments.find`

GaneshSPatil: @jyotisingh -- will write one.

PR #3179, opened 2017-02-15
Changing the page_header to be a fixed element at the top of the page

ibnc:
* This will allow easier navigation back to the job/stage/pipeline level when looking at a long console log

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_tailing.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_tailing.js
@@ -64,7 +64,7 @@
 
         topActionBar.pinOnScroll({
           'z-index':      100,
-          top:            41,
+          top:            91,
           requiredScroll: 250
         });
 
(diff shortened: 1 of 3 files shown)


naveenbhaskar:
image

agents job run history page is broken.
agents page -> click on any agent name

ibnc:
@naveenbhaskar Thanks for taking a look! I just pushed a fix, and validated things locally

ibnc:
Fixed the conflicts.

arvindsv:
I took a look at this. Every page I tried looks good. I tried to put in as much data as I can to get it to scroll. It seems to work fine. One place where the header still scrolled was the preferences page (click on the username on the top right and then "Preferences" - http://localhost:8153/go/tab/mycruise/user).

@naveenbhaskar - Even without this, I think the rest of it is ready to be merged.

arvindsv:
From a functionality perspective, that is.

ibnc:
@arvindsv You're totally right. That page is one of the old VM pages, and the page header style is "page-header" instead of "page_header" like everywhere else. I'll push a fix for this

PR #3182, opened 2017-02-15
add gradle generated .idea project to make dev setup quicker

marques-work:
- contains no machine-specific paths, should be portable to any machine
  - excludes workspace.xml (for obvious reasons)


marques-work:
Changed my mind - please don't merge. Probably not a good idea to do this as folks may be using different versions of IDEA, or have non-default locations for $HOME/.gradle or $HOME/.m2, etc. it's not really that hard to generate with modern versions of IDEA.

PR #3183, opened 2017-02-15
Add display names to pipelines

brbutler:
This adds a display name to pipelines in order to get some of the benefits of renaming without actually messing with the database.


brbutler:
There is one defect in this change that I'm working to resolve now. When creating a new pipeline, validation fails when switching from the first tab to the second, due to a missing URL.  I'm not sure why the save is triggering early, but I'm troubleshooting and will push a fix.

varshavaradarajan:
@brbutler - thank you for the PR. A lot has changed in the last year, so, closing this PR right now.

PR #3189, opened 2017-02-17
Adjustments to the position of the console action bar

ibnc:
Now that the page header is a fixed element again, the console action bar's positioning and scrolling are off. This PR contains fixes for both, and makes the scrolling action smoother by ensuring we take into account the height of the action bar when it becomes a fixed element. Making the scrolling smoother required changes to the pinOnScroll jquery plugin that I believe @ketan wrote; so, I'd love your feedback @ketan.

Before positioning fixes:
before

before making the scrolling smoother:
before_scroll

End result:
after


ketan:
LGTM. Merged

PR #3190, opened 2017-02-17
Styling changes to the white console log theme

ibnc:
* Changed action bar colors to be more consistent with how the theme is handled in the darker theme
* Changed the white console log bg color to be an ever so off white to provide subtle separation (#fdfdfd as opposed to #ffffff)
* Changed action bar link colors to accommodate the bg color changes.

One thing I'm not 100% confident about is the link color for the actions in the action bar. I'm open for feedback on my choice of color.

Before:
screen shot 2017-02-17 at 11 14 40 am

After:
screen shot 2017-02-17 at 11 16 19 am

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/build_detail.scss
@@ -301,7 +301,7 @@ div.build_detail_container_content {
 .console-area {
   background-color: #333;
   .white-theme & {
-    background-color: white;
+    background-color: #fdfdfd;
   }
 }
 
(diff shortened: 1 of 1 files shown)


ketan:
@ibnc — Not sure if this bit of history helps in any way :(

I believe we did experiment with a white (ish) action bar with blue(ish) text. IIRC one of the feedback that we received was that the action bar did not stand out from the background (esp on non-retina displays). When there was text behind the action bar (long console lines), the text on the action bar was not noticeable because it mangled/blurred with the console log text in the background. We figured that using an action bar color (grayish) that contrasted with the background color (whiteish) worked well for most cases.

You may need to experiment with long console lines and how the various colors help ensure that the action items stand out in most cases — on a non-retina display.

PR #3193, opened 2017-02-20
Fix for job scheduling on Agent.

bdpiparva:
Remove job from job plans if it is assignable to an agent or an elastic agent.


ketan:
LGTM. Please add a test.

Issue #3068, opened 2017-01-04
Force `file.encoding` to `utf-8`


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Currently the file encoding depends on a variety of factors, which causes unexpected behavior wrt server and db functionality. Additionally different encodings on the server and agent process also means that the console logs sometimes get rendered with unexpected glyphs.

Would like to change the encoding on the server,  using the startup scripts, which seems fairly straightforward - with an option to override it if necessary. It seems that a change on the agent startup scripts would not necessarily help users who are upgrading, we may need to consider putting in some code in the launcher to force the encoding, with some options to be able to override it.

arvindsv:
Yes, we should do this. But, we need to test it properly. We've never been able to properly reproduce a different encoding.

ketan:
Adding my comment from https://github.com/gocd/gocd/pull/3079#issuecomment-271491885

> But, as I said, this assumes all console output to be UTF-8, right?

Correct, I was talking about this with @maheshp yesterday. And realized that this does not feel like it's the right approach to the problem. Here's what we discussed —

GoCD "reads" bytes from various streams. The encoding of these streams is not directly controlled by GoCD —

* password file typically containing usernames with multi-byte chars (server)
* SCM material process execution (`git log`, `svn log`, `hg log`, etc) (server). SCM updates on the agent can probably be ignored, since GoCD does not read the output of such commands.
* Build script execution (agent)

Git, SVN and HG support a `--encoding` argument to override the commit encoding. However all of them claim that the encoding defaults to utf-8. And encourage that you users use that encoding. Git specifically stores the encoding of each commit object. 

For `git` — see [this](https://git-scm.com/docs/git-commit/#_discussion) and [this](https://git-scm.com/docs/git-log/#_discussion).
For `hg` - see [hg(1)](https://linux.die.net/man/1/hg) (search for encoding)
For `svn` - see http://svnbook.red-bean.com/en/1.7/svn.ref.svn.html

As far as reading these console outputs is concerned — I'm not sure what's the best way to proceed, I'm sure `jGit` can probably help us read the `git log` output in a better manner. We might need to look at other libraries for svn, hg and p4.

[...]

PR #3198, opened 2017-02-23
Send console log from agent to server via websocket

kierarad:
- Behind toggle existing websocket enabled toggle
 - BuildSession commands use the websocket to send console output as well
 - Ensure all console output sent via the websocket is properly formatted

In an effort to improve the responsiveness of console logs, this utilizes websockets to send console output to the server from the agent instead of buffering and appending it via HTTP PUT request. In other words, sending the console log output as the agent produces it.

This does not include any analysis or testing of the performance impact of using websockets.

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
+++ b/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
@@ -160,7 +160,7 @@ public class AgentWebSocketClientController extends AgentController {
         URLService urlService = new URLService();
         StreamConsumer buildConsole;
 
-        if (new SystemEnvironment().isConsoleLogsThroughWebsocketEnabled()) {
+        if (getSystemEnvironment().isConsoleLogsThroughWebsocketEnabled()) {
             buildConsole = new ConsoleOutputWebsocketTransmitter(webSocketSessionHandler, buildSettings.getBuildId());
         } else {
             buildConsole = new ConsoleOutputTransmitter(
--- a/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
+++ b/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
@@ -157,10 +157,8 @@ public class AgentWebSocketClientController extends AgentController {
 
     private void runBuild(BuildSettings buildSettings) {
         URLService urlService = new URLService();
-        ConsoleOutputTransmitter buildConsole = new ConsoleOutputTransmitter(
-                new RemoteConsoleAppender(
-                        urlService.prefixPartialUrl(buildSettings.getConsoleUrl()),
-                        httpService));
+        ConsoleOutputWebsocketTransmitter buildConsole = new ConsoleOutputWebsocketTransmitter(webSocketSessionHandler, buildSettings.getBuildId());
+
         ArtifactsRepository artifactsRepository = new UrlBasedArtifactsRepository(
                 httpService,
                 urlService.prefixPartialUrl(buildSettings.getArtifactUploadBaseUrl()),
@@ -187,11 +185,7 @@ public class AgentWebSocketClientController extends AgentController {
             getAgentRuntimeInfo().busy(new AgentBuildingInfo(buildSettings.getBuildLocatorForDisplay(), buildSettings.getBuildLocator()));
             build.build(buildSettings.getBuildCommand());
         } finally {
-            try {
-                buildConsole.stop();
-            } finally {
-                getAgentRuntimeInfo().idle();
-            }
+            getAgentRuntimeInfo().idle();
         }
         this.buildSession.set(null);
     }
--- a/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
+++ b/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
@@ -36,6 +36,7 @@ import com.thoughtworks.go.remote.work.RemoteConsoleAppender;
 import com.thoughtworks.go.remote.work.Work;
 import com.thoughtworks.go.server.service.AgentBuildingInfo;
 import com.thoughtworks.go.util.*;
+import com.thoughtworks.go.util.command.StreamConsumer;
 import com.thoughtworks.go.websocket.Action;
 import com.thoughtworks.go.websocket.Message;
 import com.thoughtworks.go.websocket.MessageEncoding;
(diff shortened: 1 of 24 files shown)


agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
```diff
@@ -159,19 +156,25 @@ public void processSetCookieAction() throws IOException, InterruptedException {
 
     @Test
     public void processAssignWorkAction() throws IOException, InterruptedException {
+        new SystemEnvironment().set(SystemEnvironment.WEBSOCKET_ENABLED, true);
```
ketan: You probably want to remember and reset this variable at the end of the test.

common/src/com/thoughtworks/go/websocket/ConsoleTransmission.java
```diff
@@ -0,0 +1,69 @@
...
+        this.jobIdentifier = jobIdentifier;
+    }
+
+    public ConsoleTransmission(String line, String buildId) {
+        this.line = line;
+        this.buildId = buildId;
+    }
+
+    public InputStream getLineAsStream() throws IOException {
+        return IOUtils.toInputStream(getLine(), "UTF-8");
```
ketan: There's unfortunately no guarantee that the line was `utf-8`. See https://github.com/gocd/gocd/issues/3068#issuecomment-271520246

common/src/com/thoughtworks/go/websocket/Transmission.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.websocket;
+
+import com.google.gson.annotations.Expose;
+import com.thoughtworks.go.domain.JobIdentifier;
+
+/**
+ * Created by kierarad on 2/22/17.
```
ketan: Please remove :)

common/src/com/thoughtworks/go/websocket/ConsoleTransmission.java
ibnc: What would you suggest in place of the line? Should we just use the deprecated IOUtils#toInputString?  (which is not passing an encoding at all just giving it a string)

ibnc: This issue also sounds bigger than this PR, and seems like it's going to require some thought on how gocd handles encoding + console output in gerenal.

For now, I'm thinking just using the platform default encoding should be okay? Or at least no any worse than what's there already. I was thinking of changing the method to be the following:

````
public InputStream getLineAsStream() throws IOException {
    return new ByteArrayInputStream(getLine().getBytes());
}
````

How does that sound @ketan ??

Comment (ibnc) [inline common/src/com/thoughtworks/go/websocket/ConsoleTransmission.java]: I talked to @arvindsv, and he says that the changed I asked about should be fine. I'll go ahead and push that up

**Comment by ibnc** [comment]
@ketan I've added a toggle for this work

**Comment by ibnc** [comment]
It looks like the checks "failed" because somebody cancelled some of the running stages

**Comment by ibnc** [comment]
@ketan Any chance this can be merged?

File: agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java [hunk 4]
```diff
@@ -157,10 +158,18 @@ void process(Message message) throws InterruptedException {
 
     private void runBuild(BuildSettings buildSettings) {
         URLService urlService = new URLService();
-        ConsoleOutputTransmitter buildConsole = new ConsoleOutputTransmitter(
+        StreamConsumer buildConsole;
+
+        if (new SystemEnvironment().isConsoleLogsThroughWebsocketEnabled()) {
```
jyotisingh: use `getSystemEnvironment().isConsoleLogsThroughWebsocketEnabled()` instead?

agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
```diff
@@ -159,19 +156,95 @@ public void processSetCookieAction() throws IOException, InterruptedException {
...
+        Message message = argumentCaptor.getAllValues().get(1);
         assertThat(message.getAcknowledgementId(), notNullValue());
         assertThat(message.getAction(), is(Action.ping));
         assertThat(message.getData(), is(MessageEncoding.encodeData(agentController.getAgentRuntimeInfo())));
+
+        Message message2 = argumentCaptor.getAllValues().get(0);
+        assertThat(message2.getAcknowledgementId(), notNullValue());
+        assertThat(message2.getAction(), is(Action.consoleOut));
+        assertThat(message2.getData(), is(MessageEncoding.encodeData(new ConsoleTransmission("Sleeping for 0 milliseconds", new JobIdentifier()))));
+        env.set(SystemEnvironment.WEBSOCKET_ENABLED, false);
```
jyotisingh: If one of the assertions fail, resetting of the properties will not happen, causing some other tests to fail. Mocked `systemEnvironment` is already being passed along to the controller, may be setup the expectations according to the test, and use the same instead of instantiating a new one in [AgentWebSocketClientController#runBuild](https://github.com/kierarad/gocd/blob/2a909964c401ae088def31da815b56e7d6caf475/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java#L163)

base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -184,6 +184,8 @@
     public static GoSystemProperty<Integer> GO_ENCRYPTION_API_MAX_REQUESTS = new GoIntSystemProperty("go.encryption.api.max.requests", 30);
 
     public static GoSystemProperty<Boolean> WEBSOCKET_ENABLED = new GoBooleanSystemProperty("go.agent.websocket.enabled", false);
+    public static GoSystemProperty<Boolean> CONSOLE_LOGS_THROUGH_WEBSOCKET_ENABLED = new GoBooleanSystemProperty("go.agent.console.logs.websocket.enabled", false);
```
jyotisingh: Does it make sense to set the default to `true` since this kicks in only when websockets is [turned on](https://github.com/kierarad/gocd/blob/2a909964c401ae088def31da815b56e7d6caf475/common/src/com/thoughtworks/go/work/DefaultGoPublisher.java#L82)? When we turn on websockets for any of the environments, console logging will start happening over websockets by default then. Given that we have this property available, we could always turn it off later if we find any issue with it.

ibnc: It does to me. What do you think @ketan @arvindsv ??

agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
ibnc: You're totally right. I've changed the test to use the mock of SystemEnvironment

base/src/com/thoughtworks/go/util/SystemEnvironment.java
arvindsv: Makes sense to me too, to make it true by default.

ibnc: Cool. I've made the change

jyotisingh:
@ibnc - Could you get rid of the instantiation and setting up of SystemEnvironment from [here](https://github.com/kierarad/gocd/blob/bee4e3d6d97d7e7b79873bd5a79cf08f63296845/agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java#L177). Its not required anymore

PR #3199, opened 2017-02-23
Console Log UI Improvements (foldable sections and color coded output)

marques-work:
Console Log UI improvements:

* Prefix all log lines with control metadata, identifying lines as task start, task status, STDERR, STDOUT, etc.
* Refactored build output observer into separate log line parser, log scroller, and build summary updater
* log line parser detects collapsible boundaries via prefixes with fault tolerance
* prefixes also used to colorize output (don't mind the colors, I'm not artistic)
* hide timestamps by default
* added line numbers to log output


common/src/com/thoughtworks/go/domain/builder/Builder.java
```diff
@@ -101,23 +94,23 @@ public int hashCode() {
     }
 
     public void cancel(DefaultGoPublisher publisher, EnvironmentVariableContext environmentVariableContext, TaskExtension taskExtension) {
-        publisher.consumeLineWithPrefix("Start to execute cancel task: " + cancelBuilder.getDescription());
+        publisher.consumeLineWithPrefix("Cancel task: " + cancelBuilder.getDescription());
```
jyotisingh: I think `Cancel task <task1>` would incorrectly imply that task1 was being cancelled. Users can setup and `on-cancel-task` for a given job which gets executed when that job execution has been cancelled. The description being logged here was for the on-cancel task not the original task that was being executed.

marques-work: Sorry for the late reply - I've since changed it to read "On Cancel Task: ...", which is the same wording in UI. Does that work?

common/src/com/thoughtworks/go/remote/work/ConsoleOutputTransmitter.java
```diff
@@ -86,10 +94,9 @@ public void flushToServer() {
             consoleAppender.append(result.toString());
         } catch (IOException e) {
             LOGGER.warn("Could not send console output to server", e);
-            //recreate buffer
             synchronized (buffer) {
                 sent.addAll(buffer);
-                buffer = new CircularFifoBuffer(10 * 1024);
+                buffer.clear();
```
jyotisingh: So, when the code was originally written, it used to be `buffer.clear()`. This was later changed to recreate the buffer as a part of a commit with a message "Fixed the performance problem". There were other changes made as a part of that commit, so I am not entirely sure if this was intentional or not. The change looks ok to me but I am still trying to think if anything will go wrong. Can you think of a reason?

marques-work: Hi @jyotisingh, I cannot think of a reason this would impact performance or how it would be logically more correct to instantiate a new buffer object rather than clearing the existing one. It may have been unintentional.

Worse though, the code synchronizes on `buffer`, and changing the reference to a new object defeats the purpose of the mutex -- that's why I changed it back to `clear()`.

jyotisingh: This change makes sense to me too. Its ok, ignore it for now. This would be dog-fooded to a few environments so we should get some feedback, lets just keep it at the back of our heads.

jyotisingh:
@marques-work - Java side of things look good to me. Still looking at the details of the javascript changes done, sorry I am going a little slow. Currently looking LogOutputTransformer. That lead me to try out the feature with a slightly large console log - this was generated by running rpsec in Go. The job details page load took about 4 secs. On expanding the task section, it took about 3-4 more secs for the task details to be rendered. [This](https://gist.github.com/jyotisingh/8cb7334a1716b6eef6f4c06057595829) is the log file I used if you wanted to give it a try.   

I will continue to look at the code and let you know if I find anything.

jyotisingh:
Also, the functional tests are failing, have a look when you get some time -> https://build.gocd.io/go/pipelines/value_stream_map/installers-PR/294

marques-work:
@jyotisingh Thanks for the large console log, it's given me a chance to profile the UI.

## First load performance
On first load, a huge bottleneck is actually the ANSI color coding as we have to do it line by line now, which is inefficient compared to before where we made a single pass over the entire content. But perhaps a bigger part is the side effect that we need to set the content via innerHTML. Chrome attributes 75% of the execution time parsing HTML this way.

In contrast, setting a text node to the contents of the line is significantly faster. **I'm not certain what to do about this one yet.**

The other slow parts are the sheer number of elements we create (each line is at least 2 elements - timestamp and content, plus whatever else AnsiUP injects).

Ironically, calculating styles is faster on create than it is during expand/collapse sections, most likely because most sections are automatically collapsed (for tasks that pass anyway) and thus removed from the render tree when you first open the logs.

## Expand/Collapse performance
The bottleneck with collapsing/expanding sections is mostly around style calculation and re-layout. I will try to move to class-based styling rather than attribute selectors to improve on this front. Regarding re-layout, I'll have to think about how to approach this. I'll try to come up with a way to do this via mostly composite-only properties that won't require re-layout, but no guarantees.

jyotisingh:
@marques-work - I can see some improvement in page load time for completed jobs since the console log doesn't load up by default. But since we construct a bunch of elements for each line in the console log, the performance of this page will deteriorate as the log file grows. I have added another log file to the gist I shared yesterday (https://gist.github.com/jyotisingh/8cb7334a1716b6eef6f4c06057595829#file-rather_long_legacy_console_log_that_doesnot_open_up_at_all_console-log). This is from one of our other builds. The console tab doesn't even show up for this. 

I understand that it going to be difficult to optimize the page for such large logs. I was wondering what if we did not do any of this styling for extremely large log files? By styling I mean both the new things that this PR introduces as well as the existing ( ie. ansi_up). So, the users with decent enough log size could get all the benefits. The ones with large logs would see things a plain old logs. Would that make sense?

marques-work:
@jyotisingh I couldn't open that log file in my instance running master, and I'm even having trouble opening that log file in raw mode :). Chrome refuses to scroll to the end and leaves me with a mostly blank page. I had to download to find out it was 1.3M lines log, and 68 MB in size.

So this log may not even be useful in the current implementation, even without ansi_up.

I would propose something more drastic - for logs that large, only offer a download/raw option? I'm not certain how we would treat running builds as we won't have size information before we tail the logs. I suppose since I'm already keeping line count, we could use that and abort? Or I can try to keep track of bytes? Any other ideas?

marques-work:
That said, I still have some optimizations in the queue to make the 20,000 line log file usable. I think that much is possible, and it's a reasonable size for a real build.

arvindsv:
@marques-work @jyotisingh @naveenbhaskar 

> That said, I still have some optimizations in the queue to make the 20,000 line log file usable. I think that much is possible, and it's a reasonable size for a real build.

I'd say: Think about hiding everything except the first 500 lines and the last N lines (N >= 1000) and point to the raw logs, if they want the lines in-between.

jyotisingh:
@arvindsv - that makes sense for the feature as such. But I am starting to wonder what do we target for 17.3? Is it all of the things that we have been speaking about?

arvindsv:
I don't think we should tie this to a release. If it goes beyond a release, it does. I know 17.3 might not have a lot of stuff we can talk about, but that's ok. There's work which has gone in which we are not yet comfortable turning on.

naveenbhaskar:
can somebody post a screenshot here pls

marques-work:
@naveenbhaskar Here are some screenshots:

Completed build (tasks that passed are collapsed, tasks that failed are expanded by default)
image

White theme:
image

Build in progress (task line is yellow and section is expanded because we don't know the status yet)
image

With job cancelation:
image

White theme of the above:
image

jyotisingh:
@marques-work - since 17.3 is out of the doors now, I was planning to get back to review/merge this. There seems to be a few conflicts, could you please get this rebased.

naveenbhaskar:
@marques-work .. UI looks good.. thanks

jyotisingh:
@marques-work - there are still some conflicts in the css. please have a look.

marques-work:
@jyotisingh ok, resolved all conflicts. hopefully the build is all good.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_fansi_pants.js
```diff
@@ -0,0 +1,220 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+;(function (_, c) {
+  "use strict";
+
+  function HelloMrfANSIpants() {
```
jyotisingh: Cute name but we need to change it :)

marques-work:
@jyotisingh ok, it's done :)

jyotisingh:
First of all, it does look so much more snappier than last time. 

I have been thinking a bit about the ansi-up adaptation. Adapting and making changes to it seems fine. But given that the original work (ansi_up) is still being actively maintained, I am a little concerned about how do we go about with the upgrades. Would it make sense to fork the original one, make the changes over there, so if there are some considerable changes made in ansi_up, it still allows us to upgrade and patch it up with the changes you have made. Else, we would be stuck with this and probably never be able to upgrade. Or if you have any better ideas around making sure that we have the upgrade path still open, then I am happy to hear that out too.

BTW, found a rather tiny issue:
* If there are empty lines in the console log, the line-numbers for the empty line overlap on top of the next line (lines 2309 and 2310 are overlapping, same for 2313 & 2314).

screen shot 2017-03-30 at 4 29 58 pm 1

ketan:
>
> I have been thinking a bit about the ansi-up adaptation. Adapting and
> making changes to it seems fine. But given that the original work (ansi_up)
> is still being actively maintained, I am a little concerned about how do we
> go about with the upgrades. Would it make sense to fork the original one,
> make the changes over there, so if there are some considerable changes made
> in ansi_up, it still allows us to upgrade and patch it up with the changes
> you have made.
>
Does it make sense pushing these changes upstream, or pushing some changes
upstream that allow us to customize behavior and keep the customizations to
ourselves?

marques-work:
@jyotisingh @ketan I had the same hesitations when I started modifying the library. I'd have to give it some thought - the modifications change the fundamental API (returning a heterogeneous array of nodes and strings instead of properly escaped HTML).

marques-work:
@jyotisingh @ketan I could at least start with a fork so that we have a way to merge and resolve upstream changes.

marques-work:
@jyotisingh the line number issue should be fixed now. Going through pipelines.

jyotisingh:
@marques-work - acceptance tests are failing for your build => https://build.gocd.io/go/pipelines/value_stream_map/trigger/864 . Its the changes to task label. I thought there was a functional-tests PR for that, but I cant see one now. Or may be I am confusing it with the task description one. 

>I could at least start with a fork so that we have a way to merge and resolve upstream changes.

Sounds good to me

marques-work:
@jyotisingh yeah I thought I submitted one for that too. I'll try again with the matcher. I still haven't been able to run the functional tests myself. I know @ibnc had trouble as well.

marques-work:
@jyotisingh The acceptances are passing (well, there's one failure -- but that's on master too, so it's not related to this PR).

The regressions are failing too, but don't appear to be related either.

PR #3210, opened 2017-02-27
Gear settings links should send you to the settings for the page you are on

kierarad:
* stage details page -> that stage's settings
* job details page -> that job's settings

Note: We were unsure what the behavior should be for pipelines with templates. We debated maintaining the original behavior (links would go to pipeline general tab) so that users wouldn't inadvertently update a template. However, we ultimately decided to keep the behavior the same for pipelines with and without templates. In both cases stage details links to stage settings and job details link to job settings.

### DIFF
--- a/server/src/com/thoughtworks/go/server/controller/JobController.java
+++ b/server/src/com/thoughtworks/go/server/controller/JobController.java
@@ -110,10 +110,12 @@ public class JobController {
 
     private ModelAndView getModelAndView(JobInstance jobDetail) throws Exception {
         final JobDetailPresentationModel presenter = presenter(jobDetail);
+        String pipelineName = jobDetail.getPipelineName();
         Map data = new HashMap();
         data.put("presenter", presenter);
         data.put("l", localizer);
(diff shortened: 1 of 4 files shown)


ketan:
@kierarad — see https://build.gocd.io/go/pipelines/build-windows/1421/build-server/1/jobs.

varshavaradarajan:
@kierarad , @ketan , The settings icon pointing to /go/admin/pipelines/general was done intentionally for 2 reasons -
1. Supposing a pipeline used a template, then you'd have to check whether the user had permission to edit the template before showing the gear icon. I see that you have checked whether the template name exists or not only. 

2. We'll be making the quick edit page the default page sometime, at which point, there won't be different urls to go to.

arvindsv:
@varshavaradarajan - About point 2 there, we should be able to have direct URLs with the SPA too. It's fairly standard to use `#/abc` to route to the right part of an SPA, right?

This feature would be nice to have, I feel, if we can make sure your concerns are addressed.

varshavaradarajan:
@arvindsv , yeah, I think it would be nice too. We'll need to handle templates on the SPA as well then, which is where it might get a little complicated depending on whether we render the templates read only section or not for a particular user.

Issue #3187, opened 2017-02-17
Improvements on current SPAs


GaneshSPatil:
### General

- [x] randomize test (make sure test clears the setup)
- [x] Better dealing with global singletons (Code level fix)
- [x] move to ES6 :)
- [x] better/consistent error representation on all the pages
- [x] freeze editing when saving any config
- [x] provide feedback on save

### Agents SPA

- [x] spinner on resources list until the ajax call finishes
- [x] spinner on environments list until the ajax call finishes
- [x] Memory leak on agents SPA. (leads to aw Snap!)
- [x] Retail Sort Order after page reload(the sort order should be persisted even after the page reload).

### Elastic Agent Profiles SPA

- [x] spinner when fetching plugin infos after selecting a plugin

### Config SPA

- [x] Spinner for all the ajax polls (plugins infos, scms, package repos)
- [x] Saving an SCM material with no name throws 500 server error

bdpiparva:
#### One bug on Elastic Agent Profiles SPA

2 tooltip icons on `profileId`

![screen shot 2017-03-01 at 12 36 40 pm](https://cloud.githubusercontent.com/assets/7871209/23449334/ff46145c-fe7b-11e6-97df-0b7a2d9d0b42.png)

GaneshSPatil:
Many of the above-mentioned issues have been fixed as part of **SPA componentization** work.
Closing this issue.

Will open separate issues for each issue item, as we observe them on the new SPAs

PR #3217, opened 2017-03-01
Agents Page Memory Leak Fix

GaneshSPatil:
**Pass resolver which specifies the key to be cached for the given agent**

Not passing key to the memoize method will take object reference as key by default.
Whereas on each poll we create a new agent object, and the memoize treats the newly created agent object as a new agent and will cache it separately.
This was causing memoize to cache all the objects leading to memory overflow.


GaneshSPatil:
fixes #3187

PR #3222, opened 2017-03-02
Added rails vendor directory, DS_Store files, and old agent log files to gitignore

kierarad:


.gitignore
```diff
@@ -128,9 +128,8 @@ server/webapp/WEB-INF/rails.new/log/
 server/webapp/WEB-INF/rails.new/logs/
 server/webapp/WEB-INF/rails.new/node_modules/
 server/webapp/WEB-INF/rails.new/public/assets/
-server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/doc/
-server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/extensions/
 server/webapp/WEB-INF/rails.new/webpack/gen/
+server/webapp/WEB-INF/rails.new/vendor/
```
arvindsv: As far as I know, you can't gitignore vendor. Some parts are not from the Gemfile. For instance, [see this](https://github.com/gocd/gocd/blob/03f39b7f21154fd41c3fc1f5770edb7540a19d9a/server/webapp/WEB-INF/rails.new/Gemfile#L19).

arvindsv: Even if they were all from it, I wonder if we've had to make changes in some cases. :(

Also, with a change such as this, it's possible that a `git clone` and then a build will probably fail. Not immediately of course, since git won't remove ignored files, but if any new file gets added to `vendor`, it won't get checked in.

kierarad: @arvindsv duh, that makes sense

PR #3237, opened 2017-03-07
Update test for console log websockets toggle

kierarad:
@jyotisingh Per your request on the previous PR #3198, I've changed the [test](https://github.com/kierarad/gocd/blob/bee4e3d6d97d7e7b79873bd5a79cf08f63296845/agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java#L177) so that we aren't instantiating a new system environment. 

It is necessary to set the properties, because the mock system environment available in the controller is not what is being called in this test. I accounted for your comment that "if one of the assertions fail, resetting of the properties will not happen, causing some other tests to fail" by moving the property clearing to the tear down so that it will happen regardless of is it passes or fails.

### DIFF
--- a/agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
+++ b/agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
@@ -37,8 +37,6 @@ import com.thoughtworks.go.util.SystemEnvironment;
 import com.thoughtworks.go.websocket.*;
 import com.thoughtworks.go.work.SleepWork;
 import org.apache.http.client.HttpClient;
-import org.apache.http.client.methods.HttpPut;
-import org.eclipse.jetty.util.IO;
 import org.eclipse.jetty.websocket.api.RemoteEndpoint;
 import org.junit.After;
 import org.junit.Rule;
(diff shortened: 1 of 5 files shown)


agent/test/unit/com/thoughtworks/go/agent/AgentWebSocketClientControllerTest.java
```diff
@@ -106,6 +102,8 @@
     @After
     public void tearDown() {
         GuidService.deleteGuid();
+        System.clearProperty("go.agent.websocket.enabled");
```
arvindsv: Can you change this to something like:

```
System.clearProperty(SystemEnvironment.WEBSOCKET_ENABLED.propertyName());        
System.clearProperty(SystemEnvironment.CONSOLE_LOGS_THROUGH_WEBSOCKET_ENABLED.propertyName());
```

so that the keys are not duplicated?

kierarad:
@arvindsv that makes sense. I've updated the test

kierarad:
@arvindsv @jyotisingh just checking, is there anything else you'd like me to update on this in order for it to be merged?

arvindsv:
@kierarad I added [a commit](https://github.com/kierarad/gocd/commit/33afcdc9517d42e4b461ac257149ab4782f9e8cd) to remove a couple more duplicated keys. @jyotisingh I think you can take a look at this and merge it in, if it's ok.

kierarad:
@arvindsv oh thanks! I can't believe I missed that

PR #3240, opened 2017-03-08
PluginAuthenticationProvider finds or creates a user on successful authentication

bdpiparva:
* UserDao and UserService can findOrCreate users

#### Why findOrCreate over addUserIfDoesNotExist?

This handles scenarios where a user exists in DB with all the attributes defined, but on authentication the plugin returns User with only the username, e.g file based authentication.


maheshp:
This is not required as we will rely on the user details sent by the plugin. The idea is to create the user with the attributes sent by plugin.  Thereafter on successful authentication do not load the user from DB but rather use the User details given by plugin for the session.

Issue #3218, opened 2017-03-01
Emoji in console output rendered as ‘?’


mrmanc:
##### Issue Type

- Bug Report

##### Summary

In the past we were able to see emoji output properly in our console log. At some point one of the upgrades broke this, and now emoji output renders as question marks.

###### Basic environment details

* Go Version: `16.10.0 (4131-730ff1867576754414cc632957f344d0263bc06d)`
* JAVA Version: `1.8.0_102`
* OS: `Linux 2.6.32-358.el6.x86_64`
* Browser vendor and version (if relevant): `Chrome  56.0.2924.87 (64-bit)`

##### Steps to Reproduce

Create a pipeline with a job which runs a command like this: `

```
<job name="test-output">
            <tasks>
              <exec command="/bin/echo">
                <arg>-e</arg>
                <arg>"😍"</arg>
                <runif status="passed" />
              </exec>
            </tasks>
            <resources>
              <resource>bash</resource>
            </resources>
          </job>
```

##### Expected Results
A 😍 should appear on the console out

##### Actual Results
This appears:

`20:47:04.061 "?"`

Inspecting the page and altering the character in the dom to the correct emoji works, so this is not a typeface issue.

##### Possible Fix
Using the wrong character encodings?

arvindsv:
cc @marques-work and @ibnc - Thought this might interest you, since you're working on https://github.com/gocd/gocd/pull/3199.

marques-work:
Definitely interesting - but I'll probably make it a separate PR as the code for #3199 is ready (IMO) for merge now.

marques-work:
Definitely an encoding issue. I was able to reproduce this on MacOS X. This issue doesn't happen when logs are transmitted from agent -> server via WebSockets, which forces UTF-8, so the problem probably occurs in the RemoteAppender (or in its vicinity).

ibnc:
@marques-work The websocket shouldn't be forcing UTC-8 encoding. It'll use whatever the system default is on that machine (which I assume is UTC-8 on your mac)

marques-work:
Right - which is what I meant; MacOS X's default is UTF-8. Sorry that wasn't more clear. I've submitted #3241 so that RemoteConsoleAppender does the same thing.

mrmanc:
Thanks for such a quick fix! 😍

ankitsri11:
Go Version: 17.3.0 (4700-277f37e8b9775a203b73c2340f0cfb9e9f19c8be)
[...]

PR #3241, opened 2017-03-08
Send console lines using the default encoding for the platform instea…

marques-work:
…d of forcing ISO-8859-1

Should fix https://github.com/gocd/gocd/issues/3218 (Emoji in console output rendered as '?') on
platforms set to UTF-8.

### DIFF
--- a/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
+++ b/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 ThoughtWorks, Inc.
+ * Copyright 2017 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
+++ b/common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
@@ -23,6 +23,7 @@ import org.apache.http.entity.StringEntity;
 import org.apache.log4j.Logger;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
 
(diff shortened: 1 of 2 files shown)


common/src/com/thoughtworks/go/remote/work/RemoteConsoleAppender.java
```diff
@@ -42,8 +43,7 @@ public void append(String content) throws IOException {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug("Appending console to URL -> " + consoleUri);
             }
-
-            putMethod.setEntity(new StringEntity(content));
+            putMethod.setEntity(new StringEntity(content, Charset.defaultCharset()));
```
ketan: 👍

server/src/com/thoughtworks/go/server/service/ConsoleService.java
```diff
@@ -116,7 +117,7 @@ public boolean updateConsoleLog(File dest, InputStream in) throws IOException {
 
         LOGGER.trace("Updating console log [" + dest.getAbsolutePath() + "]");
         try (BufferedWriter writer = new BufferedWriter(new FileWriter(dest, dest.exists()))) {
-            IOUtils.copy(in, writer);
+            IOUtils.copy(in, writer, Charset.defaultCharset());
```
ketan: This assumes that `Charset.defaultCharset()` on agent and server is identical, which could have other issues when the encodings are not identical (or compatible)

Instead of using a `Writer` to write out the `InputStream`, it'd probably be better if we open `dest` as a (buffered) `FileOutputStream` and write to it instead, so as to do away with gotchas with encodings.


```java
try (OutputStream out = new BufferedOutputStream(new FileOutputStream(dest, dest.exists()))) {
  IOUtils.copy(in, out);
}
```

ketan:
LGTM —

screen shot 2017-03-09 at 4 29 02 pm
screen shot 2017-03-09 at 4 29 07 pm
screen shot 2017-03-09 at 4 29 12 pm

PR #3245, opened 2017-03-10
Use PluginInfos V3 on the elastic agent profiles page

ketan:
* Introduced a cleaner model to represent the plugin info objects
* PluginInfos is no longer a singleton. The expectation is that
  it is passed in from the top level all the way down to where it
  is usage is required

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/single_page_apps/elastic_profiles.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/single_page_apps/elastic_profiles.js
@@ -16,8 +16,9 @@
 
 var $                     = require('jquery');
 var m                     = require('mithril');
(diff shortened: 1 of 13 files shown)


server/webapp/WEB-INF/rails.new/webpack/single_page_apps/elastic_profiles.js
```diff
@@ -36,5 +46,5 @@ $(function () {
     );
   };
 
-  Promise.all([PluginInfos.init('elastic-agent')]).then(onSuccess, onFailure);
+  Promise.all([PluginInfos.all()]).then(onSuccess, onFailure);
```
GaneshSPatil: can we say `PluginInfos.all().then(onSuccess, onFailure)`?
no need of `Promise.all `

PR #3248, opened 2017-03-11
Remove unnecessary string conversion and the possibility of issues with encoding

ketan:


config/config-server/src/com/thoughtworks/go/config/MagicalGoConfigXmlWriter.java
```diff
@@ -104,7 +100,7 @@ public String toString(Document document) throws IOException {
     public void verifyXsdValid(Document document) throws Exception {
         ByteArrayOutputStream buffer = new ByteArrayOutputStream();
         XmlUtils.writeXml(document, buffer);
-        InputStream content = toInputStream(buffer.toString());
+        InputStream content = new ByteArrayInputStream(buffer.toByteArray());
```
ketan: I guess you're right about encoding.

PR #3249, opened 2017-03-11
Upgrade jetty from 9.2.18 to 9.4.6

ketan:


jetty9/src/com/thoughtworks/go/server/AssetsContextHandler.java
```diff
@@ -1,76 +0,0 @@
...
-import org.eclipse.jetty.server.handler.ResourceHandler;
-import org.eclipse.jetty.servlets.gzip.GzipHandler;
-import org.eclipse.jetty.webapp.WebAppContext;
-
-import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
-
-public class AssetsContextHandler extends ContextHandler {
```
jyotisingh: So, this was used to serve assets through Jetty instead of rails when run in production mode. Was removing this intentional? If so, then you need to handle asset serving in production mode.

jyotisingh:
Went through the jetty changes that have happened between 9.2.18 and 9.4.2 . While most of it looks good, there is one thing that might need a little more exploration:
* Only TLS 1.2 Supported [Link](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/VERSION.txt#L688)

There are some open issues on jetty too, still trying to see if they would apply to us or not:
* https://github.com/eclipse/jetty.project/issues/1196 (check [comment](https://github.com/eclipse/jetty.project/issues/1196#issuecomment-268641772)). 

Although, we do have a way of restricting ciphers enabled for Go server, we do not do so by default. This was done to make sure existing installations do not stop working. While we could take a call about enabling only certain ciphers by default, we need to check if those work as expected with multiple versions (not just the latest) of java 8 (openjdk & oracle) and compatibility with the supported browser versions too. Since we do not support java 7, a lot of these might not be an issue for us. But we still need to check for the rest. If you have already done that exercise, then it would be nice to put it up here for future reference.

ketan:
> * Only TLS 1.2 Supported [Link](https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/VERSION.txt#L688)

According to [this](https://www.eclipse.org/jetty/documentation/9.4.x/configuring-ssl.html), there's a way to override it:

> TLS v1.0, v1.1 and SSL v3 are no longer supported by default. If your Jetty implementation requires these protocols for legacy support, they can be enabled manually.

* https://github.com/eclipse/jetty.project/issues/1196 (check [comment](https://github.com/eclipse/jetty.project/issues/1196#issuecomment-268641772)). 

I'm unsure about this too...

ketan:
Closing in favor of #4630

PR #3253, opened 2017-03-13
Remove global resources/environments singleton used in agents SPA (#3187)

GaneshSPatil:


server/webapp/WEB-INF/rails.new/webpack/models/agents/environments.js
```diff
@@ -14,29 +14,34 @@
...
 
-Environments.init = (selectedAgents) => {
-  $.ajax({
-    method:     'GET',
-    url:        Routes.apiv1AdminInternalEnvironmentsPath(),
-    beforeSend: mrequest.xhrConfig.forVersion('v1')
-  }).then((data) => {
-    Environments.list = getSortedEnvironments(data, selectedAgents);
-  }).always(m.redraw);
+var Environments = function () {
```
ketan: You probably want to call `Environments.all()` to return a list of things. In this case, might as well be an array of (sorted) strings.

Additionally, this still has an issue in that you're not waiting on the ajax promise to complete,

server/webapp/WEB-INF/rails.new/webpack/models/agents/environments.js
```diff
@@ -14,29 +14,34 @@
...
+
+  this.initialize = (selectedAgents) => {
+    $.ajax({
+      method:     'GET',
+      url:        Routes.apiv1AdminInternalEnvironmentsPath(),
+      beforeSend: mrequest.xhrConfig.forVersion('v1')
+    }).then((data) => {
+      self.list(getSortedEnvironments(data, selectedAgents));
+    }).always(m.redraw);
+  };
```
ketan: Return a promise instead, and wait on it?

server/webapp/WEB-INF/rails.new/webpack/models/agents/resources.js
```diff
@@ -14,29 +14,33 @@
...
+
+  this.initialize = (selectedAgents) => {
+    $.ajax({
+      method:     'GET',
+      url:        Routes.apiv1AdminInternalResourcesPath(),
+      beforeSend: mrequest.xhrConfig.forVersion('v1')
+    }).then((data) => {
+      self.list(getSortedResources(data, selectedAgents));
+    }).always(m.redraw);
+  };
```
ketan: Same as above?

PR #3257, opened 2017-03-13
Remove unwanted dependencies

ketan:


ketan:
@bdpiparva build is green, mind taking a look?

ketan:
@arvindsv @jyotisingh I remember being part of conversations where this was particularly written for luau, however it was unused outside of the now dead luau integration.

PR #3259, opened 2017-03-14
Ensure that logs are written to a different directory

ketan:


jyotisingh:
There are a few users who periodically backup/archive the logs. Their scripts will break after this change. If we are going ahead with this, then I think we should mark it as a breaking change in the release notes.

On a side note, why do this only for linux based installations? Why not windows for instance which logs inside the installation folder too?

ketan:
> On a side note, why do this only for linux based installations? Why not windows for instance which logs inside the installation folder too?

You're right. I'm being short-sighted here. Closing PR and will submit another PR to cleanup logging...

PR #3261, opened 2017-03-14
Providing support for Web based authentication

maheshp:
* Authorization plugins have endpoints to handle web-based
  authentication
* WebBasedAuthFilter responsible to get the identity provider url and redirect
* PreAuthFilter handles requests for pre-authenticated requests by
  interacting with authorization plugin
* PluginInteraction callbacks handles request headers
* ToDo - Integrate the above filters to filter chain

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
@@ -197,9 +197,10 @@ public class AuthorizationMessageConverterV1 implements AuthorizationMessageConv
     }
 
     @Override
-    public String identityProviderUrlRequestBody(List<SecurityAuthConfig> authConfigs) {
+    public String authorizationServerRedirectUrlRequestBody(String pluginId, List<SecurityAuthConfig> authConfigs, String siteUrl) {
         Map<String, Object> requestMap = new HashMap<>();
         requestMap.put("auth_configs", getAuthConfigs(authConfigs));
+        requestMap.put("authorization_server_callback_url", authorizationServerCallbackUrl(pluginId, siteUrl));
 
         return GSON.toJson(requestMap);
     }
@@ -212,4 +213,7 @@ public class AuthorizationMessageConverterV1 implements AuthorizationMessageConv
         return template;
     }
 
+    private String authorizationServerCallbackUrl(String pluginId, String siteUrl) {
+        return String.format("%s/go/plugin/%s/authenticate", siteUrl, pluginId);
+    }
 }
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
@@ -17,6 +17,7 @@
 package com.thoughtworks.go.plugin.access.authorization;
 
 import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import com.thoughtworks.go.config.PluginRoleConfig;
 import com.thoughtworks.go.config.SecurityAuthConfig;
 import com.thoughtworks.go.plugin.access.authentication.models.User;
@@ -31,6 +32,7 @@ import com.thoughtworks.go.plugin.domain.common.VerifyConnectionResponse;
 import org.apache.commons.collections.map.HashedMap;
 import org.apache.commons.lang.StringUtils;
 
+import java.lang.reflect.Type;
 import java.util.*;
 
 public class AuthorizationMessageConverterV1 implements AuthorizationMessageConverter {
@@ -152,11 +154,6 @@ public class AuthorizationMessageConverterV1 implements AuthorizationMessageConv
     }
 
     @Override
-    public String processGetRoleConfigsRequest(String requestBody) {
-        return (String) GSON.fromJson(requestBody, Map.class).get("auth_config_id");
-    }
-
-    @Override
     public String getProcessRoleConfigsResponseBody(List<PluginRoleConfig> roles) {
         List<Map> list = new ArrayList<>();
         for (PluginRoleConfig role : roles) {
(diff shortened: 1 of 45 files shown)


jyotisingh:
Few thoughts - some of it applies to auth extension in general not just this PR:

* Starting notification API, we have started explicitly creating version numbered packages for extensions to keep all contract classes in one place for a given version. We need to do the same for Auth Extension too. `Capabilities` is acting as a contract between server and plugin. Its the same object which is being passed along to the upper layers. The issue with this would come when we need to version the auth extension. Since the contract is being spilled over to other layers, it will make creation of new versions difficult, or even an accidental change in one of these classes will end up breaking the existing contract. Creating versioned packages such as the ones done for notification extensions make it a little clear about what classes are part of a contract, and what isn't. Hence making writing new versions easier. Ditto for `AuthenticationResponse`, `SupportedAuthType`, `PluginProfileMetadataKey` etc. 

Plugin Development issues:

* Live reload of invalid plugins fail. icon wasn't implemented. implemented it and dropped new jar inside server. Live reload fails as the updated call first tries to unload the plugin, but the plugin wasn't loaded to begin with and hence exception is thrown, causing the load to not happen at all - I will log a separate issue for this.

* Request name `go.cd.authorization.identity-provider-access` was confusing to me, I couldn't figure out what was i expected to do as a part of this call. Its only upon looking at the internal implementation in gocd and speaking to you, I could figure out that this call was supposed to return the authorization access code from the third party system which would be sent back to the system handles authentication (in my case github).

* Request for `go.cd.authorization.identity-provider-url` is supposed to return the url that Go will redirect to for the authentication to take place. In my case (github), the url looked like `https://github.com/login/oauth/authorize?client_id=<github_client_id>&response_type=code&redirect_uri=<gocd_plugin_auth_url>&scope=read:org` . The url here includes the redirect-uri which is where github would redirect to on successful authentication. I assume this is going to be the case with all oauth based plugins if not all web-based ones. 
Go expects the url to be something like `https://localhost:8154/go/plugin/cd.go.authorization.githubtest/authenticate` where `cd.go.authorization.githubtest` is the plugin-id. The url should now be hardcoded in all oauth based plugins or setup as a configuration in Go's config. Either of them does not seem right. The other bit being the domain itself. People using a proxy might want to specify the corresponding url for their setup. Would it make sense to allow users to configure the domain part in Go (may be use siteurl/secureSiteUrl) and pass along the path (ie. `go/plugin/cd.go.authorization.githubtest/authenticate`) from go-server to the plugins? With this, in future if we wanted to change the url to something else, we could do that without worrying about breaking all existing plugins.

* During one of the many runs of login attempts using the plugin I wrote, I was locked out of Go. Meaning unable to login using regular password file based auth too. Login from chrome wasn't working, but this worked from incognito window in chrome and firefox. something wrong with cookies? I did not get a chance to debug this further, but there was some exception thrown by the plugin when PreAuthenticatedAuthenticationProvider made a call to it.

* confusion about which roles to be returned as a part of [authenticate-user](https://plugin-api.gocd.io/17.3.0/authorization/#authenticate-user) call. I assumed that github roles were to be returned, but it was supposed to be the mapped ones in Go. May be we could make it clear in the documentation

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
```diff
@@ -168,6 +165,46 @@ public String getProcessRoleConfigsResponseBody(List<PluginRoleConfig> roles) {
...
+    public String authenticateUserRequestBody(Map<String, String> credentials, List<SecurityAuthConfig> authConfigs, List<PluginRoleConfig> roleConfigs) {
+        Map<String, Object> requestMap = new HashMap<>();
+
+        requestMap.put("credentials", credentials);
+        requestMap.put("auth_configs", getAuthConfigs(authConfigs));
+        requestMap.put("role_configs", getRoleConfigs(roleConfigs));
+        return GSON.toJson(requestMap);
+    }
+
+    @Override
```
jyotisingh: Rename method to `authorizationServerRedirectUrl` and contract to `authorization_server_redirect_url` instead of `identity_provider_url `?

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverterV1.java
```diff
@@ -168,6 +165,46 @@ public String getProcessRoleConfigsResponseBody(List<PluginRoleConfig> roles) {
...
+    public String getIdentityProviderUrl(String responseBody) {
+        return (String) new Gson().fromJson(responseBody, Map.class).get("identity_provider_url");
+    }
+
+    @Override
+    public String authorizationServerRedirectUrlRequestBody(String pluginId, List<SecurityAuthConfig> authConfigs, String siteUrl) {
+        Map<String, Object> requestMap = new HashMap<>();
+        requestMap.put("auth_configs", getAuthConfigs(authConfigs));
+        requestMap.put("authorization_server_callback_url", authorizationServerCallbackUrl(pluginId, siteUrl));
```
jyotisingh: This is the Go server url, so shouldn't we call it 'gocdServerAuthorizationCallbackUrl' instead of 'authorizationServerCallbackUrl'?

jyotisingh:
Found two issues:
* Login throws a 404. Had to add `setAlwaysUseDefaultTargetUrl(true);` [here](https://github.com/maheshp/gocd/blob/d6c9932705a05984d94b118dbbcab84fc6e6cc6b/server/src/com/thoughtworks/go/server/security/PreAuthenticatedRequestsProcessingFilter.java#L47) for it to work.

* The authorities are being granted appropriately for the user, but somehow admin pages are forbidden for a user with admin role. Home page has the admin menu enabled and the services respond correctly for [isUserAdmin](https://github.com/gocd/gocd/blob/b056800839ad8e2dcf706e414e620800562e7a5e/server/src/com/thoughtworks/go/server/service/GoConfigService.java#L896) calls. 
I could make it work by changing [this](https://github.com/gocd/gocd/blob/b056800839ad8e2dcf706e414e620800562e7a5e/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L281) line from `/admin/**=ROLE_SUPERVISOR` to `/admin/**=ROLE_USER`, but that would be wrong. I did not debug this further.

Setup: Tested this with a self-written github auth plugin. Config in Go:
```
    <security>
      <passwordFile path="../manual-testing/ant_hg/password.properties" />
      <authConfigs>
        <authConfig id="github" pluginId="cd.go.authorization.skeleton" />
      </authConfigs>
      <roles>
        <pluginRole name="go_admins" authConfigId="github">
          <property>
            <key>organization</key>
            <value>gocd</value>
          </property>
          <property>
            <key>team</key>
            <value>Committers</value>
          </property>
        </pluginRole>
      </roles>
      <admins>
        <user>admin</user>
        <role>go_admins</role>
      </admins>
    </security>
```

jyotisingh:
Pending documentation:
* plugin-api.gocd.io needs to be updated with the updated contract changes that were made as a part of this PR

plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMetadataStore.java
```diff
@@ -56,4 +56,8 @@ public static AuthorizationMetadataStore instance() {
         }
         return plugins;
     }
+
+    public Set<String> getPluginsThatSupportsWebBasedAuthentication() {
```
maheshp: Should we consider returning the PluginInfo objects instead of just the set of pluginIds

server/webapp/WEB-INF/vm/auth/login.vm
```diff
@@ -155,11 +155,19 @@
...
-            #foreach ($authentication_plugin_id in $authentication_plugin_registry.getPluginsThatSupportsWebBasedAuthentication())
-              <a href="/go/plugin/interact/$authentication_plugin_id/index"><img
-                  src="$authentication_plugin_registry.getDisplayImageURLFor($authentication_plugin_id)"
-                  alt="$authentication_plugin_registry.getDisplayNameFor($authentication_plugin_id)"/></a>
-            #end
+            <div class="authorization-plugins">
+              #foreach ($model in $security_auth_config_service.getAllConfiguredWebBasedAuthorizationPlugins())
+                  <a href="/go/plugin/$model.pluginId()/login"><img class="authorization-plugin-icon" src="$model.imageUrl()" alt="$model.name()"/></a>
+              #end
+              #if($security_auth_config_service.getAllConfiguredWebBasedAuthorizationPlugins().size() > 0 && $authentication_plugin_registry.getPluginsThatSupportsWebBasedAuthentication().size() > 0)
```
maheshp: Calling getAllConfiguredWebBasedAuthorizationPlugins second time would build the list twice

plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationMetadataStore.java
jyotisingh: I thought about it, but since `getPluginsThatSupports` returned a `Set<String>`, I left it as it. But sure this could be changed

Issue #3262, opened 2017-03-14
Improve logging


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Identified several issues with agent-bootstrapper, agent-launcher and agent logging —

* no documented way to configure logging
* a lot of setup is "programatic", with little ways to configure it
* inconsistent how logging is configured across the 3 systems

After some thought here's what I think will work well —

There will be 2 files —

* `config/agent-bootstrapper-log4j.properties`
* `config/agent-log4j.properties`

If the file is present, the bootstrapper and agent will load it up, if not they will gracefully fallback to loading the same files from the classloader, and print a message (to stderr?) that indicates how logging may be configured.

Since the agent-launcher works in the same process as the bootstrapper, we need to ensure that `config/agent-launcher-log4j.properties` does not mess up the log4j configuration via either

* classloader isolation (already in place), might mean multiple copies of logging jars [NOT DONE]
* threadgroup isolation

jyotisingh:
Will this also take care of the configuring the logs directory for the agent (bootstrapper+launcher+real_agent) for all supported platforms?

ketan:
Yes, I'm hoping to be able to configure it via `-Dgocd.agent.log.dir=...` and set it in the appropriate log4j file —

```properties
log4j.appender.FOO.File=${gocd.agent.log.dir}/go-agent.log
```

varshavaradarajan:
Multiple agents on the same machine ended up logging to /var/log/go-agent/go-agent.log, I think, because somewhere we are hardcoding the location to be /var/log.. mostly in agent.sh. That should also be taken care of, I think. This comes into picture for deb and rpm installations and not for zip.

jyotisingh:
PR - https://github.com/gocd/gocd/pull/3267

PR #3267, opened 2017-03-15
Improvements to logging (#3262)

ketan:


varshavaradarajan:
- stop-agent.sh - Change the LOG_FILE path to `$AGENT_WORK_DIR/logs/agent-bootstrapper.log`
- the path for `autoregister.properties`, `guid.txt` etc is hardcoded to be `config` directory. Make the `config` directory configurable on the agent side as well. This will involve change agent.sh to include the system property `cruise.config.dir` as a startup argument
- the `go-agent-stderr.log` and `go-agent-stdout.log` should be moved to the logs directory.

agent/properties/MANIFEST.MF
```diff
@@ -1,3 +0,0 @@
-Class-Path: lib/ant-launcher.jar config/
```
varshavaradarajan: Any reason for this file to be removed. I started the Development Server and tried to access the agents page. Got the exception `Could not load compiled manifest from /Users/varshavs/gocd/server/webapp/WEB-INF/rails.new/public/assets/webpack/manifest.json - have you run `rake webpack:compile`? (original error No such file or directory - /Users/varshavs/gocd/server/webapp/WEB-INF/rails.new/public/assets/webpack/manifest.json)`

jyotisingh:
Did not test it by bringing up an agent, but the code looks good to me..about the deletion of agent manifest file have you checked if the packaged agent MANIFEST file continues to include all the required info. This is what I see in the current version:

```
Manifest-Version: 1.0
Go-Version: 17.4.0-4712
Implementation-Title: agent
Target-Compatibility: 1.8
Implementation-Version: 17.4.0-4712
Source-Compatibility: 1.8
Go-Agent-Bootstrap-Class: com.thoughtworks.go.agent.AgentProcessParent
 Impl
Go-Revision: c943bb57d7d7d18c1d3cf9a9bd3385d56edbdd37
ImplementationVersion: 17.4.0-4712
Main-Class: com.simontuffs.onejar.Boot
One-Jar-Main-Class: com.thoughtworks.go.agent.AgentMain
```

ketan:
> have you checked if the packaged agent MANIFEST file continues to include all the required info

All jars are packaged via the `Jar` gradle task, and `MANIFEST.MF` content defined in the task itself. None of them slurp in configuration from a `MANIFEST.MF` file.

jyotisingh:
Then I guess its all good.

jyotisingh:
the only thing is build hasn't run, so may be you would want to see that through, esp. the functional tests

development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
```diff
@@ -39,8 +40,9 @@
 
 public class DevelopmentServer {
     public static void main(String[] args) throws Exception {
+        LogConfigurator logConfigurator = new LogConfigurator(DEFAULT_LOG4J_CONFIGURATION_FILE);
+        logConfigurator.initialize();
```
varshavaradarajan: No need for this. `LoggingInitializer` will take care of it. The statement `Could not find file 'config/log4j.properties'. Attempting to load from classpath.` appears twice otherwise

installers/go-agent/release/agent.sh
```diff
@@ -115,12 +115,14 @@ if [ "$PRODUCTION_MODE" == "Y" ]; then
...
+	    LOG_DIR=$AGENT_WORK_DIR/logs
     fi
 else
-    LOG_DIR=$AGENT_WORK_DIR
+    LOG_DIR=$AGENT_WORK_DIR/logs
 fi
 
+mkdir -p "${LOG_DIR}"
+
 STDOUT_LOG_FILE=$LOG_DIR/${SERVICE_NAME}-bootstrapper.out.log
```
varshavaradarajan: There is a `LOG_FILE` variable that is exported unnecessarily and stop-agent.sh also sets it unnecessarily. I did a `git grep -w "LOG_FILE"` in the project and found no usages in the code, except for `JavaApplicationStub64`. I think we can remove it.

development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
ketan: This happens because the main program uses a different classloader than the webapp. So the logger is actually initialized twice. Side-effect of how logging is setup, unfortunately nothing (simple) that can be done about it.

agent/build.gradle
```diff
@@ -55,7 +55,6 @@ clean.doFirst {
   delete 'agent-plugins.zip'
   delete 'tfs-impl.jar'
   delete 'go-agent.log'
```
varshavaradarajan: Doesn't this have to be logs directory now?

varshavaradarajan:
@ketan, https://github.com/gocd/gocd/blob/master/agent-launcher/test/com/thoughtworks/go/agent/launcher/AgentLauncherImplTest.java#L64 needs to change lest it leaves behind a the logs directory during cleanup?

PR #3276, opened 2017-03-16
Add icon for elastic agents on agent screen

kyleolivo:
Elastic agents provide an icon related to the type of agent in use.
This change just adds the icon to the beginning of the agent listing
on the agents page.

screen shot 2017-03-16 at 3 23 35 pm

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
@@ -62,16 +62,20 @@ const AgentRowWidget = {
     let agentStatus       = agent.status();
     const isBuildingAgent = !agent.buildDetails().isEmpty();
     const isElasticAgent  = agent.isElasticAgent();
+    let pluginInfos = args.pluginInfos;
     let selectAgentCheckbox;
+    let elasticAgentIcon;
     let hostnameLink      = <span>{agent.hostname()}</span>;
-
-    if (isElasticAgent) {
-      agent.fetchElasticAgentIcon();
-    }
 
     if (isBuildingAgent) {
       agentStatus = <f.link onclick={ctrl.buildDetailsClicked.bind(ctrl, agent.uuid())}
                             class="has-build-details-drop-down">{agent.status()}</f.link>;
+    }
+
+    if (isElasticAgent && pluginInfos()) {
+      let elasticAgentIconURL = pluginInfos().findById(agent.elasticPluginId()).imageUrl;
+      elasticAgentIcon =
+        <img class="agent-icon" src={elasticAgentIconURL} />;
     }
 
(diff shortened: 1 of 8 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
```diff
@@ -61,9 +61,14 @@ const AgentRowWidget = {
     const environments    = joinOrNoneSpecified(agent.environments());
     let agentStatus       = agent.status();
     const isBuildingAgent = !agent.buildDetails().isEmpty();
+    const isElasticAgent  = agent.isElasticAgent();
     let selectAgentCheckbox;
     let hostnameLink      = <span>{agent.hostname()}</span>;
 
+    if (isElasticAgent) {
+      agent.fetchElasticAgentIcon();
```
ketan: This could be dangerous with hundreds of agents being polled every 10 seconds. For such things, it's better to pull down the plugin infos object via `PluginInfos.all()` in `single_page_apps/agents.js`. Sort of like we do it [here](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/webpack/single_page_apps/elastic_profiles.js#L29-L48)

This plugin infos object can be passed onto this same view where you can decide which icon you want to render.

server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
```diff
@@ -86,6 +91,9 @@ const AgentRowWidget = {
         <td key="checkbox">
           {selectAgentCheckbox}
         </td>
+        <td key="agentIcon">
+          <img class='agent-icon' src={ isElasticAgent ? agent.elasticAgentIcon() : ''} />
```
ketan: Prefer to not render the image instead of an image with no src on it. See how `selectAgentCheckbox` is rendered for admins only.

This would require removing the CSS rule for empty `src` attribute.

ketan: Prefer to not render the image instead of an image with no src on it. See how `selectAgentCheckbox` is rendered for admins only.

server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
ketan: This has another drawback that a redraw (because of a user action) before the ajax call finishes will cause a second ajax call possibly at a 60hz frequency and totally overwhelm the server.

kyleolivo: Totally makes sense. Thanks for catching this.

PR #3277, opened 2017-03-16
Rename msx files to jsx

kyleolivo:
MSX is deprecated and has already been removed from the system.
This just renames the msx files to jsx and updates the webpack config
accordingly.


ketan:
`.jsx` files (under `app/assets/new_javascripts`) are not transformed by the `mithril_rails` plugin.

Also me and @GaneshSPatil have got other WIP branches where we're moving some of these pages to mithril 1.x. So I can't merge this in and cause loads of conflicts later. Sorry :(

PR #3278, opened 2017-03-17
Show page spinner on resources and environments dropdown until the ajax finishes (#3187)

GaneshSPatil:
screen shot 2017-03-17 at 10 41 32 am
screen shot 2017-03-17 at 10 42 09 am

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/agents/environments_list_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/agents/environments_list_widget.js.msx
@@ -47,11 +47,12 @@ const EnvironmentsListWidget = {
   },
 
   view (vnode) {
-    const ctrl = vnode.state;
-    const args = vnode.attrs;
+    const ctrl                   = vnode.state;
+    const args                   = vnode.attrs;
+    let environmentDropDownClass = "agent-button-group-dropdown env-dropdown ";
 
     const spinner = (
-      <div class="agent-button-group-dropdown env-dropdown empty-container" onclick={ctrl.closeDropdown}>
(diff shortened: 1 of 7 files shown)


ketan:
Looks good, can we handle errors in case of ajax failures (or timeouts)?

server/webapp/WEB-INF/rails.new/webpack/views/agents/environments_list_widget.js.msx
```diff
@@ -47,10 +47,19 @@ const EnvironmentsListWidget = {
   },
 
   view (vnode) {
-
     const ctrl = vnode.state;
     const args = vnode.attrs;
 
+    const spinner = (
+      <div class="agent-button-group-dropdown env-dropdown empty-container" onclick={ctrl.closeDropdown}>
```
ketan: These are the same classes as the ones defined in `environmentDropDownClass`. Dedup it? `class` attribute accepts arrays (even with null and undefined objects), and they usually work better with optional class names.

server/webapp/WEB-INF/rails.new/webpack/views/agents/resources_list_widget.js.msx
```diff
@@ -76,8 +76,18 @@ const ResourcesListWidget = {
   },
 
   view (vnode) {
+    const spinner = (
+      <div class="agent-button-group-dropdown resource-dropdown empty-container" onclick={vnode.state.closeDropdown}>
```
ketan: Same comment as above.

GaneshSPatil:
screen shot 2017-03-29 at 11 07 02 am
screen shot 2017-03-29 at 11 08 07 am

PR #3281, opened 2017-03-17
Allow pipeline group admins to create templates and be made template admins on creation

varshavaradarajan:
#3081

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
@@ -88,6 +88,9 @@ public class BasicCruiseConfig implements CruiseConfig {
     @IgnoreTraversal
     private List<PartialConfig> partials = new ArrayList<>();
 
+    @IgnoreTraversal
+    private transient Map<CaseInsensitiveString, Map<CaseInsensitiveString, Authorization>> allTemplatesWithAssociatedPipelines;
+
     public BasicCruiseConfig() {
         strategy = new BasicStrategy();
     }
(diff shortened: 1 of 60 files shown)


config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -1308,26 +1309,39 @@ public PipelineConfig findPipelineUsingThisPipelineAsADependency(String pipeline
     }
 
     @Override
-    public Map<CaseInsensitiveString, List<CaseInsensitiveString>> templatesWithPipelinesForUser(String username, List<Role> roles) {
-        HashMap<CaseInsensitiveString, List<CaseInsensitiveString>> templateToPipelines = new HashMap<>();
-        for (PipelineTemplateConfig template : getTemplates()) {
-            if (isAuthorizedToAccessTemplate(new CaseInsensitiveString(username), roles, template)) {
-                templateToPipelines.put(template.name(), new ArrayList<>());
+    public List<TemplatesToPipelines> templatesWithPipelinesForUser(String username, List<Role> roles) {
```
jyotisingh: `username` is being passed along as string everywhere and then being converted to `CaseInsensitiveString` over here. But if you check all the callers, they already have the `CaseInsensitiveString` object for username, might as well pass that one. Given this is used by the service alone, and that the logic doesn't belong to the domain, would it make sense to move it to the service?

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -1308,26 +1309,39 @@ public PipelineConfig findPipelineUsingThisPipelineAsADependency(String pipeline
...
+            if (isAuthorizedToAccessTemplate(new CaseInsensitiveString(username), roles, templateConfig)) {
+                templatesToPipelines.add(new TemplatesToPipelines(templateConfig.name()));
             }
         }
+
         for (PipelineConfig pipelineConfig : getAllPipelineConfigs()) {
             CaseInsensitiveString name = pipelineConfig.getTemplateName();
-            if (pipelineConfig.hasTemplate() && templateToPipelines.containsKey(name)) {
-                templateToPipelines.get(name).add(pipelineConfig.name());
+            TemplatesToPipelines template = findTemplateToPipelines(templatesToPipelines, name);
```
jyotisingh: this will become too expensive for larger configs, switch back to HashMap?

config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
jyotisingh: This entire method has 4 levels of nested iterations. Moreover, this is called once for every user. consider moving it elsewhere and may be caching some of this.

jyotisingh:
The APIs and old UI seem to be taken care of. We need to create an issue for to handle the UI restrictions in the new Template and Pipeline SPAs too. @varshavaradarajan - could you create one and add it to 17.4 for now?

jyotisingh:
Did some sanity testing around this, works as expected. If a logged-in regular user is made a template-view user and the person navigated to the templates-index page, a 403 is thrown for them. Logging out and then logging in works.

jyotisingh:
@varshavaradarajan - Build seems to be failing: https://build.gocd.io/go/pipelines/build-linux-PR/1798/build-server/1

varshavaradarajan:
https://github.com/gocd/functional-tests/pull/190

PR #3282, opened 2017-03-17
Use WebSocket() to tail build logs from the browser

marques-work:
Introduce a JSR-356 ServerEndpoint for console log tailing

CAVEAT: the effect of this change will be more appreciated with the agent -> server
console log websocket transmission feature enabled (i.e. `-Dgo.agent.websocket.enabled=true`).

Also replaces ConsoleOut with ConsoleStreamer; this means we no longer read the entire file content into memory before sending, but rather stream lines out through the websocket connection.


server/src/com/thoughtworks/go/server/websocket/ConsoleLogEndpoint.java
```diff
@@ -0,0 +1,87 @@
...
+        try {
+            session.close();
+        } catch (IOException e) {
+            LOGGER.warn("{} failed to close session.", this);
+        }
+    }
+
+    @OnError
+    public void onError(Session session, Throwable error) {
+        LOGGER.error(toString() + " closing session because an error was thrown", error);
```
ketan: Avoid string concatenation? `LOGGER.error("{} closing ...", this, error)`

server/webapp/WEB-INF/vm/build_detail/build_detail_page.vm
```diff
@@ -78,7 +78,7 @@
               <div class="sidebar_history">
                   #parse("sidebar/_sidebar_build_list.vm")
               </div>
-              <div class="job_details_content">
+              <div class="job_details_content" data-pipeline="${presenter.pipelineName}" data-pipeline-label="${presenter.pipelineLabel}" data-stage="${presenter.stageName}" data-stage-counter="${presenter.stageCounter}" data-job="${presenter.id}" data-build="${presenter.buildName}">
```
ketan: Directly embed the URL in a data attr instead of individual components?

server/src/com/thoughtworks/go/server/controller/ArtifactsController.java
```diff
@@ -239,28 +236,6 @@ public ModelAndView putArtifact(@RequestParam("pipelineName") String pipelineNam
 
     /* Other URLs */
 
-    @RequestMapping(value = "/**/consoleout.json", method = RequestMethod.GET)
-    public ModelAndView consoleout(@RequestParam("pipelineName") String pipelineName,
-                                   @RequestParam("pipelineLabel") String counterOrLabel,
-                                   @RequestParam("stageName") String stageName,
-                                   @RequestParam("buildName") String buildName,
-                                   @RequestParam(value = "stageCounter", required = false) String stageCounter,
-                                   @RequestParam(value = "startLineNumber", required = false) Integer start
```
arvindsv: I think it's just an internal redirect from [here](https://github.com/gocd/gocd/blob/5478a5460cd4fcb1a7380eb0e4ab3fbbeacb4166/server/webapp/WEB-INF/urlrewrite.xml#L202). So, it's past the acegi bit.

server/src/com/thoughtworks/go/server/websocket/ConsoleLogEndpoint.java
```diff
@@ -0,0 +1,87 @@
...
+                       @PathParam(value = "stageCounter") String stageCounter,
+                       @PathParam(value = "jobName") String jobName) throws Exception {
+        this.session = session;
+        LOGGER.debug("{} connected.", this);
+
+        Long start = (Long) config.getUserProperties().get("startLine");
+        ConsoleLogSender sender = (ConsoleLogSender) config.getUserProperties().get("sender");
+        JobIdentifier jobIdentifier = getJobIdentifier(config, pipelineName, pipelineLabel, stageName, stageCounter, jobName);
+
+        LOGGER.info("Sending logs for {} starting at line {}", jobIdentifier, start);
```
marques-work: Oh I missed that one. I meant info just for my dev server.

common/src/com/thoughtworks/go/domain/ConsoleStreamer.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Iterator;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+/**
+ * Created by marqueslee on 3/11/17.
```
marques-work: I'll remove that. IDEA generates that for me and I never pay attention to it. I'll prob replace it with a real javadoc

server/src/com/thoughtworks/go/server/websocket/ConsoleLogEndpoint.java
marques-work: Sure, I think I wasn't sure whether or not it would treat the args like it does for debug and info; I didn't want the the throwable to be just thr.toString(), and I saw that there's a signature for `Logger.error(String message, Throwable throwable)` which I know does the right thing. If Logger.error(String message, Object... args) knows how to chop that last arg as a throwable and append the stacktrace instead, I'll change it.

server/src/com/thoughtworks/go/server/websocket/ConsoleLogSender.java
```diff
@@ -0,0 +1,109 @@
...
+
+        console.stream(new Consumer<String>() {
+            @Override
+            public void accept(String line) {
+                try {
+                    if (buffer.size() >= BUFFER_SIZE) {
+                        flushBuffer(buffer, webSocket);
+                    } else {
+                        buffer.add(line);
+                    }
```
marques-work: We buffer up to 1000 or end of stream, and then we flush that afterward too. I wanted to avoid sending too many messages (i.e. one per line) at once (and avoid layout thrashing).

server/webapp/WEB-INF/vm/build_detail/build_detail_page.vm
marques-work: I could do that, but in PR 3199, I split out the split out the build observer into three observers as it was really doing 3 different things.

Those fragments are used in different capacities; TabsManager uses the components to build a UID, console log log uses it for building the status URL (which will become the websocket URL if both PRs merge), and Timer uses one component as well. It was more just planning ahead.

marques-work: I just realized I repeated "split out the" twice in a row. Brain fart :)

server/build.gradle
```diff
@@ -190,6 +190,7 @@ dependencies {
   compileOnly project(':app-server')
   compileOnly group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: '1.47'
   compileOnly group: 'org.eclipse.jetty.websocket', name: 'websocket-server', version: versions.jetty
+  compile group: 'org.eclipse.jetty.websocket', name: 'javax-websocket-server-impl', version: versions.jetty
```
ketan: This should probably be `compileOnly` to avoid pulling down all of jetty. And move these dependencies in `jetty9` for packaging in prod, and `development-server` for the dev mode.

If you look for `org.eclipse.jetty.websocket` in all gradle files, you'll see what needs to be done.

marques-work: @ketan just to confirm, the websocket jars (jetty's impl + jsr-356 api) only need to be in the `jetty9` and `development-server` modules? These do not need to be in `server`?

marques-work: @ketan nevermind, I just tried: gradle will not compile without these declared in `server`.

marques-work:
This isn't quite ready yet - smoke tests indicate packaging issues preventing start up. Will continue to work on this.

marques-work:
This would be ready to merge, but it would be easier (for me anyway) to merge this after https://github.com/gocd/gocd/pull/3199

jyotisingh:
@marques-work - #3199 is merged :) 
Rebase this one?

marques-work:
@jyotisingh yep, just finished rebasing/fixing this :). Sending through pipelines...

marques-work:
@jyotisingh have any time to review this PR?

ketan:
@marques-work — I've just rebased and force pushed this PR on top of current master to start a review. If you're pushing any new commits, be sure to reset your branch before anything.

marques-work:
@ketan thanks for the heads up - I hadn't planned on any further commits to this PR, but I'll reset now just in case.

ketan:
I'm going to create a checklist in this comment to keep track of things that need fixing, or improvements.

When I add a new item to the list, I'll add a new comment so you know something needs addressing.

marques-work:
@ketan sounds good :)

ketan:
### Unnecessary polling when job is not assigned to agent

When job is not yet assigned to an agent (for an agent not being available) the websocket is immediately disconnected, and retried in 500ms.

This causes unnecessary connections on the server. Alternatives are to either:
- leave the websocket connection open (indefinitely)
- piggyback on the existing `jobStatus.json` ajax call to determine if the websocket connection should be initialized. Check `building_info.current_status` attribute in the JSON response.

ketan:
### Having to set a very very long `proxy_(read|send)_timeout` on NGINX

I'm using the NGINX configuration below (`docker run -it --rm -p80:80 -v$(pwd)/default.conf:/etc/nginx/conf.d/default.conf nginx:alpine`)

This will run NGINX on port 80, to simulate a similar situation with actual end users.

Once the NGINX configuration is setup, I'm running the following job:

```xml
<exec command="bash" args="-c 'while true; do date; sleep 30; done'" />
```

With this combination, the job generates output at an interval of 30s. The `proxy_(read|write)_timeout` is set to 10s. This will cause NGINX to disconnect the connection to GoCD since there's no output for 30 seconds between the `sleep`s.

Having to set a very long read/write timeout may be undesirable when the GoCD's threadpool is under stress and NGINX won't be able to establish any more connections.

Can we instead send a ping frame every say 10 seconds (configurable via `SystemEnvironment`)? Might mean having to have an `ScheduledExecutorService.scheduleAtFixedRate()` to do that. See `spring-cruise-remoting-servlet.xml` on how you might define lifecycle of timers.

```nginx
# default.conf

map $http_upgrade $connection_upgrade {
    default upgrade;
    ''      close;
}

server {
    listen       80;
    server_name  localhost;

    keepalive_timeout   65;
    gzip                on;
    gzip_comp_level     6;
    gzip_types          text/plain text/css text/javascript application/x-javascript application/javascript application/json text/json text/xml application/xml;

    location / {
        # Change the IP to your needs
        proxy_pass              http://192.168.0.2:8153;
        proxy_set_header        Host            $host;
        proxy_set_header        X-Real-IP       $remote_addr;
        proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header        X-Forwarded-Proto $scheme;
        proxy_connect_timeout   150;
        proxy_send_timeout      100;
        proxy_read_timeout      100;

        proxy_buffering         on;
        proxy_buffer_size       128k;
        proxy_buffers           100  128k;
    }

    location /go/client-websocket/ {
      proxy_pass              http://192.168.0.2:8153;
      proxy_set_header        Host            $host;
      proxy_set_header        X-Real-IP       $remote_addr;
      proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header        X-Forwarded-Proto $scheme;
      proxy_connect_timeout   150;
      proxy_send_timeout      10;
      proxy_read_timeout      10;

      # this is intentionally turned off, or websocket connections will be buffered, leading to unexpected behavior
      proxy_buffering         off;

      # begin ugprade to websocket if requestsd
      proxy_http_version      1.1;
      proxy_set_header        Upgrade $http_upgrade;
      proxy_set_header        Connection $connection_upgrade;
      # end ugprade to websocket if requested
    }
}

```

ketan:
### Possibility send more info "other stuff" with the same connection

This does not need to be addressed as part of the PR, but since you've worked extensively around this area — how do we go about eliminating the `jobStatus.json` XHR call, and push stuff from the server, instead of pulling it from the browser?

ketan:
## Unable to handle abnormal connection close

Using the NGINX config [above](https://github.com/gocd/gocd/pull/3282#issuecomment-302603980). Change the read/send timeout to a high value — `1d`. Open the browser on a console log and kill the docker container `CTRL+C`. The close code is set to `1006`. 1006, according to [the spec](https://tools.ietf.org/html/rfc6455#section-7.4.1) means —

> 1006
>
>   1006 is a reserved value and MUST NOT be set as a status code in a
>   Close control frame by an endpoint.  It is designated for use in
>   applications expecting a status code to indicate that the
>   connection was closed abnormally, e.g., without sending or
>   receiving a Close control frame.

marques-work:
@ketan re: **Unable to handle abnormal connection close**

I saw this even during normal close (i.e. after console logs finished) in Chrome, where it would receive a 1006. We never explicitly set 1006 anywhere in our code. I only got a proper 1000 when I launched the dev server with the debugger and gave it some pause as I was stepping through, so I thought it might be some timing mismatch. As I wasn't able to reliably get a 1000 under normal circumstances, I defined our own explicit error code for real errors and changed the JS to make decisions on that, while being tolerant to 1006.

Of course, that error code won't cover this scenario, and I relied on the the JS to just keep trying after connections dropped unexpectedly.

marques-work:
@ketan I pushed 2 commits to address the bulk of the issues.

### Checklist

- [x] Unnecessary polling when job is not assigned to agent.
    - WS connection waits/checks for ~ 10 sec until file exists before forcing reconnection, greatly reducing "reconnect storm"
    - halfway compromise between an indefinitely open connection and a polling mechanism
- [x] Having to set a very very long proxy_(read|send)_timeout on NGINX.
    - PING frames periodically sent during idle periods
    - runs in same thread as part of the main `sendLog()` loop (as opposed to scheduled service); can't think of any scenarios where this difference would have a meaningful effect
- [ ] Possibility send more info "other stuff" with the same connection.
- [x] Unable to handle abnormal connection close.
    - as I mentioned earlier, we don't set `1006` ever. I actually think this is the correct behavior, as 1006 is the code which indicates the connection was closed without receiving a close frame, which would be exactly the case in your scenario above.
    - from https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent:

| Status code | Name | Description |
| ----------- | ------- | ----------- |
| 1006 | CLOSE_ABNORMAL | **Reserved.** Used to indicate that a connection was closed abnormally (that is, with no close frame being sent) when a status code is expected. |

marques-work:
That said, I'm still looking into the authentication bit to see whether or not I really need to reimplement the endpoint in Jetty's native websocket (i.e. if I need to give up on refactoring authentication code to provide more confidence that the same logic is being exercised).

ketan:
I'm going to review this today and will be pushing commits/rebasing to this branch.

ketan:
The PR looks good and I'm going to merge it AS-IS (after rebasing+squashing).

I have some concerns about lack of compression (and the native web-socket compression known to be buggy with some browser implementations and versions).

Usually, moderately sized console logs will compress with at-least a ratio of 10:1 ratio. Larger console logs may compress better based on the characteristics of the console log, the compression algorithm and window etc...

My concern is with large console logs going uncompressed across the wire on a web-socket connection, this may add several seconds to render the console logs as they arrive over the wire.

Given that the browser implementations are buggy with respect to web-socket compression, I'm tempted to send zlib/gzip binary data in a web-socket packet (instead of text) and decompress it on the client side in JS (assuming there's a decent decompression library available).

Thoughts?

ketan:
Adding a note that safari 10.1.1 supports a non-standard `x-webkit-deflate-frame` compression algorithm, chrome/FF supports `permessage-deflate`

PR #3284, opened 2017-03-18
Make agents page more responsive

ketan:


ketan:
Obligatory screenshots:

### 800x600

screen shot 2017-03-18 at 4 15 54 pm

### 1024x768

screen shot 2017-03-18 at 4 16 01 pm


### 1280x1024

screen shot 2017-03-18 at 4 16 13 pm

naveenbhaskar:
fixed

naveenbhaskar:
fixed for 1280 X 760... should we fix for screensize less than that?

ketan:
It seems to render fine upto 800x600 (see first screenshot). I don't feel we need to support anything below that.

PR #3293, opened 2017-03-21
Add option to make console logs fullscreen on job details page

kierarad:


kierarad:
I retriggered the pipeline, but the status checks are reflecting the first run (1412) for build-linux-PR/build-server. I'm not sure how to update the checks here?

https://build.gocd.io/go/pipelines/trigger/794/do-nothing/1/pipeline

PR #3300, opened 2017-03-23
Can disable material updates as part of integration tests

maheshp:
**Testing do not merge**
* As part of some of our integration tests we create a material instance
 e.g ScheduleServiceRunOnAllAgentIntegrationTest. There have been
 intermittent tests failures when material creation happens concurrently
 thro test as well as MaterialUpdateService.
* Adding an option to turn off material update through tests


maheshp:
Can try doing this by marking auto update to false for these materials, closing for now.

Issue #3295, opened 2017-03-22
Support GET/INDEX API calls for Remote Environments.


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Currently, the GET/INDEX call returns the partial environment from cruise-config.
Add ability to get merged environment config (cruise-config + config-repo) using API.

##### GET environment (merged)

```
curl 'https://ci.example.com/go/api/admin/environments/my_environment' \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd.v2+json'
```

##### GET local environment (cruise-config)

```
curl 'https://ci.example.com/go/api/admin/environments/my_environment?remote=false' \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd.v2+json'
```

##### GET remote environment (config-repo)

```
curl 'https://ci.example.com/go/api/admin/environments/my_environment?remote=true' \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd.v2+json'
```

GaneshSPatil:
@jyotisingh @ketan -- can you look at the query param and give some feedback regarding the API calls?

ketan:
LGTM. Do we have a situation where one might need to see the remote environment in a specific origin, perhaps by passing `origin_name=xxx` param?

GaneshSPatil:
@ketan -- _origin type_ is necessary for getting the editable part(when origin is local) or the entire config(when the origin is merged) of the environment.

There could be a situation where a single environment is defined in multiple origins (cruise-config, multiple config-repos, etc), in such cases, user might want to see the partial environment based on the origin, such as,
`origin_name=local` - to get the editable part of the environment.
`origin_name=config_repo1, config_repo2` - to get the remotely defined part of the environment.

GaneshSPatil:
The point of adding support to **GET environment along with the remote information** is to get the entire environment config using API (current API only returns the partial environment from config.xml even if the same environment is defined in multiple origins).

For Environments specified in multiple origins (`config repository` + `config.xml`),
API should return :
- Editable part of the environment config
- Entire environment config

##### GET editable part of the environment:
```
curl 'https://ci.example.com/go/api/admin/environments/my_environment' \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd.v3+json'
```

##### GET the entire environment:
```
curl 'https://ci.example.com/go/api/admin/environments/my_environment?withremote=true' \
      -u 'username:password' \
[...]

Issue #3301, opened 2017-03-23
Support managing config-repo entities through API


GaneshSPatil:
##### Issue Type

- Feature proposal

#### Environments API 

- [x] Environments GET support for config repo environments(#3295).
- [ ] Environments GET should list `origins` (`cruise-config.xml` or the config-repo).
_Example_: 
```
"origins": [
    {
      "type": "local",
      "file": {
        "_links": {
          "self": {
            "href": "http://localhost:8153/go/admin/config_xml"
          },
          "doc": {
            "href": "https://api.gocd.org/#get-configuration"
          }
        },
        "name": "cruise-config.xml"
      }
    },
    {
      "type": "config repo",
      "repo": {
        "_links": {
          "self": {
            "href": "http://localhost:8153/go/api/admin/config_repos/repo2"
          },
          "doc": {
            "href": "https://api.gocd.org/#config-repos"
          },
          "find": {
            "href": "http://localhost:8153/go/api/admin/config_repos/:id"
          }
        },
        "id": "repo2"
      }
    }
  ]
```
- [ ] Environments GET should show `origin` attribute on individual config entity.
_Example_: 
```
{ 
    "name": "dev",
    "pipelines": [
        {
            'name': 'up43',
            'origin': {
                "type": "config repo",
                "repo": {
                    "_links": {
                        "self": {
                          "href": "http://localhost:8153/go/api/admin/config_repos/repo2"
                        },
                        "doc": {
                          "href": "https://api.gocd.org/#config-repos"
                        },
                        "find": {
                          "href": "http://localhost:8153/go/api/admin/config_repos/:id"
                        }
                    },
                    "id": "repo2"
                }
            }
        }
    ]
}
```

- [ ] Environments INDEX should link to both `local` and `withremote` environment config urls
_Example_: 
```
{
 "_embedded": {
    "environments": [
      {
        "_links": {
          "self": {
            "href": "https://ci.example.com/go/api/admin/environments/foobar"
          },
          "self_with_remote": {
            "href": "https://ci.example.com/go/api/admin/environments/foobar?withremote=true"
          },
          "doc": {
            "href": "https://api.gocd.io/#environment-config"
          },
          "find": {
            "href": "https://ci.example.com/go/api/admin/environments/:environment_name"
          }
        },
        "name": "Dev",
      }
   ]
}
```

- [x] Environments INDEX should support config repo environments.

#### Pipeline config API

- [x] Pipeline config GET support for config repo pipelines.
[...]

PR #3303, opened 2017-03-23
Environments GET API "withconfigrepo" query param

GaneshSPatil:
Environments GET returns the local(cruise-config.xml) environment config.
Environments GET with `withremote=true` query param returns merged environment config.

more information: https://github.com/gocd/gocd/issues/3295#issuecomment-288630637


GaneshSPatil:
Environments API new response JSON - https://github.com/gocd/gocd/issues/3301#issuecomment-289998171

varshavaradarajan:
I don't see the expected API response that you have provided in https://github.com/gocd/gocd/issues/3301#issuecomment-289998171. Were you waiting for https://github.com/gocd/gocd/pull/3330 to be merged?

server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module ApiV3
+  module Admin
+    class EnvironmentsController < BaseController
+      before_action :check_admin_user_and_401
+      before_action :load_local_environment, only: [:put, :patch, :destroy]
```
varshavaradarajan: Is it possible to show the errors for when a user tries to update a remote environment so as to be consistent with pipeline config api?

server/webapp/WEB-INF/rails.new/app/presenters/api_v3/admin/environments/environment_config_representer.rb
```diff
@@ -0,0 +1,83 @@
...
+          opts[:url_builder].apiv3_admin_environment_url(name: environment.name)
+        end
+
+        link :doc do |opts|
+          'https://api.gocd.io/#environment-config'
+        end
+
+        link :find do |opts|
+          opts[:url_builder].apiv3_admin_environment_url(name: '__environment_name__').gsub(/__environment_name__/, ':environment_name')
+        end
```
varshavaradarajan: You had also shown `self_with_config_repo` and `find_with_config_repo` in the proposed json. Don't forget to add them.

GaneshSPatil:
@varshavaradarajan -- this PR is about adding a new API version and endpoint to support config repo environment

I will send another PR regarding the change in the JSON structure

server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
GaneshSPatil: @varshavaradarajan -- Update operation of a config repo environment is not similar as of pipeline config, because Go supports partial environments (but not partial pipelines).

The environment config defined partially in config repository is not allowed to update.
which is taken care of by [environment patch command](https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/config/update/PatchEnvironmentCommand.java#L143-L147).

varshavaradarajan: But according to your json, if an environment's origin ends up being `remote` or `config_repo` instead of `merged`(ie specified partially in local and in remote), then it should show an error?

GaneshSPatil: there wont be an `origin` attribute on an environment. but it will have a list of origins where it has defined.
for an environment defined only in config repo -- one can only add new configuration, but can not modify the existing one.

for a merged environment (defined in config.xml and  config repo) -- one can only modify configurations only from cruise-config.xml

So, at controller level, we only have information about the type as - `local` or `remote` which is not good enough to decide whether the editable part of environment is modified or not.

server/src/com/thoughtworks/go/server/service/EnvironmentConfigService.java
```diff
@@ -21,6 +21,7 @@
 import com.thoughtworks.go.config.commands.EntityConfigUpdateCommand;
 import com.thoughtworks.go.config.exceptions.GoConfigInvalidException;
 import com.thoughtworks.go.config.exceptions.NoSuchEnvironmentException;
+import com.thoughtworks.go.config.merge.MergeEnvironmentConfig;
```
varshavaradarajan: Unused import

server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+        if is_with_remote.nil? and !is_query_param_provided
+          load_local_environment(environment_name)
+        elsif is_with_remote and is_with_remote.downcase == 'true'
+          load_merged_environment(environment_name)
+        else
+          return render_not_found_error
+        end
+
+        json = Admin::Environments::EnvironmentConfigRepresenter.new(@environment_config).to_hash(url_builder: self)
+        render DEFAULT_FORMAT => json if stale?(etag: etag_for(environment_config_service.getEnvironmentForEdit(environment_name)))
```
varshavaradarajan: Why are you fetching the environment config again? etag_for(@environment_config) should suffice.

varshavaradarajan:
### Pending:
1. Environment Config Index call - add a query param to show the remote parts of every environment config as well as remotely defined environment config.

2. Change the representation to include the list of origins where the env config is defined.

server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
GaneshSPatil: @varshavaradarajan -- we need to fetch only the editable part of the environment cause we compute the ETag only for the entity specified in `config.xml`.

`@environment_config` variable contains the merged environment whose ETag won't be same as of one from xml.

Issue #3064, opened 2017-01-03
Getting pipeline config doesn't work for pipelines created via config repos


juhasipo:
##### Issue Type

- Bug Report

##### Summary

I'm trying to fetch pipeline configs for my pipelines in GoCD with the
 `GET /go/api/admin/pipelines/:pipeline_name` API method.

This works fine for my old pipelines created using the GoCD UI but it doesn't find any of the pipelines created via config repo.

###### Basic environment details

* Go Version: GoCD Server 16.12.0
* JAVA Version: 1.7.0_121
* OS: Linux 3.13.0-32-generic

###### Additional Environment Details

Config repos created with gocd-yaml-config-plugin 0.4

##### Steps to Reproduce

1. Create pipeline via config repo e.g. `test-pipeline`
2. Try to fetch pipeline config via GoCD API with admin credentials: `curl 'https://<gocd host>/go/api/admin/pipelines/test-pipeline' -u 'username:password' -H 'Accept: application/vnd.go.cd.v3+json' -i`

##### Expected Results

Response should contain config for pipeline `test-pipeline`

##### Actual Results

Server responses with HTTP code 404 and body:

```JSON
{
  "message": "Either the resource you requested was not found, or you are not authorized to perform this action."
}
```

arvindsv:
@varshavaradarajan Thoughts?

varshavaradarajan:
@jyotisingh and I had discussed it a bit. Since the pipeline configuration for a pipeline specified using a config repo plugin is not present as part of our config xml, showing a more appropriate error message should do? 

But I also think that if a user is has admin permissions to a pipeline group that has pipelines configured via a config repo, shouldn't we just allow that user to view (not edit) the configuration of the config repo pipeline? Or does that defeat the point of having the pipeline configuration inside a git repository?

arvindsv:
Let's see what @tomzo thinks. I don't mind being able to GET but not POST, PUT, PATCH, etc.

tomzo:
> But I also think that if a user is has admin permissions to a pipeline group that has pipelines configured via a config repo, shouldn't we just allow that user to view (not edit) the configuration of the config repo pipeline?

I thought about it earlier. I think GET should be supported to view current configuration, which btw, currently  would be the only way to see what got actually configured.
Secondly there could be a `origin` field in the pipeline config object returned by API to indicate where this pipeline was configured. That would be a first step for #2772

PR #3306, opened 2017-03-23
Pipeline config api show config repo pipeline

GaneshSPatil:


server/test/unit/com/thoughtworks/go/server/service/PipelineConfigServiceTest.java
```diff
@@ -64,6 +64,7 @@ public void setUp() throws Exception {
         pluggableTaskService = mock(PluggableTaskService.class);
         when(goConfigService.getCurrentConfig()).thenReturn(cruiseConfig);
         when(goConfigService.getConfigForEditing()).thenReturn(cruiseConfig);
```
varshavaradarajan: remove `getConfigForEditing` as that's no longer called?

varshavaradarajan:
Related to https://github.com/gocd/gocd/issues/3064

PR #3201, opened 2017-02-24
More human-readable task descriptions in console logs

marques-work:

PR #3307, opened 2017-03-23
Revert "More human-readable task descriptions in console logs"

jyotisingh:
Reverts gocd/gocd#3201


marques-work:
@jyotisingh I may have to write a proper shell escape. I imagine the `quoteArguments()` method (which I used to output the task lines) must not be used in the critical path because it's so fragile with respect to arguments. I'll submit a new PR for the human-readable descriptions that uses a better shell escape method.

marques-work:
@jyotisingh or perhaps I could build tokens with `translateCommandLine(String command)` and assemble a task description that way.

marques-work:
@jyotisingh Oh, I see how this was intended to be used. conceptually, args are parsed with `translateCommandLine(String command)` and mapped to `quoteArgument(String argument)`. That should be a straightforward fix.

PR #3315, opened 2017-03-24
Fix performance issue on the pipeline config SPA

ketan:
The foundation tooltip that was initialized, was never being unloaded
when the DOM was being redrawn. This caused a lot of perf hit when
redrawing the page.


ketan:
This has already been handled in `form_helper` in the webpack directory.

PR #3323, opened 2017-03-27
Move away from 'x minutes/days/months/years ago' time format

ibnc:
Specifically, moving to a format that includes the exact date and time a pipeline/job was triggered.

Job history:
job_history_sidebar

I'm not 100% sure what the format should be for the pipelines page. I have 2 ideas:

Option 1 (what's in this pr):
option_1

Option 2:
option_2

This PR doesn't contain any changes to the pipeline instance page. I'm still not 100% sure what to do about that page, but I wanted to get feedback on what I've got so far. What do you think @arvindsv @naveenbhaskar 

Pipeline instance page:
screen shot 2017-03-27 at 4 16 24 pm


naveenbhaskar:
I prefer option one. in option 2 "triggered" is repeating which clutters the section

arvindsv:
I don't mind the combined option too. But, the "about X hours ago" bit on both pages is a bit of a lie. I believe it shows the time of the latest commit or something like that. It might be useful to change that to be accurate.

arvindsv:
Don't know if it makes sense to hide the year if it's the same as the current one.

ibnc:
@arvindsv @naveenbhaskar I'll go with option 1 then :)

@arvindsv: You are correct in saying that the "x hours ago" bit on the pipeline history page is how long ago the material was updated. 

But you're incorrect in regards to the dashboard. The "x hours ago" displayed on the dashboard for each pipeline is actually when the pipeline was scheduled.

Honestly, I thought they both indicated the same thing until you said something about it, and then I took a look. My instinct is to leave the pipeline history page alone, and just go ahead with the modifications to the dashboard and job history sidebar. Thoughts?

arvindsv:
Ok, I agree.

arvindsv:
Wait. Are you saying the pipeline history page will still say "14 hours ago"? I know the time is wrong, but should it really be inconsistent - compared to the dashboard? I'd think not. What do you think?

arvindsv:
Ah, I see this:

> This PR doesn't contain any changes to the pipeline instance page. I'm still not 100% sure what to do about that page, but I wanted to get feedback on what I've got so far.

Alright.

ibnc:
@arvindsv Yeah, it will still say the old "x hours ago" format. 

The pipeline history time currently indicates when the material was updated, but it's not super clear just by looking at it. I've been using Go for a few years now, and I've always assumed that the time info on the dashboard and pipeline history page were indicating the same thing. I only now know that the pipeline history time info is about the materials because I looked at the code. 

I guess I'm unclear what you mean the time is wrong on the pipeline history page. How is it wrong?

arvindsv:
Oh, just wrong in the way you mentioned. In that it's not the trigger time.

ibnc:
@arvindsv Should I go ahead and change the pipeline history time to be the same as the dashboard?

arvindsv:
Yeah, I don't see why it should be the different from the others. As you know, that page might completely change soon. So, don't spend too much time on it. If it's a small change, I'd just do it.

ibnc:
@arvindsv It wasn't too complicated. How does this look?

pipeline_history


I had to remove the timezone info for the pipeline history page to keep everything on one line. idk if that's a big deal or not

ibnc:
@arvindsv I went ahead and pushed it up. Let me know if anything needs to change on how I'm displaying the date on the pipeline history page

ibnc:
@jyotisingh @ketan @arvindsv After these changes go in, there won't be anyplace in the code that uses the jquery lib timeago. Am I good to remove that library?

arvindsv:
@ibnc 

> I had to remove the timezone info for the pipeline history page to keep everything on one line. idk if that's a big deal or not

I think it's ok. But, there's always so much space on the right there, that I think we should use it better and make the first column wider. I'll leave it to @naveenbhaskar and you.

jyotisingh:
@ibnc - all usages of that library would be gone with this PR, so we should be good to remove jQuery.timeago. 

Just ran through the changes made in the PR, looks good to me. Few things though:
* PR build needs fixing -> https://build.gocd.io/go/pipelines/value_stream_map/trigger/846
* Need a functional tests PR for updating the assertions for the scheduled times. 
* Just curious about one [change](https://github.com/gocd/gocd/pull/3323/files#diff-a1ecf7ea1525fb9d88dc69eb7216f142R27) though, the date format has changed from `dd MMM, yyyy` to `MMM dd yyyy`. I don't have any strong preference for one over the other. Any specific reason for making that change? If you prefer sticking to the updated format, could you also submit a PR for fixing the functional tests as there are a few existing usages of `Java::JavaUtil::Date#to_long_display_date_time`.  `git grep to_long_display_date_time` should tell you the usages.

ibnc:
@jyotisingh 

I'm just starting my day, and will look into the failing tests.

On the date format. I'm not attached to it really. It just read more colloquially to me. I'll switch back for ease of testing.

ibnc:
@jyotisingh I've pushed a fix for the broken tests, and switched back to the previous date format. I've removed jquery.timeago, and will push it once the PR build passed. 

I do have a question on one of your bullet points though. What specifically do you mean by the "Need a functional tests PR for updating the assertions for the scheduled times." bullet point?

arvindsv:
I guess things like [this](https://github.com/gocd/functional-tests/blob/1f607d4cb277da4ddb5e9f7e0ecd7a1f63565a03/src/test/java/com/thoughtworks/cruise/page/OnPipelineDashboardPage.java#L696)? Not sure whether date is being verified though.

server/webapp/WEB-INF/rails.new/app/helpers/pipelines_helper.rb
```diff
@@ -30,7 +30,7 @@ def run_stage_label stage
 
   def trigger_message(actual_date_time, pim)
     if pim.class != com.thoughtworks.go.presentation.pipelinehistory.PreparingToScheduleInstance
-      "(Triggered&nbsp;#{l.string('BY')}&nbsp;<span class='who'>#{h pim.getApprovedBy()}</span>&nbsp;<span class='time'></span><input type='hidden' value='#{actual_date_time}'/>)"
+      "(Triggered&nbsp;#{l.string('BY')}&nbsp;<span class='who'>#{pim.getApprovedBy()}</span>&nbsp;on&nbsp;<span class='time'>#{actual_date_time.to_long_display_date_time()}</span><input type='hidden' value='#{actual_date_time}'/>)".html_safe
```
ketan: Can we kill the parentheses around the text — it just seems to add to clutter.

Also - `html_safe` feels dangerous because we're rendering something we don't control (the login name). Can we instead use [`content_tag`](http://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag) instead

common/src/com/thoughtworks/go/util/TimeConverter.java
```diff
@@ -127,6 +127,18 @@ private static SimpleDateFormat getDateFormatterWithTimeZone() {
         return new SimpleDateFormat("d MMM yyyy HH:mm 'GMT' Z", Locale.ENGLISH);
     }
 
+    public String getHumanReadableString(Date date) {
+
+        return date == null ? ConvertedTime.NOT_AVAILABLE.toString() : new SimpleDateFormat("dd MMM, yyyy 'at' HH:mm:ss", Locale.ENGLISH)
```
ketan: Creating new instances of `SimpleDateFormat` is expensive. Formatting dates using that is expensive too. Can't make it static because it's not thread safe either.

Consider using a static instance of `org.joda.time.format.DateTimeFormatter` instead.

server/webapp/WEB-INF/rails.new/app/helpers/pipelines_helper.rb
ibnc: Removing the h method was an accident. Adding the h method back in should resolve the problem as well. 

Let me know if that isn't enough, or if you'd prefer to just switch to content_tag

ibnc:
@ketan Changes made. Let me know if everything looks good

ibnc:
@ketan @arvindsv @jyotisingh @naveenbhaskar Do y'all think this PR is ready to merge, or is there some work that I still need to do?

arvindsv:
I'm fine with it, the way it is. I've only seen the screenshots, though. Haven't tried it or looked at the code.

ketan:
I'll merge it today.

PR #3324, opened 2017-03-28
Display Plugin name instead of plugin id in dropdown. Aded spec to validate dropdown selection

bdpiparva:
Elastic profile modal now have plugin name in dropdown instead of plugin id.

Added missing spec to validate dropdown selection change

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/elastic_profiles/elastic_profile_modal_body.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/elastic_profiles/elastic_profile_modal_body.js.msx
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-let m             = require("mithril");
-let f             = require('helpers/form_helper');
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/elastic_profiles/elastic_profile_modal_body.js.msx
```diff
@@ -42,7 +42,7 @@ const ElasticProfileModalBody = {
       return vnode.attrs.pluginInfos().mapPluginInfos((pluginInfo) => {
         return {
           id:   pluginInfo.id(),
-          text: pluginInfo.id()
+          text: pluginInfo.about().name() || pluginInfo.id()
```
GaneshSPatil: can we add the version as well in the describe? something such as _"Docker Elastic Agent Plugin [v0.5]"_

bdpiparva: ok

PR #3359, opened 2017-04-05
Pipeline config SPA toggle

GaneshSPatil:
* Render pipeline-config SPA using _mithril-1.1.1_ by default
* Use 'old_view=true' query param to switch the pipeline config page to be rendered using _mithril-0.2.5_


maheshp:
Sorry dint get why we need this toggle?

jyotisingh:
In case something breaks on the new one?

GaneshSPatil:
@maheshp @jyotisingh -- yes
we are keeping both pages for few days, making the new page as default, and if something breaks on the new page, we should have an option to go to the old page.

jyotisingh:
Realized there are no automated tests for the pipeline config SPA yet. Merging this.

PR #3361, opened 2017-04-05
Authorization PluginInfo has role_settings only if the plugin can aut…

maheshp:
…horize

* Authorization plugins have additional capabilities for
'can_verify_connection' and 'can_authorize'
* PluginInfo has role_settings only if plugin can_authorize

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/models/Capabilities.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/models/Capabilities.java
@@ -33,17 +33,12 @@ public class Capabilities {
     private final boolean canSearch;
 
     @Expose
-    @SerializedName("can_verify_connection")
-    private final boolean canVerifyConnection;
-
-    @Expose
     @SerializedName("can_authorize")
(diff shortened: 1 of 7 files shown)


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/models/Capabilities.java
```diff
@@ -32,9 +32,19 @@
     @SerializedName("can_search")
     private final boolean canSearch;
 
-    public Capabilities(SupportedAuthType supportedAuthType, boolean canSearch) {
+    @Expose
+    @SerializedName("can_verify_connection")
```
jyotisingh: @maheshp - As discussed with you, as of now I think the only plugin which would answer to this in negative would be password file auth plugin. However in that case too, usually we see people adding a path that doesn't exist or the file permissions for 'go' user is not setup properly. Since the inbuilt one does not have an option for checking if the setup is proper, people often end up saving their configs with a wrong file and get themselves blocked out. So, I think it would be useful to have such a check even for the password plugins. I agree about the changes around `can_authorize` call, that makes sense.

PR #3364, opened 2017-04-06
BuildSession Mercurial Materials Support

ibnc:
As the title states, this PR is adding mercurial materials support for the build session. I modeled this code after how git materials are supported by the build session work. So, let me know if the git materials support isn't actually completed, and I have more work to do.

### DIFF
--- a/domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
+++ b/domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
@@ -50,11 +50,33 @@ public class HgMaterialUpdater {
         return compose(
                 mkdirs(workDir).setTest(test("-nd", workDir)),
                 cleanWorkingDir(workDir),
-                cmdClone(workDir));
+                cmdClone(workDir).setTest(isNotRepository(workDir))
+        );
+    }
+
+    private BuildCommand isRepoUrlChanged(String workDir) {
+        return test("-neq",
+                material.getUrlArgument().forCommandline(),
+                exec("hg", "showconfig", "paths.default").setWorkingDirectory(workDir));
+    }
+
+    private BuildCommand isBranchChanged(String workDir) {
+        return test("-neq",
+                material.getBranch(),
+                exec("hg", "branch").setWorkingDirectory(workDir));
+    }
+
+    private BuildCommand isNotRepository(String workDir) {
+        return test("-nd", new File(workDir, ".hg").getPath());
     }
 
     private BuildCommand cleanWorkingDir(String workDir) {
-        return cleandir(workDir).setTest(test("-d", workDir));
+        return compose(
+                cleandir(workDir).setTest(isNotRepository(workDir)),
+                cleandir(workDir).setTest(isRepoUrlChanged(workDir)),
+                cleandir(workDir).setTest(isBranchChanged(workDir))
+
+        ).setTest(test("-d", workDir));
     }
 
     private BuildCommand cmdClone(String workingDir) {
--- a/domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
+++ b/domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
@@ -20,8 +20,11 @@ import com.thoughtworks.go.config.materials.mercurial.HgMaterial;
 import com.thoughtworks.go.domain.BuildCommand;
 import com.thoughtworks.go.domain.materials.Revision;
 import com.thoughtworks.go.domain.materials.RevisionContext;
+import com.thoughtworks.go.domain.materials.svn.MaterialUrl;
 import com.thoughtworks.go.util.command.UrlArgument;
+
 import java.io.File;
+
 import static com.thoughtworks.go.domain.BuildCommand.*;
 
 public class HgMaterialUpdater {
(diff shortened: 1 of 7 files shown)


arvindsv:
The git material build session support is complete.

ibnc:
@jyotisingh @ketan Either of you have time to review this PR?

domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+        this.material = material;
+    }
+
+    public BuildCommand updateTo(String baseDir, RevisionContext revisionContext) {
+        Revision revision = revisionContext.getLatestRevision();
+        String workingDir = material.workingdir(new File(baseDir)).getPath();
+        UrlArgument url = material.getUrlArgument();
+        return compose(
+                echoWithPrefix("Start updating %s at revision %s from %s", material.updatingTarget(), revision.getRevision(), url.forDisplay()),
+                cloneIfNeeded(workingDir),
```
arvindsv: The command-line is not protected (meaning, it's not marked as a secret). So, if you have a password in the URL and the command prints out the URL for some reason (failure), it'll print it out. The equivalent of [this line](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java#L44) is what's missing.

domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+                mkdirs(workDir).setTest(test("-nd", workDir)),
+                cleanWorkingDir(workDir),
+                cmdClone(workDir).setTest(isNotRepository(workDir))
+        );
+    }
+
+    private BuildCommand isRepoUrlChanged(String workDir) {
+        return test("-neq",
+                material.getUrlArgument().forCommandline(),
+                exec("hg", "showconfig", "paths.default").setWorkingDirectory(workDir));
```
arvindsv: This took a while to figure out. When a URL has a password, the left-side of this expression is: `https://username:password@some-url` and the right-side (output of `hg showconfig`) is always: `https://username@some-url`. So, it'll never match. So, the directory will [be deleted](https://github.com/gocd/gocd/pull/3364/files#diff-7986ebd144f3606fc14a74aff84a36c7R76) and re-cloned.

The equivalent of [this](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/config/materials/mercurial/HgMaterial.java#L230) is missing.

arvindsv: The call to `MaterialUrl.sameUrl` is also missing. That seems to trim the URL, remove trailing slashes, etc.

arvindsv:
Oh, also, I expect that the difference between [this line in the old code](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/mercurial/HgCommand.java#L77) and [this, in the new code](https://github.com/gocd/gocd/pull/3364/files#diff-7986ebd144f3606fc14a74aff84a36c7R46) will cause the update to fail, since there is no call to `hg pull`.

ibnc:
@arvindsv I've made the requested changes. Thanks for taking a look at this :)

Let me know if you see any other issues!

domain/src/com/thoughtworks/go/domain/materials/mercurial/HgMaterialUpdater.java
```diff
@@ -0,0 +1,95 @@
...
+        return compose(
+                mkdirs(workDir).setTest(test("-nd", workDir)),
+                cleanWorkingDir(workDir),
+                cmdClone(workDir).setTest(isNotRepository(workDir))
+        );
+    }
+
+    private BuildCommand isRepoUrlChanged(String workDir) {
+        return test("-neq",
+                new MaterialUrl(material.getUrlArgument().forCommandline()).toString(),
```
arvindsv: I think this will still not do what we need. :( `MaterialUrl` does not remove the password. `HgUrlArgument` does. It needs to be used in combination.

ibnc: @arvindsv What about something like this?? 
```
private BuildCommand isRepoUrlChanged(String workDir) {
    HgUrlArgument url = new HgUrlArgument(material.getUrl());
    return test("-neq",
            new MaterialUrl(url.defaultRemoteUrl()).toString(),
            exec("hg", "showconfig", "paths.default").setWorkingDirectory(workDir));
}
```

arvindsv: Looks good to me. If you can make a small change: Add a `getUrl` method to `MaterialUrl` and use that, so it is clear. `toString` is prone to change if something changes in the class, and my preference is to use it just for logging/debugging.

ibnc: @arvindsv Done. I'll trigger a pr build for this

PR #3368, opened 2017-04-07
Save agent hostname and ip address as part of build history

kyleolivo:
If an agent was removed from the Go config, the UI would display that job as having not
been assigned an agent. This change stores the hostname and ip address of the agent
that ran the build as part of the build history.


common/src/com/thoughtworks/go/domain/AgentInstance.java
```diff
@@ -331,6 +333,14 @@ public boolean canBeDeleted() {
         return isDisabled() && !(isBuilding() || isCancelled());
     }
 
+    public Map<String, String> getAssignmentInfo() {
```
jyotisingh: Why keep all of this info in a map and access it via string keys? a simple POJO should do right?

jyotisingh:
@kyleolivo - Since, we get the agent data from db now, the agent information will appear in the UI which is what the original intent was. The info is a link and would point to the agent detail page. But if the agent has been removed from the config, the clicking on the link will lead to a 404. So, this would create a new problem. If the agent is not part of the config, we might not want to display the link at all if it leads to a 404. Or if its easy to show the agent details page without the agent being part of the config, then it would be nice. 

@ketan and I were talking about this PR earlier today and one of the things that both of us felt was that the agent data (ie. hostname and ip) is saved at a job level. Since these are for an agent, shouldn't they belong to the Agents table which already has the uuid as the primary key. He found a bug that a record for elastic-agent was not being inserted in the agents table. We fixed that, @ketan will send out a PR for the same. Do you think moving the two columns that were added as a part of this PR to agents table would be ok?

kyleolivo:
@jyotisingh Clicking the link for a removed agent should take the user to the same "agent not found" page that exists today. That said, as you suggested, it may be better to remove the link entirely if the agent info can't be reached. I can look into this.

As for your second comment about storing the data in the Agents table, this seems more appropriate on the surface, but if an agent is deleted from the config (as I believe elastic agents would be, as well as any agent manually deleted from the UI) then the record in the Agents table is removed. So the agent information would be inaccessible and we'd have the same issue that we have today. I tried to confirm this behavior today, but have been having trouble getting the elastic agents to work.

What do you think?

common/src/com/thoughtworks/go/domain/AgentInstance.java
kyleolivo: Yeah, can do.

ketan:
> but if an agent is deleted from the config (as I believe elastic agents would be, as well as any agent manually deleted from the UI) then the record in the Agents table is removed.

That is not true, all agents show up in the agents table, we don't seem to have any code that removes stuff from this table. Like @jyotisingh mentioned, there was a bug that was preventing elastic agents from even showing up in that table. I've fixed that in 581ee39877bb7ebf49ce657e6abd9e14a5b248f0, so you should probably rebase this PR on current master.

There's nothing special about elastic agents, except for some [autoregistration properties](https://github.com/gocd/gocd/blob/b66bddc26b130e15e46f9cb7a838b01d28014945/common/src/com/thoughtworks/go/config/AgentAutoRegistrationProperties.java#L34-L34) and how the [runtime info](https://github.com/gocd/gocd/blob/3c20169531d5716d8d641fc9f5714b30fd3aaa56/agent/src/com/thoughtworks/go/agent/AgentController.java#L153-L157) is created on the agent side.

You should be able to check the behavior of an elastic agent by simply running this command with one of the agent images listed at https://hub.docker.com/r/gocd/

```bash
docker run -e GO_EA_SERVER_URL=... \
           -e GO_EA_AUTO_REGISTER_KEY=... \
           -e GO_EA_AUTO_REGISTER_ENVIRONMENT=... \
           -e GO_EA_AUTO_REGISTER_ELASTIC_AGENT_ID=... \
           -e GO_EA_AUTO_REGISTER_ELASTIC_PLUGIN_ID=... \
           ... \
           IMAGE_ID
```

marques-work:
@jyotisingh @ketan I believe this one is ready for review again. A lot of cleanup/refactoring has happened since, and this time, all the tests passed :)

server/src/com/thoughtworks/go/server/persistence/AgentDao.java
```diff
@@ -53,32 +53,38 @@ public AgentDao(SessionFactory sessionFactory, GoCache cache, TransactionTemplat
...
+
+    public Agent agentByUuid(String uuid) {
+        String key = agentCacheKey(uuid);
+        Agent agent = (Agent) cache.get(key);
+
+        if (agent == null) {
             synchronized (key) {
-                cookie = findAgentCookieByUuid(agentIdentifier);
-                cache.put(key, cookie);
+                agent = fetchAgentByUuid(uuid);
```
jyotisingh: Need double checking to be done here

server/src/com/thoughtworks/go/server/service/AgentService.java
```diff
@@ -98,9 +97,16 @@ public void initialize() {
     }
 
     public void sync(Agents agents) {
+        syncWithDB(agents);
```
jyotisingh: This would be super expensive. `AgentService.sync` is invoked by `AgentChangeListener.onEntityConfigChange` which is invoked everytime an agent's config changes. Imagine in case of build.gocd setup itself when we bring up and tear down many hundred elastic agents each day. Each of those operations is going to sync every configured agent with the db. Is that really required? The other thing being, that callback is still part of config save operation, which means any other config save operation would be in queue until the db sync finishes off.

server/src/com/thoughtworks/go/server/controller/JobController.java
```diff
@@ -105,15 +109,22 @@ public ModelAndView jobDetail(@RequestParam("pipelineName") String pipelineName,
...
+
+        if (null == agent) {
+            agent = Agent.fromConfig(goConfigService.agentByUuid(instance.getAgentUuid()));
+        }
+
+        return getModelAndView(instance, agent);
     }
 
-    private ModelAndView getModelAndView(JobInstance jobDetail) throws Exception {
+    private ModelAndView getModelAndView(JobInstance jobDetail, Agent agent) throws Exception {
```
jyotisingh: The `agent` being passed is not used in the method. `presenter(jobDetail)` continues to use the one from config.

server/src/com/thoughtworks/go/server/service/AgentService.java
```diff
@@ -368,6 +374,17 @@ public String assignCookie(AgentIdentifier identifier) {
         return cookie;
     }
 
+    public Agent findAgentObjectByUuid(String uuid) {
+        Agent agent = agentDao.agentByUuid(uuid);
```
jyotisingh: Given the original logic was to always lookup in config, was wondering if we should lookup the config by default and if not found, make a trip to db. Does that make sense to you?

jyotisingh:
@kyleolivo, @arvindsv  - I am not sure if it makes sense or not, just speaking out loud. 

I was thinking about how to avoid the expensive [syncWithDB](https://github.com/kyleolivo/gocd/blob/45def9a0c171e386a59e5fa4be5dcf66f10945e0/server/src/com/thoughtworks/go/server/service/AgentService.java#L104) call which is made after every agent save (insane number of times for setups using elastic agents) or even config save through the old UI. Imagine the plight of H2db with this.

One of the thoughts I have is, are we ok displaying potentially stale information for some agents which are not part of the config anymore. So:
(i) for the agents that are still part of the config, we could display the most accurate information about hostname, IP etc from the [agentInstances](https://github.com/gocd/gocd/blob/354576dca0f4b67af898a123aac000775ddbb710/server/src/com/thoughtworks/go/server/service/AgentService.java#L66) in memory. 
(ii) for the ones which are not part of the config - there are two sub categories here:
(a) Agents that registered, ran a few jobs, were killed (such as elastic agents) : for these, the information could be stored at the time of registration in db (instead of getCookie call). We should display the information from db, which is still accurate. 
(b) Agents that registered with a hostname, changed the hostname/IP to something else later, ran a few jobs and then died - the information is fetched from db, this would be stale though. [syncWithDB](https://github.com/kyleolivo/gocd/blob/45def9a0c171e386a59e5fa4be5dcf66f10945e0/server/src/com/thoughtworks/go/server/service/AgentService.java#L104) fixes data for just these ones if I am not wrong. Wondering if we can let this go. 

Basically, if the agent is not registered anymore, we show the information we have about the agent when if first registered. What do you think?

I am suggesting something along the lines of this patch, its not complete, but if you guys think it makes sense, then I could work with @kyleolivo on this.
[agent_save.patch.txt](https://github.com/gocd/gocd/files/1000229/agent_save.patch.txt)

arvindsv:
@jyotisingh - I don't think I have enough context about this to say anything meaning, either way. I think, ideally, the agent data should be out of config and tracked in some other way, so that config saves are not triggered every time this happens. Given that, the scenarios you outlined makes sense to me - since the ones which are alive and well are shown properly.

kyleolivo:
@jyotisingh Thanks for your comments on this. I'm in agreement with the suggestions you've made. Let me try to wrap up the team foundation build session work today and then I'll make these changes.

server/src/com/thoughtworks/go/server/controller/JobController.java
kyleolivo: Sure does. Removing.

marques-work:
@jyotisingh @arvindsv  @kyleolivo 
>(b) Agents that registered with a hostname, changed the hostname/IP to something else later, ran a few jobs and then died - the information is fetched from db, this would be stale though. syncWithDB fixes data for just these ones if I am not wrong. Wondering if we can let this go.

IMO, I would think that's ok. If the agents are already purged/torn down, even if the name is stale, they are inaccessible anyway. I can't think of a good reason to care because if one were to try to investigate an issue on a such an agent, it's already a dead end.

kyleolivo:
@jyotisingh We've (@kierarad and myself) removed the syncWithDB call as you mentioned in (b). But we are curious about (a). In order for a registration call and a cookie association call to save partial information in the DB, we'll need to allow null values for the agent columns (all but the uuid). It isn't clear to me what the benefit is of saving the agent metadata at registration versus when the cookie is stored (although I agree that logically it seems like a more appropriate place). The cookie and associated metadata currently gets stored right before an agent begins to process work, which seems alright. What do you think?

jyotisingh:
@kyleolivo - Sorry, it's me a long time to get back to this one. 

>The cookie and associated metadata currently gets stored right before an agent begins to process work, which seems alright
Sure, I guess that should be alright.

I rebased this PR. I pushed a new commit(3b521b6) to this PR. 
The main change in the commit was updating `findAgentObjectByUuid` to first lookup for the agent in config and if not found, then look up the db. This is because we do not update the db if the hostname/IP of an agent changes after the getCookie call. For agents which are not part of the config, it's ok to show the stale information, but not for the one's which are still part of the config.
Apart from this, the rest are minor changes. Let me know if you have any concerns about that specific commit. If you find it ok then I will go ahead and merge this PR.

kyleolivo:
That change seems fine to me. Thanks!

PR #3373, opened 2017-04-07
Update crel to 3.0.1 which contains a bugfix for event binding

marques-work:


GaneshSPatil:
@ketan -- if we are modifying the file, don't we need to add our copyright license on the file?

marques-work:
@GaneshSPatil  we aren't modifying it - just updating it from the latest copy from crel's repo.

PR #3374, opened 2017-04-07
BuildSession SVN Materials Support

ibnc:


common/src/com/thoughtworks/go/buildsession/TestCommandExecutor.java
```diff
@@ -37,6 +37,14 @@ public boolean execute(BuildCommand command, BuildSession buildSession) {
...
         }
 
+        if ("-in".equals(flag)) {
+           return captureSubCommandOutput(command, buildSession).contains(left);
+        }
+
+        if ("-nin".equals(flag)) {
+            return !captureSubCommandOutput(command, buildSession).contains(left);
+        }
+
```
arvindsv: Need to make a corresponding change in the golang agent. Or at least let @barrowkwan know.

domain/src/com/thoughtworks/go/domain/materials/svn/SvnMaterialUpdater.java
```diff
@@ -0,0 +1,139 @@
...
+
+      return compose(
+              echoWithPrefix(format("Start updating %s at revision %s from %s", material.updatingTarget(), revision.getRevision(), url.forDisplay())),
+              secret(url.forCommandline(), url.forDisplay()),
+              secret(material.getPassword(), "*********************"),
+              secret(material.currentPassword(), "*********************"),
+              cleanupAndUpdate(workingDir, revision).setTest(shouldDoCleanupAndUpdate(workingDir)),
+              freshCheckout(workingDir, revision).setTest(isNotRepository(workingDir)),
+              freshCheckout(workingDir, revision).setTest(test("-nd", workingDir)),
+              freshCheckout(workingDir, revision).setTest(repoUrlChanged(workingDir)),
```
arvindsv: It felt, to me, that this could checkout 3 times in certain cases. But, after some thought, it looks like it won't, because of the order of those three statements. I think it would be nice to change this to:

```java
freshCheckout(workingDir, revision).setTest(not(shouldDoCleanupAndUpdate(workingDir))),
```

But, that has a problem too, where the second check is affected by the previous action. So, your approach is fine.

domain/src/com/thoughtworks/go/domain/materials/svn/SvnMaterialUpdater.java
```diff
@@ -0,0 +1,139 @@
...
+       String userName = material.getUserName();
+       if (!StringUtils.isBlank(userName)) {
+           argList.add("--username");
+           argList.add(userName);
+           if (!StringUtils.isBlank(material.getPassword())) {
+               argList.add("--password");
+               argList.add(material.getPassword());
+           }
+       }
+
```
arvindsv: Can remove this empty line.

domain/src/com/thoughtworks/go/domain/materials/svn/SvnMaterialUpdater.java
```diff
@@ -0,0 +1,139 @@
...
+              freshCheckout(workingDir, revision).setTest(isNotRepository(workingDir)),
+              freshCheckout(workingDir, revision).setTest(test("-nd", workingDir)),
+              freshCheckout(workingDir, revision).setTest(repoUrlChanged(workingDir)),
+              echoWithPrefix(format("Done.\n"))
+      );
+   }
+
+   private BuildCommand freshCheckout(String workingDir, Revision revision) {
+      return compose(
+              cleandir(workingDir).setTest(test("-d", workingDir)),
```
arvindsv: I wonder if it makes sense to echo a message here saying we're recreating the repo since something changed.

common/src/com/thoughtworks/go/buildsession/TestCommandExecutor.java
ibnc: @arvindsv I opened a pull request for these changes on the golang agent (https://github.com/gocd-contrib/gocd-golang-agent/pull/47)

PR #3377, opened 2017-04-10
Add performance log configuration to `log4j.properties`

ketan:


server/properties/src/config/log4j.properties
```diff
@@ -50,3 +50,14 @@ log4j.appender.ShineFileAppender.MaxFileSize=10240KB
 log4j.appender.ShineFileAppender.MaxBackupIndex=50
 log4j.appender.ShineFileAppender.layout=org.apache.log4j.PatternLayout
 log4j.appender.ShineFileAppender.layout.conversionPattern=%d{ISO8601} %5p [%t] %c{1}:%L - %m%n
+
+# Comment this out to get to performance debug logs.
+# This will generate a lot of logs, so do not turn this on, unless you have been advised to
+#log4j.logger.PerformanceLogger=DEBUG, PerformanceLoggerAppender
```
varshavaradarajan: Put this above with the WebRequestsAppender and leave it commented. You can uncomment the below lines so that users don't have to.

ketan: Done. Can you relook.

jyotisingh:
since the log location is configured as `logs/go-server-perf.log`, who takes care of setting this to `/var/log/go-server/go-server-perf.log` for linux installs?

varshavaradarajan:
For deb and rpm installations, the log path gets replaced in LinuxPackagingTask.groovy#132 - (this is not checked in) 
```
log4j.additivity.PerformanceLogger=false
log4j.appender.PerformanceLoggerAppender=org.apache.log4j.RollingFileAppender
log4j.appender.PerformanceLoggerAppender.File=/var/log/go-server/go-server-perf.log 
...
```

Issue #3363, opened 2017-04-05
On Job Details -> Console log tab the "Back to Top" link not accessible when health message is shown


rajiesh:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: `17.4.0 (4819-af91a90e96c75469ad2857ce0fea83f23edfb236)`

##### Steps to Reproduce

1. Navigate to Job Details -> Console log tab of any job
2. Ensure the server health message shows up with Error & Warnings
3. The "Back to Top" link on bottom right corner of the console logs hides behind the health message.

Here is the screenshot from build.go.cd

<img width="1385" alt="screen shot 2017-04-05 at 6 53 18 pm" src="https://cloud.githubusercontent.com/assets/6533924/24707579/cc412e6c-1a31-11e7-8d6b-6d81401ce347.png">

ibnc:
@naveenbhaskar What're your thoughts on how to resolve this issue?

I'm thinking one of the two solutions:

Move the errors/warnings box to the bottom left of the page:


<img width="1440" alt="errors_to_left" src="https://cloud.githubusercontent.com/assets/5726224/24766968/3ce99912-1ab2-11e7-9764-cf8458e20220.png">

Moving the back to the top button to the bottom left of the console area:

<img width="1440" alt="back_to_top_to_left" src="https://cloud.githubusercontent.com/assets/5726224/24766972/3f43c6ce-1ab2-11e7-9aca-23bd6896c1cd.png">


Thoughts?

arvindsv:
@naveenbhaskar Let's discuss this tomorrow. We need a resolution one way or another.

naveenbhaskar:
I was thinking of moving the error/warning to the top of navigation bar. The existing position is not very noticeable.

![image](https://cloud.githubusercontent.com/assets/1627961/24846724/0d479fd0-1dda-11e7-9888-7899c48dce28.png)

Another option could be moving "back to top" to the top sticky panel of the console and rename to "scroll to top". reachability for "back to top" is same if its bottom or top if console. Actually its easy if somebody want to move top and bottom at once to check things. 

![image](https://cloud.githubusercontent.com/assets/1627961/24846811/a103553e-1dda-11e7-82a5-5fb0b64739f7.png)

what do you think?

ibnc:
I like the first option @naveenbhaskar. It makes it more noticeable, and I don't think changing the way that people use the console area should be done lightly.


I'll start working on this

ibnc:
[...]

PR #3378, opened 2017-04-10
[Bug fix] Moving error/warnings box to be up in the navigation

ibnc:
These Changes have two benifits:
* These Changes make it more discoverable
* These Changes fix an issue with it overlapping wiht the "back to top" button on the console log (issues: #3363)

Screenshot:
error_warnings_box_position

These changes were made with the advisement of @naveenbhaskar. There's a mockup in the issue

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/new-theme.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/new-theme.scss
@@ -100,6 +100,26 @@ h4 {
 }
 
 .application_nav {
+  .error_messaging_counter {
+    float: right;
+    .cruise_messages {
+      position: static;
+      padding-top: 9px;
+
+      a {
+        border: initial;
+        border-radius: 4px;
+        background-color: #a50000;
+        font-weight: 700;
+        line-height: 11px;
+
+        span {
+          line-height: 11px;
+        }
+      }
+    }
+  }
+
   .tabs {
     margin-top:    0px;
     margin-bottom: 0;
--- a/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/new-theme.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/new-theme.scss
@@ -105,6 +105,7 @@ h4 {
     .cruise_messages {
       position: static;
       padding-top: 9px;
+      box-shadow: none;
 
       a {
         border: initial;
--- a/server/webapp/WEB-INF/vm/shared/_header.vm
+++ b/server/webapp/WEB-INF/vm/shared/_header.vm
@@ -136,8 +136,10 @@
                             #end
                         </li>
                       </ul>
-                      <div id="cruise_message_counts" class="cruise_messages"></div>
-                      <div id="cruise_message_body" style="display:none;" class="cruise_message_body"></div>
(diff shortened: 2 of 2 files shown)


maheshp:
While working on it do you mind fixing it on the new mithril pages(agents and pipeline_config quick edit page), currently errors and warnings don't show up on these pages.

ketan:
> While working on it do you mind fixing it on the new mithril pages(agents and pipeline_config quick edit page), currently errors and warnings don't show up on these pages.

That'll be a lot more work, I feel. There is no API to get the server health messages.

Since (almost) every page polls the server 2 times every 5 seconds, once for the actual page contents, and another time to get the server health messages.

Since we have websockets in core now, I feel that having push based APIs (to start with, just apis for server health messages) will actually help reduce a lot of the unnecessary overhead on the server.

I have some thoughts on how we can start with simple poll based JSON APIs and migrate over to websockets later. I'll open a PR with initial thoughts if it makes sense to do this now.

maheshp:
Oh, I was looking at https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/shared/_error_messaging_counter.html.erb and thought it involved making a ajax call to get data from /go/server/messages.json

Anyways, if its lot more work it can be done as part of a different ticket.

Nevertheless, after moving to errors on the header but not showing on the agents page would be more evident. I think we need to fix this soon.

ketan:
> Oh, I was looking at https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/shared/_error_messaging_counter.html.erb and thought it involved making a ajax call to get data from /go/server/messages.json

That is a JSON containing html inside it. It's the erb/html template that makes it an order of magnitude expensive to render (vs similar object serialized as json)

```json
{
    "cruise_message_counts" : {"html" : "    \u003Ca onclick=\"Modalbox.show($('cruise_message_body'), {title: 'Error and warning messages', overlayClose:false})\"\u003E\n        \u003Cspan class=\"messages\"\u003E\n                    \u003Cspan class=\"error_count\"\u003EErrors: 1\u003C/span\u003E\n        \u003C/span\u003E\n    \u003C/a\u003E\n"},
    "cruise_message_body" : {"html" : "      \u003Cdiv class=\"error\" data-pipelines=\"[\u0026quot;build-windows-private\u0026quot;,\u0026quot;build-linux-private\u0026quot;,\u0026quot;installers-private\u0026quot;]\"\u003E\n        \u003Cdiv class=\"message\"\u003EModification check failed for material: URL: https://mirrors.gocd.io/git/gocd/gocd-private, Branch: master [Apr-10 21:07:56]\u003C/div\u003E\n        \u003Cdiv class=\"description\"\u003EFailed to run git clone command STDERR: fatal: https://mirrors.gocd.io/git/gocd/gocd-private/info/refs not found: did you run git update-server-info on the server?\u003C/div\u003E\n      \u003C/div\u003E\n"}
}
```

The convention is that the JS polling code basically goes through each of the k/v pairs looks up HTML elements using the css selector in the key, and plonks the value as the html body in the element —

```javascript
_.map(json, function(value, key){
  jQuery(key).html(value);
});
```

maheshp:
@ibnc Removed the box-shadow for cruise_messages

Issue #2666, opened 2016-09-09
Upgrade to a newer (supported) perforce client (and server) for tests


ketan:
##### Issue Type
- Bug Report
- Feature proposal
##### Summary

The current version of perforce that we use `r09.2` is fairly ancient and there is no support available for it, and the binaries are no longer available to download.

I'm opening this issue for someone to investigate what it takes to support a newer perforce version with GoCD.

ketan:
Moving to testing with `v15.1` of perforce. Although we've just upgraded the version of the p4 repository files, the `p4` commands that are executed remain unchanged. So "in theory", we would continue to work with older versions of `p4`.

PR #3379, opened 2017-04-10
Upgrading perforce test data

ibnc:
See #2666 for some details. I did this work because I need the test data upgraded in order to start working on perforce material support for the BuildSession.

@ketan You opened the issue on upgrading the test data. What're your thoughts?


ibnc:
@ketan The perforce tests are gonna fail until the version on the go-agents is upgraded to at least 2013.3. 

Here's the version info for my local machine where all perforce tests pass:

`
Rev. P4D/DARWIN90X86/2016.2/1498579 (2017/03/21).
`

common/test/unit/com/thoughtworks/go/helper/P4TestRepo.java
```diff
@@ -110,7 +110,7 @@ public void stop() {
     }
 
     private ProcessWrapper startP4dInRepo(File tempRepo) throws IOException, CheckedCommandLineException {
-        CommandLine command = createCommandLine("p4d").withArgs("-C0", "-r", tempRepo.getAbsolutePath(), "-p", String.valueOf(port));
+        CommandLine command = createCommandLine("p4d").withArgs("-r", tempRepo.getAbsolutePath(), "-p", String.valueOf(port));
```
ketan: Is this a mac-only fix (case insensitive filesystem)?

ibnc: It very well might be. Should I change it back?

ibnc: The server couldn't start locally until I removed that option. Removing the option might be causing some of the perforce failures on build.gocd

ibnc:
@ketan I'm a little at a loss as to why the tests are failing on build.gocd (see: https://build.gocd.io/go/tab/build/detail/build-linux-PR/1558/build-non-server/1/common-split-runInstance-3)

I'm using the same version locally, and they're passing just fine. The only real difference is the OS. I develop on a mac.... Any ideas?? I've sorta fried my brain working on this today.

ketan:
The best way to reproduce this would be to run it on a vagrant box. Just run `vagrant up` — you might need to pull down the p4 binaries into the VM.

It might also help if we log the exit code along with the stdout/stderr when blowing up (even if to temporarily debug it)

ibnc:
@ketan I figured out what was wrong. I didn't upgrade the db files with the `-C0` flag. 

The tests are now passing on build.gocd. The Windows tests are failing though, but that's to be expected because its perforce version hasn't been upgraded for Windows.

This should be good to merge. I'm not sure how to go about upgrading perforce on Windows

ketan:
> This should be good to merge. I'm not sure how to go about upgrading perforce on Windows

I'll take that up. We have tests which run in other internal environments which also need an upgrade. I'll work with @junaidshah on it. But expect this to be merged in post the release.

PR #3381, opened 2017-04-11
Remove unnecessary web filter

ketan:
Jetty already does web-request logging, and does a better
job at it. Plus most tools can slurp the logs and provide
much meaningful information and insight than our custom logger


arvindsv:
Ok. It was for consistency with other logging of this kind, but it makes sense to remove it, since the web logs are easier to turn on now.

Issue #2812, opened 2016-10-25
Jobs run on elastic agents show misleading agent information on job run page


maheshp:
##### Issue Type
- Bug Report
##### Summary

The job run page shows a misleading Agent information `Not yet assigned (ip:Unknown)` if the job runs on an elastic agent.

The expected agent information is agent_name:(ip_address) which is linked to the agents page. The info displayed for a normal agent might not make sense for elastic agents since these might be deleted post job run, but the current info is misleading as well.

kierarad:
@maheshp Is this still an issue? I was unable to recreate it. The build summary information showed the agent for me. See screenshots - one is a build in progress, the other a completed build. 
![screen shot 2017-03-22 at 1 19 58 pm](https://cloud.githubusercontent.com/assets/3814794/24219004/8c4ce718-0f02-11e7-94d8-7773adb52786.png)
![screen shot 2017-03-22 at 1 18 58 pm](https://cloud.githubusercontent.com/assets/3814794/24219008/90a826c4-0f02-11e7-83af-a88fe181605b.png)

arvindsv:
@kierarad I still see it on build.gocd.io (see [this](https://build.gocd.io/go/tab/build/detail/build-linux/1558/build-non-server/1/agent)). As the title said, it's for elastic agents, only.

kierarad:
@arvindsv ok I'll dig deeper. The screenshots are with an elastic agent (using the docker-elastic-agents plugin)

jyotisingh:
@kierarad - so the issue is not isolated to elastic agents. 
Basically, the agent information for a completed job shows correctly, as long as the agent is still part of Go's config. As soon as you delete it from the config, the agent information from job details page goes away too. 
For regular agents, such a case would happen only if someone manually removes the agent from the config, so this behavior was sort of not a big deal as people wouldn't delete agents from config very often. However, with elastic agents, esp things like docker the agents are torn off and removed from config a min or so after job completes (during the server ping call). So, there is a very high chance that you would see a message like "Not yet assigned" for a completed job if you use elastic agents. Since, most of Go's builds happen on elastic agents, we pretty much always see that message, and its mis-leading. 
If you still have your setup running where you took the screenshot, you would see the same behavior. You probably took a screenshot before the server-ping call to the plugin happened which is why the agent was still alive.

kierarad:
@jyotisingh - can you confirm the expected behavior?
[...]

PR #3386, opened 2017-04-11
Change message displayed when agent is missing

kyleolivo:
Previously a job that never had an agent assigned or a job that once
had an agent that had since gone missing would display the same message:
"Agent not yet assigned". This message is misleading when displayed as
part of a job that was scheduled, built, and completed. A longer term
solution might persist more agent metadata, but in the short term we
can display a more accurate state message.

See existing issue:
https://github.com/gocd/gocd/issues/2812


kyleolivo:
On the verge of completing a better solution for this. Closing this PR.

Issue #3387, opened 2017-04-12
Issues in quick edit page after Mithrill upgrade


ankitsri11:
This is the placeholder for all issues identified in Quick Edit page after Mithrill upgrade

Issue Type: Bug

Summary:
Stage/Jobs and tasks drag and drop feature is not working in pipeline quick edit page.

Steps:
- Click on pipeline settings gear icon
- Click on "quick edit" link
- try to re-order the stages by drag and drop
- Also try re-ordering jobs/tasks by drag and drop

Expected behaviour:
It should allow to re-order stages/jobs/tasks 

* Go Version: 17.4.0 (4832-8d0eebe4f8ca201d9d79c82586e34a64ff83b12b)
* Browser vendor and version (if relevant): Chrome Version 57.0.2987.133 (64-bit)

Workaround:
Use 'old_view=true' query param to switch the pipeline config page and it should work

ankitsri11:
https://github.com/gocd/gocd/pull/3359

maheshp:
Unable to toggle to Normal Edit

maheshp:
Elastic Profile Id autocomplete not working

maheshp:
Job settings alignment 

![screen shot 2017-04-12 at 3 17 16 pm](https://cloud.githubusercontent.com/assets/1061051/24951777/34ca7982-1f93-11e7-8778-cae758968c85.png)

maheshp:
Cannot drag tasks.

maheshp:
Stage operate permissions, users autocomplete not working

maheshp:
Command lookup autocomplete needs to be fixed

ankitsri11:
Job deletion is not working properly. Throws below exception:

```
vendor-and-helpers.chunk-14e1271….js:24 Uncaught TypeError: t.isBlank is not a function
    at Object.changed (vendor-and-helpers.chunk-14e1271….js:24)
    at Object.remove (vendor-and-helpers.chunk-14e1271….js:24)
    at Object.removeJob (pipeline_configs-0bf74dde66628da03f03.js:1)
    at HTMLButtonElement.N.r (vendor-and-helpers.chunk-14e1271….js:1)
```

In order to delete job, you will need hover on the job name after deleting it.

rajiesh:
Just Consolidating the auto-complete issue which @maheshp raised

Auto-Complete feature is broken at all places(user do not get auto-complete suggestions), Verified for,

* Command Lookup
[...]

PR #3388, opened 2017-04-12
Fix broken dragula config (#3387)

GaneshSPatil:
* Upgrade dragula library to latest version (v3.7.2)
* Fix drag and drop stages widget
* Fix drag and drop tasks widget


GaneshSPatil:
PR build https://build.gocd.io/go/pipelines/value_stream_map/trigger/949

PR #3389, opened 2017-04-12
Fix broken Normal Edit Link (#3387)

GaneshSPatil:
* Fix the broken link of normal edit page
* Make use of new_modal to render the confirm proceed modal
* Add a spec to verify the Normal edit link


GaneshSPatil:
PR pipeline https://build.gocd.io/go/pipelines/value_stream_map/trigger/951

Issue #3371, opened 2017-04-07
Console log UI takes approx 10 secs to load for jobs with console log of 40-50K lines


rajiesh:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: `17.4.0 (4823-6c8c66c4085fce499389c817732bedf570bc4c97)`

##### Steps to Reproduce

With Console log UI improvements(https://github.com/gocd/gocd/pull/3199) the console log tab takes ~10 sec to load for jobs with log of 40-50K lines

From a user perspective its a bit confusing when they see the log tab without any contents for ~10 secs, since there are no visual representation to say its loading the logs. Can the load time be reduced for huge logs? or having a spinner as visual representation for loading a good idea in case of delay?

`Installer-tests` and `UploadInstallers` pipelines in build.go.cd are places were we can see this issue

fire:
Maybe it's possible to fold the console logs? https://blog.travis-ci.com/2013-05-22-improving-build-visibility-log-folds/

rajiesh:
Verified on 17.4.0 (4888-cbd790d635c52dc6e0765cc2b4c8e52745e00271). The Console log loading message gives users indication of the loading. However the page loading still takes time.

Also the animation for `Waiting for console logs..` message increases the page CPU utilization, during test when we profiled on the Chrome browser for the tab which opens the console log page, we could observe ~30% CPU utilization by that page.

We will close this issue for now, if users feel the waiting and the CPU utilization annoying will re-open the issue.

Issue #3384, opened 2017-04-11
Perf issues on console log


ketan:
##### Issue Type

- Bug Report

##### Summary

When scrolling a large console log, the browser is sluggish, and scrolling is delayed by several seconds.

##### Steps to reproduce

* open a largish console log (50K likes) — try this one https://build.gocd.io/go/tab/build/detail/UploadInstallers/274/UploadInstallers/1/UploadInstallers
* open all console log folds
* scroll the page up and down a couple of times in a second
* watch the scroll happen over several seconds

My gut feel is that this is because re-layout of the `console-action-bar` causes a re-calculation of the position of all the elements in the console log (inside `buildoutput_pre`).

![screen shot 2017-04-11 at 2 23 48 pm](https://cloud.githubusercontent.com/assets/10598/24900733/678a81d6-1ec2-11e7-9b67-40b3470530a4.png)

PR #3391, opened 2017-04-12
Console log performance improvements

marques-work:
Remove line numbers - this has a significant impact on scrolling performance and expand/collapse performance. No measurable change to first render.

Wrap foldable section lines (except header line) in a separate element to improve CSS rule matching performance for expand/collapse. This improves CSS style calculations significantly. I saw a drop from 750 -> 120ms in style calc times on a 50,000 line log. Not the biggest contributor, but improvement is visually perceptible.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_foldable_section.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_foldable_section.js
@@ -78,10 +78,10 @@
     }
 
     function insertLine(cursor, prefix, line) {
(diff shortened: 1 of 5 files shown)


marques-work:
Should fix https://github.com/gocd/gocd/issues/3371 and https://github.com/gocd/gocd/issues/3384

ketan:
@marques-work — this does not quite fix #3371 because of lack of feedback about network activity...

marques-work:
@ketan - good point, I'll add a "waiting for logs" message or spinner.

Issue #3367, opened 2017-04-06
Console log UI do not provide Expand/Collapse all option for completed jobs


rajiesh:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: `17.4.0 (4823-6c8c66c4085fce499389c817732bedf570bc4c97)`

##### Steps to Reproduce

As part of the new Console log improvements option for Expand/Collapse All is provides on Jobdetails->Console log UI. However  `Expand All` and `Collapse All` links are available all time when the job is in progress. 

However for completed jobs those options are not available when the console log is loaded first(user visit the console log of that particular job run for first time). If from the same browser session user visit the console log UI of the same job run again the `Expand/Collapse all` option is available

##### Expected Results

`Expand All` and `Collapse All` links should be available for completed jobs in first visit

marques-work:
### More specific reproduction steps
This specifically happens when the console tab is not the current tab when first navigating to the job details page. For instance, if you click on a job and the "Artifacts" tab is the initial tab, clicking to the Console or Failure tabs to view logs will not properly detect the presence of foldable sections.

1. Go to a completed job
2. Click the Artifacts tab so that the session remembers the current tab selection.
3. Refresh the page to simulate navigating there for the first time (optional, for the paranoid)
4. Click the Console tab. The collapse/expand all link will not be present, but should be.

This is due to the fact that we defer all console building until the logs become visible. Particularly, we defer the execute of the DOM building, but not the detection logic that controls the visibility of the collapse/expand links in the action bar. We either need to also defer the detection logic, or explicitly call the detection logic when we switch tabs.

arvindsv:
I'd also like to consider whether there's value in always loading the "Console" tab when loading this page. It always confuses me when it switches to some random tab. I'm sure there's some meaning to that behavior but I haven't been able to find it. I'd like to not make that decision though. :) Maybe @jyotisingh wants to.  Or @rajiesh or @marques-work.

arvindsv:
Also, this page might go away / get redesigned. :) Woo hoo.

marques-work:
@arvindsv I'd vote for a completely separate console log page :)

arvindsv:
Will wait for @rajiesh to decide if it's ok. Reopening.

rajiesh:
Verified on 17.4.0 (4888-cbd790d635c52dc6e0765cc2b4c8e52745e00271). Console log loads up with all foldable sections collapsed and Expand All option available as default for completed jobs

PR #3392, opened 2017-04-12
Ensure collapse/expand all link is always present

marques-work:
Should fix https://github.com/gocd/gocd/issues/3367


ketan:
Unless I'm missing something. The default state of the page is collapsed and the link says "collapse all" instead of "expand all"?

marques-work:
@ketan before https://github.com/gocd/gocd/pull/3398 the last section would be expanded; now that this is not the case, it does make sense to flip the default state. I'll do that now.

arvindsv:
Merging this so @marques-work is not blocked. @ketan needs to take a look.

ketan:
> Merging this so @marques-work is not blocked. @ketan needs to take a look.

LGTM. I was just waiting on the build to go green.

PR #3405, opened 2017-04-17
BuildSession Perforce Material Support

ibnc:


ibnc:
This PR is waiting on https://github.com/gocd/gocd/pull/3379 to be merged

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+        String workingDir = material.workingdir(new File(baseDir)).getPath();
+        String revision = revisionContext.getLatestRevision().getRevision();
+        String clientName = material.clientName(new File(workingDir));
+
+        return compose(
+                secret(material.getPassword()),
+                export("P4PORT", material.getServerAndPort(), false),
+                export("P4USER", material.getUserName(), false),
+                export("P4PASSWD", material.getPassword(), true),
+                export("P4CLIENT", clientName, false),
```
arvindsv: The use of export here (four lines) could be problematic. What this does is make those environment variables available to the user, later in the process. What we really need is for the `exec` in `constructClient` to accept environment variables, which it uses, but does not echo or export for user-defined tasks. @ibnc - Do you think that makes sense?

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public BuildCommand updateTo(String baseDir, RevisionContext revisionContext) {
+        String workingDir = material.workingdir(new File(baseDir)).getPath();
+        String revision = revisionContext.getLatestRevision().getRevision();
+        String clientName = material.clientName(new File(workingDir));
+
+        return compose(
+                secret(material.getPassword()),
+                export("P4PORT", material.getServerAndPort(), false),
+                export("P4USER", material.getUserName(), false),
```
arvindsv: This should be done only in this case: `!StringUtil.isBlank(material.getUserName())` based on the [old code](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L148).

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+                export("P4PASSWD", material.getPassword(), true),
```
arvindsv: This should be done only in this case: `!material.getUseTickets() && !StringUtil.isBlank(material.getPassword())` based on the [old code](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L151).

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+        this.material = material;
+    }
+
+    public BuildCommand updateTo(String baseDir, RevisionContext revisionContext) {
+        String workingDir = material.workingdir(new File(baseDir)).getPath();
+        String revision = revisionContext.getLatestRevision().getRevision();
+        String clientName = material.clientName(new File(workingDir));
+
+        return compose(
+                secret(material.getPassword()),
```
arvindsv: There's a [call to `login`](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L114) which needs to happen once, and it's missing.

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+                + "View:\n"
+                + material.p4view(clientName);
+        return exec("p4", "client", "-i").setExecInput(clientArgs);
+    }
+
+    private BuildCommand sync(String workingDir, String revision) {
+        return exec("p4", "-d", workingDir, "sync", "@" + parseLong(revision));
+    }
+
+    private BuildCommand cleanWorkingDir(String workingDir) {
```
arvindsv: This doesn't seem to accurately reflect what's happening [here](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/config/materials/perforce/P4Material.java#L313). Is this the new way of doing it? Earlier, it was done only if the repo ID has changed. Now, it's done all the time.

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,77 @@
...
+                + "Root: " + new File(workingDir).getAbsolutePath() + "\n\n"
+                + "Options: clobber rmdir\n\n"
+                + "LineEnd: local\n\n"
+                + "View:\n"
+                + material.p4view(clientName);
+        return exec("p4", "client", "-i").setExecInput(clientArgs);
+    }
+
+    private BuildCommand sync(String workingDir, String revision) {
+        return exec("p4", "-d", workingDir, "sync", "@" + parseLong(revision));
```
arvindsv: Based on whether the directory is cleaned, etc. there's a [different sync happening](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L101).

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
ibnc: I don't know if I completely agree with that. 

If the password is not blank but we're not using tickets, then we're still using the password. We send it as an argument. see [old code](https://github.com/gocd/gocd/blob/e06f8526b4b3effa2604c96940256a928ff45137/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L158)

Personally, I think it makes more sense to be consistent, and always use the env variable for passwords.

ibnc: @arvindsv

arvindsv: Hmm. Are you sure? Isn't that `line.withNonArgSecret()` call made just to make sure that the password is in the secrets, so that it is not echoed?

ibnc: @arvindsv Ya know, I think you're right. I didn't look close enough

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
ibnc: I don't think the client is enough to execute other p4 commands. The port would still be needed unless I'm mistaken. So, I'd have to pass it along with every p4 exec

I can def see the password be available through an env var as a problem, but I'm not sure the other ones would be?

arvindsv: If the user doesn't need to see the rest of the four, I don't think they should be provided. At this point, we don't have the ability to pass along env vars directly to the "exec". If we get that ability, we can pass all four, every time, along with the call.

arvindsv: I mean, if we added to the exec command, along with "args", an "env" param.

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
ibnc: This comment, and the one above it are related.

Cleaning in that way removes things not in the repository, and reverts changes to the specified revision. So, we don't need to have 2 types of syncs because we don't actually deleted the directory containing the checked out repo

arvindsv: Ok. Does this take care of the p4 repo ID changing? Like the old one does? I can check before merging.

domain/src/com/thoughtworks/go/domain/BuildCommand.java
```diff
@@ -144,7 +144,9 @@ public static BuildCommand error(String message) {
     @Expose
     private BuildCommand onCancel;
     @Expose
-    private String[] execInput;
+    private String[] execInput = new String[0];
+    @Expose
+    private Map<String, String> commandEnvVars = new HashMap<String, String>();
```
arvindsv: At some point later, I might move both of these out to the ExecCommand's inputs, because this feels weird to me. A BuildCommand can be anything, including "test" and "reportCompletion". Adding execInput and commandEnvVars here seems wrong, because they are both specific to ExecCommand only. Am I making sense?

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
ibnc: @arvindsv Here's the ref page for [p4 clean](https://www.perforce.com/perforce/r14.2/manuals/cmdref/p4_clean.html)

From what I can see, it'll do everything we need.... unless I'm missing something

domain/src/com/thoughtworks/go/domain/BuildCommand.java
ibnc: Yeah, that makes total sense. It felt a little weight to me too

ibnc:
@arvindsv Any other comments? Or does this PR look good enough to merge??

arvindsv:
@ibnc - Thanks. I think this is fine. I need to verify that `p4 clean` process and if I find anything, I'll let you know. Apart from that, everything is fine.

domain/src/com/thoughtworks/go/domain/materials/perforce/P4MaterialUpdater.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private Map<String, String> envVars() {
+        Map<String, String> env = new HashMap<>();
+        env.put("P4PORT", material.getServerAndPort());
+        env.put("P4CLIENT", clientName);
+        if (!StringUtil.isBlank(material.getUserName())) {
+            env.put("P4USER", material.getUserName());
+        }
+        if (material.getUseTickets() && !StringUtil.isBlank(material.getPassword())) {
```
arvindsv: Comparing [this](https://github.com/gocd/gocd/blob/e116224b74be46f46d7b66950d4b408fcb665df3/domain/src/com/thoughtworks/go/domain/materials/perforce/P4Client.java#L169) and the line here, it looks like this should be:

```
if (!material.getUseTickets() && !StringUtil.isBlank(material.getPassword())) {
```

Without this, `P4PASSWD` is not set.

arvindsv: @ibnc I've been trying to verify and merge all of the PRs. I've been able to merge (locally) and verify that SVN, Mercurial and Package Material are all fine. This one is giving me some trouble, though. Apart from the issue I mentioned about the `P4PASSWD`, I'm struggling to get this to actually work. Here is my setup:

```
<pipeline name="Test">
  <materials>
    <p4 port="workshop.perforce.com:1666" username="abc_1234" password="password" dest="mat1" materialName="p4a">
      <view><![CDATA[//guest/abc_1234/testp41/main... //client1/...]]></view>
    </p4>
  </materials>
  <stage name="defaultStage">
    <jobs>
      <job name="defaultJob" runOnAllAgents="true">
        <tasks>
          <exec command="ls">
            <arg>-laR</arg>
            <runif status="passed" />
          </exec>
        </tasks>
      </job>
    </jobs>
  </stage>
</pipeline>
```

With the old code, the checkout actually happens and the files are visible. With this new code, the files aren't even checked out. Build passes, though. I wrote an ugly little wrapper over P4 to log everything that's happening. It looks like this:

```sh
#!/bin/bash

file="/tmp/p4.commands.${PPID}"
out="$(mktemp)"
err="$(mktemp)"

trap 'rm -f "$out" "$err"' EXIT

echo "---------------------" >>"$file"
echo "Command: p4 $@" >>"$file"
echo "PWD: $(pwd)" >>"$file"
env | grep '^P4' | sed -e 's/^/  /' >>"$file"
echo >>"$file"

if test -s /dev/stdin
then
   input="$(cat)"
   if [ -n "$input" ]; then
       echo -e "INPUT:\n$(echo "$input" | sed -e 's/^/    | /')" >>"$file"
   fi
fi

echo "$input" | exec /usr/local/bin/p4.old "$@" >"$out" 2>"$err"

echo -e "OUTPUT:\n$(cat "$out")\n" >>"$file"
echo -e "ERR:\n$(cat "$err")" >>"$file"

cat "$out"
cat "$err" >&2
```

I've attached the output of this for the old agent and the new one (I mean, p4 over buildsession -
 P4MaterialUpdater).

If you have time to take a look, that's fine. If not, I can take a look and fix it. Let me know. Thanks!

[p4.new.txt](https://github.com/gocd/gocd/files/1134052/p4.new.txt)
[p4.old.txt](https://github.com/gocd/gocd/files/1134051/p4.old.txt)

Not sure if it's a setup issue on my side, but the fact that this is a "run on all agents" job and it runs properly on the old agent is making me think it's not a setup issue.

arvindsv: Once this works, I plan to check scenarios where the URL is changed, etc.

ibnc: You're totally right!

ibnc:
@arvindsv What's the status on this PR?

arvindsv:
I haven't been able to find anything missing in this. Merging.

arvindsv:
Thank you very much, @ibnc - both for the work on this and the reminder.

PR #3416, opened 2017-04-19
Fix JS errors when navigating to single_page_app pages where controll…

marques-work:
…er_name != "pipeline_configs"

The previous logic caused pipeline_configs.js to load for other single_page_app pages, which would
throw errors as the script tried to perform operations on elements that did not exist.


arvindsv:
Looking at this commit and [this change](https://github.com/gocd/gocd/commit/bfefd6fc133b6146f3e2bbbdcf506412e5f9f963#diff-ece1b629796718ab9f569c1d688c0a49R12), I agree with this. Merging.

@maheshp should take a look too.

PR #3417, opened 2017-04-20
Displaying pipeline schedule time as localtime on dashboard, environm…

jyotisingh:
…ents page and pipeline-history page. If the server is put up in a different timezone, displaying absolute server times on these pages, creates a lot of confusion. Hence the localtime. Upon hover over, the users can see the server time with the timezone information included


server/webapp/WEB-INF/rails.new/app/helpers/pipelines_helper.rb
```diff
@@ -30,7 +30,7 @@ def run_stage_label stage
 
   def trigger_message(actual_date_time, pim)
     if pim.class != com.thoughtworks.go.presentation.pipelinehistory.PreparingToScheduleInstance
-      "Triggered&nbsp;#{l.string('BY')}&nbsp;<span class='who'>#{h pim.getApprovedBy()}</span>&nbsp;on&nbsp;<span class='time'>#{actual_date_time.to_long_display_date_time()}</span><input type='hidden' value='#{actual_date_time}'/>"
+      "Triggered&nbsp;#{l.string('BY')}&nbsp;<span class='who'>#{h pim.getApprovedBy()}</span>&nbsp;on&nbsp;<span class='time'></span><input type='hidden' value='#{actual_date_time}'/>"
```
arvindsv: Do we really need this part? `<input type='hidden' value='#{actual_date_time}'/>`?

jyotisingh: I am not sure why was it there in the first place, so just left it. I can check.

jyotisingh: So, it was added to handle displaying of title on the dashboard (on hover over). We don't use the hidden input field for that anymore, so got rid of the input tag. Mingle story - 4554 (Old issue tracker)

jyotisingh:
I have added a dependency for a new library moment.js to get the times. Its about 16KB so we should be ok I guess. But hoping to get this in before the final performance run to see the impact, else will have to set something up explicitly for this.

arvindsv:
I think it should be fine. The perf run won't really check this, I guess. Won't the JS file be precompiled too, anyway, into application.js? If it's too close, and we really need to, we could put in a flag, just in case.

jyotisingh:
The perf run wont check this. I meant to say that we could use the perf setup to verify if this causes any issues. Yes, it would be precompiled into the application.js. For now, not thinking about a flag. Lets see if we need one.

PR #3423, opened 2017-04-25
Remove console log's loading bar when build is finished

arvindsv:
Remove it even if there's no data available (cancelled before assignment).


arvindsv:
Umm, someone should review this well. Behavior is fine, but I'm not very familiar with this. Also, the animation spikes up CPU quite a bit on both Chrome and Firefox. I tried a few things, but it didn't help. We might have to replace it with an animated image. :( Naveen might be able to create one, if we decide to go down that path.

/cc @marques-work

maheshp:
@marques-work we had users complain bout CPU spike when we used animation on the dashboard, as @arvindsv mentioned we saw a spike on the console logs page as well with animation. For now we are turning off animation as part of https://github.com/gocd/gocd/pull/3424, we might have to think of an alternative.

PR #3427, opened 2017-04-25
Server time on dashboard and history pages are now in the same format…

jyotisingh:
… as that on stage details page ie. 'DD MMM YYYY [at] HH:mm:ss [ZZ]', eg: '23 Apr 2017 at 19:24:01 [+0530]'. The local time on both these pages would read like '23 Apr 2017 at 19:24:01 Local Time'. Decided to use 'Local Time' instead of the timezone name as there is no good way to reliably get the timezone name from the timezone offset because more than one timezone names can be mapped to the same timezone offset.


ankitsri11:
Verified on Go Version: 17.4.0 (4892-ab17b819e73477a47401744fa64f64fda55c26e8)

PR #3433, opened 2017-04-26
The "on" in the timestamp line should probably be lowercase

marques-work:
Even though this word starts the line, it feels more correct to have this word lowercased because it's a continuation of the sentenced started by "Triggered by..."


arvindsv:
Merged because it is a small change. Needs more thought by @jyotisingh and @naveenbhaskar.

PR #3437, opened 2017-04-28
Github Organisation Webhook Support

stephen-murby:
I have started to flesh out an implementation to support organisation webhooks triggering material updates. 

I wanted to get some design / implementation feedback early on, as this may not be the best approach to take.

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/util/command/UrlArgument.java
+++ b/config/config-api/src/com/thoughtworks/go/util/command/UrlArgument.java
@@ -16,10 +16,10 @@
 
 package com.thoughtworks.go.util.command;
 
+import com.thoughtworks.go.config.ConfigAttributeValue;
+
 import java.net.URI;
 import java.net.URISyntaxException;
-
-import com.thoughtworks.go.config.ConfigAttributeValue;
 
 import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
 
@@ -61,7 +61,7 @@ public class UrlArgument extends CommandArgument {
     protected String hostInfoForDisplay() {
         try {
             URI uri = new URI(url);
-            if (uri.getUserInfo()!=null) {
+            if (uri.getUserInfo() != null) {
                 //(String scheme, String userInfo, String host, int port, String path, String query, String fragment)
                 uri = new URI(uri.getScheme(), clean(uri.getUserInfo()), uri.getHost(), uri.getPort(), null, null, null);
             }
@@ -75,7 +75,7 @@ public class UrlArgument extends CommandArgument {
     protected String hostInfoForCommandline() {
         try {
             URI uri = new URI(url);
-            if (uri.getUserInfo()!=null) {
+            if (uri.getUserInfo() != null) {
                 //(String scheme, String userInfo, String host, int port, String path, String query, String fragment)
                 uri = new URI(uri.getScheme(), uri.getUserInfo(), uri.getHost(), uri.getPort(), null, null, null);
             }
@@ -90,7 +90,7 @@ public class UrlArgument extends CommandArgument {
         String[] userAndPassword = userInfo.split(":");
         StringBuilder result = new StringBuilder();
         result.append(userAndPassword[0]);
-        if (userAndPassword.length>1) {
+        if (userAndPassword.length > 1) {
             result.append(":");
             result.append("******");
         }
@@ -102,7 +102,7 @@ public class UrlArgument extends CommandArgument {
     }
 
     public String replaceSecretInfo(String line) {
-        if(forCommandline().length() > 0){
+        if (forCommandline().length() > 0) {
             line = line.replace(hostInfoForCommandline(), hostInfoForDisplay());
         }
 
@@ -126,4 +126,14 @@ public class UrlArgument extends CommandArgument {
         return path;
     }
 
+    public String withoutCredentials() {
+        try {
+            URI uri = null;
+            uri = new URI(url);
+            uri = new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment());
+            return uriToDisplay(uri);
+        } catch (URISyntaxException e) {
+            return url;
+        }
+    }
 }
--- a/domain/src/com/thoughtworks/go/config/materials/ScmMaterial.java
+++ b/domain/src/com/thoughtworks/go/config/materials/ScmMaterial.java
@@ -76,7 +76,7 @@ public abstract class ScmMaterial extends AbstractMaterial {
     }
 
     //most of the material such as hg, git, p4 all print the file from the root without '/'
-    //but subverion print it with '/', we standarize it here. look at the implementation of subversion as well.
+    //but subversion print it with '/', we standarize it here. look at the implementation of subversion as well.
 
     public boolean matches(String name, String regex) {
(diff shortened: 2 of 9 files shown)


server/src/com/thoughtworks/go/server/materials/postcommit/github/GithubPostCommitHookImplementer.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private String decodePayload(Map params) throws UnsupportedEncodingException {
+        String payload = ((String)params.get("payload"));
+        return URLDecoder.decode(payload, "utf-8");
+    }
+
+    boolean isUrlEqual(String paramRepoUrl, GitMaterial material) {
+        String materialUrl = material.getUrlArgument().forCommandline();
+        return validators.perform(paramRepoUrl, materialUrl);
```
stephen-murby: GitHub passes a number of different url fields through in the payload ( url | git_url | ssh_url | svn_url ) each with their respective protocol at the head of the string. 

Is it reasonable to add a material to the prunedCollection regardless of which protocol matches the material url in the cruise config? Doing so means that if you have more than one pipelined configured to point at the same repository, but using a different protocol (http, git, etc...), all of them would be triggered/notified for each push.

arvindsv: For the record, there is some conversation about this here (with @stephen-murby and @ketan): https://gitter.im/gocd/gocd?at=590c5b9b0a783b6c0a68ff8c

ketan: > Is it reasonable to add a material to the prunedCollection regardless of which protocol matches the material url in the cruise config? Doing so means that if you have more than one pipelined configured to point at the same repository, but using a different protocol (http, git, etc...), all of them would be triggered/notified for each push.

Yes. The intent of the webhook is to notify that "something happened".

Additionally, from the payload (for gocd/gocd), it looks like GitHub sends these URLs:

```
    "git_url": "git://github.com/gocd/gocd.git",
    "ssh_url": "git@github.com:gocd/gocd.git",
    "clone_url": "https://github.com/gocd/gocd.git"
```

However, to be human friendly, GitHub allows cloning even with the `.git` suffix on the URLs. So it'd be expected that you generate all combinations of `['git://github.com/', 'git@github.com:', 'https://github.com', 'http://github.com']` and `gocd/gocd` and an optional tailing `.git`.

```
$ ruby -e "puts ['git://github.com/', 'git@github.com:', 'https://github.com/', 'http://github.com/'].product(['gocd/gocd.git', 'gocd/gocd']).collect(&:join)"
git://github.com/gocd/gocd.git
git://github.com/gocd/gocd
git@github.com:gocd/gocd.git
git@github.com:gocd/gocd
https://github.com/gocd/gocd.git
https://github.com/gocd/gocd
http://github.com/gocd/gocd.git
http://github.com/gocd/gocd
```

stephen-murby:
@ketan I have done some refactoring today, and I think the pull request is in a good state; my only concern is that when I ran all of ~2500 units tests, 54 failed. None of them were in a similar location within the code base to where I have been working, lots of them were go-agent tests.

Should I look into fixing those broken tests?

.gitignore
```diff
@@ -152,3 +152,5 @@ out/
 config/config-server/warnings.log
 server/dev
 **/.DS_Store
+/github_organisation.json
```
ketan: Was this accidental?

stephen-murby: Very much so. Good spot

server/src/com/thoughtworks/go/server/materials/postcommit/github/GitHubPushEvent.java
```diff
@@ -0,0 +1,57 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        GitHubPushEvent that = (GitHubPushEvent) o;
+        return Objects.equals(commits, that.commits) &&
+                Objects.equals(repository, that.repository);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(commits, repository);
+    }
```
ketan: Consider using `generate equals and hashcode" in intellij instead of this impl.

server/src/com/thoughtworks/go/server/materials/postcommit/github/GitHubRepository.java
```diff
@@ -0,0 +1,53 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        GitHubRepository that = (GitHubRepository) o;
+        return Objects.equals(url, that.url);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(url);
+    }
```
ketan: Consider using `generate equals and hashcode" in intellij instead of this impl

server/src/com/thoughtworks/go/server/materials/postcommit/github/GitHubRepository.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class GitHubRepository {
+    String url;
+    String fullName;
+
+    public String getUrl() {
+        return url;
+    }
+
+    public void setUrl(String url) {
```
ketan: From what I can see setters are only used within tests. The team prefers using constructor injection instead of setters, and making the instance variables final —

```
class GitHubRepository {
  private final String url;
  private final String fullName;
  public final GitHubRepository() {
    // no-op constructor for GSON
  }

  public final GitHubRepository(String url, String fullName) {
    this.url = url;
    this.fullName = fullName;
  }
}
```

Ditto for `GitHubPushEvent` above.

server/test/unit/com/thoughtworks/go/server/materials/postcommit/github/GitHubRepositoryTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.materials.postcommit.github;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class GitHubRepositoryTest {
```
ketan: It's probably OK to delete this test, the equals and hashcode is only used within tests, from what I can see...

server/src/com/thoughtworks/go/server/materials/postcommit/github/GitHubWebHookMessageParser.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Map;
+
+public class GitHubWebHookMessageParser {
+
+    public GitHubPushEvent parse(Map params) throws UnsupportedEncodingException {
+        String pushEventPayload = urlDecodePayload(params);
+        Gson gson = new GsonBuilder().create();
+        GitHubPushEvent gitHubPushEvent = gson.fromJson(pushEventPayload, GitHubPushEvent.class);
+        // A push event will have a commits field, other event type payloads will not.
+        return gitHubPushEvent.getCommits() == null ? null : gitHubPushEvent;
```
ketan: Since you're not actually making use of `getCommits()` it might make sense to rename this to `hasCommits()` instead.

stephen-murby:
@ketan Thank you for checking that over, some useful pointers. I will fix them up and let you know when I am done.

PR #3440, opened 2017-05-01
Report durations for each task

marques-work:
- ouput duration in milliseconds on the status line
    - server-side durations are the most accurate; trying to guess durations based on timestamp diffs between starting and
      ending task lines, strictly done on the client-side, will not always work -- and here's why:
      - we don't stamp lines with full datetimes, so it's tricky calculating duration when timestamps start before
        and end after midnight
      - without dates, we cannot calculate super-long durations (24+ hrs -- sounds ridiculous, but probably happens in real
        life, unfortunately)
  - foldable section JS should locate and parse raw durations into something human-friendly
  - show parsed duration on the task line so that they are visible when tasks are collapsed


ketan:
Also add exit code, in case of non-zero exit code (separate PR)

marques-work:
@ketan might be good for a separate PR; but yeah that could be useful as sometimes just saying "failed" is not informative enough. I'd also like to get a UX review by @naveenbhaskar on this PR, and perhaps if this is also the best way to show non-zero exit codes too.

marques-work:
@ketan part of my push to make it a separate PR is that I think I have historically sent massive PRs, and I'd like to try to make mine smaller so it's easier to review :)

ketan:
@marques-work — got conflicts after merging the upgrades to ansi_up. Mind resolving?

marques-work:
@ketan rebased and resolved conflicts :)

server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_foldable_section.js
```diff
@@ -64,37 +70,80 @@
...
     }
 
-    function insertBasic(cursor, line) {
-      var output = c("dd", {class: "log-fs-line"}, ansi.ansi_to(line, formatter));
+    function lpad3(num) {
+      return ("00" + num).slice(-3);
+    }
+
+    // no, moment.js doesn't make this easy (at least not without some plugin)
+    function humanizeMilliseconds(duration) {
```
ketan: From a maintenance point-of-view, would much rather prefer to use [moment-duration-format](https://github.com/jsmreese/moment-duration-format).

It seems I can do this, which feels much nicer


```
var x = moment.duration(10000000123, 'ms')
console.log(x.format('d [days] h:m:s.S'))
=> 115 days 17:46:40.123
```

Might mean for us to do something like this, or something better?

```
d = moment.duration(someDuration, 'ms')
if (d <= 1 second) {
  return d.format('s.S [sec]')
} else if (d <= 1 minute) {
  return d.format('m [min] s.S [sec]')
} else if (d <= 1 hour) {
  ...
}
...
```

server/webapp/WEB-INF/rails.new/app/assets/javascripts/log_output_transformer.js
```diff
@@ -124,23 +124,23 @@
 
           if (!currentSection.type()) {
             currentSection.assignType(prefix);
-            currentLine = writer.insertHeader(currentSection, prefix, line);
+            currentLine = writer.insertHeader(currentSection, prefix, timestamp, line);
```
ketan: I'm finding it a bit difficult to understand what this additional does, or was it just a refactoring to cleanup some code?

marques-work: TLDR: yes, it was part of a refactoring/cleanup.

Long story: I moved the timestamp writing to the `FoldableSection.LineWriter` because I intended `LineWriter` to encapsulate all the formatting logic. I felt that `timestamp` could be categorized as so.

Just for some context about intended division of responsibilites:
  - `FoldableSection.Cursor` keeps track of where we are in the DOM, and detects when sections start & end
  - `LogOutputTransformer` is responsible for orchestrating the process by parsing lines, managing the cursor, and delegating the actual line output to `FoldableSection.LineWriter`.

Of course, in theory, I could have parsed out the timestamps in the `LineWriter` instead of `LogOutputTransformer`, but I felt that if LogOutputTransformer is already parsing lines to detect prefixes, it may as well parse timestamps as well and just pass it to the `LineWriter`. I'm not 100% certain that was the right choice, seeing that it's mostly pedantic. Open to a different implementation.

server/webapp/WEB-INF/rails.new/app/assets/stylesheets/new-theme.scss
```diff
@@ -3070,12 +3070,6 @@ button.edit-environment {
   padding:   0;
 }
 
-.buildoutput_pre * {
```
marques-work: yeah, all those styles were duplicated elsewhere, and this selector's weight was overriding the simple class selector I used for duration elements, and effectively clobbered the `font-size` rule I was setting.

server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_foldable_section.js
marques-work: Fair enough. It's probably worth adding the extra plugin for that sake. I'll do that.

marques-work:
@ketan rebased, and builds are green

ketan:
@marques-work — extracted `humanizeMilliseconds` into a plugin by itself that can be reused elsewhere. See https://github.com/gocd/gocd/pull/3445.

ketan:
Here's how it looks like (`sleep` commands FTW) —

screen shot 2017-05-09 at 2 45 09 pm

d-led:
looks nice! Thanks!

Issue #3438, opened 2017-04-28
Agents with websocket connection enabled often move to lost contact status


jyotisingh:
##### Issue Type
- Bug Report

##### Environment
Performance server

##### Description

Agents with websocket connection enabled often move to lost contact status.  No error reported in agent logs.

jyotisingh:
@ibnc , @arvindsv - While running the performance pipeline with websocket connection enabled, noticed the following:

Many of the agents (running with websockets enabled) lost contact with the perf server. Spent a little time with the logs and code. I think I know what's happening. 
Many of the agents have this entry in their logs:
```
2017-04-28 06:32:16,360 [WebSocketClient@267572650-3315] DEBUG thoughtworks.go.agent.AgentWebSocketClientController:266 - [172.16.38.74/172.16.38.74:8154] closed. code: 1001, reason: Idle Timeout
```

So, the idle-timeout for websocket connections is set to 60secs. If the agent is sitting idle for about 60 secs, then the connection times out. When the connection drops, [this](https://github.com/gocd/gocd/blob/a83d1a5858bcd9f6a9d475940cddd38bba4073c1/server/src/com/thoughtworks/go/server/websocket/AgentRemoteHandler.java#L134) line in `AgentRemoteHandler` would mark the agent as 'lost contact'. The loop thread on the agent runs every 10secs, so about 10 secs later when the 'loop' thread is active again, it finds that the web-socket connection has dropped, so it attempts to re-connect ([code](https://github.com/gocd/gocd/blob/a83d1a5858bcd9f6a9d475940cddd38bba4073c1/agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java#L105)). 

```
2017-04-28 06:32:24,463 [loopThread] INFO  thoughtworks.go.agent.WebSocketClientHandler:53 - Connecting to websocket endpoint: wss://172.16.38.74:8154/go/agent-websocket
```

If you go to the agents page on dashboard, and wait for a bit, you would see the agent come up over there. These are also able to pick and execute jobs. All agents go through this cycle. 

[...]

PR #3458, opened 2017-05-04
Retry to send websocket messages if there's an error

ibnc:
When the socket disconnects due to being idle, we still attempt to send messages before waiting for the agent loop to reconnect the websocket. This PR contains retry logic around sending the messages to better ensure we don't lose messages.

This is in relation to https://github.com/gocd/gocd/issues/3438

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/WebSocketSessionHandler.java
+++ b/agent/src/com/thoughtworks/go/agent/WebSocketSessionHandler.java
@@ -65,15 +65,18 @@ public class WebSocketSessionHandler {
     }
 
     private void send(Message message) {
-        for (int retries = 1; retries < 6; retries++) {
+        for (int retries = 1; retries <= systemEnvironment.getWebsocketSendRetryCount(); retries++) {
             try {
-                LOG.info("{} attempt {} to send message: {}", sessionName(), retries, message);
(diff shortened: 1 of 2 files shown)


jyotisingh:
@ibnc - In the current code, if `send` fails, the method would throw an exception which is thrown all the way. With retries added, if the sixth retry fails, the error is not logged anywhere, so looking at the logs, we would have no way of knowing if the 6th try passed or failed. Also, the exception is eaten up, so we need to check all the callers of this method to see if anything will go amiss if the message is not sent but the `send` does not throw an exception. 

The assumption here is, the session would be back in 6 seconds. It might not happen that way. 

Also, instead of retries, would it make sense to wait to send the messages until the connection opens up?

If you want to pursue the retry way, then atleast throw 6 behind a `SystemEnvironment` property, so if there was a need to change the value on performance pipeline, you dont have to wait for a new build to come out.

ibnc:
@jyotisingh You're totally right about the exception. I'll make sure that get's throw after all of the reties have been exhausted.

I thought about going with waiting for the connection to open, but I couldn't think of a good way to implement it. Primarily because we could theoretically wait forever, and we'd never reach the wait that we use to wait for acknowledgement.

PR #3461, opened 2017-05-05
fixing preferences SPA after AuthenticationHelper was moved

marques-work:


marques-work:
@ketan this fixes preferences SPA (https://github.com/gocd/gocd/pull/3360) after this PR was merged https://github.com/gocd/gocd/pull/3456

PR #3462, opened 2017-05-05
BuildSession Package Material Support

ibnc:
Which amounts to simply not throwing an error if the pipeline has a package material.


arvindsv:
Looks ok. I'll merge them all together. I need to verify that the environment variables are set properly. From the code, it looks like they are.

Issue #3375, opened 2017-04-08
Cancel job has no expected effect


xmik:
### Issue Type
- Bug Report

### Summary
When there is a task being already run, I can cancel its job using GoCD UI. I expected that this would terminate the task and so the process on go-agent vm should no longer exist.

#### Reproduction
I have a small GoCD pipeline configuration in yaml:
```yaml
pipelines:
  gocd-cancel-task:
    group: gocd
    label_template: "${git[:8]}"
    materials:
      git:
        git: "git@git.example.com:gocd/gocd-cancel-task.git"
        branch: master
    stages:
      - bash_loop:
          clean_workspace: true
          jobs:
            bash_loop:
              resources:
                - docker
                - small
              tasks:
               - exec:
                  command: bash
                  arguments:
                    - -c
                    - while true; do date; sleep 1; done
```
(The git repository contains only the file with GoCD pipeline configuration).
On GoCD server it resulted in this output:
```
# removed some initial lines
11:01:35.820 [go] Start to build gocd-cancel-task/d759430b/bash_loop/1/bash_loop on small-docker-worker-1 [/var/lib/go-agent/2]
11:01:35.820 [go] Current job status: passed.

11:01:35.820 [go] Start to execute task: <exec command="bash" timeout="0" >
<arg>-c</arg>
<arg>while true; do date; sleep 1; done</arg>
</exec>.
11:01:35.824 Sat Apr  8 11:01:35 UTC 2017
11:01:36.829 Sat Apr  8 11:01:36 UTC 2017
# and so on...
11:01:57.908 Sat Apr  8 11:01:57 UTC 2017
```
Then I cancelled the job through UI and so the output is:
```
11:01:58.013 [go] Start to execute cancel task: NULL
11:01:58.013 [go] Task is cancelled
11:01:58.912 Sat Apr  8 11:01:58 UTC 2017
11:01:59.915 Sat Apr  8 11:01:59 UTC 2017
11:02:00.919 Sat Apr  8 11:02:00 UTC 2017
# and so on...
11:03:40.287 Sat Apr  8 11:03:40 UTC 2017
```
The task is still being run. I ssh'ed into the go-agent vm and there was task process still running:
```
$ ps aux | grep while
go       19446  0.0  0.0  11244  2896 ?        S    11:01   0:00 bash -c while true; do date; sleep 1; done
go       19795  0.0  0.0  12944   976 pts/0    S+   11:02   0:00 grep --color=auto while
```
[...]

PR #3463, opened 2017-05-07
#3375 fix default cancel task when configured via configrepo

tomzo:
A task configured with configrepo, like this
```
              tasks:
               - exec:
                  command: bash
                  arguments:
                    - -c
                    - while true; do date; sleep 1; done
```

Was converted to configuration where on-cancel task was null. But it should have been kill-all.
Added unit test and verified cancel works as expected with yaml configuration.
```
pipelines:
  gocd-cancel-task:
    group: gocd
    label_template: "${git[:8]}"
    materials:
      git:
        git: "git@git.ai-traders.com:gocd/gocd-cancel-task.git"
        branch: master
    stages:
      - style_unit:
          clean_workspace: true
          jobs:
            bash_loop:
              tasks:
               - exec:
                  command: bash
                  arguments:
                    - -c
                    - while true; do date; sleep 1; done
```

### DIFF
--- a/server/src/com/thoughtworks/go/config/ConfigConverter.java
+++ b/server/src/com/thoughtworks/go/config/ConfigConverter.java
@@ -134,7 +134,8 @@ public class ConfigConverter {
 
     private void setCommonTaskMembers(AbstractTask task, CRTask crTask) {
         CRTask crTaskOnCancel = crTask.getOnCancel();
-        task.setCancelTask(crTaskOnCancel != null ? toAbstractTask(crTaskOnCancel) : null);
+        if(crTaskOnCancel != null)
+            task.setCancelTask(toAbstractTask(crTaskOnCancel));
         task.runIfConfigs = toRunIfConfigs(crTask.getRunIf());
     }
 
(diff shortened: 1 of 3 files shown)


tomzo:
Added a small fix to improve logging too https://github.com/tomzo/gocd-yaml-config-plugin/issues/23

server/src/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -134,7 +134,10 @@ public PluggableTask toPluggableTask(CRPluggableTask pluggableTask) {
 
     private void setCommonTaskMembers(AbstractTask task, CRTask crTask) {
         CRTask crTaskOnCancel = crTask.getOnCancel();
-        task.setCancelTask(crTaskOnCancel != null ? toAbstractTask(crTaskOnCancel) : null);
+        if(crTaskOnCancel != null)
```
jyotisingh: @tomzo - does it make sense to call the setter only if `crTask.getOnCancel()` is not null. The only reason being, if we ever decide to change the default value of `AbstractTask.onCancelConfig`, then there wont be multiple places to make the fix.

tomzo: Good point - `AbstractTask.onCancelConfig` had already the kill-all task in there by default. 
I updated the commit, so now task is set only when `crTask.getOnCancel()` is null. Otherwise stays default.
@jyotisingh  please merge if you are fine with it now.

Issue #3472, opened 2017-05-09
Change duration format


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Currently, GoCD formats durations using `hh:mm:ss` format in most places. This is done using the [`fromSecondsToHHMMSS`](https://github.com/gocd/gocd/blob/3bf6e2d6a7a0710abc28ea2859fd22e5240d247c/server/webapp/WEB-INF/rails.new/app/assets/javascripts/time_converter.js#L18) helper method.

We now have an [improved duration formatter](https://github.com/gocd/gocd/blob/3bf6e2d6a7a0710abc28ea2859fd22e5240d247c/server/webapp/WEB-INF/rails.new/spec/javascripts/human_readable_duration_spec.js) that can be used instead.

rajiesh:
Verified on 17.5.0 (5094-facde1c814310d1f9c463c8945c5145a994d0f39)

PR #3445, opened 2017-05-02
Display Duration on VSM for stages

rohitggarg:
As part of software maturity assessment, we would like to extract stats from GoCD about a particular repository's VSM with regards to how much time it takes to either successfully deploy to production or fail.
We also need to extract the data regarding mean time to recover. Due to these requirements, we need a feature in GoCD which can quickly display the times on hover for each stage on the VSM itself.
Thanks,
RohitG

PR #3474, opened 2017-05-09
Display Duration on VSM

rohitggarg:
Per #3445 the approach has a problem on the API side when build state is unknown.
1. Added the check for Build state being unknown
2. Removed the Display part so that the diff is only on the API side.

Please take a cue from this one and update #3472

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -11,6 +11,3 @@ WORKDIR /root/work
 VOLUME /root/.m2
 VOLUME /root/.gradle
 
-ENTRYPOINT '/root/work/gradlew'
-
-CMD ['tasks']
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -1,0 +1,16 @@
+FROM nimmis/java-centos:oracle-8-jdk 
+
+RUN yum install -y git
+RUN yum install -y epel-release 
+RUN yum install -y nodejs npm
+
+ENV JAVA_HOME /usr
+
+WORKDIR /root/work
+
+VOLUME /root/.m2
+VOLUME /root/.gradle
+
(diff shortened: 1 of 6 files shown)


ketan:
Just trying to confirm that you want me to point to this difference between this PR and #3472


```patch
diff --git a/server/webapp/WEB-INF/rails.new/app/models/value_stream_map_model.rb b/server/webapp/WEB-INF/rails.new/app/models/value_stream_map_model.rb
index 069f2dd24f..c4e4b04820 100644
--- a/server/webapp/WEB-INF/rails.new/app/models/value_stream_map_model.rb
+++ b/server/webapp/WEB-INF/rails.new/app/models/value_stream_map_model.rb
@@ -112,7 +112,11 @@ class VSMPipelineInstanceModel
     @counter = counter
     @locator = ""
     @locator = pdg_path_partial.call name, counter unless counter == 0
-    @stages = stages.map { |stage| VSMPipelineInstanceStageModel.new(stage.getName(), stage.getState().to_s, stage.getDuration().getTotalSeconds(), stage.getCounter(), name, counter, stage_detail_path_partial) }
+    @stages = stages.map do |stage|
+      state = stage.getState()
+      duration = stage.getDuration().getTotalSeconds() if state != StageState::Unknown
+      VSMPipelineInstanceStageModel.new(stage.getName(), state.to_s, duration, stage.getCounter(), name, counter, stage_detail_path_partial)
+    end
   end
 end
```

rohitggarg:
That is correct + The tests as well

ketan:
> That is correct + The tests as well

Perfect, I'll lift just the diff I pasted in the comment above, and any corresponding tests.

ketan:
Closing in favor of #3472.

Issue #3407, opened 2017-04-18
Pipeline name should not be capitalized on pipeline history page


rajiesh:
##### Issue Type

- Bug Report

##### Summary

As mentioned by @arvindsv :

On [pages like this one](https://build.gocd.io/go/tab/pipeline/history/gocd-www.go.cd), because of this [line](https://github.com/gocd/gocd/blob/35776381e8e217183c67dfd97d57c3231c8d16c6/server/webapp/WEB-INF/rails.new/app/assets/stylesheets/css/global.scss#L48), the title (pipeline name) is capitalized. That's wrong. What should show as: gocd-www.go.cd shows up as: Gocd-Www.Go.Cd

##### Proposed solution

This issue was existing on older builds too(verified on 16.5.0). Discussed with @naveenbhaskar and decided to make the title as `Pipeline: <pipeline-name>` by preserving the case of the pipeline name

marques-work:
@rajiesh should be fixed now

PR #3485, opened 2017-05-10
Update pipeline history header style

marques-work:
Should fix https://github.com/gocd/gocd/issues/3407

  - don't capitalize the pipeline name
  - prefix the title with "Pipeline: "
    - yes, I cheated by using a :before pseudoelement for a CSS-only solution
      because I did not want to break any functional tests, in case this is validated.
      What can I say, I'm cheap with my commits :)


marques-work:
@naveenbhaskar @arvindsv @rajiesh anyone have time to review this tiny PR?

PR #3487, opened 2017-05-10
Ignore NullPointerException when there is no callback for an acknowledgementId

ibnc:
This related to https://github.com/gocd/gocd/issues/3438

The situation under which the error is thrown is harmless. When the websocket connection times out due to being idle too long, we reestablish the connection. Whenever we establish the connection, the agent clears out all of the pending callbacks waiting for acknowledgement. 

The changes here in this PR will remove the error from the logs

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/WebSocketSessionHandler.java
+++ b/agent/src/com/thoughtworks/go/agent/WebSocketSessionHandler.java
@@ -119,7 +119,11 @@ public class WebSocketSessionHandler {
     void acknowledge(Message message) {
         String acknowledgementId = MessageEncoding.decodeData(message.getData(), String.class);
         LOG.debug("Acknowledging {}", acknowledgementId);
-        callbacks.remove(acknowledgementId).call();
+        try {
+            callbacks.remove(acknowledgementId).call();
+        } catch (NullPointerException e) {
+           LOG.debug("Acknowledgement {} was already acknowledged or removed", acknowledgementId);
+        }
     }
 
     void clearCallBacks() {


jyotisingh:
I am not sure about merging this just yet because we still dont know if its ok to clear out all those pending callbacks. What were we supposed to do with those callbacks in the first place? Will anything go wrong if we do not execute them? I understand too many NullPointerExceptions in the logs is bad, but I think we might just be eliminating the symptom without considering the entire picture.

ibnc:
@jyotisingh I don't know the rationality behind clearing the callbacks upon establishing the connection. 

Here's what I do know.

If we stop clearing the callbacks, then there are two cases for the callbacks that predate the connection being reestablished.
Case 1) The agent receives an acknowledgement
  The callback is then removed, and the waiting thread moves on
Case 2) The agent doesn't receive the callback (because it was sent during the time the socket was closed)
   The thread waiting for that callback will wait the configured time (GO_WEBSOCKET_ACK_MESSAGE_TIMEOUT), then bomb. (See https://github.com/gocd/gocd/blob/master/agent/src/com/thoughtworks/go/agent/WebSocketSessionHandler.java#L95)

I very well might be wrong. Let me know if you see a flaw in my assessment

arvindsv:
@GaneshSPatil - Assigned to you, since you analyzed this and had some opinion about it.

GaneshSPatil:
@arvindsv @jyotisingh @ibnc -- clearing callbacks isn't a good option, as an acknowledgement for console log might not be received and a chunk of log might be missed.

related issue: https://github.com/gocd/gocd/issues/3701

ketan:
@GaneshSPatil ping?

GaneshSPatil:
@ketan, me and @jyotisingh are working on it. 
On a WebSocket reconnect, we shouldn't clear callbacks, as an acknowledgment for console log might not be received and a chunk of log might be missed.This is done [here](https://github.com/jyotisingh/gocd/commit/0cd74da1c53084ebd527d775be6d53c75f75f062#diff-a553f309aeac071663cf5e76ec33abfdL104).

Also, there are other changes to the agent-server-websocket communication.
more info:https://github.com/jyotisingh/gocd/commit/0cd74da1c53084ebd527d775be6d53c75f75f062


@ibnc @jyotisingh can we close this PR?

jyotisingh:
Yes, we should close this one.

PR #3494, opened 2017-05-12
Fixed issue with deletion of gocd role users

bdpiparva:
* Fixed issue with deletion of role users
* Improved error handling on role config SPA

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/webpack/views/roles/roles_widget_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/webpack/views/roles/roles_widget_spec.js
@@ -122,6 +122,12 @@ describe("RolesWidget", () => {
     "properties": []
   };
 
+  const textElements = (selector) => {
+    return $.map($(selector).contents().filter(function () {
+      return this.nodeType === 3;
+    }), $.text);
+  };
+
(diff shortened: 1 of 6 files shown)


server/webapp/WEB-INF/rails.new/spec/webpack/views/roles/roles_widget_spec.js
```diff
@@ -369,26 +375,27 @@ describe("RolesWidget", () => {
...
+      simulateEvent.simulate($('.reveal .role-user').get(0), 'keyup');
+      simulateEvent.simulate($('.reveal .add-role-user-button').get(0), 'click');
+
       $('.reveal .role-user').val("bob");
       simulateEvent.simulate($('.reveal .role-user').get(0), 'keyup');
       simulateEvent.simulate($('.reveal .add-role-user-button').get(0), 'click');
-      m.redraw();
 
-      expect($('.reveal .tag.current-user-tag')).toBeInDOM();
+      expect(textElements('.reveal .tag')).toEqual(["alice", "bob"]);
```
ketan: `expect($('selector')).toContain('alice, bob')` should be sufficient, no?

bdpiparva: $('selector') will return html nodes

PR #3496, opened 2017-05-12
Upgrade JRuby to latest 1.7.27

ketan:
See http://jruby.org/2017/05/11/jruby-1-7-27


ketan:
This apparently breaks because this version of JRuby bundles a newer bouncycastle jar, which breaks stuff. Will merge this PR as part of #3503.

PR #3504, opened 2017-05-12
Fix for gocd role icon on role config spa

bdpiparva:


server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/roles.scss
```diff
@@ -167,6 +167,13 @@ $role-icon-hover-color: #000;
   }
 }
 
+.gocd-role-icon {
+  background:          image-url("cruise.ico") no-repeat center;
```
ketan: Can we use a high resolution png instead of the 16px icon?

https://github.com/gocd/plugin-api.go.cd/blob/master/source/layouts/layout.erb#L24.

bdpiparva: sure

PR #3507, opened 2017-05-12
Separate p4 test data for windows and linux

ibnc:


ibnc:
@ketan Okay, I've got the perforce tests passing on Windows. 

See:
https://build.gocd.io/go/tab/build/detail/build-windows-PR/1398/build-non-server/2/common-split-runInstance-4
https://build.gocd.io/go/tab/build/detail/build-windows-PR/1398/build-non-server/2/common-split-runInstance-3

PR #3513, opened 2017-05-15
Turn off inbuilt ldap passwordfile

jyotisingh:
Disabling Ldap and PasswordFile auth when go.security.inbuilt.auth.enabled is turned off - this includes turning off the providers and disallow save. This is to allow deprecation of inbuilt ldap and passwordfile support, instead of just removing them off. When we decide to drop support, the flag, associated classes, auth-providers and xsd could be updated.

PR build is failing because the functional tests use passwordfile auto. https://github.com/gocd/functional-tests/pull/182 takes care of migrating the existing configs to authConfig format.


server/webapp/WEB-INF/rails.new/app/views/admin/server/index.html.erb
```diff
@@ -93,8 +94,14 @@
                         <span id="password_file_path_error_message"></span>
                     </div>
                 </div>
+                  <%= render :partial => 'shared/form_required_message.html', :locals => {:scope => {}} %>
+
+              <% else -%>
+                  <div class="information">
+                    <span>Support for inbuilt LDAP and Password file have been deprecated in favour of the bundled plugins for the same.</span>
```
jyotisingh: Does this look ok?

screen shot 2017-05-23 at 3 49 42 pm

maheshp: looks good, thanks

PR #3517, opened 2017-05-16
Make TFS work with BuildSession architecture

kyleolivo:
* Mostly just a wrapper around the existing TFSMaterial class.
* Chose not to use the 'tf' command line tool due to previous performance problems (per Duck)
* Created a generic 'Plugin' BuildCommand that is used to invoke the existing TFSMaterial for TFS checkouts. This can be used with Task and SCM plugins in the build session as well.


arvindsv:
@kyleolivo - I remember working with you on this. It looks ok to me, from an idea and implementation perspective. Will need to let @barrowkwan know about these changes, once they're set to be merged.

PR #3520, opened 2017-05-16
Implement console log annotations in BuildSession for colored, foldable logs

marques-work:


common/src/com/thoughtworks/go/buildsession/TaskCommandExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+import java.time.Duration;
+import java.time.Instant;
+
+import static com.thoughtworks.go.util.command.ConsoleLogTags.*;
+import static java.lang.String.format;
+
+/**
+ * Executes a task, reporting its description and result
+ */
+public class TaskCommandExecutor implements BuildCommandExecutor {
```
arvindsv: After having seen the rest of the PR and parts such as [this](https://github.com/gocd/gocd/pull/3520/files#diff-f21aae1907b779d5291d580fc02aa492R103), I see what you're trying to do. But, the whole point of the BuildSession work was to come up with composable units, using which we can build a full "job" or a "build plan". By moving to specialized executors, it seems less pure - and other agent implementations will need to keep changing this.

Are you sure about this? Do you really think that these two are real abstractions in this grammar? I mean, "task" and "jobResult"? Or, does it just make it easier to write the build plan?

arvindsv:
@marques-work If you have any thoughts about my comments, let me know!

ketan:
@arvindsv ping?

GaneshSPatil:
@arvindsv @marques-work -- any updates on this PR?

arvindsv:
@GaneshSPatil I'm going to close this. It obviously needs work and more thought.

Issue #3464, opened 2017-05-08
Stage shows up as passed even when the job failed


jyotisingh:
##### Issue Type
- Bug Report

##### Summary

Stage is showing up as passed when one of the jobs inside it fails.  Basically, the stage failed, but rerun of the job fixed the build. While looking at job instance 1, the stage status should show up as failed and only the second run should show up as passed, but both the runs show up as passed. That seems wrong.

##### Environment
go05

https://go05.thoughtworks.com/go/pipelines/peg-gocd-in-enterprise/138034/defaultStage/1 (Run 1 should show as failure)
https://go05.thoughtworks.com/go/pipelines/peg-gocd-in-enterprise/138034/defaultStage/2 (Run  2 passed)

ketan:
Reproducible on `16.7.0` as well. I took a current DB dump and applied the following SQL to rollback to a schema that's compatible with 16.7.0.

```sql
ALTER TABLE pipelines ADD locked BOOLEAN DEFAULT false NOT NULL;

```

PR #3522, opened 2017-05-17
Use a different SQL to load pipeline history based on stage counter (#3464)

ketan:


ketan:
No tests yet, but I wanted to get @jyotisingh to take a quick look to see if the direction is right.

jyotisingh:
@ketan - Sorry, took me a while to look at this. I think something like this would work.

jyotisingh:
The approach looks good to me. But, I haven't done a thorough analysis of the new queries itself, let me know when this is ready for a full review.

GaneshSPatil:
@ketan @jyotisingh -- any updates on this?

ketan:
No updates, this is a bit risky, right now. I'll work on this over the next few days to be merged in 18.8.

Issue #3519, opened 2017-05-16
Add Role SPA validation is broken


marques-work:
##### Bug Report

Validation on Add Role is incomplete:

- Does not handle validating special characters, instead XHR receives a 500, resulting in "unknown error" message
- Does not fully prevent blank names from being submitted (allows submission of blank on 2nd try, then gives "There was an unknown error performing the operation")

##### Environment
Seen on local dev environment. Haven't tried on build.gocd.io (no privileges).

##### Steps to Reproduce

**Special characters**:
1. Go to http://localhost:8153/go/admin/security/roles and then open the "Add Role" modal dialogue
2. Enter "!" for the name, then save.

**Expected**: display validation message about valid names
**Actual**: displays generic error message reporting unknown error (500 xhr response)

**Handling blanks**:
1. Go to http://localhost:8153/go/admin/security/roles and then open the "Add Role" modal dialogue
2. Enter one or more spaces for the name, then save.
3. Error message "Name must be present" is displayed.
4. Click save again, without changing the input.

**Expected**: validation message "Name must be present" should persist
**Actual**: Modal closes, and a generic error message "There was an unknown error performing the operation" is displayed (500 xhr response)

arvindsv:
@bdpiparva @rajiesh - Is this fixed by #3523? If so, can we close this?

PR #3523, opened 2017-05-17
Validate entity before save or update on SPA

bdpiparva:
This will perform client side validataion of entity on `create` or
`update` call and will not make an ajax call if entity is havin any validation errors.


jyotisingh:
@bdpiparva - something like `foo*` continues to throw the exception mentioned in https://github.com/gocd/gocd/issues/3519

bdpiparva:
@jyotisingh --- Ok will check

PR #3556, opened 2017-05-29
Fix for plugin zip version check

jyotisingh:


server-launcher/build.gradle
```diff
@@ -165,15 +165,14 @@ task defaultCommandSnippetsZip(type: Zip, dependsOn: [':server-launcher:cloneCom
...
-  inputs.property('distVersion', rootProject.distVersion)
   inputs.property('gitRevision', rootProject.gitRevision)
 
   outputs.file file
 
   doLast {
     file.getParentFile().mkdirs()
     file.withWriter { out ->
-      out.write("${rootProject.version}(${rootProject.distVersion}-${rootProject.gitRevision})")
+      out.write("${rootProject.version} (${rootProject.gitRevision})")
```
ketan: This file is also used in backups, not sure if it affects the outcome.

base/src/com/thoughtworks/go/util/ZipUtil.java
```diff
@@ -128,6 +129,24 @@ public void unzip(ZipInputStream zipInputStream, File destDir) throws IOExceptio
...
+            while (zipEntry != null) {
+                if (zipEntry.getName().equalsIgnoreCase(filePath)) {
+                    extractTo(zipEntry, zipInputStream, tempFolder);
+                    return FileUtil.readContentFromFile(new File(tempFolder, zipEntry.getName()));
+                }
+                zipEntry = zipInputStream.getNextEntry();
+            }
+        } finally {
+            IOUtils.closeQuietly(zipInputStream);
+            FileUtils.deleteQuietly(tempFolder);
```
ketan: See https://stackoverflow.com/questions/15667125/read-content-from-files-which-are-inside-zip-file for doing this without an intermediate tempfile..

server-launcher/build.gradle
jyotisingh: what backups @ketan? Did you mean `BackupService`? If not, could you point it out?

base/src/com/thoughtworks/go/util/ZipUtil.java
jyotisingh: sure

ketan:
> what backups @ketan? Did you mean BackupService? If not, could you point it out?

Yes, `BackupService`.

jyotisingh:
`BackupService` writes to a file by the same name ie. version.txt and places it inside the current backup so that users would be aware of which version was the server on when the corresponding backup was taken, but its totally unrelated to this.

jyotisingh:
https://build.gocd.io/go/pipelines/value_stream_map/trigger/1235
@ketan - good to merge?

PR #3571, opened 2017-06-06
Download filebased authentication plugin from latest git release

bdpiparva:


bdpiparva:
@jyotisingh - We will still make a release when there is a issue with plugin right ?

jyotisingh:
Correct, but the release would be made once the QAs are happy with it. For Development purpose you would also want to use the latest one, no? 
BTW, why did you want to make this change? Does this not work in some scenario?

ketan:
> BTW, why did you want to make this change? Does this not work in some scenario?

Yes — the curl command has been failing for quite some time now. and when it fails, it does not blow up for lack of `--fail` argument. Also, we've changed the domains, so the URL needs to be changed to build.gocd.org

ketan:
IMO all that is needed is this —

```patch
diff --git a/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java b/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
index 678f2686b8..117b614c0f 100644
--- a/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
+++ b/development-utility/development-server/src/com/thoughtworks/go/server/DevelopmentServer.java
@@ -89,7 +89,7 @@ public class DevelopmentServer {
         if (passwordFilePluginJar.exists()) {
             System.out.println("Found a local copy of passwordFile plugin, using it.");
         } else {
-            new ProcessRunner().command("curl", "-L", "https://build.gocd.io/go/files/plugins/latest/build/latest/file-authentication-plugin/filebased-authentication-plugin.jar", "-u 'view:password'", "--output", passwordFilePluginJar.getAbsolutePath()).failOnError(true).run();
+            new ProcessRunner().command("curl", "--fail", "--silent", "--location", "https://build.gocd.org/go/files/plugins/latest/build/latest/file-authentication-plugin/filebased-authentication-plugin.jar", "--user", "view:password", "--output", passwordFilePluginJar.getAbsolutePath()).failOnError(true).run();
         }
         new ZipUtil().zipFolderContents(pluginsDist, new File(classpath(), "plugins.zip"));
     }
```

Issue #3577, opened 2017-06-08
Avoid starting up gocd windows service by default


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Currently installing the GoCD agent on windows will bring up the agent process, and create a guid.txt and possibly autoregister with the server (if the file is present).

This makes it very difficult to bake an AMI out of the windows instance with just the gocd agent and relevant configuration to bring up the instance later.

rajiesh:
Verified on 17.6.0 (5142-be1ff52e07d80323fcc615864a64f3afe83b7016)

PR #3585, opened 2017-06-14
Feature/skip agent start on win install

drewsonne:
Added a switch to allow the agent not to be started after install on windows agents.

Used in this case for building AMI's.


drewsonne:
This addresses #3577

ketan:
There's a [build](https://build.gocd.org/go/pipelines/value_stream_map/build-linux/1717) that kicked off with your changes.

You should see the installers show up at https://www.gocd.org/download/?experimental=true in a couple of hours with version `17.6.0-5114`.

PR #3596, opened 2017-06-19
set merging order of config partials matching order of appearing in xml

stCarolas:
Case: there are few pipelines, defined in config-repos. One of them (lets call it pipeline A) depends on another (pipeline B). When server loads config, it checks and merges every pipeline one by one. Pipeline internally stores in hashsets, so they merge with main config in random order - and server throw "Pipeline not exists" exception if pipeline A was merged before pipeline B 

Tests would be later, when I would understand test structure

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/BasicCruiseConfig.java
@@ -45,12 +45,14 @@ import java.util.concurrent.ConcurrentMap;
 
 import static com.thoughtworks.go.util.ExceptionUtils.bomb;
 import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
+import org.apache.log4j.Logger;
 
 /**
  * @understands the configuration for cruise
  */
 @ConfigTag("cruise")
 public class BasicCruiseConfig implements CruiseConfig {
+    private static final Logger LOGGER = Logger.getLogger(BasicCruiseConfig.class);
     @ConfigSubtag
     @SkipParameterResolution
     private ServerConfig serverConfig = new ServerConfig();
@@ -115,6 +117,7 @@ public class BasicCruiseConfig implements CruiseConfig {
             return;
         }
         partList = removePartialsThatDoNotCorrespondToTheCurrentConfigReposList(partList);
+        LOGGER.trace("[ Config Merge ] corresponded list of partials: " + partList);
 
         if (this.strategy instanceof MergeStrategy)
             throw new RuntimeException("cannot merge partials to already merged configuration");
@@ -1055,8 +1058,8 @@ public class BasicCruiseConfig implements CruiseConfig {
     }
 
     private Set<MaterialConfig> getUniqueMaterials(boolean ignoreManualPipelines, boolean ignoreConfigRepos) {
-        Set<MaterialConfig> materialConfigs = new HashSet<>();
-        Set<Map> uniqueMaterials = new HashSet<>();
+        Set<MaterialConfig> materialConfigs = new LinkedHashSet<>();
+        Set<Map> uniqueMaterials = new LinkedHashSet<>();
         for (PipelineConfig pipelineConfig : pipelinesFromAllGroups()) {
             for (MaterialConfig materialConfig : pipelineConfig.materialConfigs()) {
                 if (!uniqueMaterials.contains(materialConfig.getSqlCriteria())) {
@@ -1072,6 +1075,7 @@ public class BasicCruiseConfig implements CruiseConfig {
         }
         if (!ignoreConfigRepos) {
             for (ConfigRepoConfig configRepo : this.configRepos) {
+                LOGGER.trace("[ Config Merge ] check for unique material: " + configRepo);
                 MaterialConfig materialConfig = configRepo.getMaterialConfig();
                 if (!uniqueMaterials.contains(materialConfig.getSqlCriteria())) {
                     materialConfigs.add(materialConfig);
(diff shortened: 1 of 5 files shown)


arvindsv:
/cc @tomzo @jyotisingh - They know this best. My feedback:

1. Use slf4j logger instead.

2. With that, use: `LOGGER.trace("Some message {}", material)`, which prevents the `toString` from being called unless trace logging is turned on.

Apart from that the idea of using a LinkedHashSet makes sense to me.

arvindsv:
@tomzo - If you agree with this, please leave a comment rather than merging it. It's possible we might do a security-related fix. That reminds me, that you're not on that list. I'll add you now.

tomzo:
Hi @stCarolas 

> When server loads config, it checks and merges every pipeline one by one. Pipeline internally stores in hashsets, so they merge with main config in random order - and server throw "Pipeline not exists" exception if pipeline A was merged before pipeline B

You are correct that merging order is random. But let's be clear on what happens on **server start**, in short
1. A thread pool of several MDU workers is allocated, which (among other things) will be parsing config repos. We have no control of how long that will take and which repo will end first.
2. When any config repo parsing ends, a merge is attempted. Regardless if it fails or not, that partial is cached until next merge.
3. At some point all repos are parsed and valid merged config is accepted.

I think  your changes will improve the current situation, (assuming that you have config-repos in order of dependencies), but there is no guarantee. In your example with A,B pipelines, this is what would happen:
1. A, then B are both put on a queue of materials scheduled for update.
2. There are many threads unloading that queue, and since A was placed there a bit faster, it might be picked up from the queue before B.
3. Parsing A and B are basically racing now.
4. Maybe A finishes before B, because it started earlier. Maybe not..

Anyhow these changes are not dangerous in any way. Tests expecting that finishing order is always the same as order of config repos would be flaky. But you can test that material updates were **scheduled** in order.

jyotisingh:
I echo what @tomzo says about this. The PR assumes the pipeline dependency is in the same order in which the config-repos are defined in the XML. In many cases you cannot really define a load order for config repos when they contain inter-dependent pipelines. 
@stCarolas you are right about the error message showing up when the server starts up, but that error should eventually go away as the rest of the config repos get loaded up. Doesn't that happen? Or are you trying to make sure that the error doesn't show up at all?

stCarolas:
@arvindsv  slf4j -ok, simply made it like others in the module. Also dont seen lobmok, no objections against it?

@tomzo @jyotisingh  in my case (a go-server in a docker container and a private git server) config loading and parsing took about 10-30 sec per config, about 2-3 minutes totally for 9 pipelines. De-facto the parsing was done one by one, no parallels and no racing.
No, the error didnt disappear after 4 hours and only 6 pipeline from 9 was merged and shown on a dashboard. I think the config-repos are periodically scheduled for update, an update of corresponding partial happens, but this update doesnt trigger merge with the main config. I can send a matching log.
So, questions:
1) Only one try for merging a config-repo is expected?
2) If using LinkedHashSet doesnt seem like a stable solution, how to fix the situation? I think giving another try to merge is dangerous because it can take a lot of time and resources for a complicated dependency tree (and the startup merge error would be shown for a long time in this case, that may embarrassed user). Maybe some tag in the xml like bulk-container, which should load a nested config-repo as one config update command?

tomzo:
> No, the error didnt disappear after 4 hours and only 6 pipeline from 9 was merged and shown on a dashboard

Then you definitely have invalid config references, pipeline dependencies or alike. I would suggest that you remove all config repos from xml, then add them one by one. Starting from the one with no dependencies. There will be eventually errors after adding one of them. The error message will tell what is wrong, more or less.

> 1. Only one try for merging a config-repo is expected?

Merge of all repos and main is attempted each time when any of the parts has changed.

> 2. how to fix the situation?

You need to fix your configuration. There is no bug in GoCD that would not allow you to make complex dependencies between config repos. The only limitation is that you cannot reference pipeline defined in config repo from xml. 

> config loading and parsing took about 10-30 sec per config, about 2-3 minutes totally for 9 pipelines

That sounds very long. I have a server which has about 160 config repos, about 400 pipelines and startup takes 4-5 minutes.

> Maybe some tag in the xml like bulk-container, which should load a nested config-repo as one config update command?

That is an option. We decided to not do it initially, because most users don't care and actually do not want to specify dependencies between config repos explicitly.
Recently I thought that we could introduce something like config scopes:
```xml
<config-repos>
<config-scope references-to-xml-allowed="true">
<config-repo>...</config-repo>
<config-repo>...</config-repo>
</config-scope>
<config-scope references-to-xml-allowed="false">
<config-repo>...</config-repo>
<config-repo>...</config-repo>
</config-scope>
</config-repos>
```
Each scope would be self-sufficient, it could be parsed and validated in its own context without others participating.

arvindsv:
> ok, simply made it like others in the module. Also I don't see a lombok there, no objections against adding it, generally?

I'd rather not add it (lombok), for a small change like this.

jyotisingh:
@stCarolas - what we are trying to say is if all the configs together are valid, then GoCD should be able to eventually load all of them up. As of now, we aren't aware of any bug that would cause one of the config-repos to not load up because of the load order. So, at least it seems that may be one of your configs is infact invalid. If so, the 'errors & warnings' popup should have the details. Could you check what does it say. Also, have you tried your config with this PR, does everything load up as expected?

stCarolas:
@jyotisingh @tomzo understood. 

The 'Errors & Warnings' popup says about errors in another pipelines (not interested), also it says

> Pipeline "TARIFF_API_MASTER" does not exist. It is used from pipeline "DEPLOY_TARIFF_TO_INT".;; Pipeline with name 'TARIFF_API_MASTER' does not exist, it is defined as a dependency for pipeline 'DEPLOY_TARIFF_TO_INT'

But TARIFF_API_MASTER is shown on the dashboard
Yes, I had tried my config with this PR and yes, these two pipelines were loaded as expected. Actually my team works with my build of the go-server

Anyway, I think first I should write (or find one) a test for this case (on weekend, I think)- this discussion without the test is too abstract

tomzo:
I am closing this.
As said before it's conceptual error to think that there is a fixed order of loading config-repos.
Even so if there was a bug with pipelines not getting to valid state *eventually* then more users would report it.

PR #3601, opened 2017-06-20
Add a `webhookSecret` attribute on `<server/>` element

ketan:


config/config-api/src/com/thoughtworks/go/config/ServerConfig.java
```diff
@@ -37,6 +37,7 @@
     @ConfigAttribute(value = "purgeUpto", optional = true, allowNull = true) private Double purgeUpto;
     @ConfigAttribute(value = "jobTimeout", optional = true) private String jobTimeout = "0";
     @ConfigAttribute(value="agentAutoRegisterKey", optional = true, allowNull = true) private String agentAutoRegisterKey;
+    @ConfigAttribute(value="webhookSecret", optional = true, allowNull = true) private String webhookSecret;
```
stephen-murby: Maybe this should be optional false, and allowNull false? Given that it will be a generated value.

arvindsv:
@ketan Looks good to me. The comment by @stephen-murby is relevant. I guess it depends on the way the config validation happens. Either way, I think this will end up working, since the `ensure...` method will get called.

arvindsv:
Some conversation about this here: https://gitter.im/gocd/gocd?at=59492409ad5b0377046ec41e

config/config-api/src/com/thoughtworks/go/config/ServerConfig.java
ketan: The `ensureWebhookSecretExists` will be invoked on a config save everytime, so although it's marked `optional`, a value will be generated even if intentionally left blank.

PR #3605, opened 2017-06-22
Filter to restrict GoCD access from Artifacts 

bdpiparva:
Added ArtifactFilter to ensure that artifact rendered in browser do not have access to any other resources apart from other artifacts.


server/webapp/WEB-INF/applicationContext-acegi-security.xml
```diff
@@ -180,6 +180,7 @@
     <!-- End CAS Stuff -->
 
     <bean id="removeAdminPermissionFilter" class="com.thoughtworks.go.server.security.RemoveAdminPermissionFilter" autowire="autodetect"/>
+    <bean id="artifactsFilter" class="com.thoughtworks.go.server.security.ArtifactsFilter" autowire="autodetect"/>
```
arvindsv: `denyNonArtifactCallsFromWithinArtifactsFilter`?

bdpiparva: What about `ArtifactRequestingNonArtifactDenialFilter` ?

arvindsv: I prefer `artifactsFilter` to that. :) The one I suggested seems easier (for me) to understand what is happening. In this case, I'll need to do some mental gymnastics to try and understand what it means.

The problem, imo, is that the "ArtifactRequestingNonArtifact" is the part you want someone to read together and then the "DenialFilter" adds to that. However, when I read it first, I read it as: "ArtifactRequesting" (pause) "NonArtifactDenialFilter" - and got lost.

maheshp: I named it `DenyGoCDAccessForArtifactsFilter` sounds ok?

arvindsv:
Looks good to me.

server/src/com/thoughtworks/go/server/security/ArtifactsFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import static org.apache.commons.lang.StringUtils.isBlank;
+import static org.springframework.security.ui.FilterChainOrder.EXCEPTION_TRANSLATION_FILTER;
+
+public class ArtifactsFilter extends SpringSecurityFilter {
+
+    @Override
+    protected void doFilterHttp(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
+
+        if (isRequestFromArtifact(request) && !requestingAnArtifact(request)) {
```
4cad: You should probably allow the login URL and the homepage, otherwise it will loop on redirect to the login until the browser realizes it is looping and terminates the loop.

bdpiparva: @4cad  --- Thanks for the feedback. This filter will block request that came from an artifact and requesting non artifact resource. We tried following scenarios:

1) Accessing login or home page from an artifact: Filter will block any non artifact calls from an artifact as I mentioned above.
     * In following screenshot `attack_alert.html` is an artifact accessing login and home page through ajax calls.

        screen shot 2017-06-23 at 11 04 43 am

        (Assuming, this will anser you question)     


2) Accessing an artifact after session termination:
    a. Open an artifact
    b. From browser use different tab and, end the current session
    c. Reload artifact page[a]

    On reload of artifact page user will be redirected to login page. After successful authentication user will be redirected back to the artifact page
   
3) Accessing artifact from API with credentials: Working as expected

Let me know, If you need more clarification.

4cad: Yes, AJAX doesn't follow redirects. My point was that if you click on a link on an artifact oage that points to a non-artifact Go page it triggers a redirect loop.

I just wanted to bring this to your attention, it is probably a low-value place to focus anymore work.

screen shot 2017-06-23 at 7 25 30 am

Issue #995, opened 2015-04-01
Go Agent not using proxy settings from Java Configuration


cross311:
My team has been using Go-Cd for our AWS resources for a number of months now and have finally gotten approval to have Go deploy our private datacenter resources.  Our datacenter is setup to only allow outbound traffic through a proxy.  This proxy is not invisible (not automatically used) it must be set via internet options or the (java control panel)[https://www.java.com/en/download/help/proxy_setup.xml].

After setting up the proxy on the machine that has the go agent installed, the installation still fails with connection timed out.  We also see no activity in the proxy logs as to it attempting to use the proxy.  We are able to access the go server via internet explorer on the machine due to it running requests properly through the outbound proxy.

Has anyone else experienced issues with the go agent not using proxy settings?  Is there a parameter that I am missing?

For information I even tried to edit the agent.cmd to pass in the proxy info when executing the registration phase.

``` sh
"%JAVA_CMD%" -Dhttp.proxyHost=%proxyhostURL% -Dhttp.proxyPort=%proxyPortNumber% -jar "%AGENT_DIR%\agent-bootstrapper.jar" %GO_SERVER% %GO_SERVER_PORT%
```

Any information or help would be much appreciated.

arvindsv:
Can't claim to have ever tried it, but the Go Agent connects first to port 8153 of the server, using HTTP (only one request) and then, immediately connects to port 8154 using HTTPS. You might have some luck setting https.proxyHost. Can't say for sure.

cross311:
@arvindsv Thanks for the quick reply.  We tried setting both.  However the agent launcher does not get past the original http request over 8153.

matt-richardson:
Have you considered switching to a transparent proxy?
World of pain with old school proxies.

cross311:
@matt-richardson Yes actually I have proposed this change.  However the folks in charge of the datacenter are not willing to change the architecture for this.

arvindsv:
Took a quick look at this, and it turns out that Apache HTTP Client 3.1, which is used by Go internally to setup a connection between the Go Agent and the Go Server, does not honor http.proxyHost. Reference: [This](http://www.mail-archive.com/httpclient-user@jakarta.apache.org/msg02586.html).

[...]

PR #3613, opened 2017-06-27
Add support for http proxy

harsimranmaan:
It is now possible to run the agent and server in
an environment where network access requires a proxy.
While this has not been tested to work against a proxy
that requires authentication, it is an extraction of a
setup which works with a proxy setup.

Set go.proxy.url=http:/proxy.host.com:8080 to enable
proxy access.

Addresses #995

### DIFF
--- a/base/src/com/thoughtworks/go/agent/common/ssl/GoAgentServerHttpClientBuilder.java
+++ b/base/src/com/thoughtworks/go/agent/common/ssl/GoAgentServerHttpClientBuilder.java
@@ -18,6 +18,7 @@ package com.thoughtworks.go.agent.common.ssl;
 
 import com.thoughtworks.go.util.SslVerificationMode;
 import com.thoughtworks.go.util.SystemEnvironment;
+import org.apache.http.HttpHost;
 import org.apache.http.config.SocketConfig;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.conn.ssl.TrustStrategy;
@@ -64,7 +65,11 @@ public class GoAgentServerHttpClientBuilder extends GoAgentServerClientBuilder<C
         }
 
         sslContextBuilder.loadKeyMaterial(agentKeystore(), keystorePassword().toCharArray());
-
+        String proxyUrl = systemEnvironment.getProxyUrl();
+        if (!proxyUrl.isEmpty()) {
+            HttpHost proxy = HttpHost.create(proxyUrl);
+            builder.setProxy(proxy);
+        }
         SSLConnectionSocketFactory sslConnectionSocketFactory = new SSLConnectionSocketFactory(sslContextBuilder.build(), hostnameVerifier);
         builder.setSSLSocketFactory(sslConnectionSocketFactory);
         return builder.build();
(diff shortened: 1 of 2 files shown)


ketan:
If we're baking support for proxies, I'd rather that it be complete to support:

* proxy auth
* SOCKS proxy 

There's a bunch of examples at http://hc.apache.org/httpcomponents-client-ga/httpclient/examples/org/apache/http/examples/client/ that you may find inspiration from to build support for these things.

You will also need to test these with proxies that support auth (docker works great for such situations). Please paste the docker/proxy config files so you (and the gocd team) can test it out with proxy support with a simple `docker run -v /path/to/some.conf:/etc/proxy/some.conf -p.... CONTAINER_ID`

Thanks

harsimranmaan:
@ketan, I deliberately did not try to  support proxy from the env(http[s]_proxy,no_proxy,all_proxy)/conf as I wasn't sure if there would be enough demand. Hence, I used the go namespace for the property name. I'd be happy to add more cases but that would run the risk of breaking backwards compatibility unless we guard the proxy toggle with another custom property.
eg: We would not want to break the installations where the environment has a proxy config defined but where go-cd went from ignoring something like `/etc/sysconfig/proxy` to honoring it after we started reading in the config.
Let me know if the complete support is needed indeed and if that is something we want to do in one shot or assess the usage before committing to the complete implementation. I'd be happy to contribute more patches on this.

On a lighter note, I was a bit surprised to learn that go-cd is written in java instead of golang ;).

ketan:
My intention was to not configure proxy via environment variables, but via system properties. The reason being that these system properties are universally understood by most things java. (GoCD being written in java, instead of go-lang. Speaking of which, we had the name go, before golang ;).

My idea was to support proxies via the [well documented system properties](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html). These are the equivalent of the `*_proxy` environment variables you mention above. Since plugins that run on GoCD may also need to make network calls, they will most likely support these system properties out-of-the-box.

* http.proxyHost
* http.proxyPort
* http.nonProxyHosts
* https.proxyHost
* https.proxyPort
* socksProxyHost
* socksProxyPort


My guess is that we need the equivalent of [these lines](https://github.com/apache/httpcomponents-client/blob/4.5.2/httpclient/src/main/java/org/apache/http/impl/client/HttpClientBuilder.java#L1147-L1148):

```java
public class GoAgentServerHttpClientBuilder ... {
    public CloseableHttpClient build() throws Exception {
        HttpClientBuilder builder = HttpClients.custom();
        if (useProxy()) {
           builder.setRoutePlanner(new SystemDefaultRoutePlanner(schemePortResolverCopy, ProxySelector.getDefault()));
        }
    }

    boolean useProxy() {
        // check if the variables above are defined
    }
}
```

ketan:
@harsimranmaan 

Are you still interested in contributing to this PR?

ketan:
Closing as there is no interest in this PR.

dnewhook:
Hi there - new to github so bare with me.  Is this feature available in the current gocd release?  Where do I need to set go.proxy.url?  Would it be in the AGENT_STARTUP_ARGS as "-Dgo.proxy.url="? Very interested in being able to configure agents to use a forward proxy in order to reach the server.

ketan:
This is not available as the feature has not been implemented. I'm happy to
provide you an experimental build if you are willing to help test it.

dnewhook:
I'd be happy to help test an experimental build.

ketan:
Closing in favor of #3847

ketan:
A build is still running, you should see some installers in another 10 minutes or so at https://build.gocd.org/go/tab/build/detail/installers-PR/633/dist/1/dist. Please make sure to use both the server and agent from that build.

To connect an agent through the proxy, you'd need to setup 2 variables on your agent:

```
AGENT_BOOTSTRAPPER_JVM_ARGS='-Dhttp.proxyHost=... -Dhttp.proxyPort=...'
AGENT_STARTUP_ARGS='-Dhttp.proxyHost=... -Dhttp.proxyPort=...'
```

Issue #3608, opened 2017-06-23
Add ability to provide plugin order for authentication 


maheshp:
##### Issue Type

- Feature enhancement

##### Summary

In `17.5.0` GoCD core support for File based and LDAP/AD authentication is moved to plugins which implement the new Authorization Plugin Endpoint. Previously GoCD provided support for these authentication mechanism through Spring `AuthenticationProvider` which had a specific order in which they were executed, the FileBased AuthProvider was executed before the LDAP provider. 

With authentication moved to plugins there is no way for users to define the order of plugins to be used for authentication which is undesirable.

Possible solutions,
-  Honour the order of `<authConfig>` defined in the config while authenticating.
- To be more explicit give an option in `<authConfig>` to define the order, 
`<authConfig id="file" pluginId="cd.go.authentication.passwordfile" order="1">`
`<authConfig id="ldap" pluginId="com.thoughtworks.gocd.authorization.ldap" order="2">`

ketan:
i thought we were authentication in the order defined in the config file. And was assuming there would be some ability to "drag and re-order" instead of explicitly setting an `order` attribute..

maheshp:
No we are not using the order defined in the config file. We spoke about few options for ordering but nothing was finalised. This issue can help us capture the desired behaviour.

maheshp:
After discussing with @arvindsv we have decided on relying on `<authConfig/>` order in config_xml to determine the order in which plugins will to be used for authentication.

```
<authConfigs>
  <authConfig id="file1" pluginId="cd.go.authentication.passwordfile">
   ...
  </authConfig>
  <authConfig id="ldap" pluginId="cd.go.authentication.ldap">
  ...
  </authConfig>
  <authConfig id="file1" pluginId="cd.go.authentication.passwordfile">
  ...
  </authConfig>
</authConfigs>
```
For the above example authConfigs, the order in which plugins would be called for authentication would be,
1. Password file plugin with authConfig id `file1`
2. LDAP plugin with authConfig id `ldap`
3. Password file plugin with authConfig id `file2`

jyotisingh:
Sounds fair. Just one thing to point out. Earlier on password file auth used to happen first and then ldap. Config migration migrates ldap first (91.xsl) and then password-file (92.xsl). So, the order would be inverted from what it used to be. Not sure if that would be an issue, just thought I would point out.

maheshp:
[...]

PR #3626, opened 2017-06-30
Ensure order of auth_config is used to authenticate users #3608

maheshp:
* With this commit the auth_config order in config xml is used to
  derive the order in which plugins are used for authentication.


jyotisingh:
Pending: Update documentation of plugin-api to specify that the authConfigs passed to `authenticateUser` call would have just one authConfig, and the roleConfigs would be the roles corresponding to the authConfig, so the plugins need not do further processing on these.

PR #3629, opened 2017-07-03
Remove tfs-impl-10

varshavaradarajan:
@arvindsv , @ketan mentioned you had some concerns around this.


server/test/unit/com/thoughtworks/go/server/controller/AgentRegistrationControllerTest.java
```diff
@@ -65,7 +65,6 @@ public void setUp() throws Exception {
         goConfigService = mock(GoConfigService.class);
 
         when(systemEnvironment.getSslServerPort()).thenReturn(8443);
-        when(systemEnvironment.get(SystemEnvironment.TFS_SDK_10)).thenReturn(true);
```
jyotisingh: Won't removing that expectation break any of the tests? May be trigger a PR build, but also run the tests locally for quick feedback

varshavaradarajan: Removed the checks for the system property. Will keep an eye on the PR build.

jyotisingh:
@varshavaradarajan - PR looks good, will merge after regression-tests passes.
Pending: https://github.com/gocd/docs.go.cd/issues/99
As far as I remember, we did not announce the JVM arg that is being removed as a part of this PR, but double check if the documentation for the same.

varshavaradarajan:
Regression passed, @jyotisingh. Will make the doc changes.

Issue #2024, opened 2016-03-08
Remove unused asset files


ketan:
There seem to be a lot of unused assets (indirectly referenced by unused css).

I ran a quick script on the apache logs on build.go.cd to see what assets were actually served out (since Oct 11).

```
$ zcat *-access.log*.gz | grep ' 200 ' | grep /go/assets/ | grep -Ev '.(css|js)' | awk '{print $7}' | cut -f1 -d'?' | sed -e 's/-[[:xdigit:]]\{64\}//g' -e 's/-[[:xdigit:]]\{32\}//g' | sort -u

/go/assets/andare/comment-triangle.png
/go/assets/andare/gears-loader.gif
/go/assets/andare/ico-16x16-alerts.png
/go/assets/andare/ico-16x16.png
/go/assets/andare/ico-20x16-webkit-select.png
/go/assets/andare/icon-mix.png
/go/assets/andare/icon-plugin-default-bundled.png
/go/assets/andare/icon-plugin-default.png
/go/assets/andare/nav-sprites-pipeline.png
/go/assets/andare/nav-sprites.png
/go/assets/andare/nav-user-triangle.png
/go/assets/andare/thumbs-materials-100x100.png
/go/assets/arrow_closed.jpg
/go/assets/arrow_open.jpg
/go/assets/back-to-top.png
/go/assets/bg_agents_page_header.jpg
/go/assets/bg_page_intro_arrow.gif
/go/assets/bg-texture-dark.png
/go/assets/cruise.ico
/go/assets/directory.gif
/go/assets/file.gif
/go/assets/font-awesome/fontawesome-webfont.eot
/go/assets/font-awesome/fontawesome-webfont.svg
/go/assets/font-awesome/fontawesome-webfont.ttf
/go/assets/font-awesome/fontawesome-webfont.woff
/go/assets/font-awesome/fontawesome-webfont.woff2
/go/assets/g9/backgrounds/bg_actionpanel_shadow.png
/go/assets/g9/backgrounds/bg_glass_gradient.png
/go/assets/g9/backgrounds/bg_gradient_50.png
/go/assets/g9/backgrounds/bg_header_default.png
/go/assets/g9/backgrounds/bg_state_unknown_big.png
/go/assets/g9/backgrounds/bg_state_unknown.png
/go/assets/g9/backgrounds/bg_status_diagonal.png
/go/assets/g9/backgrounds/bg_status_gradient.png
/go/assets/g9/backgrounds/details_stage_selected.png
[...]

PR #3631, opened 2017-07-03
Removed some unused assets.

varshavaradarajan:
Ran the following on the assets file attached to get the unused assets.
I have not deleted everything shown by the below command. To be discussed with @jyotisingh 
```
while read i; do; echo -n "$i "; git grep -q $i; echo $?; done <~/assets.txt | grep " 1"
```
[assets.txt](https://github.com/gocd/gocd/files/1120274/assets.txt)


ketan:
I'd logged #2024 earlier to find unused assets from the apache log files, you may want to cross your files with the files from running the command on build.gocd.org.

PR #3639, opened 2017-07-10
Add package-lock.json file

GaneshSPatil:
* Add package-lock.json file for cross-package-manager compatibility
* package-lock.json file is supported from npm v5.0.0
* No need of yarn package manager anymore, as npm(v5.0.0) comes bundled along with node(v8.0)

Related Links:
* [npm 5](http://blog.npmjs.org/post/161081169345/v500)
* [node-v8-release-notes](https://nodejs.org/en/blog/release/v8.0.0)
* [package-lock.json description](https://github.com/npm/npm/pull/16441)

#### Note:
Those who are on an older version of npm (> v5.0.0) can still continue using `package.json` or `yarn.lock`(using yarn) for packages polling, `package-lock.json` file wont affect the existing behavior.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/package-lock.json
+++ b/server/webapp/WEB-INF/rails.new/package-lock.json
@@ -1,0 +1,6695 @@
+{
+  "name": "webpack-rails-example",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "accepts": {
+      "version": "1.3.3",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz",
+      "integrity": "sha1-w8p0NJOGSMPg2cHjKN1otiLChMo=",
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.15",
+        "negotiator": "0.6.1"
+      }
+    },
+    "acorn": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.1.1.tgz",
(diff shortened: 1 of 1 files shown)


ketan:
> No need of yarn package manager anymore, as npm(v5.0.0) comes bundled along with node(v8.0)

I believe we're still on node 6. If you see no issues moving over to node 8 on windows and linux, feel free to update the relevant cookbooks and move over node 8 with npm5. I believe we still need yarn for most of our doc sites, so you may want to consider migrating those too (to avoid having to deal with too many toolchains).

Additionally:

* is there any way to assert that all transitive dependencies in the `package-lock.json` are same as in `yarn.lock`?
* there is an additional `package.json` and `yarn.lock` in the tools dir that you will need to migrate
* you're missing removing the yarn.lock files
* updating of tasks that perform `yarn install`

> Those who are on an older version of npm (> v5.0.0) can still continue using package.json or yarn.lock(using yarn) for packages polling, package-lock.json file wont affect the existing behavior.

This is only used by dev teams, so I think it's perfectly OK to not have to support multiple versions if that's required.

GaneshSPatil:
@ketan -- upgrading to any higher version of node is failing due to some native binding code.
I have tried upgrading it to `v6.11.0`, `v7.0.0` and `v8.0.0`, it fails for the same reason on all the versions.
Need some time to investigate the failure cause.
Closing this PR for now, once we manage to upgrade node to the latest version, I'll reopen it.
Also, 
`v8.1.3` is **`Latest Current Version`** of node (includes npm `v5.0.3 `)
`6.11.0` is **`Latest LTS Version`** (includes npm `v3.10.10`)

PR #3642, opened 2017-07-10
[WIP] Support for implementation if multiple extensions in a plugin

jyotisingh:
Right now plugins users are forced to install multiple plugins to implement things around the same tool, because of the restriction in GoCD which allows for just a single implementation for a given extension in a given plugin jar. For instance, for someone following github PR workflow would need to install both https://github.com/ashwanthkumar/gocd-build-github-pull-requests (implements SCM extension) and https://github.com/gocd-contrib/gocd-build-status-notifier (implements Notification extension) to make things work. This PR would allow merging those two plugins into a single one. 

* This PR introduces support for implementation of different extensions as a part of the same plugin jar
* This PR does NOT introduce support for multiple implementation of the same extension in a single jar.

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/PluginRequestHelper.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/PluginRequestHelper.java
@@ -25,7 +25,7 @@ public class PluginRequestHelper {
             throw new PluginNotFoundException(format("Did not find '%s' plugin with id '%s'. Looks like plugin is missing", extensionName, pluginId));
         }
         try {
-            String resolvedExtensionVersion = pluginManager.resolveExtensionVersion(pluginId, goSupportedVersions);
+            String resolvedExtensionVersion = pluginManager.resolveExtensionVersion(pluginId, goSupportedVersions, extensionName);
             DefaultGoPluginApiRequest apiRequest = new DefaultGoPluginApiRequest(extensionName, resolvedExtensionVersion, requestName);
             apiRequest.setRequestBody(pluginInteractionCallback.requestBody(resolvedExtensionVersion));
             apiRequest.setRequestParams(pluginInteractionCallback.requestParams(resolvedExtensionVersion));
--- a/plugin-infra/go-plugin-access/test/com/thoughtworks/go/plugin/access/authentication/AuthenticationExtensionTest.java
+++ b/plugin-infra/go-plugin-access/test/com/thoughtworks/go/plugin/access/authentication/AuthenticationExtensionTest.java
@@ -69,7 +69,7 @@ public class AuthenticationExtensionTest {
         pluginSettingsConfiguration = new PluginSettingsConfiguration();
         requestArgumentCaptor = ArgumentCaptor.forClass(GoPluginApiRequest.class);
 
-        when(pluginManager.resolveExtensionVersion(PLUGIN_ID, Arrays.asList("1.0"))).thenReturn("1.0");
+        when(pluginManager.resolveExtensionVersion(PLUGIN_ID, Arrays.asList("1.0"), authenticationExtension.extensionName())).thenReturn("1.0");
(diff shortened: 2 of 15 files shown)


ketan:
Open questions:

* How does a hypothetical plugin "foo" implementing Notification and SCM extensions using separate `@Extension` annotations get the plugin settings?
* When a plugin wants to send the messages get/validate plugin settings or get plugin settings view, which extension should it go and ask?


```
@Extension
class FooNotication implements GoPlugin {

}

@Extension
class FooSCM implements GoPlugin {

}
```

arvindsv:
I was wondering if it would be easier to change [GoPluginIdentifier](https://github.com/gocd/gocd/blob/a814c191c94f9c8f1c52eed4ae1c193e2c329a11/plugin-infra/go-plugin-api/src/com/thoughtworks/go/plugin/api/GoPluginIdentifier.java#L27) to declare that it supports multiple extension points and still have only one `@Extension`. The GoPluginApiRequest interface [already has](https://github.com/gocd/gocd/blob/a814c191c94f9c8f1c52eed4ae1c193e2c329a11/plugin-infra/go-plugin-api/src/com/thoughtworks/go/plugin/api/request/GoPluginApiRequest.java#L31) the extension name, which can be used by the plugin to decide how to handle the request.

arvindsv:
That way, the settings can also just be one message for the whole plugin to handle.

jyotisingh:
@ketan - yeah, I am yet to handle plugin settings call, not very sure how to at this point. One option I was thinking about was to get the settings from each of the implementations and display it one after the other in the form that shows up. This would require updating the plugin settings UI & services and accommodating it in the API.

@arvindsv - Changing `GoPluginIdentifier` would have been the easiest thing to do, but that would mean breaking all the existing plugins, unless you can think of a way of doing so without breaking the plugins.

One other option is to have GoPluginV2 where we could define some of these interfaces a little better, but that is a much larger change.

arvindsv:
Yeah, you're right. It's almost the only Java part of the infrastructure left and it's not possible to upgrade easily. I'm wondering whether it's better to do a V2, with better support for this and other things - than do something like this. But, as you said, it's a lot of work (and thought).

If you're going down this path, I think the way to handle settings is to treat this as two different plugins very early on in the process and hide the fact that it is one plugin pretending to be multiple from everything else. So, there should be two PluginLoaded events and maybe change the ID a bit to denote two different extensions within that plugin and handle it transparently. So, that'll mean two different plugin settings.

The problem with that approach is that it doesn't reduce the complexity for a user of this plugin. It'll still have too many settings at too many places.

jyotisingh:
@arvindsv - Yes, if we split up the settings based on number of extensions supported, we need to provide a way for the user to go to each of them - both ux and API wise. And like you mentioned it would be too many places to keep the settings. That's why I was thinking about consolidating them, but that might be mucking around with multiple settings at Go's end. 

If you guys think Goplugin V2 is a better approach then we could start a conversation around how to go about that. I am fine closing this one.

arvindsv:
My preference would be V2. And some thinking about whether this approach will work when we switch out of OSGi.

maheshp:
I would prefer a plugin to do one thing only.

Apart from having to include multiple jars, what else are we trying to solve?

jyotisingh:
Closing this PR as we agreed to have a version 2 of the Plugin implementation to support this and also handle plugin settings properly

arvindsv:
@maheshp said:

> I would prefer a plugin to do one thing only.
> Apart from having to include multiple jars, what else are we trying to solve?

For instance, see this: https://www.gocd.org/2015/07/07/GitHub-Integration/

I would expect that to be one GitHub integration plugin, which takes care of everything. Configuration does not need to be repeated and the experience will be much better.

maheshp:
> I would expect that to be one GitHub integration plugin, which takes care of everything. Configuration does not need to be repeated and the experience will be much better. Instead there are 4 different plugins.

It makes sense if we are proposing a single configuration for integration with Github(or something similar), otherwise from the user perspective you would still need to configure multiple things. If I understand right that would mean a significant change to way plugins work today.

arvindsv:
Yes, single config. Single plugin settings. In my opinion.

PR #3647, opened 2017-07-12
Support withconfigrepo query param for environments INDEX

GaneshSPatil:
* 'api/admin/environments' returns local environment configs
* 'api/admin/environments?withconfigrepo=true' returns merged (local + config_repo) environments configs

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
@@ -22,32 +22,26 @@ module ApiV3
       before_action :check_for_stale_request, only: [:put]
 
       def index
-        is_query_param_provided = params.length > 2
         is_with_remote = params[:withconfigrepo]
         environment_names = environment_config_service.environmentNames()
 
-        if is_with_remote.nil? and !is_query_param_provided
-          load_local_environments(environment_names)
-        elsif is_with_remote and is_with_remote.downcase == 'true'
+        if is_with_remote and is_with_remote.downcase == 'true'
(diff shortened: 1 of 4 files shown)


server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
```diff
@@ -68,7 +80,7 @@ def patch
 
         env_vars = params[:environment_variables] || {}
 
-        env_vars_to_add = (env_vars[:add] || []).map { |env_var|
+        env_vars_to_add = (env_vars[:add] || []).map {|env_var|
```
ketan: use `do ... end` blocks instead of `{}` for multi line blocks.

server/webapp/WEB-INF/rails.new/app/controllers/api_v3/admin/environments_controller.rb
```diff
@@ -22,7 +22,19 @@ class EnvironmentsController < BaseController
...
+        is_with_remote = params[:withconfigrepo]
+        environment_names = environment_config_service.environmentNames()
+
+        if is_with_remote.nil? and !is_query_param_provided
+          load_local_environments(environment_names)
+        elsif is_with_remote and is_with_remote.downcase == 'true'
+          load_merged_environments(environment_names)
+        else
+          return render_not_found_error
+        end
```
ketan: You may be able to simplify a lot of this if you simply ask the environment config service. Additionally, since all the tests are simply mocks, it's easier to read the tests too..

```ruby
environment_config_service.getLocalEnvironments()
environment_config_service.getMergedEnvironments()
```

PR #3651, opened 2017-07-13
Add 'id' as a unique identifier on config-repo tag

GaneshSPatil:
* New config-repo tag example:

```
<config-repo plugin="json.config.plugin" id="9271a1c8-b1ee-4a76-868a-36ffaa349ecc">
	<git url="https://github.com/bob/gocd-json-config-example.git" />
</config-repo>
```

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
@@ -23,6 +23,7 @@ import com.thoughtworks.go.domain.config.Configuration;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
 import com.thoughtworks.go.util.StringUtil;
 
+import java.util.ArrayList;
 import java.util.Map;
 
 /**
@@ -109,6 +110,11 @@ public class ConfigRepoConfig implements Validatable {
         if (repo != null ? !repo.equals(that.repo) : that.repo != null) {
             return false;
         }
+
+        if (id != null ? !id.equals(that.id) : that.id != null) {
+            return false;
+        }
+
         if (configProviderPluginName != null ? !configProviderPluginName.equals(that.configProviderPluginName) : that.configProviderPluginName != null) {
             return false;
         }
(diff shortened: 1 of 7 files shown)


config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
```diff
@@ -71,6 +79,16 @@ public String getConfigProviderPluginName() {
         return configProviderPluginName;
     }
 
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        if(StringUtil.isBlank(id))
+            id = null;
```
varshavaradarajan: this should be randomly generated, no? See, ensureIdExists method in SCM.java

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1502,6 +1502,20 @@ public void shouldMigrateLdapToAuthConfigAsPartOfMigration91() throws Exception
     }
 
     @Test
+    public void shouldAddIdOnConfigRepoAsPartOfMigration94() throws Exception {
+        String configXml = "<cruise schemaVersion='93'>" +
+                "<config-repos>\n" +
+                "   <config-repo plugin=\"json.config.plugin\">\n" +
```
varshavaradarajan: Did you test with multiple config-repos?

varshavaradarajan:
Can you also send a PR to functional tests to update the schema version? We can merge the two together.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
GaneshSPatil: @varshavaradarajan -- do we need to test it with multiple `config-repo` tags?
if yes, how it is different from testing with just one?

config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
ketan: We don't do that with elastic agent and auth plugins. Users are expected to provide an identifier.

GaneshSPatil:
@varshavaradarajan -- We also need to remove the duplicate config-repo which is currently allowed.
We'll do it as part of the same PR

Do not merge until then

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
varshavaradarajan: Its not. I am not saying write a test. Just asking you to confirm that there's no unexpected behavior when you have multiple config repos

config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
varshavaradarajan: Not sure if the xsd validation takes cate of things like id="                ". There should be a validation on the model for the id then.

PR #3670, opened 2017-07-19
Fix config repo flaky tests

GaneshSPatil:


server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/admin/config_repos_controller_spec.rb
```diff
@@ -75,11 +81,6 @@
         login_as_admin
         expect(controller).to allow_action(:get, :show)
       end
-
-      it 'should allow pipeline group admin users, with security enabled' do
```
varshavaradarajan: You should write should disallow group admins. For all actions.

Issue #3671, opened 2017-07-19
Config repo change plugin attribute to pluginId


GaneshSPatil:
##### Current:
```
<config-repo plugin="json.config.plugin" id="repo1">
     <git url="https://github.com/tomzo/gocd-json-config-example.git" />
</config-repo>
```
##### Expected:
```
<config-repo pluginId="json.config.plugin" id="repo1">
     <git url="https://github.com/tomzo/gocd-json-config-example.git" />
</config-repo>
```

##### Current:
```
{
    "id": "repo-1",
    "plugin": "json-config-plugin"
}
```
##### Expected:
```
{
    "id": "repo-1",
    "plugin_id": "json-config-plugin"
}
```

GaneshSPatil:
- [ ] Change "plugin" attribute of "config-repo" element to "pluginId"
- [ ] Change config-repo API representers to show "plugin" as "Plugin_id"
- [ ] config repo API doc PR to show "plugin" as "plugin_id"
- [ ] Change Add config repository doc under [gocd-json-config-plugn](https://github.com/tomzo/gocd-json-config-plugin#add-configuration-repository)
- [ ] Change Add config repository doc under [gocd-yaml-config-plugn](https://github.com/tomzo/gocd-yaml-config-plugin#setup)

GaneshSPatil:
- [ ] Update [config-repo docs](https://docs.gocd.org/17.8.0/advanced_usage/pipelines_as_code.html#storing-pipeline-configuration-in-yaml) on docs.go.cd
Related issue: https://github.com/gocd/docs.go.cd/issues/110

PR #3672, opened 2017-07-19
Config repo change plugin attribute to plugin_id

GaneshSPatil:


GaneshSPatil:
issue: https://github.com/gocd/gocd/issues/3671

PR #3678, opened 2017-07-20
Remove warnings

cv:
This is a simple PR to remove warnings generated by IntelliJ IDEA inspections. I took care to make each automated warning removal as a separate commit, so it'd be slightly easier to review (the combined diff might be a little overwhelming, mind).

No behavior is expected to change as a result of this.


ketan:
Some of the usages of lamdas and method references may fail because of the old version of spring we use.

I've kicked off a build so you should be able to see some failures if there are any. Please login using the guest login icon on the login page or use `view:password` as the login credentials.

cv:
Thanks, @ketan! I'll go through the failing tests - looks like there were some related to lambdas, indeed - and update the PR accordingly.

ketan:
@cv — is there still interest in cleaning this?

cv:
I think I'd have to start over - probably a lot easier than solving the merge conflicts manually. Closing this and booking some time later this month for it. Wish me luck! :)

PR #3686, opened 2017-07-24
Spring bean reinit bug

jyotisingh:
Fixing the bug wherein DbDeploy migration would run multiple times
* This used to happen because HttpService had a autowired contructor with required false and a default constructor. As a part of the first initialization on the server side if a db-migration failed(for that matter anything threw an exception), spring would destroy all the beans that it created until it hits a class with more than one candidate-constructors. It would now use the second constructor, and retry the flow. From a class instantiation point of view this seems fine, but since db-migration is linked to the initilization it wouldn't work for us
* Added a test to ensure that we do not add something like this by mistake in future.

### DIFF
--- a/common/src/com/thoughtworks/go/util/HttpService.java
+++ b/common/src/com/thoughtworks/go/util/HttpService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 ThoughtWorks, Inc.
+ * Copyright 2017 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package com.thoughtworks.go.util;
 
 import com.thoughtworks.go.agent.common.ssl.GoAgentServerHttpClient;
-import com.thoughtworks.go.agent.common.ssl.GoAgentServerHttpClientBuilder;
 import com.thoughtworks.go.domain.FetchHandler;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.logging.Log;
@@ -33,24 +32,17 @@ import org.apache.http.entity.mime.content.ByteArrayBody;
 import org.apache.http.entity.mime.content.FileBody;
 import org.apache.http.message.BasicNameValuePair;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
 
 import javax.servlet.http.HttpServletResponse;
 import java.io.*;
 import java.util.Arrays;
 import java.util.Properties;
 
-@Component
 public class HttpService {
     private HttpClientFactory httpClientFactory;
     private static final Log LOGGER = LogFactory.getLog(HttpService.class);
-    public static final String GO_ARTIFACT_PAYLOAD_SIZE = "X-GO-ARTIFACT-SIZE";
 
-    public HttpService() {
-        this(new GoAgentServerHttpClient(new GoAgentServerHttpClientBuilder(new SystemEnvironment())));
-    }
-
-    @Autowired(required = false)
+    @Autowired
     public HttpService(GoAgentServerHttpClient httpClient) {
         this(new HttpClientFactory(httpClient));
     }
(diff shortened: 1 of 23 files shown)


server/test/integration/com/thoughtworks/go/server/SpringBeanCandidateConstructorIntegrationTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.hamcrest.core.IsNull.notNullValue;
+import static org.junit.Assert.fail;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = {
+        "classpath:WEB-INF/applicationContext-global.xml",
+        "classpath:WEB-INF/applicationContext-dataLocalAccess.xml",
+        "classpath:WEB-INF/applicationContext-acegi-security.xml"
+})
+public class SpringBeanCandidateConstructorIntegrationTest {
```
ketan: Do we need something similar on the agent?

jyotisingh: Yes, I thought about it, but got a little lazy. Even though a reinitialization at the agent end wouldn't hurt, but it does make sense. Will do

agent/src/com/thoughtworks/go/agent/AgentHTTPClientController.java
```diff
@@ -34,6 +34,7 @@
 import com.thoughtworks.go.util.SystemEnvironment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
```
maheshp: unused import

server/test/integration/com/thoughtworks/go/server/SpringBeanCandidateConstructorIntegrationTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Autowired
+    private ApplicationContext applicationContext;
+
+    @Test
+    public void shouldEnsureThatNoBeanHasMoreThanOneCandidateConstructors() throws IllegalAccessException, InvocationTargetException, InstantiationException, NoSuchMethodException {
```
maheshp: Put a note around issue with multiple candidate constructor?

common/src/com/thoughtworks/go/util/HttpService.java
```diff
@@ -40,15 +39,9 @@
...
 public class HttpService {
     private HttpClientFactory httpClientFactory;
     private static final Log LOGGER = LogFactory.getLog(HttpService.class);
-    public static final String GO_ARTIFACT_PAYLOAD_SIZE = "X-GO-ARTIFACT-SIZE";
-
-    public HttpService() {
-        this(new GoAgentServerHttpClient(new GoAgentServerHttpClientBuilder(new SystemEnvironment())));
-    }
 
     @Autowired(required = false)
```
maheshp: remove `required = false`

server/src/com/thoughtworks/studios/shine/cruise/stage/StagesQuery.java
```diff
@@ -25,6 +25,8 @@
 import com.thoughtworks.studios.shine.semweb.BoundVariables;
 import com.thoughtworks.studios.shine.semweb.Graph;
 import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
```
maheshp: Unused imports

server/test/integration/com/thoughtworks/go/server/SpringBeanCandidateConstructorIntegrationTest.java
jyotisingh: `shouldEnsureThatNoBeanHasMoreThanOneCandidateConstructorsToPreventSpringFromBacktrackingAndUsingOtherCandidateConstructorsInCaseOfAnExceptionBeingThrownInAConstructorOrPostConstructMethod` ?

Issue #3619, opened 2017-06-28
API to delete all disabled users


varshavaradarajan:
##### Issue Type

- Feature enhancement

##### Summary

Currently we have to make individual API calls to delete disabled users. Does it make sense to provide an option to delete all disabled users?

varshavaradarajan:
https://github.com/gocd/gocd/pull/3720

PR #3694, opened 2017-07-25
API to delete all disabled users.

varshavaradarajan:
#3619

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/UserService.java
+++ b/server/src/com/thoughtworks/go/server/service/UserService.java
@@ -258,6 +258,33 @@ public class UserService {
         }
     }
 
+    private List<String> getAllDisabledUsers() {
+        ArrayList<String> disabledUsers = new ArrayList<>();
+        Collection<User> allUsers = allUsers();
+        for (User user : allUsers) {
+            if (!user.isEnabled()) {
+                disabledUsers.add(user.getName());
+            }
+        }
+        return disabledUsers;
+    }
+
+    public void deleteAllDisabledUsers(HttpLocalizedOperationResult result) {
+        List<String> allDisabledUsers = getAllDisabledUsers();
+        if (allDisabledUsers.isEmpty()) {
+            result.unprocessableEntity(LocalizedMessage.string("NO_DISABLED_USERS"));
+            return;
+        }
+        for (String allDisabledUser : allDisabledUsers) {
+            deleteUser(allDisabledUser, result);
+            if (!result.isSuccessful()) {
+                return;
+            }
+        }
+        result.setMessage(LocalizedMessage.string("RESOURCES_DELETE_SUCCESSFUL", "user(s)", allDisabledUsers));
+    }
+
+
     public enum SortableColumn {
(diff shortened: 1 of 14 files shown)


ketan:
I assumed that deleting users is not permissible, because we have references to users in other tables.

@arvindsv @jyotisingh — care to throw some light?

ketan:
Arguably, there is [this API](https://api.gocd.org/current/#delete-a-user) to delete a user, written by yours truly...

arvindsv:
@ketan - As far as I know, there is no foreign key from any table to the users table. So, delete should be fine. @jyotisingh might know better.

The only problem we had with the delete is that, with the default value of the "Allow users that exist in LDAP, password file or via plugin to log into GoCD, even if they haven't been explicitly added to GoCD" flag being "true", a deleted user might be able to log back in, whereas a disabled user would not have been allowed to.

arvindsv:
Having said that ... the "delete all disabled users" seems like too specific an API to exist. I'd expect people to write a script which goes through all the disabled users, by using the API and submitting "delete user" calls instead. We could look at a "bulk delete users" call, if needed - but this one seems too specific.

I try to see whether API calls are composable. This one doesn't seem to be.

varshavaradarajan:
Bulk delete makes sense. But why wouldn't you want to give an option to delete all disabled users too?

arvindsv:
I'll invoke the "lowly-product-manager" defence here and will let all of you decide. This one is more technical than anything. As I said, to me, it feels like a weirdly specific action to have as an API. But, if you can convince the others, good for you. :)

varshavaradarajan:
Can you tell me what you mean by "weirdly" specific? Why can't an API be specific?

ketan:
I agree with what @arvindsv has to say. A bulk delete API makes sense if you're asking it to delete objects with specific identifiers. Like [this](https://api.gocd.org/current/#bulk-update-agents) or [this](http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html).

What makes me queasy about this api is that it applies an operation (delete) to objects that match a filter (disabled).

varshavaradarajan:
> What makes me queasy about this api is that it applies an operation (delete) to objects that match a filter (disabled).

Why? I think we should have a bulk delete too. But why not this too?

ketan:
>> What makes me queasy about this api is that it applies an operation (delete) to objects that match a filter (disabled).

> Why? So what if its a filter? Why not perform an operation on things that match a filter? I think we should have a bulk delete too. But why not this too?

The idea in such a case would be to introduce a resource called a `filter`, and perhaps call a delete with that filter id:

```http
POST /api/users/filters?where='state=disabled'
```

Response:
```
Location: /api/users/filter/1234
```

Then you perform a delete:

```
DELETE /api/users?filterId=1234
```

Alternatively, do a delete by query API - https://www.elastic.co/guide/en/elasticsearch/reference/1.4/docs-delete-by-query.html

varshavaradarajan:
Closing this. Will open another PR for bulk delete of disabled users.

Issue #3650, opened 2017-07-13
Websocket console log says "Waiting for console logs" when console log is removed


ketan:
##### Issue Type

- Bug Report

##### Summary

- run a build
- navigate to the job detail page
- remove console.log from the artifacts dir
- refresh page
- see the message "Waiting for console logs..."

rajiesh:
Verified on 17.12.0 (5605-89a5ade1e9497ffc255dfa744476466c09beca06)

PR #3695, opened 2017-07-25
Should close websocket connection when console.log file not found (#3650)

GaneshSPatil:
Note: need to handle it at client side

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_socket.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_socket.js
@@ -64,6 +64,13 @@
     function maybeResume(e) {
       if (fatal) return;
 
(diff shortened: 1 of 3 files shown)


server/webapp/WEB-INF/rails.new/app/assets/javascripts/console_log_socket.js
```diff
@@ -64,6 +64,13 @@
     function maybeResume(e) {
       if (fatal) return;
 
+      if (e.type === "close" && e.code === 1011) {
+        startLine = 0;
+
+        console.error('Something went wrong:'+ e.reason);
+        return;
+      }
```
GaneshSPatil: @ketan -- Merging this for now.
This can be fixed as a part of **Custom WebSocket Wrapper** that we have written which falls back to polling and has a better error handling!

PR #3709, opened 2017-07-28
Simplify how agent.jar output is copied to file.

ketan:
The current implementation does the equivalent of:

`java -jar agent.jar > go-agent-stdout.log 2>go-agent-stderr.log`

It does this by messing around with log4j NDC and filters which are
configured programatically, and any new appenders initialized later by
changing log4j.properties at runtime are not configured.

The new implementation, simply gets around this problem by directly
writing to a `RollingFileAppender` without going through the rest of the
logging framework.

Logs are optionally written to a `ConsoleAppender` if a system property
is set.


ketan:
@varshavaradarajan — [this code](https://github.com/gocd/gocd/blob/ecc7290c4a53ac82b4fb23348111944b3bb85d8a/installers/go-agent/release/agent.sh#L114-L122) is now redundant because of [this line](https://github.com/gocd/gocd/pull/3709/files#diff-39e686fcc3bd93e97576561a2a949bbdL65).

I'm going to submit another commit to cleanup the script.

agent-process-launcher/src/com/thoughtworks/go/agent/AgentConsoleLogThread.java
```diff
@@ -16,67 +16,62 @@
...
             if (!keepRunning && flushedAfterStop) {
                 break;
             }
             try {
                 logConsoleError(reader);
-                Thread.sleep(1000);
+                Thread.sleep(100);
             } catch (Exception e) {
-                LOG.error("Error occured while capturing console output from agent process: ", e);
+                System.err.println("Error occured while capturing console output from agent process: ");
```
jyotisingh: We seem to catch and ignore all exceptions inside `logConsoleError` so may be we dont even need this catch block. But wondering if its really ok to ignore all exceptions [here](https://github.com/gocd/gocd/pull/3709/files#diff-60758573102b6e2af9173fe53220083cR67).  Why not leave it to the original throws IOException in `logConsoleError` and handle the exception in here ie. LOG.error and System.err.println to take care of both docker based installations and regular rpm based ones.

agent-process-launcher/src/com/thoughtworks/go/agent/AgentConsoleLogThread.java
```diff
@@ -16,67 +16,62 @@
...
+
+    private void pipeLogs(BufferedReader reader) {
         while (true) {
             if (!keepRunning && flushedAfterStop) {
                 break;
             }
             try {
                 logConsoleError(reader);
-                Thread.sleep(1000);
+                Thread.sleep(100);
```
jyotisingh: Do you think sleeping for a second is too slow?

PR #3721, opened 2017-08-01
Clear out all OAuth token on auth config change

bdpiparva:
`OauthTokenSweeper` is a config change listner. Which clears out OAuth
tokens on security config change. This class is using SecurityConfig#hasSecurityMethodChanged
to determine change in security config

Updated `SecurityConfig#hasSecurityMethodChanged` to check config change
in auth config along with inbuilt password file and ldap

### DIFF
--- a/config/config-api/src/com/thoughtworks/go/config/SecurityConfig.java
+++ b/config/config-api/src/com/thoughtworks/go/config/SecurityConfig.java
@@ -126,9 +126,12 @@ public class SecurityConfig implements Validatable {
         if (newSecurity == null) {
             return true;
         }
+
         boolean ldapChanged = !ObjectUtil.equal(ldapConfig, newSecurity.ldapConfig());
         boolean passwordFileChanged = !ObjectUtil.equal(passwordFileConfig, newSecurity.passwordFileConfig());
-        return ldapChanged || passwordFileChanged;
+        final boolean securityAuthConfigChanged = !ObjectUtil.equal(securityAuthConfigs, newSecurity.securityAuthConfigs());
+
+        return ldapChanged || passwordFileChanged || securityAuthConfigChanged;
     }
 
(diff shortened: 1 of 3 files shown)


config/config-api/src/com/thoughtworks/go/config/SecurityConfig.java
```diff
@@ -126,9 +126,12 @@ public boolean hasSecurityMethodChanged(SecurityConfig newSecurity) {
         if (newSecurity == null) {
             return true;
         }
+
         boolean ldapChanged = !ObjectUtil.equal(ldapConfig, newSecurity.ldapConfig());
         boolean passwordFileChanged = !ObjectUtil.equal(passwordFileConfig, newSecurity.passwordFileConfig());
-        return ldapChanged || passwordFileChanged;
+        final boolean securityAuthConfigChanged = !ObjectUtil.equal(securityAuthConfigs, newSecurity.securityAuthConfigs());
+
+        return ldapChanged || passwordFileChanged || securityAuthConfigChanged;
```
maheshp: I think any changes to security config should not affect oAuth, so wondering if we need to clean the tokens at if there is a change in security. But, we need to give a thought bout the expected behaviour when security is enabled and disabled.

jyotisingh:
@bdpiparva - closing this one. We agreed on not clearing off the tokens when the config changes. With the support for gadgets gone, we shouldn't even need this.

ketan:
> @bdpiparva - closing this one. We agreed on not clearing off the tokens when the config changes. With the support for gadgets gone, we shouldn't even need this.

I thought that the understanding was that we remove the OAuthTokenSweeper altogether?

jyotisingh:
@ketan  - yes, thats what I meant by not clearing off the tokens. That will be a separate PR I guess.

PR #3722, opened 2017-08-01
Add a toggle to make pipeline quick edit page as a default page

varshavaradarajan:


server/resources/available.toggles
```diff
@@ -7,9 +7,9 @@
          "value": false
       },
       {
-         "key": "pipeline_config_single_page_app_key",
```
maheshp: Should we be removing this toggle? From what I remember we had this toggle since this feature is not complete and this gives the ability for admins to turn this off.

arvindsv: I'm not sure anyone used it. If we're going to make it the default and we're still not sure, we should keep it. If not, it can go.

PR #3723, opened 2017-08-01
Modify build scripts to work in a windows environment

thecontrarycat:
These are the changes I had to make to the build so that I could compile from the command line on Windows.

build.gradle => change git:// url to https:// 
inline-partials.rake => change Windows-style paths to Unix-style
rails.gradle => ensure yarn has been ran without arguments first (to restore the packages) before we try to run webpack

### DIFF
--- a/server/rails.gradle
+++ b/server/rails.gradle
@@ -16,7 +16,6 @@
 
 
 import com.thoughtworks.go.build.ExecuteUnderRailsTask
-import com.thoughtworks.go.build.NpmInstallTask
 import com.thoughtworks.go.build.PrepareRailsCommandHelper
 
 task pathingJar(type: Jar) {
@@ -54,8 +53,16 @@ private boolean isWindows() {
   org.gradle.internal.os.OperatingSystem.current().isWindows()
 }
 
-task npmInstall(type: NpmInstallTask) {
+task npmInstall(type: Exec) {
+  standardOutput = System.out
+  errorOutput = System.err
+
+  commandLine = [isWindows() ? "yarn.cmd" : "yarn", "install"]
   workingDir = project.railsRoot
+
+  doFirst {
+    println "[${workingDir}]\$ ${executable} ${args.join(' ')}"
+  }
 }
 
 prepare.dependsOn
--- a/build.gradle
+++ b/build.gradle
@@ -137,7 +137,7 @@ allprojects {
   rootProject.ext.gitRevision = GIT_REVISION
   rootProject.ext.copyrightYear = COPYRIGHT_YEAR
 
-  rootProject.ext.commandRepoUrl = System.getenv('COMMAND_REPO_URL') ? System.getenv('COMMAND_REPO_URL') : "git://github.com/gocd/go-command-repo"
+  rootProject.ext.commandRepoUrl = System.getenv('COMMAND_REPO_URL') ? System.getenv('COMMAND_REPO_URL') : "https://github.com/gocd/go-command-repo"
 
   group = 'com.thoughtworks.go'
   version = project.fullVersion
--- a/server/inline-partials.rake
+++ b/server/inline-partials.rake
@@ -14,7 +14,7 @@
 # limitations under the License.
 ##########################################################################
 
-RAILS_INTERPOLATED_VIEWS = (ENV['INPUT_DIR'] or raise 'INPUT_DIR not defined')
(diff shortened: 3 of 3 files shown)


thecontrarycat:
Note: *I don't have a Linux/Mac environment to test against, so I'm not 100% sure that these changes won't break the build on those platforms*

server/rails.gradle
```diff
@@ -135,6 +135,16 @@ task compileAssets(type: ExecuteUnderRailsTask) {
   }
 
   doLast {
+	project.exec {
```
ketan: I believe this should not be needed since there's already a dependency on `npmInstall` around line 113.

`npmInstall` should ideally be called `yarnInstall` since we've moved over to yarn, but that's a separate issue.

If you can remove this block of code, and squash the PR. I'm happy to merge it.

thecontrarycat: You'd think, but it doesn't run on my machine.

```11:07:21.381 [DEBUG] [org.gradle.internal.progress.DefaultBuildOperationExecutor] Completing Build operation 'Snapshot task inputs for :server:npmInstall'
11:07:21.381 [DEBUG] [org.gradle.internal.progress.DefaultBuildOperationExecutor] Build operation 'Snapshot task inputs for :server:npmInstall' completed
11:07:21.381 [INFO] [org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter] Caching disabled for task ':server:npmInstall': Caching has not been enabled for the task
11:07:21.381 [DEBUG] [org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter] Determining if task ':server:npmInstall' is up-to-date
11:07:21.390 [INFO] [org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter] Skipping task ':server:npmInstall' as it is up-to-date (took 0.009 secs).
11:07:21.390 [DEBUG] [org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter] Removed task artifact state for {} from context.
11:07:21.390 [DEBUG] [org.gradle.api.internal.tasks.execution.ExecuteAtMostOnceTaskExecuter] Finished executing task ':server:npmInstall'
11:07:21.390 [LIFECYCLE] [class org.gradle.internal.buildevents.TaskExecutionLogger] :server:npmInstall UP-TO-DATE
11:07:21.390 [DEBUG] [org.gradle.internal.progress.DefaultBuildOperationExecutor] Completing Build operation 'Task :server:npmInstall'
11:07:21.390 [DEBUG] [org.gradle.internal.progress.DefaultBuildOperationExecutor] Build operation 'Task :server:npmInstall' completed
11:07:21.390 [INFO] [org.gradle.execution.taskgraph.DefaultTaskPlanExecutor] :server:npmInstall (Thread[Task worker Thread 3,5,main]) completed. Took 1.37 secs.```

**Comment by thecontrarycat** [comment]
I've updated `rails.gradle` so the `npmInstall` task now directly calls `yarn.cmd install` rather than relying upon `NpmInstallTask.groovy` to do it.  This fixes the issue on Windows of the packages not being restored during build (and thus breaking `:server:compileAssets`)

**Comment by ketan** [comment]
I'm not sure I understand why `NpmInstallTask` does not work on windows. I've been using that task on windows without any hiccups...

**Comment by ketan** [comment]
I think I may know why you're having to do this. I believe you may be running `./gradlew clean prepare` which randomly causes the `npmInstall` task to think it's UP-TO-DATE. This is a known issue with gradle (https://github.com/gradle/gradle/issues/1168). I suggest that you do run `./gradlew clean` by itself, with no other tasks on the command line.

**Comment by thecontrarycat** [comment]
I'm actually running `gradlew clean genericAgentZip genericServerZip` (as the developer docs suggest).

Sounds like you're proposing that I run 
```
gradlew clean
gradlew genericAgentZip genericServerZip
```
which seems a bit of an annoyance given we can work around this bug by rewriting the task.

ketan:
> which seems a bit of an annoyance given we can work around this bug by rewriting the task.

Either way, it breaks because of a known gradle issue. I'm going to apply all changes except for the `npmInstall` task. I'm watching for a gradle 4.2 to fix the `npmInstall` issue

ketan:
> build.gradle => change git:// url to https://

Could you tell me why this change was needed?

PR #3731, opened 2017-08-03
Eager load a quick request handler

arvindsv:
The Jetty Server is started empty, with no handlers and that's quick. It takes ports 8153 and 8154 and starts returning 404 for everything. Immediately, the welcome file handler is started. It starts handling /go (redirecting to /go/home) which won't work, because the Spring web app (and Rails) haven't started yet. Once they're up, the redirection works.

Till that time, we can show a message or something more clever, while waiting for the rest of the app to come up.

This is based on how Jetty's hot reloads work. The idea (for this PR) is that there should be no change in behavior due to this. Any scripts which were checking whether 8153/8154 are responding to decide whether the server is up, will need to change.

Need to check:

* Will the server process stop properly if something goes wrong?
    - If config migration fails
    - If some other validation fails

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ development-utility/development-agent/target/
 development-utility/development-server/target/
 installers/target/
 jetty9/target/
+jetty9/logs/
 plugin-infra/go-plugin-access/target/
 plugin-infra/go-plugin-activator/target/
 plugin-infra/go-plugin-api-internal/target/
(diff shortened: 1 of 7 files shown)


arvindsv:
Can be merged. [Everything](https://build.gocd.org/go/pipelines/value_stream_map/trigger/1443) succeeded.

arvindsv:
Squashed changes.

ketan:
@arvindsv — mind rebasing and kicking off a build?

arvindsv:
Done. [All green](https://build.gocd.org/go/materials/value_stream_map/caef7f8c7e79ac4acdc4e41c1a0aafc7fbd2ef0ff8cf0a18371457b7c5aa0757/eed66f0f7205679ac87f679ade4ba8c36c466f18).

arvindsv:
I'll verify today that agents handle this properly during an upgrade. That's the other aspect I was thinking of - but forgot to do. I believe they'll get a 404 and retry.

Issue #3401, opened 2017-04-17
Share go server info (UUID, Site URL, Secure Site URL etc..)with plugin


bdpiparva:
##### Issue Type

- Feature proposal

##### Summary

Share go server identity with plugins to tag plugin generated resources for go server.

#### Why we need it ?

Consider that two go server(for two different team) running and ECS plugin installed on both the server and configured using AWS credentials.

There is no way to identify resources are created using which plugin.

#### Why plugin is not taggin resource with self generated UUID ?

* Plugin will lose it state on go server restart load or unload of plugin

Qinmu-Xie:
#3341

ketan:
Added via #3732.

PR #3732, opened 2017-08-04
Add a request processor to allow plugins to query the server info (#3401)

ketan:


ketan:
With this in place, a request for `go.processor.server-id.get` will return the following JSON:


```json
{
  "server_id": "df0cb9be-2696-4689-8d46-1ef3c4e4447c",
  "site_url": "http://example.com:8153/go",
  "secure_site_url": "https://example.com:8154/go"
}
```

PR #3734, opened 2017-08-04
Admins can edit pipelines from VSM now

maheshp:
* Adding the pipeline gear icon on VSM to ease pipeline edit for Admins

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/GoConfigService.java
+++ b/server/src/com/thoughtworks/go/server/service/GoConfigService.java
@@ -148,7 +148,7 @@ public class GoConfigService implements Initializer, CruiseConfigProvider {
     }
 
     public boolean canEditPipeline(String pipelineName, Username username) {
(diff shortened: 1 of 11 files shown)


server/src/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -147,6 +147,15 @@ private boolean canEditPipeline(String pipelineName, Username username, Localize
         return canEditPipeline(pipelineName, username, result, findGroupNameByPipeline(new CaseInsensitiveString(pipelineName)));
     }
 
+    public boolean canEditPipeline(String pipelineName, Username username) {
+        if (getCurrentConfig().hasPipelineNamed(new CaseInsensitiveString(pipelineName)) &&
```
jyotisingh: `getCurrentConfig().hasPipelineNamed` internally uses `getPipelineConfigByName` which goes over the list of pipeline-groups and hence could become too slow. Could you change the implementation of `hasPipelineNamed` to use `pipelineConfigByName`. `pipelineConfigByName` uses a cached map to get this information.

PR #3746, opened 2017-08-09
Disallow setting of job status/result from an unassigned agent

jyotisingh:


Webysther:
@jyotisingh you can add in project? I can help close some issues.

jyotisingh:
@Webysther - Did you mean you could update the status of some issues? If so, just comment on the issue with the details and one of us could update the status of the issue accordingly.

If you were talking about closing the issue by sending code fixes for them, then it would be great. We follow a PR based workflow. So, just send across any PR. We would be happy to review/merge it. Thanks for showing interest.

jyotisingh:
[All green](https://build.gocd.org/go/pipelines/value_stream_map/trigger/1468)

PR #3752, opened 2017-08-10
Disabled agents are refreshed now

### LINKED ISSUE #3747: Jobs hung forever
##### Issue Type

- Bug Report

##### Summary

Steps to reproduce
1. Disable an agent which is running a job
2. Delete the agent

The Job remains in the state that was communicated to the server before the agent dies and is hung forever. 

The job is not rescheduled since disabled agents are not refreshed [refer](https://github.com/gocd/gocd/blob/master/common/src/com/thoughtworks/go/domain/AgentInstance.java#L171)

maheshp:
* This fixes #3747

### DIFF
--- a/common/src/com/thoughtworks/go/domain/AgentInstance.java
+++ b/common/src/com/thoughtworks/go/domain/AgentInstance.java
@@ -168,7 +168,7 @@ public class AgentInstance implements Comparable<AgentInstance> {
     }
 
     public void refresh(final AgentRuntimeStatus.ChangeListener changeListener) {
-        if (agentConfigStatus == AgentConfigStatus.Pending || agentConfigStatus == AgentConfigStatus.Disabled) {
+        if (agentConfigStatus == AgentConfigStatus.Pending) {
             return;
         }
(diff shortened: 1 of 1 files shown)


common/src/com/thoughtworks/go/domain/AgentInstance.java
```diff
@@ -168,7 +168,7 @@ public boolean canApprove() {
     }
 
     public void refresh(final AgentRuntimeStatus.ChangeListener changeListener) {
-        if (agentConfigStatus == AgentConfigStatus.Pending || agentConfigStatus == AgentConfigStatus.Disabled) {
```
GaneshSPatil: was there any specific reason why disabled agents were not refreshing the runtime status?

maheshp: Not sure, couldn't find a test for it as well

maheshp:
There are few more changes required to fix the hung jobs issue, closing this and sending a new PR with the required changes.

PR #3753, opened 2017-08-10
Stage and job related warnings should not be logged for a rerun stage/job

jyotisingh:
A rerun stage would already contain state from the previous run. Currently every rerun of a job/stage adds a lot of unnecessary noise to the server logs


GaneshSPatil:
@jyotisingh -- These warnings logs have been added for debugging purpose of the improper `stage`/`job` state.
If we don't see these warnings on `build.gocd.org` going forward in couple of weeks/months, we should get rid of them.

PR #3761, opened 2017-08-16
Removing ldap support from core

bdpiparva:
Depends on: [PR#3850](https://github.com/gocd/gocd/pull/3850)

* Use bundled [Ldap Plugin](https://github.com/gocd/gocd-ldap-authentication-plugin). Existing Ldap configurations have been moved to `Authorization Configuration`  in **_Admin_** > **_Security_**


maheshp:
Probably we should hold on to merging this till `17.10.0` since we need to support this till September as mentioned in the `17.5.0` release notes.

bdpiparva:
@ketan ---  I will rebase this on top of https://github.com/gocd/gocd/pull/3850

Issue #3756, opened 2017-08-11
Incorrect sort order of Free space values in agents tab


ashokgowtham:
##### Issue Type

Bug Report,  Feature enhancement

##### Summary

Free space values in agents tab are not sorted correctly (does not consider MB/GB, just sorts based on number) which causes wrong ordering.

###### Basic environment details

* Go Version: 17.5.0 (5095-3130b5ba96f8155ae86fefb261a559d4c7493326)
* JAVA Version: 1.8.0_121
* OS: Centos, Linux 3.10.0-327.el7.x86_64
* Browser vendor and version (if relevant): Chrome 59

##### Expected Results
Sorting should be done numerically considering the units (4MB is < 1GB)

##### Actual Results
An agent with 48.36 MB is listed at the last after all other agents (having free space in GBs), when sorted in ascending order.

PR #3762, opened 2017-08-16
Fixes sort bug present for freeSpace column in agents table

yankurk:


yankurk:
#3756

ketan:
Do you think a test is in order? https://github.com/gocd/gocd/blob/3fef18c6c79dafa81e9833ebe2044bcbd02ec504/server/webapp/WEB-INF/rails.new/spec/webpack/models/agents/agents_spec.js#L154

ketan:
If a new test isn't needed, perhaps modify an existing test to cover this behavior?

ashokgowtham:
Seems like the build has passed previously
https://build.gocd.org/go/pipelines/trigger/1486/do-nothing/1
But was cancelled after a retrigger. So actually all checks passed

PR #3766, opened 2017-08-18
Elastic agent capabilities

maheshp:
***WIP Do Not Merge**


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentExtension.java
```diff
@@ -132,12 +133,26 @@ public String getStatusReport(String pluginId) {
             throw new UnsupportedOperationException(format("Plugin `%s` implements Elastic Agent V1, `StatusReport` endpoint is not supported in this version.", pluginId));
```
maheshp: I think we would need this since V1 does not support this endpoint, the expectation from plugins implementing V2 is to support `get-capabilities` request

bdpiparva: Ok, I think we should also add check for plugin capabilities after version check

PR #3767, opened 2017-08-18
Elastic agent id on job details page

bdpiparva:
* This will help user to identify elastic agent on status report page

screen shot 2017-08-18 at 12 37 15 pm

### DIFF
--- a/server/src/com/thoughtworks/go/server/controller/JobController.java
+++ b/server/src/com/thoughtworks/go/server/controller/JobController.java
@@ -16,7 +16,10 @@
 
 package com.thoughtworks.go.server.controller;
 
-import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.config.AgentConfig;
+import com.thoughtworks.go.config.CaseInsensitiveString;
+import com.thoughtworks.go.config.Tabs;
+import com.thoughtworks.go.config.TrackingTool;
 import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.domain.Properties;
 import com.thoughtworks.go.i18n.Localizer;
@@ -45,6 +48,7 @@ import java.util.*;
 import static com.thoughtworks.go.server.controller.actions.JsonAction.jsonFound;
 import static com.thoughtworks.go.util.ExceptionUtils.bomb;
 import static com.thoughtworks.go.util.GoConstants.ERROR_FOR_PAGE;
+import static com.thoughtworks.go.util.StringUtil.isBlank;
 import static com.thoughtworks.go.util.json.JsonHelper.addDeveloperErrorMessage;
 
 /*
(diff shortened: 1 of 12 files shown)


server/src/com/thoughtworks/go/server/controller/JobController.java
```diff
@@ -192,6 +196,15 @@ public ModelAndView handleRequest(@RequestParam(value = "pipelineName") String p
         return jsonFound(json).respond(response);
     }
 
+    private AgentConfig agentConfig(String agentUuid) {
+        if (isBlank(agentUuid)) {
+            return null;
+        }
+
+        final AgentInstance agentInstance = agentService.findAgent(agentUuid);
+        return agentInstance == null ? null : agentInstance.agentConfig();
```
jyotisingh: That null check might not work as `agentService.findAgent(agentUuid)` returns a `NullAgentInstance` object if it cannot find a matching agent, see [this]( https://github.com/bdpiparva/gocd/blob/f31eeae5f61ce974be4d679f1b7d6b1f4c5ed63c/common/src/com/thoughtworks/go/server/domain/AgentInstances.java#L90). Might want to change that to `agentInstance.isNullAgent()`

bdpiparva: I think null check is fine because `NullAgentInstance` also has the `NullAgent`. WDYT ?

jyotisingh: What I meant is `agentInstance == null` will always be false since `agentService.findAgent(agentUuid)` will never return null.

jyotisingh:
There are two places where this info would be useful. While looking at the 'Status report' page if the plugin supports it, and if you list all your agents using `docker ps` or equivalent command. 

@bdpiparva - While `Agent` and `Elastic Agent Id` are both available on the page, `Elastic Agent Id` seems like something that you could easily miss spotting with it being at the bottom of the section. You and I are aware of it and so would find it handy, but how do we make it known to users that this is when the information would be useful. Also, I don't know if that's the right label meaning if the users would understand it. 
Just check with @arvindsv and @naveenbhaskar on that.

arvindsv:
> @bdpiparva - While Agent and Elastic Agent Id are both available on the page, Elastic Agent Id seems like something that you could easily miss spotting with it being at the bottom of the section. You and I are aware of it and so would find it handy, but how do we make it known to users that this is when the information would be useful. Also, I don't know if that's the right label meaning if the users would understand it.

Yes, I was thinking about that too. Either it needs more context, or it needs to link to something which provides that context.

arvindsv:
A quick summary of a conversation related to this: The elastic agent ID seems out of context, and it's quite possible that this information is relevant only during a run. For an old run, this might not be important, because the agent will probably be gone.

One approach I was thinking of was to show the agent UUID in the plugin status page, but @bdpiparva and @ketan reminded me that the plugin doesn't know the agent UUID. It knows only about the plugin ID.

So, an approach could be to ask the plugin to provide a little template to show in a popup, when the agent link in clicked in the job details and agents pages. This does complicate the interface a bit, but it allows the user to not have to know things like the elastic agent ID and try and correlate that information on the plugin status page, where the ID might no longer be present.

maheshp:
Having a template makes sense to me, probably show basic information about the container instance and the container would help.

Also, should we consider adding a icon on the job details page to identify elastic agents?

arvindsv:
I wonder if it will be a template or a real-time call to the plugin, giving it an elastic agent ID. Template will be useful when the server has the data and the plugin just needs to tell it how to display it. In this case, it feels like the plugin has the data too.

Sure, adding an icon makes sense.

maheshp:
My bad, when you mentioned template I din't read it right.

I was thinking of adding a new elastic agent plugin endpoint, for a given agent id or elastic agent id it should return a html snippet with the relevant details for the agent.

arvindsv:
Yes, that.

ketan:
As mentioned in this comment (https://github.com/gocd/gocd/pull/3767#issuecomment-324002903). Closing in favor of the plugin providing some contextual popup, or something better.

Issue #3588, opened 2017-06-15
Agent logs location


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

We made some changes to the way logging was done on the agent in 17.4.0. This affects the agents installed via rpm or deb. The logs will be present in multiple locations - /var/log/go-agent, /var/lib/go-agent/logs

###### Basic environment details

* Go Version: `17.4.0`
* JAVA Version: `Available from http://<go-server-host>/go/about`
* OS: `Available from http://<go-server-host>/go/about`
* Browser vendor and version (if relevant): `Hopefully not Internet Explorer 7 :)`

##### Steps to Reproduce

1. Install an agent using the yum repository
2. Start the agent

##### Expected Results

logs to be in one location - /var/log/go-agent

##### Actual Results

Logs are across multiple locations - /var/log/go-agent, /var/lib/go-agent/logs

varshavaradarajan:
This is only when the agent upgrades from a version before 17.4.0.

PR #3763, opened 2017-08-16
Additional system properties that can be passed to the bootstrapper

ketan:
Primarily setup in order to pass `-Dgo.agent.console.stdout` to the
bootstrapper process.

PR #3768, opened 2017-08-18
Log to the appropriate locations on the agent. (#3588)

varshavaradarajan:


agent-launcher/src/com/thoughtworks/go/agent/launcher/AgentLauncherImpl.java
```diff
@@ -57,6 +58,7 @@ public AgentLauncherImpl(AgentProcessParentRunner agentProcessParentCreator) {
 
     public int launch(AgentLaunchDescriptor descriptor) {
         Thread shutdownHook = null;
+        new SystemEnvironment().initializeAgentLogDir();
```
ketan: We probably need do this in `AgentBootstrapper#main` as well, just before `LogConfigurator.initialize()` as well.

agent/resources/config/agent-log4j.properties
```diff
@@ -25,6 +25,6 @@ log4j.logger.org.apache.http.wire=ERROR
 log4j.appender.FILE=org.apache.log4j.RollingFileAppender
 log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
 log4j.appender.FILE.layout.ConversionPattern=%d{ISO8601} [%-9t] %-5p %-16c{4}:%L %x- %m%n
-log4j.appender.FILE.File=logs/go-agent.log
+log4j.appender.FILE.File=${go.agent.log.dir}/go-agent.log
```
ketan: Please change `LinuxPackagingTask#prepareFileSystem` to work with this change.

base/src/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -833,6 +833,13 @@ public long getReAuthenticationTimeInterval() {
         return REAUTHENTICATION_TIME_INTERVAL.getValue();
     }
 
+    public void initializeAgentLogDir() {
+        if (getPropertyImpl("go.agent.log.dir") == null) {
+            String logDir = getEnvironmentVariable("LOG_DIR") == null ? "logs" : getEnvironmentVariable("LOG_DIR");
+            setProperty("go.agent.log.dir", logDir);
+        }
```
ketan: Make a similar change in `AgentOutputAppender#getEffectiveLogDirectory`

agent-launcher/src/com/thoughtworks/go/agent/launcher/AgentLauncherImpl.java
varshavaradarajan: Don't think this is required. It won't help people who don't upgrade their bootstrappers, If someone does use a new bootstrapper, then, the location will either be logs or /var/log/go-agent anyway. Which is why I didn't modify the`agent-bootstrapper-log4j.properties`to use the system property. But, if you think we must do it just for the sake of consistency...

ketan:
Yes, I suggested it for sake of consistency.

agent-process-launcher/src/com/thoughtworks/go/agent/AgentOutputAppender.java
```diff
@@ -75,7 +75,7 @@ private RollingFileAppender rollingAppender(String file) throws IOException {
     }
 
     private static String getEffectiveLogDirectory(String file) {
-        String logDir = System.getenv(LOG_DIR);
+        String logDir = new SystemEnvironment().getPropertyImpl("go.agent.log.dir");
```
jyotisingh: That's not correct. `System.getenv(LOG_DIR)` is not the same as `System.getProperty('go.agent.log.dir')`. agent.sh has `LOG_DIR` set to logs or '/var/log/go-agent/'

varshavaradarajan: The property is set at the time of initialization in AgentLauncherImpl. Just verified that it works. Will check for linux

agent/resources/config/agent-log4j.properties
varshavaradarajan: Is this required to do? We initialize the system property in agent bootstrapper, launcher and AgentMain. Supposing somebody using the agent rpm sets the go.agent.log.dir property, then, because we have hardcoded the /var/log/go-agent bit in all 3 log4j properties, this system property won't be used.

varshavaradarajan:
@ketan , @jyotisingh - Because we first check if the system property is set, https://github.com/gocd/gocd/pull/3768/files#diff-d4a6f9a20334b19a72d0fc579b989559R837, passing the system property `go.agent.log.dir` using the `AGENT_BOOTSTRAPPER_JVM_ARGS` as per https://github.com/gocd/gocd/pull/3763 will only result in the following - 
1. `go-agent-bootstrapper.out.log` will be in the location specified by the LOG_DIR env var (because STD_OUT_LOG_FILE is being set in agent.sh based on the environment variable)
2. `go-agent.log` will be in the location specified by the LOG_DIR env var (because the system property specified using `AGENT_BOOTSTRAPPER_JVM_ARGS`  will not be passed to the agent process.
3. go-agent-bootstrapper.log, go-agent-launcher.log, go-agent-stderr.log, go-agent-stdout.log will be in the location specified by the system property passed to AGENT_BOOTSTRAPPER_JVM_ARGS. 

So, I think that we should get rid of the check for whether the system property is set (https://github.com/gocd/gocd/pull/3768/files#diff-d4a6f9a20334b19a72d0fc579b989559R837)  before setting the system property to the LOG_DIR environment variable. This means that the only way to configure the agent log directory will be through the environment variable. We'll have to change agent.sh somewhere around https://github.com/gocd/gocd/blob/master/installers/go-agent/release/agent.sh#L114 to allow users to truly configure the LOG_DIR environment variable. For those people who don't upgrade their bootstrapper, they won't have an option to configure the environment variable. But the changes that is made in this pull request should enable them to solve the multiple logs location issue. 

Is this okay?

jyotisingh:
Always relying on LOGS_DIR environment variable to set the system property `go.agent.log.dir` sounds okay to me. 

I think the current PR still doesn't take care of setting the property when a older version of bootstrapper(actually launcher) is installed on the machine, and hence the stdout and stderr logs would still go to the `logs` folder instead of `/var/log/go-agent/` folder for instance. Just verify if this works.

ketan:
Closing in favor of #3831

PR #3769, opened 2017-08-18
Pipeline Group API should return template name for pipelines using a template instead of stage names.

varshavaradarajan:


varshavaradarajan:
Apparently the `api/config/pipeline_groups` is used in rendering the VSM which needs to show the stage names.

Issue #3074, opened 2017-01-04
Provide local API for the agent


lenucksi:
##### Issue Type
- Feature proposal

##### Summary
To make usage with elastic agent deployments nicer I would like to be able to retrieve the current status of an agent via an API that is available on the machine the agents runs on. 
This would remove the need to ask the server for the agent's status when this knowledge is available locally and only needed locally.

Such an API would ideally allow the following:
 - Provide information similar to the agent object illustrated here: https://api.gocd.io/current/#the-agent-object
 - Provide measures to force an agent to stop taking new work or tell the server to retire it.

moritzheiber:
A use-case for this would be environments where there are "native" ways for determining an instance's/service's "health" (akin to [AWS ASGs](http://docs.aws.amazon.com/autoscaling/latest/userguide/healthcheck.html), [Kubernetes liveness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/)) but there is not way to reliably tell an agent's status from looking at the resources available on the host running the agent itself.

ketan:
Would you like to provide examples of expected API responses (status+body) in various situations?

moritzheiber:
I really like the way [Vault is handling health checks](https://www.vaultproject.io/api/system/health.html). Essentially, it would be a request to an endpoint exposed on the agent:

```
curl -X GET http://localhost:8155/health
```
with a few parameters perhaps:

- `notconnectedok`: Still return `200` if an agent is not connected to a server 
- `notactiveok`: Still return `200` when an agent is connected but inactive

but that would just be icing on the cake, honestly. For now, a simple endpoint telling me "I'm alive, connected to a server and ready to execute builds (i.e. "active"), would suffice.

ketan:
> but that would just be icing on the cake, honestly. For now, a simple endpoint telling me "I'm alive, connected to a server and ready to execute builds (i.e. "active"), would suffice.

Are you sure that'll suffice?

On the server end, we publish multiple statues for the agent — Idle, Building, LostContact, Missing, Building, Unknown.

Most are self explanatory, but the `LostContact` and `Missing` are interesting in that `Missing` is when an agent first disconnects, if the agent stays `Missing` for about 5 minutes, we assume it to have `LostContact` this is to deal with situations like intermittent network outages, the server being unavailable for short intervals of time (restarts etc). Would you need something more detailed than a boolean indicating "active"?

moritzheiber:
[...]

PR #3781, opened 2017-08-25
Health check api for agent (#3074)

ketan:
This adds 2 endpoints for testing the health of the agent <-> server comm

* Perform a `GET` on `/health/{v1,latest}/isConnectedToServer`, will
  return status `200` along with text `OK!` if the agent is:
  - able to establish HTTP(s) contact
  - is authorized by the server (by an admin, or by auto-registration)
  The endpoint will return status `500` in any other cases.
  Two consecutive failing pings will be treated as having failed healthcheck.

These health checks will (by default) bind to port 8152 on localhost,
but can be configured by the following system properties:

* `go.agent.status.api.enabled`. Defaults to `true`. Set to false to
  disable health check api endpoint
* `go.agent.status.api.bind.host`. Defaults to `localhost`. Set to a
  specific ip address or hostname to bind to that host. Set to `0.0.0.0`
  to bind to all network interfaces.
* `go.agent.status.api.bind.port`. Defaults to `8152`. Set to `0` to
  use an ephemeral port, which will be displayed in a log statement.

In case the http server is unable to bind (usually because of port
conflicts, or multiple agents running on the same macchine), a warning
will be emitted to the console log and agent startup will continue.

### DIFF
--- a/agent/src/com/thoughtworks/go/agent/statusapi/HttpHandler.java
+++ b/agent/src/com/thoughtworks/go/agent/statusapi/HttpHandler.java
@@ -14,9 +14,19 @@
  * limitations under the License.
  */
 
-package com.thoughtworks.go.config;
+package com.thoughtworks.go.agent.statusapi;
 
-public interface AgentRegistry {
-    String uuid();
-    String serverBaseUrl();
+import fi.iki.elonen.NanoHTTPD;
+import fi.iki.elonen.NanoHTTPD.Response.Status;
+
+abstract class HttpHandler {
+    NanoHTTPD.Response process() {
+        if (isPassed()) {
+            return NanoHTTPD.newFixedLengthResponse(Status.OK, "text/plain; charset=utf-8", "OK!");
+        } else {
+            return NanoHTTPD.newFixedLengthResponse(Status.SERVICE_UNAVAILABLE, "text/plain; charset=utf-8", "Bad!");
+        }
+    }
+
+    protected abstract boolean isPassed();
 }
--- a/agent/build.gradle
+++ b/agent/build.gradle
@@ -33,6 +33,7 @@ dependencies {
   compile group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: '1.47'
   compile group: 'commons-configuration', name: 'commons-configuration', version: '1.10'
   compile group: 'org.eclipse.jetty.websocket', name: 'websocket-client', version: versions.jetty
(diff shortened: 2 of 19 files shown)


ketan:
@varshavaradarajan — can you check if this health check is going to be sufficient to see if an agent is running well and connected?

agent/src/com/thoughtworks/go/agent/statusapi/HttpHandler.java
```diff
@@ -0,0 +1,32 @@
...
+
+import fi.iki.elonen.NanoHTTPD;
+import fi.iki.elonen.NanoHTTPD.Response.Status;
+
+abstract class HttpHandler {
+    NanoHTTPD.Response process() {
+        if (isPassed()) {
+            return NanoHTTPD.newFixedLengthResponse(Status.OK, "text/plain; charset=utf-8", "OK!");
+        } else {
+            return NanoHTTPD.newFixedLengthResponse(Status.SERVICE_UNAVAILABLE, "text/plain; charset=utf-8", "Bad!");
```
varshavaradarajan: I don't think this should be 503. If the agent is in pending state (i.e if /health/isConnectedToServer is false), then the service is available, its just that the agent is not connected to the server.

moritzheiber:
Honestly, the latest changes completely disregard the intent behind the issue. Now, again, one has to check in two different places at once to figure out whether an agent is healthy..

ketan:
> Honestly, the latest changes completely disregard the intent behind the issue. Now, again, one has to check in two different places at once to figure out whether an agent is healthy.

I don't believe that's the case. Let me clarify:

Based on your comment (https://github.com/gocd/gocd/issues/3074#issuecomment-307735492):

> but that would just be icing on the cake, honestly. For now, a simple endpoint telling me "I'm alive, connected to a server and ready to execute builds (i.e. "active"), would suffice.

That's precisely what this PR will now do:

`/isConnected` endpoint that answers whether the agent is ready to build, i.e. it is:
- able to connect over HTTP(s) AND
- is approved to run builds, either by an admin or via auto-registration

The `/isApproved` which was added in the first commit has been removed because there's no way to answer that question until the agent is able to establish contact with the server. We can re-look at other endpoints based on feedback.

Is this not what you're expecting?

ketan:
I've updated my initial comment(https://github.com/gocd/gocd/pull/3781#issue-252836564) to clearly document the behavior.

moritzheiber:
Okay, that makes sense, thank you for clarifying.

PR #3794, opened 2017-08-29
Use a different system property for onejar

ketan:


installers/go-agent/win/config/wrapper-agent.conf
```diff
@@ -64,8 +64,7 @@ wrapper.app.parameter.2=-serverUrl
 wrapper.app.parameter.3=%GO_SERVER_URL%
 
 # Additional properties to pass to the JVM
-wrapper.java.additional.1=-Done-jar.main.class=com.thoughtworks.go.agent.bootstrapper.AgentBootstrapper
```
ketan: @jyotisingh — can I get your OK on this? We don't need to override this property, because there's no need to override the default classname.

PR #3800, opened 2017-09-01
Handle boolean and numeric value as a string from JSON to ConfigurationValue.java deserialization

bdpiparva:
Handle boolean and numeric value as a string ,which allows to design UI for a plugin  with following html elements

* Html element having numeric value e.g. `<input type=number/>`
* Html element having boolean value e.g. `<input type=checkbox/>`

Previously, it was failing to deserialize JSON to `ConfigurationValue` object with following error:

```
NameError (no constructor for arguments (org.jruby.RubyBoolean.True) on
Java::ComThoughtworksGoDomainConfig::ConfigurationValue):
  app/presenters/api_v1/config/plugin_configuration_property_representer.rb:42:in
`value='
  app/presenters/api_v1/base_representer.rb:104:in `from_hash'
  app/presenters/api_v1/base_representer.rb:104:in `from_hash'
  app/helpers/api_v1/profiles_controller_actions.rb:43:in `create'
  lib/jetty_weak_etag_middleware.rb:35:in `call'
  lib/catch_json_parse_errors.rb:8:in `call'

```


ketan:
@bdpiparva — is this ready to merge now? (the first note still says WIP)

PR #3805, opened 2017-09-01
Improve classloading and resource management

ketan:
Prevents file-descriptor and classloader leaks by ensuring that:

* classloaders for agent-launcher and agent are closed properly
* temporary jar files that are extracted are removed after closing the
  classloaders
* the file watcher from log4j is replaced with a custom implementation

### DIFF
--- a/agent-bootstrapper/src/com/thoughtworks/go/agent/bootstrapper/DefaultAgentLauncherCreatorImpl.java
+++ b/agent-bootstrapper/src/com/thoughtworks/go/agent/bootstrapper/DefaultAgentLauncherCreatorImpl.java
@@ -16,10 +16,11 @@
 
 package com.thoughtworks.go.agent.bootstrapper;
 
-import com.thoughtworks.cruise.agent.common.launcher.AgentLauncher;
 import com.thoughtworks.cruise.agent.common.launcher.AgentLaunchDescriptor;
+import com.thoughtworks.cruise.agent.common.launcher.AgentLauncher;
 import com.thoughtworks.go.agent.common.util.Downloader;
 import com.thoughtworks.go.agent.common.util.JarUtil;
+import com.thoughtworks.go.util.FileUtil;
 import com.thoughtworks.go.util.SystemUtil;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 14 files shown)


agent-bootstrapper/src/com/thoughtworks/go/agent/bootstrapper/DefaultAgentLauncherCreatorImpl.java
```diff
@@ -16,80 +16,92 @@
...
 
-        launcherDestroyed = true;
+    private void createTempLauncherJar() {
+        try {
+            inUseLauncher.getParentFile().mkdirs();
+            Files.copy(new File(Downloader.AGENT_LAUNCHER).toPath(), inUseLauncher.toPath());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        LauncherTempFileHandler.startTempFileReaper();
```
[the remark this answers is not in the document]
jyotisingh: The current PR cleans up the temp-jars at two points, during a graceful shutdown and during bootstrapper restart. If the bootstrapper is an older version say 17.5 but the server was upgraded to 17.11, and a force agent shutdown happens, the temp-jars and the dependencies dir wouldn't be deleted. Since the boostrapper is an older version, these files would be left behind. 
`LauncherTempFileHandler` used to make an attempt at cleaning those up. If not temp-file-reaper, we need something else in place to clean up those files.

agent-bootstrapper/src/com/thoughtworks/go/agent/bootstrapper/LauncherTempFileHandler.java
```diff
@@ -57,8 +56,9 @@ private void reapFiles() {
             for (String fileName : fileSet) {
                 File file = new File(fileName);
                 FileUtils.deleteQuietly(file);
-                JarUtil.cleanup(fileName);
-                if (!file.exists() && !JarUtil.tempFileExist(fileName)) {
+                File depsDir = new File("data", fileName);
```
jyotisingh: There are too any places that we use the `data` folder and refer to it via the string. Can we instead create a constant atleast on the agent side as a part of this PR. Its difficult to see where all we use the data folder. The only way to do so is to look up for the string `"data"` and that gives back way too many results.

base/src/com/thoughtworks/go/logging/LogConfigurator.java
```diff
@@ -82,13 +80,35 @@ protected void initializeFromXMLResource(URL resource) {
     }
 
     protected void initializeFromPropertiesFile(File log4jFile) {
-        watchDog = new PropertyWatchdog(log4jFile.getAbsolutePath());
-        watchDog.setDelay(DELAY);
+        CustomFileWatchdog watchdog = new CustomFileWatchdog(log4jFile) {
```
jyotisingh: Wasn't able to startup Go-server using zip installer after this change, it fails with 
```
Caused by: java.lang.NoClassDefFoundError: com/thoughtworks/go/logging/LogConfigurator$1
	at com.thoughtworks.go.server.util.GoLauncher.main(GoLauncher.java:41)
	... 6 more
```

ketan: Fixed

PR #3808, opened 2017-09-04
Handle NullPointerException under GoCache

GaneshSPatil:
* First create call to any entity using API fails due to 'entity cache removal' request throwing a null pointer exception
  This happens due no no parent key GO_ETAG_CACHE present in the GoCache.
  Check for the parent GO_ETAG_CACHE key before removal of the entity subkey


varshavaradarajan:
@GaneshSPatil , will merge this and the other PR you're about to send together. Don't want multiple builds to kick off.

PR #3809, opened 2017-09-04
Update material watchlist for config repo materials when a config repo is managed using API

GaneshSPatil:


server/test/unit/com/thoughtworks/go/server/materials/SCMMaterialSourceTest.java
```diff
@@ -150,6 +151,32 @@ public void shouldReloadSchedulableMaterialsOnConfigChange() {
...
+    public void shouldReloadSchedulableMaterialsOnConfigRepoChange() {
+        Set<MaterialConfig> schedulableMaterialConfigs = new HashSet<>(Arrays.asList(svnMaterial.config()));
+
+        when(goConfigService.getSchedulableSCMMaterials()).thenReturn(schedulableMaterialConfigs);
+        when(materialConfigConverter.toMaterials(schedulableMaterialConfigs)).thenReturn(new HashSet<>(Arrays.asList(svnMaterial)));
+
+        Set<Material> materials = source.materialsForUpdate();
+
+        assertThat(materials.size(), is(1));
+        assertTrue(materials.contains(svnMaterial));
```
varshavaradarajan: This isn't required.  Its been tested in another test. Since you are mocking the same thing after this, its not necessary to check if schedulable materials contained something before. https://github.com/gocd/gocd/pull/3809/files#diff-4844feeb7ab629be0b9f429e848090dbR166 ensures that there had been another shedulable material present.

server/test/unit/com/thoughtworks/go/server/materials/SCMMaterialSourceTest.java
```diff
@@ -150,6 +151,26 @@ public void shouldReloadSchedulableMaterialsOnConfigChange() {
     }
 
     @Test
+    public void shouldReloadSchedulableMaterialsOnConfigRepoChange() {
+        Set<MaterialConfig> schedulableMaterialConfigs = new HashSet<>(Arrays.asList(svnMaterial.config()));
+
+        when(goConfigService.getSchedulableSCMMaterials()).thenReturn(schedulableMaterialConfigs);
+        when(materialConfigConverter.toMaterials(schedulableMaterialConfigs)).thenReturn(new HashSet<>(Arrays.asList(svnMaterial)));
```
varshavaradarajan: This setup is also not needed, right?

PR #3822, opened 2017-09-05
Cleanup `SystemEnvironment` and friends

ketan:


ketan:
Keep reviewing, don't merge until I say so.

PR #3823, opened 2017-09-05
Remove old config save code (Fixes #3094)

### LINKED ISSUE #3094: Cleanup config save old code
Config Save Optimisation https://github.com/gocd/gocd/pull/2912 adds a new flow for config save, with this there is duplicate code to save config. 

With 17.1 users have an option to toggle between optimised flow or the old flow using the system environment ```optimize.full.config.save```. The reason to retain the old flow was to take feedback on the new changes for few releases and then cleanup the old code.

The cleanup should be done as part of 17.3 or 17.4 release unless we find any issues with the new changes.

ketan:


ketan:
@maheshp I'm having a hard time fixing one of the tests, I believe it's one line of code somewhere, but I can't put my finger on it, can you take a look at this?

maheshp:
@ketan I had a quick look at the config flow, one of the things I remember is cleaning up the old config save code in [GoConfigMigration](https://github.com/gocd/gocd/blob/master/config/config-server/src/com/thoughtworks/go/config/GoConfigMigration.java#L101). 

My intent while creating #3094 was to remove all the code pertaining to old full config save flow, which would involve changes to config migration, config save, tests etc. 

With this PR we would remove the invocation of the old config save methods but some of the old code will still be around. 

I think if we intend to cleanup we should go through all the different flows and get rid of old config save code, which I think could be time consuming.

What do you think?

ketan:
> I think if we intend to cleanup we should go through all the different flows and get rid of old config save code, which I think could be time consuming.

We can't be keeping old code which is unused. So one way or the other it needs to go, not sure what's the path of least resistance to get to it.

maheshp:
Roughly I think this might be few days of work, since I have context I can have a look at it whenever we prioritise this work.

PR #3825, opened 2017-09-06
[WIP] Prevent update of agentInstance with a stale runtime info

GaneshSPatil:
* There is a chance of a race condition between BuildAssignmentService.assignWorkToAgent and AgentInstances.updateAgentRuntimeInfo. Basically, BuildAssignmentService.assignWorkToAgent would assign some job say J1 to an agent A1. Immidiately after this AgentInstances.updateAgentRuntimeInfo updates the agentA1  to be idle. For Websocket enabled agents, the agent A1 would be considered for assignment again as the current status is idle. This causes cancellation of the job J1 that was assigned to the agent A1 moments ago. While this can happen on all instances, this is seen more often when the server is under heavy load and processing slowly.

* This commit attempts to record the AgentRuntimeInfo.lastUpdatedTime and uses that to discard updates if the reported runtime info is found to be older .


GaneshSPatil:
merging this as [PR build](https://build.gocd.org/go/pipelines/value_stream_map/trigger/1580) passed.

Issue #3814, opened 2017-09-04
Change logging framework from log4j 1.2


ketan:
##### Issue Type

- Feature enhancement

##### Summary

We've pretty much moved off log4j 1.2 to slf4j for the `Logger` interface, except for these 3 places:

- [`LogConfigurator`](https://github.com/gocd/gocd/blob/6aef9b9b1414cff903ac3b60826d012d52f2abe1/base/src/com/thoughtworks/go/logging/LogConfigurator.java)

    This one merely reloads the log configuration automatically when things change, something that comes out-of-the-box with log4j2 and logback

- [`DefaultPluginLoggingService`](https://github.com/gocd/gocd/blob/6aef9b9b1414cff903ac3b60826d012d52f2abe1/plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/service/DefaultPluginLoggingService.java)

    Removing this would mean either:

    - (preferred approach) all logs go to the same file (`go-server.log`), probably with a marker containing the plugin ID

   -OR-

    - we can apply a NDC/MDC+filter to send plugin logs to a single `plugin.log` file

    For developers building plugins, they can always specify additional appenders to filter their own log files to help ease development. For end users wanting to diagnose problems they may either grep for plugin markers in the main log, or use a (temporary) appender for a specific plugin.

- [`AgentOutputAppender`](https://github.com/gocd/gocd/blob/6aef9b9b1414cff903ac3b60826d012d52f2abe1/agent-process-launcher/src/com/thoughtworks/go/agent/AgentOutputAppender.java)

    This currently uses log4j's `RollingFileAppender` because there's no other library (that I'm aware of) which offers a similar capability, but this can be replaced with another (custom?) implementation fairly easily without affecting any end-user configuration.

Since log4j1.2 has been EOL'ed for the past 2 years and we've suffered a few bugs to classloader leaks along with issues like [this](#3521).

The choice of logging frameworks leave us with 2 possible options logback(preferred) and log4j2, both of which will require a migration of the `log4j` properties that are shipped with GoCD. For most users who do not change these files, the change in the underlying framework will not affect an upgrade. For users who have customized the log4j.properties files for some reason, they will have to "migrate" over their changes to a newer (xml?) format that's compatible with log4j2 or logback.

[...]

PR #3831, opened 2017-09-06
Move over to logback (#3814)

ketan:


ketan:
TODO:

- [ ] verify effect of log level level changes on plugin logs
- [ ] verify log rotation
- [ ] verify `<include />` directive in log configuration
- [ ] verify if log files are being reloaded

varshavaradarajan:
As discussed - 
1. There is the cruise.config.dir and cruise.config.file system properties. Make use of one everywhere for consistency. This could go into https://github.com/gocd/gocd/pull/3822 after we find out the repercussions of getting rid of one of these system properties.  
2. Rename go.agentBootstrapper.log.dir and go.agentLaucher.log.dir to go.agent.log.dir
3. Check for the system property first and then the env vars as we intend to use the system property in the future in favor of environment variables.

ketan:
Documentation is over at https://github.com/gocd/docs.go.cd/pull/116

server/properties/src/config/logback.xml
```diff
@@ -0,0 +1,106 @@
...
+    xmlns="http://ch.qos.logback/xml/ns/logback"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://ch.qos.logback/xml/ns/logback http://ch.qos.logback/xml/ns/logback/logback.xsd"
+    debug="${gocd.server.logback.debug:-false}"
+    scan="${gocd.server.logback.scan:-true}"
+    scanPeriod="${gocd.server.logback.scanPeriod:-5 seconds}"
+>
+
+  <appender name="FileAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
+    <file>${gocd.server.log.dir:-logs}/go-server.log</file>
```
varshavaradarajan: This will be messed up on linux. It should be /var/log/go-server in linux. People only configure the gocd.server.log.dir if they don't want the default.

varshavaradarajan: Remember the logic in LinuxPackagingTask to replace logs with /var/log/go-server..?

ketan: I suppose you're hinting at setting `gocd.server.log.dir` somewhere in `server.sh`?

varshavaradarajan: Yep. If /var/log/go-server directory exists, then set this system property to /var/log/go-server

varshavaradarajan:
Got this on Windows - 
```
WARN   | wrapper  | 2017/09/13 22:12:18 | Unable to write to the configured log directory: logs (The system cannot find the path specified. (0x3))
WARN   | wrapper  | 2017/09/13 22:12:18 |   The directory does not exist.
WARN   | wrapper  | 2017/09/13 22:12:18 | Unable to write to the configured log file: logs\go-server-wrapper.log (The operation completed successfully. (0x0))
WARN   | wrapper  | 2017/09/13 22:12:18 |   Falling back to the default file in the current working directory: wrapper.log

```

ketan:
> Got this on Windows -
> 
> WARN   | wrapper  | 2017/09/13 22:12:18 | Unable to write to the configured log directory: logs (The system cannot find the path specified. (0x3))
> WARN   | wrapper  | 2017/09/13 22:12:18 |   The directory does not exist.
> WARN   | wrapper  | 2017/09/13 22:12:18 | Unable to write to the configured log file: logs\go-server-wrapper.log (The operation completed successfully. (0x0))
> WARN   | wrapper  | 2017/09/13 22:12:18 |   Falling back to the default file in the current working directory: wrapper.log

I suspect it's an existing problem and not something introduced by this change: https://github.com/gocd/gocd/blob/42e34bd327d415b4ea44f747f6980352c6532ac2/installers/go-server/win/config/wrapper-server.conf#L95-L96

varshavaradarajan:
Yes, it was introduced in 17.4.0 when we introduced the logs dir.

server/properties/src/config/logback.xml
```diff
@@ -26,63 +26,63 @@
 >
 
   <appender name="FileAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
-    <file>${gocd.server.log.dir:-logs}/go-server.log</file>
+    <file>${gocd.server.log.dir:-${GO_SERVER_LOG_DIR:-logs}}/go-server.log</file>
```
varshavaradarajan: Do we need to check for logs? That seems to be taken care of in `server.sh`

ketan: Yes, you always seem to keep forgetting windows installers and batch files.

varshavaradarajan: Oh, sorry. I didn't forget windows. I had had a conversation with @jyotisingh about making GO_SERVER_SYSTEM_PROPERTIES and GO_AGENT_SYSTEM_PROPERTIES on windows consistent with linux that would eliminate the need for this. That's another conversation. And another PR, if we ever get there.

installers/go-server/release/server.sh
```diff
@@ -78,14 +78,15 @@ SERVER_DIR="$(cd "$CWD" && pwd)"
...
-    STDOUT_LOG_FILE=go-server.out.log
-  fi
+if [ -d "${GO_SERVER_LOG_DIR}" ]; then
+  export GO_SERVER_LOG_DIR="logs"
+  mkdir -p "${GO_SERVER_LOG_DIR}"
+else
+  export GO_SERVER_LOG_DIR="/var/log/go-server"
 fi
 
+STDOUT_LOG_FILE="${GO_SERVER_LOG_DIR}/go-server.out.log"
```
varshavaradarajan: This is a breaking change for those using zip. They will have 2 go-server.out.log files. Do we have to give notice before doing it?

ketan: No notice is needed, I think. We can mention this change in the release notes.

agent-process-launcher/src/com/thoughtworks/go/agent/AgentOutputAppender.java
```diff
@@ -63,25 +73,33 @@ public void write(String line) {
...
     private static String getEffectiveLogDirectory(String file) {
-        String logDir = System.getenv(LOG_DIR);
-
-        if (isBlank(logDir)) {
-            return file;
-        } else {
-            return logDir + "/" + file;
+        String logDir;
+        // LOG_DIR is for backward compatibility with older linux installers which export `LOG_DIR`
+        logDir = System.getenv("GO_AGENT_LOG_DIR") == null ? System.getenv("LOG_DIR") : System.getenv("GO_AGENT_LOG_DIR");
```
varshavaradarajan: Don't we have to check for the system property first too?

agent-process-launcher/src/com/thoughtworks/go/agent/AgentOutputAppender.java
```diff
@@ -63,25 +73,33 @@ public void write(String line) {
     }
 
     public void close() {
```
varshavaradarajan: Seems like there are certain methods that can be pkg local

agent/resources/config/agent-logback.xml
```diff
@@ -0,0 +1,54 @@
...
+    xmlns="http://ch.qos.logback/xml/ns/logback"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://ch.qos.logback/xml/ns/logback http://ch.qos.logback/xml/ns/logback/logback.xsd"
+    debug="${gocd.agent.logback.debug:-false}"
+    scan="${gocd.agent.logback.scan:-true}"
+    scanPeriod="${gocd.agent.logback.scanPeriod:-5 seconds}"
+>
+
+  <appender name="FileAppender" class="ch.qos.logback.core.rolling.RollingFileAppender">
+    <file>${GO_AGENT_LOG_DIR:-${LOG_DIR:-${gocd.agent.log.dir:-logs}}}/go-agent.log</file>
```
varshavaradarajan: Change this to check system property first?

agent/resources/config/agent-logback.xml
```diff
@@ -0,0 +1,54 @@
...
+    <encoder>
+      <pattern>
+        ${gocd.agent.logback.defaultPattern:-%date{ISO8601} %-5level [%thread] %logger{0}:%line - %msg%n}
+      </pattern>
+    </encoder>
+
+    <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+      <fileNamePattern>
+        ${GO_AGENT_LOG_DIR:-${LOG_DIR:-${gocd.agent.log.dir:-logs}}}/go-agent.log.%d{yyyy-MM-dd}.%i.gz
```
varshavaradarajan: Same as above

build.gradle
```diff
@@ -186,11 +186,17 @@ subprojects {
         runtimeClasspath += configurations.compileOnly
       }
       resources {
-        srcDirs = ['test-resources']
+        srcDirs = ['test-resources', rootProject.findProject(':test-utils').file("resource-include-in-all-projects")]
       }
     }
   }
 
+  task cleanLogs(type: Delete) {
```
varshavaradarajan: Why does this need to be a separate task? It can go in clean itself.

agent/build.gradle
```diff
@@ -55,7 +55,6 @@ clean {
   delete 'plugins_work'
   delete 'agent-plugins.zip'
   delete 'tfs-impl.jar'
-  delete 'logs'
```
varshavaradarajan: Why not delete logs? We do it for server.

common/test/unit/com/thoughtworks/go/domain/DownloadActionTest.java
```diff
@@ -16,10 +16,10 @@
 
 package com.thoughtworks.go.domain;
 
+import ch.qos.logback.classic.Level;
```
varshavaradarajan: Copyright

build.gradle
ketan: Since tests in each project generates a logs of its own, it's easier to add it to all projects, than to add `logs` to clean of each project.

agent/build.gradle
ketan: See `cleanLogs` task added to all projects.

PR #3837, opened 2017-09-07
Fixes #3836. Updated footer links to point to installed version's api and docs pages

### LINKED ISSUE #3836: Update footer links to point to api and docs of "my" GoCD version
##### Issue Type

- Feature enhancement

##### Summary

The links in the footer point to the latest for docs and version selector for api. This results in me reading documentation different than that of my version

mikhailadvani:


mikhailadvani:
Have fixed the login page. As advised, did a search across the repo for /current. Found many links pointing to https://docs.gocd.org/current/something. Would like to tackle that separately

ketan:
> Have fixed the login page. As advised, did a search across the repo for /current. Found many links pointing to https://docs.gocd.org/current/something. Would like to tackle that separately

There's also a bunch of them pointing to `https://api.gocd.org/#foo`. Do you think you can change them to `https://api.gocd.org/VERSION/#foo` as well — mostly involve calling the singleton `com.thoughtworks.go.CurrentGoCDVersion.getInstance().goVersion()` directly in places like these https://github.com/gocd/gocd/blob/3ed499b57d196ab61cdcbe8500eb1095aa8f81c7/server/webapp/WEB-INF/rails.new/app/presenters/api_v1/shared/agent_summary_representer.rb#L27-L27

mikhailadvani:
I would propose to tackle that separately. Found a lot of links pointing to both docs.gocd.org and api.gocd.org. Would want to extract it out to a single definition as part of that activity. Will help us manage DNS changes if any in the future in a simpler and cleaner way. Views?

Issue #3849, opened 2017-09-13
Support for configuring parameters in the Configuration repository Extension 


PredatorVI:
I'm looking at using the JSON pipeline as code plugin to migrate my pipeline definitions to a JSON format in GIT.  However, we use "parameters" to make the tasks more generic.  As an example, each deployment pipeline has the following (or simliar) defined:
```
      <params>
        <param name="DEPLOY_ENV">acc</param>
      </params>
```
Subsequently this value is passed as a parameter or command-line argument to the various tasks using #{DEPLOY_ENV}.  However, because this feature isn't supported for some reason, It is complicating our pipeline definitions.  

It would be nice to have this supported and available ultimately through the pipeline as code mechanisms.

rajiesh:
Verified on 17.11.0 (5520-9f6909e2f64b07d2dce5cecd4ea5b92b8e19d6b1)

varshavaradarajan:
The json expected by the GoCD server from any config repo plugin is - 

```
{
  "group": "...",
  "name": "...",
  "label_template": "...",
  "template": null,
  "enable_pipeline_locking" : false, # boolean
  "parameters": [
    {
      "name": "param",
      "value": "parameter_value"
    }
  ],
  "mingle": {
    "base_url": "...",
    "project_identifier": "..."
  },
  "tracking_tool": null,
  "timer": {
    "spec": "..."
  },
  "environment_variables": [],
  "materials": [
    ...
  ],
  "stages": [
    ...
  ]
}
```
#### Referencing an existing template 

```
{
  "group": "...",
  "name": "...",
  "label_template": "...",
  "enable_pipeline_locking" : false,
  "template": "existing-template",
  "parameters": [
    {
      "name": "param",
      "value": "parameter-value"
    }
  ],
  "materials": [
    ...
  ]
}
```

PR #3851, opened 2017-09-14
Config repo support for parameters and referencing templates from pipelines.

varshavaradarajan:


varshavaradarajan:
#3849

tomzo:
@varshavaradarajan This looks good. I don't see anything wrong in these changes. But I think it would be nice to add one integration test case at higher level with main XML config with some template and a config-repo pipeline referencing it. That way we would have a test for all preprocessors used together.
Have a look at [these tests](https://github.com/gocd/gocd/blob/4a7363bbcb7919c81de5482d8e814c714ad3f75e/server/test/integration/com/thoughtworks/go/config/CachedGoConfigIntegrationTest.java#L349) as example. You can write the remote pipeline in XML there like [here](https://github.com/gocd/gocd/blob/4a7363bbcb7919c81de5482d8e814c714ad3f75e/server/test-resources/external_git_config_repo_referencing_first/downstream.gocd.xml) (no plugin is needed). It is a hidden feature, which we have for tests. It will bypass the `ConfigConverter` but it would use preprocessors in order.

Also, @jyotisingh wrote the preprocessors part so she might have better insight here than me.

varshavaradarajan:
Yes, I am writing the integration test.

jyotisingh:
@varshavaradarajan - looks good to me. 
As @tomzo mentioned, just write an integration test. Let me know if you need any help with that.

PR #3856, opened 2017-09-15
Go Server session timeout is configurable now

maheshp:
* Session `MaxInactiveInterval` can be configured using the system property `go.server.session.timeout.seconds` defaulted to 1209600 seconds(two weeks)


jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
```diff
@@ -87,6 +88,8 @@ public void configure() throws Exception {
         server.setHandler(handlers);
         performCustomConfiguration();
         server.setStopAtShutdown(true);
+        SessionManager sessionManager = webAppContext.getSessionHandler().getSessionManager();
```
ketan: This is inconsistent with how `setCookieExpirePeriod` is setup, can we make it consistent, otherwise looks good to merge.

PR #3858, opened 2017-09-15
Adding an explicit `vsm` link #3705

maheshp:
* Moving away from linking pipeline label to vsm.
* Adding a explicit vsm link which should make it more discoverable

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
@@ -321,10 +321,10 @@ Graph_Renderer = function (container) {
         gui += '<li class="instance">';
         if (instance.label != '') {
             if (isCurrent) {
-                gui += '<h4 title="' + instance.label + '"><span class="pipeline_run_label">Instance:' + instance.label + '</span></h4>';
+                gui += '<h4 title="' + instance.label + '"><span class="pipeline_run_label">Instance: ' + instance.label + '</span></h4>';
             }
             else {
-                gui += '<h4 title="' + instance.label + '"><span class="pipeline_run_label">Instance:' + instance.label + '</span><span class="vsm_link_wrapper"><a href="' + instance.locator + '">' + 'VSM' + '</a></span></h4>';
+                gui += '<h4 title="' + instance.label + '"><span class="pipeline_run_label">Instance: ' + instance.label + '</span><span class="vsm_link_wrapper"><a href="' + instance.locator + '">' + 'VSM' + '</a></span></h4>';
             }
         }
         if(instance.locator.trim() != "") {
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
@@ -328,6 +328,9 @@ Graph_Renderer = function (container) {
             }
         }
         if(instance.locator.trim() != "") {
+            var duration =  pipelineRunCompleted(instance) ? pipelineRunDuration(instance) : 'In Progress';
+            gui += '<span class="duration">Duration: ' + duration + '</span>';
+
             gui += '<ul class="stages">';
             stagesCount = instance.stages.length;
             for (var i = 0; i < stagesCount; i++) {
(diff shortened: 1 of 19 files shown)


jyotisingh:
Since we are moving away from linking the `label` to VSM, should we also have a separate link for VSM of material VSM as well? Currently VSM for material looks like this where the `revision` links to the VSM:

screen shot 2017-10-12 at 5 53 46 pm

jyotisingh:
Currently this link seems to be floating so the dashboard so the link appears at different places if pipelines have different labelling config. Is that ok? eg:

screen shot 2017-10-12 at 6 36 06 pm

arvindsv:
No. It was supposed to be styled differently. Right-aligned. Or, next to "Changes" or something. Check with @naveenbhaskar?

server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
```diff
@@ -321,13 +321,16 @@ Graph_Renderer = function (container) {
         gui += '<li class="instance">';
         if (instance.label != '') {
             if (isCurrent) {
-                gui += '<h4 title="' + instance.label + '"><i class="label">Label: </i>' + instance.label + '</h4>';
+                gui += '<h4 title="' + instance.label + '"><span class="pipeline_run_label">Instance:' + instance.label + '</span></h4>';
```
jyotisingh: Could you write javascript tests to get this duration verified for different cases?

jyotisingh:
The `vsm` link in the changes popup on dashboard and pipeline-history page needs to be uppercase

screen shot 2017-10-12 at 6 56 16 pm

arvindsv:
Talked to @maheshp. Talked about inconsistency between VSM link on dashboard and on VSM, etc. Let's go with how it is now, and then move it if needed.

arvindsv:
Unless anyone has a better idea, before it is merged.

eeluke:
Just seeing all of this - yes, it is intentional to have VSM to the right
of "changes" with a line separating it from "changes". Also, the VSM link
should be underlined everywhere it appears. To standardize, "Changes" and
"Compare" should also be underlined. Please let me know if you have any
other questions.

Emily

maheshp:
@eeluke after talking to @naveenbhaskar we wanted to keep the VSM link position consistent and evident by having it next next to _Instance_ on dashboard, VSM and headers for stage and job detail page. 

I am happy to go with what you guys decide for now, if required we can tweak it later.

naveenbhaskar:
I believe the VSM link should be next to the instance in all the places. @eeluke 
, can you pls tell me the reason to place it on the right side?. May be i am missing something.

 if its near to instance number, its clear that they are related. on the right side, its hanging
Have a look at the image below, with options next to instance and away from instance.

image

arvindsv:
I think the idea was VSM link next to the "Changes" link, @naveenbhaskar. But, you two decide and tell us.

jyotisingh:
I don't have any opinion on where should the link be, so its totally up to you guys. 

Just thought of clarifying what @maheshp and @naveenbhaskar were trying to point at. There are two versions of pipeline instance widget that shows up in GoCD's dashboard. 
1. The pipeline widget that's displayed on the home page has the "Compare" and "Changes" link which looks like this currently:
screen shot 2017-10-13 at 5 28 26 pm

2. The pipeline widget that shows up on VSM page which doesn't have either of these links, looks like this currently:
screen shot 2017-10-13 at 5 28 33 pm

@eeluke - Are you proposing adding VSM link next to "Changes" link on the widget that shows up on the home page, but showing it next to the "Instance" on the VSM page(given that there is no 'changes' link on this page)? If so, is it okay for the link to show up at two different positions on these widgets?

arvindsv:
@jyotisingh I *think* @eeluke is saying we should stay faithful to the original proposal for this, which is [here](https://github.com/gocd/gocd/issues/3705#issue-246145020). So, yes, it will be in different places in the VSM and on the dashboard.

Just trying to save some time, since @eeluke will not be able to respond for a few more hours.

jyotisingh:
@arvindsv - cool

eeluke:
@maheshp @naveenbhaskar @arvindsv This might be too late. Sorry! Let me explain my original intent. I trust you all to make the final call.

My understanding is that on the dashboard pipeline tile, there are a few actions you can take regarding each "instance" - you can compare it to another instance by clicking the "compare" link, and you can see "changes" to the commits to that instance by clicking "changes". I thought that the VSM fit here in this list of actions you can take related to that particular instance. By moving the VSM up closer to the Instance name, we are saying that viewing the VSM is a more important "action" to take related to that instance. If that is true, I'm all for moving that up closer to the instance. If not, let's leave it in the same line as the the other actions. I don't think it's confusing to have it in different places in the VSM vs the dashboard because users view and use those tiles in different ways and it's not so different that people won't be able to find it. 

As for moving it closer to the Instance number on the VSM tile, my concern was that the link would constantly move based on the length of that instance name. If it's always aligned right, it will not shift at all.

Please let me know if you have questions.

eeluke:
This is a very small request, but can we make sure that there is a
character space between the ":" and the number or label after it.

Emily

maheshp:
@arvindsv fixed the long label by truncating it. Is this fine?

screen shot 2017-10-18 at 10 42 06 am
screen shot 2017-10-18 at 10 40 24 am
screen shot 2017-10-18 at 10 44 23 am

jyotisingh:
There are a few things which I think we need to address or atleast talk about. It did not make sense to keep this PR open for the same so opened a new issue => https://github.com/gocd/gocd/issues/3944

eeluke:
@maheshp The truncation looks good to me. BUT, the VSM link is a couple pixels below the instance name. Can we move it up?
vsm link

PR #3859, opened 2017-09-15
Page header style changes

naveenbhaskar:


ketan:
Screenshots at https://github.com/gocd/gocd/issues/3705#issuecomment-329745406

PR #3861, opened 2017-09-15
Removing password file support from core

bdpiparva:
* Use bundled [Password File Plugin](https://github.com/gocd/gocd-filebased-authentication-plugin) instead. Existing Ldap configurations have been moved to `Authorization Configuration`  in **_Admin_** > **_Security_**


ketan:
@bdpiparva — can you rebase, merge and get functional tests green?

PR #3862, opened 2017-09-15
Add support for GitLab and BitBucket webhook

ketan:


ketan:
The webhook url is `http://go-server:8153/go/api/webhooks/gitlab/notify`

Issue #3754, opened 2017-08-10
Do something about exceptions on server boot


ketan:
##### Issue Type

- Bug Report

##### Summary

This is not a bug per se, but on server startup, we see a bunch of exceptions, all of which can be ignored or logged at a different severity. This always seems gives an impression that something went wrong on startup, and since there are several stacktraces to go through, it makes it particularly bad.

```
2017-08-10 16:33:41,588  INFO [main] ConfigRepository:220 - Upgrading config file from version 81 to version 94
2017-08-10 16:33:41,716  WARN [Thread-75] PluginInfoBuilder:55 - Plugin settings configuration and view could not be retrieved. May be because the plugin doesn't have any plugin settings
java.lang.RuntimeException: Interaction with plugin with id 'cd.go.authentication.ldap' implementing 'authorization' extension failed while requesting for 'go.plugin-settings.get-view'. Reason: [The plugin sent a response that could not be understood by Go. Plugin returned with code '500' and the following response: 'Request go.plugin-settings.get-view is not supported by plugin.']
	at com.thoughtworks.go.plugin.access.PluginRequestHelper.submitRequest(PluginRequestHelper.java:41)
	at com.thoughtworks.go.plugin.access.common.AbstractExtension.getPluginSettingsView(AbstractExtension.java:60)
	at com.thoughtworks.go.plugin.access.common.PluginInfoBuilder.getPluginSettingsAndView(PluginInfoBuilder.java:48)
	at com.thoughtworks.go.plugin.access.authorization.AuthorizationPluginInfoBuilder.pluginInfoFor(AuthorizationPluginInfoBuilder.java:44)
	at com.thoughtworks.go.plugin.access.authorization.AuthorizationPluginInfoBuilder.pluginInfoFor(AuthorizationPluginInfoBuilder.java:29)
	at com.thoughtworks.go.plugin.access.common.MetadataLoader.pluginLoaded(MetadataLoader.java:41)
	at com.thoughtworks.go.plugin.infra.DefaultPluginManager$FilterChangeListener.pluginLoaded(DefaultPluginManager.java:254)
	at com.thoughtworks.go.plugin.infra.FelixGoPluginOSGiFramework$2.execute(FelixGoPluginOSGiFramework.java:354)
	at org.apache.commons.collections.CollectionUtils.forAllDo(CollectionUtils.java:389)
[...]

PR #3865, opened 2017-09-18
Silence some exceptions on server startup

ketan:
(Addresses parts of #3754)

On startup the server is sending the plugin settings config and view
messages to all plugins to build plugin info objects. This call blows up
for atleast the task and authorization plugins because these calls are
not supported. Not making these calls will silence some of the
exceptions on startup.

### DIFF
--- a/config/config-server/src/com/thoughtworks/go/service/ConfigRepository.java
+++ b/config/config-server/src/com/thoughtworks/go/service/ConfigRepository.java
@@ -25,6 +25,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.NullArgumentException;
 import org.eclipse.jgit.api.*;
 import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.api.errors.NoHeadException;
 import org.eclipse.jgit.diff.DiffFormatter;
 import org.eclipse.jgit.errors.IncorrectObjectTypeException;
 import org.eclipse.jgit.errors.MissingObjectException;
@@ -174,6 +175,8 @@ public class ConfigRepository {
                 RevCommit revision;
                 try {
                     revision = getCurrentRevCommit();
+                } catch (NoHeadException e) {
+                    return null;
                 } catch (GitAPIException e) {
                     LOGGER.info("[CONFIG REPOSITORY] Unable retrieve current cruise config revision", e);
                     return null;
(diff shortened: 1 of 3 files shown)


plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationPluginInfoBuilder.java
```diff
@@ -41,10 +41,9 @@ public AuthorizationPluginInfo pluginInfoFor(GoPluginDescriptor descriptor) {
 
         PluggableInstanceSettings authConfigSettings = authConfigSettings(descriptor.id());
         PluggableInstanceSettings roleSettings = roleSettings(descriptor.id(), capabilities);
-        PluggableInstanceSettings pluginSettingsAndView = getPluginSettingsAndView(descriptor, extension);
         Image image = image(descriptor.id());
 
-        return new AuthorizationPluginInfo(descriptor, authConfigSettings, roleSettings, image, capabilities, pluginSettingsAndView);
+        return new AuthorizationPluginInfo(descriptor, authConfigSettings, roleSettings, image, capabilities, null);
```
jyotisingh: This and [this](https://github.com/gocd/gocd/pull/3865/files#diff-4604837e13357ba5ea488420ebdbac1eR37) are not correct. The auth plugins written/used by us so far do not allow plugin-settings, but that doesn't mean that the authorization-extension itself doesn't support plugin-settings.

jyotisingh: Even if a plugin-extension supports settings, a plugin could choose not to implement it, so the exception stacktrace from the server startup won't really go away with this. One option could be to mandate plugins to provide a list of capabilities - settings being one of them. We could invoke calls only if the plugin reports having a capability to configure settings. Another option, given the lack of the first one is to use [FelixGoPluginOSGiFramework.doOnWithExceptionHandling](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/FelixGoPluginOSGiFramework.java#L275) for plugin-settings call, the [handler](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/FelixGoPluginOSGiFramework.java#L307) could return null or ignore a plugin if the call is not implemented by it.

PR #3879, opened 2017-09-20
Remove usages of some environment variables

yankurk:
* `PRODUCTION_MODE` and `MANUAL_SETTING` usages in `server.sh` and `agent.sh` were removed.
* Removed usages of undocumented variables that were setup by functional tests.


jyotisingh:
should get rid of `DAEMON=Y` from `go-agent.default` and `go-server.default` since it cannot really be overridden anymore

ketan:
> should get rid of DAEMON=Y from go-agent.default and go-server.default since it cannot really be overridden anymore

I want to leave an out for zip installers to be able to daemonize without using `service_mode` which is meant only for services

jyotisingh:
Then [this](https://github.com/gocd/gocd/pull/3879/files#diff-d223645ff4170c3d599616f9d988e240R18) line need to change. Currently it sets `DAEMON=N`, it should be `DAEMON=${DAEMON:-"N"}` may be.

ketan:
> Then this line need to change. Currently it sets DAEMON=N, it should be DAEMON=${DAEMON:-"N"} may be.

Yes, just realized that mistake just now. It's been fixed.

jar-class-loader/src/com/thoughtworks/gocd/Boot.java
```diff
@@ -43,9 +41,49 @@ private Boot(String... args) {
...
+        RuntimeMXBean bean = ManagementFactory.getRuntimeMXBean();
+        return bean.getInputArguments();
+    }
+
+    private static void log(String message) {
+        System.err.println("[" + new Date() + "] " + message);
+    }
+
+    private static String getOutFile() {
+        return System.getProperty("redirect.stdout.to.file");
```
jyotisingh: rename to `go.redirect.stdout.to.file`

PR #3884, opened 2017-09-22
Improve checksums for agent jar(s) download

ketan:
Before this commit the (md5) checksums for various jars are were encoded
in base64 and hex encoding:

    curl -s https://example.com/go/admin/latest-agent.status -I | grep MD5
    Agent-Content-MD5: xcOIJGRG6DVEndmh+ngL0w==
    Agent-Launcher-Content-MD5: YZZ/8vOLt9tme93YVwtoIg==
    Agent-Plugins-Content-MD5: 0fc9e52a2619dedfd7973959b8c92063
    TFS-SDK-Content-MD5: FDZDpIK84xCKx0EO+HZu9A==

This caused problems when checksums were verified because the agent was
always verifying passwords using base64, and the plugins zip was encoded
in hex format, causing the agent to re-download the plugins.zip every
time it was started up.

Additionally there was another bug with how the `agent-plugins.zip`
checksum was computed. The checksum was computed was basically the
equivalent of:
`cat [list of plugin.jar] | md5sum`.

Whereas it should have been:

`zip [list of plugin.jar] | md5sum`


ketan:
@gocd/committers should we move to sha1/sha256 instead of md5 for checksums?

arvindsv:
I think so. MD5 is not recommended for anything.

ketan:
I think moving to a different checksum breaks the protocol described in the initial comment, and will cause backward incompatibilities. I think a better bet is to introduce an internal api that will:

- list all jars (agent, launcher, tfs) along with some checksums (md5, sha1, sha256)
- list all plugin jars (that apply to agents) along with the same checksums

This should give the agents the ability to just download what is essential. Plugin jar updates will just require downloading (or deleting) a particular plugin, instead of all the plugins (as is currently the case)

We will obviously need to phase out compatibility with old agent-bootstrappers in 6 months to a year timeframe.

Issue #2947, opened 2016-12-06
Custom SCM material plugins do not show up in the list of materials when creating a new pipeline


alethenorio:
##### Issue Type

- Bug Report

##### Summary

When creating a new pipeline and selecting a material type, a custom written SCM plugin does  not show up in the list of material candidates

##### Environment

Gocd server running in a docker container

###### Basic environment details

* Go Version: `16.11`
* JAVA Version: `1.8`
* OS: `Docker container running Ubuntu`
* Browser vendor and version (if relevant): `Chrome/Firefox`

##### Steps to Reproduce

1. Create a plugin using the SCM extension
2. Create a new pipeline and follow the wizard until the part you have to create or select a materi
3. Click on the drop down of materials to select one
4. Notice that the custom written SCM plugin is nowhere to be found in the list

##### Expected Results

The plugin type should appear in the list below the other supported SCM materials (such as git, svn and so on) 

##### Actual Results

The material type is not present in the list and therefore cannot be selected

##### Any other info

The custom material does show up in the list when you try to add a new material to an existing pipeline so I think this is an issue with the UI not presenting in the pipeline creation section.

This means it is possible to work around the issue by creating a dummy material and then switching after the pipeline has been created but it makes for a very broken user experience

rogerioyuuki:
When I had this problem before, I found an old Github issue. I thought it was relevant to put the link here.

https://github.com/gocd/gocd/issues/976

kszatan:
I'm working on this.

mrmanc:
It looks like work on this issue in #3886 stalled, but this is still an issue and prevents you from creating a pipeline without supplying a dummy material which you then need to swap out once the pipeline is created. It would be great if anyone felt able to get #3886 merged in?

arvindsv:
@mrmanc As a part of #7920, the behaviour shown in https://github.com/gocd/gocd/pull/7920#issuecomment-608255470 will be available in the new pipeline wizard as well (@kritika-singh3 tells me). So, it will be consistent across the new and edit pipeline pages.

Note that it won't allow you to _create_ this material during the pipeline creation (or during pipeline edit). The [new pluggable SCMs page](https://github.com/gocd/gocd/pull/7919) coming in 20.3.0 will be the place to manage all of them and see where they are used, etc.

arvindsv:
@mrmanc Suggesting https://github.com/gocd/gocd/pull/7920#issuecomment-611368205 to see if we can make the process a little easier.

[...]

PR #3886, opened 2017-09-24
Show SCM materials when creating a new pipeline

kszatan:
This PR tries to address issue #2947.


rajiesh:
@kszatan , the failure of functional test case [here](https://build.gocd.org/go/tab/build/detail/acceptance-linux-PR/759/RunAcceptanceSpecs-priority/2/firefox-split-runInstance-4) is caused by an issue in the PR. The stage details page should not show both template and stage details.

Steps to reproduce,

1. Start the server with any SCM plugin included in its `<SERVER_WORKING_DIR>/plugins/external` folder
2. On Pipeline creation wizard fill in details and navigate to step 3 - Stage details
3. Could observer that the template and stage both shows up

Browser console log shows this error 

```
Error: Argument 'PluggableSCMConfigurationController' is not a function, got undefined
```

kszatan:
I have fixed this error but now it seems there's another problem. Could you please help me figure it out? https://build.gocd.org/go/tab/build/detail/acceptance-linux-PR/777/RunAcceptanceSpecs-plugins/1/firefox-split-runInstance-3

arvindsv:
@kszatan - What is happening with [this failure](https://build.gocd.org/go/files/acceptance-linux-PR/777/RunAcceptanceSpecs-plugins/1/firefox-split-runInstance-3/GaugeReports/reports/html-report/specs/CurlTaskPlugin.html) is that inline errors in plugins are getting lost.

To reproduce this, you need:

1. Download and install [this task plugin](https://github.com/gocd-contrib/task-skeleton-plugin/releases).
2. Create a new pipeline. Choose any material (even git material).
3. Choose "Curl" in the task dropdown.
4. Don't enter any URL.
5. Click Save.

Without this PR, you see this error:



With this PR, you see no error against that field.

Also, maybe more importantly, any other information you've entered into other fields gets lost. It looks like the responses from the plugin are not being bound back onto the form. Not in the new _pluggable_scm_form.html.erb, but in the older task one.

If you have some idea why, take a look at it. If not, let me know and I'll help. Thanks!

kszatan:
OK, so the problem is that PluggableSCMConfigurationModule and TaskPluginView are both Anuglar applications that try to attach to the same form. They do basically the same thing but one for pluggable materials and the other one for pluggable tasks. I'll have to merge those scripts or think of a way to make them work alongside themselves without errors.

arvindsv:
Sigh. This got missed. I saw your questions [on gitter](https://gitter.im/gocd/gocd?at=59f303caf7299e8f533b2353). Sorry about that, @kszatan. I'm going to spend some time thinking about what you said, tomorrow and reply to you.

kszatan:
No problem! I have an idea and I'm trying to make it work right now.

arvindsv:
Ok, take a look. The closest I remember to getting to this problem is with the task plugin. As you know, tasks can have on-cancel tasks. So, there can be two tasks, both plugin tasks, on the same page.

To get past this problem, I remember creating two `ng-app`s, so that there could be two separate Angular contexts on the same page. What is left of this code seems to be in [task_plugin_view.js](https://github.com/gocd/gocd/blob/3eebade2e7b833542c92ea0a027870aeb5399c14/server/webapp/WEB-INF/rails.new/app/assets/javascripts/task_plugin_view.js) and is used in many places, such as [this](https://github.com/gocd/gocd/blob/3eebade2e7b833542c92ea0a027870aeb5399c14/server/webapp/WEB-INF/rails.new/app/views/admin/tasks/pluggable_task/_form.html.erb#L20).

Would something like this work in this case?

kszatan:
> So, there can be two tasks, both plugin tasks, on the same page.
... and in the same form! That's why there's this whole GOINPUTNAME html rewriting.

> Would something like this work in this case?
The problem I'm facing is that TaskPluginView attaches on the closest form (it has to b/c of how Angular form validations work) and when you have another plugin template in the same form (for example handled by PluggableSCMConfigurationModule) they will clash. Fortunately, the code inside TaskPluginView does what needs to be done for every plugin template (changing input names and wiring the template data returned from server). I'm writing a general Angular app to handle all plugin templates in a form at once, automatically, instead of executing inline JS.

The biggest problem is that I'm new to Angular and really learned the tool to fix this issue. It's a bit of a struggle :)

arvindsv:
Oh, I can imagine! Thank you for doing this. I remember my struggle when I was doing this as well - and it's not "standard Angular" to make it do this.

It also doesn't help that it's inside ERBs. We want to rewrite that to get rid of ERBs, and use Mithril + JSX/MSX but it is not done yet.

Thanks again. I'll wait for you to try this approach - and if you need anything, let me know and I will try to respond quicker this time.

kszatan:
OK, the latest commit solves this problem!

kszatan:
@arvindsv It seems to be failing on some acceptance tests https://build.gocd.org/go/pipelines/value_stream_map/build-linux-PR/2305. I need a feedback.

arvindsv:
@kszatan I took a look - and it seems fine to me. It's strange that all the functional tests failed to start the server. It's obviously nothing to do with your change. Maybe something has changed/needs to change in the way the server is started in the functional tests. I'm looking into it.

arvindsv:
@kszatan What's happening is that this branch/PR needs to be rebased. Schema version of the config has been updated to 98 in the latest code, due to another PR. So, this one is failing to start the server in functional tests, because the tests have the config at a higher version.

I tried a rebase of this, and it seems simple. Just a couple of ruby spec files conflicted. If you rebase and push, I can trigger a new run. Thanks!

arvindsv:
:( Well ... I lost track of this and it languished here and ended up conflicting, as would be expected. Sorry, @kszatan. It was my fault. I am closing this with the assumption that this has been too long for you to have context about this.

If not, let me know and I'll try and fix conflicts, since I led to this happening. Thank you! Apologies again.

kszatan:
@arvindsv No problem. I got an impression that there's a new UI coming for GoCD and there's no interest in merging this PR. If that's not the case then I might have another crack at this when I find some time in the following weeks.

arvindsv:
@kszatan Makes sense. If you don't mind pinging me here or on Gitter or on email (any of my commit messages will have the address), before starting again, I can give you some information about the latest plans of the UI and what's going on.

At this point, there are no plans for a new UI for the wizard, but we worked on improving the UI around the dashboard, etc. as you might have seen. We are also moving away from Ruby / ERBs to single page applications. That might actually make this work (potentially) easier - since we can use the proper pipeline config API to create a pipeline, instead of messing around with the ruby controllers. One challenge might be to get the SCM plugin UI to show here. However, we are making some changes to the UI framework to create better components, etc. which might help.

Once you get back to me, I can provide some help. We can have a quick chat before you restart. Thanks again.

PR #3900, opened 2017-09-29
Fix TFS build command

arvindsv:
* This line seems to have got lost in a merge


arvindsv:
Merging. @rajiesh and I looked at this, and this is needed to make a functional test pass.

PR #3906, opened 2017-10-02
Fix bumpSchemaVersion task

arvindsv:
cruise-config.xsd should be copied over to numbered version, before the schemaVersion is incremented


bdpiparva:
LGTM

ketan:
> LGTM

Why's it not merged?

PR #3912, opened 2017-10-04
Health check API for Server.

varshavaradarajan:
GET `/go/api/v1/health`

Response: `{ "health": "OK" }`

This API is versioned in case we want to also show the server health messages and additional information in the future.


server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/health_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe 'show' do
+    describe 'security' do
+      it 'should allow anyone with security disabled' do
+        disable_security
+
+        expect(controller).to allow_action(:get, :show)
+      end
+
+      it 'should allow anyone with security disabled' do
```
GaneshSPatil: Change the test name?

PR #3918, opened 2017-10-05
Allow agent to re-registration only if agent is having auto registration key.

bdpiparva:
This PR make sure that agent can not re-register it self in absence of auto-register key.

### For a new agent
Server should register agent in pending state(in memory). When someone `approves` an agent then update the config.xml with agent information

#### Agent/Server restart 
* Agent already have the `certs` and `uuid` and it will not make an registration request to server. 
* What if agent is not having valid `certs` ?
  - In such case, SSL handshake between server and agent will fail and as a result agent will delete the invalid `certs` from disk. After removing `certs` it will make a registration request to server with following possibilities

    1)  Registration request with `uuid` and a valid `auto-register` key
        - In case of valid `auto-register` key server will process the registration request.
    2) Registration request with `uuid` *OR* Registration request with `uuid` and invalid `auto-register`
       - If autoregister key is provided or not valid then server won't allow such a agent and ignores the registration request with appropriate `error-message`

### DIFF
--- a/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
+++ b/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
@@ -54,7 +54,8 @@ import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
 
-import static org.apache.commons.lang.StringUtils.*;
+import static org.apache.commons.lang.StringUtils.isBlank;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 @Controller
 public class AgentRegistrationController {
@@ -194,16 +195,19 @@ public class AgentRegistrationController {
 
         try {
             final boolean shouldAutoRegisterAgent = goConfigService.serverConfig().shouldAutoRegisterAgentWith(agentAutoRegisterKey);
+            final boolean registeredAlready = goConfigService.hasAgent(uuid);
 
-            validateAgentRegistrationRequest(uuid, shouldAutoRegisterAgent, elasticAgentId, elasticPluginId);
+            validateAgentRegistrationRequest(uuid, shouldAutoRegisterAgent, registeredAlready, elasticAgentId, elasticPluginId);
 
             final String preferredHostname = getPreferredHostname(agentAutoRegisterHostname, hostname);
 
             AgentConfig agentConfig = new AgentConfig(uuid, preferredHostname, ipAddress);
-            agentConfig.setElasticAgentId(stripToNull(elasticAgentId));
-            agentConfig.setElasticPluginId(stripToNull(elasticPluginId));
+            if (isElasticAgent(elasticAgentId, elasticPluginId)) {
+                agentConfig.setElasticAgentId(elasticAgentId);
+                agentConfig.setElasticPluginId(elasticPluginId);
+            }
 
-            if (shouldAutoRegisterAgent) {
+            if (shouldAutoRegisterAgent && !registeredAlready) {
                 LOG.info("[Agent Auto Registration] Auto registering agent with uuid {} ", uuid);
                 GoConfigDao.CompositeConfigCommand compositeConfigCommand = new GoConfigDao.CompositeConfigCommand(
                         new AgentConfigService.AddAgentCommand(agentConfig),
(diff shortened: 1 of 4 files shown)


server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
```diff
@@ -192,36 +191,27 @@ public ModelAndView agentRequest(@RequestParam("hostname") String hostname,
...
-            if (goConfigService.serverConfig().shouldAutoRegisterAgentWith(agentAutoRegisterKey)) {
-                preferredHostname = getPreferredHostname(agentAutoRegisterHostname, hostname);
-                LOG.info("[Agent Auto Registration] Auto registering agent with uuid {} ", uuid);
-            } else {
-                if (elasticAgentAutoregistrationInfoPresent(elasticAgentId, elasticPluginId)) {
-                    throw new RuntimeException(String.format("Elastic agent registration requires an auto-register agent key to be setup on the server. Agent-id: [%s], Plugin-id: [%s]", elasticAgentId, elasticPluginId));
-                }
-            }
+            final boolean shouldAutoRegisterAgent = goConfigService.serverConfig().shouldAutoRegisterAgentWith(agentAutoRegisterKey);
+            final boolean registeredAlready = goConfigService.hasAgent(uuid);
```
jyotisingh: Ok, there is one slight problem with moving this code on top. 

If you look at the existing code, when auto-registration properties file is setup correctly for an agent, [that](https://github.com/gocd/gocd/blob/da55ed3cf9ce35641e96f4f18a8787fced3199bb/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L226) line will take care of saving the agent config to the file, so when the call to check if the agent was registered already happens [here](https://github.com/gocd/gocd/blob/da55ed3cf9ce35641e96f4f18a8787fced3199bb/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L238), it would return `true`. This causes `AgentRuntimeInfo.fromServer` to set the agent status to `Idle` which would then cause a successful registration of the agent in `agentService.requestRegistration`. 

With this PR, [that](https://github.com/bdpiparva/gocd/blob/d9c9006a9f9c2713aaf0e6cc2a9e315503df7c0b/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L197) line will return `false` for such an agent causing `agentService.requestRegistration` use a different flow. You might have to recheck for the value of the flag after the auto-registration but runs. 



 correctly, [that](https://github.com/bdpiparva/gocd/blob/d9c9006a9f9c2713aaf0e6cc2a9e315503df7c0b/server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java#L216) line will take care of saving the agent in the config, and hence the `registeredAlready` flag

server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
```diff
@@ -235,24 +224,43 @@ public ModelAndView agentRequest(@RequestParam("hostname") String hostname,
...
+            throw new RuntimeException(String.format("Elastic agent registration requires an auto-register key and it must match the auto-register key setup on the server. Agent-id: [%s], Plugin-id: [%s]", elasticAgentId, elasticPluginId));
+        }
+
+        if (partialElasticAgentInfo(elasticAgentId, elasticAgentId)) {
+            throw new RuntimeException("Elastic agents must submit both elasticAgentId and elasticPluginId");
+        }
+
+    }
+
+    private boolean isElasticAgent(String elasticAgentId, String elasticPluginId) {
```
ketan: Move this to `AgentConfig`?

server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
```diff
@@ -276,14 +284,10 @@ private void sendFile(InputStreamSrc input, HttpServletResponse response) throws
         }
     }
 
-    private boolean partialElasticAgentAutoregistrationInfo(String elasticAgentId, String elasticPluginId) {
+    private boolean partialElasticAgentInfo(String elasticAgentId, String elasticPluginId) {
```
ketan: Move this to `AgentConfig`?

bdpiparva: AgentConfig is already having `isElastic`. I will check and update it.

ketan:
I think we also need to stop scrubbing the autoregistration properties file.

jyotisingh:
@ketan - Why so? Just leaving the file behind doesn't give you anything more than what we already have. If the agent is already registered, the file doesn't even get used - doesn't matter if the contents are removed, commented out or left as is. 

Unless ofcourse you are trying to propose that we start using the contents of this file for updating the resources etc on the agent? If so, that is something we need to analyze a little more I think and that should be a separate story.

Issue #1841, opened 2016-01-26
Proposal: Improve Dashboard performance - rewrite


arvindsv:
Had to write this quickly, because @xli is working on [changing the dashboard](https://github.com/gocd/gocd/issues/1595#issuecomment-174402136), and I wanted to provide an option. :) JK. I don't expect this to be picked up right now. It is a proposal for a rewrite, albeit a clean one which can stay side-by-side with the current dashboard while it is in progress. But, it's good to have this out of my head and on digital-paper.

Broadly, the dashboard related code starts off [here](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/controllers/pipelines_controller.rb#L37), then [here](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/controllers/pipelines_controller.rb#L90-L94) and then [here](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/src/com/thoughtworks/go/server/service/PipelineHistoryService.java#L449). After the information about the pipelines is loaded, it goes through a set of ERB templates to create a [full dashboard page](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/views/pipelines/index.html.erb) and then a ["JSON"](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/views/pipelines/index.json.erb). I say JSON in quotes because it has HTML in it and is not really a JSON. Just in structure, not in spirit. Finally, there's a view cache layer [here](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/views/pipelines/_pipeline_group.html.erb#L11) and [here](https://github.com/gocd/gocd/blob/75f21bf7bebddc15fadf65651b709dbb4961ddbe/server/webapp/WEB-INF/rails.new/app/views/pipelines/index.json.erb#L9).

The current problems with this approach are:
[...]

PR #3920, opened 2017-10-05
#1841 - Quicker dashboard

maheshp:
This PR is the first step towards improving the pipeline dashboard performance by implementing a new dashboard API which is derived from the existing CCTary implementation. For more details refer the [proposal](https://github.com/gocd/gocd/issues/1841). 

This is a first cut of the dashboard API which will be tweaked further based on the requirements.


jyotisingh:
https://github.com/gocd/gocd/issues/1841

arvindsv:
> There are a few outstanding issues with the API such as it doesn't return any entries when invoked until a build happens.

Really? Not even historical information? That's ... strange. Do you mean it doesn't return any entries for a newly added pipeline? Or is completely empty till one build happens after a restart?

Issue #3928, opened 2017-10-10
Missing icons on plugins page


ketan:
##### Issue Type

- Bug Report

##### Summary

The new admin page seems to directly refer to `../assets/andare/icon-plugin-default.png`, this file is renamed to a unique file name after the asset compilation step. We need to likely use a module loader for webpack to allow it to load the images properly.

<img width="767" alt="screen shot 2017-10-10 at 6 39 37 am" src="https://user-images.githubusercontent.com/10598/31369688-6e63bb28-ad86-11e7-80ad-0a4cc51ecd17.png">

PR #3932, opened 2017-10-10
Plugin SPA Fixes 

varshavaradarajan:


server/webapp/WEB-INF/rails.new/app/views/admin/plugins/index.html.erb
```diff
@@ -1,5 +1,7 @@
 <div id="plugins"
-     is-user_an_admin="<%= @is_user_an_admin %>"
+     is-user-an-admin="<%= @is_user_an_admin %>"
+     unknown-plugin-icon-bundled="<%= image_path("andare/icon-plugin-default-bundled.png") %>"
```
ketan: Please no! — https://www.npmjs.com/package/file-loader

varshavaradarajan:
About the plugin icons on the plugin SPA page (`/go/admin/new_plugins`), there are certain scenarios to be considered before showing the right image.

Supposing a plugin is bundled and provides an image, we use the plugin's image for the icon. There may not be any indication from the icon as to whether the plugin is bundled or not. 
If the plugin does not provide an image, we default to an image which is indicative of whether the plugin's bundled. Because of this inconsistency, is it okay to not rely on the plugin icon to decide if the plugin is bundled, but provide another indication for in on the SPA? 

Currently, defaulting to the original unknown plugin image that we had. 

cc @naveenbhaskar

bdpiparva:
@varshavaradarajan Can we close this https://github.com/gocd/gocd/issues/3928 ?

PR #3933, opened 2017-10-11
Fix config repo material representer

GaneshSPatil:
* Remove attributes from config repo materials which are not required. 
* Modify XML schema definition to not support unnecessary attributes.
* Add an XSL Translation to wipe out unnecessary attributes from the materials tag.

#### Git
* `url`
* `branch`
* `materialName`

#### hg
* `url`
* `materialName`

#### tfs
* `url`
* `username`
* `domain`
* `password`
* `encryptedPassword`
* `projectPath`
* `materialName`

#### p4
* `view`
* `port`
* `username`
* `password`
* `encryptedPassword`
* `materialName`

#### svn
* `url`
* `username`
* `password`
* `encryptedPassword`
* `materialName`

### DIFF
--- a/server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
+++ b/server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
@@ -19,9 +19,11 @@ package com.thoughtworks.go.config;
 import com.thoughtworks.go.config.elastic.ElasticProfile;
 import com.thoughtworks.go.config.elastic.ElasticProfiles;
 import com.thoughtworks.go.config.materials.dependency.DependencyMaterialConfig;
+import com.thoughtworks.go.config.materials.git.GitMaterialConfig;
 import com.thoughtworks.go.config.materials.mercurial.HgMaterialConfig;
 import com.thoughtworks.go.config.pluggabletask.PluggableTask;
 import com.thoughtworks.go.config.registry.ConfigElementImplementationRegistry;
+import com.thoughtworks.go.config.remote.ConfigRepoConfig;
 import com.thoughtworks.go.config.validation.GoConfigValidity;
 import com.thoughtworks.go.domain.GoConfigRevision;
 import com.thoughtworks.go.domain.Task;
@@ -40,13 +42,17 @@ import com.thoughtworks.go.util.*;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
+import org.hamcrest.core.IsEqual;
 import org.jdom2.Document;
 import org.jdom2.filter.ElementFilter;
 import org.jdom2.input.SAXBuilder;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import org.mockito.Matchers;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
@@ -73,6 +79,8 @@ import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.Matchers.*;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsNot.not;
 import static org.hamcrest.core.IsNull.nullValue;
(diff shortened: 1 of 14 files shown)


server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/config_repo/materials/git_material_representer.rb
```diff
@@ -27,8 +27,6 @@ class GitMaterialRepresenter < ScmMaterialRepresenter
                    setter: lambda { |value, options|
                      value.blank? ? self.setBranch('master') : self.setBranch(value)
                    }
-          property :submodule_folder
-          property :shallow_clone
```
ketan: That looks like breaking compatibility?

GaneshSPatil: @ketan -- firstly, these fields were never documented.
Even if these fields are specified, those were not in use for `config repo` material! So removing these fields won't lead to broken functionality.

We've added an `xsl` migration to wipe out unnecessary fields from the material tag.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1450,6 +1450,24 @@ public void shouldConvertPluginToPluginIdOnConfigRepoAsPartOfMigration95() throw
...
+                "<config-repos>\n" +
+                "   <config-repo pluginId=\"json.config.plugin\" id=\"config-repo-1\">\n" +
+                "     <git url=\"https://github.com/tomzo/gocd-json-config-example.git\" dest=\"dest\"/>\n" +
+                "   </config-repo>\n" +
+                "</config-repos>" +
+                "</cruise>";
+
+        try {
+            migrateXmlString(configXml, 97);
+        } catch (InvocationTargetException e) {
```
maheshp: Consider using `@Rule` for this? The test does not verify that an exception is thrown.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1450,6 +1460,45 @@ public void shouldConvertPluginToPluginIdOnConfigRepoAsPartOfMigration95() throw
...
+                "<config-repos>\n" +
+                "   <config-repo pluginId=\"json.config.plugin\" id=\"config-repo-1\">\n" +
+                "     <git url=\"https://github.com/tomzo/gocd-json-config-example.git\" dest=\"dest\"/>\n" +
+                "   </config-repo>\n" +
+                "</config-repos>" +
+                "</cruise>";
+
+        try {
+            migrateXmlString(configXml, 98);
+        } catch (InvocationTargetException e) {
```
maheshp: change this to use @Rule. The test would pass if an exception is not thrown

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1450,6 +1460,45 @@ public void shouldConvertPluginToPluginIdOnConfigRepoAsPartOfMigration95() throw
...
+            assertThat(e.getTargetException().getCause().getMessage(), is("Attribute 'dest' is not allowed to appear in element 'git'."));
+        }
+    }
+
+    @Test
+    public void shouldRemoveUncesseryMaterialFieldsAsPartOfMigration98() throws Exception {
+        String configXml = "<cruise schemaVersion='96'>" +
```
maheshp: Add tests for migration of other materials as well

config/config-server/resources/cruise-config.xsd
```diff
@@ -143,19 +143,52 @@
     <xsd:complexType name="configRepoType">
         <xsd:sequence>
             <xsd:choice minOccurs="1" maxOccurs="1">
-                <xsd:element name="svn" type="svnType"/>
-                <xsd:element name="hg" type="hgType"/>
-                <xsd:element name="p4" type="p4Type"/>
-                <xsd:element name="git" type="gitType"/>
-                <xsd:element name="tfs" type="tfsType"/>
-                <xsd:element name="scm" type="scmReferenceType"/>
```
maheshp: why is `scmReferenceType` removed, it should be supported through the API as well.

config/config-server/resources/cruise-config.xsd
```diff
@@ -143,19 +143,56 @@
...
+                <xsd:element name="tfs" type="configRepoTfsType"/>
+                <xsd:element name="scm" type="configRepoScmReferenceType"/>
             </xsd:choice>
             <xsd:element name="configuration" minOccurs="0" maxOccurs="1" type="configurationType"/>
         </xsd:sequence>
         <xsd:attribute type="nameType" name="id" use="required"/>
         <xsd:attribute type="nameType" name="pluginId" use="optional"/>
     </xsd:complexType>
-
+    <xsd:complexType name="configRepoSvnType">
```
maheshp: we need to support `checkExternals`

config/config-server/resources/cruise-config.xsd
```diff
@@ -143,19 +143,56 @@
...
+        <xsd:attribute name="url" type="xsd:string" use="required"/>
+        <xsd:attribute name="username" type="xsd:string"/>
+        <xsd:attribute name="password" type="xsd:string"/>
+        <xsd:attribute name="encryptedPassword" type="xsd:string"/>
+        <xsd:attribute name="materialName" type="nameType" use="optional"/>
+    </xsd:complexType>
+    <xsd:complexType name="configRepoP4Type">
```
maheshp: add support of `useTickets`

config/config-server/resources/upgrades/98.xsl
```diff
@@ -0,0 +1,67 @@
...
+    <!-- Remove following attributes from config repo git material -->
+    <xsl:template match="/cruise/config-repos/config-repo/git/filter" />
+    <xsl:template match="/cruise/config-repos/config-repo/git/@dest" />
+    <xsl:template match="/cruise/config-repos/config-repo/git/@shallowClone" />
+    <xsl:template match="/cruise/config-repos/config-repo/git/@autoUpdate" />
+    <xsl:template match="/cruise/config-repos/config-repo/git/@invertFilter" />
+
+    <!-- Remove following attributes from config repo svn material -->
+    <xsl:template match="/cruise/config-repos/config-repo/svn/filter" />
+    <xsl:template match="/cruise/config-repos/config-repo/svn/@checkexternals" />
```
maheshp: should support `checkExternals`

config/config-server/resources/upgrades/98.xsl
```diff
@@ -0,0 +1,67 @@
...
+    <xsl:template match="/cruise/config-repos/config-repo/svn/@dest" />
+    <xsl:template match="/cruise/config-repos/config-repo/svn/@autoUpdate" />
+    <xsl:template match="/cruise/config-repos/config-repo/svn/@invertFilter" />
+
+    <!-- Remove following attributes from config repo p4 material -->
+    <xsl:template match="/cruise/config-repos/config-repo/p4/filter" />
+    <xsl:template match="/cruise/config-repos/config-repo/p4/@useTickets" />
```
maheshp: should support `useTickets`

config/config-server/resources/cruise-config.xsd
```diff
@@ -143,19 +143,58 @@
...
             </xsd:choice>
             <xsd:element name="configuration" minOccurs="0" maxOccurs="1" type="configurationType"/>
         </xsd:sequence>
         <xsd:attribute type="nameType" name="id" use="required"/>
         <xsd:attribute type="nameType" name="pluginId" use="optional"/>
     </xsd:complexType>
-
+    <xsd:complexType name="configRepoSvnType">
+        <xsd:attribute name="url" type="xsd:string" use="required"/>
+        <xsd:attribute name="username" type="xsd:string"/>
```
jyotisingh: `autoUpdate` has been removed from the configRepo material types. We need to allow users to be able to configure this for multiple reasons, one being that they might want to turn off autoUpdate and setup post-commit hooks for such materials. That should be allow. Another being if the fingerprint matches that of a material which is part of a pipeline, Go server wouldn't allow two different values for this flag and cause the same to fail.

GaneshSPatil: @jyotisingh -- I guess, `auto_update` for config repo materials is always set to true.
Not sure though why it should be always `true`.

https://github.com/gocd/gocd/blob/master/config/config-api/src/com/thoughtworks/go/config/remote/ConfigRepoConfig.java#L177-L182

config/config-server/resources/cruise-config.xsd
```diff
@@ -143,19 +143,58 @@
...
+        <xsd:attribute name="port" type="xsd:string" use="required"/>
+        <xsd:attribute name="username" type="xsd:string"/>
+        <xsd:attribute name="password" type="xsd:string"/>
+        <xsd:attribute name="encryptedPassword" type="xsd:string"/>
+        <xsd:attribute name="useTickets" type="xsd:boolean" use="optional"/>
+        <xsd:attribute name="materialName" type="nameType" use="optional"/>
+    </xsd:complexType>
+    <xsd:complexType name="configRepoGitType">
+        <xsd:attribute name="url" type="xsd:string" use="required"/>
+        <xsd:attribute name="branch" type="xsd:string" use="optional"/>
```
jyotisingh: any specific reason for getting rid of `shallowClone`? I think it should be available.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1450,6 +1455,187 @@ public void shouldConvertPluginToPluginIdOnConfigRepoAsPartOfMigration95() throw
...
+                "   </config-repo>\n" +
+                "</config-repos>" +
+                "</cruise>";
+
+        assertThat(configXml, containsString("dest=\"dest\""));
+        assertThat(configXml, containsString("<filter>"));
+        assertThat(configXml, containsString("autoUpdate='false'"));
+        assertThat(configXml, containsString("invertFilter='true'"));
+
+        String migratedContent = migrateXmlString(configXml, 97);
```
jyotisingh: I guess you need to update all the tests to use schemaVersion 99 or whatever is available when you next rebase, and trigger the PR pipeline. This test must be failing currently since the version that migrates these nodes in your PR is 98 not 97 :)

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/config/config_repo/materials/git_material_representer.rb
jyotisingh: `submodule_folder` was a wrongly added attribute, `shallow_clone` needs to come back.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1450,6 +1455,187 @@ public void shouldConvertPluginToPluginIdOnConfigRepoAsPartOfMigration95() throw
...
+                "</cruise>";
+
+        assertThat(configXml, containsString("<filter>"));
+        assertThat(configXml, containsString("dest='dest'"));
+        assertThat(configXml, containsString("autoUpdate='false'"));
+        assertThat(configXml, containsString("invertFilter='true'"));
+
+        String migratedContent = migrateXmlString(configXml, 97);
+
+        assertThat(migratedContent, not(containsString("<filter>")));
```
jyotisingh: May be you could convert the migratedContext to Config object using `GoConfigHolder holder = loader.loadConfigHolder(migratedContent);` and then assert that the config-repo object has the right attributes set, say autoUpdate etc.

jyotisingh:
rebase - since version 98 is already merged with master.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
GaneshSPatil: @jyotisingh -- we also need to check whether the attributes are removed from the config.
Will keep the existing assertions and will add new assertions for `null` values for the `attributes` which are not supported on `config-repo` material

PR #3940, opened 2017-10-16
Disallow a authenticated user from re-authentication

maheshp:
* Before this commit a authenticated user can open the login page and
  re-authenticate, this behviour differs between various auth mechanism
  used.
* With this commit a authenticated user is re-directed to home page
  on trying to access the login page or posting the authentication credentials.

### DIFF
--- a/server/src/com/thoughtworks/go/server/security/DisallowExternalReAuthenticationFilter.java
+++ b/server/src/com/thoughtworks/go/server/security/DisallowExternalReAuthenticationFilter.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2017 ThoughtWorks, Inc.
+ *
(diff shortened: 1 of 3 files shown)


maheshp:
Post these changes, the existing session takes precedence over a new session.  

E.g scenario:
1. A unauthenticated user opens the login page in two tabs.  
2. The user authenticates to GoCD from one tab as `foo`. 
3. On the second tab with login page open the user authenticates as `bar`.
The second authentication request would be ignored and the user would be redirected to home page with the session of `foo`

arvindsv:
What happens if that second authentication request fails? I mean, wrong password for `bar`.

maheshp:
The second auth request is ignored and hence an authentication attempt is not made. 

For a user, it might look like authentication going through with invalid credentials.

PR #3952, opened 2017-10-25
update config repo plugins, json to 0.2.1, yaml to 0.6.0

tomzo:


varshavaradarajan:
@tomzo - We decided to bump up the bundled plugins version for the next release (17.12). We are close to cutting the 17.11 release and hence there is a merge freeze till that happens. Thanks for the PR. I have verified that the  yaml and json plugin works as expected too. We realize that the bundled yaml plugin will not support the template ref and parameters until the next GoCD release. But if people wish to, I suppose they can delete the plugin in the bundled folder and add the new yaml plugin in the external folder. :)

PR #3960, opened 2017-10-27
Adding pipeline count information to the about page

jyotisingh:


server/test/unit/com/thoughtworks/go/server/service/PipelineConfigServiceTest.java
```diff
@@ -180,4 +183,19 @@ public void createPipelineConfigShouldValidateAllPluggableTasks() {
...
         verify(pluggableTaskService).isValid(docker);
     }
+
+    @Test
+    public void shouldGetPipelinesCount() {
+        assertThat(pipelineConfigService.totalPipelinesCount(), is(this.cruiseConfig.allPipelines().size()));
+    }
+
+    @Test
+    public void pipelineCountShouldIncludeConfigRepoPipelinesAsWell() {
```
varshavaradarajan: Will the number of pipelines shown in `/go/api/support` be consistent with this number? Meaning, does `/go/api/support` also include config repo pipelines?

jyotisingh: yes it does

Issue #3854, opened 2017-09-15
Get rid of intermediate config xmls stored while config migration


varshavaradarajan:
##### Issue

While introducing the config migrations for auth config, we left behind `config-before-migration-91.xml` and `config-before-migration-92.xml` in case users had to rollback. Now that there haven't been any reports of failing migrations, we can remove these files.

ketan:
I've asked @bdpiparva to address this.

bdpiparva:
Yes, I will do this as part of removing password file support from core.

rajiesh:
Verified on 17.12.0 (5605-89a5ade1e9497ffc255dfa744476466c09beca06)

PR #3961, opened 2017-10-27
Get rid of intermediate config xmls stored while config migration

bdpiparva:
Removed migration backup code. https://github.com/gocd/gocd/issues/3854


maheshp:
The changes LGTM, do you also want to remove references to `systemEnvironment` since its no longer used.

PR #3963, opened 2017-10-30
Template Authorization API

varshavaradarajan:


server/src/main/java/com/thoughtworks/go/config/update/UpdateTemplateAuthConfigCommand.java
```diff
@@ -0,0 +1,40 @@
...
+    public UpdateTemplateAuthConfigCommand(PipelineTemplateConfig templateConfig, Username currentUser, SecurityService securityService, LocalizedOperationResult result, String md5, EntityHashingService entityHashingService) {
+        super(templateConfig, currentUser, securityService, result, md5, entityHashingService);
+    }
+
+    @Override
+    public void update(CruiseConfig modifiedConfig) throws Exception {
+        PipelineTemplateConfig existingTemplateConfig = findAddedTemplate(modifiedConfig);
+        TemplatesConfig templatesConfig = modifiedConfig.getTemplates();
+        existingTemplateConfig.setAuthorization(templateConfig.getAuthorization());
+        modifiedConfig.setTemplates(templatesConfig);
```
maheshp: I couldn't understand the need to get templates config and set it back again, is this required?

varshavaradarajan: No, templates need not be set. I'll change it.

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/authorization/templates_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+        def update
+          result = HttpLocalizedOperationResult.new
+          authorization = ApiV1::Admin::Authorization::AuthorizationConfigRepresenter.new(com.thoughtworks.go.config.Authorization.new).from_hash(params[:template])
+          updated_template = CLONER.deepClone(@template)
+          updated_template.setAuthorization(authorization)
+          template_config_service.updateTemplateAuthConfig(current_user, updated_template, result, etag_for(@template))
+          handle_create_or_update_response(result, updated_template)
+        end
+
+        def load_template(template_name = params[:template_name])
```
maheshp: Move these methods to private?

server/webapp/WEB-INF/rails.new/app/controllers/api_v1/admin/authorization/templates_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+        def show
+          json = ApiV1::Admin::Authorization::AuthorizationConfigRepresenter.new(@template.getAuthorization).to_hash(url_builder: self)
+          render DEFAULT_FORMAT => json if stale?(etag: etag_for(@template))
+        end
+
+        def update
+          result = HttpLocalizedOperationResult.new
+          authorization = ApiV1::Admin::Authorization::AuthorizationConfigRepresenter.new(com.thoughtworks.go.config.Authorization.new).from_hash(params[:template])
+          updated_template = CLONER.deepClone(@template)
```
maheshp: How bout getting rid of the template cloning? 
change `template_config_service.updateTemplateAuthConfig(current_user, template_name, auth_config, result, etag)'
and UpdateTemplateAuthConfigCommand update can update the auth_config as part of the update method.

What do you think?

maheshp:
Listing out few things discussed with @varshavaradarajan 

1. Since authorization is a nested resource of templates we are considering changing the routes to `/go/api/admin/templates/<template_id>/authorization`.
2. Change templates_controller to authorization_controller, namespace within templates.
3. templates authorization API will expose only GET and PUT. The idea is when a template is created it would have an empty authorization, so creating authorization is not required but can be updated.

server/webapp/WEB-INF/rails.new/config/routes.rb
```diff
@@ -241,6 +241,12 @@
           resources :auth_configs, param: :auth_config_id, except: [:new, :edit,], constraints: {auth_config_id: ALLOW_DOTS}
           resources :roles, param: :role_name, except: [:new, :edit], constraints: {role_name: ROLE_NAME_FORMAT}
         end
+
+        namespace :templates do
+          get ':template_name/authorization' => 'authorization#show', constraints: {template_name: ALLOW_DOTS}
```
maheshp: Use `TEMPLATE_NAME_FORMAT` for template_name contraints.

server/webapp/WEB-INF/rails.new/app/presenters/api_v1/admin/authorization/authorization_config_representer.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        property :isAllowGroupAdmins, as: :all_group_admins_are_view_users,
+                 setter: lambda {|val, options| self.setAllowGroupAdmins(val)}
+
+        property :adminsConfig, as: :admin,
+                 skip_render: lambda { |object, options| object.empty? },
+                 decorator: ApiV1::Admin::Authorization::UserRoleConfigRepresenter,
+                 class: com.thoughtworks.go.config.AdminsConfig
+
+        property :operationConfig, as: :operate,
```
maheshp: Should we expose this since templates authorisation do not support `operate`?

varshavaradarajan:
PR build went through -  https://build.gocd.org/go/pipelines/value_stream_map/trigger/2238 (regression SPA was a flaky one). Just rebased it now. 

@maheshp - we discussed about why the authorization validate just returns without performi. It because it will end up doing the same validation twice when using the GoConfigGraphWalker. Its the same reason why I removed `validateStageConfig` from `PipelineTemplateConfig#validate`.

varshavaradarajan:
Reason why DelegatingValidationContext was introduced

Consider validation of the roles configured in template authorization and template stage config approval with ConfigSaveValidationContext. 

Roles configured in template authorization should be validated to check if they exist. 
Roles configured in **template** stage config should **not** be validated as this could be a parameter.

Given the above 2 restrictions, using just the `ConfigSaveValidationContext#isWithTemplates()` is not enough. This method would have returned true for template authorization as well and skipped validation of template authorization. Therefore, we introduced the DelegatingValidationContext. 

Here's the PR that  introduced it - https://github.com/gocd/gocd/pull/3035

PR #3967, opened 2017-10-30
Log the size of the config repository before and after jgit gc.

varshavaradarajan:


varshavaradarajan:
@jyotisingh - For a 2G directory, takes 14ms to calculate the size. You still want me to add a system property so that calculating the size of the config repo won't affect performance?

ketan:
> @jyotisingh - For a 2G directory, takes 14ms to calculate the size. You still want me to add a system property so that calculating the size of the config repo won't affect performance?

How frequently do we check this?

varshavaradarajan:
@ketan - that depends on the property `go.config.repo.gc.cron`

ketan:
Let me rephrase that question:

What interval (min and max) do you expect users to run this?

If it's very infrequent (a few times a day), it's probably OK to take a few seconds to compute the size. If the GC is very frequent. If the cron is running a few times an hour, it's probably not OK to take a few hours to compute the directory sizes.

varshavaradarajan:
@ketan ,
Running GC few times in an hour results in bloating up the config repo (as we saw when we had set the cron to run every 5 minutes). We'll hit other perf problems by that time.

varshavaradarajan:
@jyotisingh can better answer the question about the min and max number of times people can run gc.

ketan:
@jyotisingh can we hear your thoughts? My concern is that paying a high cost of computing dir size is not much if we're not doing it "too often", can you clarify what's the frequency that we should be doing this check?

jyotisingh:
Sorry for the delay in response on this one.
Ok, so the background is in some cases we happened to notice that a GC happened but it did not reduce the size of the config.git folder. @varshavaradarajan mentioned about some instance in past wherein the config.git started bloating up on build.gocd.org. The intent was to put something in place that could report the directory size along with the rest of the repo statistics before and after the Jgit GC. 

I guess the simplest thing to do would be to do the following:

* Have the following log statements:
```
LOGGER.info("Before GC: {}", git.gc().getStatistics());
LOGGER.debug("Before GC: Size: {}", getConfigRepoDisplaySize());
```

* Add this to the logback.xml of build.gocd.org so that debug logging is enabled only for this specific class:
```
  <logger name="com.thoughtworks.go.service.ConfigRepository" level="DEBUG" />
```

That would help us get the required details from our server, we could do the same for users who complain about the directory size not reducing etc. This point of the exercise is to gather evidence. We would need to make fixes if we spot an issue after we have the logs.

@ketan - to answer your question about frequency - the default is set to once in a week, but we know of users who have configured it to run more frequently than that - say couple of times in a day.

PR #3971, opened 2017-10-31
Added missing lincense header

bdpiparva:
Added a check task under `gocd` project to check license header on source files.


ketan:
maybe use a gradle plugin that can add licenses?

ketan:
https://github.com/hierynomus/license-gradle-plugin seems to be the only one maintained. Need to verify if it works with java, js, jsx, ruby, sql and couple of others we use.

Issue #1792, opened 2016-01-12
Cannot use parameters in pipeline label templates if the parameter value contains a space


ivanmoore:
![image](https://cloud.githubusercontent.com/assets/410457/12264685/54f09ebe-b930-11e5-8827-8aa659ae4e1b.png)

When setting the pipeline label template, the tooltip (and documentation) suggests that you can refer to a parameter via the `#{NAME}` syntax.

When setting the pipleline label template, attempting to refer to a parameter where the parameter value contains a space results in the error shown above being displayed. The parameter `bob` in this case is set, as shown in the screenshot below.

![image](https://cloud.githubusercontent.com/assets/410457/12264729/9956883e-b930-11e5-8d81-aa36af83b874.png)

The error message is misleading, it is unclear that that the label cannot contain a space and it is additionally unclear that the parameter is responsible for raising the error.

ivanmoore:
Additionally when setting the value of a parameter used in a pipeline label template, there is no error message displayed when the new value contains a space, as shown below:

![image](https://cloud.githubusercontent.com/assets/410457/12265029/8ba99242-b932-11e5-9abc-d36c35b1ff53.png)

akshaydewan:
I'm thinking of possible fixes for this issue.

We can't restrict the user from entering spaces into the Parameter Value. As per the info and example in the [GoCD docs](https://docs.gocd.org/current/configuration/admin_use_parameters_in_configuration.html), spaces should be allowed here.

On the 'General Options' tab, we can improve the error message of the label template. We could show the evaluated template in the error message. So for parameter value "foo bar" it would read: `Invalid label '${COUNT}-foo bar'. Label should be composed of alphanumeric text, it should contain the builder number as ${COUNT}, can contain a material revision as ${<material-name>} of ${<material-name>[:<number>]}, or use params as #{<param-name>}.`

On the 'Parameters' tab, if the user tries to change to an invalid value after the parameter is already used in a Label Template, the Save operation fails with a cryptic error message "Save failed, see errors below". I'm not sure where on the UI we can show a better error message in this case.

Thoughts?

arvindsv:
@akshaydewan I agree with the change in message.

For the second part, I think it's a larger problem - of error messages which are lost because the "key" of the error, is not found in the page. In this case, the "key" will be something like "pipeline-label". We should find out which errors are not bound and show them next to the "Save failed" message itself, I feel.

PR #3988, opened 2017-11-07
Show evaluated value in the error message for label template (#1792)

akshaydewan:


server/webapp/WEB-INF/rails.new/spec/controllers/api_v3/admin/pipelines_controller_spec.rb
```diff
@@ -669,7 +669,7 @@
     def expected_data_with_validation_errors
       {
         enable_pipeline_locking: false,
-        errors: {label_template: ["Invalid label. Label should be composed of alphanumeric text, it should contain the builder number as ${COUNT}, can contain a material revision as ${<material-name>} of ${<material-name>[:<number>]}, or use params as \#{<param-name>}."]},
+        errors: {label_template: ["Invalid label '%s'. Label should be composed of alphanumeric text, it should contain the builder number as ${COUNT}, can contain a material revision as ${<material-name>} of ${<material-name>[:<number>]}, or use params as \#{<param-name>}."]},
```
varshavaradarajan: Should be the actual string at this point, right? Also, can you please change the original error message from `it should contain the build number ...` to `it can contain the build number ...` ?

Issue #3773, opened 2017-08-22
When console logs are not available, no message is shown


arvindsv:
##### Issue Type

- Bug Report

##### Summary

Console logs of jobs with purged logs don’t show any message. Compare [this](https://build.gocd.org/go/tab/build/detail/build-linux/1832/build-non-server/1/agent) and its [raw logs](https://build.gocd.org/go/files/build-linux/1832/build-non-server/1/agent/cruise-output/console.log).

###### Basic environment details

* Go Version: 17.8

##### Expected Results

A message which says that the logs are not available, because they were removed.

##### Actual Results

Empty. No message.

##### Any other info

Related: #3650 #3695

arvindsv:
Added 17.9 as milestone, just so we don't lose it. Not that urgent. Should be in the next one, at least, though.

GaneshSPatil:
Reproducible on both
  - [ ] Console log Polling
  - [ ] Console log Websocket

rajiesh:
Verified on 17.12.0 (5621-158fee2729f0a07475594782e6a5716ac3154f8a)

PR #3998, opened 2017-11-09
Fix waiting for console log bug

GaneshSPatil:


GaneshSPatil:
This PR addresses issue - #3650 and #3773

server/src/com/thoughtworks/go/server/controller/ArtifactsController.java
```diff
@@ -63,22 +63,28 @@
 public class ArtifactsController {
     private static final Logger LOGGER = LoggerFactory.getLogger(ArtifactsController.class);
 
+    private final ZipArtifactCache zipArtifactCache;
```
varshavaradarajan: Not necessary to be a field.

server/src/com/thoughtworks/go/server/controller/ArtifactsController.java
```diff
@@ -63,22 +63,28 @@
...
+    private final JobInstanceDao jobInstanceDao;
     private final ConsoleActivityMonitor consoleActivityMonitor;
     private final ArtifactFolderViewFactory folderViewFactory;
     private final ArtifactFolderViewFactory jsonViewFactory;
     private final ArtifactFolderViewFactory zipViewFactory;
     private ArtifactsService artifactsService;
     private RestfulService restfulService;
     private ConsoleService consoleService;
+    private final SystemEnvironment systemEnvironment;
```
varshavaradarajan: Not necessary to be a field.

server/src/com/thoughtworks/go/server/dao/JobInstanceSqlMapDao.java
```diff
@@ -272,6 +272,11 @@ public int totalCompletedJobsOnAgent(String uuid) {
         return (Integer) getSqlMapClientTemplate().queryForObject("totalCompletedJobsOnAgent", arguments("uuid", uuid).asMap());
     }
 
+    @Override
+    public boolean isJobCompleted(JobIdentifier jobIdentifier) {
+        return mostRecentJobWithTransitions(jobIdentifier).isCompleted();
```
varshavaradarajan: I think we can use the latestCompletedJobs, restrict the count to 1 instead of querying for transitions. Since we have the job identifier, we can get the pipeline name, stage name, job name that's required for the latestCompletedJobs method? What do you think?

server/src/com/thoughtworks/go/server/service/ConsoleService.java
```diff
@@ -43,16 +44,18 @@
     private ArtifactDirectoryChooser chooser;
     public static final int DEFAULT_CONSOLE_LOG_LINE_BUFFER_SIZE = 1024;
     private ArtifactsDirHolder artifactsDirHolder;
+    private JobInstanceDao jobInstanceDao;
```
varshavaradarajan: This is unused

server/src/com/thoughtworks/go/server/service/ConsoleService.java
```diff
@@ -69,6 +72,14 @@ public File consoleLogArtifact(LocatableEntity jobIdentifier) throws IllegalArti
         return chooser.findArtifact(jobIdentifier, getConsoleOutputFolderAndFileName());
     }
 
+    public boolean doesLogExists(JobIdentifier jobIdentifier) {
```
varshavaradarajan: doesLogExists => doesLogExist

server/src/com/thoughtworks/go/server/websocket/ConsoleLogSender.java
```diff
@@ -113,11 +113,7 @@ public void process(final SocketEndpoint webSocket, JobIdentifier jobIdentifier,
     }
 
     private boolean doesLogExists(JobIdentifier jobIdentifier) {
-        try {
-            return consoleService.consoleLogFile(jobIdentifier).exists();
-        } catch (IllegalArtifactLocationException e) {
-            return false;
-        }
+        return consoleService.doesLogExists(jobIdentifier);
```
varshavaradarajan: Replace all occurrences of `consoleService.consoleLogFile(jobIdentifier).exists()` with something like above.

server/src/com/thoughtworks/go/server/dao/JobInstanceSqlMapDao.java
GaneshSPatil: @varshavaradarajan -- `latestCompletedJobs` for the given identifier can't be used.
Let's say for a job identifier, build 1 passed and build 2 is in progress, in such scenarios,

`latestCompletedJobs` will return `build1`, whereas,
`mostRecentJobWithTransitions` will return `build2`

rajiesh:
Should we also not consider the scenario of `waiting for console logs...` message being shown when the job is yet to be assigned. For a new user this message for not yet assigned job will be confusing. May be we should consider showing up `waiting for assignment...`

Issue #4007, opened 2017-11-13
VSM is messed up


VeeTeeDev:
Generation of the VSM is sometimes messed up.

I've tested this on:
* Internet Explorer 11
* Edge
* Chrome
* Firefox dev edition

It's only for some specific elements. Others are getting visualised just right.

https://gist.github.com/VeeTeeDev/13fe767a5cab623f71d4161db13ef32b

![bug2](https://user-images.githubusercontent.com/9404890/32726997-472b912e-c87b-11e7-9f1b-51927f76693f.png)

In some cases the element is getting displayed on top of the header of the page.

![bug4](https://user-images.githubusercontent.com/9404890/32727255-398eefd8-c87c-11e7-9c5e-2a531962b24f.PNG)

It's not because of the warning that header-e2e did not run yet. Cause this is a screenshot from a rebuild of the setup. It was showing the same problems before.

This was tested on 17.9 and 17.11 but I only did all the different browser tests on 17.11

VeeTeeDev:
Thinking about it... My pipelines are named 'footer' and 'header'.
Seems that GoCD is adding the id="footer" to the div. Removing the id of the div made it look much better :-)

VeeTeeDev:
Hey guys,

This is still not released ?

PR #4015, opened 2017-11-15
#4007 - Make VSM work with pipeline named "header"

arvindsv:
* Prefix all VSM node IDs with "id_" to prevent conflicts with existing DOM nodes such as "header", "footer", etc.

* This is just a temporary solution. This JS file needs a full rewrite, which does not do direct HTML-munging and a bunch of other anti-patterns.

@maheshp I've just put this in the 17.12 milestone, but feel free to move it back a release, if you can't take a look in time. 

This is for #4007.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
+++ b/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js
@@ -53,6 +53,10 @@ Graph_Renderer = function (container) {
         container.scrollLeft(0);
     }
 
+    function sanitizeForID(value) {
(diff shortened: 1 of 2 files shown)


maheshp:
@arvindsv looks like the changes affect the dependency arrows, I think thinks around [sortDependencyArrows](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_renderer.js#L516) needs to be taken care of as well.

Used build.gocd backup to generate the below VSM.

screen shot 2017-11-16 at 9 33 11 am

arvindsv:
Closing. Needs more work/thought.

PR #4017, opened 2017-11-16
Fixed misalignment of 'Pipeline' label on history page header for fir…

maheshp:
…efox


ketan:
Why was this an issue only in FF, and how does this fix not affect other browsers?

maheshp:
From what @naveenbhaskar said looks like FF is not considering default position and needs an explicit `left` position. This could also because some other css is affecting the position behaviour.

Issue #4018, opened 2017-11-16
Duplicated logs at the console on GoCD v17.11.0 on Safari and Firefox browsers


wlsf82:
##### Issue Type
- Bug Report

##### Summary
On v17.11.0 of GoCD, when running a pipeline and visualizing the logs in the console on Safari (or Firefox) browser, they get duplicated.
Note: After refreshing the page, the duplications are gone.
Note 2: This is not perceived in Chrome, for instance.

###### Basic environment details
* Go Version: `17.11.0 (5520-9f6909e2f64b07d2dce5cecd4ea5b92b8e19d6b1)`
* JAVA Version: `1.8.0_131`
* OS: `Linux 3.13.0-123-generic`
* Browser vendor and version: `Safari version 11.0.1 (13604.3.5)`

##### Steps to Reproduce
1. Run any pipeline on Safari (or Firefox) with GoCD v17.11.0 and start looking at the console tab of the running job.

##### Expected Results
No duplications should be displayed in logs of the console.

##### Actual Results
Duplications are displayed in the logs of the console.
Note: Sometimes duplications are displayed while the job is running and sometimes they are only displayed after the job is finished.

##### Screenshots
<img width="695" alt="screen shot 2017-11-16 at 10 22 03" src="https://user-images.githubusercontent.com/2768415/32886947-a66a834a-cac2-11e7-9600-2fc6537752b1.png">
<img width="497" alt="screen shot 2017-11-16 at 10 21 56" src="https://user-images.githubusercontent.com/2768415/32886948-a6828f26-cac2-11e7-98c5-de394a77c64f.png">

##### Other info
Notes from Gitter conversation:
```
Ketan Padegaonkar @ketan 10:37
what browser are you on, BTW?

Walmyr @wlsf82 10:38
I’m using Safari. Let me test on Chrome

Ketan Padegaonkar @ketan 10:38
that might be a culprit.
safari is known to have several issues with websockets (which is what accidentally got turned on in 17.11)

Walmyr @wlsf82 10:40
yeah, it doesn’t seem to happen on Chrome

Vierbergen Tim @VeeTeeDev 10:41
I think I was on Chrome or Firefox
not sure
but I'll check later

@wlsf82 — we can reproduce the issue on firefox. Its due websocket connection dropping off while the console logs are streamed.
`The connection to wss://build.gocd.org/go/console-websocket/<BUILD_URL>?startLine=0 was interrupted while the page was loading.`
```

[...]

PR #4024, opened 2017-11-20
Fix multiple logs showing on build details page tab (#4018)

GaneshSPatil:
* Fix broken websocket ping
* Fix multiple eventlisteners being added for the same event
* Add tests for websocket wrapper

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
@@ -1,0 +1,321 @@
+/*
+ * Copyright 2017 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("Websocket Wrapper", function () {
+  var MockedWebsocket, oftenWrapper, originalOften;
+
+  beforeEach(function () {
+    oftenWrapper = {
+      often: jasmine.createSpy(),
+      start: jasmine.createSpy(),
+      wait:  jasmine.createSpy(),
+      done:  jasmine.createSpy()
+    };
+
+    originalOften = often;
+
+    often = oftenWrapper.often;
+    start = oftenWrapper.start;
+    wait  = oftenWrapper.wait;
+
+    often.and.returnValue(oftenWrapper);
+    start.and.returnValue(oftenWrapper);
+    wait.and.returnValue(oftenWrapper);
+
+    MockedWebsocket = function () {
+      this.close = function () {
+      }
+    };
+  });
+
+  afterEach(function () {
+    MockedWebsocket = undefined;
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
```diff
@@ -0,0 +1,362 @@
...
+    assertEquals(15000, WebSocketWrapper.TIMEOUT_DEFAULT);
+    assertEquals(5000, WebSocketWrapper.TIMEOUT_START);
+  });
+
+  it("test_websocket_wrapper_adds_events_on_the_base_websocket", function () {
+    var options       = {};
+    options.Websocket = MockedWebsocket;
+    var myWebsocket   = new WebSocketWrapper(options);
+
+    assertEquals('function', typeof myWebsocket.ws.onopen);
```
ketan: `expect(_.isFunction(thing)).eq(true)`

server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
```diff
@@ -0,0 +1,362 @@
...
+    assertEquals('function', typeof myWebsocket.ws.onclose);
+    assertEquals('function', typeof myWebsocket.ws.onmessage);
+    assertEquals('function', typeof myWebsocket.ws.onerror);
+  });
+
+  it("test_websocket_wrapper_should_emit_onopen_event_when_websocket_connection_is_established", function () {
+    var options       = {};
+    options.Websocket = MockedWebsocket;
```
ketan: `var options = {Websocket: MockedWebsocket}`, and inline the variable?

server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
```diff
@@ -0,0 +1,362 @@
...
+    var myWebsocket   = new WebSocketWrapper(options);
+    myWebsocket.on("open", mySinon.stub);
+
+    assertEquals(false, mySinon.isCalled);
```
ketan: can we not use spies and stubs from jasmine?

server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
```diff
@@ -0,0 +1,362 @@
...
+    });
+
+    myWebsocket.on("beforeInitialize", mySinon.stub);
+
+    assertEquals(false, mySinon.isCalled);
+    assertEquals(0, mySinon.callCount);
+
+    myWebsocket.ws.onerror({data: "Boom!"});
+
+    setTimeout(function () {
```
ketan: This is a bad idea — because the callback is invoked after the test finishes, exceptions raised will likely be marked on a wrong test. Worse, if the callback is invoked after all tests finish execution, the exception will never be recognized.

Check https://jasmine.github.io/2.3/introduction#section-Asynchronous_Support

server/webapp/WEB-INF/rails.new/spec/javascripts/websocket_wrapper_spec.js
```diff
@@ -0,0 +1,318 @@
...
+
+  beforeEach(function () {
+    oftenWrapper = {
+      often: jasmine.createSpy(),
+      start: jasmine.createSpy(),
+      wait:  jasmine.createSpy(),
+      done:  jasmine.createSpy()
+    };
+
+    often = oftenWrapper.often;
```
ketan: This will overwrite the global, and leave it as is. Might have an impact on other tests. Might make sense to rollback these vars to their original state?

ketan:
When a build is out, can we ask @wlsf82 to test it out on his environment?

GaneshSPatil:
@ketan - will do!

PR #4025, opened 2017-11-20
Do not send JobPlans to agents

maheshp:
* Currently the BuildAssignment object serialized and sent to agent consists of DefaultJobPlan, the JobPlan object is something internal to server and should not be passed to agent.
* BuildAssignment is a DTO passed to agent with all the necessary information for agent to 
execute a job.
* Moved publish artifcats behaviour from JobPlan to ArtifactsPublisher
* Ensure build assignment has the required environment_variables resolved from JobPlan and 
   Environment. The env variables are resolved in the following order
   1. apply -> variables_for_an_environment
   2. apply and override -> variables_for_job_derived_from_job_plan
   3. apply and override -> trigger_variables_for_a_job.

### DIFF
--- a/agent/test/unit/com/thoughtworks/go/agent/JobRunnerTest.java
+++ b/agent/test/unit/com/thoughtworks/go/agent/JobRunnerTest.java
@@ -101,25 +101,6 @@ public class JobRunnerTest {
         verifyNoMoreInteractions(resolver);
     }
 
-    private BuildWork getWork(JobConfig jobConfig) {
-        CruiseConfig config = new BasicCruiseConfig();
-        config.server().setArtifactsDir("logs");
-        String stageName = "mingle";
-        String pipelineName = "pipeline1";
-        config.addPipeline(BasicPipelineConfigs.DEFAULT_GROUP, new PipelineConfig(new CaseInsensitiveString(pipelineName), new MaterialConfigs(), new StageConfig(
-                new CaseInsensitiveString(stageName), new JobConfigs(jobConfig))));
-
-        String pipelineLabel = "100";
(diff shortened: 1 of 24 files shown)


ketan:
BTW — this sort of a change will likely cause objects in flight to not be serialized/deserialized properly and the server and might throw up unexpected `NotSerializableException` exceptions.

maheshp:
Valid point @ketan. Just wondering in what scenario can something like this happen, trying to list out few scenarios.  

1. If an agent in building during a server upgrade, post completion of the build the agent would upgrade first before picking up a new job.
2. If an agent is idle during the server upgrade, the agent would upgrade once the server comes up before picking up a new job.

ketan:
> If an agent in building during a server upgrade, post completion of the build the agent would upgrade first before picking up a new job.

My point (and I was not clear about this) is that — users should expect to see some exceptions when the server upgrades this time around.

In this particular case, I *believe* the agent will simply [`exit()`](https://github.com/gocd/gocd/blob/72ea08b1f0a578181aeb4b4742b3f49c370cf73e/agent/src/com/thoughtworks/go/agent/service/AgentUpgradeService.java#L98) while a build is running. Obviously in such a case, the agent will again attempt to ask for a build, and the server will assign it again (assuming that the agent went offline). You may need to verify this hypothesis, or @jyotisingh can simply answer this :)

jyotisingh:
@maheshp - Rebase the PR. Functional tests are failing since you haven't rebased, the schema version on master has changed ([VSM](https://build.gocd.org/go/pipelines/value_stream_map/trigger/1852)).

Issue #3984, opened 2017-11-06
Latest version 17.11.0 fails to migrate config


pedrocarrico:
##### Issue Type
- Bug Report

##### Summary

I'm upgrading the server to the latest version (17.11.0) coming from the previous (17.10.0) and the server can't start because it can't migrate the config.

##### Environment

The server environment is running on a docker container on an EC2 instance.

###### Basic environment details

* Go Version: `17.11.0 (5520-9f6909e2f64b07d2dce5cecd4ea5b92b8e19d6b1)`
* JAVA Version: `1.8.0_151`
* OS: `Linux 4.4.41-35.53.amzn1.x86_64`
* Browser vendor and version (if relevant): `Not relevant`

##### Steps to Reproduce

1. Have a GoCD server version 17.10.0 with a pipeline configuration with a task like the one pasted below
2. Have a pipeline just like the 
3. Upgrade to latest GoCD server version 17.11.0
4. See that it fails upon starting up with the errors described below.

##### Expected Results

GoCD would be able to migrate the config from the previous version.

##### Actual Results

GoCD server fails to start.

##### Possible Fix

I don't really know 😞 

##### Log snippets

```log
2017-11-06 16:35:19,318 INFO  [main] ConfigRepository:220 - Upgrading config file from version 95 to version 97
2017-11-06 16:35:19,527 ERROR [Thread-75] PluginInfoBuilder:55 - Failed to fetch Plugin Settings metadata for plugin cd.go.authentication.ldap. Maybe the plugin does not implement plugin settings and view?
2017-11-06 16:35:19,544 ERROR [Thread-75] PluginSettingsMetadataLoader:75 - Failed to fetch Plugin Settings metadata for plugin cd.go.authentication.ldap. Maybe the plugin does not implement plugin settings and view?
2017-11-06 16:35:19,810 INFO  [main] ConfigRepository:229 - Finished upgrading config file
2017-11-06 16:35:19,818 INFO  [main] GoConfigMigrator:105 - [Config Save] Starting Config Save post upgrade using FullConfigSaveNormalFlow
2017-11-06 16:35:20,745 ERROR [Thread-75] PluginSettingsMetadataLoader:75 - Failed to fetch Plugin Settings metadata for plugin yum. Maybe the plugin does not implement plugin settings and view?
[...]

PR #4026, opened 2017-11-20
Update server side validation to not use file system -#3984

bdpiparva:
* See https://github.com/gocd/gocd/issues/3984\#issuecomment-342476130 for more information.
* Using filenames to check for nesting and allowed directory
configurations instead of the filesystem as the checks on the server's
filesystem wouldn't apply on the agent side which is where the Task
working directory and material destination directory would be used.

### DIFF
--- a/base/src/com/thoughtworks/go/util/FilenameUtil.java
+++ b/base/src/com/thoughtworks/go/util/FilenameUtil.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.util;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+
+public class FilenameUtil {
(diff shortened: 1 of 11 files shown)


base/src/com/thoughtworks/go/util/FilenameUtil.java
```diff
@@ -0,0 +1,38 @@
...
+package com.thoughtworks.go.util;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+
+public class FilenameUtil {
+    public static boolean isSubdirectoryOf(String parent, String subdirectory) throws IOException {
```
ketan: Did you forget to remove the old impl? Also, this method name is ambiguous, IMO. I don't know if there are cases where this answer requires us to actually resolve file paths — can we confirm?

base/src/com/thoughtworks/go/util/FilenameUtil.java
```diff
@@ -0,0 +1,38 @@
...
+        final String normalizedParentPath = FilenameUtils.normalize(parent + File.separator);
+        final String normalizedSubDirPath = FilenameUtils.normalize(subdirectory + File.separator);
+        return StringUtils.isNotBlank(normalizedParentPath) && StringUtils.isNotBlank(normalizedSubDirPath) && normalizedSubDirPath.startsWith(normalizedParentPath);
+    }
+    public static boolean isFolderInsideSandbox(String path) {
```
ketan: Can we instead rename this to `isNormalizedPathOutsideWorkingDir` or something similar to indicate that this fundamentally performs path normalization. `insideSandbox` seems to invoke the feeling that it actually does resolution of file paths.

base/src/com/thoughtworks/go/util/FilenameUtil.java
jyotisingh: Didn't forget to remove it, its been left behind intentionally. There are some very specific cases of this validation being done wrt server only directories and similar stuff for agent-only directories which need file system level resolution. Comment about method renaming makes sense, will do.

base/src/com/thoughtworks/go/util/FilenameUtil.java
jyotisingh: cool

Issue #4005, opened 2017-11-13
Is the VNC option necessary on Go Agents? 


varshavaradarajan:
##### Summary

Currently, while starting the go agent via  https://github.com/gocd/gocd/blob/master/installers/go-agent/release/agent.sh , if VNC is set to Y, an attempt is made to start the vnc server with display number 3. 

Issues: 

1. Suppose vnc is not installed, this script proceeds with starting the agent anyway. 
2. Suppose the go user doesn't have the vnc password set yet, then, it gets prompted and unless someone enters the password, the go agent does not get started
3. Suppose the display is not available, the script proceeds with starting the agent.

While 1 and 3 require some configuration to be done outside of go and 2 hinders automation, is it necessary to take care of starting the vnc server  while starting the go agent? 
I think that this option was provided in case the agent was running some tests which needed a display. But the argument is that anything that is needed on the agent is to be maintained completely by the user. We don't install any other package or configure anything else while starting the go agent, so why should vnc be an exception? Why start the vnc server when the user has to anyway take care of making sure its starts properly? 

If we do get rid of this, the users who have been using this won't get affected until they upgrade the bootstrapper. Thoughts?

arvindsv:
I don't mind getting rid of it. It does seem out of place.

ketan:
# Alternatives to the `VNC` flag

This flag was introduced in the configuration file to 

(a) allow the GoCD agent process to be able to access an X11 window system for running UI tests and
(b) to allow users to connect to the running vnc server remotely to diagnose problems

This added an additional overhead on having to **interactively** setup a VNC password the first time VNC was executed by the GoCD agent.

It is now recommended that users wrap their tasks with the [`xvfb-run`](http://manpages.ubuntu.com/manpages/xenial/man1/xvfb-run.1.html) command as follows:

**Before:**

```bash
rake ci:jasmine
```

**After:**

```bash
xvfb-run rake ci:jasmine
```

PR #4028, opened 2017-11-21
Remove the code related to starting vncserver. (#4005)

varshavaradarajan:
* Users having the old bootstrapper will not be affected by this change.

#4005

### DIFF
--- a/installers/go-agent/release/agent.sh
+++ b/installers/go-agent/release/agent.sh
@@ -95,7 +95,6 @@ AGENT_DIR="$(cd "$CWD" && pwd)"
 
 AGENT_MEM=${AGENT_MEM:-"128m"}
 AGENT_MAX_MEM=${AGENT_MAX_MEM:-"256m"}
-VNC=${VNC:-"N"}
 AGENT_WORK_DIR="${AGENT_WORK_DIR:-$AGENT_DIR}"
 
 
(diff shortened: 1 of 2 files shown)


ketan:
Also, missed [this line](https://github.com/gocd/gocd/blob/791933f87a9b3f2573ca14c4aadf23cc842b140b/installers/go-agent/release/go-agent.default#L3).

ketan:
Also, missed [this line](https://github.com/gocd/gocd/blob/791933f87a9b3f2573ca14c4aadf23cc842b140b/installers/go-agent/release/go-agent.default#L3).

installers/go-agent/release/agent.sh
```diff
@@ -119,13 +118,6 @@ else
   PID_FILE="$AGENT_WORK_DIR/go-agent.pid"
 fi
 
-if [ "$VNC" == "Y" ]; then
```
ketan: Thinking out aloud, should we warn if this variable is set, perhaps with some like to a documentation about possible fallbacks?

varshavaradarajan: Really? I thought a note in the release notes should be fine. Also, is this even a change that would require us to add a deprecation note given that only people who upgrade the bootstrapper would face the problem? I don't mind adding the warning though. What fallbacks? We should ask people to start the vncserver on their own, right?

ketan: > What fallbacks? We should ask people to start the vncserver on their own, right?

That's what I meant as "fallback", not automated, but something users can use to replace this functionality.

ketan: I'm not suggesting that we do this, I'm just thinking if there are things we can possibly do to make it easy for users.

ketan:
@varshavaradarajan — when you have time, can you take a look and see if the warning is good enough?

PR #4030, opened 2017-11-21
Plugins SPA - enabled by default.

varshavaradarajan:
* The old plugins page is available at go/admin/old_plugins

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/plugins.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/plugins.scss
@@ -106,6 +106,8 @@ $elastic-profile-icon-hover-color: #000;
     text-overflow: ellipsis;
     white-space:   nowrap;
     padding-right: 30px;
+    font-weight:   400;
+    font-size:     rem-calc(13px);
 
     .value {
       font-weight: 400;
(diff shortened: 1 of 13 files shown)


ketan:
I believe you'd raised some concern about icons for bad plugins, is that logged somewhere?

varshavaradarajan:
https://github.com/gocd/gocd/pull/3932#issuecomment-336365888

server/webapp/WEB-INF/rails.new/webpack/views/plugins/plugins_widget.js.msx
```diff
@@ -87,18 +87,7 @@ const PluginsWidget = {
     };
 
     this.edit = function (pluginInfo) {
-      const newPluginSetting   = Stream(new PluginSetting({pluginId: pluginInfo.id()}));
-      const setPluginSetting = function () {
```
maheshp: Is this behaviour tested, should we change the tests for this?

server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/plugins.scss
```diff
@@ -106,6 +106,8 @@ $elastic-profile-icon-hover-color: #000;
     text-overflow: ellipsis;
     white-space:   nowrap;
     padding-right: 30px;
+    font-weight: 400;
```
maheshp: format this

Issue #3982, opened 2017-11-06
GO_SERVER_PORT environment variable is not honored


EbenZhang:
##### Issue Type

- Bug Report

##### Summary

After upgraded to v17.11.0, GO_SERVER_PORT environment variable is not honored.

###### Basic environment details

Go Server Version: | 17.11.0 (5520-9f6909e2f64b07d2dce5cecd4ea5b92b8e19d6b1)
JVM version: | 1.8.0_102
OS Information: | Windows 10 10.0
Usable space in artifacts repository: | 503 GB
Database schema version: | 1708001
GOCD Service command line: `E:\GoServer\cruisewrapper.exe -s E:\GoServer\config\wrapper-server.conf`

###### Additional Environment Details

See attached:  [goenv.txt](https://github.com/gocd/gocd/files/1444684/goenv.txt)

##### Steps to Reproduce

- Make sure port 80 is not occupied
- Make sure the logon user for GOCD Server Service has the admin permission
- Start the GOCD service

##### Expected Results
The server is listening on port 80

##### Actual Results
The server is listening on port 8153

##### Any other info
It was working well before I upgraded it to v17.11.0

akshaydewan:
Hi @EbenZhang, which version of Go Server were you using before upgrading to v17.11.0? (This seems to be broken in v17.10 as well)

varshavaradarajan:
Closing as 17.12.0 was released.

EbenZhang:
Tried on 17.12.0 (5626-cb7df2ffe421e43f2a682a7a323cb3a3e30734cc).  It still doesn't work.

ketan:
In 17.12.0 We've changed how port numbers assignment works on windows:

Please edit the file `config/wrapper-properties.conf` in the server install directory and add the following to override the defaults:

```
wrapper.java.additional.9=-Dcruise.server.port=8153
wrapper.java.additional.10=-Dcruise.server.ssl.port=8154
```

EbenZhang:
it works with the configuration in the wrapper-properties.conf.  Thanks.

EbenZhang:
@ketan every time after upgraded the config file gets overwritten to default port. Is there a way to keep the port settings?

ketan:
[...]

PR #4038, opened 2017-11-27
Read Go Server ports from wrapper-server.conf to fix Windows service …

akshaydewan:
…(#3982)

Tested this change locally and by generating the windows installer.

@jyotisingh Do you think this is alright, or should we take some other approach to fix this?

### DIFF
--- a/installers/go-server/win/config/wrapper-server.conf
+++ b/installers/go-server/win/config/wrapper-server.conf
@@ -47,10 +47,6 @@ set.JVM_DEBUG_PORT=5005
 set.GC_LOG_LOCATION="%CRUISE_SERVER_DIR%\go-server-gc.log"
 set.JAVA_HOME=%GO_SERVER_JAVA_HOME%
 
-# Read Go Server Ports from Environment, or set default
-set.default.GO_SERVER_PORT=8153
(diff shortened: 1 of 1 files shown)


installers/go-server/win/config/wrapper-server.conf
```diff
@@ -64,8 +68,8 @@ wrapper.java.additional.5=-Duser.language=en
 wrapper.java.additional.6=-Duser.country=US
 wrapper.java.additional.7="-Dcruise.config.dir=%CRUISE_SERVER_DIR%\config"
 wrapper.java.additional.8="-Dcruise.config.file=%CRUISE_SERVER_DIR%\config\cruise-config.xml"
-wrapper.java.additional.9=-DreservedForFuture.1
-wrapper.java.additional.10=-DreservedForFuture.2
+wrapper.java.additional.9=-Dcruise.server.port=%GO_SERVER_PORT%
+wrapper.java.additional.10=-Dcruise.server.ssl.port=%GO_SERVER_SSL_PORT%
```
ketan: @akshaydewan — as discussed. can you inline the environment variables here and remove the declaration on line 50 so we can merge this PR?

Issue #4039, opened 2017-11-28
Enhance api support


jyotisingh:
* Add list of enabled plugins along with the kind of extension and addons to api/support

rajiesh:
Verified on 17.12.0 (5621-158fee2729f0a07475594782e6a5716ac3154f8a)

PR #4042, opened 2017-11-29
Include plugin information in /api/support (#4039)

akshaydewan:


jyotisingh:
For api/support we care only about the plugins which got loaded since those are the only ones which could potentially create issues for the server. I see that this PR fetches only those (since it uses the metadata stores), so status is not really required I think. 
Version makes sense. 
What information do we get from the bundled state? We know which ones are bundled so everything else would be external anyhow.

ketan:
State tells us if the plugin has a problem loading up, or initializing - in
which case, depending on the context, we may ignore that plugin from
further analysis. Bundled state to know if someone has overridden a plugin
from  external folder. I don't think I'd need more info. I checked the
plugin info API, and did not see any other useful info.

akshaydewan:
Added version, bundled state and status. Here's what the JSON structure looks like:
```
"Plugin information": {
        "plugins": [
            {
                "id": "yum",
                "type": "",
                "version": "2.0.3",
                "bundled_plugin": true,
                "status": {
                    "state": "INVALID",
                    "messages": [
                        "Plugin with ID (yum) is not valid: Incompatible with current operating system 'Mac OS X'. Valid operating systems are: [Linux]."
                    ],
                    "rootCauseIfInvalid": null
                }
            },
            {
                "id": "cd.go.authentication.passwordfile",
                "type": "authorization",
                "version": "1.0.1-48",
                "bundled_plugin": true,
                "status": {
                    "state": "ACTIVE",
                    "messages": [],
                    "rootCauseIfInvalid": null
                }
            },
            {
                "id": "cd.go.authentication.ldap",
                "type": "authorization",
                "version": "1.0.1-49",
                "bundled_plugin": true,
                "status": {
                    "state": "ACTIVE",
                    "messages": [],
                    "rootCauseIfInvalid": null
                }
            },
            {
                "id": "yaml.config.plugin",
                "type": "configrepo",
                "version": "0.6.0",
                "bundled_plugin": true,
                "status": {
                    "state": "ACTIVE",
                    "messages": [],
                    "rootCauseIfInvalid": null
                }
            },
            {
                "id": "json.config.plugin",
                "type": "configrepo",
                "version": "0.2.1",
                "bundled_plugin": true,
                "status": {
                    "state": "ACTIVE",
                    "messages": [],
                    "rootCauseIfInvalid": null
                }
            }
        ]
    }
```

Issue #3442, opened 2017-05-02
Console log not being rendered properly (Encoding issue)


jyotisingh:
##### Issue Type

- Bug Report

##### Environment
Could reproduce this on build.gocd.io. Not reproducible on my laptop.

##### Steps to Reproduce

1. Copy the attached console.log.txt file to the artifact location of any of the completed jobs on the server.
2. Rename the file to console.log. Amend file permissions so that 'go' user has access to the file(if applicable).
3. Refresh the job details page, switch to console tab

##### Expected Results
Expected the tree structure to show up

##### Actual Results
See some gibberish instead
<img width="778" alt="screen shot 2017-05-02 at 11 58 22 am" src="https://cloud.githubusercontent.com/assets/1046342/25606403/b5db95c4-2f2e-11e7-81b6-f147fe5e3f97.png">

[console.log.txt](https://github.com/gocd/gocd/files/969581/console.log.txt)

ketan:
> Could reproduce this on build.gocd.io. Not reproducible on my laptop.

Very much reproducible — https://build.gocd.io/go/tab/build/detail/gocd-docs.go.cd-master/9/Build/1/build_job

marques-work:
I've tried without success to reproduce this locally. I suspect it will never happen on MacOS X as the default is always UTF-8. We must be changing the encoding before sending the response because the file is valid UTF-8 (according to iconv). I'll dig into this some more.

ketan:
> I've tried without success to reproduce this locally. I suspect it will never happen on MacOS X as the default is always UTF-8. We must be changing the encoding before sending the response because the file is valid UTF-8 (according to iconv). I'll dig into this some more.

That's werid. I was seeing junk chars instead of the fancy npm tree on the URL I pasted in the comment before. I suspect something changed with chrome. Next time, I'll be sure to grab a screenshot if I come across this.

marques-work:
@ketan there might be something more to it. I'm on Chrome Version 57.0.2987.133 (64-bit). I see junk characters with the URL, but not when I take the same file and use it in the dev instance...

ketan:
Kidding me — I see junk on that page now.

![screen shot 2017-05-10 at 3 36 00 pm](https://cloud.githubusercontent.com/assets/10598/25893688/64d6d01c-3596-11e7-974c-503d4f823c15.png).

If it helps, here are the request/response headers: Clearly mentioning the encoding as utf-8.

[...]

PR #4044, opened 2017-11-30
Fix console log character encoding

### LINKED ISSUE #3926: Bug: Console logs stop updating when non-ascii characters are encountered
**Bug Report**

**Summary**

Console logs containing non-ASCII characters are truncated at the point that the non-ASCII characters occur. Only the portion of the console log that is generated prior to the occurrence of the non-ASCII characters is displayed. The log is still generated and view-able once the stage completes via the "raw output" link, however the non-ASCII characters are replaced with this symbol "�". 

We discovered this issue when upgrading from version 17.2 to version 17.10 and were forced to back out of the upgrade as a result of the inability to follow the progress of any stages that display non-ASCII characters in the console log.  Many of these stages are lengthy, causing the stage to appear stalled since the log is no longer updating and the "raw output" is not available until the stage completes.

The raw output in 17.2 also replaces non-ASCII characters with "�", but does process and display them correctly in the scrolling log.

The specific characters in our logs that may be contributing to the problem are: æ, ä, ö, ê, à, ü, ö, ß, ó, í, ç, á, 

**Environment**

GoCD 17.10
- non-ASCII characters cause console log to stop rendering "real time"
- non-ASCII characters are replaced by "�" in the "raw output"

GoCD 17.2
- non-ASCII characters are replaced by "�" in the "raw output"

Windows Server 2008 R2
Number of pipelines 343
Number of agents 24

 AND 

Windows 7
Number of pipelines 195
Number of agents 12

Firefox Version 56.0
Google Chrome Version 61.0.3163.100
Microsoft Edge 38.14393.1066.0

**Steps to Reproduce**

1. Generate an output containing multiple instances of non-ASCII characters
2. View the console log real time
3. View the raw output

**Expected Results**

Scrolling console log should continue to scroll and correctly display all characters.  Raw output should correctly display all characters.

**Actual Results**

Scrolling console log stops scrolling at the point at which the non-ASCII characters are being processed.  Raw output replaces non-ASCII characters with "�".

**Possible Fix**

Render the characters in the scrolling console log in the same manner that they were rendered in 17.2 and below (I'm not sure exactly where the change took place that caused the problem between versions 17.2 and 17.10).

**Log snippets**

Log is generated normally until this point: (captured from the raw output since the scrolling log truncates prior to displaying this info).  This is just a small portion of the log - there are several occurrences of non-ASCII characters in this log.

image

ketan:
Fixes #3926, #3442


agent/src/com/thoughtworks/go/agent/AgentWebSocketClientController.java
```diff
@@ -166,7 +166,8 @@ private void runBuild(BuildSettings buildSettings) {
             buildConsole = new ConsoleOutputTransmitter(
                 new RemoteConsoleAppender(
                     urlService.prefixPartialUrl(buildSettings.getConsoleUrl()),
-                    httpService)
+                    httpService,
+                    getSystemEnvironment().consoleLogCharsetAsCharset())
```
jyotisingh: This would be set at the server side not agent. The charset that was passed along from the server to the agent needs to be passed along to the Console appender

common/src/main/java/com/thoughtworks/go/buildsession/ExecCommandExecutor.java
```diff
@@ -83,6 +84,7 @@ private CommandLine createCommandLine(String cmd) {
         } else {
             commandLine = CommandLine.createCommandLine(cmd);
         }
+        commandLine.withEncoding(new SystemEnvironment().consoleLogCharset());
```
ketan: @jyotisingh — we'd likely need to inject the charset to all `BuildCommandExecutor` impls, or add it to `BuildSession` to make it available to all executors.

Thoughts?

jyotisingh: I prefer adding it to the BuildSession. One way would be to pass along the charset with [BuildSettings](https://github.com/gocd/gocd/blob/6889503814329ba12794b9a8a86971afa84bfc3c/common/src/main/java/com/thoughtworks/go/domain/BuildSettings.java#L20) and then use it to construct the `BuildSession`. That way in future too if any other CommandExecutor needs access to commandline, it could just get hold of the charset from `BuildSession`

ketan:
For testing this PR, create a file (`/tmp/unicode.txt`) with some unicode characters, emojis. I used [this file](https://gist.github.com/ketan/9dc86e34f30ed9c560d029385cce063a), make sure to click "raw" and download using wget, copy/pasting is not recommended.

Use the following config.xml snippet:

```xml
    <pipeline name="up42">
      <materials>
        <git url="test-repo" />
      </materials>
      <stage name="up42_stage">
        <jobs>
          <job name="up42_job">
            <tasks>
              <exec command="bash" args="-c 'while true; do cat /tmp/unicode.txt; sleep 1; done'" />
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
```

### On 18.1 of GoCD server and agent:

* set the `file.encoding` system property to `cp1252` (to simulate GoCD running on windows). This needs to be done on both the server and agent.
* Run the build
* When viewing the console logs, the non-ascii part the output will appear garbled

### On latest (experimental) GoCD

* set the `file.encoding` system property to `cp1252` (to simulate GoCD running on windows). This needs to be done on both the server and agent.
* Run the build
* the console output will not be garbled

### On latest (experimental) GoCD

* set the `go.console.log.charset` system property to `cp1252` (to force console log encoding to be that used by windows). This needs to be done on the server, doing it on the agent has no effect
* Run the build
* the console output will be garbled — because the unicode text is being rendered using `cp1252` encoding. This is expected behaviour.

adityasood:
Tested the above steps

1. On 18.1 of GoCD server and agent:
The non-ascii part the output appeared garbled after file.encoding is set to cp1252 on both GoCD server and agent

2. On latest (experimental) GoCD git sha (6029...)
The console output is not garbled after file.encoding is set to cp1252 on both GoCD server and agent

3. On latest (experimental) GoCD git sha (6029...)
The console output is garbled after go.console.log.charset is set to cp1252 on GoCD server

lobster2012-user:
It works for me. 
```
wrapper.java.additional.100=-Dgo.console.log.charset=CP866
```

PR #4045, opened 2017-11-30
Fixing build work test

maheshp:
Cannot find a better way of asserting `GO_SERVER_URL` variable is overridden by job env variables.


ketan:
Can we maybe pull out the line number index of the 2 messages and compare the index?

PR #4048, opened 2017-12-03
Clean Shell Scripts

Androbin:
You need a space after the `[` and before the `]`.
The order of the `2>&1` and the redirect matters. The `2>&1` has to be last.
Double quote array expansions to avoid re-splitting elements.

### DIFF
--- a/manual-testing/ant_hg/make-one-change.sh
+++ b/manual-testing/ant_hg/make-one-change.sh
@@ -2,7 +2,7 @@
 if [ "$USER" == '' ]; then
     USER=ManualUser
 fi
-if ["$FOLDER" == '']; then
+if [ "$FOLDER" == '' ]; then
   echo 
   FOLDER=dummy
 fi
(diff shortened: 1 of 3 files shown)


ketan:
Hi @Androbin — thanks for the PR. I just realized that the stuff in `manual-testing` is unused. I'll simply cherry-pick the changes to the jruby script directly to master and close this PR.

ketan:
Applied directly at 8fa812d5fee4b82d6abc1134065bb0ceba5fd702

Androbin:
Thanks @ketan ! You said that `manual-testing` is unused. But what about `server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/nokogiri-1.6.6.2-java/test_all` ?

ketan:
> Thanks @ketan ! You said that manual-testing is unused. But what about server/webapp/WEB-INF/rails.new/vendor/bundle/jruby/1.9/gems/nokogiri-1.6.6.2-java/test_all ?

That is provided by the [nokogiri gem](https://github.com/sparklemotion/nokogiri/blob/v1.6.2/test_all). Not something we wrote, or something we use.

PR #4056, opened 2017-12-06
Notify plugins on plugin_settings change

maheshp:
* With this commit, GoCD notifies plugins on any change to plugin settings.
* Errors from PluginSettings change notifcation are logged and ignored
  since not all plugins handle the notification request.


server/src/com/thoughtworks/go/server/service/PluginService.java
```diff
@@ -160,6 +161,21 @@ private void update(Username currentUser, LocalizedOperationResult result, Plugi
...
     }
 
+    private void notifyPluginSettingsChange(PluginSettings pluginSettings) {
+        String pluginId = pluginSettings.getPluginId();
+
+        for (GoPluginExtension extension : extensions) {
+            if (extension.canHandlePlugin(pluginId)) {
+                try {
+                    extension.notifyPluginSettingsChange(pluginId, pluginSettings.getSettingsMap());
+                    break;
```
bdpiparva: I think `break` should be outside of try block or It should be in finally to break loop once `extension.canHandlePlugin(pluginId)` you have extension for a pluginId.

maheshp:
Closed this in favour of https://github.com/gocd/gocd/pull/4084

Issue #3444, opened 2017-05-02
Cleanup multiple PluginInfo Builders


maheshp:
##### Issue Type

- Tech Debt

Currently there are three different ways to build PluginInfo. This needs to be cleaned up and moved to having a single way of building the PluginInfo e.g AuthorizationPluginInfoBuilder [1] 

The PluginInfos are readonly objects and there is no need to build this objects for each API request, this can be built when a plugin is loaded and added to store. A plugin unload should remove it from the store.

[1] https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/authorization/AuthorizationPluginInfoBuilder.java

maheshp:
Done as part of https://github.com/gocd/gocd/pull/4066

PR #4066, opened 2017-12-07
Removing plugin_info v2 endpoint deprecated in 17.9.0

maheshp:


ketan:
@maheshp — I'm going to wait on you to remove the service code related to plugin info v2. If the tests pass, can you merge it yourself?

maheshp:
This commit just removes the endpoints in rails, we need to cleanup unused classes on the services side, can be done as part of https://github.com/gocd/gocd/issues/3444

Issue #4071, opened 2017-12-08
Test artifacts configured using config-repo is uploaded to GoCD server's build artifacts' default location.


bdpiparva:
##### Issue Type
- Bug Report

##### Summary

I have job with artifacts configured using json config repo plugin(https://github.com/tomzo/gocd-json-config-plugin). Here is the configuration

```json
{
  "environment_variables": [],
  "format_version": 1,
  "group": "configrepo-example",
  "materials": [
    {
      "type": "git",
      "url": "https://github.com/bdpiparva/config_repo_sample"
    }
  ],
  "name": "firstpipe",
  "stages": [
    {
      "clean_working_directory": false,
      "environment_variables": [],
      "fetch_materials": true,
      "jobs": [
        {
          "artifacts": [
            {
              "source": "*.json",
              "type": "test"
            },
            {
              "source": "*.zip",
              "type": "build"
            }
          ],
          "environment_variables": [],
          "name": "build",
          "properties": [],
          "resources": [],
          "run_instance_count": null,
          "tabs": [],
          "tasks": [
            {
              "arguments": [],
              "command": "ls",
              "on_cancel": null,
              "run_if": "passed",
              "type": "exec",
              "working_directory": null
            }
          ],
          "timeout": 0
        }
      ],
      "name": "build",
      "never_cleanup_artifacts": false
    }
  ]
}
```

By default it should upload test artifacts under `testoutput` folder but this is not true in case of if pipeline is coming from config-repo.

@tomzo --- is this intentional behaviour ?. I can see(https://github.com/gocd/gocd/blob/master/server/src/com/thoughtworks/go/config/ConfigConverter.java#L501-L504) that all artifact without destination directory is uploaded in the artifacts directory instead of `testoutput` dir.

tomzo:
Hi,

> By default it should upload test artifacts under testoutput folder but this is not true in case of if pipeline is coming from config-repo.

I did not know about this behavior until now. So it is a bug. As you noticed that should be fixed in `ConfigConverter` somewhere around those lines.

bdpiparva:
This issue is fixed in #4083

PR #4077, opened 2017-12-11
Have different classes to represent artifact config and artifact plans

bdpiparva:
* Updated ConfigConvertor to convert  CRArtifact to ArtifactConfig based
on type(#4071)


config/config-api/src/com/thoughtworks/go/config/ArtifactConfig.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public void setDestination(String destination) {
+        this.destination = StringUtils.trim(destination);
+    }
+
+    public boolean equals(Object other) {
+        return this == other || other != null && other instanceof ArtifactConfig && equals((ArtifactConfig) other);
+    }
+
+    private boolean equals(ArtifactConfig other) {
```
varshavaradarajan: Inline method in public boolean equals

PR #4085, opened 2017-12-12
Allowing configuration of session cookie secure flag

jyotisingh:
* Leaving the default as secure flag to not be set as it could break existing setups, but users should be able to set it using -Dgo.sessioncookie.secure=Y

### DIFF
--- a/jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
+++ b/jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
@@ -109,11 +109,13 @@ public class Jetty9Server extends AppServer {
     }
 
     @Override
-    public void setSessionAndCookieExpiryTimeout(int sessionAndCookieExpiryTimeout) {
-        SessionCookieConfig cookieConfig = webAppContext.getSessionHandler().getSessionManager().getSessionCookieConfig();
-        cookieConfig.setHttpOnly(true);
-        cookieConfig.setMaxAge(sessionAndCookieExpiryTimeout);
+    public void setSessionCookieConfig() {
+        int sessionAndCookieExpiryTimeout = systemEnvironment.sessionTimeoutInSeconds();
         SessionManager sessionManager = webAppContext.getSessionHandler().getSessionManager();
+        SessionCookieConfig sessionCookieConfig = sessionManager.getSessionCookieConfig();
+        sessionCookieConfig.setHttpOnly(true);
+        sessionCookieConfig.setSecure(systemEnvironment.isSessionCookieSecure());
+        sessionCookieConfig.setMaxAge(sessionAndCookieExpiryTimeout);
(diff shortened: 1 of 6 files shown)


jetty9/src/com/thoughtworks/go/server/Jetty9Server.java
```diff
@@ -109,11 +109,13 @@ public void addExtraJarsToClasspath(String extraClasspath) {
...
-        SessionCookieConfig cookieConfig = webAppContext.getSessionHandler().getSessionManager().getSessionCookieConfig();
-        cookieConfig.setHttpOnly(true);
-        cookieConfig.setMaxAge(sessionAndCookieExpiryTimeout);
+    public void setSessionCookieConfig() {
+        int sessionAndCookieExpiryTimeout = systemEnvironment.sessionTimeoutInSeconds();
         SessionManager sessionManager = webAppContext.getSessionHandler().getSessionManager();
+        SessionCookieConfig sessionCookieConfig = sessionManager.getSessionCookieConfig();
+        sessionCookieConfig.setHttpOnly(true);
+        sessionCookieConfig.setSecure(systemEnvironment.isSessionCookieSecure());
+        sessionCookieConfig.setMaxAge(sessionAndCookieExpiryTimeout);
```
arvindsv: Did we talk about making this configurable, so that it can be set to `0`, so that it's a proper session cookie? That is, it's wiped out on browser close?

jyotisingh: There were some discussions around it a few times but I get a feeling we might have missed talking to you about it. The value for max-age was hardcoded to 2 weeks, but as part of 17.11 this was made configurable to allow users to reduce this to to a shorter period, that could mean setting it to `-1` to make it a proper session cookie, or to set it to say 1 hr(`3600`) or so if someone wanted it to be that way. This PR allows configuring it to be marked as secure. It would work if the installation was accessible via https only, else it would disallow users from logging in when using http with the flag set. The default behavior for both of these are the same as they used to be, intentionally.  Any concerns?

arvindsv: Sounds good, thanks. Is this documented?

jyotisingh: https://docs.gocd.org/current/configuration/dev_authentication.html 
Look for `go.server.session.timeout.seconds`, doesn't mention about `-1` - may be we should. Yet to add a note about the secure flag.

arvindsv: Ok!

jyotisingh: @arvindsv - Thanks for probing. We found an issue with the current implementation. Submitted a PR for the same -> https://github.com/gocd/gocd/pull/4099
Submitted a PR for the docs too -> https://github.com/gocd/docs.go.cd/pull/133

PR #4087, opened 2017-12-12
Split Resource and ResourceConfig

bdpiparva:


arvindsv:
Just out of curiosity: Why?

bdpiparva:
@arvindsv --- We required to split the Config from PersistentObject for artifacts. So we decided to make this change for all hybrid representation(Resources and EnvironmentVariables) we have for config and PersistentObject for consistency.

Issue #4091, opened 2017-12-13
Pass along JobIdentifier for Elastic Agent Plugins


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Pass along `JobIdentifier` to the elastic agent plugins for making better decisions while:
* `create-agent` request
    - `JobIdentifier` information will be used by the plugins for making better decisions about creating a new container for the current `create-agent` request.
If the plugin has already scheduled a request for creating an agent for the given `jobIdentifier`, on the next `create-agent` call, the plugin can skip the agent creation.

* `should-assign-work` request
    - `JobIdentifier` information will be used by the plugins for making better decisions about assigning a job to a particular agent.
The plugin can make a decision based on whether the current agent was brought up for the given jobIdentifier or not.

GaneshSPatil:
Related: https://github.com/gocd/kubernetes-elastic-agents/issues/3

GaneshSPatil:
Pass along JobIdentifier to the elastic agent plugins for making better decisions while:

* status-report request
    - JobIdentifier information will be used by the plugins for highlighting the container which is running the current job on the status report page.

    - Currently, navigating from `build_details_page` to `status_report_page` doesn't show the relation between which container is running the current job. 

    - JobIdentifier is optional for status-report page. It is only provided while navigating from `build_details_page`. JobIdentifier won't be provided while navigating from `elastic_profile_page` and `plugins_page`.

PR #4094, opened 2017-12-14
Pass along JobIdentifier to createAgent request for elasticAgent plugins #4091

GaneshSPatil:


jyotisingh:
@GaneshSPatil - we should pass along a jobIdentifier at the time of `should-assign-work` call too. Plugins could choose to honor it or not. That's the only way you for plugins to ensure only one job runs on a given container for instance.

PR #4095, opened 2017-12-14
[WIP] Split EnvironmentVariableConfig and DTO

bdpiparva:


varshavaradarajan:
The environment variables at the environment level will be taken from the environment config at the time of build assignment because the pipeline to which the job belongs can be removed from the environment before build assignment happens.  What breaks as a result of this is the job feeds API which we said we'll not fix - https://github.com/gocd/gocd/issues/4097

PR #4098, opened 2017-12-15
Removing unnecessary pipeline-scheduled topic

jyotisingh:
A message for each scheduled pipeline would be sent to this topic when there were 0 consumers for this. Hence deleting this.

For my future reference (git sha from old repo):
pipeline-schedule topic was added (bb602fcae4) to post messages about scheduled pipelines with a plan to consume these messages by the Elastic search module. A SearchListener was added (b6d592d0e8) to consume these messages. However, it appears that the plan to implement elastic search was scrapped at a later point(092b5b5b5), leaving behind the queue and related code.


arvindsv:
You know ... I always wondered about this every time I explained or explored something around scheduling. I thought it was something I didn't understand and couldn't find the consumers of. :)

+1 for removing this.

PR #4109, opened 2017-12-19
Fix test failures caused due to #4094

GaneshSPatil:
* Pass Matchers.any() matcher instead of null for a stub method in tests


GaneshSPatil:
Build passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/1966

Issue #4076, opened 2017-12-11
Pipeline Pause/Unpause API


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Create a new Pipeline Pause/Unpause API.

Problems with current API:
- [x]  Current API is not versioned under API namespace.
- [x] Returns empty response body ("") in case of success.
- [x] Returns a text response in case of errors.
- [x] Allows pausing a pipeline even when the pipeline is already paused.
- [x] Allows unpausing a pipeline even when the pipeline is already unpaused.

These problems should be fixed as part of the newer version of API.

ketan:
@rajiesh — can we get a round of testing on this and close?

rajiesh:
Verified on 18.2.0 (6208-f0f03120fc42d2ae1f230dd89996a3dc086912c5)

PR #4110, opened 2017-12-19
Introduce pipeline pause/unpause API v1 #4076

GaneshSPatil:


server/webapp/WEB-INF/rails.new/spec/controllers/api_v1/pipeline_operation_controller_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      allow(@pipeline_pause_service).to receive(:pause).with(@pipeline_name, @pause_cause, @user, anything) do |pipeline_name, pause_cause, user, result|
+        result.setMessage(LocalizedMessage.string('PIPELINE_PAUSE_SUCCESSFUL', pipeline_name));
+      end
+
+      post_with_api_header :pause, pipeline_name: @pipeline_name, pause_cause: @pause_cause
+      expect(response).to be_ok
+    end
+
+    it 'should show errors occurred while pausing the pipeline' do
```
varshavaradarajan: Should move to a different describe block? Add route tests?

server/webapp/localize.xml
```diff
@@ -794,4 +794,8 @@
     <entry key="CAN_NOT_UPDATE_RESOURCES_ON_ELASTIC_AGENT">Resources on elastic agents with uuids [{0}] can not be updated.</entry>
     <entry key="PENDING_AGENT_INVALID_OPERATION">Pending agents [{0}] must be explicitly enabled or disabled when performing any operations on them</entry>
     <entry key="NO_OPERATION_PERFORMED_ON_AGENTS">No Operation performed on agents.</entry>
+    <entry key="PIPELINE_PAUSE_SUCCESSFUL">Pipeline ''{0}'' paused successfully.</entry>
```
varshavaradarajan: `localize_ja.xml` hasn't been deleted, so, you need to add these entries there too?

arvindsv: Shouldn't we delete `localize_ja.xml`?

varshavaradarajan: Yes. I remember @GaneshSPatil and me talking about it. I suppose another PR for it? @ketan wanted to get rid of even localize.xml, I think

ketan: > Shouldn't we delete localize_ja.xml?

I vote for you for president...

PR #4113, opened 2017-12-19
PluginSettingsRequestProcessor delegates response building to extensions

maheshp:
* Moving away from PluginSettingsRequestProcessor handling multiple
  versions of plugin and building the response body.
* Plugin extensions would be responsible to build the plugin settings
  JSON based on the appropriate plugin version.


arvindsv:
img_5745

This is an image which explains what this PR is doing. Plugin settings fetch requests from the plugin also send the same version as any other request. Since PluginSettingsRequestProcessor was not tied to the extension and its version, adding a new version (like v3.0 in the image) would cause plugins which implement the new v3.0 version to fail, without there being a corresponding 3.0 in PluginSettingsRequestProcessor.

This PR makes that connection between the two more explicit. As in something [like this](https://github.com/gocd/gocd/pull/4113/files#diff-872b5d5179ca863e0bc9e10c46639493R56).

arvindsv:
@ketan You asked for context around this PR. The comment above could help. Let one of us know if it doesn't.

PR #4115, opened 2017-12-20
Pass along jobIdentifier as part of get-status-report request to the elastic agent plugins

GaneshSPatil:
* JobIdentifier is only passed when the request for  'status report page' made from 'build details page'.
* JobIdentifier wont be passed to the plugin when the request for 'status report page' is made from 'plugin settings page' or 'elastic profile page'.

More information: https://github.com/gocd/gocd/issues/4091#issuecomment-352979314

### DIFF
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentExtension.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentExtension.java
@@ -129,8 +129,13 @@ public class ElasticAgentExtension extends AbstractExtension {
         });
     }
 
-    public String getStatusReport(String pluginId) {
+    public String getStatusReport(String pluginId, JobIdentifier identifier) {
         return pluginRequestHelper.submitRequest(pluginId, REQUEST_STATUS_REPORT, new DefaultPluginInteractionCallback<String>() {
+            @Override
+            public String requestBody(String resolvedExtensionVersion) {
+                return getElasticAgentMessageConverter(resolvedExtensionVersion).statusReportRequestBody(identifier);
+            }
+
             @Override
             public String onSuccess(String responseBody, String resolvedExtensionVersion) {
                 final ElasticAgentExtensionConverterV2 converter = (ElasticAgentExtensionConverterV2) getElasticAgentMessageConverter(resolvedExtensionVersion);
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentMessageConverter.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentMessageConverter.java
@@ -48,4 +48,6 @@ public interface ElasticAgentMessageConverter {
     String validateRequestBody(Map<String, String> configuration);
 
     com.thoughtworks.go.plugin.domain.common.Image getImageResponseFromBody(String responseBody);
+
+    String statusReportRequestBody(JobIdentifier identifier);
 }
--- a/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/v1/ElasticAgentExtensionConverterV1.java
+++ b/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/v1/ElasticAgentExtensionConverterV1.java
@@ -103,6 +103,11 @@ public class ElasticAgentExtensionConverterV1 implements ElasticAgentMessageConv
     }
 
     @Override
+    public String statusReportRequestBody(JobIdentifier identifier) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
     public ValidationResult getValidationResultResponseFromBody(String responseBody) {
         return new JSONResultMessageHandler().toValidationResult(responseBody);
     }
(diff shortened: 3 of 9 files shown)


GaneshSPatil:
@bdpiparva @ketan @arvindsv -- can you take a look at the proposed option of *optionally passing job identifier to the status report*? 

We have used the jobIdentifier information to show _which pod is building the current job_. [here](https://github.com/gocd/kubernetes-elastic-agents/pull/9) is the PR.

plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/elastic/ElasticAgentExtension.java
```diff
@@ -129,8 +129,13 @@ public ValidationResult onSuccess(String responseBody, String resolvedExtensionV
         });
     }
 
-    public String getStatusReport(String pluginId) {
+    public String getStatusReport(String pluginId, JobIdentifier identifier) {
```
maheshp: Probably I wouldn't modify the existing status report. 
I would look at having a separate view(job status report) when on the jobs details page which would give information specific to the job with information as status of the agent, errors if any while creating an agent etc.

ketan: What does it even mean to say "get status report for a job"? I'd think a status report should show the status of all nodes in the cluster along with the pods/services and what job they map to.

I believe one of the problems is that a container may be started for a particular job, but may possibly be assigned to another job because `should-assign-work` returns true. I don't believe there's any way to know what job an elastic agent is running, unless the server **explicitly** tells the plugin about it.

You could certainly make **assumptions** about it based on the job identifiers passed to the  `create-agent` and `should-assign-work` — but they'll be just that — assumptions, and they'll be wrong in some cases...

arvindsv: Another option is to do this in JS, as the job ID will be in the URL. The page (either through JS or using the passed in identifier) can decide whether to highlight the job/pod/service by hiding everything else (a job status report) or keeping everything else intact and highlighting the job defined by the passed in ID.

arvindsv: It's true that the server is the only one that knows which GoCD agent a job is running on. Isn't it also the one that knows which elastic agent ID that GoCD agent has? If so, instead of depending on job ID, shouldn't it be sending elastic agent ID and tell the status report to highlight that?

arvindsv: Maybe worth talking about it during a call.

maheshp: Writing down what I am thinking just that I don't forget.

When I am on the Job details page I would prefer to see details of an agent in context of the specific job rather than seeing an entire Status report. If the user finds the agent information insufficient the user should be given a option to view the entire status report. 
Anyways lets talk later.

GaneshSPatil:
@arvindsv @ketan @maheshp @bhupendra -- Writing down what we spoke above "status report page".

### `Container Status Report` Endpoint for EA plugins

* Container Status Report endpoint returns a container-specific information, such as:
    - The container Configuration.
    - The GoCD Agent logs running on the container.
    - Errors/Warnings related to scheduling of the container or GoCD agent.
    - more plugin specific information. For Example, k8s Elastic Agent plugin can show the events of the pod.

* Container status report page can be Accessed from:
    - Build Details Page.
    - For Elastic Agents, on the default Agent Details Page (when clicked on agent's name from agents page), Add another (`container details`) tab to show Container Status Report, which will give more information about the elastic agent's container. (_Note:_ `Agent Details Tab` contains information about the GoCD Agent, `Container status report Tab` will contain the information about the container running the GoCD Agent).
    - Clicking on a container name from "Plugin Status Report" page should navigate to "Container Status Report" page.

* Container Status Report Page should contain links to:
    - Job Details Page of the job currently executing by the container
    - "Status Report Page"
       - In case of a cluster, link to the cluster information.
       - Link to the instance information on which current container is running.
    - Agent Details Tab too?

* For Container Status Report Page, pass along `Elastic Agent ID` instead of passing the `Job Identifier` as the context. GoCD server knows which agent is building the current job, it should be possible for the plugins to identify and locate the container based on elastic agent id.

#### Question/Need opinion?
* Data Streaming is not possible as per the current implementation of plugin status report page, this might lead to showing stale information on the status report.
Should we consider doing something similar to what we are planning for analytics plugin reports?
(we can go with the static page way for now, but going forward?)

GaneshSPatil:
As per the discussion on elastic agents status report page, we decided following:
* Do not pass `job identifier` to the `status-report` page
   - `status-report` is the simple get status report request with an empty request body.
   - `status-report` will provide the information in the context of the cluster. 

* introduce `elastic-agent-status-report` endpoint.
   - `elastic-agent-status-report` is the elastic agent specific status report request. It is intended to provide information regarding a specific elastic agent (in the context of elastic agent/container).
   - `elastic-agent-status-report` will receive the elastic agent id as part of the request body.

Closing this PR as the current changes are not relevant anymore.

PR #4117, opened 2017-12-21
PluginAssetsLoader is a service now

maheshp:
* Cached static assets path are now appended with SHA-256 fingerprints.
* Fingerprints are calculated from the actual plugin zip streams.
* Errors while caching assets are bubbled up, this would ensure plugin
  would be unloaded since the plugin would not be functional in absence
  of assets.

### DIFF
--- a/server/src/com/thoughtworks/go/server/service/plugins/PluginAssetsService.java
+++ b/server/src/com/thoughtworks/go/server/service/plugins/PluginAssetsService.java
@@ -19,34 +19,45 @@ package com.thoughtworks.go.server.service.plugins;
 import com.thoughtworks.go.plugin.access.analytics.AnalyticsExtension;
 import com.thoughtworks.go.plugin.infra.PluginChangeListener;
 import com.thoughtworks.go.plugin.infra.plugininfo.GoPluginDescriptor;
+import com.thoughtworks.go.util.ExceptionUtils;
 import com.thoughtworks.go.util.FileUtil;
 import com.thoughtworks.go.util.StringUtil;
 import com.thoughtworks.go.util.ZipUtil;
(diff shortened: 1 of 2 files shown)


server/src/com/thoughtworks/go/server/service/plugins/PluginAssetsService.java
```diff
@@ -83,20 +98,45 @@ private void cacheStaticAssets(String pluginId) {
...
             LOGGER.error("Failed to extract static assets from plugin: {}", pluginId, e);
+            ExceptionUtils.bomb(e);
+        }
+    }
+
+    private String calculateHash(byte[] data) {
+        return sha2Digest(data);
+    }
+
+    private String sha2Digest(byte[] bytes) {
```
ketan: `DigestUtils.sha256Hex(data)` will be less code :)

server/src/com/thoughtworks/go/server/service/plugins/PluginAssetsService.java
```diff
@@ -83,20 +98,45 @@ private void cacheStaticAssets(String pluginId) {
...
+
+        return null;
     }
 
     private void deleteExistingAssets(String pluginId) {
         LOGGER.info("Deleting cached static assets for plugin: {}", pluginId);
         try {
-            FileUtil.deleteDirectoryNoisily(new File(pluginStaticAssetsDirPath(pluginId)));
-        } catch (IOException e) {
+            FileUtil.deleteDirectoryNoisily(new File(pluginStaticAssetsRootDir(pluginId)));
```
ketan: `FileUtils.deleteDirectory` does the same thing..

PR #4123, opened 2017-12-21
Multiple extensions in one plugin

arvindsv:
This is an attempt to have multiple extension be part of a single plugin JAR.

### Expected behavior changes:

- No changes for existing plugins. All existing plugins should continue to work exactly as they have been.

### What this does:
- Removes `PluginDescriptorAware` class. It was a weird, deprecated class that was the basis of almost all plugin infrastructure tests. They've all been moved to use `GoPlugin`, which is now the only interface that plugins can extend using `@Extension`.

- Make `Task` not be a potential plugin extension point.

- Make `DefaultGoPluginActivator` register plugin extension class references (OSGi service references) using both plugin ID and extension type. This is key. Earlier, references [were registered ](https://github.com/gocd/gocd/blob/77d0c147c3993791d015ba61862dc24a9b1c658e/plugin-infra/go-plugin-activator/src/com/thoughtworks/go/plugin/activation/DefaultGoPluginActivator.java#L118) using only the plugin ID. So, it was not possible to differentiate between multiple extension classes, within the same plugin.

- Remove unnecessary methods from plugin infra (such as [doOnIfHasReference](https://github.com/gocd/gocd/blob/77d0c147c3993791d015ba61862dc24a9b1c658e/plugin-infra/go-plugin-infra/src/com/thoughtworks/go/plugin/infra/PluginManager.java#L38)) and make all the remaining methods be aware of extension type. Pass in extension type from all usages.

- The plugin infos API v4 has been updated to send multiple extensions. The two changes are:
    1. "type" is no longer at plugin level. It's at an extension level.
    2. "extension_info" was an attribute. It's now a list called "extensions".

Compare the [old one](https://api.gocd.org/current/#the-plugin-info-object) with the new one (below):

<details>
       <summary>New JSON of plugin infos API:</summary>

```json
{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/admin/plugin_info"
    },
    "find": {
      "href": "http://localhost:8153/go/api/admin/plugin_info/:plugin_id"
    },
    "doc": {
      "href": "https://api.gocd.org/#plugin-info"
    }
  },
  "_embedded": {
    "plugin_info": [
      {
        "_links": {
          "self": {
            "href": "http://localhost:8153/go/api/admin/plugin_info/multi-extension-plugin"
          },
          "doc": {
            "href": "https://api.gocd.org/#plugin-info"
          },
          "find": {
            "href": "http://localhost:8153/go/api/admin/plugin_info/:plugin_id"
          }
        },
        "id": "multi-extension-plugin",
        "status": {
          "state": "active"
        },
        "plugin_file_location": "/path/to/server/plugins/external/multi-extension-plugin-1.0.jar",
        "bundled_plugin": false,
        "about": {
          "name": "Multi Extension Plugin",
          "version": "0.0.1",
          "target_go_version": "17.12.0",
          "description": "Can a plugin implement multiple extension points?",
          "target_operating_systems": [

          ],
          "vendor": {
            "name": "GoCD team",
            "url": "https://www.gocd.org"
          }
        },
        "extensions": [
          {
            "type": "notification",
            "plugin_settings": {
              "configurations": [
                {
                  "key": "url",
                  "metadata": {
                    "secure": false,
                    "required": false
                  }
                },
                {
                  "key": "user",
                  "metadata": {
                    "secure": false,
                    "required": false
                  }
                }
              ],
              "view": {
                "template": "<div>some html</div>"
              }
            }
          },
          {
            "type": "task",
            "display_name": "MultiPluginTask",
            "task_settings": {
              "configurations": [
                {
                  "key": "url",
                  "metadata": {
                    "secure": false,
                    "required": false
                  }
                },
                {
                  "key": "user",
                  "metadata": {
                    "secure": false,
                    "required": false
                  }
                }
              ],
              "view": {
                "template": "<div class=\"form_item_block\">\n  <label>Username:<span class='asterix'>*</span></label>\n  <input ng-model=\"user\" />\n</div>\n"
              }
            }
          }
        ]
      }
    ]
  }
}
```
</details>

### Todos

- Update [API documentation](https://github.com/gocd/api.go.cd/blob/601af9c9a2fc24431270bbd5bfb975931560f7eb/source/includes/plugin_info/_000-describe.md.erb).

- ~~Scenario: Consider how to properly handle a plugin with one "bad" extension and one good extension. Mark the plugin itself as "bad". However, it will already be loaded into metadata loaders.~~ [Fixed: [here](https://github.com/gocd/gocd/pull/4123/commits/0f4bc13653490b2d72ff1a54cd017bf03d9142cf)]

- ~~There are some usages such as [this](https://github.com/gocd/gocd/blob/11a92de581b706cac7069bc1f87af99fef50dfc2/server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/pluginsettings/PluginSettingsRequestProcessor.java#L89-L98), which will fail. That is because, if an extension can handle a plugin, it does not mean that it can handle the validate settings message.~~ [Fixed: Using the extension sent as a part of the request]

- ~~Scenario: A plugin has two extensions, both of which implement one kind of extension. What happens? It'll fail [here](https://github.com/arvindsv/gocd/blob/51f92da80fb1d738bfe9b81932a7b279c4f94aed/plugin-infra/go-plugin-infra/src/main/java/com/thoughtworks/go/plugin/infra/FelixGoPluginOSGiFramework.java#L242) just like earlier. Is that good enough?~~ [I think it's fine]

- Show multiple extensions in a plugin properly in the plugins page:

  ~~The UI pages still assume that all plugins only have one extension, and show only the first one. [This example plugin which implements multiple extension points](https://github.com/arvindsv/gocd-multi-extension-plugin-example/tree/943ef174ee25b97a5e9c2ab0d60f5b69c98b43ca) is the one shown in the image above. One idea is to replace make it more obvious that it contains multiple extensions. Note that one or more of those extensions can have configurations. It's ideal if only one of them has it - but it is not enforced.~~ This is now enforced. Only one extension in the plugin can support configuration save/validate, etc. If more than one extension has that, you'll see a message like this:

- If a plugin fails to provide a valid PluginIdentifier (say it throws an exception), you'll see a message like this:

image

### For review:

Commit: [Remove PluginDescriptorAware.](https://github.com/gocd/gocd/pull/4123/commits/0ee49a35f3a80303f6b13190461b4dc219fa1468)

  * It was deprecated.

  * Lots of plugin infra tests were written around it.

  * Move all of those to use GoPlugin.

Comments:

  - The commit message explains it. All tests are rewritten to use GoPlugin instead. The biggest [change in code](https://github.com/gocd/gocd/pull/4123/commits/0ee49a35f3a80303f6b13190461b4dc219fa1468#diff-957d63833636f69d7b1c1bd3da67835cL146) was around here. That functionality is removed.
    

<hr>

Commit: [Remove GoPluginApiMarker from Task.](https://github.com/gocd/gocd/pull/4123/commits/88e92522fde4e75b2a9263808f13f82bc2c8e514)

* Task should no longer be a candidate for a plugin to use it as an extension.

Comments:

  - Nothing significant.
  

<hr>

Commit: [Use extension type when registering plugin.](https://github.com/gocd/gocd/pull/4123/commits/c1b6f7fc19e0bc9857bc783e18b37e8e66072c0a)

  * Plugin extension classes were registered against the plugin ID only.

  * Now, they're registered using both plugin ID and extension type, to allow having multiple extension classes of different types in a single plugin.

  * Regenerate test jars which use the activator.
  
Comments:

  - The main change of this whole PR is [in the activator](https://github.com/gocd/gocd/pull/4123/commits/c1b6f7fc19e0bc9857bc783e18b37e8e66072c0a#diff-9b2d585dbaa6d8d5cdcfe25f2e705206) where every individual extension's OSGi service reference is registered against its extension type. The Constants.BUNDLE_CATEGORY constant is used for this purpose. It was chosen quite arbitrarily, but it feels ok to me. The [tests](https://github.com/gocd/gocd/pull/4123/commits/c1b6f7fc19e0bc9857bc783e18b37e8e66072c0a#diff-512f6125823be4a20f0c3d192f50e16d) explain this. This is the first time the activator calls a method in one of the classes it instantiates.
    

<hr>

Commit: [Regenerate test plugins](https://github.com/gocd/gocd/pull/4123/commits/a73f8a73a82c197a7a348f5559a6b504b21c9107)

  * Activator has changed and PluginDescriptorAware is gone.

Comments:

  - All the plugins used in the integration tests were regenerated and checked in, after some changes to their build script.

<hr>

Commit: [Remove methods used only in tests.](https://github.com/gocd/gocd/pull/4123/commits/555079742c4ae3b124b0db2dd504216de16fef10)

Comments:

  - Nothing significant. A *lot* of methods in the [GoPluginOSGiFramework](https://github.com/gocd/gocd/pull/4123/commits/555079742c4ae3b124b0db2dd504216de16fef10#diff-6e5ca830071b5be8798a611a26917b40) interface were unnecessary.

<hr>

Commit: [Remove ability to filter change events.](https://github.com/gocd/gocd/pull/4123/commits/240f9ed387538f3027ea6d750b1ab8950865cf93)

  * Plugins can only be of type GoPlugin.class. There is no point in allowing registering listeners for other types.
  
Comments:

  - It was assumed that there could be multiple classes of type GoPluginMarker. That was true when we started. It's no longer the case. We have GoPlugin as the only extensible class. So, there's no point waiting on those other classes. So, that filtering [was removed](https://github.com/gocd/gocd/pull/4123/commits/240f9ed387538f3027ea6d750b1ab8950865cf93#diff-a287701eed50f83f4d530f1bde187b84L134).
  

<hr>

Commit: [Make hasReferenceFor aware of extension type.](https://github.com/gocd/gocd/pull/4123/commits/67fe8672408d5b17b87590b010e08e3322064334)

Comments:
  
  - Now that the extensions in a plugin have been registered with extension type, we can differentiate between the service references. However, that implies that the calls themselves tell us not just which plugin they want to send a message to (or check), but *which extension* in that plugin they want to address. This commit begins that by making sure that all calls to `hasReferenceFor` pass in the extension type.
  
  - [This is where](https://github.com/gocd/gocd/pull/4123/commits/67fe8672408d5b17b87590b010e08e3322064334#diff-1ae84b0623746bc9743efec07e22f907R236) the extension type registered in the activator is used.

<hr>

Commit: [Make resolveExtensionVersion aware of extension type.](https://github.com/gocd/gocd/pull/4123/commits/33cd0d05a7c5d6536281135c6a1b7b1b148af1e5)

Comments:

  - Similar to the other commit which changed hasReferenceFor, this changes all usages of resolveExtensionVersion to include the extension type. The [main change is here](https://github.com/gocd/gocd/pull/4123/commits/33cd0d05a7c5d6536281135c6a1b7b1b148af1e5#diff-a287701eed50f83f4d530f1bde187b84). Since `resolveExtensionVersion` uses `doOn`, which has not been changed yet, there's no change in the body of the method yet.
  
  - Since most usages are from the corresponding extension classes (like TaskExtension, ConfigRepoExtension, etc), it's easy to know what extension type to pass to each call. PluginRequestHelper was also used by multiple extensions, and the extension type [was available](https://github.com/gocd/gocd/pull/4123/commits/33cd0d05a7c5d6536281135c6a1b7b1b148af1e5#diff-d7d606f53ab18a8b566c72fe556d4c85R43) there as well.
  

<hr>

Commit: [Make pluginManager.submitTo aware of extension type.](https://github.com/gocd/gocd/pull/4123/commits/1328b02a59b8da16571bb92bd9f8db5b0dfc15ea)

Comments:

  - Very similar to the other commits like hasReferenceFor and resolveExtensionVersion. The [main change is here](https://github.com/gocd/gocd/pull/4123/commits/1328b02a59b8da16571bb92bd9f8db5b0dfc15ea#diff-a287701eed50f83f4d530f1bde187b84). The rest are mostly fixing the tests.
  

<hr>

Commit: [Remove pluginManager#doOn.](https://github.com/gocd/gocd/pull/4123/commits/57c9a2ba3cdc1a8f9530d1e4f44db457b513ece8)

Comments:

  - Nothing significant.

<hr>

Commit: [Make osgiFramework#doOn aware of extension type.](https://github.com/gocd/gocd/pull/4123/commits/f1e3528be13346d6187f8c4569fd6861100ab1c6)

Comments:

  - Finally the [OSGi framework](https://github.com/gocd/gocd/pull/4123/commits/f1e3528be13346d6187f8c4569fd6861100ab1c6#diff-1ae84b0623746bc9743efec07e22f907) is passed the extension type, which it can use to find the right service reference among multiple service references (extensions) in the plugin. The [old method](https://github.com/gocd/gocd/pull/4123/commits/f1e3528be13346d6187f8c4569fd6861100ab1c6#diff-1ae84b0623746bc9743efec07e22f907L230) which was just using the plugin ID is removed.
  

<hr>

Commit: [Plugin infos APIs aware of extensions.](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1)

  * Plugin Infos APIs v3 and v4 are now aware of the possibility of multiple extensions.
  
  * v3 is unchanged (it assumes that a plugin can only have one extension).
  
  * v4 changes "extension_info" to "extensions" and supports multiple extensions.
  
Comments:
  
  - New class "CombinedPluginInfo" [is introduced](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-7959c6157349c08d1ef1fdf352b34173). It can contain multiple "PluginInfo" objects (usually, implementers of "AbstractPluginInfo", such as SCMPluginInfo and AnalyticsPluginInfo).
    
  - Ideally all these "PluginInfo" objects should be called "ExtensionInfo", but I felt that it'll make this change seem much bigger than it is and didn't do it.

  - [BadPluginInfo](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-fbb6998fbb2de4496d5f2050a134c669) is also a CombinedPluginInfo now.

  - CombinedPluginInfo now knows [how to get the "image"](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-7959c6157349c08d1ef1fdf352b34173R67) of a plugin. This behavior was earlier in the [DefaultPluginInfoFinder](https://github.com/gocd/gocd/blob/cba9fd72860ec1858d63ea27a8251275da6b1fcd/server/src/main/java/com/thoughtworks/go/server/service/plugins/builder/DefaultPluginInfoFinder.java#L103). That seemed like a weird place for it. It was also doing something around verifying that the hash matches. That seemed like something the controller should do. So, it has been moved [into that controller](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-417693e10f52fd5441e57ca8d3619b5d).
    
  - The most significant change in this commit is probably [in DefaultPluginInfoFinder](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-3f308b1b89975c08375833b8ebb3571c). It's not a big change. Just looks like that because of Java generics and the inability to use lambdas because of our old Spring code. It just groups multiple PluginInfo objects by plugin ID, and puts them into a CombinedPluginInfo object. If a plugin has only one extension, then the CombinedPluginInfo will only have one PluginInfo. The [tests](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-b77804909aaa6992a1ca55b513c18b38) describe this behavior.
    
  - The changes made to the plugin infos controllers [for v3](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-4765218b4ee322ef817ea3ab3fdf34e1) and [for v4](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-7d3a64b5c7cbc42da14a64d7ce686a30) are identical. They just make it use CombinedPluginInfo which is what DefaultPluginInfoFinder now returns.
    
  - The [representer for v3](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-4f966b7082b639fb0b3b99f0fc79b894) was changed to expect a CombinedPluginInfo but its external interface is still the same. It assumes that there is only one extension in the plugin and continues to use "extension_info" just like before. You can see that there's no change in the output, [in the v3 specs](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-828ee0ae135c2bf0089e5a9439b5049c).
  
  - The [representer for v4](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-d7d8ae58acfae152c4bc61b14e4dad6eR74) is changed to have a collection called "extension", instead of "extension_info". The contents are nearly identical to earlier, except that ["type" has been moved](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-8ed09188777e669d317981ded1e7d283) into the extension level. You can see these changes represented [in the v4 specs](https://github.com/gocd/gocd/pull/4123/commits/0c2024796a5e4a963e63c017aaa5ed4025ecdabe?w=1#diff-dfba9137ca232bc3ab08806ec3e9978a).

<hr>

Commit: [Make plugins page work with new APIs.](https://github.com/gocd/gocd/pull/4123/commits/ff66af68d61929b0985d8b6d78a463269ef412d1?w=1)

  * Plugins page now expects the output from the new API ("extensions" as an array instead of "extension_info").
    
  * It still does not understand multiple extensions in a plugin. It assumes there is only one and uses it (`extensions[0]`).

Comments:

  - Most changes are in specs, like [here](https://github.com/gocd/gocd/pull/4123/commits/ff66af68d61929b0985d8b6d78a463269ef412d1?w=1#diff-70c3891b4d10bad3ffe0317abfa666c1R163) and [here](https://github.com/gocd/gocd/pull/4123/commits/ff66af68d61929b0985d8b6d78a463269ef412d1?w=1#diff-70c3891b4d10bad3ffe0317abfa666c1R187), where the expected object is changed to have "extensions" instead of "extension_info" and "type" is moved into the extension level.
  
  - The most significant code change is [in models/shared/plugin_infos.js](https://github.com/gocd/gocd/pull/4123/commits/ff66af68d61929b0985d8b6d78a463269ef412d1?w=1#diff-e1b4b5ebc615227cdd10d6eb42b7b12c). All "extension_info" usages have been changed to "extensions[0]". One usage of "data.type" is changed to "data.extensions[0].type". So, this is how the assumption of a plugin having one extension is enforced.
  

<hr>

Commit: [Make plugin settings API work with combined plugins](https://github.com/gocd/gocd/pull/4123/commits/483be88654c55c5cc5dd69cdd8aa70a14193fe14)

* Currently, assumes that there is only one extension in the plugin.

Comments:

  - Nothing significant. Like other places, it [assumes there is only one extension](https://github.com/gocd/gocd/pull/4123/commits/483be88654c55c5cc5dd69cdd8aa70a14193fe14#diff-c643ecb67470f48ca36fe62b00ef061aR60), for now. Will need to be changed to support multiple plugins.

<hr>

Commit: [Ensure only one extension has plugin settings](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1)

* When multiple extensions are present in a plugin, only one of them can have settings.

* Keep track of which extension handles settings, and send further calls (validation of settings) to that extension in the plugin.

* When a plugin fails to load during pluginLoad listener callback, send a pluginUnloaded notification to all listeners.

Comments:

  - [PluginSettingsMetadataLoader](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1#diff-949cc70e2a570690357085187c3b40e2R67) checks that only one extension responds to the settings request. If not, it fails. Tests are [here](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1#diff-15145cb7dc1ad168a4a2be05a12b204dR151).

  - There's a [reasonably big change](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1#diff-1ae84b0623746bc9743efec07e22f907R117) in what happens when a plugin listener fails. Earlier the plugin was marked invalid (in the plugin registry) and it was unloaded. But, all of the listeners were not sent a plugin unloaded event. So, especially for metadata stores, they would continue to have information about a plugin (such as settings view template) and would continue to show them in the UI and elsewhere. Now, all of them are sent an unloaded event and they are supposed to handle it and remove data they inserted into their own stores.

  - [PluginService](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1#diff-7da686785c37f232a02c70711082f0c6R108) and other places now use the extension that was decided earlier as the one that handles settings, when validating settings, etc.

<hr>

Commit: [Rename method getPluginSettingsFor.](https://github.com/gocd/gocd/pull/4123/commits/9066d9d2d6fd7387e973c6fbb2a0bcf59379a308?w=1)

* Just a rename: getPluginSettingsFor => createPluginSettingsFrom.

Comments:

  - Nothing significant.

<hr>

Commit: [Rename method getPluginSettingsFor.](https://github.com/gocd/gocd/pull/4123/commits/f7b60ab4014cc2ffe7288fe23637f901fba28607?w=1)

* Just a rename: getPluginSettingsFor => loadPluginSettingsFor.

Comments:

  - Nothing significant. This was the second method named that way.

<hr>

Commit: [Rename method: getPluginSettings.](https://github.com/gocd/gocd/pull/4123/commits/d64ce276907b3e567d9f7fe5c55e12b379805f60?w=1)

* Just a rename: getPluginSettings => loadStoredPluginSettings.

Comments:

  - Nothing significant.

<hr>

Commit: [Rename method: createPluginSettings.](https://github.com/gocd/gocd/pull/4123/commits/2b9aa54270cf9c05944652acea53469153ae7a2b?w=1)

* Just a rename: createPluginSettings => savePluginSettings.

Comments:

  - Nothing significant.

<hr>

Commit: [Refactor to make it clearer.](https://github.com/gocd/gocd/pull/4123/commits/3430278040e8b4c857ad1b8b0c0e963164498efd?w=1)

Comments:

  - Nothing significant.

<hr>

Commit: [Notify correct extension upon settings change](https://github.com/gocd/gocd/pull/4123/commits/47fd37bdd04eef8728fa9f78737c2de369c0e553?w=1)

* Only the extension which handles settings related messages gets a notification about plugin settings change.

Comments:

  - Earlier, all extensions in a plugin were sent a notification if plugin settings changed. Now, only the extension which handles settings [gets that notification](https://github.com/gocd/gocd/pull/4123/commits/47fd37bdd04eef8728fa9f78737c2de369c0e553?w=1#diff-7da686785c37f232a02c70711082f0c6L157). Pretty simple reuse of the changes in [the earlier commit](https://github.com/gocd/gocd/pull/4123/commits/f229447586a8fd17beba7dc8c0c83a4099bf7fbc?w=1).

<hr>

Commit: [Handle server info requests properly.](https://github.com/gocd/gocd/pull/4123/commits/0f25a3f56db3065f8d4eed30016edde8e584dc16?w=1)

* In a multi-extension plugin, when a server info request arrives, it needs to be routed to the right extension, based on the extension that the plugin expects to send it to.

Comments:

  - Main change is [here](https://github.com/gocd/gocd/pull/4123/commits/0f25a3f56db3065f8d4eed30016edde8e584dc16?w=1#diff-893c3e7e84a51401a4e0e73f7c9cfc8f). Uses the extension name from the identifier to decide which extension to use in the plugin, to parse the message.

<hr>

Commit: [Handle plugin settings requests properly.](https://github.com/gocd/gocd/pull/4123/commits/99f8ca501f40e1925d6992af8080c728ab2b5bd6?w=1)

* In a multi-extension plugin, when a plugin settings request arrives, it needs to be routed to the right extension, based on the extension that the plugin expects to send it to.

Comments:

  - Main change is [here](https://github.com/gocd/gocd/pull/4123/commits/99f8ca501f40e1925d6992af8080c728ab2b5bd6?w=1#diff-e53248d91fef93e3663b3f2ff2758338). Uses the extension name from the identifier to decide which extension to use in the plugin, to parse the message.

<hr>

Commit: [Refactor: Sort properties in JS object.](https://github.com/gocd/gocd/pull/4123/commits/1c3b3579d93a1f89b1fc09039f27d9cda0fb7bc8?w=1)

* Sort some properties in alphabetical order to make it easier to modify the structure of the plugin info JS object.

Comments:

  - Nothing significant.

<hr>

Commit: [Refactor: Add missing data to JS object.](https://github.com/gocd/gocd/pull/4123/commits/89cbe0eb950566d774848a3c91108a19a7597857?w=1)

* Make all plugin info JS objects similar by adding properties which are not present and making them null. For instance, adding "pluginSettings: null" to task plugin info.

Comments:

  - Nothing significant.

<hr>

Commit: [Handle multiple extension in plugin info.](https://github.com/gocd/gocd/pull/4123/commits/3afc76c21c8fdafa50dee7085d4c18270e61da00?w=1)

<details>
  <summary>Details</summary>
  
Earlier PluginInfos.PluginInfo JS object looked like this:
```
{
  about: ...,
  bundledPlugin: ...,
  id: ...,
  imageUrl: ...,
  pluginFileLocation: ...,
  pluginSettings: ... (or null),
  status: ...,
  type: "scm",

  scmSpecificProperty1: ...,
  scmSpecificProperty2: ...,
}
```

Now, it looks like this:
```
{
  about: ...,
  bundledPlugin: ...,
  id: ...,
  imageUrl: ...,
  pluginFileLocation: ...,
  pluginSettings: ... (or null),
  status: ...,
  types: ["scm", "elastic-agent"],

  scm: {
    scmSpecificProperty1: ...,
  },
  elastic-agent: {
    elasticAgentSpecificProperty1: ...,
    elasticAgentSpecificProperty2: ...,
  },
}
```
</details>

Comments:

  - The [spec changes](https://github.com/gocd/gocd/pull/4123/commits/3afc76c21c8fdafa50dee7085d4c18270e61da00?w=1#diff-70c3891b4d10bad3ffe0317abfa666c1) might look like a lot, but they aren't. It's just that the JSON objects for the specs have moved to the top of the file. The only real change is a spec to verify the [multi extension plugin](https://github.com/gocd/gocd/pull/4123/commits/3afc76c21c8fdafa50dee7085d4c18270e61da00?w=1#diff-70c3891b4d10bad3ffe0317abfa666c1R867).

  - The main change is in [plugin_infos.js](https://github.com/gocd/gocd/pull/4123/commits/3afc76c21c8fdafa50dee7085d4c18270e61da00?w=1#diff-e1b4b5ebc615227cdd10d6eb42b7b12c), which now understands a plugin with multiple extensions and handles extension-specific settings properly. It seems easiest to review this using the [split diff](https://github.com/gocd/gocd/pull/4123/commits/3afc76c21c8fdafa50dee7085d4c18270e61da00?diff=split&w=1#diff-e1b4b5ebc615227cdd10d6eb42b7b12c). The most important change is [around here](https://github.com/arvindsv/gocd/blob/3afc76c21c8fdafa50dee7085d4c18270e61da00/server/webapp/WEB-INF/rails.new/webpack/models/shared/plugin_infos.js#L95-L113).

<hr>

Commit: [Refactor: Fix typo.](https://github.com/gocd/gocd/pull/4123/commits/cb9f6943283daaa6a7a4878fa2cc67b7d815c91a?w=1)

Comments:

  - Nothing significant.


arvindsv:
@ketan - FYI: My current thinking is that it'll be easier to keep this PR to what exists right now. We know of some cases where multiple extensions in a plugin will not work properly (UI is one such place). But, most changes needed to support it is in place - and this continues to work with single extension plugins. Also, plugin infos API v4 is changed.

So, I'll work on a separate PR for the todos, unless you think differently.

@ankitsri11 has been working on verifying that this continues to work with each kind of plugin, and he has some API calls which create pipelines, etc. to prove that this works. We will attach that tomorrow, to this.

arvindsv:
@ketan @ankitsri11 found a bug around plugin settings API that I'll need to fix.

arvindsv:
@maheshp @jyotisingh I've rebased it, but since there were some changes in this area, I'll need to go through all the commits and make sure nothing is missed.

arvindsv:
@jyotisingh You need to take a look at:

https://github.com/gocd/gocd/pull/4123/commits/fb43e97cae9a564d9e1329b5a069da13248e08d0
https://github.com/gocd/gocd/pull/4123/commits/e970a97eeee1ac84686d3a380a9bb98f736c7433

and if you want:
https://github.com/gocd/gocd/pull/4123/commits/504557928b55d47ba6ba264c32f5862d3a072f83

Those are the commits that have changed since you saw them last and the changes we talked about are made there.

PR #4124, opened 2017-12-21
PluginAssetsService is a metadata change listener

maheshp:
* Renamed PluginAssetsService to AnalyticsPluginAssetsService since it
  is specific to Analytics Plugin.
* AnalyticsPluginAssetsService is a metadata load listener, once a
  analytics plugin_info is built the static assets are cached in GoCD.
* AssetsService updates the analytics plugin_info with the assets path.
* Cleaned up usages of `FileUtil` and `StringUtil`

### DIFF
--- a/plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/analytics/AnalyticsMetadataStore.java
+++ b/plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/analytics/AnalyticsMetadataStore.java
@@ -34,4 +34,10 @@ public class AnalyticsMetadataStore extends MetadataStore<AnalyticsPluginInfo> {
         return store;
     }
 
+    public void updateAssetsPath(String pluginId, String assetPath) {
+        AnalyticsPluginInfo pluginInfo = getPluginInfo(pluginId);
+        if (pluginInfo != null) {
+            pluginInfo.setStaticAssetsPath(assetPath);
+        }
+    }
 }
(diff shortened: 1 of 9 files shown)


plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/analytics/AnalyticsMetadataStore.java
```diff
@@ -34,4 +34,10 @@ public static AnalyticsMetadataStore instance() {
         return store;
     }
 
+    public void updateAssetsPath(String pluginId, String assetPath) {
+        AnalyticsPluginInfo pluginInfo = getPluginInfo(pluginId);
+        if (pluginInfo != null) {
```
arvindsv: Isn't it fatal if pluginInfo is null? There's only one chance to set static assets. I'd think we would rather not silently fail.

arvindsv:
Context: We need to load and cache static assets from the analytics plugin, once it is loaded. Before this PR, the object that does that ([AnalyticsPluginAssetsService](https://github.com/gocd/gocd/pull/4124/files#diff-a5423a1ea01ee624bf17fa7d057ae07cR45)) was a PluginChangeListener. Problem is that the object that loads the plugin ([MetadataLoader](https://github.com/gocd/gocd/blob/21c89b7f251b1cba6fbc28b7d5981d0ce6f877da/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/common/MetadataLoader.java#L39)) is also a PluginChangeListener. So, it's possible that a plugin and its metadata is not yet loaded, when AnalyticsPluginAssetsService tries to ask it for its static resources. So, the order between the two needs to be consistent.

PluginMetadataChangeListener is new.  It's a listener which will be called back by the MetadataLoader, once the plugin metadata is loaded. Then, the static assets can be cached by the AnalyticsPluginAssetsService. Order is guaranteed.

plugin-infra/plugin-metadata-store/src/com/thoughtworks/go/plugin/access/analytics/AnalyticsMetadataStore.java
maheshp: Well theoretically this should never happen, but you are right we as well not silently fail if this happens.

Issue #3828, opened 2017-09-06
Plugin remains in a block state when multiple agents are having same `elasticAgentId`


bdpiparva:
##### Issue Type

- Bug Report

##### Summary

Plugin remains in a block state when multiple agents are having same `elasticAgentId`

As part of `ServerPing` plugin makes a disable agent call to server with list of elastic agent ids. In case of multiple agents are having same UUID, server fails to disable the agents and throws an `IllegalStateException`. This will prevent plugin to delete old container and plugin can not start new containers.

### Possible fix:

1. Put check on agent registration, if agent with same `elasticAgentId` exist then do not register the new agent.
2. Ignore the agents with duplicate `elasticAgentId` and disable the other agents.

### Action needed at plugin

Make sure plugin always creates container with `unique id`

GaneshSPatil:
@jyotisingh -- Yes. There won't be multiple agents with the same elasticAgentId. So this issue won't happen.

PR #4133, opened 2017-12-27
Do not allow agent to register with duplicate elastic agent id (#3828)

GaneshSPatil:
[Agent Registration] Do not allow agent registration if agent with same elastic agent id already exists
[XSL Migration] Add a xslt transformation to remove an agent with duplicate elastic agent id.

### DIFF
--- a/config/config-server/resources/upgrades/103.xsl
+++ b/config/config-server/resources/upgrades/103.xsl
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<!--
+  ~ Copyright 2017 ThoughtWorks, Inc.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
+    <xsl:template match="/cruise/@schemaVersion">
+        <xsl:attribute name="schemaVersion">103</xsl:attribute>
+    </xsl:template>
+    <!-- Copy everything -->
+    <xsl:template match="@*|node()">
+        <xsl:copy>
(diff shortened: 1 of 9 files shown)


config/config-server/resources/upgrades/103.xsl
```diff
@@ -0,0 +1,30 @@
...
+        <xsl:attribute name="schemaVersion">103</xsl:attribute>
+    </xsl:template>
+    <!-- Copy everything -->
+    <xsl:template match="@*|node()">
+        <xsl:copy>
+            <xsl:apply-templates select="@*|node()"/>
+        </xsl:copy>
+    </xsl:template>
+
+    <xsl:template match="agent[@elasticAgentId = preceding-sibling::agent/@elasticAgentId]"/>
```
ketan: What if the preceeding sibling is not the dup, but some other sibling is?

bdpiparva: @ketan --- It think will check for all siblings. see http://xsltransform.net/3MEbY6h/3.

server/test/integration/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1854,6 +1854,22 @@ public void shouldSkipParamResoulutionForElasticConfig_asPartOf100To101Migration
         );
     }
 
+
+    @Test
+    public void shouldRemoveAgentWithDuplicateElasticAgentId_asPartOf102To103Migration() throws NoSuchMethodException, IllegalAccessException, InvocationTargetException {
+        String configXml = "<cruise schemaVersion='102'>" +
+                "<agents>\n" +
+                "    <agent hostname=\"hostname\" ipaddress=\"127.0.0.1\" uuid=\"c46a08a7-921c-4e77-b748-6128975a3e7d\" elasticAgentId=\"16649813-4cb3-4682-8702-8e202824dd73\" elasticPluginId=\"elastic-plugin-id\" />\n" +
```
bdpiparva: @GaneshSPatil --- Change the order of duplicate agents. Just to make sure that translation will remove duplicates in any order.

server/src/com/thoughtworks/go/server/controller/AgentRegistrationController.java
```diff
@@ -233,6 +233,10 @@ public ResponseEntity agentRequest(@RequestParam("hostname") String hostname,
                 return new ResponseEntity<>("Elastic agents must submit both elasticAgentId and elasticPluginId.", UNPROCESSABLE_ENTITY);
             }
 
+            if (elasticAgentIdAlreadyRegistered(elasticAgentId, elasticPluginId)) {
```
bdpiparva: Can we move this check inside [AgentConfig#validate](https://github.com/gocd/gocd/blob/master/config/config-api/src/com/thoughtworks/go/config/AgentConfig.java#L83-L111) ?

PR #4141, opened 2018-01-01
Moving DatabaseConfiguratin to :addon-api:database module

ashwanthkumar:
As discussed with @ketan on go-cd-dev mailing list.


ketan:
On further thought, it looks like `DatabaseConfiguration` should be an internal implementation of h2, and let other `Database` implementations do what they like.

`DatabaseConfiguration#loadConfiguration` does not exactly seem like API worthy. I'll make that change on another PR and close this one out.

ketan:
@ashwanthkumar — see 4c8eed1ba406f214b50580ef51232113ee204b71 on master.

ashwanthkumar:
Thanks @ketan this is even better.

PR #4146, opened 2018-01-02
Utilize static assets path for the analytics plugins

ibnc:
* Specifically as the root for the analytics data view path

### DIFF
--- a/plugin-infra/go-plugin-access/test/com/thoughtworks/go/plugin/access/analytics/AnalyticsExtensionTest.java
+++ b/plugin-infra/go-plugin-access/test/com/thoughtworks/go/plugin/access/analytics/AnalyticsExtensionTest.java
@@ -16,12 +16,12 @@
 
 package com.thoughtworks.go.plugin.access.analytics;
 
-import com.thoughtworks.go.plugin.api.info.PluginDescriptor;
 import com.thoughtworks.go.plugin.api.request.GoPluginApiRequest;
 import com.thoughtworks.go.plugin.api.response.DefaultGoPluginApiResponse;
 import com.thoughtworks.go.plugin.domain.analytics.AnalyticsData;
 import com.thoughtworks.go.plugin.domain.analytics.AnalyticsPluginInfo;
 import com.thoughtworks.go.plugin.infra.PluginManager;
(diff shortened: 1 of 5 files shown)


ibnc:
@maheshp Would you mind taking a look when you get a chance?

plugin-infra/go-plugin-domain/src/com/thoughtworks/go/plugin/domain/analytics/AnalyticsData.java
```diff
@@ -36,11 +38,22 @@ public String getViewPath() {
         return viewPath;
     }
 
+    public String getFullViewPath() {
+        if (null == assetRoot || "".equals(assetRoot)) {
```
maheshp: Use `StringUtils.isBlank`

plugin-infra/go-plugin-access/test/com/thoughtworks/go/plugin/access/analytics/AnalyticsExtensionTest.java
```diff
@@ -78,13 +90,23 @@ public void shouldTalkToPlugin_To_GetCapabilities() throws Exception {
     public void shouldTalkToPlugin_To_GetPipelineAnalytics() throws Exception {
         String responseBody = "{ \"view_path\": \"path/to/view\", \"data\": \"{}\" }";
         when(pluginManager.submitTo(eq(PLUGIN_ID), requestArgumentCaptor.capture())).thenReturn(new DefaultGoPluginApiResponse(SUCCESS_RESPONSE_CODE, responseBody));
+        AnalyticsPluginInfo pluginInfo = mock(AnalyticsPluginInfo.class);
```
arvindsv: This seems like an unnecessary mock of a simple class. Can you consider this:

```java
AnalyticsPluginInfo pluginInfo = new AnalyticsPluginInfo(
    new GoPluginDescriptor(PLUGIN_ID, null, null, null, null, false),
    null, null, null);
    pluginInfo.setStaticAssetsPath("/assets/root");
    metadataStore.setPluginInfo(pluginInfo);
```

Ideally we should have a PluginInfoMother.

PR #4147, opened 2018-01-02
Display Pipeline Analytics Icon If Supported

ibnc:
The icon will only display if an available analytics plugin supports pipeline analytics.

Note: This only works for ERB pages. Another PR will be created to support mithril SPAs.

@maheshp Mind taking a look when you get a chance?


maheshp:
Have we decided on showing the icon on the dashboard? Having this icon at a pipeline instance level might not seem right since this is not a instance level information.

What happens when there are multiple plugins which support pipeline analytics?

arvindsv:
I think it's reasonable to have it somewhere on the dashboard, at least for now. You're right that it can be moved to the pipeline level, rather than the pipeline instance level.

For now, if there are multiple plugins, it's ok to show only the first one. We can later decide what to do in such a case. It'll need some input from the design team as well.

server/webapp/WEB-INF/rails.new/lib/analytics_support.rb
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+##########################################################################
+
+module AnalyticsSupport
+  include JavaImports
+
+  def with_pipeline_analytics_support(&block)
+    return unless block_given?
+
+    default_plugin_info_finder.allPluginInfos(PluginConstants.ANALYTICS_EXTENSION).each do |plugin|
```
maheshp: Wondering if this method should be moved to helper. 

You could as well use  `AnalyticsMetadataStore#allPluginInfos()` to go over analytics plugin_infos over using the `default_plugin_info_finder`.

maheshp:
Ok @arvindsv lets go with an icon on the dashboard for now.

server/webapp/WEB-INF/rails.new/lib/analytics_support.rb
arvindsv: Yes. Should be a helper, I guess.

They talked about it. It's marked deprecated, and it looked like default_plugin_info_finder was starting to be a place to get info about plugins. No? We thought you would know best. So, you decide.

maheshp: I think in this case we should use the `AnalyticsMetadataStore` since we are dealing only with the Analytics Plugin.

ibnc: @maheshp @arvindsv The analytics_support file gets included in the application helper. Me and @marques-work figured that the analytics helper methods should be accessible from the everywhere. We created a module to include just for organizational reasons. 

That being said, we can totally move the methods into the application_helper directly if y'all would prefer

ibnc: @maheshp Never mind. Me and @arvindsv talked about it. I'm moving the method into the pipelines helper

ibnc:
@maheshp I've made the changes you suggested. Let me know if anything else needs to change

PR #4155, opened 2018-01-04
Introduce iframe window messaging endpoint helper

marques-work:
Added ExtensionIFrameEndpoint helper object to:

- ensure listener is established in parent window
- handle dispatch of messages based on keys
- simple wrappers to send and reply to key-based messages

Refactored analytics modal script to use this.

@maheshp could you review this when you get a chance? thanks!

### DIFF
--- a/development-utility/development-server/src/main/java/com/thoughtworks/go/server/DevelopmentServer.java
+++ b/development-utility/development-server/src/main/java/com/thoughtworks/go/server/DevelopmentServer.java
@@ -40,6 +40,7 @@ public class DevelopmentServer {
         LogConfigurator logConfigurator = new LogConfigurator(DEFAULT_LOGBACK_CONFIGURATION_FILE);
         logConfigurator.initialize();
         copyDbFiles();
+        copyPluginAssets();
         File webApp = new File("webapp");
         if (!webApp.exists()) {
             throw new RuntimeException("No webapp found in " + webApp.getAbsolutePath());
@@ -96,6 +97,11 @@ public class DevelopmentServer {
         if (!new File("db/h2db/cruise.h2.db").exists()) {
             FileUtils.copyDirectoryToDirectory(new File("db/dbtemplate/h2db"), new File("db/"));
         }
+    }
+
+    private static void copyPluginAssets() throws IOException {
+        File classPathRoot = new File(DevelopmentServer.class.getProtectionDomain().getCodeSource().getLocation().getPath());
+        FileUtils.copyFile(new File("webapp/WEB-INF/rails.new/app/assets/javascripts/plugin-endpoint.js"), new File(classPathRoot, "plugin-endpoint.js"));
     }
 
     private static void assertActivationJarPresent() throws IOException {
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -543,6 +543,10 @@ task war(type: Jar) {
     into('WEB-INF/classes')
   }
 
+  from("${project(':server').projectDir}/webapp/WEB-INF/rails.new/app/assets/javascripts/plugin-endpoint.js") {
+    into('WEB-INF/classes')
+  }
+
   from("${project(':config-server').sourceSets.main.resources}/cruise-config.xsd") {
     into '/'
   }
--- a/server/src/main/java/com/thoughtworks/go/server/service/plugins/AnalyticsPluginAssetsService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/plugins/AnalyticsPluginAssetsService.java
@@ -35,8 +35,7 @@ import javax.servlet.ServletContext;
 import javax.xml.bind.DatatypeConverter;
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileOutputStream;
-import java.io.InputStream;
+import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.security.MessageDigest;
 import java.util.Base64;
@@ -46,8 +45,10 @@ import java.util.zip.ZipInputStream;
 
 @Service
 public class AnalyticsPluginAssetsService implements ServletContextAware, PluginMetadataChangeListener {
-    public static final String HASH_ALGORITHM = "SHA-256";
+    private static final String HASH_ALGORITHM = "SHA-256";
     private static final Logger LOGGER = LoggerFactory.getLogger(AnalyticsPluginAssetsService.class);
+    private static final String PLUGIN_ENDPOINT_JS = "plugin-endpoint.js";
+
     private AnalyticsExtension analyticsExtension;
     private ServletContext servletContext;
     private ZipUtil zipUtil;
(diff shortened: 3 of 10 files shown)


kierarad:
@maheshp this is marques (not kiera) - let's figure out a time to sync up and review the code.

maheshp:
@marques-work I was talking to @ketan to get his opinion. One of the things he brought up was around the plugin-endpoint.js caching. All assets under `public` have a long lived cache. That would mean any changes to `plugin-endpoint.js` would not reflect since the file name is appended by a hash. Adding a hash to the file name would mean the plugins will not be able to refer to them.

The other thing to consider/think about is how would the plugin handle changes to the `plugin-endpoint.js`, wouldn't changes to the JS API break things for the plugin. In that case should we consider versioning the JS API as well?

marques-work:
@maheshp @ketan interesting point. Even though it's installed in the hashed directory for the plugin, it's still possible for it to go stale so long as the plugin assets' content doesn't change.

I suppose we will need to version these then, and possibly maintain several versions.

Or, update the SHA-256 digest with the plugin-endpoint.js contents as well. That would certainly take care of all caching issues.

ketan:
> Even though it's installed in the hashed directory for the plugin

"hashed directory for the plugin" seems interesting, care to elaborate?

ketan:
>> Even though it's installed in the hashed directory for the plugin
> 
> "hashed directory for the plugin" seems interesting, care to elaborate?

My point is that if the base directory out of which everything is served works to bust the cache, we might be good and no changes may be needed. `/go/assets/${PLUGIN_ID}-${LONG_HASH}/foo.{js,css,img}`

The LONG_HASH might just mean a (secure) random string generated everytime the plugin is loaded.

marques-work:
@ketan - sure :).

When we pull the zipped static assets payload from the plugin, we explode it into a directory whose name is the SHA-256 hash of zip payload bytes. That ensures that when the plugin assets content changes, the URI path to the assets changes too, which takes care of browser caching issues. This happens each time an analytics plugin is loaded, which is, for the most part, once at startup.

However, the plugin-endpoint.js content is not included in the SHA-256 digest, so if someone upgrades GoCD (and the content of that script changes), but keeps the same plugin version, one could still load a stale plugin-endpoint.js from the iframe.

Unless of course, we calculate the SHA-256 against the plugin-endpoint.js script as well.

marques-work:
@ketan that was my initial thought too, but the plugin-endpoint.js is **copied from GoCD** into the hashed directory, which means if only that file updates, the browser may use a stale version unless we generate the hash on the combined content of the plugin assets zip payload and plugin-endpoint.js

marques-work:
@ketan I see what you mean -- let the LONG hash be unique and unpredictable instead of content based. That would work too. I guess our initial thoughts were to invalidate only when content actually changes and utilize the browser hash when we can. For our plugin, we have few assets so it might be ok. However, we can't guarantee other plugin authors won't load a bunch of stuff.

Using a GUID instead is a very simple change though, I admit :).

ketan:
> @ketan I see what you mean -- let the LONG hash be unique and unpredictable instead of content based. That would work too. I guess our initial thoughts were to invalidate only when content actually changes and utilize the browser hash when we can. For our plugin, we have few assets so it might be ok. However, we can't guarantee other plugin authors won't load a bunch of stuff.

I hear your concern about plugin authors loading a bunch of stuff, and the cache busting on clients because of a server reboot for no good reason.

I think doing this might be a good enough middle-ground:

* package the `plugin-endpoint.js` in assets dir (for the rails app). Likely part of `application.js` or something based on whatever the sprockets gem does.
* package the `plugin-endpoint.js` AS-IS (without compression, or sha in the name) somewhere on the classpath (like how it's being done [here](https://github.com/gocd/gocd/pull/4155/files#diff-2af3da569e27d5fad9d8c14613c35a66R546))
* the base-dir for your plugin assets becomes a function of zip file contents and the contents of the original `plugin-endpoint.js` on the classpath

This will avoid having to go through the `RailsAssetService` for resolving the actual path at runtime.

marques-work:
@ketan yep, that's _EXACTLY_ what I was referring to :)

> we generate the hash on the combined content of the plugin assets zip payload and plugin-endpoint.js

But I probably didn't spell it out so clearly as you did.

The change to do that would be simple, as should involve just updating the hash with a second byte array.

maheshp:
@marques-work @ketan generating the hash using the plugin_endpoint.js would solve the caching issue, but I am concerned with how to handle the JS API changes.

Instead, does it makes sense to version the JS API and make it available to the plugin authors(maybe thro the plugin-api.gocd.org). The plugins can use this JS and bundle it as part of its static assets.

GoCD would use the appropriate JS API based on the plugin version to pass messages to the plugin iframe.

marques-work:
@maheshp right, back to the versioning question.

My thoughts are that we'll probably need to version as this JS API will be public so that plugin developers can use it. perhaps the api version could be a top-level namespace for the supported keys?

e.g. `v1.analytics.pipeline-charts`

maybe we could have several send methods that prepend the version to the key:

`sendv1()`, `sendv2()`, etc..

or just have separate `plugin-endpoint-v1.js`, `plugin-endpoint-v2.js` that plugin developers can include.

ketan:
> Instead, does it makes sense to version the JS API and make it available to the plugin authors(maybe thro the plugin-api.gocd.org). The plugins can use this JS and bundle it as part of its static assets.

https://www.npmjs.com/ — that's the convention for distributing JS code, and all tooling works with this.

Might mean pulling out that JS into a git of its own with its own release cycle.

marques-work:
@ketan wouldn't that lead to situations where the plugin-endpoint.js on the plugin is different than on the server? We decided that by serving it through Go server, we'd avoid this type of incompatibility.

Granted there's other incompatibilities issues to still deal with, like a plugin expecting a higher version of the JS API than the version of GoCD provides.

marques-work:
@maheshp @ketan additionally, this script is just defining the connector and a means to define an API, _but not the API itself._ The API definition will likely be another script, packed into application.js that lays out the messages GoCD supports for plugins. Plugins won't need to import the API, they just need to write conforming messages against that standard.

maheshp:
I was thinking the JS API version should be tied to the plugin version. Say a plugin implementing v1.0 of the endpoint will have to use JS API v1.0 and would get the corresponding messages. The script as you rightly said would just be a connector to handle these messages.

ketan:
I don't have enough context around this, so perhaps I'm wrong. But my understanding it that sooner or later, the server (but not the plugin) will have backward compatibility to deal with — which means a couple of versions of the endpoint.js on the server and some sort of a way to negotiate what version of the JS the plugin uses.

We do this with most API contracts, and I don't believe this is any different.

marques-work:
@ketan yeah, that's certainly correct as well, and we may end up with several versions of the connector living on the server under different namespaces. Needs some thought.

PR #4157, opened 2018-01-04
Environment Index Page - Change view to show the environment config for each environment.

varshavaradarajan:
@arvindsv - As per what we discussed, 
- The environments index page now shows the list of pipelines, list of agents and env vars associated with each environment. 
- Only super admins get the option of editing the environment config to add/remove pipelines, agents and env vars. 
- Other users get the view of all the pipeline names that are associated with the environment (even if they do not have the permission to view the pipeline config

From the new view, the new information that non admin users now get is which agent is associated with a specific environment. I think that that's alright to show, as that information would have anyway been available from the agents page.  Let me know if you have any concerns. 

screen shot 2018-01-04 at 11 42 15 pm


varshavaradarajan:
The URL for this page is still `/go/environments`. Should it move under under the admin dropdown because all config edit pages are under admin? The only problem with that is the config listing view for environments is available for non admins too (without the edit and add buttons). But if this page moves under the admin dropdown, non admins will not be able to view it. 

After discussion with @arvindsv , the new view for the page does not seem important enough to be in the main header. What could be the ideal location for this, @gocd/committers ?

maheshp:
Guess this was discussed earlier. I think for the environments we should have 2 views.
1. A toggle on the dashboard to group all pipelines by either environment or pipeline_group. This will be part of the new dashboard SPA.
2. The environments edit page listed under the admins tab, which will be accessible only by admins to manage the environments.

Since the new dashboard page is still WIP we can consider having this page under the admins tab for managing the environment. The current environment page can be just a readonly view which groups pipelines by environment. 

Once the dashboard adds the ability to group pipelines by environment we can remove existing page.

arvindsv:
I don't mind removing/hiding the current environments page. I think the only reason people use that environments page is to edit environments (which point 2 in @maheshp's comment will solve). I'd like to hear from people using it differently, if that's not the case. To get to hear from them, what better way that to remove it?

varshavaradarajan:
Closing this. Can be reopened if needed.

PR #4161, opened 2018-01-06
Missed to clean few stuff in (PR#4138).

bdpiparva:


server/src/main/java/com/thoughtworks/go/server/security/ReAuthenticationFilter.java
```diff
@@ -44,7 +43,7 @@ public ReAuthenticationFilter(SystemEnvironment systemEnvironment, TimeProvider
     protected void doFilterHttp(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
         Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
 
-        if (!systemEnvironment.isReAuthenticationEnabled() || (authentication == null) || !authenticatedUsingAuthorizationPlugin(authentication)) {
+        if (!systemEnvironment.isReAuthenticationEnabled()) {
```
maheshp: Shouldn't we retain the check for `(authentication == null)`?

server/src/main/java/com/thoughtworks/go/server/security/providers/PluginAuthenticationProvider.java
```diff
@@ -132,7 +132,7 @@ private User getUserDetailsFromAuthorizationPlugins(String username, UsernamePas
 
     private GoUserPrinciple getGoUserPrinciple(User user, String loginName, boolean requiresPeriodicReAuthentication) {
```
maheshp: Remove `requiresPeriodicReAuthentication` from method params.

PR #4164, opened 2018-01-08
Elastic agent status report endpoint

GaneshSPatil:
**TO BE MERGED for 18.2**
##### Done
* Add a route to display agent status report page for each elastic agent
* Redirect to agent status report page from build details page

##### Pending
* Replace/Add a link from agent details page to agent status report (_only for elastic agents_).


server/webapp/WEB-INF/rails.new/app/controllers/admin/status_reports_controller.rb
```diff
@@ -32,16 +32,27 @@ def show
       render_plugin_error e
     end
 
+    def show
+      @view_title = 'Plugin Status Report'
```
bdpiparva: Can we change it to `Agent status report(agentid)` ?

server/webapp/WEB-INF/vm/build_detail/_build_detail_summary_jstemplate.vm
```diff
@@ -29,7 +29,11 @@
     <span>Not yet assigned</span>
   {/if}
   #if ( $elasticProfilePluginId && $userHasAdministratorRights )
-    <a href="$req.getContextPath()/admin/status_reports/$elasticProfilePluginId" class="btn-primary btn-small status-report-btn-small">Check Plugin Status</a>
+    {if build.build_assigned_date != -1}
```
bdpiparva: @gocd/committers --- With this implementation if plugin has not implemented `agent-status-report` call it will raise `UnhandledRequestTypeException`.  We will update the document saying both the call is mandatory if plugin supports [status_report](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/elastic/Capabilities.java#L26) capability. Is this okay ?

ketan:
Seems reasonable. Unless you are looking to add multiple capabilities for
each report page.

server/webapp/WEB-INF/rails.new/app/controllers/admin/status_reports_controller.rb
```diff
@@ -32,16 +32,27 @@ def show
...
+      @page_header = 'Plugin Status Report'
+      @agent_status_report = elastic_agent_extension.getAgentStatusReport(params[:plugin_id], params[:elastic_agent_id])
```
maheshp: What should be the behaviour when an agent is not assigned?

GaneshSPatil: @maheshp -- If the job is not assigned yet, it shows the *plugin status report* instead of *Agent Status Report* button. The check is done [here](https://github.com/GaneshSPatil/gocd/blob/7ec2e32c0010955951689b33a42e49474867748e/server/webapp/WEB-INF/vm/build_detail/_build_detail_summary_jstemplate.vm#L32-L36)

Going forward, we will add *Errors and Warnings* sections to the EA endpoint, which will tell errors regarding agent, job assignment failure, cluster, etc.

maheshp: Please correct me if I am wrong. I thought an agent status report would be more useful in cases where an agent is not yet assigned. I thought the plugin through the agent status report should be able to answer questions around the status of the agent creation, wouldn't this be possible since we send the job identifier to the plugin.

server/webapp/WEB-INF/rails.new/app/controllers/admin/status_reports_controller.rb
maheshp: Also, should the actions be `agent_status` and `plugin_status` over `index` and  `show`?

GaneshSPatil:
Changes done as part of the current PR:

* Introduce Agent Status Report endpoint
  - Add `agent-status-report` endpoint in Elastic Agent Extension V2
  - Following information is sent to plugin as part of request body
    - `Elastic Agent Id`
    - `Job Identifier`

* Define route for Agent Status Report Page
  - Using Elastic Agent ID
    Route URL: `/go/admin/status_reports/<PLUGIN_ID>/<ELASTIC_AGENT_ID>`
  - Using Job ID (in case of elastic agent id is not available)
    Route URL: `/go/admin/status_reports/<PLUGIN_ID>/unassigned?job_id=<JOB_ID>`

**Note**: Based on the information available in the context from which the `agent-status-report` page is referred, either one of the elastic_agent_id or job_identifier field will be sent to plugin as part of request body.

Issue #4135, opened 2017-12-28
Move over roles controller and representers to java


ketan:
##### Issue Type

- Improvement

##### Summary

The current set of representers using JRuby have a severe limitation in terms of performance. We've written a [jrepresenter](https://github.com/gocd/jrepresenter/) library to mimic some of the features provided by roar+representer. The java bit is not as concise as jruby, but does the job well enough and is not as bad as I'd initially thought it'd be.

We also took this opportunity to move the controllers over to java for a couple of reasons —
- There is not much code in the rails controllers to justify using ruby (or rails)
- Doing it in java makes it easier to mock/stub things with the safety net of java
- Upgrades (of rails and rspec) don't matter anymore because we'd be using the servlet API, which does not change, and break compatibility

I'm opening this issue to show some of the code we've written and gather feedback on (a) how the representers are written and (b) how the controllers are written.

**Representers:**

- Some example code [here](https://github.com/akshaydewan/gocd/tree/spikes/spark/config/config-representers/src/com/thoughtworks/go/config/representers). Ignore the *Serializer and *Deserializer classes.
- Most of the code is annotation driven, but for things where the annotations get in the way, it's possible to write a [simple serializer](https://github.com/akshaydewan/gocd/blob/a0a932857d8a71d80728ce73ed41aae4b250d345/config/config-representers/src/com/thoughtworks/go/config/representers/PluginRoleConfigRepresenter.java#L74-L82) to do the job
- [the tests](https://github.com/akshaydewan/gocd/tree/a0a932857d8a71d80728ce73ed41aae4b250d345/config/config-representers/test/com/thoughtworks/go/config/representers). Because representer tests mostly involve comparing lists and maps and other json structures, they have been written in groovy. The the readability of creating simple lists and maps is much better in groovy as compared to java. Don't worry, any valid java syntax is valid groovy :) **We are not using groovy for the production code, only the test code.**

**Controllers:**

[...]

PR #4167, opened 2018-01-08
Create a RolesControllerV1 implementation using Spark and JRepresenter (#4135)

akshaydewan:


akshaydewan:
Further enhancements to be done:
- [ ] Ignore trailing slash in request URLs (same as current behavior in other controllers)
- [ ] In PluginRole representer, show encrypted value if secure
- [ ] Figure out if the design of `setDefaultHeaders` in `BaseController` can be improved

ketan:
I want to put in some notes and conventions that might help reduce friction for someone writing new controllers.

---

### Controllers

There are 2 things to wiring up controllers:

* a `@Component` that is wired up using spring. [This component](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api-v1/src/main/java/com/thoughtworks/go/apiv1/admin/security/spring/RolesControllerV1.java) must not use any lambdas, java 8 method references etc because of limitations of current spring version. This goes in a `.spring` package and is component-scanned [here](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/server/webapp/WEB-INF/applicationContext-global.xml#L31-L32)

   This component should simply hold a reference to a delegate that processes the incoming request

* the [actual delegate](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api-v1/src/main/java/com/thoughtworks/go/apiv1/admin/security/RolesControllerV1Delegate.java) will setup routes, and process the incoming requests. It uses a wrapper around the `ServletRequest/Response` objects. This should not go in the `.spring` package as it will contain lambdas and method references.

---

### Controller tests

[Controller tests](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api-v1/src/test/groovy/com/thoughtworks/go/apiv1/admin/security/RolesControllerV1DelegateTest.groovy) will be written in groovy + junit5 + assertj. For several reasons:

- junit 5 allows nested tests, like rspec's `describe` inside `describe`
- groovy because its dynamic, and since a lot of the controller tests involve asserting on JSON, it's easier to deal with maps and multi-line strings in groovy
- assertj because it provides an experience that is closer to rspec, than junit assertions, and significantly increases the readability of tests

The [controller tests](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api-v1/src/test/groovy/com/thoughtworks/go/apiv1/admin/security/RolesControllerV1DelegateTest.groovy) have some convention that will again help reduce friction, and keep your experience as close to rails as possible:

```java
// traits are like ruby modules that are `include`

class FooControllerDelegateTest implements
 // provides a mock `securityService` instance with some helpers to allow security setup.
 SecurityServiceTrait, 
 // provides a of the dsl to create and dispatch requests and assert on responses
 ControllerTrait<RolesControllerV1Delegate> 
{
  // tests for the `index` action. must be `@Nested` for junit to know
  @Nested class Index {
    // must implement `SecurityTestTrait`
    // provides a simple structure to setting up your security and asserting if the response was 200 or 401
    @Nested class Security implements SecurityTestTrait {

      // kind of like rspec's `it 'should foo'`
      // make sure the return type is void, or junit will ignore the test
      @Test
      void 'should foo'() {

      }

    }

    // what you can do as user
    @Nested class AsUser { }

    // what you can do as admin
    @Nested class AsAdmin { }
  }
}
```

---

### gradle modules

There are currently 3 gradle modules that we've introduced.

- [`config-representers`](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/config/config-representers/build.gradle), this will include jrepresenter code to convert config objects to json, and vice-versa. Its only dependencies are `config-api` and `jrepresenter`
- [`api`](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api/build.gradle) — the plumbing for hooking spark into the webapp. It also contains some plumbing used by any controller tests
- [`api-v1`](https://github.com/gocd/gocd/blob/bbc8e27a2a96131449c5569a9d5fa00e29d061cc/api/api-v1/build.gradle) — the module for v1 of the api. Currently only has partial support for roles api.

We can add api-v2, v3 etc as and when needed. This is mostly done to keep boundaries and avoid accidental sharing of code between api versions.

PR #4169, opened 2018-01-09
Include pipeline stages as part of stage status notification

maheshp:
* With the current stage status notification message, the plugin gets
  data only for a specific stage of the pipeline. This information might
  not suffice for plugins which might collate data at a pipeline level,
  e.g analytics plugin.
* With this commit, the stage status notification message consists of
  the pipeline stage names.


ibnc:
@maheshp Any idea when this PR will be reviewed/merged?

maheshp:
@ibnc was planning to get this after we released `18.1.0` which should be out soon. Also wanted to see if we need to add any additional information apart from the stage names.

maheshp:
Don't see a need for this now, closing the PR.

PR #4174, opened 2018-01-12
Add Analytics Dashboard for Analytics Plugins

kierarad:
Analytics "dashboard" is top level (mithril) page.

- Link to "Analytics" included in the header if any installed plugins have "supports_analytics_dashboard" capability.
- Creates iframe for each plugin with "supports_analytics_dashboard" capability. (We will revisit allowing one plugin to have multiple iframes)
- Fetches dashboard analytics from plugin 
- Initiates messaging between iframe and parent window


server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
```diff
@@ -17,7 +17,22 @@
...
+
+  def index
+    @view_title = 'Analytics'
+    @plugin_ids = default_plugin_info_finder.allPluginInfos(PluginConstants.ANALYTICS_EXTENSION).inject([]) do |memo, plugin|
+      memo << plugin.getDescriptor().id() if plugin.getCapabilities().supportsAnalyticsDashboard()
+      memo
+    end
+  end
+
+  def dashboard
```
maheshp: How do we handle authorization for dashboard analytics? We might end up showing metrics for users who do not view permission e.g for a pipeline?

arvindsv: Doesn't [this line](https://github.com/gocd/gocd/pull/4174/files/7c3f36fd447638c9323102d961d407572d7d31d9#diff-e79bcabcd354070f287977abbc39415dR23) do that?

maheshp: Guess you are right it will take care of auth when we access an individual pipeline, but the dashboard page would still list out pipelines for which user might not have permissions to view.

arvindsv: Oh, I see. For `index`. Hmm. Yeah. We will have to think about it. That's more of a concern for the plugin, but we will have to see if we need to provide anything special in the infrastructure for the plugin to say, "Show this only to users who can see all pipelines, or this group" etc.

For now, the dashboard listing pipelines is more of a spike. It's not likely to be in the final setup. It's not very useful by itself. The infrastructure to have a top-level dashboard (or dashboards) should be the focus.

maheshp: I agree, lets ignore the data from the plugin.

> The infrastructure to have a top-level dashboard (or dashboards) should be the focus.

I think for building the infrastructure we should have few more concrete things. Some of questions I had in mind were,
1. Will a plugin support a single analytics dashboard or multiple? How do we handle/list multiple dashboards by a plugin?
2. How do we handle multiple plugins which support dashboard analytics?
3. If a plugin supports multiple dashboards should we introduce a new message in the plugin endpoint to list the dashboards supported by the plugin.

We can get on a call discuss some of these if it helps.

arvindsv: Yes, I talked briefly about this with Kiera and Marques, last week. Some of it depends on design, and some of it, I'm happy to make some assumptions that'll help us move on. For instance, for now, we can assume one plugin only, which has multiple dashboards.

We will eventually need to handle multiple plugins with multiple dashboards, and decide how to show them all and in what order, etc. We might need to provide a way to configure that, as a part of settings somewhere.

maheshp:
@kierarad the PR looks ok, few specs are broken can you please fix it.

The Analytics link needs to be added to velocity page headers as well, the partial for it is https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/vm/shared/_header.vm
You can verify this on the pipeline history page.

server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
kierarad: @maheshp @arvindsv I just want to clarify that as it currently stands we actually support multiple plugins. We don't support multiple dashboards. At some point we can definitely revisit supporting multiple dashboards for one plugin (and therefore have support for both multiple plugins and multiple dashboards per plugin).

arvindsv: Ok. Yes, Mahesh was talking about possibly having a new call which asks for a list of dashboards to show and then either showing them or showing a dropdown. We can decide later. For now, multiple plugins, and one dashboard in each. I understand.

kierarad:
@maheshp we've fixed the specs, added additional test coverage, and added the link in the velocity template. Can you review and let us know if there is anything else we need to do to make it mergeable?

PR #4184, opened 2018-01-15
Move java controller and representer to API module

GaneshSPatil:
* Create a separate module per versioned API endpoint


development-utility/development-server/src/main/java/com/thoughtworks/go/server/DevelopmentServer.java
```diff
@@ -36,6 +36,10 @@
  */
 
 public class DevelopmentServer {
+    static {
```
GaneshSPatil: remove this

PR #4187, opened 2018-01-15
Fix error representers for Role and ConfigProperty

akshaydewan:


api/api-roles-config-v1/src/test/groovy/com/thoughtworks/go/representers/config/rolev1/ConfigurationPropertyRepresenterTest.groovy
```diff
@@ -28,6 +29,12 @@ import static org.assertj.core.api.Assertions.assertThat
 
 class ConfigurationPropertyRepresenterTest {
 
+  @BeforeClass
+  static void init() {
+    //TODO need to find a better workaround
+    System.setProperty("json-unit.libraries", "gson");
+  }
```
akshaydewan: @ketan The test became flaky on my machine again, so I added this.

akshaydewan:
Note: jrepresenter needs to be bumped up before merging this

api/api-roles-config-v1/src/test/groovy/com/thoughtworks/go/representers/config/rolev1/ConfigurationPropertyRepresenterTest.groovy
bdpiparva: Look at this http://stefanbirkner.github.io/system-rules/#EnvironmentVariables. It provides junit rules for system properties and env.

PR #4188, opened 2018-01-16
Introduce Pipeline pause unpause API using java controller

GaneshSPatil:
issue: #4076
Related PR: https://github.com/gocd/gocd/pull/4110


ketan:
This has been merged, but will be released for 18.2 after some feedback from based on the new dashboard.

PR #4203, opened 2018-01-19
Sort pipelines on pipeline and environments pages.

markhibberd:
We are currently working around this with a browser extension, if pipelines are not sorted, things start jumping around when you have lots of pipelines from lots of config-repos (we are using the yaml config plugin).

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/views/environments/_environment.html.erb
+++ b/server/webapp/WEB-INF/rails.new/app/views/environments/_environment.html.erb
@@ -13,7 +13,7 @@
         <% if scope[:pipeline_models].empty? %>
             <span><%= l.string('NO_PIPELINES_CONFIGURED_FOR_ENVIRONMENT')-%></span>
         <% else %>
-            <% scope[:pipeline_models].each do |pipeline_model_in_environment| %>
+            <% scope[:pipeline_models].sort_by { |pipeline| pipeline.getLatestPipelineInstance().getName() }.each do |pipeline_model_in_environment| %>
                 <div id="<%= env_pipeline_dom_id(pipeline_model_in_environment) %>" class="pipeline hidereveal_collapsed">
                     <% unless scope[:omit_pipeline] %>
                         <% cache(view_cache_key.forEnvironmentPipelineBox(pipeline_model_in_environment), {:subkey => "environment_html", :skip_digest => true}) do %>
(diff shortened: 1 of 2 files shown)


ketan:
This might be a breaking change for a lot of users. It may make sense to add an option to allow sorting of pipelines, probably using a query param or something.

@arvindsv thoughts?

markhibberd:
@ketan At the moment it doesn't even return the same order every time, in what way could someone be depending on it?

ketan:
> @ketan At the moment it doesn't even return the same order every time, in what way could someone be depending on it?

I see your point — I was not aware of the fact that the order is different everytime. I can investigate why this is so and get back to you.

markhibberd:
@ketan Thanks, I don't mind so much about the sorting, just that it becomes hard to find things when you have to look in a different place everytime, so consistent but not sorted is just as good.

ketan:
@jyotisingh @tomzo any idea about the sorting?

tomzo:
> At the moment it doesn't even return the same order every time

That's true, this is basically a bug.

> any idea about the random order of pipelines?

Sorting config-repo pipelines alphabetically is the easiest option. Although, this change will sort all pipelines, even if they are only in XML, so that's kind-of breaking.
Perhaps the change should be somewhere in [this method](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java#L340), sorting the pipelines in each group, but only if that group has config-repo pipelines.

Some users want the [pipeline order to match with order in yaml](https://github.com/tomzo/gocd-yaml-config-plugin/issues/4), but the real issue is probably order of pipelines added among many config-repos. The harder solution would be something like this:
Given some xml pipelines and multiple repositories:
```xml
<pipelines>
<!-- defines pipelines X Z -->
</pipelines>
<configrepo id="a">
<!-- defines pipelines L K -->
</configrepo>
<configrepo id="b">
<!-- defines pipelines G H -->
</configrepo>
```
The order of pipelines would be `X Z L K G H`. Because 
 * we just put xml pipelines first, like "normal" server does.
 * config-repo plugins return a list of pipelines, in order in which plugin defines. (This would allow to implement https://github.com/tomzo/gocd-yaml-config-plugin/issues/4)
 * we sort pipelines first by index within `<config-repos>` section, then as returned by plugin.

But with the magical xml loader I am not sure if we can actually get `index within <config-repos>`.

arvindsv:
> Sorting config-repo pipelines alphabetically is the easiest option. Although, this change will sort all pipelines, even if they are only in XML, so that's kind-of breaking.

+1 👍 - Yes, it is a breaking change for those who don't use config-repo pipelines. So, this change will need to be specifically for that - and not at the generic level that this PR is doing it at.

PR #4204, opened 2018-01-19
Remove jrepresenter

akshaydewan:
Since the boilerplate needed to generate boilerplate was almost equal to the generated boilerplate, we're removing jrepresenter.


api/api-base/src/main/java/com/thoughtworks/go/api/representers/RequestContext.java
```diff
@@ -0,0 +1,64 @@
...
+    public Link build(String name, String template, Object... args) {
+        try {
+            String href = String.format(template, args);
+            return getLink(name, href);
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public Link buildWithNamedArgs(String name, String template, Map<String, Object> args) {
```
ketan: Unused method?

api/api-base/src/test/groovy/com/thoughtworks/go/api/utils/JsonUtil.groovy
```diff
@@ -14,17 +14,18 @@
...
+import com.thoughtworks.go.api.util.GsonTransformer
+import groovy.json.JsonOutput
 
-import java.util.function.BiFunction;
+class JsonUtil {
 
-public class CaseInsensitiveStringSerializer implements BiFunction<CaseInsensitiveString, RequestContext, String> {
+  static JsonObject jsonObjectFrom(map) {
+    def jsonString = JsonOutput.toJson(map)
+    GsonTransformer.instance.fromJson(jsonString, JsonElement.class).asJsonObject
```
ketan: As we've seen before, the behavior of all JSON libraries is subtly different. Can we move this into the GsonTransformer itself?

```java
JsonElement toJsonElement(Object o) {
  return GSON.toJsonTree(o);
}
```

api/api-roles-config-v1/src/main/java/com/thoughtworks/go/apiv1/admin/security/representers/RoleRepresenter.java
```diff
@@ -16,87 +16,79 @@
...
 
-        @Override
-        public List<Link> getLinks(Role model, RequestContext requestContext) {
-            return Arrays.asList(
-                    DOC,
-                    requestContext.build("self", "/go/api/admin/security/roles/%s", model.getName()),
-                    requestContext.build("find", "/go/api/admin/security/roles/:role_name")
-            );
+        if ("gocd".equals(type)) {
+            model = GoCDRoleConfigRepresenter.fromJSON(jsonObject.get("attributes").getAsJsonObject());
```
ketan: This seems to assume that `attributes` is always present in the JSON, even if the value is blank and is a potential for a NPE.

api/api-roles-config-v1/src/test/groovy/com/thoughtworks/go/apiv1/admin/security/RolesControllerV1DelegateTest.groovy
```diff
@@ -533,7 +532,8 @@ class RolesControllerV1DelegateTest implements SecurityServiceTrait, ControllerT
         ]
         def body = [
           name: 'blackbird',
-          type: 'plugin'
+          type: 'plugin',
+          attributes: [:]
```
ketan: Remove the `attributes` node to allow testing for no attributes?

api/api-base/src/main/java/com/thoughtworks/go/api/representers/RequestContext.java
akshaydewan: Planning to use this in the dashboard API for better readability. I can remove it now and add it later, or let it stay?

ketan: Let it stay, I suppose.

PR #4208, opened 2018-01-22
Log progress of pluggable artifact upload and download on console log

bdpiparva:


common/src/main/java/com/thoughtworks/go/domain/builder/FetchPluggableArtifactBuilder.java
```diff
@@ -63,10 +66,10 @@ public FetchPluggableArtifactBuilder(RunIfConfigs conditions, Builder cancelBuil
         this.metadataFileLocationOnServer = source;
     }
 
-    public void build(DefaultGoPublisher publisher, EnvironmentVariableContext environmentVariableContext, TaskExtension taskExtension, ArtifactExtension artifactExtension) {
+    public void build(DefaultGoPublisher publisher, EnvironmentVariableContext environmentVariableContext, TaskExtension taskExtension, ArtifactExtension artifactExtension, PluginRequestProcessorRegistry pluginRequestProcessorRegistry) {
         downloadMetadataFile(publisher);
-
         try {
+            pluginRequestProcessorRegistry.registerProcessorFor(CONSOLE_LOG.requestName(), new ArtifactRequestProcessor(publisher));
```
varshavaradarajan: extract a variable for CONSOLE_LOG.requestName() ?

common/src/test/java/com/thoughtworks/go/remote/work/ArtifactRequestProcessorTest.java
```diff
@@ -0,0 +1,67 @@
+/*
```
varshavaradarajan: Move this to artifact package? Move ArtifactsPublisherTest also the the artifact package

common/src/main/java/com/thoughtworks/go/domain/builder/FetchPluggableArtifactBuilder.java
bdpiparva: I think this is fine.

PR #4211, opened 2018-01-22
Updated web_hook_controller.rb to include gitlab@ urls

DavidWylie:
Many Gitlab installations use gitlab@ as a git user.  this was once a default user for gitlab installs.  
Due to this the urls for ssh connections from gitlab can include this user.  
Changing the user is a non trivial event.

This adds gitlab to this list so that it can be used in gocd.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/api/web_hooks/git_lab_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/api/web_hooks/git_lab_controller.rb
@@ -1,5 +1,5 @@
 ##########################################################################
-# Copyright 2017 ThoughtWorks, Inc.
+# Copyright 2018 ThoughtWorks, Inc.
 #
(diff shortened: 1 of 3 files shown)


ketan:
Since this code is specific to gitlab, it should ideally belong to `git_lab_controller.rb` and not the superclass `web_hook_controller.rb` — as this will affect other webhook implementations. Also, please update the corresponding [spec file](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/spec/controllers/api/web_hooks/git_lab_controller_spec.rb)

DavidWylie:
@ketan Thanks for the feedback.  Ive copied the function to the gitlab webhook controller and added the changes needed and added it to the spec tests.

varshavaradarajan:
PR build - https://build.gocd.org/go/pipelines/value_stream_map/trigger/2181

DavidWylie:
Thanks.

PR #4216, opened 2018-01-23
Support multiple metrics per plugin on analytics dashboard

kierarad:
- Changed "supports analytics dashboard" capability to be a list of metrics supported by
  plugin vs a boolean (and renamed accordingly)
- Create iframe for each dashboard metric (now we support multiple plugins and multiple
  dashboard metrics per plugin)
- Changed signature of getAnalyticsDashboard() to require a "metric" parameter
- Updated dashboard and pipeline routes to reflect new parameterization (and to make them
  distinct from each other)
- Added test coverage for AnalyticsMessageConverter and Capabilities


maheshp:
The changes look mostly ok to me. I am just wondering, for the dashboard metrics would a plugin require some extra parameters apart from the metric name and if we should support it.

kierarad:
@maheshp it's definitely possible that a dashboard metric might require extra parameters. It's hard to think of a scenario off the top of my head though. Maybe that is something we can revisit when we find a case where more parameters are needed?

server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
```diff
@@ -24,14 +24,15 @@ class AnalyticsController < ApplicationController
 
   def index
     @view_title = 'Analytics'
-    @plugin_ids = default_plugin_info_finder.allPluginInfos(PluginConstants.ANALYTICS_EXTENSION).inject([]) do |memo, plugin|
-      memo << plugin.getDescriptor().id() if plugin.getCapabilities().supportsAnalyticsDashboard()
+    @plugin_ids = default_plugin_info_finder.allPluginInfos(PluginConstants.ANALYTICS_EXTENSION).inject({})do |memo, plugin|
```
arvindsv: This should no longer be called `@plugin_ids`? Right, @kierarad and @maheshp? It looks like it's a hash.

Issue #4031, opened 2017-11-22
Deprecation notice for plugin upload form


varshavaradarajan:
##### Issue Type

- Deprecation Reminder

##### Summary

- Now that the plugins SPA will be enabled by default, the old_plugins page (/go/admin/old_plugins - after https://github.com/gocd/gocd/pull/4030 ) should be removed in 3 months time. 

- [ ] The plugin upload feature, which was always in beta will be removed in 3 months time along with the old page. A deprecation notice in the release notes for 
- [ ] The system property `go.plugin.upload.enabled` which was false by default all this while should be removed
- [ ] Remove https://docs.gocd.org/current/beta/plugin_upload.html from docs.

PR #4226, opened 2018-01-25
Remove old plugins page

ketan:
* Remove plugins toggle
* Remove (beta) upload plugin feature, along with feature toggle


varshavaradarajan:
https://github.com/gocd/gocd/issues/4031

PR #4230, opened 2018-01-25
Fix jasmine build on windows

GaneshSPatil:
* IE 11 doesnt support String.prototype.includes method.
* Replace the usage of includes with indexOf


GaneshSPatil:
Jasmine spec passed for `linux` and `windows`
https://build.gocd.org/go/pipelines/value_stream_map/trigger/2193

Issue #4206, opened 2018-01-19
Parameterization of storeId and artifactId in templates and pipelines


varshavaradarajan:
##### Issue Type
- Feature enhancement

##### Summary

Pluggable artifact configuration can be configured at a job level in templates and pipelines. When using this configuration with templates, this configuration can be parameterized. Specifically, the store id. Given that the store id can be parameterized, different pipelines using the templates can define different store ids for stores from different plugins. So, essentially, for a job in the template, the specific plugin properties that need to be configured will be based on what the store id is. This creates a problem because users may end up specifying properties that are not useful for a plugin (will result in validation errors on all the pipelines, when configured through the UI and API). Moreover, for the view, we cannot even determine what the plugin view for configuration should be if we cannot determine what the plugin id is from the store id, as it is parameterized.

Similarly, for fetch pluggable artifact, since the artifactId can be parameterized, plugin specific properties could potentially result in not being useful and the fetch artifact view cannot be determined.  

We decided to work around this by adding the plugin id attribute on the pluggable artifact configuration and the fetch pluggable artifact configuration. 

- pluginID attribute cannot be parameterized. 
- Validate whether the plugin id associated with the store id or artifact id is the same as the plugin id configured by the user. 

Here's how the configuration will look like - 

```json
<pluggableArtifact id="demo-image" storeId="#{param}" pluginId="cd.go.artifact.docker">
  <property>
    <key>BuildFile</key>
    <value>image_info.json</value>
  </property>
</pluggableArtifact>
```
```json
<fetchPluggableArtifact artifactId="#{another-param}" pluginId="cd.go.artifact.docker" pipeline="#{pipeline}" stage="build" job="build_docker_image">
  <property>
    <key>Destination</key>
    <value>somewhere-on-the-agent</value>
  </property>
</fetchPluggableArtifact>
```

cc @arvindsv @bdpiparva

arvindsv:
The fact that `storeId` is parameterizable is the one causing the problem here, right? If `storeId` is parameterizable, that tells users that stores can change across pipelines. That's not correct, given that the params will be different per store (and hence plugin).

So, disallowing that, will make all of this work well. Right? This is my ideal:

```xml
<artifactStore pluginId="plugin.1" id="cannot-be-param">
...
</artifactStore>
```

```xml
<artifact id="demo-image" storeId="cannot-be-param">
  <property>
    <key>BuildFile</key>
    <value>image_info.json</value>
  </property>
</artifact>
```

```xml
[...]

PR #4239, opened 2018-01-25
[WIP] Have a same tag-name to define publish artifact configurations.

bdpiparva:
Use same tags to configure build, test or pluggable artifact.

#### Examples
```xml
<artifacts>
  <artifact type="build" src="build/libs" dest="dist"/>
  <artifact type="test" src="test-results" dest="test"/>
  <artifact type="plugin" id="installer" storeId="docker">
    <property>
        <key>BuildFile</key>
        <value>docker-build.json</value>
    </property>
  </artifact>
</artifacts>
```


arvindsv:
@varshavaradarajan has been chasing me to agree to adding a pluginId next to storeId, to make it easier to use with templates (to be able to parameterize storeId and still be able to show a UI). Might need to consider that too.

varshavaradarajan:
@arvindsv , that's another PR and its pending discussion with @jyotisingh - https://github.com/gocd/gocd/issues/4206. I was just trying to find arguments against the default param thing so that we don't delve into it only to find out later that we can't do it. Apparently you think I have strong opinions. I am just a lowly developer.

arvindsv:
Ha ha! :) Tsk, tsk.

You made good points, though, in arguing for it.

config/config-api/src/main/java/com/thoughtworks/go/config/ArtifactConfig.java
```diff
@@ -17,111 +17,204 @@
...
         return errors().isEmpty();
     }
 
+    @Override
     public void validate(ValidationContext validationContext) {
-        if (!StringUtils.isBlank(destination) && (!(destination.equals(DEFAULT_ROOT.getPath()) || new FilePathTypeValidator().isPathValid(destination)))) {
+        if (artifactType == ArtifactType.plugin) {
+            validatePluggableArtifact(validationContext);
+        } else {
+            validateBuiltInArtifact(validationContext);
```
varshavaradarajan: Add more validations - 
1. If id and store id are given, src and dest must not be given
2. If src and destination are provided, pluggable artifact properties, id and store id must not be given.

Issue #4237, opened 2018-01-25
Pipeline unlocking API


ketan:
##### Issue Type

- Feature enhancement

##### Summary

Add ability to unlock a pipeline using an API endpoint.

ketan:
Use this controller https://github.com/gocd/gocd/blob/master/api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1Delegate.java

ketan:
@rajiesh — can we get a round of testing on this and close?

rajiesh:
Verified on 18.2.0 (6215-1a3f4adf5f908dcf4186c24f65c24fcda50d8e78)

PR #4243, opened 2018-01-29
Implement pipeline unlock API (#4237)

ketan:
```
curl http://localhost:8153/go/api/pipelines/up42/unlock \
    -X POST \
    -H 'Accept: application/vnd.go.cd.v1+json' \
    -H 'Confirm: true'
```

The response will contain a JSON string with a `message` key, and status
code will be:
 - 200 OK on success
 - 404 if pipeline is not found
 - 409 Conflict in case of errors with unlocking pipeline


ketan:
@GaneshSPatil — for you to review, most of the changes are refactorings, the crux of the code is in the delegate.

Also changed the status code to "409 conflict", which is different from the earlier "406 not acceptable" since the semantics of 409 are different.

> The HyperText Transfer Protocol (HTTP) 406 Not Acceptable client error response code indicates that a response matching the list of acceptable values defined in Accept-Charset and Accept-Language cannot be served.

PR #4249, opened 2018-01-29
Make analytics only visible to admins (temporary)

ibnc:
@maheshp @arvindsv Mind taking a look?


arvindsv:
@ibnc Can you look into adding a toggle instead? Something like: `should_show_analytics`. Take a look at `Toggles.java` and the file `available.toggles`. It's easier to control and easier to remove. This changes behavior in too many places.

ibnc:
@arvindsv and I talked, and decided that a toggle isn't needed

Issue #4254, opened 2018-01-30
Dashboard tile


ketan:
##### Issue Type

- Feature enhancement

##### Summary

List of things we need on the dashboard tile:

- [x] Show Pipeline Name.
- [x] Pipeline Name redirects to Pipeline History Page.
- [x] Show Pipeline Settings icon (Depending on the Pipeline permissions).
- [x] Show Pipeline Instance label.
- [x] Show Compare Pipeline Link.
- [x] Show Changes drop-down. As part of changes, show which material caused the pipeline to trigger. (#4114)
- [x] Linkify issue numbers in changes dropdown (#4192)
- [x] Show VSM link
- [x] Show Triggered By Information
- [x] Show Triggered On Information
- [x] Show Stages Status
- [x] Show pipeline paused by and pipeline paused cause when pipeline is paused
- [x] Show Pipeline operations response message.
    - `Operations`: Trigger, pause, unpause pipeline
    - `Example`: `Request for scheduling pipeline up42 accepted successfully.`
- [x] Show multiple stages instance for parallel pipeline executions. (Multiple stages instance).

arvindsv:
When a pipeline has never run, it looks like this:

<img width="291" alt="2018_02_02_15-37-44" src="https://user-images.githubusercontent.com/172235/35753791-100c5b30-082f-11e8-98a7-70796eb6ac06.png">

Should we consider showing a message?

eeluke:
Yes, we should. White area below action buttons should contain a short message like "You haven't run this pipeline yet. Click the play button to run pipeline."

GaneshSPatil:
##### Show on hover information for:
- [ ] Pipeline Name. (In case of long pipeline name, last few characters of pipeline name are stripped. On hover of the pipeline name, show the entire pipeline name).
- [ ] Scheduled at server time.
- [ ] Stage Instance Status.

FredrikWendt:
Regarding "You haven't run this pipeline yet. Click the play button to run pipeline." <- can we turn this off? The play button doesn't respect fan-in-fan-out, which cause confusion (much later when things don't work as expected, so the cost is really high for this). See #1924

I'm asking for the default "call to action" to NOT trigger the one flaw that breaks the biggest, single (unique) selling point of GoCD over other tools.

GaneshSPatil:
Done as part of **New Dashboard** feature in GoCD `v18.3.0`

Issue #4107, opened 2017-12-19
Pipeline times are now shown as full timestamps rather than time deltas


brendanheywood:
##### Issue Type

- Bug Report

##### Summary

I consider this a major usability regression bug for such a tiny technical change. Previously in gocd you would get a really useful and easy to understand time deltas such as '5 minutes ago'.

After our last upgrade these are now an ugly full timestamp, which forces you to go and look at a calendar and a clock to know what both the current time and date is, and then do some mental arithmetic it in order to extract useful meaning from it. If you want the exact timestamp you should still be able to hover over it and get the timestamp.

![image](https://user-images.githubusercontent.com/187449/34133469-05396a76-e4a9-11e7-8db6-3bbc6679cd28.png)

##### Environment

Go Server Version:	17.11.0 (5520-9f6909e2f64b07d2dce5cecd4ea5b92b8e19d6b1)
JVM version:	1.8.0_141
OS Information:	Linux 3.13.0-132-generic
Database schema version:	1708001

##### Steps to Reproduce

1. Go to the root page, or any pipeline page
2. Look at timestamps

##### Expected Results
 * Dates should be deltas in human parsable natural language like they were previously

##### Actual Results
* Dates are in a full timestamp 

##### Possible Fix
* Go back to the way it was! These deltas should also update in browser over time if you leave the window open.
* At the very least make this a config option

##### Code snippets/Screenshots

I would suggest the best path forward is introduce one  settings which toggles from the moment.js format strings to the .fromNow(); format.

'DD MMM YYYY [at] HH:mm:ss [Local Time]' eg

https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/build_summary_observer.js#L60

I still think the delta is the more sensible default, but as long as it could be overridden I'd be happy.

PR #4257, opened 2018-01-30
Show pipeline scheduled at information in the readable format (#4254)

GaneshSPatil:
**Note**: On the older pipelines page, the `scheduled_at` date formatting is done [here](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/pipelines/index.html.erb#L29) using [moment](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/moment-2.18.1.js) and [moment-duration-format](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/assets/javascripts/lib/moment-duration-format-1.3.0.js)

* Add `moment` and `moment-duration-format` to package.json.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/package.json
+++ b/server/webapp/WEB-INF/rails.new/package.json
@@ -19,6 +19,8 @@
     "lodash": "^4.2.1",
     "lodash-inflection": "^1.5.0",
     "mithril": "^1.1.1",
+    "moment": "^2.20.1",
+    "moment-duration-format": "^2.2.1",
     "repeat": "^0.0.6",
     "underscore.string": "^3.3.4"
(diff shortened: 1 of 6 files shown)


ketan:
There was a recent issue that was raised in https://github.com/gocd/gocd/issues/4107.

Can we add a tooltip with relative time wherever we render the timestamp?

<span title="10 minutes ago">10 Nov 2017 at 12:55:28 Local Time</span>

@naveenbhaskar @eeluke — do you have any thoughts on what's the best way to convey timestamp information to end users?

PR #4258, opened 2018-01-30
Notify on agent runtime status or config status change

maheshp:
**WIP do not merge**


common/src/main/java/com/thoughtworks/go/server/domain/AgentInstances.java
```diff
@@ -40,15 +40,15 @@
...
-        this.changeListener = changeListener;
+    public AgentInstances(AgentStatusChangeListener agentStatusChangeListener) {
+        this.agentStatusChangeListener = agentStatusChangeListener;
         this.systemEnvironment = new SystemEnvironment();
     }
 
-    public AgentInstances(AgentRuntimeStatus.ChangeListener changeListener, SystemEnvironment systemEnvironment, AgentInstance... agentInstances) {
-        this(changeListener);
+    public AgentInstances(SystemEnvironment systemEnvironment, AgentInstance... agentInstances) {
+        this(null);
```
jyotisingh: Since this constructor is being used only in tests so far, its fine to pass along `null` for `AgentStatusChangeListener`, but could you instead pass along that `null` value from tests instead of hardcoding that in the constructor here.

server/src/main/java/com/thoughtworks/go/server/messaging/plugin/AgentStatusChangeNotifier.java
```diff
@@ -0,0 +1,69 @@
...
+        if (isAnyPluginInterestedInAgentStatus()) {
+            AgentNotificationData agentNotificationData = notificationDataFrom(agentInstance);
+            pluginNotificationQueue.post(new PluginNotificationMessage(NotificationExtension.AGENT_STATUS_CHANGE_NOTIFICATION, agentNotificationData));
+        }
+    }
+
+    private AgentNotificationData notificationDataFrom(AgentInstance agentInstance) {
+        return new AgentNotificationData(agentInstance.getUuid(),
+                agentInstance.getHostname(),
+                elasticAgentId(agentInstance),
```
jyotisingh: What do we plan to do with the `elasticAgentId`?

jyotisingh: Reason for asking this is `elasticAgentId` by itself is not very useful. If all we care about is if the agent is elastic, then may be we could pass along a flag for that. If we do care about the `elasticAgentId` then most likely we would also care about the `elasticPluginId` too.

common/src/main/java/com/thoughtworks/go/server/domain/AgentInstances.java
maheshp: Done

server/src/main/java/com/thoughtworks/go/server/messaging/plugin/AgentStatusChangeNotifier.java
maheshp: I don't see a use case for using the elastic_agent_id apart from being able to distinguish between elastic and non-elastic agents. You are right that just the elastic_agent_id might not be of much use.
Changed this to pass a flag `is_elastic` for now.

PR #4262, opened 2018-01-30
Improved Error Handling for Analytics

kierarad:
Instead of just failing silently, we bubble up the errors to the UI.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
@@ -44,7 +44,10 @@ class AnalyticsController < ApplicationController
   end
 
   private
+
   def render_plugin_error e
-    render :text => "Error generating analytics from plugin - #{params[:plugin_id]}", status: 500
+    message = (e.getMessage() =~ /^Interaction with plugin with id/) ? e.getCause().getMessage() : e.getMessage()
+    Rails.logger.error("#{e.getMessage}\n#{com.google.common.base.Throwables.getStackTraceAsString(e)}")
+    render :text => message, status: 500
   end
 end
--- a/server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
+++ b/server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
@@ -44,10 +44,10 @@ class AnalyticsController < ApplicationController
   end
 
   private
-
   def render_plugin_error e
     message = (e.getMessage() =~ /^Interaction with plugin with id/) ? e.getCause().getMessage() : e.getMessage()
     Rails.logger.error("#{e.getMessage}\n#{com.google.common.base.Throwables.getStackTraceAsString(e)}")
-    render :text => message, status: 500
+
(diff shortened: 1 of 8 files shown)


server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
```diff
@@ -44,7 +44,10 @@ def pipeline
   end
 
   private
+
   def render_plugin_error e
-    render :text => "Error generating analytics from plugin - #{params[:plugin_id]}", status: 500
```
maheshp: I am just wondering if we should retain the generic error message over using the exception message. User needs to know that something went wrong, not sure if the specific message from the exception would be useful to them. The logs should provide appropriate information for the admins to take action.

What do you think?

arvindsv: I think it should be both:

1. Case 1: The plugin itself fails to respond properly. GoCD should show a meaningful message.

2. Case 2: The plugin understands that the settings are not provided. It should return a proper view path to an error page, with more details. Using the mechanism we use for displaying graphs.

kierarad: Per your comments @maheshp and @arvindsv I opted to keep a generic error message for the user (which would cover case 1), but go ahead and log the exception message for admins. 

As far as case 2, since we are using the mechanism provided for displaying graphs no changes to core were required. The changes have been made in the plugin.

kierarad:
@arvindsv @maheshp are we ok to merge this?

arvindsv:
@maheshp tried this earlier today and it wasn't working for him. It seemed to show some random message, and more importantly it seemed to be outside of the iframe sandbox.

I tried it just now, and it works. So, I think @maheshp and I will both have to look at it and see if there's any problem. We will take a look. Sorry for the delay, @kierarad.

server/webapp/WEB-INF/rails.new/app/controllers/analytics_controller.rb
```diff
@@ -45,6 +45,9 @@ def pipeline
 
   private
   def render_plugin_error e
+    message = (e.getMessage() =~ /^Interaction with plugin with id/) ? e.getCause().getMessage() : e.getMessage()
```
maheshp: `message` is not used.

server/webapp/WEB-INF/rails.new/webpack/views/analytics/plugin_iframe_widget.js.msx
```diff
@@ -34,8 +36,13 @@
...
     view(vnode) {
-      const attrs = {src: vnode.attrs.model.view()};
-      return <iframe sandbox="allow-scripts" {...attrs}></iframe>;
+      const model = vnode.attrs.model;
+      const attrs = {src: model.view()};
+      const errorAttrs = model.errors() ? {"data-error-text": model.errors()} : {};
+
+      return <div class="frame-container" {...errorAttrs}>
+        <iframe sandbox="allow-scripts" {...attrs}></iframe>
+      </div>;
```
maheshp: The analytics widgets are not tested, can you please add tests around it.

PR #4263, opened 2018-01-31
analytics modals should have an overlay to prevent opening multiple modals

marques-work:
small change to prevent users from opening too many modals


maheshp:
@marques-work I was wondering if we really need these changes. Ideally we should not be able to open multiple modals. I think with analytics we can open multiple modals as the background is not greyed out. Do you think we should fix that?

marques-work:
@maheshp it's true, this could be solved by an overlay. It might be an extra click for the user - but I'd be ok with that as it's a familiar interface

marques-work:
@maheshp so in short, we should probably go that route :).

marques-work:
@maheshp done

PR #4272, opened 2018-01-31
Remove default value for modification username and comment

ketan:
Currently set to `Unknown` and is an invalid value for
modifications of type pipeline dependency


ketan:
@jyotisingh and @akshaydewan — this is merged as per our original discussion. Fixed usages where the assumption was that the fields were non-null.

Issue #3899, opened 2017-09-29
User permissions and cookie


jackTheRipper:
##### Issue Type

Bug Report

##### Summary

When you set some credentials to a user, the cookies aren't cleaned properly.

Because of it, it creates multiple redirections and makes the Gocd's UI unavailable until you cleaned manually the cookies

##### Environment

Chrome Version 61.0.3163.100 (Official Build) (64-bit)

###### Basic environment details

* Mac OS

###### Additional Environment Details

none
##### Steps to Reproduce

1. Ask a new user to login using his Gmail account
2. Grant him access to some specific roles
3 Try to login in Gocd again

##### Expected Results

We should be able to log in again in Gocd after setting the roles without having to clean up the cookies.

##### Actual Results

The browser complains about too many redirections

##### Possible Fix

clean up the browser cookies before redirecting to the login page

##### Log snippets

none

##### Code snippets/Screenshots

none

##### Any other info

none

jackTheRipper:
This issue becomes really annoying. Any updates?

jackTheRipper:
The new release doesn't help. The issue is still here. Will it be fixed at some point?

maheshp:
@jackTheRipper can you let us know the plugin used for authentication. And is this issue only for a first time login?

jackTheRipper:
Hi @maheshp, the plugin we use is the one for the OAuth with Google. It happens everytime we change / update the user roles.

maheshp:
@jackTheRipper I was not able to reproduce this issue, need some help with reproducing it.

These are the steps I followed,
1. I am using version 2.0.0 of [Google OAuth Plugin](https://github.com/gocd-contrib/google-oauth-authorization-plugin/releases/tag/2.0.0) built using the new Authorization Plugin Endpoint. 
2. I was able to configure the plugin as per the given [instructions](https://github.com/gocd-contrib/google-oauth-authorization-plugin/blob/master/INSTALL.md)
3. Post all the configuration I am able to login to application using my google credentials.
4. Upon login a user is created(say `google_user`)  and I can see the user details in the User Summary page.
5. Now as an admin (on a different browser) I add roles to the user `google_user`.
6. As a `google_user` I am able to logout and login multiple times without deleting any of the cookies.

Please let me know the steps followed by you were different from what I have mentioned.

jackTheRipper:
@maheshp : we are using the plugin `Google oauth authorization` for Go 17.5.0 (version 1.0.0-1)

[...]

PR #4274, opened 2018-02-01
Fixed infinite redirection on unsuccessful reauth using web based auth plugin

bdpiparva:
* Change PreAuthenticatedAuthenticationProvider to throw
  BadCredentialsExceptions on unsuccessful authentication. This will
  ensure the existing session with PreAuthenticationToken to be
  invalidated and redirected to login page. Invalidating the session
  ensures existing AuthToken is removed from Security context and thereby
  avoids infinite redirection.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/security/providers/PreAuthenticatedAuthenticationProvider.java
+++ b/server/src/main/java/com/thoughtworks/go/server/security/providers/PreAuthenticatedAuthenticationProvider.java
@@ -74,10 +74,15 @@ public class PreAuthenticatedAuthenticationProvider implements AuthenticationPro
     private Authentication doAuthenticate(PreAuthenticatedAuthenticationToken preAuthToken) {
         String pluginId = preAuthToken.getPluginId();
 
-        AuthenticationResponse response = authenticateUser(preAuthToken);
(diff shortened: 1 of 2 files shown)


bdpiparva:
https://github.com/gocd/gocd/issues/3899

JorritSalverda:
I'm testing experimental build 18.2.0-6000 that has this fix. Will update once I'm passed the 30 minutes of idle time.

JorritSalverda:
So far so good, I'll upgrade more of our go.cd servers to this version and see if anyone still complains about the redirect issue.

ketan:
A word of caution: this is an experimental build, warrant does not apply, and some things may not.qork as expected :)

JorritSalverda:
Could it be that something changed with respect to secrets in this experimental build? I'm using `envsubst` to replace variable placeholders in a file with values. That worked fine, but suddenly it's replaced with an empty string.

ketan:
No stuff wrt secrets has changed. Did you restore the cipher file properly?

JorritSalverda:
Downgrading doesn't solve it, so it's probably an issue in my agent. The logs properly show the environment variable to be set. If there's a real issue I'll create a separate ticket instead of polluting this one.

PR #4278, opened 2018-02-01
Have same tag-name to define publish artifact configurations.

varshavaradarajan:


varshavaradarajan:
The builtin artifacts representation has been changed -

- A type attribute has been introduced.

Built Artifacts -
```
<artifact type="build" src="dist" destination="dest" />
```

Test Artifacts -
```
<artifact type="test" src="reports" destination="coverage" />
```

Issue #4111, opened 2017-12-19
Search feature in dashboard SPA


yankurk:
##### Summary

There should be a provision to search for pipelines based on their names in the dashboard SPA. 
- The search should result in a view of the dashboard which includes only pipelines whose names contain the search text. 
- The view should be updated when the search text is updated.
- The search result should also include the count of matched pipelines.

yankurk:
#4081

arvindsv:
@GaneshSPatil I see that it is implemented in #4275. However, when a search is active, the changes popup doesn't work. On 3e28f3316ce53fd6db85982c92a78d96371dbc26.

arvindsv:
Also, you're right: The new one, without throttling and debouncing feels very "jerky". The existing one is a little slower but smoother. The new one changes too quickly as you're typing.

arvindsv:
The CSS: It has too much padding. If you type "g" or "y", it gets cut off.

arvindsv:
Also, as implemented, it looks like it is case sensitive, unlike the older one.

arvindsv:
When search is active, pause functionality also doesn't work. I think I know what's happening. The way that the search is implemented, it seems to be still holding on to the info of the original dashboard. Try this:

1. Have four pipelines in this order:

```
- Pipeline_group_1
  - Pipeline: ABC_1
  - Pipeline: ABC_2

- Pipeline_group_2
  - Pipeline: DEF_1
  - Pipeline: DEF_2
```

2. Search for DEF.

3. Pause DEF_2.

You won't see any change in DEF_2. But, if you remove the search now, you'll see that ABC_2 is paused. I think the same thing is happening to the changes popup.

rajiesh:
Verified on 18.2.0 (6211-48d5a54cd6525e54b4692cb757a6ad9896e46392). This fix has made the search pipeline field to not accept Upper case, which is different from what the old page did, however closing this issue.

PR #4280, opened 2018-02-01
Introduce Dashboard polling

GaneshSPatil:
* Introduce AJAX API polling for new dashboard SPA
* Fetch latest dashboard data after any pipeline operation (unpause/unlock)


ketan:
We need to implement the 304 and a few other things. If you find specific
problems, can you open separate issues for them?

arvindsv:
Ok. From now on? [Please say yes. Otherwise I have [so many](https://github.com/gocd/gocd/issues/4111#issuecomment-362698482)].

Issue #4121, opened 2017-12-21
Improvements to Dashboard API V2


GaneshSPatil:
- [ ] Include html url links along with api links for:
    - [ ] Pipeline History
           - `history_url` links to Pipeline history API. 
           - Add `history_html_url` which links to Pipeline History Page.
    - [ ] Pipeline Settings
           - Currently `settings_path` links to Pipeline settings HTML page. 
           - Change `settings_path` to point to Pipeline Config API.
           - Add `settings_html_path` to point to Pipeline settings HTML page.
    -  [ ] Stage Details
           - Currently, stage instance`self` links to stages get instance API. 
           - Add stage instance `details_html_url` to point to Stage Details HTML page.

~~- [ ] Dashboard API does not return the previous stage instance related information.
       - Previous Stage instance information is necessary to show the previous stage status when the current run is in progress.~~ @arvindsv and @jyotisingh voted this feature out while building the API.

arvindsv:
@GaneshSPatil or @yankurk: I think there's a potential issue around the usage of the cache for this. Specifically [GoDashboardService.java#L101](https://github.com/gocd/gocd/blob/6752c191c77fc00e52d30326bbcd97e137734548/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java#L101) and [GoDashboardService.java#L111](https://github.com/gocd/gocd/blob/6752c191c77fc00e52d30326bbcd97e137734548/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java#L111). `cache.get` is not a safe operation here. In fact, the `get` method should be marked internal and used only in tests, or completely removed to be replaced with usages of `allEntriesInOrder`.

`allEntriesInOrder` is the only safe method which can be called to get entries from the cache. That's because the cache [can be changed](https://github.com/gocd/gocd/blob/6752c191c77fc00e52d30326bbcd97e137734548/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java#L49-L58) by other threads while it is being used. Whenever a cache update finishes, the `cacheHasChanged` method is called and the `dashboardPipelines` object is updated. A double buffer, essentially.

So, a `cache.get` can return null. That's why the usages in GoDashboardService should be rewritten to use `allEntriesInOrder`.

/cc @jyotisingh

maheshp:
[...]

PR #4287, opened 2018-02-02
Do not expose GoDashboardCache internal data structure for access (#4121)

GaneshSPatil:
* Use 'allEntries' method instead of 'cache.get' to get all the entries safely from the cache

More information: https://github.com/gocd/gocd/issues/4121#issuecomment-360196822

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java
@@ -20,7 +20,10 @@ import com.thoughtworks.go.config.CaseInsensitiveString;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-import java.util.*;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 /* Understands how to cache dashboard statuses, for every pipeline. */
 @Component
@@ -39,11 +42,7 @@ public class GoDashboardCache {
     public GoDashboardCache(TimeStampBasedCounter timeStampBasedCounter) {
         this.timeStampBasedCounter = timeStampBasedCounter;
         cache = new LinkedHashMap<>();
-        dashboardPipelines = new GoDashboardPipelines(new ArrayList<>(), timeStampBasedCounter);
-    }
-
-    public GoDashboardPipeline get(CaseInsensitiveString pipelineName) {
-        return cache.get(pipelineName);
+        dashboardPipelines = new GoDashboardPipelines(new HashMap<>(), timeStampBasedCounter);
     }
 
     public void put(GoDashboardPipeline pipeline) {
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCache.java
```diff
@@ -57,12 +53,16 @@ public void replaceAllEntriesInCacheWith(List<GoDashboardPipeline> newPipelinesT
         cacheHasChanged();
     }
 
-    public GoDashboardPipelines allEntriesInOrder() {
+    public GoDashboardPipelines allEntries() {
         return dashboardPipelines;
     }
 
+    public List<GoDashboardPipeline> allEntriesInOrder() {
```
arvindsv: I would remove this method and all the corresponding tests. If there is no need for it, why keep it?

server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
```diff
@@ -91,24 +88,25 @@ public void updateCacheForAllPipelinesIn(CruiseConfig config) {
     }
 
     private GoDashboardPipelineGroup dashboardPipelineGroupFor(PipelineConfigs pipelineGroup, PipelineSelections pipelineSelections, Username user) {
-        GoDashboardPipelineGroup goDashboardPipelineGroup = new GoDashboardPipelineGroup(pipelineGroup.getGroup(), resolvePermissionsForPipelineGroup(pipelineGroup));
+        GoDashboardPipelines allPipelines = cache.allEntries();
```
arvindsv: You should really be doing this once, in `allPipelineGroupsForDashboard` method.

arvindsv:
For good measure, do this too:

```diff
diff --git a/server/src/main/java/com/thoughtworks/go/domain/cctray/CcTrayCache.java b/server/src/main/java/com/thoughtworks/go/domain/cctray/CcTrayCache.java
index 178a2c65cf..e6b325eb41 100644
--- a/server/src/main/java/com/thoughtworks/go/domain/cctray/CcTrayCache.java
+++ b/server/src/main/java/com/thoughtworks/go/domain/cctray/CcTrayCache.java
@@ -38,7 +38,7 @@ public class CcTrayCache {
         this.orderedEntries = new ArrayList<>();
     }
 
-    public ProjectStatus get(String projectName) {
+    ProjectStatus get(String projectName) {
         return cache.get(projectName);
     }
 
@@ -52,7 +52,7 @@ public class CcTrayCache {
         cacheHasChanged();
     }
 
-    public void replaceAllEntriesInCacheWith(List<ProjectStatus> projectStatuses) {
+    void replaceAllEntriesInCacheWith(List<ProjectStatus> projectStatuses) {
         this.cache.clear();
         this.cache.putAll(createReplacementItems(projectStatuses));
         cacheHasChanged();
```

arvindsv:
Looks good now. Merge when you want.

PR #4299, opened 2018-02-06
Fixed random rspec failures

maheshp:
* Analytics capabilities constructor was changed from from (boolean,
  boolean) to (boolean, list). The plugin info representer spec
  continued to pass (boolean, boolean). Due to this, the 'TrueClass' was
  extended to include Enumerable and List modules. This led to issues
  when flattening a array with boolean value 'true' during asset compile.
* Some minor refactoring.


server/webapp/WEB-INF/rails.new/app/views/analytics/index.html.erb
```diff
@@ -1 +1 @@
-<div class="analytics-container" data-plugin-ids="<%= @plugin_ids.to_json %>"></div>
+<div class="analytics-container" data-supported_dashboard_metrics="<%= @supported_dashboard_metrics.to_json %>"></div>
```
ketan: Data attributes should use hyphens, not underscores.

maheshp: Oops, changing it

PR #4296, opened 2018-02-06
New Dashboard: Fixes for pause popup

yankurk:
Issues fixed in this PR related to pause popup in the new dashboard:
- Auto focus on the text field when the popup opens
- Trigger pause and close popup when enter is pressed in the text field
- Close popup upon pressing escape key
- Do not retain pause message entered once the popup is closed

PR #4304, opened 2018-02-07
Improve modal ui

akshaydewan:
It would be good to merge #4296 before this so that I can rebase and incorporate those changes.

Changes in this PR:
* Fix the modal height when the view scrolls
* Made the header sticky when it scrolls
* Make the close button part of the header so that it doesn't scroll out of view

Needs to be tested on different browsers and OSes. I have tested on the latest Chrome, FF and Safari on MacOS.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_modal.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_modal.scss
@@ -15,15 +15,20 @@
  */
 
 .reveal {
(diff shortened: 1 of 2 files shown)


ketan:
Can we make the footer sticky as well, so the buttons are always accessible?

akshaydewan:
> Can we make the footer sticky as well, so the buttons are always accessible?

Makes sense. I'll give it a shot.

akshaydewan:
Added fixes for IE and Edge. Checked on IE 11 and Edge 14 on Windows 10.

PR #4341, opened 2018-02-13
Personalize api fix

### LINKED ISSUE #4339: Personalize API improvement
##### Issue Type

- Feature enhancement

##### Summary

The personalize API returns selected pipeline names based on blacklist value. The list should contain names of pipelines that the user had previously checked, regardless of blacklist value.

##### Steps to Reproduce

1. Create three pipelines, say `one`, `two` and `three` under a pipeline group say `first`.
2. Under personalize dropdown select checkbox for `Show newly created pipelines`
3. Also check pipelines `one` and `two`
4. Click on apply.
4. Make an api call as such:
```
curl 'http://localhost:8153/go/api/internal/pipeline_selection' -H 'Accept: application/vnd.go.cd.v1+json' -u 'username:password'
```

##### Expected Results

The api response should be:
```
{
  "selections": [
    "one",
    "two",
  ],
  "blacklist": true,
  "pipelines": {
    "first": [
      "one",
      "two",
      "three",
    ]
  }
}
```

##### Actual Results

The api response is:
```
{
  "selections": [
    "three"
  ],
  "blacklist": true,
  "pipelines": {
    "first": [
      "one",
      "two",
      "three",
    ]
  }
}
```

yankurk:
#4339

### DIFF
--- a/api/api-dashboard-v2/src/main/java/com/thoughtworks/go/apiv2/dashboard/DashboardControllerDelegate.java
+++ b/api/api-dashboard-v2/src/main/java/com/thoughtworks/go/apiv2/dashboard/DashboardControllerDelegate.java
@@ -66,7 +66,7 @@ public class DashboardControllerDelegate extends ApiController {
         Long userId = currentUserId(request);
         Username userName = currentUsername();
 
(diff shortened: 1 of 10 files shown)


yankurk:
The current implementation of personalized view has been left untouched in this PR. Ideally, we would want to change the current implementation to match the implementation adopted in the API.

Issue #4081, opened 2017-12-11
Dashboard SPA


GaneshSPatil:
### Pipelines Dashboard Header
- [x] Implement Pipeline Search (#4111 )
- [ ] Implement Pipeline Selection (Personalize View)
  - [x] Internal API (#4235) (#4312)
  - [ ] Dropdown on dashboard (#4252)
  - [ ] SPA for managing personalizations (#4236)

### Pipelines Groups
- [x] Show Pipeline Groups Depending on the permissions

### Pipeline
- [ ] Dashboard tile (#4254)
- [ ] ~~~Show Current Pipeline Status (Passed: LAST_STAGE, Building: CURRENT_STAGE)~~~
- [ ] ~~~Show previous stage run status when the stage is currently running.~~~
- [x] Errors and Warnings
  - [x] Frontend (#4232) (#4265)
  - [x] API (#4231) (#4256)
- [x] Implement Pause a pipeline
  - [x] API (#4076) (#4188)
  - [x] Frontend (#4254)
- [x] Implement Unpause a pipeline 
  - [x] API (#4076) (#4188)
  - [x] Frontend (#4254)
- [x] Implement Pipeline unlock feature.
  - [x] API (#4237) (#4243)
  - [x] Frontend (#4254)
- [x] Implement Trigger a pipeline
- [x] For Trigger with options
  - [x] API to perform the trigger (replacement for https://api.gocd.org/current/#scheduling-pipelines) (#4346) (#4319)
- [x] For Trigger with options, show options for trigger
  - [x] API (#4233) (#4276) (#4314)
    - [x] Materials
    - [x] Environment Variables
    - [x] Secure Environment Variables
  - [x] Frontend (#4234)
    - [x] Materials (#4320)
    - [x] Environment Variables (#4328)
    - [x] Secure Environment Variables (#4328)
- [x] Material Search
  - [x] API  (#4270) (#4284)
  - [x] Frontend  (#4271)

### Icons
- [ ]  Pipeline Group
- [ ] Pipeline
- [ ] Pipeline Settings
- [ ] Compare
- [ ] Changes Dropdown
- [ ] Stage Status
    - [ ] Failed
    - [ ] Passed
    - [ ] Running
- [ ] Failed icon in the status when the previous run of the stage has failed and current run is in 
progress.

### Buttons
- [x] Trigger
- [x] Trigger with options
- [x] Pause
- [x] Unpause

GaneshSPatil:
* Technical debt:
[...]

PR #4345, opened 2018-02-13
Fix dashboard pipeline search performance issues

GaneshSPatil:
* Use debounced dashboard search operation

Related #4081


yankurk:
I tested it out with build.gocd.org data. The implementation also looks good to me.

PR #4354, opened 2018-02-15
Personalize api fix

yankurk:
* If no personalization was previously setup the api currently returns a 500 internal error.
* The api will now return all the pipelines a user can see if blacklist value is true and it returns no pipelines if blacklist value is false, given that personalization wasn't setup for a user earlier.


yankurk:
Related to #4339

PR #4357, opened 2018-02-15
Pass the trackingTool to MaterialRevisionWidget

akshaydewan:
It was probably lost during refactoring or merging.


akshaydewan:
Added a test and rebased with master.

Issue #4236, opened 2018-01-25
View for managing pipeline personalizations


ketan:
##### Issue Type

- Feature enhancement

##### Summary

See https://github.com/gocd/gocd/issues/4235. Need a view (and possibly modal on the dashboard for quick access) for managing pipeline personalizations.

GaneshSPatil:
This feature has been introduced in GoCD as part of release v18.3.0.

PR #4358, opened 2018-02-15
Implement Personalize View for quicker Dashboard (#4236) (#4081)

GaneshSPatil:


yankurk:
Tested out the PR locally for different functionalities. Looks good to me. There is one functionality pending wherein given the personalize dropdown is visible and a user clicks anywhere else on the screen, the dropdown should be hidden.

PR #4359, opened 2018-02-15
Make improvements to all modals on the old pages

ketan:
They now look more like the modals on the new pages with a lot more room
on it


ketan:
Visually inspected all the usages of the modal that I could find, everything looks good.

Issue #3989, opened 2017-11-07
StageConverter: Time looks like Zulu-Time, but isn't


ralfhergert:
https://github.com/gocd/gocd/blob/3cfb0a62fccabcc9e345df39cbede86c6cbea6db/plugin-infra/go-plugin-access/src/com/thoughtworks/go/plugin/access/notification/v1/StageConverter.java#L31

The suffix 'Z' is defined as 'Zulu time' which means the time should be rendered as being in the timezone UTC+-0. But this pattern and how it is used renders in fact the time using the default locale. If the default locale is not the same as UTC+-0 appending the letter 'Z' creates in fact a wrong date.

A possible solution can be to actually include the correct timezone. For instance with the pattern:
`new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ")`
in which Z stands for an RFC822 timezone definition.

arvindsv:
```
$ TZ=Europe/Paris jrunscript -l js -e "print((new java.text.SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSSZ\")).format(new java.sql.Timestamp(java.lang.System.currentTimeMillis())))"
2017-11-07T17:43:56.852+0100
$ TZ=Europe/Paris jrunscript -l js -e "print((new java.text.SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")).format(new java.sql.Timestamp(java.lang.System.currentTimeMillis())))"
2017-11-07T17:44:07.795Z

$ TZ=UTC jrunscript -l js -e "print((new java.text.SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSSZ\")).format(new java.sql.Timestamp(java.lang.System.currentTimeMillis())))"
2017-11-07T16:44:19.677+0000
$ TZ=UTC jrunscript -l js -e "print((new java.text.SimpleDateFormat(\"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\")).format(new java.sql.Timestamp(java.lang.System.currentTimeMillis())))"
2017-11-07T16:44:24.176Z
```

Agree.

/cc @bdpiparva Thoughts?

bdpiparva:
[...]

PR #4360, opened 2018-02-15
NotificationExtension v3 use 'UTC' timestamps #3989

maheshp:
* Changed all timestamps in Stage Status notification to use 'UTC'.

### DIFF
--- a/plugin-infra/go-plugin-access/src/test/java/com/thoughtworks/go/plugin/access/notification/v3/JsonMessageHandler3_0_Test.java
+++ b/plugin-infra/go-plugin-access/src/test/java/com/thoughtworks/go/plugin/access/notification/v3/JsonMessageHandler3_0_Test.java
@@ -36,7 +36,6 @@ import java.sql.Timestamp;
 import java.text.SimpleDateFormat;
 import java.util.*;
 
-import static com.thoughtworks.go.plugin.access.notification.v2.StageConverter.DATE_PATTERN;
 import static java.util.Arrays.asList;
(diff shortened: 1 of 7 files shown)


plugin-infra/go-plugin-access/src/test/java/com/thoughtworks/go/plugin/access/notification/v3/JsonMessageHandler3_0_Test.java
```diff
@@ -330,7 +331,7 @@ private String errorMessageForNotifyResult(String message) {
     }
 
     private Date getFixedDate() throws Exception {
-        return new SimpleDateFormat(DATE_PATTERN).parse("2011-07-13T19:43:37.100Z");
+        return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ").parse("2011-07-13T19:43:37.100+0530");
```
bdpiparva: Can we use calendar to build date object -

```
Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("GMT"));
        calendar.set(2011,07,13,19,43,37);
        calendar.set(Calendar.MILLISECOND,100);
        return calendar.getTime();
```

If we take above approach then we can remove https://github.com/gocd/gocd/pull/4360/files#diff-21f722c49e082ef198454f08e361be18R358 and https://github.com/gocd/gocd/pull/4360/files#diff-21f722c49e082ef198454f08e361be18R104 from test since we have a fixed date in `createPipeline ()`

PR #4363, opened 2018-02-15
Closes personalize dropdown when user clicks anywhere else on the screen

yankurk:
This PR adds to the functionalities introduced in #4358


GaneshSPatil:
The drop-down needs to be closed on clicking `apply` button.
Can you make that change as part of the same PR and remove the showing spinner code?

PR #4367, opened 2018-02-15
Improve the look and feel of the trigger-with-options modal

ketan:


ketan:
Before:

screen shot 2018-02-15 at 11 06 43 pm

After:

screen shot 2018-02-15 at 11 07 04 pm

Issue #4348, opened 2018-02-14
New Dashboard Issues


GaneshSPatil:
#### Issues with new Dashboard:

- [x] Performance issues with stage status animation. (`Building`, `Failing`).
- [x] Changes popup positioning issue in case of multiple pipeline instances.
- [x] Show server Scheduled at time on hover of local scheduled_at time.
- [x] Show action disabled mouse on hover of disabled gear icon.
- [x] Implement Etag for Dashboard API.
- [x] Disable Trigger buttons when the first stage is building.
- [x] Disable Trigger buttons when the pipeline is locked.
- [x] Disable Trigger buttons when the pipeline is paused.

yankurk:
Some other issues:

- [ ] Remove pipeline operations related buttons (trigger, trigger-with-options and pause/unpause) for users who don't have operate permissions (#4376)
- [ ] If no permissions is setup for a pipeline group, by default, all users should have operate permissions (#4374)

PR #4373, opened 2018-02-19
[New Dashboard] Changes related to pipeline operation related buttons

yankurk:
#4348

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/webpack/views/dashboard/pipeline_widget_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/webpack/views/dashboard/pipeline_widget_spec.js
@@ -111,26 +111,37 @@ describe("Dashboard Pipeline Widget", () => {
   });
 
   describe("Pipeline Header", () => {
-    beforeEach(mount);
     afterEach(unmount);
 
     it("should render pipeline name", () => {
+      mount();
       expect($root.find('.pipeline_name')).toContainText('up42');
     });
 
     it("should show pipeline name on hover", () => {
+      mount();
       expect($root.find('.pipeline_name').get(0).title).toEqual('up42');
     });
 
     it("should link history to pipeline history page", () => {
+      mount();
       expect($root.find('.pipeline_header>div>a')).toContainText('History');
       const expectedPath = `/go/tab/pipeline/history/${pipelinesJson[0].name}`;
       expect($root.find('.pipeline_header>div>a').get(1).href.indexOf(expectedPath)).not.toEqual(-1);
     });
 
     it("should link to pipeline settings path", () => {
+      mount();
       const expectedPath = pipeline.settingsPath;
       expect($root.find('.edit_config').get(0).href.indexOf(expectedPath)).not.toEqual(-1);
+    });
+
+    it("should align history link to the left if pipeline operation buttons aren't present", () => {
+      mount(false, false, {}, {}, false, false);
+      const historyLink      = $root.find('.pipeline_header>div>a');
+      const operationButtons = $root.find('.pipeline_operations');
+      expect(operationButtons).not.toBeInDOM();
+      expect(historyLink).toHaveClass("no_operation_buttons");
     });
 
   });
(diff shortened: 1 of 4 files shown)


yankurk:
As part of this PR:
* Trigger buttons are disabled if a pipeline is paused
* Trigger and pause/unpause buttons are not shown if user doesn't have operating rights

server/webapp/WEB-INF/rails.new/spec/webpack/views/dashboard/pipeline_widget_spec.js
```diff
@@ -540,6 +547,17 @@ describe("Dashboard Pipeline Widget", () => {
         const modalTitle = $('.modal-title:visible');
         expect(modalTitle).toHaveText(`${pipeline.name} - Trigger`);
       });
+
+      it("should disable button when pipeline is paused", () => {
+        unmount();
```
ketan: Can we not mount in a setup method if it requires us to unmount before a test?

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_operations_widget.js.msx
```diff
@@ -148,26 +148,33 @@ const PipelineOperationsWidget = {
     }
 
     let isTriggerDisabled = false;
-    if (!pipeline.canOperate || pipeline.isFirstStageInProgress() || pipeline.isLocked) {
+    if (!pipeline.canOperate || pipeline.isFirstStageInProgress() || pipeline.isLocked || pipeline.isPaused) {
       isTriggerDisabled = true;
     }
 
+    let operationRelatedButtons = "";
+    if(pipeline.canAdminister || pipeline.canOperate) {
```
ketan: Run lint?

yankurk:
There is one case I overlooked in this PR. It is related to this [commit ](https://github.com/gocd/gocd/commit/5caef8ecee0c71c5bbed3d2a6e46930fd4f0201f) which says that if there are operators defined for the first stage of a pipeline, those operators override the operators defined in the pipeline group level. However, the operators defined at pipeline group level can still pause the pipeline in such scenarios. This means that trigger and pause are treated differently. I'm working on a fix for such scenarios on the new dashboard.

yankurk:
These are the states that operation buttons can be in the new dashboard:
1) For users with operate permissions (and by default for admins):
screen shot 2018-02-20 at 12 11 08 pm

2) For users with pause permission but no operate permission (see 5caef8ecee0c71c5bbed3d2a6e46930fd4f0201f)
screen shot 2018-02-20 at 12 12 37 pm

3) For users with only view permission
screen shot 2018-02-20 at 12 14 58 pm

4) Trigger buttons disabled due to pipeline being paused
screen shot 2018-02-20 at 12 11 35 pm

5) Trigger buttons disabled because the first stage is still in progress
screen shot 2018-02-20 at 12 12 00 pm

ketan:
@yankurk — Removing buttons from the page does not feel right, because users don't know why they do not see the button. It's better to show the button, but have it disabled, preferably with a tooltip to say why it's been disabled.

In most cases, when something does not show up, users may think that they have done something wrong, or there's a bug with the system.

yankurk:
@ketan -- The decision to remove the buttons is also tied to whether we show pipeline and pipeline group edit gear icon for non admin users #4375. If we decide to keep the trigger buttons in disabled state, we need to keep the edit gear icon in disabled state for non-admins too.

GaneshSPatil:
@yankurk -- I guess, we decided not to remove any element and disable them based on the permissions.
Same applies to settings gear and operation buttons too.

maheshp:
I was thinking bout guest users on build.gocd, there is no way a guest user will be able to operate or edit a pipeline. In these cases wondering if it makes sense to hide those buttons over disabling them?

arvindsv:
As long as it is clear why it is disabled, in the tool tip, I think it's fine to keep them. Showing a proper reason ("You do not have permission" vs "This pipeline is about to trigger") might not be very easy.

naveenbhaskar:
Disabling and hiding elements 
case 1:  User has no way to do anything with the elements. In that case, we dont have to show the element. If we show the element and then disable it, that means, user can do something to make it enabled. for eg: config button can be used only with permission. if there is no permission for a user, then lets not show it.

case 2 : user can do some action to make the button enabled. In this case, button has to be visible

if there is something which cant be used by the user, there is no point in showing that. even if we show a tooltip there, its just an element in the UI which is not usable at all.

arvindsv:
@naveenbhaskar The action that a user can perform to make it enabled is to convince an admin they need it. :)

GoCD's old dashboard, for a view user, looks quite bare and it doesn't show the possibilities that exist. I'd strongly consider disabling it, rather than hiding it. If we're hiding it, we can hide it by changing `opacity` to 0, so that the box sizes won't get affected, etc.

Finally, it is your call though, since you need to consider consistency (gear icon, etc).

GaneshSPatil:
Closing this PR as we have decided to show the buttons and disabled message tooltip on hover

PR #4375, opened 2018-02-19
[New dashboard] Hide pipeline settings and pipeline group settings icon for non admin users

yankurk:
* Showing the settings icon in disabled mode for non admin users makes sense only if it is possible for such users to elevate previleges for themselves. Since this is not possible it doesn't make sense to show the settings icon. Hence, it was removed.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
@@ -48,9 +48,10 @@ const DashboardWidget = {
         );
       });
 
-      const pipelineGroupSettingsIcon = (pipelineGroup.canAdminister)
-        ? (<a href={pipelineGroup.path} class="edit_config pipeline-group_edit-config"></a>)
-        : (<a href="#" class="edit_config pipeline-group_edit-config disabled"></a>);
+      let pipelineGroupSettingsIcon;
+      if (pipelineGroup.canAdminister) {
+        pipelineGroupSettingsIcon = (<a href={pipelineGroup.path} class="edit_config pipeline-group_edit-config"></a>);
+      }
 
       return (
         <div class="pipeline-group">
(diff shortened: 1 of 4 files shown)


yankurk:
Linking #4348

GaneshSPatil:
@yankurk -- according to the new way of writing an SPA, we decided to show the buttons/links onto the Page and disable them based on the permissions.

yankurk:
@GaneshSPatil Why is it that we decided to disable it? I talked to @naveenbhaskar and he says that since unless an admin gives view users the appropriate permissions, those users will never be able to operate the pipeline. It would make sense to disable those buttons if users themselves could at some point change permissions to operate the pipeline.  
For example, 
screen shot 2018-02-19 at 2 33 08 pm
In the screenshot, delete link is disabled because there are still pipelines in the pipeline group. The link will be active if the admin removes those pipelines. Since there is a possibility that the delete link could activate based on actions of the logged in user, it makes sense to disable it. 

There is no way for a logged in view users to update permissions. Therefore, looks like those operations related buttons need to be removed.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_header_widget.js.msx
```diff
@@ -40,7 +40,7 @@ const PipelineHeaderWidget = {
       const settingsPath = vnode.attrs.isQuickEditPageEnabled ? pipeline.quickEditPath : pipeline.settingsPath;
       settingsButton     = (<a class={`edit_config`} href={settingsPath}/>);
     } else {
-      settingsButton = (<a class={`edit_config disabled`}/>);
+      settingsButton = "";
```
ketan: Leaving it undefined is better.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
```diff
@@ -50,7 +50,7 @@ const DashboardWidget = {
 
       const pipelineGroupSettingsIcon = (pipelineGroup.canAdminister)
         ? (<a href={pipelineGroup.path} class="edit_config pipeline-group_edit-config"></a>)
-        : (<a href="#" class="edit_config pipeline-group_edit-config disabled"></a>);
+        : "";
```
ketan: leaving it undefined is better.

GaneshSPatil:
@yankurk -- closing this PR as we've decided to show the pipeline operations buttons and show the disabled message tooltip on hover.

Issue #4374, opened 2018-02-19
[New Dashboard] Pipeline groups without any permissions defined should be operable for other users


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Pipeline groups without any permissions defined should be operable for other users

###### Basic environment details

* Go Version: `18.2.0 (6115-1db2ab42cf730dc6f47f7426642e6d78cef68409)`

##### Steps to Reproduce

1.  Create 2 pipeline groups - G1 and G2
2. Enable admin permissions for G1 to user U1, but leave G2 without any permissions
3. Login as U1(or any other valid users) and visit the `http://go-server/go/dashboard`
4. User do not have operate permissions on G2

##### Expected Results

User should have operate permissions on G2 since its not restricted by any permissions

rajiesh:
Verified on 18.2.0 (6211-48d5a54cd6525e54b4692cb757a6ad9896e46392)

PR #4382, opened 2018-02-19
Provides operate permissions to all users if no permissions are set on pipeline group

yankurk:
To maintain consistency with current implementation, if no permissions are set on a pipeline group all users are given operate permissions instead of only pipeline admins.


yankurk:
This fixes issue #4374 .

arvindsv:
@yankurk and I worked on this. Merging.

PR #4386, opened 2018-02-20
Expose new Logging API's to plugin to use slf4j parameterized messages

maheshp:


maheshp:
Yeah will verify with couple of plugins, ideally it should not break since we are not changing the existing API.

maheshp:
Verified logging on ldap authentication and file based authentication plugin.

PR #4388, opened 2018-02-20
Small fix to artifact tab CSS

arvindsv:
Before: See [this tab](https://build.gocd.org/go/tab/build/detail/performance/304/benchmark/1/thread_gc_analysis#tab-artifacts).

After:


naveenbhaskar:
Looks good to me. Pls merge

PR #4389, opened 2018-02-21
Copy over some more code from the test-set plugin to improve test detection in intellij

ketan:


arvindsv:
@ketan I'm not sure I understand this, but IntelliJ did. And did the right thing. It looks ok to me, from a behavior perspective.

For context:

Some tests show up when you look for production usages, since they're a new source set in Gradle. A patch like the one below would tell IntelliJ that they're tests and not source. This PR also does something similar.

```diff
diff --git a/build.gradle b/build.gradle
index 17d05ebc8f..7f694dad57 100644
--- a/build.gradle
+++ b/build.gradle
@@ -211,6 +211,15 @@ subprojects {
   apply plugin: 'base'
   apply plugin: "com.github.hierynomus.license"

+  apply plugin: 'idea'
+  idea {
+    module {
+      testSourceDirs += file('src/test-fast/java')
+      testSourceDirs += file('src/test-integration/java')
+      testSourceDirs += file('src/test-shared/java')
+    }
+  }
+
   license {
     skipExistingHeaders = true
     ext.year = Calendar.getInstance().get(Calendar.YEAR)
```

ketan:
> Some tests show up when you look for production usages, since they're a new source set in Gradle. A patch like the one below would tell IntelliJ that they're tests and not source. This PR also does something similar.

We're managing sourcesets (and dependencies between them) using the (forked) testset plugin, so it needs to be managed using that plugin, instead of doing it in a gradle file.

PR #4390, opened 2018-02-21
#4081 - JSON streaming using Jackson APIs.

ketan:


arvindsv:
1. We need to consider what happens when something throws an exception in the middle of writing. I guess client-side JS, etc. will need to handle it and consider it invalid. They probably do, but worth checking.

2. In `GoDashboardPipelineGroup#etag`, the usage of `permissions.hashCode()` is a little iffy. It's not guaranteed to be really unique. It should work, most times though.

ketan:
I was just looking for a cheap way to compute an etag, without really concatenating a huge string. `hashCode` seemed good enough, do you have a better idea about how etag for permissions can be computed?

ketan:
I suppose the only place where long string appears with `Users` is the `AllowedUsers` implementation. I suppose it should not be a huge string (probably a hundred users, at best). Plus we can probably throw lazy initialization of the etag, if that's fine with you.

arvindsv:
At this scale, it should be fine. A small change in users or permissions is unlikely to be enough for the integer to roll over and still reach the same value. I'd leave it.

> Plus we can probably throw lazy initialization of the etag, if that's fine with you.

What do you mean?

ketan:
> > Plus we can probably throw lazy initialization of the etag, if that's fine with you.
> 
> What do you mean?

Assuming that the permissions model does not change between requests, we can compute something that's a more precise (than `hashcode`, and probably more expensive to compute) and store it as a local variable on the permission object, so as to not compute the `Permissions#etag` on every request.

arvindsv:
@ketan Yes. Having a lazy `Permissions#etag` would be fine.

PR #4391, opened 2018-02-21
Changes stage cancelled sign color

yankurk:
After this change cancelled stages look like:
screen shot 2018-02-21 at 12 19 55 pm
instead of:
screen shot 2018-02-21 at 12 20 34 pm

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
@@ -846,6 +846,7 @@ body {
     &.cancelled {
       @include icon-before($cancelled-icon, 12px, 0);
       background: $building;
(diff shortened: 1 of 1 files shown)


GaneshSPatil:
@yankurk -- does it make sense to change all the occurrences of `#333` with  variable [`tab-head-color `](https://github.com/yankurk/gocd/blob/cd51fe8af6d6d94e1a36096f29b41aa3709f2cc9/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss#L46)

And probably rename the variable to a common name too?

yankurk:
There are other places where `#333` has been used. I'll replace all of them in another PR @GaneshSPatil

ketan:
> There are other places where #333 has been used. I'll replace all of them in another PR @GaneshSPatil

Please extract to a variable?

yankurk:
@ketan -- I've made those changes as a part of #4394.

Issue #4310, opened 2018-02-08
Add Server 'scheduled_at' Time to Dashboard API


GaneshSPatil:
##### Issue Type
- Feature enhancement

##### Summary
Currently, the Dashboard API does not include server time or the timezone related information.
Add server `server_scheduled_at` time information to show the server triggered time on the pipeline instance.

Current Dashboard API returns:
```
{
  "_embedded": {
    "pipelines": [
      {
        "_embedded": {
          "instances": [
            {
              "label": "2185",
              "triggered_by": "Triggered by changes",
              "scheduled_at": "2018-02-07T13:51:38Z"
            }
          ]
        }
      }
    ]
  }
}
```

Change the API response to include server time:
```
{
  "_embedded": {
    "pipelines": [
      {
        "_embedded": {
          "instances": [
            {
              "label": "2185",
              "triggered_by": "Triggered by changes",
              "scheduled_at": "1518011498341",
              "scheduled_at_server_time": "07 Feb, 2018 at 08:51:38 [-0500]"
            }
          ]
        }
      }
    ]
  }
}
```

Related Links:
* https://github.com/gocd/gocd/issues/4254#issuecomment-363733672
* [Pipeline instance Server Time On old Page](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/views/pipelines/_pipeline_instance_details.html.erb#L2)
* [Server Time formatting](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/lib/extensions/java_util_date.rb#L26-L28)

Related issue: #4254 #4121

rajiesh:
Verified on 18.2.0 (6211-48d5a54cd6525e54b4692cb757a6ad9896e46392)

PR #4392, opened 2018-02-21
Show server scheduled at time (#4310)

GaneshSPatil:
* Add 'scheduled_at_server_time' field in go dashboard API response
* Show server scheduled at time on hover of scheduled at local time
* Use time_formatter lib at client side for a common date format at one place


yankurk:
@GaneshSPatil -- I have tested it and the changes look good to me.

GaneshSPatil:
Do not include server timezone information as part of the Dashboard API response, the server timezone information is necessary on other pages too (Example: Stage details page).
Add timezone information as part of body param so that its available on all the SPAs.

Closing this PR. will send another PR with changes.

PR #4405, opened 2018-02-22
Remove h2-1.2.127.jar.

varshavaradarajan:
The h2-1.2.127.jar jar was used to migrate from the old default HSQL db to h2 db. Since h2 has been the default for years now, this is no longer needed.


ketan:
For existing installs, this will leave behind the jar on disk. You should remove it in this method so it'll get purged on an upgrade:

https://github.com/gocd/gocd/blob/e759cb91a89a0041ea088e32950f711f7f5cf16a/server/src/main/java/com/thoughtworks/go/server/util/GoLauncher.java#L54-L63

PR #4406, opened 2018-02-22
Show server scheduled at time (#4310)

GaneshSPatil:
* Add 'data-timezone' body attribute which contain server timezone UTC offset information
* Show server scheduled at time on hover of scheduled at local time
* Use time_formatter lib at client side for a common date format at one place

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/helpers/time_formatter.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/helpers/time_formatter.js
@@ -15,9 +15,29 @@
  */
 
 
+const $      = require('jquery');
 const moment = require("moment");
 require("moment-duration-format");
 
-module.exports = function (time) {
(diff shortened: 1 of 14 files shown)


server/webapp/WEB-INF/rails.new/webpack/helpers/time_formatter.js
```diff
@@ -15,9 +15,23 @@
...
+  format: (time) => {
+    return moment(time).format('DD MMM YYYY [at] HH:mm:ss [Local Time]');
+  },
+
+  serverTimeZone: () => {
+    return parseInt(JSON.parse($('body').attr('data-timezone')));
+  },
+
+  formatInServerTime: (time) => {
+    const format             = '[Server Time:] DD MMM, YYYY [at] HH:mm:ss Z';
```
ketan: Can we stick to the same format as local time:

```
DD MMM YYYY [at] HH:mm:ss [Server Time]
```

PR #4408, opened 2018-02-22
Optimize the SQL executed for the new dashboard

ketan:


ketan:
There is still one gotcha with the `where pipelines.names in ()` clause in the `getPipelinesForDashboard` query, in that a pipeline that exists in the config, but never ran will incur a significant cost. So as a second level optimisation, we may possibly not pass in a pipelinename that we know has never run.

jyotisingh:
@ketan - I was thinking we could get what we have so far merged once the build passes and get some feedback from our dogfood environments. The bit mentioned in https://github.com/gocd/gocd/pull/4408#issuecomment-367659564 could be handled in a separate PR. Does that sound ok?

ketan:
> The bit mentioned in #4408 (comment) could be handled in a separate PR. Does that sound ok?

Yes, that note was meant for a reminder in case we stumble on this in the future.

Issue #4398, opened 2018-02-21
Possible corner case with new dashboard


arvindsv:
##### Issue Type

- Bug Report

##### Summary

This is unverified, but I just thought about it and felt it is good to capture it.

1. Have a GoCD server with a few pipelines.
2. Run a pipeline (or do something) to decrease the usable size on disk (in either the DB volume or artifacts volume).

I expect that the old dashboard will recognize this and make all pipelines un-trigger-able. I don't think the new dashboard will. Admittedly, this is a corner case. Also, upon a trigger, it will be rejected. However, there will be a difference in the two dashboards.

We need to talk about it and decide what to do, if it really does happen.

In code, we need to [consider all of these](https://github.com/gocd/gocd/blob/b85614cc0c132563e24c27a7febfc88e0ab9bb1c/server/src/main/java/com/thoughtworks/go/server/service/SchedulingCheckerService.java#L179-L193) (AboutToBeTriggeredChecker, ArtifactsDiskSpaceFullChecker and DatabaseDiskSpaceFullChecker).

/cc @rajiesh

varshavaradarajan:
This is handled in the BuildCauseProducerService - https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L152

varshavaradarajan:
Ah, you mean disable the trigger buttons.

arvindsv:
Yes. As [this](https://github.com/gocd/gocd/blob/b85614cc0c132563e24c27a7febfc88e0ab9bb1c/server/src/main/java/com/thoughtworks/go/server/service/PipelineHistoryService.java#L552) would have done in the old dashboard.

arvindsv:
I'm ok if we say that we won't disable it. There are ways of disabling it without causing the crippling performance of running `DiskSpaceChecker` for *every* pipeline, for every dashboard request. We can talk about those if we want to.

varshavaradarajan:
Yes, I think @ketan and I had spoken about not running the disk checkers for every pipeline that gets triggered. If we can get a list of pipelines that are meant to be scheduled and run the disk space checker, it should be fine, I think? Because all we are checking is that the space available is not below the configured threshold. Not what happens when one of those triggered pipelines uploads lots of artifacts.

arvindsv:
[...]

PR #4411, opened 2018-02-22
Do not perform disk space check for every pipeline

ketan:
In order to disable the trigger buttons, a disk check was being
performed for every pipeline on the dashboard and that was impacting
performance of the dashboard. Both old and new dashboards were being
impacted.


arvindsv:
Paired with @ketan on this. Merging.

arvindsv:
Related to #4398. I think I'm fine closing #4398 as "wont fix". I don't think it's particularly important to make the pipelines not triggerable. But, in such a situation, triggering should show a message, which it does not today. That's because the scheduling API is async.

Issue #4403, opened 2018-02-22
Indication of Config repo Pipeline on New Dashboard


GaneshSPatil:
##### Issue Type

- Feature proposal

##### Summary
Currently, on pipelines page (old dashboard), there is no clear indication of a pipeline defined in config repository.

#### Pipelines defined in config repo:
- [ ] should show an icon indicating pipeline is defined in config repository.
- [ ] should show config repository name and revision in which it is defined/updated.
- [x] should disable gear icon stating a message `pipelines defined in config repository cannot be edited`.
- [ ] should link config-repository icon to config repository status page (#4402).

Related issues: #4081 #4351

PR #4417, opened 2018-02-23
Disable Pipeline settings icon for pipelines defined in config repo

GaneshSPatil:


api/api-dashboard-v2/src/main/java/com/thoughtworks/go/apiv2/dashboard/representers/PipelineRepresenter.java
```diff
@@ -33,25 +35,25 @@
...
-            .addChild("pause_info", getPauseInfoNEW(model))
-            .add("can_operate", model.isPipelineOperator(usernameString))
-            .add("can_administer", model.canBeAdministeredBy(usernameString))
-            .add("can_unlock", model.canBeOperatedBy(usernameString))
-            .add("can_pause", model.canBeOperatedBy(usernameString));
+                .addLinks(linksWriter -> addLinks(linksWriter, model))
+                .add("name", model.name().toString())
+                .add("last_updated_timestamp", model.getLastUpdatedTimeStamp())
+                .add("locked", model.model().getLatestPipelineInstance().isCurrentlyLocked())
+                .addChild("pause_info", getPauseInfoNEW(model))
```
ketan: Rename that method to `getPauseInfo`

api/api-dashboard-v2/src/main/java/com/thoughtworks/go/apiv2/dashboard/representers/PipelineRepresenter.java
```diff
@@ -33,25 +35,25 @@
...
+                .add("can_operate", model.isPipelineOperator(usernameString))
+                .add("can_administer", model.canBeAdministeredBy(usernameString))
+                .add("can_unlock", model.canBeOperatedBy(usernameString))
+                .add("can_pause", model.canBeOperatedBy(usernameString))
+                .addChild("origin", getOrigin(model.getOrigin()));
```
ketan: If all we're doing is to disable the edit button, can I suggest that we use the `can_administer` or just make this field a boolean to reduce the size and cost of serialization?

GaneshSPatil: @ketan -- can we handle only the permissions related stuff with `can_administer`?
Regarding sending the origin attribute, some more context is provided on issue #4403.

We will require the origin attribute to show more information about the config repo in which the pipeline is defined.

ketan: Since we don't know the specifics that are needed, can we leave it be for now?

Issue #438, opened 2014-08-10
Is there a way to make particular folders `allowed` for DirectoryCleaner? 


yaman:
I am trying to build a project written in golang. To build golang projects, you need to define `GOPATH` and checkout code you want to build into `$GOPATH/src/gitlab.fancycompany.com/yourgroup/yourproject` in order to make your golang imports work.

Thus I set `$GOPATH` to my agents build root, like `GOPATH=/var/go-agent/myapi` . And define checkout folder of git material relative to agents build root as `src/gitlab.fancycompany.com/yourgroup/yourproject` which checks out to `$GOPATH/src/gitlab.fancycompany.com/yourgroup/yourproject` . And `cleanup working directory` option is unchecked.

I don't want `go get` to get dependencies everytime I run it, instead, I want `go get` to leverage existing(already downloaded) dependencies under `$GOPATH/src/github...` and not to download again  and again to shorten build process, otherwise getting some dependencies takes 10 minutes...

But my agent removes files it seems irrelevant to build, like, it deletes `src/github.com` but it keeps `src/gitlab.fancycompany.com`. I checked the DirectoryCleaner code several times with tests but could not understand the logic behind. What is the reason it is allowing or not allowing is a mystery to me.

Is there a way to tell agent not to delete particular files under build root folder?

webmaster128:
Did you solve the problem or did you give up expecting a solution? I think the issue is very much valid. I also have different projects in my GO CD setup that randomly run and don't run working directory cleanups. In some cases the Cleaning step is entirely skipped and sometimes build files I want to keep are deleted.

abdulkadiryaman:
no couldnt solve it. changed the way we implemented the pipeline at the
end....

PR #4419, opened 2018-02-23
Upgrade h2 to 1.4.197.

varshavaradarajan:
Disable MV_STORE. @jyotisingh's PR - https://github.com/gocd/gocd/pull/3730 will take care of making use of mvstore.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
+++ b/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
@@ -111,8 +111,9 @@ public class SystemEnvironment implements Serializable, ConfigDirProvider {
     public static GoSystemProperty<Integer> RESOLVE_FANIN_MAX_BACK_TRACK_LIMIT = new CachedProperty<>(new GoIntSystemProperty("resolve.fanin.max.backtrack.limit", 100));
     public static GoSystemProperty<Integer> MATERIAL_UPDATE_INACTIVE_TIMEOUT = new CachedProperty<>(new GoIntSystemProperty("material.update.inactive.timeout", 15));
 
-    public static GoSystemProperty<Integer> H2_DB_TRACE_LEVEL = new GoIntSystemProperty("h2.trace.level", 1);
+    public static GoSystemProperty<Integer> H2_DB_TRACE_LEVEL = new GoIntSystemProperty("h2.trace.level", 3);
     public static GoSystemProperty<Integer> H2_DB_TRACE_FILE_SIZE_MB = new GoIntSystemProperty("h2.trace.file.size.mb", 16);
+    public static GoSystemProperty<Integer> H2_DB_LOB_TIMEOUT_MINUTES = new GoIntSystemProperty("h2.lob.timeout.minutes", 5);
     private static GoSystemProperty<String> CRUISE_DATABASE_DIR = new GoStringSystemProperty("cruise.database.dir", DB_DEFAULT_PATH);
 
     public static final String MATERIAL_UPDATE_IDLE_INTERVAL_PROPERTY = "material.update.idle.interval";
@@ -690,6 +691,10 @@ public class SystemEnvironment implements Serializable, ConfigDirProvider {
 
     public int getCruiseDbTraceFileSize() {
         return H2_DB_TRACE_FILE_SIZE_MB.getValue();
+    }
+
+    public int getLobTimeoutInMilliSeconds() {
+        return (H2_DB_LOB_TIMEOUT_MINUTES.getValue() * 60 * 1000);
     }
 
     public Level pluginLoggingLevel(String pluginId) {
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/LicenseReport.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/LicenseReport.groovy
@@ -36,7 +36,7 @@ class LicenseReport {
     'dom4j BSD license',
     'Public Domain',
     'Similar to Apache License but with the acknowledgment clause removed',
-    'The H2 License, Version 1.0',
+    'MPL 2.0 or EPL 1.0',
     'The OpenSymphony Software License 1.1',
   ]
 
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/SpdxLicense.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/SpdxLicense.groovy
@@ -129,7 +129,7 @@ enum SpdxLicense {
   DOTSEQN("Dotseqn", "Dotseqn License"),
   DSDP("DSDP", "DSDP License"),
   DVIPDFM("dvipdfm", "dvipdfm License"),
-  EPL_1_0("EPL-1.0", "Eclipse Public License 1.0", "Eclipse Public License - v 1.0", "Eclipse Public License - Version 1.0"),
+  EPL_1_0("EPL-1.0", "Eclipse Public License 1.0", "Eclipse Public License - v 1.0", "Eclipse Public License - Version 1.0", "EPL 1.0"),
   ECL_1_0("ECL-1.0", "Educational Community License v1.0"),
   ECL_2_0("ECL-2.0", "Educational Community License v2.0"),
   EGENIX("eGenix", "eGenix.com Public License 1.1.0"),
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -135,7 +135,7 @@ dependencies {
   copyOnlyTestData project(':test:test-agent')
 
   compile group: 'commons-dbcp', name: 'commons-dbcp', version: '1.4'
-  compile group: 'com.h2database', name: 'h2', version: '1.3.168'
+  compile group: 'com.h2database', name: 'h2', version: '1.4.197'
   compile group: 'net.sf', name: 'dbdeploy', version: '2.11.1'
   compile(group: 'org.hibernate', name: 'hibernate-ehcache', version: project.versions.hibernate) {
     exclude(module: 'ehcache-core')
@@ -668,7 +668,7 @@ task verifyWar(type: VerifyJarTask) {
         "go-plugin-infra-${project.version}.jar",
         "gson-${project.versions.gson}.jar",
         "guava-${project.versions.guava}.jar",
-        "h2-1.3.168.jar",
+        "h2-1.4.197.jar",
         "hibernate-commons-annotations-3.2.0.Final.jar",
         "hibernate-core-${project.versions.hibernate}.jar",
         "hibernate-ehcache-${project.versions.hibernate}.jar",
--- a/server/src/main/java/com/thoughtworks/go/server/database/H2Database.java
+++ b/server/src/main/java/com/thoughtworks/go/server/database/H2Database.java
@@ -135,13 +135,15 @@ public class H2Database implements Database {
     }
 
     private String dburl(Boolean mvccEnabled) {
-        return "jdbc:h2:" + systemEnvironment.getDbPath() + "/" + configuration.getName()
+        return "jdbc:h2:" + systemEnvironment.getDbPath().getAbsolutePath() + "/" + configuration.getName()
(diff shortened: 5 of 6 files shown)


jyotisingh:
1. A new version `1.4.197 (2018-03-18)` of h2db came out 2 days ago, may be we should upgrade to that. 

2. Could you bring up a local server with some existing data (may be from build.gocd) first with this patch(ie. h2 1.4.196/1.4.197), wait for the server to come up fully. Now, bring down the server and bring it up with h2 1.3.168 but with the upgraded database just to see if things are still compatible. Just in case we have to rollback this change if/when we find a potential issue with the upgraded version of h2, we need to see if the database would still be compatible or not. 

3. Went through the change-logs for the versions between 1.3.168 and 1.4.197, a few things stood out that we need to check if it works as expected or not. Went through their issues and our code to see if they apply to us or not. The ones I think do not apply have been striked-out, the others need to be checked/tested. 

	Version 1.4.177 Beta (2014-04-12)
	* Enable the new storage format for dates (system property "h2.storeLocalTime").
		    For the MVStore mode, this is always enabled, but with version 1.4
		    this is even enabled in the PageStore mode.
        -- We have a few timestamp columns, need to see if this affects us in anyway
	~~* Implicit relative paths are disabled (system property "h2.implicitRelativePath"),
		    so that the database URL jdbc:h2:test now needs to be written as jdbc:h2:./test.~~
	~~“select ... fetch first 1 row only" is supported with the regular mode.
		    This was disabled so far because "fetch" is now a keyword.
		    See also Mode.supportOffsetFetch.~~
	~~* Remove support for the limited old-style outer join syntax using "(+)".
		    Use "outer join" instead.
		    System property "h2.oldStyleOuterJoin".~~

	https://github.com/h2database/h2database/blob/version-1.4.178/h2/src/docsrc/html/changelog.html
	* ~~In server mode, appending ";autocommit=false" to the database URL was working,
		    but the return value of Connection.getAutoCommit() was wrong.~~

	https://github.com/h2database/h2database/blob/version-1.4.181/h2/src/docsrc/html/changelog.html
	* ~~Auto-scale MAX_MEMORY_ROWS and CACHE_SIZE settings by the amount of available RAM. Gives a better
		    out of box experience for people with more powerful machines.~~
	* Issue 567: H2 hangs for a long time then (sometimes) recovers.
		    Introduce a queue when doing table locking to prevent session starvation.
	~~* There was a way to prevent a database from being re-opened,
		    by creating a column constraint that references a table with a higher id,
		    for example with "check" constraints that contains queries.
		    This is now detected, and creating the table is prohibited.
		    In future versions of H2, most likely creating references to other
		    tables will no longer be supported because of such problems.~~
	* Large result sets now always create temporary tables instead of temporary files.
	* When using the PageStore, opening a database failed in some cases with a NullPointerException
		    if temporary tables were used (explicitly, or implicitly when using large result sets).
	* Databases created with version 1.3.175 and earlier
		    that contained foreign keys in combination with multi-column indexes
		    could not be opened in some cases.
		    This was due to a bugfix in version 1.3.176:
		    Referential integrity constraints sometimes used the wrong index. 
         -- Not a breaking change, but we need to see if the fixes made for these somehow affect/break anything for us since we do have a few multi-column indexes (idx_pipeline_name_id, idx_stages_name_pipelineId, idx_builds_agentId_stageid_name etc)

	https://github.com/h2database/h2database/blob/version-1.4.187/h2/src/docsrc/html/changelog.html
	* References to BLOB and CLOB objects now have a timeout.
		    The configuration setting is LOB_TIMEOUT (default 5 minutes).
		    This should avoid growing the database file if there are many queries that return BLOB or CLOB objects,
		    and the database is not closed for a longer time.
	-- We do have a few CLOBs in our tables jobagentmetadata, artifactplans.

	https://github.com/h2database/h2database/blob/version-1.4.188/h2/src/docsrc/html/changelog.html
	~~* A thread deadlock detector (disabled by default) can help
		    detect and analyze Java level deadlocks.
		    To enable, set the system property "h2.threadDeadlockDetector" to true.~~ - we could explore this later
	~~* Fix for issue #143, deadlock between two sessions hitting the same sequence on a column.~~ NA happens for multithreading mode only
	~~* Timestamps in the trace log follow the format (yyyy-MM-dd HH:mm:ss) instead
		of the old format (MM-dd HH:mm:ss). Patch by Richard Bull.~~

	https://github.com/h2database/h2database/blob/version-1.4.193/h2/src/docsrc/html/changelog.html
	~~* WARNING: THE MERGE BELOW WILL AFFECT ANY 'TIMESTAMP WITH TIMEZONE' INDEXES. You will need to drop and recreate any such indexes.
	PR #364: fix compare TIMESTAMP WITH TIMEZONE~~

	* PR #359: Added missing source encoding. Assuming UTF-8.

	https://github.com/h2database/h2database/blob/version-1.4.196/h2/src/docsrc/html/changelog.html
	~~Issue #438: Fix slow logging via SLF4J (TRACE_LEVEL_FILE=4).~~

varshavaradarajan:
> Enable the new storage format for dates (system property "h2.storeLocalTime").
For the MVStore mode, this is always enabled, but with version 1.4
this is even enabled in the PageStore mode.
-- We have a few timestamp columns, need to see if this affects us in anyway

Brought up the server with h2-1.3.168 and scheduled a build, brought up the same db with h2-1.4.197 and scheduled the build. The timestamp format was the same. 

> Issue 567: H2 hangs for a long time then (sometimes) recovers.
Introduce a queue when doing table locking to prevent session starvation.

I would say that the performance run (after benchmarking with h2-1.3.168, of course) will indicate if db calls are being slow.

> References to BLOB and CLOB objects now have a timeout.
The configuration setting is LOB_TIMEOUT (default 5 minutes).
This should avoid growing the database file if there are many queries that return BLOB or CLOB objects,
and the database is not closed for a longer time.
-- We do have a few CLOBs in our tables jobagentmetadata, artifactplans.

Related issue: https://github.com/h2database/h2database/issues/634
During a perf run, we should take a heap dump and analyze what the right LOB_TIMEOUT default for GoCD should be if there are a lot of references to lobs in the memory? 

> When using the PageStore, opening a database failed in some cases with a NullPointerException
if temporary tables were used (explicitly, or implicitly when using large result sets).

Um, did a `kill -9` while it was adding modifications in a transaction and tried restarting the server. Started fine. May be, I have to try a larger result set. I am not sure how else to test this yet. 

> Databases created with version 1.3.175 and earlier
that contained foreign keys in combination with multi-column indexes
could not be opened in some cases.
This was due to a bugfix in version 1.3.176:
Referential integrity constraints sometimes used the wrong index.
-- Not a breaking change, but we need to see if the fixes made for these somehow affect/break anything for us since we do have a few multi-column indexes (idx_pipeline_name_id, idx_stages_name_pipelineId, idx_builds_agentId_stageid_name etc)

I am not sure how to check this. 

> PR #359: Added missing source encoding. Assuming UTF-8.

Tried to change to locale and inserted strange characters via a commit message. Seem to not affect the db. Is that a sufficient check?

@jyotisingh - need your help with a couple of scenarios mentioned above.

varshavaradarajan:
> Databases created with version 1.3.175 and earlier
that contained foreign keys in combination with multi-column indexes
could not be opened in some cases.
This was due to a bugfix in version 1.3.176:
Referential integrity constraints sometimes used the wrong index.
-- Not a breaking change, but we need to see if the fixes made for these somehow affect/break anything for us since we do have a few multi-column indexes 

Ran `explain analyze` on the following queries on tables that use multi column indexes. Most seemed the same. 

1. findPipelineByNameAndCounter
2. findEarlierPipelineThatPassedForStage
3. mostRecentPipeline - used 22% of the index in h2-1.3 and 21% of the index with h2-1.4. 
4. getLastSuccessfulStageInPipeline
5. getMostRecentPassedStageId
6. findStageIdByPipelineAndStageNameAndCounter

7. getLastSuccessfulBuildIdOnAgent 

On h2-1.3

```
BUILDS.BUILDS_DATA read: 12 (42%)
BUILDS.IDX_BUILDS_AGENTID_STAGEID_NAME read: 6 (21%)
PIPELINES.IDX_PIPELINE_NAME_ID read: 5 (17%)
PIPELINES.PIPELINES_DATA read: 2 (7%)
STAGES.STAGES_DATA read: 3 (10%)
```

On h2-1.4

```
BUILDS.BUILDS_DATA read: 12 (25%)
BUILDS.IDX_BUILDS_AGENTID_STAGEID_NAME read: 6 (12%)
PIPELINES.IDX_PIPELINE_NAME_ID read: 5 (10%)
PIPELINES.PIPELINES_DATA read: 6 (12%)
STAGES.STAGES_DATA read: 18 (38%)
```

Um, I don't yet know the consequences of the above change.

jyotisingh:
> Could you bring up a local server with some existing data (may be from build.gocd) first with this patch(ie. h2 1.4.196/1.4.197), wait for the server to come up fully. Now, bring down the server and bring it up with h2 1.3.168 but with the upgraded database just to see if things are still compatible. Just in case we have to rollback this change if/when we find a potential issue with the upgraded version of h2, we need to see if the database would still be compatible or not.

You haven't responded about this one. Out of the list, this the most critical thing to test before we merge. Soon enough after the merge, all dogfood environments would get upgrade and we might not be in a position to rollback (in case of issues) if this doesn't work. 

> During a perf run, we should take a heap dump and analyze what the right LOB_TIMEOUT default for GoCD should be if there are a lot of references to lobs in the memory?

About this, well I don't even know how does this apply to us and what does it really mean. So, as far as testing is concerned, I don't think a blind perf run will find things for us or not. I mostly wanted you to read up a but about this and see what it really means. More so, what happens when a timeout really happens, will something fail, do we need any specific handling for these timeouts in code etc.

>> Large result sets now always create temporary tables instead of temporary files.
When using the PageStore, opening a database failed in some cases with a NullPointerException
if temporary tables were used (explicitly, or implicitly when using large result sets).

>Um, did a kill -9 while it was adding modifications in a transaction and tried restarting the server. Started fine. May be, I have to try a larger result set. I am not sure how else to test this yet.

It says "Large result sets", insert modifications do not really return large result sets. Look for some queries which return a large result set, may be dashboard query or a job detail kind of query. This also requires a bit of reading up about what do they mean by large, or when exactly would a temp table be used. Is there anything that we need to handle when such temp tables are used?

> Ran explain analyze on the following queries on tables that use multi column indexes. Most seemed the same.
> Um, I don't yet know the consequences of the above change.

As long as the indexes are being used correctly, we should be ok for merging the PR. We will get some feedback from the dogfooding and performance runs. 

We need to test a few things out before the release once this is merged. But the most important aspect for me to merge this PR is point 1. So, as long as you can confirm that works well, I can go ahead with merging this PR.

varshavaradarajan:
> You haven't responded about this one. Out of the list, this the most critical thing to test before we merge. Soon enough after the merge, all dogfood environments would get upgrade and we might not be in a position to rollback (in case of issues) if this doesn't work.

@jyotisingh - yes. Did this with build.go.cd's db. It brought up the server just fine both times. 

@jyotisingh - can we have the perf runs before merging, please?

ketan:
@jyotisingh @varshavaradarajan — wanted to see where we are with this PR. Is a perf run the only thing pending?

jyotisingh:
Yes, @varshavaradarajan wanted to get a perf run done before getting this merged in. Once the release related testing on the perf environment is done, we should be able to get a run for this

varshavaradarajan:
Perf run summary - 

The max time taken for db queries has reduced from 709ms to 437ms averaging all db queries. Comparing with the previous perf run with h2, we did not see any db queries taking longer than usual.
The CPU usage and response time seem to have increased a bit. We need to have another perf run to verify though. 
Memory utilization compared with the previous run has decreased a bit as well. From 8.91G to 8.16G. This is not indicative of anything but probably the snapshot being taken right after GC.  

@rajiesh and I are planning to have a shorted perf run targeting only the db intensive operations and pages, first with h2-1.3 and then with h2-1.4

varshavaradarajan:
Had another performance run with the thread profiler. The CPU usage is comparable to what was being reported for h2-1.3. So, this is good to merge. I'll rebase and merge this in a while.

cc @ketan , @jyotisingh

PR #4421, opened 2018-02-23
Make PluginEndpoint request/response and define analytics.jobs contract

kierarad:
**PluginEndpoint now implements and request-response approach only.**

This is a breaking API change.

Now all defined listeners (including onInit) receive a transport object instead of a reply() fn.

The transport object has 2 methods: request() and respond(); Plugin authors may use the transport
object to make requests and add response callbacks:

PluginEndpoint.onInit(initialData, transport) {

  // some code here
  // ...

  transport.request("analytics.job.history", myParams).
    done(function success() { /* do something */}).
    fail(function error() { /* handle errors */).
    always(function onComplete() { /* cleanup */);
});

And on GoCD server pages, we register listeners to handle the response back:

PluginEndpoint.define({

  "analytics.job.history": function(message, transport) {
    // do some processing

    // currently, response() requires a request ID passed through the message
    // so that the plugin-side scripts can associate responses with requests
    transport.respond(message.head.reqId, responseData);
  }

});

As one can see, the API resembles jQuery.ajax() requests.

**The analytics jobs endpoint allows us to get the data for a job history chart.**

### DIFF
--- a/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/analytics/AnalyticsExtension.java
+++ b/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/analytics/AnalyticsExtension.java
@@ -32,6 +32,7 @@ import org.springframework.stereotype.Component;
 
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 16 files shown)


marques-work:
@maheshp @jyotisingh Let me know when you want to sync up on this; I'd be happy to go over the changes with you over Zoom, Hangouts, or Slack

PR #4433, opened 2018-02-26
Analytics Dashboard iframes no longer scroll

ibnc:


arvindsv:
@ibnc I'm merging this, but this is just a part of the problem. This works like `overflow: hidden`. However, we need to make sure that the whole chart is shown here. That might need some understanding between the plugin and core about the size of that iframe, or a proper responsive chart. Some of this is on the plugin side.

PR #4442, opened 2018-02-27
Show message on dashboard while cache is being prepped

ketan:
The server will render a status 202 (Accepted) to indicate that the
request will be processed in some time.

The JS SPA will understand this status code and render an appropriate
message on the dashboard.


ketan:
Screenshot after this PR:

screen shot 2018-02-27 at 9 08 23 pm

Issue #4441, opened 2018-02-27
[New dashboard] Trigger with options UI issues


arvindsv:
##### Issue Type

- Bug Report

##### Summary

The trigger with options box still has a few issues.

1. Sometimes it goes beyond the screen.
2. When you select a commit which is not the latest one, the popup stays and incorrectly says: “Last 1 commits listed in chronological order”.

###### Basic environment details

* GoCD Version: `18.2.0 (6175-1058e75b18e8a645dd71702851994a010789f450)`
* Java Version: `1.8.0_144`

###### Information

1. [x] Screenshot 1: It is cut off. Doesn't show 5 commits.

2. [x] Screenshot 2: It shows the wrong message. I chose the 3rd commit shown. There are two problems here:
      1. It says: "Last 1 commits listed in chronological order" (that is incorrect).
      2. It keeps that box open.

3. [x] Screenshot 3: When the previous commit was a long one, it breaks in interesting ways. You can also see parts of line 2 in the first commit's message in the dropdown.

4. [x] ~~Screenshot 4: When there is only one commit, the dropdown doesn't show at all.~~ [arvindsv: Not an issue.]

Screenshot 1:
<img width="1425" alt="2018_02_27_08-07-22" src="https://user-images.githubusercontent.com/172235/36730408-4fce57d6-1b95-11e8-8d7e-2a33746dbd80.png">

Screenshot 2:
<img width="1426" alt="2018_02_27_08-07-38" src="https://user-images.githubusercontent.com/172235/36730410-51fd94f4-1b95-11e8-9511-5e9bb49bed44.png">

Screenshot 3:
<img width="1201" alt="2018_02_27_08-28-22" src="https://user-images.githubusercontent.com/172235/36731364-80df7d20-1b98-11e8-9675-6d07d10621e2.png">

Screenshot 4:
<img width="1201" alt="2018_02_27_08-38-11" src="https://user-images.githubusercontent.com/172235/36731735-97f25cf2-1b99-11e8-896c-aa2b522c9fd3.png">

arvindsv:
I need to retry Screenshots 3 and 4 with the latest code. Will do now.

arvindsv:
Maybe we can move the "Revision to trigger with" up, so that it obscures the box, but has enough space for the last 5 commits.

varshavaradarajan:
- [ ] Revision to trigger with misaligned on the old dashboard.

[...]

PR #4443, opened 2018-02-27
Trigger with options modal fixes (#4441)

ketan:
* The commit message popup now scrolls properly and is cropped near the
   window bottom
* The comment in the modal and popup are cropped to 5 lines and 3 lines
  respectively
* Tweak the margins to make more space and align things a bit better


ketan:
This fixes 1 and 3 in bug #4441

PR #4447, opened 2018-02-28
Change http status when not enough disk space is available

ketan:
Earlier used to return 400 (bad request), but now returns a
507 (insufficient storage). This is done because a server going out of
disk space is not a user error (4xx), so a 5xx code seems more
appropriate.


ketan:
Moving this to 18.3, is that ok @varshavaradarajan @arvindsv?

arvindsv:
Yes.

PR #4456, opened 2018-02-28
#4417 - Treat pipeline with null config origin as local.

arvindsv:
Scenario to trigger this:

1. Quick edit a pipeline.
2. Save.
3. Go to new dashboard.


arvindsv:
@GaneshSPatil I'm merging this, but you need to take a look later.

PR #4460, opened 2018-03-01
Explicitly add config origin on the submitted pipeline config for the Update Pipeline Config API request

GaneshSPatil:
The User submitted pipeline config JSON is deserialized using [PipelineConfigRepresenter](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/presenters/api_v5/admin/pipelines/pipeline_config_representer.rb).

As the origin attribute is not required while updating/creating the pipeline config, the deserialized pipeline config object does not have a config origin attribute on it.

Explicitly add the [FileConfigOrigin](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/remote/FileConfigOrigin.java) origin on the user submitted pipeline config object.

It is safer to expect that all the deserialized pipeline config objects are defined in config.xml as the check for remotely defined pipeline is performed prior to deserialization [here](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/app/controllers/api_v5/admin/pipelines_controller.rb#L22)


varshavaradarajan:
Closing in favour of https://github.com/gocd/gocd/pull/4465

Issue #2780, opened 2016-10-12
Improvements to EntityConfigUpdateCommand


varshavaradarajan:
##### Issue Type

- Feature enhancement
##### Summary

There can be some improvements made to the EntityCommands to provide more clarity for validation.
1. Have a `preValidate` and `postValidate` call instead of just isValid(). The reason for this is that, there are some validations, without which the entity should not even be updated in the cruise config and based on which, the subsequent validations (that is, the post-update validations should happen.) Most common of the pre validate can be that the identifying attribute should not be null.
2. Pull a `isAuthorized` method out of the canContinue method. 
3. Move the isAuthorized method to goConfigService so that given the username and privileges we want, we can avoid duplication in all entity commands.
4. The `canContinue` method can deal with stale requests etc.

akshaydewan:
I'm planning to pick this up and raise PRs in small chunks.

PR #4465, opened 2018-03-01
Set config origin while performing post validation updates on pipeline config and environment config

varshavaradarajan:
@GaneshSPatil - I'll send another PR to address https://github.com/gocd/gocd/issues/2780 because I don't think we should do everything listed there.


varshavaradarajan:
Closing this stale PR.

PR #4470, opened 2018-03-01
Elastic agent extension v3

bdpiparva:
This adds new capability to support agent status report for elastic agent plugin. As part of this version following calls are modified.

- Get capabilities
    * Added new capability(`supports_agent_status_report`) to support agent status report.
- Get agent status report (plugin will get this call only if `supports_agent_status_report` is set to true in capabilities)
    * From job details page user can request for agent status report. The plugin will get job identifier or elastic-agent-id based on agent registration status.
- Create agent
   * Plugin will now get job identifier as an additional param in the request body. This will help plugin to tag container with job identifier.
- Should assign work
   * Plugin will get job identifier for which this call is being made.

### Breaking changes

In elastic-agent extension v1 and v2 request has `go.cd.elastic-agent` as prefix which is now changed to `cd.go.elastic-agent`  in v3


server/src/test-fast/java/com/thoughtworks/go/server/service/ElasticAgentPluginServiceTest.java
```diff
@@ -242,6 +259,57 @@ public void shouldNotAssignJobToAnAgentBroughtUpByADifferentElasticPlugin(){
...
+        final Capabilities capabilities = new Capabilities(false, true);
+        final GoPluginDescriptor descriptor = new GoPluginDescriptor("cd.go.example.plugin", null, null, null, null, false);
+        elasticAgentMetadataStore.setPluginInfo(new ElasticAgentPluginInfo(descriptor, null, null, null, capabilities));
+
+        when(registry.getAgentStatusReport("cd.go.example.plugin", null, "some-id"))
+                .thenReturn("<div>This is a agent status report snippet.</div>");
+
+        final String pluginStatusReport = service.getAgentStatusReport("cd.go.example.plugin", null, "some-id");
+
+        assertThat(pluginStatusReport, is("<div>This is a agent status report snippet.</div>"));
```
varshavaradarajan: Rename the variable to agentStatusReport

server/src/test-fast/java/com/thoughtworks/go/server/service/ElasticAgentPluginServiceTest.java
```diff
@@ -242,6 +259,57 @@ public void shouldNotAssignJobToAnAgentBroughtUpByADifferentElasticPlugin(){
...
+    }
+
+    @Test
+    public void shouldErrorOutWhenPluginDoesNotAgentSupportStatusReport() {
+        final Capabilities capabilities = new Capabilities(true, false);
+        final GoPluginDescriptor descriptor = new GoPluginDescriptor("cd.go.example.plugin", null, null, null, null, false);
+        elasticAgentMetadataStore.setPluginInfo(new ElasticAgentPluginInfo(descriptor, null, null, null, capabilities));
+
+        thrown.expect(RuntimeException.class);
```
varshavaradarajan: Isn't this `UnsupportedOperationException` now?

server/src/test-integration/java/com/thoughtworks/go/server/controller/JobControllerIntegrationTest.java
```diff
@@ -190,14 +190,14 @@ public void jobDetailModel_shouldHaveTheElasticProfilePluginIdWhenAgentIsNotAssi
         JobInstance job = stage.getFirstJob();
         GoPluginDescriptor.About about = new GoPluginDescriptor.About("name", "0.1", "17.3.0", "desc", null, null);
         GoPluginDescriptor descriptor = new GoPluginDescriptor("plugin_id", null, about, null, null, false);
-        ElasticAgentMetadataStore.instance().setPluginInfo(new ElasticAgentPluginInfo(descriptor, null, null, null, new Capabilities(true)));
+        ElasticAgentMetadataStore.instance().setPluginInfo(new ElasticAgentPluginInfo(descriptor, null, null, null, new Capabilities(false,true)));
```
varshavaradarajan: Indent code.

server/webapp/WEB-INF/rails.new/spec/controllers/admin/status_reports_controller_spec.rb
```diff
@@ -67,11 +67,11 @@
...
 
-    it 'should return the status report for an available plugin' do
-      elastic_agent_extension = instance_double('com.thoughtworks.go.plugin.access.elastic.ElasticAgentExtension')
-      allow(elastic_agent_extension).to receive(:getStatusReport).with(elastic_plugin_id).and_return('status_report')
+    it 'should return the plugin status report for an available plugin' do
+      elastic_agent_plugin_service = instance_double('com.thoughtworks.go.server.service.ElasticAgentPluginService')
+      allow(elastic_agent_plugin_service).to receive(:getPluginStatusReport).with(elastic_plugin_id).and_return('status_report')
 
-      allow(controller).to receive(:elastic_agent_extension).and_return(elastic_agent_extension)
+      allow(controller).to receive(:elastic_agent_plugin_service).and_return(elastic_agent_plugin_service)
```
varshavaradarajan: Move the stub to the before each block to avoid duplication.

server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/elasticagent/ElasticAgentRequestProcessor.java
```diff
@@ -16,135 +16,56 @@
...
-                public AgentMetadata apply(ElasticAgentMetadata obj) {
-                    return toAgentMetadata(obj);
-                }
-            }).collect(Collectors.toList());
-        }
-
-        String responseBody = elasticAgentExtension.getElasticAgentMessageConverter(goPluginApiRequest.apiVersion()).listAgentsResponseBody(metadata);
-        return DefaultGoApiResponse.success(responseBody);
+    public ElasticAgentRequestProcessor(PluginRequestProcessorRegistry registry, AgentService agentService, AgentConfigService agentConfigService) {
+        this(registry, new ElasticAgentRequestProcessorV1(agentService, agentConfigService),
```
jyotisingh: I think this is incorrect. While the elastic agent extension version has changed over time, the processor contract hasn't changed at all. Why create three versions for this then?

server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/elasticagent/v2/ElasticAgentProcessorConverterV2.java
```diff
@@ -0,0 +1,59 @@
...
+import com.thoughtworks.go.plugin.access.elastic.v2.AgentMetadataDTO;
+import com.thoughtworks.go.server.service.plugins.processor.elasticagent.ElasticAgentProcessorConverter;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class ElasticAgentProcessorConverterV2 implements ElasticAgentProcessorConverter {
+    private static final Gson GSON = new GsonBuilder().excludeFieldsWithoutExposeAnnotation().create();
+    private static final AgentMetadataConverterV2 agentMetadataConverter = new AgentMetadataConverterV2();
```
jyotisingh: `AgentMetadataDTO` and `AgentMetadataConverterV2` are defined by the EA extension V2. Reusing it here would mean going back to tying down elastic-agent extension with processor's contract. We should use a different one here, its alright if it looks exactly the same. Also, like I mentioned earlier. Since the processor's contract hasn't changed, we don't really need three versions of the processor, converters or this DTO.

server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/elasticagent/ElasticAgentRequestProcessor.java
bdpiparva: Our skeleton plugin uses same API constant to represent processor version and the extension version. That means the plugin that we have in gocd-contrib is already using "2.0" API version of the processor.

plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/elastic/v2/ElasticAgentExtensionConverterV2.java
```diff
@@ -29,78 +30,35 @@
...
 public class ElasticAgentExtensionConverterV2 implements ElasticAgentMessageConverter {
-    private static final Gson GSON = new Gson();
-
-    public static final String VERSION = "2.0";
+    private static final Gson GSON = new GsonBuilder().excludeFieldsWithoutExposeAnnotation().create();
+    private CapabilitiesConverterV2 capabilitiesConverterV2 = new CapabilitiesConverterV2();
+    private AgentMetadataConverterV2 agentMetadataConverterV2 = new AgentMetadataConverterV2();
 
     @Override
     public String createAgentRequestBody(String autoRegisterKey, String environment, Map<String, String> configuration, JobIdentifier jobIdentifier) {
```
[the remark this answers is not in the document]
bdpiparva: None of the plugins(at least plugins that we maintain) uses job_identifier as part of v2. Considering that extension v2 is not publicly announced previously I think so we don't need v4.

PR #4475, opened 2018-03-02
Enable the quick dashboard when the toggle is turned on

### LINKED ISSUE #4450: [New Dashboard] Enabling of `quicker_dashboard_key` toggle do not trigger dashboard cache creation
##### Issue Type

- Bug Report

##### Summary

Enabling of `quicker_dashboard_key` toggle do not trigger dashboard cache creation. Empty dashboard page is shown to users. Need to perform a full config save to get it triggered and dashboard page to load with pipelines

* Go Version: `18.2.0 (6175-1058e75b18e8a645dd71702851994a010789f450)`

##### Steps to Reproduce

1. On an existing GoCD setup with multiple pipeline and instances of it enable the toggle `quciker_dashboard_key`
2. Visit `<go-server>/go/dashbaord` page
3. Empty dashboard is shown up

##### Expected Results

Dashboard page should showup all the pipelines

##### Actual Results

Only on a full config save all pipelines are shown. Or if there are any pipeline runs, then only that specific pipeline is shown

ketan:
Hook in a listener in the `FeatureToggleService` which notifies
when the toggle is turned on, and pretends like a full config save happened

Fixes #4450

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 ThoughtWorks, Inc.
+ * Copyright 2018 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -65,7 +65,7 @@ public class GoDashboardService {
     }
 
     public void updateCacheForPipeline(CaseInsensitiveString pipelineName) {
-        if (!Toggles.isToggleOn(Toggles.QUICKER_DASHBOARD_KEY)) {
+        if (featureToggleDisabled()) {
             return;
         }
         PipelineConfigs group = goConfigService.findGroupByPipeline(pipelineName);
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
```diff
@@ -65,7 +65,7 @@ public void visit(PipelineConfigs group) {
     }
 
     public void updateCacheForPipeline(CaseInsensitiveString pipelineName) {
-        if (!Toggles.isToggleOn(Toggles.QUICKER_DASHBOARD_KEY)) {
+        if (featureToggleDisabled()) {
```
varshavaradarajan: rename this to`isFeatureToggleDisabled` ?

server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
```diff
@@ -75,14 +75,18 @@ public void updateCacheForPipeline(CaseInsensitiveString pipelineName) {
...
 
     public void updateCacheForPipeline(PipelineConfig pipelineConfig) {
-        if (!Toggles.isToggleOn(Toggles.QUICKER_DASHBOARD_KEY)) {
+        if (featureToggleDisabled()) {
             return;
         }
         updateCache(goConfigService.findGroupByPipeline(pipelineConfig.name()), pipelineConfig);
     }
 
+    public boolean featureToggleDisabled() {
```
varshavaradarajan: Does this belong here?

ketan: It's only temporary. I think.

server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
ketan: Done. And merged.

PR #4479, opened 2018-03-05
Show new dashboard link on old dashboard when toggle(quicker_dashboard_key) is on

AksMKR:
- Added link from old dashboard to new.


server/webapp/WEB-INF/rails.new/app/views/pipelines/index.html.erb
```diff
@@ -1,28 +1,32 @@
 <% @view_title = "Pipelines" -%>
-<% @page_header = '<h1 class="entity_title">Pipelines</h1>'
+<% if is_quicker_dashboard_toggle_enabled?
+     @page_header = '<h1 class="entity_title">Pipelines</h1><a href="/go/dashboard" class="toggle-new-view">New Dashboard</a>'
```
varshavaradarajan: The `<h1>` pipelines part need not repeat itself.

PR #4485, opened 2018-03-05
Truncate the label template on the dashboard (Fixes #4483)

### LINKED ISSUE #4483: Pipeline counter is not truncated
##### Issue Type

- Bug Report

##### Summary

When the counter is too long, the counter is not truncated using ellipsis.

screen shot 2018-03-05 at 9 07 41 pm

ketan:


GaneshSPatil:
@ketan -- as we are truncating the pipeline label, it would be good to show the full pipeline label on hover.


**note**: We've used title tag to display [pipeline name](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_header_widget.js.msx#L55). Similarly we can do it for pipeline label

PR #4491, opened 2018-03-06
increasing SPA timeout from 5s to 60s considering the load etc on an active server

jyotisingh:


bdpiparva:
Need to change the timeout value in the documentation: https://github.com/gocd/docs.go.cd/pull/152

PR #4501, opened 2018-03-08
Ensure cancelled stage has lastTransitionedTime

maheshp:
* The lastTransitionedTime on a stage is set using a database trigger
  upon insert of buildStateTransitions. As part of normal stage
  transition in ScheduleService#updateJobStatus the stage object in
  cache is invalidated and reloaded which will have the
  lastTransitionedTime. But in case a stage is cancelled through
  ScheduleService#cancelAndTriggerRelevantStages a stage being cancelled
  is not synced with db to get the lastTransitionTime and hence
  StageStatus notifications have a null lastTransitionedTime.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java
@@ -51,6 +51,7 @@ import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.support.TransactionCallback;
 import org.springframework.transaction.support.TransactionSynchronizationAdapter;
 
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


jyotisingh:
That description is not fully accurate. Its not just a cache issue. When a new stage record gets inserted in db along with the associated jobs and transitions, for the insertion of the `scheduled` transition, the db trigger ([Migration_230007](https://github.com/gocd/gocd/blob/6995ea7176e8c759ebd9a8cc3e790a69ec48cb1f/server/src/main/java/com/thoughtworks/go/server/sqlmigration/Migration_230007.java)) gets invoked which is supposed to update the `lastTransitionedTime` on the stage record. The trigger is invoked, however it fails to update the stage with `lastTransitionedTime`  for some reason that I am unable to understand. Hence the value remains to be `null` in db too. What I mean is if you check the update count of the [sql executed](https://github.com/gocd/gocd/blob/6995ea7176e8c759ebd9a8cc3e790a69ec48cb1f/server/src/main/java/com/thoughtworks/go/server/sqlmigration/Migration_230007.java#L48) by this trigger, it would be `0` for a new stage being inserted, and `1` for the further updates that happen to the transitions.

Even if the trigger worked, the stage object that is passed along doesn't get updated, so I am going ahead with merging this PR.

Issue #4448, opened 2018-02-28
[New Dashboard] Pipeline with lock behavior specified continues to be locked after run completion


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Pipeline with lock behavior specified continues to be locked after run completion. A full cache regeneration only removes the invalid data ie., do a full config save and then visit dashboard page the lock is removed

###### Basic environment details

* Go Version: `18.2.0 (6175-1058e75b18e8a645dd71702851994a010789f450)`

##### Steps to Reproduce

1. Create a pipeline with lock behavior set(either `lockOnFailure` or `unlockWhenFinished`)
2. Trigger the pipeline and wait for it to complete
3. Visit the new dashboard page - Pipeline is still locked and trigger options disabled
4. Do a full config save
5. Now on new dashboard page pipeline lock is released

##### Expected Results

Pipeline lock should be released as per the lock behavior set

##### Actual Results

Pipeline continues to be locked even after completion

rajiesh:
Verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4507, opened 2018-03-09
Notify listeners about lock status change only after the transaction is complete (#4448)

akshaydewan:
Since the DB cache is invalidated in the `afterCommit` hook, it was possible for the event listeners to get stale information. With this change the lock status event listeners will get notified only after the transaction has completed.


akshaydewan:
Don't merge this yet, since there seems to be a functional test failure. I'll look into it.

jyotisingh:
I think I have asked this before, but I seem to be forgetting. Why is doing this in `afterCommit` not good enough since we care about committed transactions alone?

jyotisingh:
If we must to do this as part of `afterCompletion` hook instead of `afterCommit`, then we need to fix notification in case of `stageStatusChanged` event too, but like I said I do not know/remember why this should be done in `afterCompletion`.

akshaydewan:
We are invalidating the cache in the `afterCommit` hook, and we need to notify listeners after that is done. If we have multiple `afterCommits`, the order is unpredictable. So one way around it is to notify listeners in the `afterCompletion` hook.

jyotisingh:
Spoke to @akshaydewan about the two comments:
https://github.com/gocd/gocd/pull/4507#issuecomment-376093430 - we need the notifications to be sent after all the afterCommit blocks have been executed. Since there are multiple afterCommit blocks registered across different classes for a given nested transaction, the only way to ensure all of them are executed before notifying listeners is to do it in the afterCompletion block which comes a little later in the lifecycle after all registered afterCommits have been executed. 

https://github.com/gocd/gocd/pull/4507#issuecomment-376094759 - While we are not aware of any issues with `stageStatusChanged`, going by the theory about unpredictable order of afterCommit, we need to address this too, but changing things around that would require a lot more effort (going through scenarios etc) and can be taken care of in a separate PR.

Merging this PR.

PR #4508, opened 2018-03-09
WIP : Pipeline search filtering

AksMKR:
Pipeline search filtering based on:
1. Pipeline Group Name
2. Pipeline Name
3. Pipeline status

### DIFF
--- a/server/webapp/WEB-INF/rails.new/spec/webpack/models/dashboard/dashboard_spec.js
+++ b/server/webapp/WEB-INF/rails.new/spec/webpack/models/dashboard/dashboard_spec.js
@@ -66,13 +66,14 @@ describe("Dashboard", () => {
     });
 
     it("it should filter dashboard provided filter text", () => {
-      expect(dashboard.getPipelineGroups()[0].pipelines).toEqual(['up42']);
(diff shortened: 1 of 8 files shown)


ketan:
@arvindsv maybe I'm wrong here. But this looks like duct-tape for the problem about not having named personalizations. Once we have named personalizations, my assumption is that the number of pipelines on the dashboard will come down to a manageable level where search *may* not be needed.

arvindsv:
@ketan I'll need to see a screenshot. If this is what I think it is, then you're partially right. It is a bit of duct tape. However, if you have hundreds of pipelines, then even to create named personalization, you might need something like this. Right? Once you do create it, you shouldn't need to use this much - that's true.

ketan:
> @ketan <https://github.com/ketan> I'll need to see a screenshot. If this
> is what I think it is, then you're partially right. It is a bit of duct
> tape. However, if you have hundreds of pipelines, then even to create named
> personalization, you might need something like this. Right? Once you do
> create it, you shouldn't need to use this much - that's true.
>

For the named personalization, depending on how the UI is, search may be
needed for the interface that lists pipelines, not for the "dashboard".

Issue #4515, opened 2018-03-12
Pipeline pause API throws a NullPointerException


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

If the pipeline pause API documented in https://api.gocd.org/18.2.0/#pause-a-pipeline is used with the `X-GoCD-Confirm` and no data like below, the response returns a NullPointerException: 

```
curl 'http://localhost:8153/go/api/pipelines/up42/pause' \
      -H 'Accept: application/vnd.go.cd.v1+json' \
      -H 'Content-Type: application/json' \
      -H 'X-GoCD-Confirm: true' \
      -X POST
```
Response: 

```
{"error":"Payload data is not valid JSON: java.lang.NullPointerException"}
```

ketan:
A workaround currently would be to submit an empty JSON "{}" with a content-type of `application/json`.

rajiesh:
Verified on 18.3.0 (6578-31b0ad88fec5af12e88b913612eb6c141abaee0e)

PR #4516, opened 2018-03-12
Pipeline pause API - handle the case when no data is posted to the API. (#4515).

varshavaradarajan:


api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1Delegate.java
```diff
@@ -100,9 +100,12 @@ public void setupRoutes() {
 
     public String pause(Request req, Response res) throws IOException {
         HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
-        JsonReader requestBody = GsonTransformer.getInstance().jsonReaderFrom(req.body());
+        String pauseCause = null;
         String pipelineName = req.params("pipeline_name");
-        String pauseCause = requestBody.optString("pause_cause").orElse(null);
+        if (!StringUtils.isBlank(req.body())) {
```
ketan: I'd probably handle the null in the `GsonTransformer` instead of here — avoids similar problem in other cases.

Issue #4454, opened 2018-02-28
Cannot remove empty pipeline group from personalize dropdown


bdpiparva:
##### Issue Type
- Bug Report

##### Summary
1. Create an empty pipeline group(foo). This won't show up on the dashboard.

- In new dashboard 
  -  In a personalization dropdown, you will see the `foo`. Click on select none and then click on apply. After refreshing the page you will see `foo` selected.
  - You can not uncheck the pipeline group when there is no pipeline.

![new-dashbord](https://user-images.githubusercontent.com/7871209/36788091-a6db4732-1cb2-11e8-843a-1b889078435b.gif)

- In old dashboard
  -  In a personalization dropdown, you will see the `foo`. Click on select none and then click on apply. After refreshing the page you will see `foo` selected.
  - You can also try unchecking` foo` manually and then apply it. you will see the same behavior.

![old-dashboard](https://user-images.githubusercontent.com/7871209/36787959-20a64f72-1cb2-11e8-9426-9af0d140f729.gif)

AksMKR:
I went through the issue. There are 2 problems present here
1. In personalize dropdown, if we select None and Apply then all the pipeline groups will get unselected. But the empty pipeline groups will still stay selected. This issue can be fixed with few tweaks.
2. We are showing empty pipeline groups in personalize menu but not on dashboard. This might bring confusion to users. Either we can show the empty pipeline groups on both personalize menu and dashboard, or hide them on both.

@naveenbhaskar @arvindsv What's your call on this? How should we go about it?

varshavaradarajan:
@yankurk and I had spoken about this issue before he went on leave. I think he was working on it or has a WIP branch.

varshavaradarajan:
https://github.com/yankurk/gocd/tree/personalize-dropdown is the WIP branch for this

rajiesh:
verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4517, opened 2018-03-12
Skipping the empty pipeline groups from pipeline selection response

AksMKR:
1. Changes made to **PipelineSelectionsRepresenter**

https://github.com/gocd/gocd/issues/4454

### DIFF
--- a/api/api-pipeline-selection-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineselection/representers/PipelineSelectionsRepresenter.java
+++ b/api/api-pipeline-selection-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineselection/representers/PipelineSelectionsRepresenter.java
@@ -33,10 +33,14 @@ public class PipelineSelectionsRepresenter {
             .add("blacklist", pipelineSelectionResponse.getSelectedPipelines().isBlacklist())
             .addChild("pipelines", pipelineGroupsWriter -> {
                 pipelineSelectionResponse.getPipelineConfigs().forEach(pipelineConfigs -> {
-                        List<String> pipelineNames = pipelineConfigs.getPipelines().stream().map(PipelineConfig::getName).map(CaseInsensitiveString::toString).collect(Collectors.toList());
+                    List<String> pipelineNames = pipelineConfigs
+                            .getPipelines().stream()
+                            .map(pipelineConfig -> pipelineConfig.getName().toString())
+                            .collect(Collectors.toList());
+                    if (!pipelineNames.isEmpty()) {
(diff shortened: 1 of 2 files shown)


api/api-pipeline-selection-v1/src/test/groovy/com/thoughtworks/go/apiv1/pipelineselection/representers/PipelineSelectionsRepresenterTest.groovy
```diff
@@ -55,7 +55,6 @@ class PipelineSelectionsRepresenterTest {
 
       JsonFluentAssert.assertThatJson(actualJson).isEqualTo([
         pipelines: [
-          grp1: [],
```
ketan: It's not obvious looking at the test that empty groups should not be rendered. I'd probably write an explicit test which makes the behavior explicit.

AksMKR: Added separate testcase for empty pipeline groups.

api/api-pipeline-selection-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineselection/representers/PipelineSelectionsRepresenter.java
```diff
@@ -33,10 +33,15 @@ public static void toJSON(OutputWriter writer, PipelineSelectionResponse pipelin
             .add("blacklist", pipelineSelectionResponse.getSelectedPipelines().isBlacklist())
             .addChild("pipelines", pipelineGroupsWriter -> {
                 pipelineSelectionResponse.getPipelineConfigs().forEach(pipelineConfigs -> {
-                        List<String> pipelineNames = pipelineConfigs.getPipelines().stream().map(PipelineConfig::getName).map(CaseInsensitiveString::toString).collect(Collectors.toList());
+                    List<String> pipelineNames = pipelineConfigs
+                            .getPipelines().stream()
+                            .map(PipelineConfig::getName)
+                            .map(CaseInsensitiveString::toString)
```
GaneshSPatil: can we iter once over the list instead of twice?

`map(pipeline -> pipeline.getName().toString())` ?

ketan: It's a stream, so it's not actually iterating twice :)

PR #4523, opened 2018-03-14
[New Dashboard] Analytics Icon on the new Dashboard

ibnc:
This adds the analytics icon to the new dashboard

Screenshot 1:
screen shot 2018-03-14 at 1 13 38 pm

Screenshot 2:
screen shot 2018-03-14 at 1 14 17 pm


ibnc:
What would be the best way to merge in core to resolve the conflicts?

Specifically, how should I change how I'm grabbing the plugins supporting pipeline analytics (see: https://github.com/gocd/gocd/pull/4523/files#diff-09f4b6f802453d49ff058e064dd98519R28)

@maheshp @jyotisingh @arvindsv @ketan

GaneshSPatil:
@ibnc -- IMO, for an SPA, we shouldn't rely on controller injecting div data params and fetch the required information through API, which makes an SPA to rely only on API data.

we should fetch the `plugin-info` for analytics type and if any of analytics plugin supports pipeline level analytics, we should display the analytics icon.
We've done something similar on [agents SPA](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/webpack/single_page_apps/agents.js#L92).

Also, based on the admin permissions for the pipeline, the analytics icon should be shown.

can pair on the same if you want and get it merged quickly?

ibnc:
@GaneshSPatil Okay, I'm now grabbing the needed data through PluginInfos, and the icon should now only display if the current user as an admin.

would you mind taking a look?

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/NewDashboardDelegate.java
```diff
@@ -57,6 +60,7 @@ public ModelAndView index(Request request, Response response) {
             put("viewTitle", "Dashboard");
             put("isQuickEditPageEnabled", Toggles.isToggleOn(Toggles.PIPELINE_CONFIG_SINGLE_PAGE_APP) && Toggles.isToggleOn(Toggles.QUICK_EDIT_PAGE_DEFAULT));
             put("isNewDashboardPageEnabled", Toggles.isToggleOn(Toggles.NEW_DASHBOARD_PAGE_DEFAULT));
+            put("isUserAnAdmin", securityService.isUserAdmin(currentUsername()));
```
GaneshSPatil: can we show pipeline level analytics for pipeline group admins as well?

arvindsv: The check whether user is admin (for pipeline analytics in GoCD's dashboard) is just a decision made for the purposes of not showing our commercial plugin till it is released. We don't need to call it out as such. We can name it: "shouldShow...".

ibnc:
@GaneshSPatil I made the change that @arvindsv suggested, merged in core, and the PR builds are passing. Can this be merged?

ibnc:
@maheshp @GaneshSPatil where are we at with this PR being mergeable?

maheshp:
@ibnc sorry bout this, either @GaneshSPatil or I will look into it today.

Issue #4489, opened 2018-03-06
[New Dashboard] If invalid revision specified do not trigger pipeline


rajiesh:
##### Issue Type

- Bug Report

##### Summary

If specified an invalid revision from trigger with options modal, pipeline should not be scheduled and also should show up some valid message to indicate the same

###### Basic environment details

* Go Version: `18.2.0 (6211-48d5a54cd6525e54b4692cb757a6ad9896e46392)`

rajiesh:
Verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

Issue #4488, opened 2018-03-06
[New Dashboard] If for a material no revision selected on trigger with options, schedules with latest revision


rajiesh:
##### Issue Type

- Bug Report

##### Summary

On New dashboard Trigger with options modal if user do not specify specific revision for any material the latest revision is picked for scheduling. However the existing behavior was run with last run revision. Need to retain it on the new dashboard page too for consistency.

If at all we decide to change the behavior to run with latest revision, it has to be consistent too across both dashboard page as long as we support old dashboard

rajiesh:
Verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d). Current behavior is to run with last run revision of the materials for which no revision is selected

PR #4525, opened 2018-03-15
Trigger with revisions issues (resolved)

AksMKR:
1. If for a material no revision selected on trigger with options, will schedule with last run revision. (#4488)
2. If invalid revision specified do not trigger pipeline and show error message to user. (#4489 )
3. If the pipeline never ran before, and no revision is selected, then trigger the pipeline with latest revision (since there is no last run revision).


yankurk:
I've tested the functionality out locally. Looks good to me.

PR #4528, opened 2018-03-15
Make plugin-initiated requests from plugin-endpoint.js send a generic message key based on API version

marques-work:
The idea here is that we can send a generic request/message key from the plugin to the parent page. This is a good fit to the generic routing done by @maheshp and @jyotisingh. Params are passed through to the model, and eventually the actual AJAX request to the static route.

I also made the key versioned: `api.analytics.v1`

I was thinking that this would be convenient if we wanted to support multiple API versions as one would just add new handlers for each version.

### DIFF
--- a/scripts/clean-workingdir.sh
+++ b/scripts/clean-workingdir.sh
@@ -6,6 +6,6 @@ WORKING_DIR=$(cd `dirname $0`/.. && pwd)
 
 cd $WORKING_DIR
 
-rm -rf agent/config agent/pipelines server/pipelines server/db/config.git
+rm -rf server/test-repo agent/config agent/pipelines server/pipelines server/db/config.git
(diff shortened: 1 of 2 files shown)


scripts/clean-workingdir.sh
```diff
@@ -6,6 +6,6 @@ WORKING_DIR=$(cd `dirname $0`/.. && pwd)
 
 cd $WORKING_DIR
 
-rm -rf agent/config agent/pipelines server/pipelines server/db/config.git
+rm -rf server/test-repo agent/config agent/pipelines server/pipelines server/db/config.git
```
marques-work: Don't mind this -- it's just a script I use for cleaning my dev environment without killing my IDEA settings. Didn't think it was worth a separate PR, so I'm smuggling this one in here :)

Issue #4524, opened 2018-03-15
Feature toggle key `new_dashboard_page_default` is `false` by default


rajiesh:
##### Issue Type

- Bug Report

##### Summary

The feature toggle `new_dashboard_page_default` should be `true` by default to make new dashboard the default landing page. But it is false now.

Also if user visit `/go/old_dashboard` page there is an option to navigate to new dashboard, however it redirects to `/go/dashboard` page which does not exists. It should redirect to `/go/pipelines` instead

###### Basic environment details

* Go Version: `18.3.0 (6275-6750bf21edde35d4d590ada2532fa426b02a7db7)`

rajiesh:
This commit makes the toggle key  `new_dashboard_page_default` to be true by default -  https://github.com/gocd/gocd/commit/bcf0e7ae82b677ba63b7069f4b720ebe599ca149

However the startup experience so far for users has been to show the pipeline creation wizard on first access of go server home page with no pipelines configured. That is changed now, user land up on the empty new pipeline dashboard page. 

However clicking on `old dashboard` link takes to the pipeline creation wizard. Good to maintain same behavior with new dashboard page too

arvindsv:
I agree. I guess https://github.com/gocd/gocd/pull/4646 takes care of this.

rajiesh:
Verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4534, opened 2018-03-19
Fix urlrewrite for the dashboard routes. (#4524)

varshavaradarajan:
* /go/dashboard points to the new dashboard
* /go/old_dashboard points to the old dashboard
* /go/pipelines points to either the new or the old dashboard based on the toggle 'new_dashboard_page_default'
* urlrewrite handles the routes to dashboard with a trailing slash
* Add the old dashboard button on the new dashboard.


bdpiparva:
### When `quicker_dashboard_key` == `new_dashboard_page_default` == true
- The default page is `New dashboard` with link to the `old dashboard`
- Also `old dashboard` has a link to `new dashboard`

### When `quicker_dashboard_key == true`  and `new_dashboard_page_default == false`
- The default page is `Old dashboard` with link to the `New dashboard`

### When `quicker_dashboard_key == false`
- The default page is `Old dashboard` and it does not have a link to the `New dashboard`

PR #4537, opened 2018-03-19
Show no analytics plugin installed message when no analytics plugin is installed and user lands on /go/analytics page

GaneshSPatil:
screen shot 2018-03-19 at 6 05 58 pm


jyotisingh:
What if a plugin is installed but not configured or doesn't support dashboard metrics?

maheshp:
I think there are couple of scenarios we need to consider while rendering the top level analytics,
1. Plugin installed, supports only dashboard metrics.
2. Plugin installed, supports only pipeline metrics.
3. Plugin installed, but not configured.
4. No plugins installed, user access the `/go/analytics ' page directly.

PR #4538, opened 2018-03-19
Avoid making DB call for admin/latest-agent.status call

bdpiparva:
As part of `admin/latest-agent.status (HEAD)` call we end up looking in DB for `_go_agent_HOSTNAME` user which is not there. All the agents are making this call in 10 seconds interval. This PR will not check in DB if a username starts with `_go_agent_`.

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/server/domain/Username.java
+++ b/domain/src/main/java/com/thoughtworks/go/server/domain/Username.java
@@ -59,7 +59,7 @@ public class Username implements Serializable {
         return username;
     }
 
-    public boolean hasDistinctDisplayName(){
+    public boolean hasDistinctDisplayName() {
         return !displayName.equals(username.toString());
     }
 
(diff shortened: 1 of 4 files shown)


ketan:
Would you ideally not put this filter on the endpoint?

ketan:
> Would you ideally not put this filter on the endpoint?

And on other endpoints in the controller that download the jar and zip files.

bdpiparva:
> And on other endpoints in the controller that download the jar and zip files.  Would you ideally not put this filter on the endpoint? 

Currently, we don't have filter chain defined in acegi XML for  admin/latest-agent.status (HEAD). So by default it is using filter chain defined for `/**`

ketan:
> Currently, we don't have filter chain defined in acegi XML for admin/latest-agent.status (HEAD). So by default it is using filter chain defined for /**

I'm suggesting that we add one :)

bdpiparva:
@ketan  --- Agreed, at this point in time apart from `admin/latest-agent.status (HEAD)` other URL also requires different filter-chain. I have created an [Issue#4541](https://github.com/gocd/gocd/issues/4541) for this. We should do the same for all other routes and that requires some more effort.

bdpiparva:
As of now, I am merging this PR for fix the issue. We have to address the issue for other urls.

Issue #4474, opened 2018-03-02
SPAs popup a login dialog after session timeout


ketan:
##### Issue Type

- Bug Report

##### Summary

The SPAs will popup the basic auth dialog because of the API endpoints send a `WWW-Authenticate: Basic realm="GoCD"` header to force username password base authentication.

![screen shot 2018-03-02 at 11 08 12 am](https://user-images.githubusercontent.com/10598/36884995-4788bb76-1e0a-11e8-8182-2edd5d74052f.png)

A fix appears to be to add some sort of sniffer for `X-Requested-With: XMLHttpRequest` in the incoming request and not send the `WWW-Authenticate` header here https://github.com/gocd/gocd/blob/58886bb2b85294bb94e805651bb90ca75fae330f/server/src/main/java/com/thoughtworks/go/server/security/BasicProcessingFilterEntryPoint.java#L39-L55

jyotisingh:
@ketan - why is this problem isolated to SPAs?

ketan:
I'm not sure why — I've not investigated it yet for lack of bandwidth.

jyotisingh:
I am not against the solution you are suggesting for this. However, I don't see it happening for other API calls. Which is why I think we should spend some time trying to find out if this happens only for SPAs or in other scenarios too and also understand the reason for the same. We could leave this open until one of has some bandwidth to investigate this a bit if that's ok with you.

varshavaradarajan:
@jyotisingh - This is not an issue on the old page because while `WWW-Authenticate` header is added to all requests, for requests like `go/pipelines.json`, `go/server_health_messages.json`, trigger pipeline on old dashboard page etc; on 401, we redirect to login page `go/auth/login` from the client side.

PR #4543, opened 2018-03-20
Do not add the WWW-Authenticate header for XHRs.

varshavaradarajan:


varshavaradarajan:
https://github.com/gocd/gocd/issues/4474

PR #4546, opened 2018-03-21
[WIP] Change the sleep command used in tests run on windows

akshaydewan:


arvindsv:
@akshaydewan Can you revert the changes in [this commit](https://github.com/gocd/gocd/commit/d98a5c637c60b5d7604c120e0232db937f25c22d) then? If you think it should work now. I'd mentioned on gitter that I had ignored them.

akshaydewan:
@arvindsv The windows tests are [still failing](https://github.com/gocd/gocd/pull/4549) after this change. We'll need to investigate it further.

PR #4550, opened 2018-03-22
[Analytics Endpoint] Define 'go.cd.analytics.v1.link-to' transport request handler for the plugin

GaneshSPatil:
* Analytics charts can make use of 'go.cd.analytics.v1.link-to' endpoint to send a request to the parent window for any GoCD core link.
* Currently supported link:
    - job_details_page

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/analytics/helpers/gocd_link_support.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/analytics/helpers/gocd_link_support.js
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const PIPELINE_NAME_KEY    = 'pipeline_name';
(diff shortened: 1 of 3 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/analytics/helpers/gocd_link_support.js
```diff
@@ -0,0 +1,35 @@
...
+//--- FetchFromParams
+const pipelineName    = (params) => params[PIPELINE_NAME_KEY];
+const pipelineCounter = (params) => params[PIPELINE_COUNTER_KEY];
+const stageName       = (params) => params[STAGE_NAME_KEY];
+const stageCounter    = (params) => params[STAGE_COUNTER_KEY];
+const jobName         = (params) => params[JOB_NAME_KEY];
+
+module.exports = {
+  "job_details_page": (params) => {
+    const jobDetailsPagePath = `/go/tab/build/detail/${pipelineName(params)}/${pipelineCounter(params)}/${stageName(params)}/${stageCounter(params)}/${jobName(params)}`;
```
marques-work: @GaneshSPatil Can we assume the `/go/` will always be the context path? Isn't that configurable per instance?

server/webapp/WEB-INF/rails.new/webpack/views/analytics/helpers/gocd_link_support.js
```diff
@@ -0,0 +1,35 @@
...
+    window.open(jobDetailsPagePath, '_black');
```
marques-work: @GaneshSPatil do you mean `_blank`?

server/webapp/WEB-INF/rails.new/webpack/views/analytics/helpers/gocd_link_support.js
GaneshSPatil: @marques-work -- there has been some inconsistency with the way we are using webappContextPath.
There are a lot of places in the codebase where we have hardcoded the `/go` in the URLs.

I have created a separate [GitHub issue](https://github.com/gocd/gocd/issues/4558) to address this.
[here](https://gitter.im/gocd/tw-private-chat?at=5ab896e67685a046389d0102) is some discussion around the same.

PR #4551, opened 2018-03-22
Add a loading spinner to analytics frames in the analytics dashboard

marques-work:
Spinner also reveals title of the metric

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_analytics-common.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_analytics-common.scss
@@ -1,0 +1,47 @@
+.frame-container {
+  position: relative;
+
+  iframe {
+    height: 100%;
+    width:  100%;
+    border: 0;
+  }
+
+  .loading-overlay {
+    display:    none;
+    width:      auto;
+    height:     auto;
+    text-align: center;
+    background: #fff;
+    position:   absolute;
+    top:        0;
+    right:      0;
+    bottom:     0;
+    left:       0;
+    z-index:    200;
+
+    .loading-message {
+      position:    relative;
+      font-size:   14px;
+      font-weight: 300;
+      top:         calc(50% + 60px);
+    }
+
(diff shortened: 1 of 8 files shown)


marques-work:
@arvindsv @jyotisingh @maheshp @GaneshSPatil Would any of you mind reviewing this PR? The broken build is not related, and is broken in master as of a4bf2d1de7d15a3912a9e2f9f2276fde57be0235

GaneshSPatil:
@marques-work -- The graph title which is displayed along with the spinner looks little odd. In terms of styles and positioning.

* We are showing the spinner along with the title and on fetching data, we replace the entire content with the graph (which has the data).
* The problem with showing the title (which was provided as part of capability) and later replacing it with the chart, is:
  - The styles (font style, size, etc) of the title could be different.
     (The title rendered by GoCD can be completely different in look from one rendered graph)
  - The title itself might change.
     (Currently, we render the title which was provided as part of the capabilities. And expect the plugin to use the same title. There is no mechanism to verify the same. Also, there might be a genuine need to render a different chart title than one provided as part of capabilities. Eg: Job Run History Chart which contains job identifier in the title)

##### Just a thought:
As the server already knows about the graph being rendered. The server can show the graph container with the title and expect the plugin to provide the contents of the graph (**_without the graph title_**).

This can help in:
*  Relying on the title which is provided as part of the capabilities, provides consistency in terms of title styles and also the chart UX across all.
* For the dynamic information (such as _job identifier_, _agent hostname/UUID_), the chart should use a subtitle.

maheshp:
> As the server already knows about the graph being rendered. The server can show the graph container with the title and expect the plugin to provide the contents of the graph (without the graph title).

@GaneshSPatil I think we should leave it to the plugin to show the entire content with the heading. Yes the styles change between what is shown earlier and once we have the actual content. Does it make sense to not show it as a title but use the title to indicate which graph is being loaded.

@marques-work @tinozza thoughts?

fullsizerender

marques-work:
@tinozza how should this be styled? Any thoughts? This is probably desirable for the first release.

marques-work:
Updated to Zeplin design (option 2):

image

PR #4556, opened 2018-03-26
Remove the unmaintained japanese localization and some unused localization keys

ketan:


ketan:
Used this script to remove the unwanted keys:

```
$ cat server/webapp/localize.xml | egrep -o 'key=([^>])*>' | awk -F '"' '{print $2}' | xargs -I X bash -c "if [ \"\$(git grep X | wc -l)\" = '1' ]; then sed -i '/X/d' server/webapp/localize.xml; echo removed X; fi"
```

server/webapp/WEB-INF/rails.new/app/controllers/environments_controller.rb
```diff
@@ -37,20 +37,20 @@ def new
...
-    redirect_with_flash(l.string("ADD_ENVIRONMENT_SUCCESS", [@environment.name()]), :action => :show, :name => @environment.name().to_s, :class => 'success') if @result.isSuccessful()
+    redirect_with_flash("Added environment '#{@environment.name}'", :action => :show, :name => @environment.name().to_s, :class => 'success') if @result.isSuccessful()
   end
 
   def update
     @environment = environment_config_service.getEnvironmentForEdit(params[:name])
     @environment.setConfigAttributes(params[:environment])
     if @environment.name().blank?
-      render_error_response l.string("ENVIRONMENT_NAME_REQUIRED"), 400, true
+      render_error_response 'Environment name is required', 400, true
```
varshavaradarajan: There are multiple places where a localized string could be used (Line 40 here). If we decide to change the message, we'd have to change it in multiple places. I am not sure how many such localize strings are there that are used in multiple places. I am assuming that you have done this for only certain localize strings and not for all.

ketan: Yes, so far only for certain strings. I think what you're suggesting is to de-dup some of the strings, which I find useful. But given that we're moving over to SPAs, it's automatically taken care of. I don't particularly care for some of the old controllers and pages which have not changed in ages.

ketan: Also — This is still a WIP. I'm basically in the middle of "inlining" all the constants in localize.xml (or as many as I can without making too much of a mess). I might need to extract some helper methods in a class if there's significant de-duplication that's needed and if it's worth it. So far, I've identified maybe a handful of messages that have a lot of duplication (in production code).

Issue #4457, opened 2018-02-28
[New Dashboard] Possible perf improvement


arvindsv:
##### Issue Type

- Feature enhancement

##### Summary

Currently the full dashboard gets refreshed twice at startup. First time [here](https://github.com/gocd/gocd/blob/27df50b57d0eaeb151f4df5d77d5c39a02aa6bcb/server/src/main/java/com/thoughtworks/go/config/CachedGoConfig.java#L221) and the second time [here](https://github.com/gocd/gocd/blob/27df50b57d0eaeb151f4df5d77d5c39a02aa6bcb/server/src/main/java/com/thoughtworks/go/config/CachedGoConfig.java#L84), both of which end up calling [GoDashboardConfigChangeHandler#call(CruiseConfig config)](https://github.com/gocd/gocd/blob/27df50b57d0eaeb151f4df5d77d5c39a02aa6bcb/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardConfigChangeHandler.java#L40).

Marked this as 18.3.

PR #4560, opened 2018-03-26
Preventing multiple dashboard cache creation calls during startup - #4457

jyotisingh:
Dashboard cache was being recreated thrice during server startup - once while registering config-change-listener for GoDashboardActivityListener, then for featureToggleService and finally for CachedGoConfig.onTimer. Making the cache recreation happen only if the config has really changed, using md5 checksum to check the same.

~~PS: Do not merge, found an issue with this.~~
Update: The issue was partials (things configured in config repo) are not taken into account for the MD5 checksum that we store at the configCache level. This meant notifications wouldn't be sent if things were changed in config-repo. The checksum now has 2 components - md5sum of main-config, and the checksum of the partials. Only of the checksum changes that we send a notification.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/GoConfigHolder.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/GoConfigHolder.java
@@ -16,18 +16,73 @@
 
 package com.thoughtworks.go.config;
 
+import com.thoughtworks.go.config.remote.PartialConfig;
+import com.thoughtworks.go.util.CachedDigestUtils;
+
+import java.util.List;
+import java.util.Objects;
+
 public class GoConfigHolder {
     public final CruiseConfig config;
-    public final CruiseConfig configForEdit;
-    public CruiseConfig mergedConfigForEdit;
+    private final CruiseConfig configForEdit;
+    private CruiseConfig mergedConfigForEdit;
+    private Checksum checksum;
 
     public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit) {
-        this.config = config;
-        this.configForEdit = configForEdit;
+        this(config, configForEdit, null, null);
     }
 
-    public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit, CruiseConfig mergedConfigForEdit) {
-        this(config, configForEdit);
+    public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit, CruiseConfig mergedConfigForEdit, List<PartialConfig> partials) {
+        this.config = config;
+        this.configForEdit = configForEdit;
+        setMergedConfigForEdit(mergedConfigForEdit, partials);
+    }
+
+    private String md5(List<PartialConfig> partials) {
+        if (partials == null || partials.isEmpty()) {
+            return null;
+        }
+        return CachedDigestUtils.md5Hex(partials);
+    }
+
+    public Checksum getChecksum() {
+        return checksum;
+    }
+
+    public CruiseConfig getMergedConfigForEdit() {
+        return mergedConfigForEdit;
+    }
+
+    public void setMergedConfigForEdit(CruiseConfig mergedConfigForEdit, List<PartialConfig> partials) {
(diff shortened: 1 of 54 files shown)


config/config-api/src/main/java/com/thoughtworks/go/config/GoConfigHolder.java
```diff
@@ -16,18 +16,73 @@
...
+        setMergedConfigForEdit(mergedConfigForEdit, partials);
     }
 
-    public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit, CruiseConfig mergedConfigForEdit) {
-        this(config, configForEdit);
+    private String md5(List<PartialConfig> partials) {
+        if (partials == null || partials.isEmpty()) {
+            return null;
+        }
+        return CachedDigestUtils.md5Hex(partials);
```
ketan: Is there a reason we don't attempt serialization of `PartialConfig` using MagicalXmlWriter?

config/config-api/config/cipher
```diff
@@ -0,0 +1 @@
+57abfb79942adf10
```
ketan: Remove and ignore this?

config/config-api/src/main/java/com/thoughtworks/go/util/CachedDigestUtils.java
```diff
@@ -49,6 +51,21 @@ public static String sha256Hex(String string) {
         return digest;
     }
 
+    public static String md5Hex(Object serializable) {
+        return objectPools.computeDigest(DigestObjectPools.MD_5, new DigestObjectPools.DigestOperation() {
+            public String perform(MessageDigest messageDigest) {
+                try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
```
ketan: This can possibly create a huge byte array in memory, also since the implementation is such that the array needs "resizing" once it nears capacity.

A better alternative is to use `DigestOutputStream` wrapped around a `NullOutputStream` like here:
https://github.com/gocd/gocd/blob/5c7c6e382613f4efd79fbe5ce771284296c48500/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardPipelineGroup.java#L51-L74

This essentially has a memory footprint that is determined by whatever is needed by the digest algorithm.

config/config-api/src/main/java/com/thoughtworks/go/config/GoConfigHolder.java
```diff
@@ -16,18 +16,73 @@
...
+
+    public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit, CruiseConfig mergedConfigForEdit, List<PartialConfig> partials) {
         this.config = config;
         this.configForEdit = configForEdit;
+        setMergedConfigForEdit(mergedConfigForEdit, partials);
     }
 
-    public GoConfigHolder(CruiseConfig config, CruiseConfig configForEdit, CruiseConfig mergedConfigForEdit) {
-        this(config, configForEdit);
+    private String md5(List<PartialConfig> partials) {
```
ketan: Given that the partials are ordered based on whatever the plugin returns (which is again a function of order of files as returned by the filesystem) — should this be a list, or a ordered set, to keep consistent hashing when just the order changes, with no other changes.

GaneshSPatil:
What's the status on this PR?

@ketan @arvindsv @maheshp  -- This PR was required to avoid multiple dashboard cache creation calls on server startup. Possibly improving the startup performance. As of now, we haven't heard from anyone complaining about the increase in startup time with the new dashboard feature. Also, not sure if the PR have any other side effects.

Should we spend some more time on this PR and get it merged? or close it?

maheshp:
@GaneshSPatil I haven't verified if the cache is created multiple times but that seems to be the case as mentioned in the issue #4457 . I think this has not been reported so far as this happens only at server startup. 

Since there has been effort towards fixing this, I think somebody should take this forward and see if it feasible to be merged. I will mark it for 18.11.0 for now.

bdpiparva:
Closing the PR in favor of issue #5603.

Issue #4563, opened 2018-03-27
[New Dashboard] Filter icon shown twice in filter view section


yankurk:
##### Issue Type
- Bug Report

##### Summary

There are two filter icons shown in the filter view section:
<img width="456" alt="screen shot 2018-03-27 at 9 57 15 am" src="https://user-images.githubusercontent.com/23699743/37946853-4f0af96e-31a5-11e8-8334-b20ddacc6f4a.png">

PR #4566, opened 2018-03-27
[New dashboard] Filter related fixes

yankurk:
* Removes extra filter button (#4563)
* Fixes position of personalize dropdown spinner box


yankurk:
### Before:
screen shot 2018-03-29 at 11 27 35 am

### After fix:
screen shot 2018-03-29 at 11 24 58 am

jyotisingh:
The left-hand-side of the container doesn't seem aligned with the LHS of search textbox, but merging it since it fixes the other issues with the screen.

Issue #4567, opened 2018-03-27
[New Dashboard] Change font of commit message to be the same as commit author and SHA


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

To maintain uniformity on the new dashboard, the font-family of the commit message should be the same as the commit author and the material revision.

rajiesh:
Verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4568, opened 2018-03-27
Remove the monospace font from commit message to make it the same as commit author and revision.

varshavaradarajan:


varshavaradarajan:
@yankurk and @AksMKR  - can you please check if this change doesn't affect anything else?

ketan:
> @yankurk and @AksMKR - can you please check if this change doesn't affect anything else?

It should not. It's only used on the new pages.

varshavaradarajan:
https://github.com/gocd/gocd/issues/4567

PR #4573, opened 2018-03-27
Fix broken agents page due to change in the plugin infos (multiple extension support)

GaneshSPatil:
* Plugins can now support multiple extensions, hence look only for elastic agent capabilities for a given plugin

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
@@ -49,7 +49,7 @@ const supportsAgentStatusReport = function (pluginInfos, pluginId) {
 
   const pluginInfo = pluginInfos().findById(pluginId);
   if (pluginInfo) {
-    return pluginInfo.capabilities().supportsAgentStatusReport();
+    return pluginInfo.extensions()['elastic-agent'].capabilities().supportsAgentStatusReport();
   }
 
   return false;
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/agents/agent_row_widget.js.msx
```diff
@@ -49,7 +49,7 @@ const supportsAgentStatusReport = function (pluginInfos, pluginId) {
 
   const pluginInfo = pluginInfos().findById(pluginId);
   if (pluginInfo) {
-    return pluginInfo.capabilities().supportsAgentStatusReport();
+    return pluginInfo.extensions()['elastic-agent'].capabilities().supportsAgentStatusReport();
```
ketan: No null check?

GaneshSPatil: @ketan -- I guess the null check is not required.  There has been some discussion around the same on [tw-private](https://gitter.im/gocd/tw-private-chat?at=5aba178726a769820b2b2210) channel.

varshavaradarajan: What @GaneshSPatil and I discussed on the call - the elastic agent type plugin info will always have a capabilities object. A capabilities object with default values for supports_agent_report and support_status_report will be created while deserializing the plugin infos API response. Even for plugins using EA V1. As long as line 51 checks for pluginInfo being null, we should be sorted out.

Issue #4564, opened 2018-03-27
Misaligned Trigger with options spinner


GaneshSPatil:
##### Issue Type

- Bug Report

##### Description
The page spinner for the trigger with options information shown before the trigger_info is rendered, is misaligned.

##### Steps to reproduce:
1. Click on trigger with options for a pipeline
2. the page spinner is misaligned.

![screen shot 2018-03-27 at 10 51 52 am](https://user-images.githubusercontent.com/15275847/37948249-e9c594c6-31ac-11e8-9de3-075ef1591ce4.png)

rajiesh:
verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4582, opened 2018-03-28
fixed the height for trigger options modal when spinner is being shown

AksMKR:
#4564


server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
```diff
@@ -1047,6 +1047,10 @@ body {
   }
 }
 
+.empty-modal {
```
ketan: Why not set this on the trigger with options class itself?

PR #4583, opened 2018-03-29
Various fixes for analytics modal on pipelines page

marques-work:


varshavaradarajan:
Reverting this as it caused all the modals to break. The alignment of the buttons are off.

varshavaradarajan:
Need to check the modals in different places - plugins page, profiles, etc

Issue #4561, opened 2018-03-27
ellipsize the pipeline name on the dashboard


ketan:
##### Issue Type

- Bug Report

##### Summary

For a long pipeline name, it gets overlapped with the analytics icon:

<img width="288" alt="screen shot 2018-03-27 at 8 55 04 am" src="https://user-images.githubusercontent.com/10598/37945197-a7d9775e-319c-11e8-9e04-d18d97fb2422.png">

GaneshSPatil:
@ketan @naveenbhaskar -- I guess we need to rethink about the positioning of the plugin icons.
There could be minimum 2 icons (always) if one has analytics plugin installed.
The number of the icons could increase, based on:
* Pipeline is locked
* More analytics plugins are installed.

In the above-mentioned scenarios, there will always be a slight possibility of pipeline-name overlapping with the icons.

**Note**: As of now, we just have analytics plugin icon displayed on the pipeline, going forward we might have more plugin icons, example: icons indicating: deployment-pipeline, config-repo-pipeline, etc

rajiesh:
verified on 18.3.0 (6530-f9dc89e397d1b134f4d124e6381f8764e38e0a1d)

PR #4595, opened 2018-04-02
Sets pipeline name container width dynamically based on number of icons

yankurk:
#4561


yankurk:
This PR works for the current scenario where there are maximum of three icons possible: settings, pipeline lock and analytics icon. The width of the pipeline name section is dynamically adjusted based on the number of icons. A different approach is needed if in the future there is a need for more icons.

yankurk:
Some screenshots:
For scenario with two icons:
screen shot 2018-04-02 at 1 29 38 pm

For scenario with three icons:
screen shot 2018-04-02 at 1 29 52 pm

bdpiparva:
Can we wrap the `pipeline_name` and `pipeline_actions` in a parent and either use proper floats to do this — or — use table layouts? Ask @naveenbhaskar for help here, if you need to.

screen shot 2018-04-03 at 10 45 38 am

yankurk:
@bdpiparva What you have suggested is definitely better than what the current solution is. I talked to @naveenbhaskar about it and we were able to adjust the pipeline name width dynamically using css only. There are issues with IE (I suspect its flex). So, yes it does seem possible to do it using css and html structure tweaks. I'll send a different PR once the issues with IE are fixed. Keeping this PR open until then.

yankurk:
No workaround for IE could be identified. Hence, we will need to go with the current approach. If we need more icons, we will need to rethink the solution.

bdpiparva:
This issue is fixed using `display:flex`. All the CSS properties used in the PR is supported by IE11 and we have tested it.

- Windows 10 

screen shot 2018-04-09 at 2 19 18 pm

- Windows 7(Old version of IE11)

screen shot 2018-04-09 at 2 18 12 pm

PR #4600, opened 2018-04-03
Removed trailing slash from old_dashboard link

maheshp:
* Having the trailing slash on the old_dashboard led to wrong root path
  while loading the static analytics page. This would give a 404.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
@@ -91,7 +91,7 @@ const DashboardWidget = {
                                             vm={vnode.state.pipelineSelectionVM}/>);
     }
 
-    const oldDashboardLink = (<a class="toggle-old-view" href={"/go/old_dashboard/"}>Old Dashboard</a>);
+    const oldDashboardLink = (<a class="toggle-old-view" href={"/go/old_dashboard"}>Old Dashboard</a>);
     return (
       <div class="pipeline_wrapper">
         <div class="page_header" key="page-header">


varshavaradarajan:
@maheshp - perhaps adding the regex `(/?)` in urlrewrite for the old_dashboard is better? I mean, people often specify urls with trailing slash.

varshavaradarajan:
@maheshp - the urlrewrite takes care of the trailing slash and just checked by clicking on the `Old Dashboard` button. Works. So, not sure why we need this PR.

maheshp:
Sorry, I should have been more elaborate in the PR description. The issue is while fetching the pipeline analytics through the dashboard, where the path to analytics html is a relative path.

With - `http://localhost:8153/go/old_dashboard/` if a request is made to fetch the pipeline analytics the path turns out to be `http://localhost:8153/go/old_dashboard/assets/...` which would return a 404.

With - `http://localhost:8153/go/old_dashboard` (without trailing slash) if a request is made to fetch the pipeline analytics the path turns out to be `http://localhost:8153/go/assets/...`

GaneshSPatil:
@maheshp -- with the current fix, it would reduce the probability of the error happening (but won't eliminate the issue completely). 

With this fix, If the user clicks on the toggle dashboard button, the page would be redirected to `/go/old_dashboard` and things will work as expected.

But if a user manually lands on the URL `/go/old_dashboard/` or modifies the URL from `/go/old_dashboard` to `/go/old_dashboard/`, things will continue to break.

In addition to this fix, does it make sense to add another URL rewrite rule to redirect from `/go/old_dashboard/` to `/go/old_dashboard` ?
So  at URL rewrite:
- `/go/old_dashboard`  serves the old pipelines page
- `/go/old_dashboard/`  redirects to `/go/old_dashboard`

maheshp:
@GaneshSPatil I thought this would be simple fix for now since the old_dashboard would go away soon.  I think it is very unlikely for someone to add a trailing slash manually.

Issue #4604, opened 2018-04-04
Add an xsl tranform to convert <pluggableArtifact> to <artifact type="external">


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

The PR - https://github.com/gocd/gocd/pull/4278 removed the new XML tag `<pluggableArtifact>` and introduced a type attribute on `<artifact>`. Since the XSD having the element `<pluggableArtifact>` had already been checked in and released in 18.2, we need to add a config migration to convert `<pluggableArtifact>` to `<artifact type="external">` for those who are upgrading to 18.3. Even though we haven't announced `<pluggableArtifact>` as something that users can configure, it is possible that it is being used.  

The conversion - 

```xml
<pluggableArtifact id="artifactId" storeId="artifactStoreId">
  <property>
    <key>plugin_provided_config_key</key>
    <value>some_value</value>
  </property>
</pluggableArtifact>
```

to

```xml
<artifact id="artifactId" storeId="artifactStoreId" type="external">
  <property>
    <key>plugin_provided_config_key</key>
    <value>some_value</value>
  </property>
</artifact>
```

PR #4605, opened 2018-04-04
Edit 106.xsl to add the xsl transform to convert pluggable artifact to type external.

varshavaradarajan:
* 106.xsl had a bug that would have left any cruise-config.xml containing pluggableArtifact xml tag invalid.
* Users already having config xml with schema version 106 need not have to worry about the xsl this commit introduces as their config is already in a valid state - either they have manually converted pluggableArtifact tag to artifact type external or they did not have any pluggable artifact config.


varshavaradarajan:
https://github.com/gocd/gocd/issues/4604

Issue #4577, opened 2018-03-28
Have the same xml tag to represent fetch task for built in artifact and pluggable artifacts


varshavaradarajan:
##### Issue Type

- Feature enhancement

##### Summary

Currently, the fetch tasks look like - 

Built in fetch tasks 
```xml
<fetchartifact pipeline="pipelineName" stage="stageName" job="jobName" srcdir="dirName" dest="dest" />
<fetchartifact pipeline="pipelineName" stage="stageName" job="jobName" srcfile="fileName" dest="dest" />
```

Pluggable Artifact's fetch task
```xml
<fetchPluggableArtifact pipeline="pipelineName" stage="stageName" job="jobName" artifactId="artifactId">
  <configuration>
    <property>
      <key>plugin_provided_configuration_key</key>
      <value>foo_what_else</value>
    </property>
  </configuration>
</fetchPluggableArtifact>
```

Since the above two aren't different kinds of tasks, having different xml tags to represent them seems wrong. Planning on having an additional `type` attribute to distinguish between  - 

```xml
<fetchartifact type="external" pipeline="pipelineName" stage="stageName" job="jobName" artifactId="artifactId">
  <configuration>
    <property>
      <key>plugin_provided_configuration_key</key>
      <value>foo_what_else</value>
    </property>
  </configuration
</fetchartifact>
<fetchartifact type="builtin" pipeline="pipelineName" stage="stageName" job="jobName" srcdir="dirName" dest="dest" />
<fetchartifact type="builtin" pipeline="pipelineName" stage="stageName" job="jobName" srcfile="dirName" dest="dest" />
```

Need to take care of
1. The pipeline config API
2. Quick edit view to show the type attribute while configuring the fetch tasks
3. Old fetch task view to show the type attribute. 

Pending discussion with @naveenbhaskar around how the view should look like. We'll update this issue with mockups.

rajiesh:
Verified on 18.4.0 (6640-1fb2812cfa45eee8502e420443f2b46d81c51f98)

PR #4610, opened 2018-04-04
Add type attribute on fetchartifact (#4577)

varshavaradarajan:
* XSL to add type attribute with value builtin for existing fetchartifact tags in the xml
* XSL to convert fetchPluggableArtifact tag to fetchartifact with type external
* Use AttributeAwareConfigTag to differentiate between different types of fetchartifact tasks.

To check that fetchartifact is not broken -
- [x] Quick edit and pipeline config API
- [x] Template config API
- [x] Config repos
- [x] XML


varshavaradarajan:
@maheshp and I discussed what the representation should look like for this - 

We decided upon having - 

```
<fetchartifact pipeline="upstream" stage="defaultStage" job="defaultJob" srcfile="file_to_fetch" dest="dest_on_agent" origin="gocd" />

<fetchartifact pipeline="upstream" stage="defaultStage" job="defaultJob" srcdir="dir_to_fetch" dest="dest_on_agent" origin="gocd" />

<fetchartifact pipeline="upstream" stage="defaultStage" job="defaultJob" artifactId="artifactId" origin="external" />

```
An additional attribute called `origin` on the fetchartifact config will help us in differentiating between builtin and external artifacts to fetch while deserialize the config. The acceptable values for this is `gocd` and `external`

PR #4611, opened 2018-04-04
Remove dependency on spring-security

bdpiparva:
# Summary

The new implementation of spring security is a collection of the filter chain. The entry point for this is defined in `web.xml`

```xml
<filter>
    <filter-name>mainFilterChain</filter-name>
    <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
    <init-param>
      <param-name>targetBeanName</param-name>
      <param-value>mainFilterChain</param-value>
    </init-param>
</filter>
...
<filter-mapping>
    <filter-name>mainFilterChain</filter-name>
    <url-pattern>/*</url-pattern>
</filter-mapping>
```

## MainFilterChain
The main filter chain is a collection of filter chain and global filters. 

1. **ModeAwareFilter:** This filter is to block incoming `POST`, `PUT`, `PATCH` or `DELETE` request when a server is running in passive mode
2. **CreateSessionFilterChain:** This filter chain contains 2 filters -

    | Name | Description |
    | ------------------------------------ | ----------------------------------------------------------------------------------------------- |
    | AlwaysCreateSessionFilter            | Filter will create a new session if one not exist                                               |
    | ApiSessionReduceIdleTimeoutFilter    | This is to reduce `MaxInActiveInterval` on session for `/api/*`, `/cctray.xml`.                 |
3. **LocaleResolver:** Setup the locale for the request
4. **RememberLastRequestUrlFilterChain:** The filter remembers the last requested url(`GET` and `HEAD` only) except following URLs 
    - `/cctray.xml`
    - `/api/*`
    - `/remoting/**`
    - `/agent-websocket/**`
    - `/auth/**`
    - `/plugin/*/authenticate`
    - `/plugin/*/login`
    - `/assets/**` 
    - `/server/messages.json` 
    
5. **AuthenticationFilterChain:** The authentication filter chain contains a list of URLs and authentication filter for the URL.
It also contains some additional filters that applies before authentication filter such as - `InvalidateAuthenticationOnSecurityConfigChangeFilter` and `ReAuthenticationFilter`
  * `X509AuthenticationFilter`: This is to perform authentication using X509 certificate provided by the agent. 
  * `InvalidateAuthenticationOnSecurityConfigChangeFilter`: Invalidates authentication token for all user's when security config is changed.
  * `ReAuthenticationWithRedirectToLoginPage`: Performs re-authentication using existing `AuthenticationToken` when token is not valid or expired. On success, the filter will update the existing `AuthenticationToken` and continues to next filter in the chain. On failure, it will redirect the user to `/go/auth/login` page.
  * `ReAuthenticationWithChallenge`: Performs re-authentication using existing `AuthenticationToken` when a token is not valid or expired. On success, the filter will update the existing `AuthenticationToken` and continues to next filter in the chain. On failure, it will prepare a response as follow -
    
    - Show challenge(`WWW-Authenticate`) when security is enabled and request is api request but not an Ajax request
    - Set response status to `401`
    - Set `Content-Type` based on request `Accept`  header
    - Generate access denied message(`json or xml`) based on `Accept` header
    
  * `BasicAuthenticationWithChallengeFilter`: Performs the authentication if a request is not authenticated and `BASIC` auth credentials are provided in the request.
  * `ReAuthenticationWithRedirectToLoginPage`: Performs the authentication if a request is not authenticated and `BASIC` auth credentials are provided in the request.

| Ant url patter | Filters |
| ---------------------------------------- | ------------------------------------------------------------------------------------------- |
| `/remoting/**`                           | `X509AuthenticationFilter` |
| `/agent-websocket/**`                    | `x509AuthenticationFilter` |
| `/add-on/*/api/**`                       | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithRedirectToLoginPage` and`OauthAuthenticationFilter` |
| `/api/config-repository.git/**`          | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithChallenge` and `BasicAuthenticationWithChallengeFilter` |
| `/cctray.xml`                            | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithChallenge` and `BasicAuthenticationWithChallengeFilter` | 
| `/api/**`                                | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithChallenge` and `BasicAuthenticationWithChallengeFilter` |
| `/api/version`                           | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithRedirectToLoginPage` and `BasicAuthenticationWithRedirectToLoginFilter` |
| `auth/*`                                 | This url is open for all hence, `NoOpFilter` will applied to it. | 
| `/plugin/*/login`                        | This url is open for all hence, `NoOpFilter` will applied to it. |
| `/plugin/*/authenticate`                 | This url is open for all hence, `NoOpFilter` will applied to it. |
| `/**`                                    | `InvalidateAuthenticationOnSecurityConfigChangeFilter`, `ReAuthenticationWithRedirectToLoginPage` and `BasicAuthenticationWithRedirectToLoginFilter` |

6. **AssumeAnonymousUserFilter:** When `AuthenticationFilterChain` fails to authenticate request the filter will authenticate using anonymous authentication token.
7. **ThreadLocalUserFilter:** This will update the user in ThreadLocal for application use and at after serving request it will remove the user from ThreadLocal.
8. **AuthorizeFilterChain:** The filter chain contains mappings of URLs to required authority to access the URL. 
It also uses `ResponseHandler` to generate access denied message based on the request.
9. **DenyGoCDAccessForArtifactsFilterChain:** It will block GoCD access from artifacts.  
10. **ArtifactSizeEnforcementFilterChain:** The filter will check if enough disk space available while uploading artifact. The filter applies to `/files/**` and `/remoting/files/**`
11. **FlashLoadingFilter:** Load the flash from the session and add it to thread local.


server/src/main/java/com/thoughtworks/go/server/cache/UserCacheFactory.java
```diff
@@ -1,54 +0,0 @@
...
-import net.sf.ehcache.Cache;
-import net.sf.ehcache.CacheManager;
-import net.sf.ehcache.config.CacheConfiguration;
-import net.sf.ehcache.config.Configuration;
-import net.sf.ehcache.store.MemoryStoreEvictionPolicy;
-import org.springframework.cache.ehcache.EhCacheFactoryBean;
-import org.springframework.stereotype.Component;
-
-@Component
-public class UserCacheFactory {
```
bdpiparva: Replaced by a simple method that creates the cache when the `CachingSubjectDnX509PrincipalExtractor` needs it — no need for this to be a spring bean.

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchain/AgentRemotingFilterChain.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.stereotype.Component;
+import org.tuckey.web.filters.urlrewrite.UrlRewriteFilter;
+
+import javax.servlet.Filter;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+
+@Component("agentRemotingFilterChain")
+public class AgentRemotingFilterChain extends FilterChainProxy {
```
bdpiparva: This will basically replace these lines:

https://github.com/gocd/gocd/blob/7672cd956ef5c62f9957592ffc543d11509a31d6/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L42-L43

The 2 filters `x509ProcessingFilter` and `x509AccessDenied` are replaced by a simple `X509AuthenticationFilter`.

The filter `agentRemotingFilterInvocationInterceptor` is replaced by a simple `AuthorizeAgentFilterChain`.

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchain/AuthorizeAgentFilterChain.java
```diff
@@ -0,0 +1,41 @@
...
+import com.thoughtworks.go.server.security.GoAuthority;
+import org.springframework.security.util.FilterChainProxy;
+
+import javax.servlet.Filter;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+
+public class AuthorizeAgentFilterChain extends FilterChainProxy {
+    public AuthorizeAgentFilterChain() {
```
bdpiparva: This is a replacement for these lines:
https://github.com/gocd/gocd/blob/7672cd956ef5c62f9957592ffc543d11509a31d6/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L198-L215

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchain/AuthorizeAgentFilterChain.java
```diff
@@ -0,0 +1,41 @@
...
+        final LinkedHashMap<String, List<Filter>> filterChainMap = new LinkedHashMap<>();
+        final VerifyAuthorityFilter verifyAuthorityFilter = new VerifyAuthorityFilter(GoAuthority.ROLE_AGENT.asAuthority());
+        filterChainMap.put("/remoting/remoteBuildRepository", Arrays.asList(verifyAuthorityFilter));
+        filterChainMap.put("/remoting/files/**", Arrays.asList(verifyAuthorityFilter));
+        filterChainMap.put("/remoting/properties/**", Arrays.asList(verifyAuthorityFilter));
+        filterChainMap.put("/agent-websocket/**", Arrays.asList(verifyAuthorityFilter));
+        filterChainMap.put("/**", Arrays.asList(new DenyAllAccessFilter()));
```
bdpiparva: Make sure that any other `/remoting/*` calls are simply denied to prevent any abuse.

server/src/main/java/com/thoughtworks/go/server/newsecurity/filters/X509AuthenticationFilter.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.web.filter.OncePerRequestFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.security.cert.X509Certificate;
+
+public class X509AuthenticationFilter extends OncePerRequestFilter {
```
bdpiparva: This replaces the old filter defined here:

https://github.com/gocd/gocd/blob/7672cd956ef5c62f9957592ffc543d11509a31d6/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L163-L185

server/src/main/java/com/thoughtworks/go/server/newsecurity/x509/CachingSubjectDnX509PrincipalExtractor.java
```diff
@@ -0,0 +1,83 @@
...
+import net.sf.ehcache.*;
+import net.sf.ehcache.config.CacheConfiguration;
+import net.sf.ehcache.store.MemoryStoreEvictionPolicy;
+import org.springframework.dao.DataRetrievalFailureException;
+import org.springframework.security.ui.preauth.x509.SubjectDnX509PrincipalExtractor;
+import org.springframework.security.ui.preauth.x509.X509PrincipalExtractor;
+
+import java.security.cert.X509Certificate;
+
+public class CachingSubjectDnX509PrincipalExtractor implements X509PrincipalExtractor {
```
bdpiparva: Since parsing the X509 certificate to extract the subjectDN is a potentially expensive operation, we cache this parsing... Effectively a replacement for these lines: https://github.com/gocd/gocd/blob/7672cd956ef5c62f9957592ffc543d11509a31d6/server/webapp/WEB-INF/applicationContext-acegi-security.xml#L175-L178

server/src/main/java/com/thoughtworks/go/server/newsecurity/x509/CachingSubjectDnX509PrincipalExtractor.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final String CACHE_NAME = "agentCertificateCache";
+    private final SubjectDnX509PrincipalExtractor delegate = new SubjectDnX509PrincipalExtractor();
+
+    private final Ehcache cache;
+
+    public CachingSubjectDnX509PrincipalExtractor() {
+        this(createCacheIfRequired());
+    }
+
+    private static Ehcache createCacheIfRequired() {
```
bdpiparva: The cache has been internalized to simplify things and reduce yet another spring component.

bdpiparva:
### Todo
- [ ] Make sure authorization filter chain has the same config as it was in acegi.xml
- [ ] Make sure user gets`403` if not authorized to access URL
---
- Authentication flow
  - [ ] Password-based plugin
  - [ ] Web-based plugin

- Reauthentication flow
  - [ ] Password-based plugin
  - [ ] Web-based plugin
---
- [ ] Basic authentication on `/api/**` and `/cctray.xml` calls
  - [ ] With cURL - `should give 401 with WWW-Authenticate`
  - [ ] With Ajax - `should give 401 without challenge or redirect`
-  Basic authentication on everything else
 - [ ] Should redirect user to login page(`/go/auth/login`)
---
- [ ] Should remember the last URL before login and redirect the user to the same page after login
- [ ] Make sure `/api/**` has short lived session

ketan:
@rajiesh do you mind giving this a go locally?

@gocd/committers can you please look at this on priority, this is a fairly big, and high risk, considering security. I'd like to not delay merging of this as much as possible.

jetty9/src/main/java/com/thoughtworks/go/server/util/Jetty9Request.java
```diff
@@ -18,10 +18,16 @@
 
 import org.eclipse.jetty.server.Request;
 
+import javax.servlet.ServletRequestWrapper;
+
 public class Jetty9Request implements ServletRequest {
     private Request request;
 
     public Jetty9Request(javax.servlet.ServletRequest request) {
+        while (request instanceof ServletRequestWrapper) {
```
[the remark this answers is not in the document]
ketan: The jetty `Request` instance gets wrapped with `ServletRequestWrapper` implementations by several filters, since the other methods in this class require access to the underlying jetty `Request` we have to unwrap it completely — hence the `while`.

maheshp:
@bdpiparva @ketan I had look at the PR and tested few scenarios, few thoughts,

1. I found it a little tough to understand all the filters applied for a given URL. I had to start with `MainFilterChain` and go over each FilterChains to identify all the filters applied.
2. Since the `MainFilterChain` applies a set of filters/filter chains if we need to ignore a particular filter for a url pattern we will have to explicitly remove it in the appropriate filter chain.
3. We are authenticating in multiple places and logic is repeated as well in some cases, e.g `AbstractReAuthenticationFilter` and `AuthenticationController`. 
4. Since the authorization filters are applied post the authentication filters should we handle authentication before authorization as done earlier.
5. I am trying to understand the benefits of moving away from Spring Security since we are following the same authentication/authorization pattern provided by spring security but have to re-implement some of the things Spring provided like SecurityContext(which is moved to SessionUtils),  AuthenticationToken, RememberLastRequestUrlFilter, AlwaysCreateSessionFilter, ThreadLocalUserFilter.

ketan:
> 1. I found it a little tough to understand all the filters applied for a given URL. I had to start with MainFilterChain and go over each FilterChains to identify all the filters applied.
> 2. Since the MainFilterChain applies a set of filters/filter chains if we need to ignore a particular filter for a url pattern we will have to explicitly remove it in the appropriate filter chain.

Do you have any ideas about how to reduce this confusion — perhaps I'm biased because I understand this in and out, and I did not understand at all the previous implementation.

Yes — this is by design — the `MainFilterChain` basically gives a high level overview of what's involved. With each individual "filter" in the `MainFilterChain` being responsible for how to perform filtering on each URL. So in a way — the "high level" responsibility is based on the functionality, and not on the url pattern.

The mechanics of doing session management, authentication and reauthentication, url remembering and everything else is different for several url patterns. The old design basically added a lot of the filters everywhere. Quite a few of these filters were naturally choosing to "skip themselves" by checking for url patterns or [having](https://github.com/gocd/gocd/blob/4eb47523d843600ae992bdda6a2b3ceddb5bfdb5/server/src/main/java/com/thoughtworks/go/server/security/GoExceptionTranslationFilter.java) [different](https://github.com/gocd/gocd/blob/4eb47523d843600ae992bdda6a2b3ceddb5bfdb5/server/src/main/java/com/thoughtworks/go/server/security/DisallowExternalReAuthenticationFilter.java) [behavior](https://github.com/gocd/gocd/blob/4eb47523d843600ae992bdda6a2b3ceddb5bfdb5/server/src/main/java/com/thoughtworks/go/server/security/BasicAuthenticationFilter.java) based on the incoming request params (violating the SRP) — this added to a lot of confusion about the actual behavior of each filter, introducing a lot of the subtle bugs that were hard to debug and the behavior hard to reason.

We could have possibly squished all the filter chains into one giant filter chain — but I think that'll increase the duplication. We also discussed moving these filterchains into a single spring XML file like what we had earlier (except we'd still have multiple filter chains), to improve the readability. But the testability goes down significantly — moving it into XML would mean having to use spring to load up everything and significantly slow down testing.

> 3. We are authenticating in multiple places and logic is repeated as well in some cases, e.g AbstractReAuthenticationFilter and AuthenticationController.

This is a few lines of duplication. Thanks for identifying — we'll look into this.

> 4. Since the authorization filters are applied post the authentication filters should we handle authentication before authorization as done earlier.

That's pretty much how it is right now. Did I miss something obvious? https://github.com/gocd/gocd/blob/7ff588712a4d4be46c8038eb05d980df69f9b16b/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/MainFilterChain.java#L44-L53

> 5. I am trying to understand the benefits of moving away from Spring Security since we are following the same authentication/authorization pattern provided by spring security but have to re-implement some of the things Spring provided like SecurityContext(which is moved to SessionUtils), AuthenticationToken, RememberLastRequestUrlFilter, AlwaysCreateSessionFilter, ThreadLocalUserFilter.

To answer your question — this is because of the api breakage due to an upgrade to 3.x — and then another when we upgrade to 4.x and later to 5.x. Using spring security would have meant having to rewrite just about every spring security filter and exception handler — and not knowing what the current filters were doing meant that we could not simply migrate them to a newer version of spring-security. (I suppose this question should have been raised a long time ago — not when a PR is complete and good to be merged.)

maheshp:
> Do you have any ideas about how to reduce this confusion — perhaps I'm biased because I understand this in and out, and I did not understand at all the previous implementation.

Sorry, I don't have one as of now. Will think about it though.

> To answer your question — this is because of the api breakage due to an upgrade to 3.x — and then another when we upgrade to 4.x and later to 5.x. Using spring security would have meant having to rewrite just about every spring security filter and exception handler — and not knowing what the current filters were doing meant that we could not simply migrate them to a newer version of spring-security. (I suppose this question should have been raised a long time ago — not when a PR is complete and good to be merged.)

I haven't got a chance to look into what it takes to upgrade, so my ignorance bout could be the reason for my comment. Since you and @bdpiparva have spent time on it and know it best I will go with you guys on it.

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java
```diff
@@ -80,6 +81,8 @@ public AuthorizeFilterChain(AllowAllAccessFilter allowAllAccessFilter,
 
                 // all apis
                 .addAuthorityFilterChain("/cctray.xml", apiAccessDeniedHandler, ROLE_USER)
+                .addAuthorityFilterChain("/server/messages.json", apiAccessDeniedHandler, ROLE_USER)
+                .addAuthorityFilterChain("/pipelines.json", apiAccessDeniedHandler, ROLE_USER)
```
ketan: These are "API" it's a internal curl request from the browser, I suppose these should be serviced by the `/**` filter at the bottom.

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/RememberLastRequestUrlFilterChain.java
```diff
@@ -34,6 +34,7 @@ public RememberLastRequestUrlFilterChain() {
                 .addFilterChain("/plugin/*/login")
                 .addFilterChain("/assets/**")
                 .addFilterChain("/server/messages.json")
+                .addFilterChain("/pipelines.json")
```
ketan: Should we just add a `*.json` and `*.xml` here?

server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java
ketan: ignore me — i see why this is needed — but we may need this for other ajax URLs from the stage detail and job detail pages too.

Issue #4608, opened 2018-04-04
Get pipeline config API returns 500 if pluggableArtifact is configured 


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

Pipeline Config get API fails with the following error if a pluggableArtifact is configured. 

```
2018-04-04 08:41:39,091 ERROR [qtp391618063-30] Rails:-2 -
NoMethodError (undefined method `source' for #<Java::ComThoughtworksGoConfig::PluggableArtifactConfig:0x7b09b76>):
  app/presenters/api_v5/base_representer.rb:102:in `to_hash'
  app/presenters/api_v5/base_representer.rb:102:in `to_hash'
  app/presenters/api_v5/base_representer.rb:102:in `to_hash'
  app/presenters/api_v5/base_representer.rb:102:in `to_hash'
  app/controllers/api_v5/admin/pipelines_controller.rb:28:in `show'
  lib/jetty_weak_etag_middleware.rb:35:in `call'
  lib/catch_json_parse_errors.rb:23:in `call'
```
###### Basic environment details

* Go Version: `18.2.0`
* JAVA Version: `8`

##### Steps to Reproduce

1. Create a new pipeline through the Pipeline create wizard
2. Go to Admin -> Config XML, add the following the snippet to the job

```
<artifacts>
  <pluggableArtifact id="artifactId1" storeId="foo">
    <property>
      <key>BuildFile</key>
      <value>artifact_info.json</value>
    </property>
  </pluggableArtifact>
</artifacts>
```
3. Add the following snippet globally before `<pipelines>`

```xml
<artifactStores>
    <artifactStore id="foo" pluginId="cd.go.artifact.docker.registry">
      <property>
        <key>RegistryURL</key>
        <value>http://foo</value>
      </property>
    </artifactStore>
  </artifactStores>
```
4. Use Get Pipeline config API - https://api.gocd.org/current/#get-pipeline-config

##### Expected Results

HTTP Status Code 400, pipeline config json

##### Actual Results

HTTP Status: 500

PR #4615, opened 2018-04-05
Render bad request if the representer throws an exception while serializing or deserializing unsupported properties for resources.(#4608)

varshavaradarajan:


server/webapp/WEB-INF/rails.new/app/presenters/api_v2/base_representer.rb
```diff
@@ -97,11 +97,21 @@ def errors
     end
 
     def to_hash(*options)
-      super.deep_symbolize_keys unless @represented.nil?
+      begin
+        super.deep_symbolize_keys
+      rescue Exception => e
+        Rails.logger.error(e)
+        raise ApiV2::BadRequest, "API Version 2 does not support the contract for this request. Are you sure you're using the right API version?"
```
ketan: This can be cause if wrong data-types are submitted by the user and passed onto java. For e.g. calling `setName(String)` and the user sends a boolean for the `name` attribute. In which case, this error may not make sense.

varshavaradarajan: To not break old APIs, we introduce new properties, different value types, etc. If a user tries to do a get of say, pipeline config API using an old version (v3) for a pipeline that's been configured using say, v6, how do we let the users know that its wrong? It throws 500 currently, we are making it a bad request in this PR. Are you saying the error message should just be `check the logs`?

ketan: Not quite, I'm suggesting that what you're telling the user may actually not be correct in situations where users submit bad data (mostly unintentionally) and get an error which says something about "not supporting contract".

ketan: To give a more concrete example:

```
curl -s https://build.gocd.org/go/api/admin/pipelines/ketan-test \
  -H 'Accept: application/vnd.go.cd.v5+json' \
  -XPATCH -d '{"label_template": true, "name": "ketan-test"}' \
  -H'Content-Type: application/json' \
  -H'If-Match: "..."'

<!DOCTYPE html>
<html>
<head>
  <title>We're sorry, but something went wrong (500)</title>
...
</html>
```

varshavaradarajan: What do you suggest then as a better error message?

jyotisingh: What if we said something along the lines of :
```
There is something wrong with your request. This would usually happen if
* there is a API version mismatch
* you submitted an incorrect data-type for one or more elements
Please check logs. 
```
May be not exactly the same words, but something along those lines so the user gets a hint of what things to look at.

PR #4624, opened 2018-04-06
Add a handler to allow analytics plugins to open external links in new windows

marques-work:


maheshp:
@arvindsv I have no issues with this new message and handler.

PR #4625, opened 2018-04-06
Fix analytics modal sizing and title for old pipelines dashboard

marques-work:
This time, I promise not to break modals everywhere :).


marques-work:
@arvindsv so, basically, your life in in my hands!

marques-work:
**Before:**

screen shot 2018-04-06 at 3 32 16 pm

**After (ta-da!):**

image

@varshavaradarajan don't kill @arvindsv on account of my commit; do it because you want to.

PR #4630, opened 2018-04-09
Upgrading jetty from 9.2.18 to 9.4.9

jyotisingh:
One potential breaking change - it has support for TLS1.2 alone. Meaning it disables (by default) all cipher suites that matches ^.*_(MD5|SHA|SHA1)$ which are the only ones which work with TLS 1.0 and TLS1.1. So if any of the client uses TLS1.1 or TLS1.0 then the communication will fail. Need to test if this is an issue for us or not.
This version fixes some known issues with JDK9
fixes a class loader leak issue on the version we were on
bunch of bug fixes and a few feature enhancements


ketan:
If we merge this — we may have to enable the weak ciphers and give a timeline for deprecation. https://githubengineering.com/crypto-deprecation-notice/ is a good example.

ketan:
FYI: I don't think I have anything else more to review. As long as the tests go green, this is good to merge. I don't know if we we have functional tests for agent/server comm over websockets. Can @rajiesh confirm?

jyotisingh:
Well, I didn't look for it. I assumed there was one based on a conversation in past, may be I am mistaken. Will let @rajiesh confirm about the automated test. I will quickly test it out on local box.

jyotisingh:
Alright, we do have a test for it => https://build.gocd.org/go/tab/build/detail/smoke-PR/1376/Smoke/2/build-command-enabled and its broken, will get that fixed. Config => https://build.gocd.org/go/admin/templates/smoke-gradle/stages/Smoke/job/build-command-enabled/environment_variables

Issue #4627, opened 2018-04-09
Introduce the `<configuration>` subtag on `<artifact type="external">`


varshavaradarajan:
##### Issue Type

- Feature enhancement

##### Summary

In order to be consistent with https://github.com/gocd/gocd/issues/4577, introduce the `<configuration>` subtag on `<artifact type="external">`. 

This includes a config migration to push the `<property>` to the `<configuration>` subtag

rajiesh:
Verified just the validation around configuration sub tag added as part of 18.4.0 release

PR #4631, opened 2018-04-09
Add configuration subtag to external artifacts. (#4627)

varshavaradarajan:


config/config-api/src/main/java/com/thoughtworks/go/config/PluggableArtifactConfig.java
```diff
@@ -150,9 +160,8 @@ public ConfigErrors errors() {
         return errors;
     }
 
-    @Override
     public boolean hasErrors() {
-        return super.hasErrors() || !errors.isEmpty();
+        return !errors.isEmpty();
```
akshaydewan: Do we need `|| configuration.hasErrors()` ?

varshavaradarajan: I think so. I missed the `configuration.validateUniqueness` in `#validate`, which adds errors to the configuration object. I'll fix this.

config/config-server/src/main/resources/cruise-config.xsd
```diff
@@ -727,7 +727,7 @@
 
   <xsd:complexType name="artifactType">
     <xsd:sequence>
-      <xsd:element name="property" minOccurs="0" maxOccurs="unbounded" type="pluginPropertyType"/>
+      <xsd:element name="configuration" minOccurs="0" maxOccurs="unbounded" type="configurationType"/>
```
akshaydewan: I think the `maxOccurs` should be `1`, since we can have only one `<configuration>`

server/src/test-integration/java/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -2104,6 +2104,61 @@ public void shouldConvertPluggableArtifactTagToArtifactWithTypeOnPluggableArtifa
...
+                + "             </job>"
+                + "             </jobs>"
+                + "         </stage>"
+                + "      </pipeline>"
+                + "    </pipelines>"
+                + "</cruise>";
+
+        String migratedContent = migrateXmlString(configXml, 106);
+        String migratedArtifact1 = "<artifact type=\"external\" id=\"artifactId1\" storeId=\"foo\"/>";
+        String migratedArtifact2 = "<artifact type=\"external\" id=\"artifactId2\" storeId=\"foo\"><configuration>"
```
akshaydewan: The test fails. storeId should be `foobar`

varshavaradarajan:
Squashed the commits. The PR build had gone through - https://build.gocd.org/go/pipelines/value_stream_map/trigger/2650

PR #4636, opened 2018-04-09
Don't need overflow-y on content container as we want the entire page to scroll as content grows

marques-work:
Before

image

After

image


arvindsv:
I guess it's still be needed, even if it's horizontal? I agree. We shouldn't need to specify overflow.

PR #4644, opened 2018-04-10
Removing pipeline instance rendering from environments page. Don't kn…

jyotisingh:
…ow of a good reason for it, its non-performant. Added a flag to enable it just in case someone really cares.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
+++ b/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
@@ -220,6 +220,8 @@ public class SystemEnvironment implements Serializable, ConfigDirProvider {
     private static GoSystemProperty<Integer> GO_SPA_REFRESH_INTERVAL = new GoIntSystemProperty("go.spa.refresh.interval", 10000);
 
     private static GoSystemProperty<Boolean> ENABLE_PIPELINE_ANALYTICS_ONLY_FOR_ADMINS = new GoBooleanSystemProperty("go.enable.pipeline.analytics.only.for.admins", false);
+    private static final GoSystemProperty<Boolean> DISPLAY_PIPELINE_INSTANCES_ON_ENVIRONMENTS_PAGE = new GoBooleanSystemProperty("gocd.environments.show.pipelines", false);
+
     private final static Map<String, String> GIT_ALLOW_PROTOCOL;
 
     static {
@@ -873,6 +875,10 @@ public class SystemEnvironment implements Serializable, ConfigDirProvider {
 
     public Integer getNotificationListenerCountForPlugin(String pluginId) {
         return Integer.parseInt(getPropertyImpl("plugin." + pluginId + ".notifications.listener.count", "1"));
+    }
+
+    public boolean displayPipelineInstancesOnEnvironmentsPage(){
+        return get(DISPLAY_PIPELINE_INSTANCES_ON_ENVIRONMENTS_PAGE);
     }
 
     public boolean enablePipelineAnalyticsOnlyForAdmins() {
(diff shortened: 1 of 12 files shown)


jyotisingh:
BTW, this is what the page would look like with this commit

screen shot 2018-04-10 at 3 45 49 pm

arvindsv:
I like it. I'd prefer we remove that meaningless globe icon and put in a gear icon next to the name. Also we can make the name not clickable.

jyotisingh:
Sure. That I can do :)

ketan:
It looks even more weird now — it's no longer a link anymore — can we keep the existing style (blue text) along with the gear on the side?

jyotisingh:
@ketan - do you mean just changing the text color to blue or change it back to a link?

anushr:
@jyotisingh Seeing all the environment pipelines was tremendously useful for us. It gave a quick snapshot of the versions deployed on each component of each environment. Very useful when one has a lot of these.

You could argue that I can see all this on the dashboard; and I can. However, it's painful to have to filter the dashboard each time to do this. We have 100+ build pipelines and around 20 deploy pipelines.

One new function that makes life a lot easier is that the pipeline search is now a `#!` on the location and can be bookmarked. This is awesome -- thank you for that! But what would really want me to not care about the Environments page anymore is for the ability to save filtered dashboard views.

In any case, I whent through your PR saw the flag `gocd.environments.show.pipelines`. I'll give that a try.

jyotisingh:
Thanks for the feedback @anushr 

We felt that the only benefit this page provided was to group pipelines based on the environments (which is valuable). I am sure you know the pipeline widget on this page wasn't the same as the one on dashboard. Based on our interactions with GoCD users so far, no one really cared about the extra info these widgets provided, so we decided to kill them since the queries used for rendering the extra info on the widgets used to perform pretty badly. 

We are aware that the filtering on Dashboard is quite restrictive and a bit difficult to use. Also the grouping is based on Pipeline-Groups alone which doesn't fit everyone's need. You are right that people would want to see their pipelines grouped by environments too. 
https://github.com/gocd/gocd/issues/3842 is a story to improve the experience around searching and filtering and saving the views for the pipelines based on filters such as groups, environments - there might other kind of filters too. Since you seem interested in this, you might want to go through that story and put in your thoughts. 

`gocd.environments.show.pipelines` flag was added considering usecase like yours, so its fair to use it meanwhile. Just so you know, once we provide support for filtering pipeline based on environments, we would be getting rid of that flag. As far as the existing environments page is concerned, we plan to kill that page too eventually. There would be a separate config-only page available under the `Admin` menu.

BenWolstencroft:
I'd like to add to this, while i agree that the page was unusable with all the pipeline data shown, i think it would be useful still to simply show a list of the pipelines (with links to said pipelines) beneath each environment header (also maintaining the No pipelines in this environment message that used to appear), this wouldn't have the overhead that the old view had, but would give an at a glance view of which of your pipelines were in which environment without having to click on the cog icon on each to find out.

ketan:
Given that the new dashboard is fairly quick now, how hard would it be to partition the data based on the environment name and render it the old way on a (new) environments page?

varshavaradarajan:
@ketan - what would be the point of showing pipeline instances on the environment page? People use the page to know which environment a pipeline belongs to. Duplicating the view here seems unnecessary? @maheshp had brought up sorting of pipelines on the dashboard based on environments.

BenWolstencroft:
Just to add to my previous message, in our use case we have approximately 1000 deploy pipelines spread over 20 or so environments, the old environments page (when it eventually loaded) was useful to show the distribution of software across those environments, if functionality is added to the dashboards page to optionally view grouped by environment or pipeline group as an in UI option this would give us the functionality and actually be more useful than the old environments page was!

PR #4660, opened 2018-04-16
Making a bunch of APIs to support pipeline-counter instead of both counter and label

jyotisingh:
Artifacts, Stage rerun, Properties API supported (but not documented) both pipeline counter and label.
In order to function, these queries required a pipeline-identifier or a stage-identifier. The information passed along in the API was enough to construct these, but since a user could technically send pipeline-label, there was a db lookup done for the pipeline using the label. I think this is un-necessary. 
[RestfulService.findJob](https://github.com/gocd/gocd/blob/98734997e24fb181f8e8a7858fba43a6d04ac207/server/src/main/java/com/thoughtworks/go/server/service/RestfulService.java#L42) takes care of checking for presence of pipeline in the config already. 

More importantly, the logic for this is very buggy. [PipelineSqlMapDao.findPipelineByCounterOrLabel](https://github.com/gocd/gocd/blob/98734997e24fb181f8e8a7858fba43a6d04ac207/server/src/main/java/com/thoughtworks/go/server/dao/PipelineSqlMapDao.java#L788) checks if the passed along string is a number or a string and uses that to decide it its a counter or a label. While this may be okay for a simple enough pipeline, it causes problems in certain cases for example: Consider `Upstream` pipeline with labeltemplate `$COUNTER`. If there is a `Downstream` pipeline which depends on `Upstream` and has labelTemplate = ${UPSTREAM}, then `PipelineSqlMapDao.findPipelineByCounterOrLabel` will always return an incorrect instance of the pipeline.

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/domain/StageIdentifierTest.java
+++ b/common/src/test/java/com/thoughtworks/go/domain/StageIdentifierTest.java
@@ -1,18 +1,18 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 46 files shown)


ketan:
Most of the changes look OK, can you rename the variable `counter` to be `pipelineCounter` to make it explicit?

jyotisingh:
Sure. 
Still testing it out, fixing bugs and broken tests as I find them. So, its not ready to merge just yet. Will let you know once I am done with all the changes

ketan:
Ping @jyotisingh ? do you need any help with this PR?

jyotisingh:
No, as far as I recollect, the work was complete. The builds were failing for some issues with the agents, so I was waiting for them to be stable again but later it went off my mind I guess. Let me rebase and trigger off a build. Will let you know if I need any help, Thanks @ketan .

akshaydewan:
Closing this in favour of PR #5256

PR #4661, opened 2018-04-16
Allow URL query parameter pass through to analytics iframes

marques-work:
- Only passes through a white list of key-val pairs (or optionally just key)
  - Currently pass through ?ui=test to toggle the ui variant feature
  - Also clean up iframe widget (DRY)


maheshp:
@marques-work we spoke to @arvindsv to get the intent of this PR. Will you be able to add some tests for the changes please.

marques-work:
Sure, I will do that

marques-work:
@maheshp done, with some additional cleanup

PR #4671, opened 2018-04-19
Refactor analytics page

GaneshSPatil:


server/webapp/WEB-INF/rails.new/webpack/single_page_apps/analytics.js
```diff
@@ -14,44 +14,51 @@
...
   PluginEndpoint.ensure("v1");
+  const analyticsElem = $('.analytics-container');
+  const pipelines     = JSON.parse(analyticsElem.attr('data-pipeline-list'));
 
-  document.addEventListener("DOMContentLoaded", () => {
-    const main = document.querySelector("[data-supported-dashboard-metrics]");
+  const onSuccess = (pluginInfos) => {
+    const plugins = pluginInfos.mapPluginInfos((p) => p.id());
+    const metrics = {};
+    pluginInfos.eachPluginInfo((p) => {
```
maheshp: Use `mapPluginInfo` instead?

GaneshSPatil: Can you provide an example, how to do it using map? I thought, we need a reduce, but the mixin does not provide reduce functionality.

PR #4677, opened 2018-04-23
Remove unwanted routes from the code

ketan:


jyotisingh:
`/rss.xml` was provided for the very old brown dashboard
`/atom` based urls were provided for the old pipeline-history and stage-details page.
None of these are available anymore, so safe to kill the urls.

PR #4679, opened 2018-04-23
Added a wrapper over the x509trustManager so we can log certificate v…

jyotisingh:
…alidation failures in server log

### DIFF
--- a/jetty9/src/main/java/com/thoughtworks/go/server/util/CustomSslContextFactory.java
+++ b/jetty9/src/main/java/com/thoughtworks/go/server/util/CustomSslContextFactory.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.util;
+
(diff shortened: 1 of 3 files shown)


ketan:
@jyotisingh - rebase and kickoff a build?

jetty9/src/main/java/com/thoughtworks/go/server/util/CustomSslContextFactory.java
```diff
@@ -0,0 +1,76 @@
...
+            this.trustManager = trustManager;
+        }
+
+        @Override
+        public void checkClientTrusted(X509Certificate[] x509Certificates, String s) throws CertificateException {
+            try {
+                trustManager.checkClientTrusted(x509Certificates, s);
+            } catch (Exception e) {
+                String name = x509Certificates.length >= 1 ? x509Certificates[0].getSubjectDN().getName() : "NONAME";
+                LOGGER.error("Client certificate was found to be invalid. Subject: [{}] Error: [{}].", name, e.getMessage(), e);
```
ketan: This PR looks good to me — however a question remains whether this log message is sufficient to tell the difference between a good certificate and a bad one. For one, the subject DN is bound to a certificate created when the agent registered. The log message does not talk about the certificate chain (which is really what we are verifying)

 Perhaps dumping the entire incoming certificate via `toString()` helps along with the `toString()` of the CA certificate chain that the server is validating the incoming certificate against?


Perhaps something along the lines of (untested code, but hopefully shows intent):


```java
        @Override
        public void checkClientTrusted(X509Certificate[] x509Certificates, String s) throws CertificateException {
            try {
                trustManager.checkClientTrusted(x509Certificates, s);
            } catch (Exception e) {
                LOGGER.error("Client certificate was found to be invalid", e);
                LOGGER.error("Client certificate chain:");
                for (int i = 0; i < x509Certificates.length; i++) {
                    X509Certificate x509Certificate = x509Certificates[i];
                    LOGGER.error("chain[" + i + "]:");
                    LOGGER.error(x509Certificate.toString());
                }

                LOGGER.error("The accepted certificates are:");
                X509Certificate[] acceptedIssuers = getAcceptedIssuers();
                for (X509Certificate acceptedIssuer : acceptedIssuers) {
                    LOGGER.error(acceptedIssuer.toString());
                }

                throw e;
            }
        }
```

PR #4684, opened 2018-04-25
Generating a uid for the new dashboard analytics frame

ibnc:
* Also adding the fetch analytics handler to both pipeline analytic
views

Pretty minor change, but needed for other work I'm doing. It also adds consistency.

@arvindsv @maheshp @GaneshSPatil could one of you look at this, and hopefully merge it soon?

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_analytics_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_analytics_widget.js.msx
@@ -24,11 +24,16 @@ const PluginEndpointRequestHandler = require('rails-shared/plugin-endpoint-reque
 const Frame                        = require('models/shared/frame');
 
 
+const models = {};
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_analytics_widget.js.msx
```diff
@@ -24,11 +24,16 @@ const PluginEndpointRequestHandler = require('rails-shared/plugin-endpoint-reque
 const Frame                        = require('models/shared/frame');
 
 
+const models = {};
+PluginEndpointRequestHandler.defineLinkHandler();
+PluginEndpointRequestHandler.defineFetchAnalyticsHandler(models);
```
GaneshSPatil: For the pipeline analytics, I assume we don't require linkHandler, but we will need FetchAnalyticsHandler for the charts with configurable time. (Eg. data for a week, month, etc)?

PR #4689, opened 2018-04-26
Making VSM handle pipeline-names caseinsensitively

jyotisingh:
Without these changes the VSM of a pipeline whose downstream pipelines referred to it using a different case would not show up the downstream pipelines at all. Same goes for the VSM of the parent SCM material.


server/src/main/java/com/thoughtworks/go/domain/valuestreammap/Node.java
```diff
@@ -24,7 +25,7 @@
 import java.util.Set;
 
 public abstract class Node implements Comparable<Node>{
-    protected final String id;
+    protected final CaseInsensitiveString id;
```
bdpiparva: IMO nodeId should be case sensitive at least for material `Fingerprint`. WDYT?

If yes then parameterize the type of Node Id

### Node.java
```java
public abstract class Node<NODEIDTYPE> implements Comparable<Node>{
protected final NODEIDTYPE id;
public Node(DependencyNodeType dependencyNodeType, NODEIDTYPE nodeId, String nodeName) {
        this.id = nodeId;
        this.type= dependencyNodeType;
        this.nodeName = nodeName;
    }
}
```

### SCMDependencyNode.java
```java
public class SCMDependencyNode extends Node<String> {
}
```

and

### PipelineDependencyNode.java
```java
public class PipelineDependencyNode extends Node<CaseInsensitiveString> {
}
```

bdpiparva: We never know what will be a material fingerprint. Fingerprint generation algorithm varies from material to material.

jyotisingh: The algorithm doesn't really change between materials, its always hash produced by sha-256 which is hex encoded to form a fingerprint. Its the parameters passed along to the algorithm which change based on the material type. So, in that sense its feels okay to consider the fingerprint to be case-insensitive. Although, the db column for fingerprint might not be case-insensitive so that's something we might have to fix too.

PR #4706, opened 2018-05-04
Optimization of `GoDashboardCurrentStateLoader`

ketan:
Do not perform the DB query to load pipelines if list of pipelines has
not changed.

A further optimization would be to load from the DB, only the pipelines
that have been newly added. Pipelines that have been removed do not
require a DB call, and we can simply purge those from the
`historyForDashboard` instance variable.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCurrentStateLoader.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCurrentStateLoader.java
@@ -59,6 +59,8 @@ public class GoDashboardCurrentStateLoader {
     private GoConfigPipelinePermissionsAuthority permissionsAuthority;
     private TimeStampBasedCounter timeStampBasedCounter;
     private boolean everLoadedCurrentState = false;
+    private PipelineInstanceModels historyForDashboard;
+    private Set<CaseInsensitiveString> previousPipelineNames;
 
     @Autowired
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCurrentStateLoader.java
```diff
@@ -76,8 +78,14 @@ public GoDashboardCurrentStateLoader(PipelineDao pipelineDao, TriggerMonitor tri
...
 
     public List<GoDashboardPipeline> allPipelines(CruiseConfig config) {
-        List<String> pipelineNames = CaseInsensitiveString.toStringList(config.getAllPipelineNames());
-        PipelineInstanceModels historyForDashboard = loadHistoryForPipelines(pipelineNames);
+        List<CaseInsensitiveString> allPipelineNames = config.getAllPipelineNames();
+
+        HashSet<CaseInsensitiveString> newPipelineNames = new HashSet<>(allPipelineNames);
+        if (!newPipelineNames.equals(previousPipelineNames)) {
+            List<String> pipelineNames = CaseInsensitiveString.toStringList(allPipelineNames);
+            historyForDashboard = loadHistoryForPipelines(pipelineNames);
```
jyotisingh: Dont we need to update `historyForDashboard` on pipeline-config-api callback?

ketan: which callback is this?

PR #4708, opened 2018-05-05
[Experimental - don't merge yet] Add Stage charts to Analytics Dashboard

marques-work:


marques-work:
@arvindsv @tinozza This PR will place stage build time chart as a companion to the pipeline build time chart in the analytics dashboard. I don't know if this is the right place, but wanted to make sure it was on your radar. Here's how it currently looks in this PR:

image

marques-work:
I don't think we want this anymore.

Issue #4698, opened 2018-04-30
Long pipeline names get truncated


JorritSalverda:
##### Issue Type

- Feature enhancement

##### Summary

With the new dashboard pipeline names get truncated to get a better design. However this makes it very unusable for longer pipeline names where the distinguishing part is in the end of the name, see screenshot.

![20180430-truncated-pipeline-names](https://user-images.githubusercontent.com/136472/39421543-d4dd86ae-4c69-11e8-9abe-aebffe3ac193.png)

Not sure how to solve this best in order to maintain the clean design while being able to read longer pipeline names without needing to hover over them first.

###### Basic environment details

* Go Version: `18.3.0`

ketan:
@naveenbhaskar @tinozza thoughts?

ketan:
For reference, the old dashboard wrapped around the pipeline name:

<img width="1186" alt="screen shot 2018-05-04 at 4 15 34 pm" src="https://user-images.githubusercontent.com/10598/39624298-6cf80b68-4fb6-11e8-8eb7-192dc844899a.png">
<img width="1417" alt="screen shot 2018-05-04 at 4 15 39 pm" src="https://user-images.githubusercontent.com/10598/39624299-6d223c76-4fb6-11e8-8629-2c810d730369.png">

PR #4709, opened 2018-05-07
[New Dashboard] Do not truncate longer pipeline name (#4698)

GaneshSPatil:
* Show the entire pipeline name with wrapped words.
* Remove onhover pipeline name tooltip as the entire pipeline name is visible

screen shot 2018-05-07 at 10 31 58 am

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_header_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_header_widget.js.msx
@@ -68,7 +68,7 @@ const PipelineHeaderWidget = {
     return (
       <div class="pipeline_header">
         <div class="pipeline_sub_header">
-          <h3 className="pipeline_name" title={pipeline.name}> {pipeline.name} </h3>
+          <h3 className="pipeline_name">{pipeline.name}</h3>
           <div className="pipeline_actions">
             {analyticsIcons}
             {settingsButton}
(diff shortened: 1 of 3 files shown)


ketan:
@GaneshSPatil — if the name is too long, does this wrap into 3 or 4 lines, should it wrap around so much?

Can we consider limiting based on the number of lines as we do for the build cause popup here:

https://github.com/gocd/gocd/blob/fc405f37e3ee24f742157b1cada57cc0801a7ab5/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss#L852-L854

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/pipeline_header_widget.js.msx
```diff
@@ -68,7 +68,7 @@ const PipelineHeaderWidget = {
     return (
       <div class="pipeline_header">
         <div class="pipeline_sub_header">
-          <h3 className="pipeline_name" title={pipeline.name}> {pipeline.name} </h3>
+          <h3 className="pipeline_name"> {pipeline.name} </h3>
```
ketan: Trim the whitespace around this, please.

GaneshSPatil:
>  if the name is too long, does this wrap into 3 or 4 lines, should it wrap around so much?

I am not sure if the pipeline name will ever exceed 3 lines (22 chars in a line, 66 chars in 3 lines).  Also wont the same problem exists if we truncate it?

_For example_: If users have pipeline-names `Org1-team1...my-fancy-pipeline1`, `Org1-team1...my-fancy-pipeline2` and so on.. Truncating the pipeline name will result in the similar problem?

GaneshSPatil:
Pipeline Name could be of length 255 chars. 
Attached a screenshot of the pipeline with 255 char long pipeline name.

screen shot 2018-05-07 at 11 08 56 am

varshavaradarajan:
@GaneshSPatil - is there a limit on the length of pipeline name while creating it?

ketan:
The XSD allows for 255 char names.

Issue #4682, opened 2018-04-24
Using gocd python library with correct credentials getting status code 401


vkrisha:
Bug

i have a script using gocd python library to connect to goserver and get pipeline status, it was working fine on GOCD 16.x but when i upgraded to 18.3 it always throwing 401 with correct credentials. can you help me on this.

ketan:
Can you provide more details on what API calls the library is making? The issue template asks for steps to reproduce that you've not provided — so I don't think we can provide a fix without that information.

ketan:
I'm sorry I don't know what python module you're referring to. Can you provide a curl or a simple script that I can run locally?

vkrisha:
i'm not using curl for this, well the script is simple
if you want to try in python,

pip install gocd
then

save the below code in a ex.py file (change server, user , password and pipline name)

from gocd import Server
  server_connection = Server("http://localhost:8153", user = "abc", password = "Enp1")
  a = server_connection.pipeline("Test1")
  b = a.history()
  print b.status_code

varshavaradarajan:
@vkrisha - are you sure the user `abc` has still got the right permission to access the API?

vkrisha:
yes, as i said it was working in go 16.x and when i have upgraded to 18.3 it is not working, even if manually login with "abc" its success.

varshavaradarajan:
@vkrisha - this python library is maintained by a community member @gaqzi . Please raise the issue there - https://github.com/gaqzi/py-gocd/issues

standingb76:
Here is the exception in debug mode:
send: 'GET /go/api/config/pipeline_groups HTTP/1.1\r\nAccept-Encoding: identity\r\nHost: something:8153\r\nConnection: close\r\nUser-Agent: py-gocd\r\n\r\n'
reply: 'HTTP/1.1 401 Unauthorized\r\n'
header: Date: Wed, 25 Apr 2018 14:57:33 GMT
header: Set-Cookie: JSESSIONID=qwdwdwd;Path=/go;Expires=Wed, 09-May-2018 14:57:33 GMT;HttpOnly
header: Expires: Thu, 01 Jan 1970 00:00:00 GMT
header: X-XSS-Protection: 1; mode=block
header: X-Content-Type-Options: nosniff
header: X-Frame-Options: SAMEORIGIN
header: X-UA-Compatible: chrome=1
header: Connection: close
Traceback (most recent call last):
  File "gocdAdmin.py", line 58, in <module>
[...]

PR #4711, opened 2018-05-07
Return before adding the WWW-Authenticate header only for unauthenticated API XHRs. (#4682)

varshavaradarajan:
* Use the BasicProcessingFilterEntryPoint to respond for non XHR API requests that are unauthenticated.


varshavaradarajan:
@ketan, @bdpiparva  - `GoExceptionTranslationFilter` and `BasicProcessingFilterEntryPoint` will be removed in #4611, right ?

ketan:
> @ketan, @bdpiparva - GoExceptionTranslationFilter and BasicProcessingFilterEntryPoint will be removed in #4611, right ?

Yes.

PR #4713, opened 2018-05-07
Pipeline case insensitivity part2

jyotisingh:
Making some cache-keys case-insensitive. Cache-keys were created based on the case provided by end-user (which need not always match the one used in config), however cache invalidation would happen for the same case as used by the config. So there were many cases of stale cached information being displayed if a pipeline was accessed via API or UI with a different case from the one in config.

Changes in this commit affects the following
* materialRepository.hasPipelineEverRunWith (scheduling - to decide if the pipeline should be triggered or not)
* stageSqlMapDao.getCount case-insensitive (get-stage-history api)
* StageSqlMapDao.getAllRunsOfStageForPipelineInstance (Stage details page)
* stageSqlMapDao.findStageHistoryForChart (stage details graph)
* stageSqlMapDao.findAllStagesFor (for preparing VSM, stage-active-checker used by unlock-api, dashboard etc)
* stageSqlMapDao.mostRecentJobsForStage (get-pipeline-status api, )
* stageDao.findStageHistoryPage, stageDao.findDetailedStageHistoryByOffset (Stage history pagination)
* pipelineDao.findLastSuccessfulStageIdentifier (get-pipeline-status api)
* Making pipeline, stage and job identifiers case insensitive
* Mutex for pipeline, stage, jobs should apply immaterial of case of names defined in config or passed along by user
* HealthState scope for pipeline, stage, job need to be case-insensitive

### DIFF
--- a/common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java
+++ b/common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java
@@ -42,19 +42,21 @@ public class HealthStateScope implements Comparable<HealthStateScope> {
     }
 
     public static HealthStateScope forPipeline(String pipelineName) {
-        return new HealthStateScope(ScopeType.PIPELINE, pipelineName);
+        return new HealthStateScope(ScopeType.PIPELINE, pipelineName.toLowerCase());
     }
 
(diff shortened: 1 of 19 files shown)


jyotisingh:
Closing this since @rajiesh thinks this might be a risky change to make.

PR #4720, opened 2018-05-09
New SCM modal does not appear for plugin when "Create New" button is clicked on pipeline config page.

akshaymayekar:
1. picking the plugin version from **about** object instead of **plugininfo**.


GaneshSPatil:
LGTM. Can be merged once the merge window is open!

PR #4722, opened 2018-05-09
Changing status code for all apis and authentication filters

bdpiparva:
1. Authentication failure
All authentication filter will return `401(UNAUTHORIZED)` in case of any failure while authenticating user.
Read more [RFC#401](https://tools.ietf.org/html/rfc7235#section-3.1)

2. User permission check at controller level and in (`AuthorizeFilterChain`)
All controller will now return `403(FORBIDDEN)` in case of a user is not authorized to access the specified resource
Read more [RFC#403](https://tools.ietf.org/html/rfc7231#section-6.5.3)

### DIFF
--- a/api/api-base/src/test/groovy/com/thoughtworks/go/api/SecurityTestTrait.groovy
+++ b/api/api-base/src/test/groovy/com/thoughtworks/go/api/SecurityTestTrait.groovy
@@ -46,7 +46,7 @@ trait SecurityTestTrait {
 
   abstract void makeHttpCall()
 
-  def assertRequestAuthorized() {
+  def assertRequestAllowed() {
     verify(controller)."${controllerMethodUnderTest}"(any(), any())
 
     ((MockHttpServletResponseAssert) assertThatResponse())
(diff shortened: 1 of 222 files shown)


api/api-base/src/test/groovy/com/thoughtworks/go/api/SecurityTestTrait.groovy
```diff
@@ -61,6 +61,6 @@ trait SecurityTestTrait {
     ((MockHttpServletResponseAssert) assertThatResponse())
       .hasContentType(controller.mimeType)
       .hasStatus(401)
```
ketan: Should be 403, no?

common/src/main/java/com/thoughtworks/go/server/service/result/OperationResult.java
```diff
@@ -43,6 +43,8 @@
 
     ServerHealthState unauthorized(String message, String description, HealthStateType id);
```
ketan: From my understanding `unauthorized` was always used instead of `forbidden`. I think it's probably safe to rename this message to `forbidden` instead. By extension, the `forbidden` method you added below should not be needed.

@jyotisingh WDYT?

jyotisingh: Yes, makes sense. 
If an unauthenticated user tries to access a url which doesn't allow anonymous access, I believe `BasicAuthenticationWithChallengeFilter` would return 401 so the request would never reach the controller, so I think we are covered for that too.

bdpiparva: > From my understanding unauthorized was always used instead of forbidden. I think it's probably safe to rename this message to forbidden instead. By extension, the forbidden method you added below should not be needed.

I will delete `unauthorized ` method

PR #4724, opened 2018-05-09
Fixing issue where a pipeline created via API would show up twice on …

jyotisingh:
…the dashboard


server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCurrentStateLoader.java
```diff
@@ -230,6 +231,17 @@ private Permissions permissionsFor(PipelineConfig pipelineConfig, Map<CaseInsens
 
     public void reset() {
         historyForDashboard = PipelineInstanceModels.createPipelineInstanceModels();
-        previousPipelineNames = new HashSet<>();
+        lastKnownPipelineNames = new HashSet<>();
+    }
+
+    public void clearEntryFor(CaseInsensitiveString deletedPipeline) {
```
ketan: Call this method around here? https://github.com/gocd/gocd/blob/e954d289bb285525e03fd8be858508f26ce0588b/server/src/main/java/com/thoughtworks/go/server/dashboard/GoDashboardCurrentStateLoader.java#L95-L102

Issue #4338, opened 2018-02-13
QuickEdit does not allow delete of stage


sriv:
##### Issue Type

- Bug Report

##### Summary

QuickEdit in a pipeline does not allow me to delete a stage. On deleting and "save" it throws an error saying:

`There are errors on the page, fix them and save`.

No network calls were made.

###### Basic environment details

* Go Version: `18.1.0`
* Browser vendor and version (if relevant): `Hopefully not Internet Explorer 7 :)` No IE, Chrome 64.

##### Steps to Reproduce

1.  Open a pipeline with multiple stages for edit in QuickEdit
2.  Select the stage to be deleted, click on the `Trash` icon on the right panel
3.  Click `Save` on the top of the page.

##### Expected Results
Should save pipelines with the edits.

##### Actual Results
It removes the stage to be deleted, and adds a new empty stage, which fails the mandatory validation.

PR #4734, opened 2018-05-14
QuickEdit does not allow delete of stage (fixed)

AksMKR:
disabling the delete button when only one stage is left in the pipeline.
Issue : https://github.com/gocd/gocd/issues/4338


ketan:
Is this still WIP?

AksMKR:
No. removed WIP tag.

server/webapp/WEB-INF/rails.new/webpack/views/pipeline_configs/stages_config_widget.js.msx
```diff
@@ -206,9 +206,19 @@ const StageConfigDefinitionWidget = {
     const stage      = vnode.attrs.stage;
     const stageIndex = vnode.attrs.pipeline().stages().indexOfStage(vnode.attrs.stage);
 
+    let removebutton = null;
+    if (vnode.attrs.pipeline().stages().isLastStage()) {
```
GaneshSPatil: can we add some tests for this block?

Issue #4736, opened 2018-05-14
After fetching external artifact, console logs do not show correct color


akshaydewan:
##### Issue Type
- Bug Report

##### Summary
On fetching an external artifact, the console logs continue to show the "In progress" color. It should show the "completed" color after the job is completed.

##### Code snippets/Screenshots
![screen shot 2018-05-14 at 14 08 48](https://user-images.githubusercontent.com/6024038/39986892-adfea726-5780-11e8-93fa-4ee0666d2b94.png)

varshavaradarajan:
@akshaydewan - it should also not show 2 expandable sections for the task and the progress of the pull. The section beginning with `Pulling docker image...` should be part of the task.

arvindsv:
@marques-work knows a lot about this. In case you need help.

PR #4738, opened 2018-05-14
Change tag of console log in ArtifactRequestProcessor to Info and Error

akshaydewan:
Issue: (#4736)

The request processor is used by the fetch task as well, so it is not appropriate to use the publish tag here.

Screenshot of a sample fetch artifact:
screen shot 2018-05-14 at 15 53 36

Screenshot of a sample publish artifact:
screen shot 2018-05-14 at 15 53 56

### DIFF
--- a/common/src/main/java/com/thoughtworks/go/remote/work/artifact/ArtifactRequestProcessor.java
+++ b/common/src/main/java/com/thoughtworks/go/remote/work/artifact/ArtifactRequestProcessor.java
@@ -21,12 +21,11 @@ import com.thoughtworks.go.plugin.api.response.DefaultGoApiResponse;
 import com.thoughtworks.go.plugin.api.response.GoApiResponse;
 import com.thoughtworks.go.plugin.infra.GoPluginApiRequestProcessor;
 import com.thoughtworks.go.plugin.infra.plugininfo.GoPluginDescriptor;
+import com.thoughtworks.go.util.command.TaggedStreamConsumer;
 import com.thoughtworks.go.work.GoPublisher;
 
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/thoughtworks/go/remote/work/artifact/ArtifactRequestProcessor.java
```diff
@@ -54,10 +53,10 @@ private GoApiResponse processConsoleLogRequest(GoPluginDescriptor pluginDescript
         final String message = format("[%s] %s", pluginDescriptor.id(), consoleLogMessage.getMessage());
         switch (consoleLogMessage.getLogLevel()) {
             case INFO:
-                goPublisher.taggedConsumeLine(PUBLISH, message);
+                goPublisher.taggedConsumeLine(TaggedStreamConsumer.OUT, message);
```
varshavaradarajan: The ArtifactRequestProcessor can be called from publish artifact as well. Make sure that the console log colour for publishing external artifacts remains consistent with publish of artifacts to gocd server.

akshaydewan: The difference is the two lines in white after `Start to upload pluggable artifact using...`. They were previously in blue. Is that acceptable? If not, we will probably have to change the contract of `ConsoleLogMessage`

varshavaradarajan: I think that having it in white is not consistent with the publish artifact to gocd server.

akshaydewan:
Closing this, we'll change the `ConsoleLogMessage` and open another PR

Issue #4737, opened 2018-05-14
secure variables are not masked in task plugins


adityasood:
##### Issue Type
- Bug Report
##### Summary
Secure variables are not masked in task plugins.
###### Basic environment details

* Go Version: 18.5.0 and 18.4.0
* JAVA Version: Java8
* OS: Mac and Linux

##### Steps to Reproduce

1. Start a GoCD server
2. Configure a pipeline to use a task plugin ( tested with groovy plugin and script executor plugin) and add a secure variable to the job and pipeline
3. Echo or use printenv to print the values of secure variables on the GoCD UI job console
4. The values are not masked

##### Expected Results
The secure variables should be masked as they are masked when using the "More" task command in GoCD 

##### Actual Results
The secure variable values are visible on the job console

##### Possible Fix
mask the secure variable values

<img width="1146" alt="screen shot 2018-05-11 at 11 17 20" src="https://user-images.githubusercontent.com/17826464/39991151-39a8c2ba-578c-11e8-8ad1-72a3e6171813.png">
<img width="1348" alt="screen shot 2018-05-14 at 3 28 07 pm" src="https://user-images.githubusercontent.com/17826464/39991172-4244ace0-578c-11e8-9554-91b29d1077e0.png">

rajiesh:
verified on 18.6.0 (6883-1013a67ccf0cda9c87e6361286708b584be42338)

Issue #4401, opened 2018-02-22
Support agent side plugin for BuildSession workfolw.


bdpiparva:
##### Issue Type
- Feature enhancement

##### Summary

Agent side plugins(Task, SCM...) are not supported when is agent set to use build session.

This should also take care of providing access to console log processor or equivalent for agent side plugins so that the plugins can choose to write to the console.log.

GaneshSPatil:
Build session has been removed as part of [this](https://github.com/gocd/gocd/commit/d45efb4d257b72bd0b801824404e785fc873e4de) commit.
This issue is no longer relevant.
Closing...

PR #4739, opened 2018-05-15
Use the safe output stream consumer to replace secrets in pluggable tasks. (#4737)

varshavaradarajan:
cc @akshaydewan

### DIFF
--- a/commandline/src/main/java/com/thoughtworks/go/util/ProcessWrapper.java
+++ b/commandline/src/main/java/com/thoughtworks/go/util/ProcessWrapper.java
@@ -46,8 +46,8 @@ public class ProcessWrapper {
         this.command = command;
         this.consumer = consumer;
         this.startTime = System.currentTimeMillis();
-        this.processOutputStream = StreamPumper.pump(process.getInputStream(), new SafeOutputConsumer(consumer), "", encoding);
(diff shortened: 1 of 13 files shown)


varshavaradarajan:
Pending: 

- [ ] Websocket enabled agents
- [ ] Agents using BuildSession

arvindsv:
Agents using build session don't support plugins (yet). Right? Or is it just the golang one?

I'd not tie those two to what you already have. Meaning I'd consider those separate PRs. Makes it easier to review and merge, imo.

varshavaradarajan:
> Agents using build session don't support plugins (yet). Right? Or is it just the golang one?

No idea. We saw something called `ExportCommandFilter` that does the masking. Need to try that out with pluggable task. You just made me realize that if that's the case, even publish and fetch of external artifacts must be supported by Build Session :)

> I'd not tie those two to what you already have. 

Okay.

varshavaradarajan:
Tried it out with BuildSession - Got the error `"com.thoughtworks.go.domain.builder.pluggableTask.PluggableTaskBuilder" does not support new build command agent` - So, pluggable tasks is not supported by BuildSession. @jyotisingh - checked with websockets. Works.

varshavaradarajan:
https://github.com/gocd/gocd/issues/4401

varshavaradarajan:
https://build.gocd.org/go/pipelines/value_stream_map/trigger/2776

PR #4757, opened 2018-05-18
#000 - Fix Mac installer to use Java 1.8 only

arvindsv:
Currently, it's looking for 1.7+. That'll make it find Java 9 etc.


arvindsv:
Merging. It's a small change and I've tried it out. /cc @rajiesh

PR #4760, opened 2018-05-21
Support for external artifact config in pipeline config API.

varshavaradarajan:


varshavaradarajan:
To test -

- [ ] Fetch artifact - within the same pipeline and if the pipeline config being saved has a dependency on an upstream pipeline. 

- [ ] pipeline config update API:  What happens if the pipeline already exists in the config? For a request like 

```json
{
                    "label_template": "${COUNT}",
                    "lock_behavior": "lockOnFailure",
                    "name": "new_pipeline",
                    "materials": [
                      {
                        "type": "git",
                        "attributes": {
                          "url": "git@github.com:sample_repo/example.git",
                          "destination": "dest"
                        }
                      }
                    ],
                    "stages": [
                      {
                        "name": "defaultStage",
                        "jobs": [
                          {
                            "name": "defaultJob",
                            "tasks": [
                              {
                                "type": "exec",
                                "attributes": {
                                  "run_if": [
                                    "passed"
                                  ],
                                  "command": "ls",
                                  "working_directory": null
                                }
                              }
                            ],
                            "artifacts": [
                              {
                                "type": "build",
                                "source": "foo",
                                "destination": "bar"
                              },
                              {
                                "type": "test",
                                "source": "fml",
                                "destination": "fml"
                              },
                              {
                                "type": "external",
                                "artifact_id": "gross",
                                "store_id": "hub",
                                "configuration": [
                                  {
                                    "key": "Image",
                                    "value": "we"
                                  }
                                ]
                              }
                            ]
                          }
                        ]
                      },
                      {
                        "name": "defaultStage2",
                        "jobs": [
                          {
                            "name": "defaultJob",
                            "tasks": [
                              {
                                "type": "exec",
                                "attributes": {
                                  "run_if": [
                                    "passed"
                                  ],
                                  "command": "ls",
                                  "working_directory": null
                                }
                              },
                              {
                                "type": "fetch",
                                "attributes": {
                                  "origin": "gocd",
                                  "stage": "defaultStage",
                                  "job": "defaultJob",
                                  "source": "dir"
                                }
                              },
                              {
                                "type": "fetch",
                                "attributes": {
                                  "origin": "external",
                                  "pipeline": "new_pipeline",
                                  "stage": "defaultStage",
                                  "job": "defaultJob",
                                  "artifact_id": "gross",
                                  "configuration": [
                                    {
                                      "key": "foo",
                                      "value": "plain"
                                    }
                                  ]
                                }
                              }
                            ]
                          }
                        ]
                      }
                    ]
                }
```

- [ ] Publish artifact config - If the store id is a parameter.

- [ ] Fetch artifact config - If either the pipeline, stage, job or artifact id is a parameter.

@akshaydewan - I guess the above applies to config repos as well. Although I don't think anything would have to change as the `@PostConstruct` is in place. Please verify.

varshavaradarajan:
What we have finally decided on encryption:
 
- Do the preprocessing twice - once at the time of encrypting secure properties (preprocessing of only params) and once when we call `preprocessAndValidate` or `preprocess` (in case of entities). This might incur a perf overhead in some cases. We need to check the config save performance. Possible checks to do before doing the param preprocessing is whether the pipeline config even has publish and fetch external artifact config

varshavaradarajan:
Questions for @arvindsv:

**Context**: In this PR, we handle encryption of secure publish and fetch plugin properties for *pipelines*. We have various config save flows that we should take care of:
- FullConfigSaveNormalFlow
- FullConfigSaveMergeFlow
- Old config save - normal flow (See GoFileConfigDataSource#writeWithLock)
- Old config save - (See GoFileConfigDataSource#writeWithLock)
- APIs

For the various config save flows that we have, should we even preprocess and attempt to encrypt plugin properties at the template level given that at the template level given that the store id or artifact id may be a param and so we might not be able to encrypt the properties? Asking this so that we don't perform expensive param preprocessing (which won't be able to replace param values for anything in a template) if we can't encrypt. We could also try to encrypt the plugin properties as is without preprocessing, but we might end up running into #4863.

For validation of a template publish or fetch external artifact config, the validation of plugin properties should happen in the context of the associated pipelines? So, for create template, we won't be requesting any plugin to perform validations, but for update of template, we have to check the associated pipelines and request the respective plugin to perform the validation, right? 

@jyotisingh, @maheshp - Have I missed anything?

PR #4763, opened 2018-05-21
Ensure that the external artifact configuration object encrypts values for secure properties.

varshavaradarajan:
* When editing through the XML, if the plugin is present, secure properties need to be encrypted.

cc @akshaydewan


config/config-api/src/main/java/com/thoughtworks/go/config/PluggableArtifactConfig.java
```diff
@@ -196,4 +213,21 @@ public String toString() {
                 ", storeId='" + storeId + '\'' +
                 '}';
     }
+
+    private boolean isSecure(String configKeyName) {
```
bdpiparva: When the plugin is not loaded, the artifact store config will be exposed as a plain-text value including secure values as well. Is it okay to send 404 when the plugin is not loaded?

This is also true for other entities as well(elastic profile, auth_configs etc...)

varshavaradarajan: Yeah. Its left as is for a reason. Since directly editing the XML is what results in this, we can't really render the config as invalid or say 404. Because this causes problems while starting the server. If the config is rendered invalid, then the previous config is picked up from config.git and this continues until a valid config xml is found. This will result in loss of user's config.

PR #4769, opened 2018-05-22
Disable agent level analytics icons based on go.enable.analytics.only.for.admins for non admin users

GaneshSPatil:


GaneshSPatil:
PR build: https://build.gocd.org/go/pipelines/value_stream_map/trigger/2824

PR #4770, opened 2018-05-22
Remove webpack-rails gem

ketan:
Use the java `WebpackAssetsService` instead.

The rails <-> webpack integration provided by gems is not suitable for
our needs and needs quite some hacky workarounds. So it's best to drop
this integration before attempting a rails upgrade.


ketan:
The only relevant change is here https://github.com/gocd/gocd/pull/4770/files#diff-ece1b629796718ab9f569c1d688c0a49R17.

PR #4771, opened 2018-05-22
Migrate analytics SPA controller to Spark

marques-work:


spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/AnalyticsDelegate.java
```diff
@@ -0,0 +1,145 @@
...
+            return GSON.toJson(analytics.toMap());
+        } catch (Exception e) {
+            LOG.error("Encountered error while fetching analytics", e);
+            throw halt(500, format("Error generating analytics from plugin - %s", request.params(":plugin_id")));
+        }
+    }
+
+    private void checkPipelineExists(Request request, Response response) {
+        if (isPipelineRequest(request)) {
+            if (null == pipelineConfigService.getPipelineConfig(request.queryParams("pipeline_name"))) {
```
ketan: Use `GoConfigService#pipelineConfigByName` — it's optimized for faster lookups, instead of iterating through all pipelines in the config.

marques-work:
@ketan is this mergeable now?

Issue #4049, opened 2017-12-04
Artifact extension to upload and fetch artifact on external repository


bdpiparva:
##### Issue Type
- Feature proposal

##### Summary

- [x] Configuration to support pluggable artifact(https://github.com/gocd/gocd/issues/4050)
- [x] Publish a artifact to an external store(https://github.com/gocd/gocd/issues/4051)
- [x] Fetch artifact using artifact plugin (https://github.com/gocd/gocd/issues/4053)
- [x] Config repo support for publish and fetch pluggable artifacts
- [x] Add a feature toggle for config views
- [x] Artifact plugins capabilities(https://github.com/gocd/gocd/issues/4052)
- [x] Do not purge content of pluggable artifact metadata from GoCD (https://github.com/gocd/gocd/pull/4218) 
- [x] Show the progress of fetch or publish artifact on the console(https://github.com/gocd/gocd/pull/4208)
- [x] Have same tag to represent pluggable artifact and builtin artifact(https://github.com/gocd/gocd/pull/4278)
- [x] XSL transform to convert `<pluggableArtifact>` to `<artifact type="external">` (https://github.com/gocd/gocd/issues/4604) PR - https://github.com/gocd/gocd/pull/4605
- [x] Have same xml tag to represent different types of fetch tasks. (https://github.com/gocd/gocd/issues/4577) 
- [x] Have configuration subtag on external artifacts (https://github.com/gocd/gocd/issues/4627)
- [ ] ~~Add pluginId attribute to pluggableArtifact and fetchPluggableArtifact elements (https://github.com/gocd/gocd/issues/4206)~~
- [x] Artifact Store Config API(https://github.com/gocd/gocd/pull/4195)
- [x] Artifact Store SPA (https://github.com/gocd/gocd/issues/4580) PR (https://github.com/gocd/gocd/pull/4628)
- [x] Console log rendering bug for fetch artifact  (https://github.com/gocd/gocd/issues/4736)
- [x] Ensure encrypted values for publish external artifact config (#4760)
- [x] Ensure encrypted values for fetch external artifact config (#4760)
- [x] Pipeline Config API support for publish and fetch pluggable artifacts
   - [x] Move Pipeline config API to spark
   - [x] Add support for publish and fetch config of artifact type external (#4760)
- [x] Template Config API support for publish and fetch pluggable artifacts (#4804)
   - [x] Move Template config API to spark
   - [x] Add support for publish and fetch config of artifact type external
[...]

PR #4779, opened 2018-05-24
Bump up version of Config Repo target_version and add migration

akshaydewan:
cc: @arvindsv


arvindsv:
@akshaydewan Looks good to me. But, shouldn't the v3_comprehensive.json file have external artifacts? Or, is that going to be changed as a part of a different PR?

Context for whoever is seeing this: This is preparation for support for [external artifacts](https://github.com/gocd/gocd/issues/4049) in config repos.

akshaydewan:
@arvindsv Since the test is migrating from v2 to v3, I can't add external artifacts there. Instead I've added another test for v3 and that one contains external artifacts.

arvindsv:
@akshaydewan Ok.

About [this](https://github.com/gocd/gocd/pull/4779/commits/a3a08e69468ef616cceec846c3aa07f1de2bba11#diff-56c2043ac5ed0f65b790cb794477dcd6R48), since it's JSON, we should be able to use something like:

```json
"configuration": {
  "Image": "gocddemo/guestbook",
  "Tag": "v5"
}
```

Right?

arvindsv:
It might not matter much, since this is the API between the plugin and the GoCD server. Not the interface between the user (`*.gopipeline.json`) and the plugin. However, unless we care about the order, we don't need it to be an array.

akshaydewan:
Currently the `ConfigConverter` uses the same format for Task plugins as well. Environment variables and Parameters also have a similar structure. If we do change it, we should try to make all of them consistent.

Btw, the JSON config repo plugin also uses a similar structure, but the YAML plugin is nicer. We can think of improving the JSON plugin (which is user facing) if not the contract between the server and plugin.

arvindsv:
Yes. On second thoughts, I think changing the user-facing side is better use of time anyway. So, I am alright with this being this way (between plugin and server).

/cc @tomzo - FYI

tomzo:
> Currently the ConfigConverter uses the same format for Task plugins as well. Environment variables and Parameters also have a similar structure. If we do change it, we should try to make all of them consistent.

@akshaydewan That's right. I like that you made the new one similar. I also think that this format is  convenient for the contracts and I would not change it.
And yes if we want JSON to be more concise, then the change should happen in plugin.

But I don't see any changes for the `ConfigConverter`, there should be some code creating new artifact config classes, is it still WIP or in another PR?

varshavaradarajan:
@tomzo - https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java#L174 and https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java#L521 

Did you mean any other change?

tomzo:
@varshavaradarajan thanks, I was looking for those. This looks OK to me now.

varshavaradarajan:
@tomzo - given that we are differentiating between fetch from gocd and external by using `artifact_origin` (See #4888), I added a migration to add `"artifact_origin":"gocd"` on all fetch tasks if the `target_version` is 2. That okay? Can you please take a look?

varshavaradarajan:
Also, @arvindsv and @tomzo - Should we retain all the migrations? I understand that people would have checked in jsons and yamls with target_version 1 and we don't want that to break. But for how long should we maintain those? Or is this something similar to API versioning? Should we throw an error if the target version is less than say, a minimum supported version?

tomzo:
> I added a migration to add "artifact_origin":"gocd" on all fetch tasks if the target_version is 2. That okay? Can you please take a look?

Yes. 
@varshavaradarajan The code looks fine, minor things I would improve:
 - It would be nice to add a test which migrates from v1 up to v3 to checks that when all steps are executed the end result is as expected.
 - I would rename those older tests a bit `shouldMigrateToV2` to `shouldMigrateV1ToV2`

And  Test `shouldDoNothingIfMigratingFromV3ToV3` is failing though.

>  Should we retain all the migrations?

IMO we keep all migrations forever. It is an equivalent of the xsd migrations for cruise-config.xml. I don't see a reason to remove them.

> Should we throw an error if the target version is less than say, a minimum supported version?

We might want to set a warning in health messages, not an error. The reason would be to force users to update to newer format to reduce the overhead of migrations on server. Otherwise user has no reason to do so until new feature is needed.

varshavaradarajan:
@tomzo - thanks, had to change the migration. Can you take a look again, please?

Regarding testing v1 to v3, there is a loop in JsonMessageHandler1_0 that calls individual migrations - that is, v1 to v2 and then v2 to v3. That's why I didn't think it was necessary to test v1 to v3 as long as we test that the migrator is called that many times. Let me know if you think otherwise. If so, I'll have to write a test in JsonMessageHandler1_0Test asserting on the json.

PR #4783, opened 2018-05-25
Introduced v4 of notification extension

maheshp:
New version introduced to support enhanced StageStatusNotification message
* StageStatusNotification message has material fingerprint.
* StageStatusNotification message has details of previous stage run if
  one exists.

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/domain/Stage.java
+++ b/domain/src/main/java/com/thoughtworks/go/domain/Stage.java
@@ -53,6 +53,7 @@ public class Stage extends PersistentObject {
     private static final StageResult DEFAULT_RESULT = StageResult.Unknown;
     private static final Cloner CLONER = new Cloner();
     private String configVersion = null;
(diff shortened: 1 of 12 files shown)


plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/notification/v3/JsonMessageHandler3_0.java
```diff
@@ -24,8 +24,9 @@
 import com.thoughtworks.go.plugin.api.response.Result;
 import org.apache.commons.lang3.NotImplementedException;
 
-import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
```
GaneshSPatil: Optimizing import in `v3/JsonMessageHandler3_0.java` as nothing has changed in this file. is it ok to undo these import changes?

PR #4803, opened 2018-06-01
GoCipher will now prefer to use AES for encryption and decryption

ketan:
Write a config migration to migrate from DES cipher-text to AES cipher-text

There is a fallback to use DES for decryption only if there is a cipher
file containing the DES cipher AND DES has been enabled via a system
property. DES is enabled by default (for now).

For new GoCD server instances, since the DES cipher file will be missing,
the GoCipher instance will no longer support DES.


ketan:
For context, here's what the implementation does:

### New installation

- There will be no way to generate or encrypt strings using DES (since there was no cipher to begin with)

### Upgrades on existing installations

- A one-time config migration that will re-encrypt *all* encrypted text in the config using AES
- There will be backward compatibility to be able to decrypt any encrypted text in config repositories that was encrypted using DES
- There will be backward compatibility if an administrator pastes any encrypted text in the config XML that was encrypted using DES, the value will simply re-encrypted using AES before being saved

### How the new AES strings will be generated

- A `cipher.aes` file will be created, it will contain a (hex encoded) 128-bit secret key used for encrypting all strings
- Every time an encryption is peformed, a random 128-bit IV (Initialization Vector) will be generated, and the plaintext encrypted to AES. The output of the AES function will look like `AES:INITIALIZATION_VECTOR_AS_BASE_64:ENCRYPTED_TEXT_AS_BASE64`. This scheme us to (a) identify and future-proof the encryption used `AES` (b) decrypt the `ENCRYPTED_TEXT_AS_BASE64` using the encoded IV in the text and the secret key in the `cipher.aes` file

PR #4804, opened 2018-06-01
Migrate template config api v4 to spark

varshavaradarajan:
* Add support for publish and fetch of external artifacts. 

This PR on top of an existing PR - #4760. To be merged only after #4760 

@ketan - the authorization at the controller level is slightly different here.  If you see the `api_v4/templates_controller.rb`, we need to authorize template view users for index and show, template admins for update and destroy and pipeline group admins for create. At the time of `setUpRoutes`, we don't seem to check for the request method before using the `apiAuthenticationHelper`. How should I proceed? 

Also, since template v4 is unannounced, I didn't create a new API version in spark. So, I have to add a toggle and route to either rails or spark based on that, right?  I suppose I need to check for the api version v4 and the toggle as well because there is a API v1 of template authorization API that I haven't migrated to spark as it was out of scope for this.


varshavaradarajan:
PR build passed - https://build.gocd.org/go/pipelines/value_stream_map/trigger/3021 and @ketan approved :)

Issue #4776, opened 2018-05-24
Fan in/out appears to stop once new commit lands


plainsane:
##### Issue Type
- Bug Report
- Documentation Report
##### Summary
When building a pipeline graph by specifying upstream pipelines as materials, works well until a new commit lands on the repo.  Once this happens, the older invactions of the pipeline no longer trigger their down streams.
##### Environment
aws ec2 instance
###### Basic environment details
amazon linux 1 ami
openjdk 8
script-executor plugin
yaml plugin
##### Steps to Reproduce
fork git@github.com:plainsane/gogogadgetcd.git
add as config repo to server xml.
wait for pipeline A to reach its first manual approval stage.
make new commit to repo and push.
wait for pipeline A to reach its first manual approval stage.
select history on pipeline A
allow instance 1 of pipeline A to continue to completion. 
##### Expected Results
i would expect pipeline B and C to trigger for the same sha as pipeline A's shaw.
##### Actual Results
nothing happens at all.  if you allow the latest instance of pipeline A to complete, things behave as expected.

##### Any other info
i have a pipeline graph of 13 pipelines that im using to control the flow of code/terraform from dev->qa->production.  i currenlty have to lock out all developers from landing code on my master branch to take code to qa, then to production (which can take half a day sometimes).  i would really like to allow things to progress from a specific sha while code is landed on master in a fail forward fashion but i currently can not do that as pipelines fail to trigger downstream once a new commit lands on master.

varshavaradarajan:
Could reproduce this with above config - The go server logs show 

```
BuildCauseProducerService:175 - Error while scheduling pipeline: CR3. Possible Reasons: (1) Upstream pipelines have not been built yet. (2) Materials do not match between configuration and build-cause.
2018-05-24 13:39:12,981 WARN  [91@MessageListener for ScheduleCheckListener] BuildCauseProducerService:175 - Error while scheduling pipeline: CR2. Possible Reasons: (1) Upstream pipelines have not been built yet. (2) Materials do not match between configuration and build-cause.
```

Need to debug further.

plainsane:
so the fact that pipeline A was built with a specific sha that has been superseded by a new sha in the repository, i can no longer progress pipeline B or C for the sha that pipeline A was built against?

i was under the impression that pipeline B and C would be schedule for the sha that pipeline A completed for stage apply.  i didnt see anytrhing in the documentation that described this scenario.

varshavaradarajan:
[...]

PR #4811, opened 2018-06-05
WIP: fix for #4776

ibnc:
Possible fix for #4776 

I'm not 100% satisfied with my solution, but I don't quite see another alternative given the information available when deciding whether or not the build should run. 

I'd really appreciate any feedback: @ketan @arvindsv @jyotisingh @varshavaradarajan

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java
@@ -367,6 +367,13 @@ public class PipelineConfig extends BaseCollection<StageConfig> implements Param
         return this.origin instanceof RepoConfigOrigin;
     }
 
+    public boolean hasSameConfigOrigin(PipelineConfig other) {
+        if (!(isConfigDefinedRemotely()))
+            return false;
+
+        return this.origin.equals(other.getOrigin());
+    }
+
     public boolean isConfigOriginFromRevision(String revision) {
         if (!(isConfigDefinedRemotely()))
             return false;
--- a/domain/src/main/java/com/thoughtworks/go/domain/MaterialRevisions.java
+++ b/domain/src/main/java/com/thoughtworks/go/domain/MaterialRevisions.java
@@ -289,6 +289,17 @@ public class MaterialRevisions implements Serializable, Iterable<MaterialRevisio
         return revision.getRevision();
     }
 
+    public List<DependencyMaterial> getDependencyMaterials() {
+        List<DependencyMaterial> mats = new ArrayList<DependencyMaterial>();
+        for(MaterialRevision materialRevision : this) {
+            Material material = materialRevision.getMaterial();
+            if (material instanceof DependencyMaterial) {
+                mats.add((DependencyMaterial) material);
+            }
+        }
+        return mats;
+    }
+
     public DependencyMaterialRevision findDependencyMaterialRevision(String pipelineName) {
         for (MaterialRevision materialRevision : this) {
             Revision revision = materialRevision.getRevision();
--- a/domain/src/main/java/com/thoughtworks/go/domain/buildcause/BuildCause.java
+++ b/domain/src/main/java/com/thoughtworks/go/domain/buildcause/BuildCause.java
@@ -20,6 +20,7 @@ import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.materials.MaterialConfigs;
 import com.thoughtworks.go.config.materials.Materials;
+import com.thoughtworks.go.config.materials.dependency.DependencyMaterial;
 import com.thoughtworks.go.config.remote.RepoConfigOrigin;
 import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.domain.materials.Material;
@@ -116,6 +117,15 @@ public class BuildCause implements Serializable {
         }
         return true;
     }
+
+    public boolean hasDependencyMaterials() {
+        return this.getMaterialRevisions().hasDependencyMaterials();
+    }
+
+    public List<DependencyMaterial> getDependencyMaterials() {
+       return this.getMaterialRevisions().getDependencyMaterials();
+    }
+
     public boolean pipelineConfigAndMaterialRevisionMatch(PipelineConfig pipelineConfig){
         if(!pipelineConfig.isConfigOriginSameAsOneOfMaterials())
         {
@@ -151,7 +161,7 @@ public class BuildCause implements Serializable {
             return;
         }
         // then config and code revision must both match
-        if(this.trigger.isForced())
+        if(this.trigger.isForced() || this.hasDependencyMaterials())
         {
             // we should not check when manual trigger because of re-runs
             // and possibility to specify revisions to run with
--- a/server/src/main/java/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java
@@ -20,6 +20,7 @@ import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.materials.MaterialConfigs;
 import com.thoughtworks.go.config.materials.Materials;
+import com.thoughtworks.go.config.materials.dependency.DependencyMaterial;
 import com.thoughtworks.go.config.remote.ConfigOrigin;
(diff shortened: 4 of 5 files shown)


tomzo:
@ibnc please elaborate what new behavior did you intend to make with this change.

In #4776 you correctly identified the root cause of this limitation:
> The actual bug here is that non configuration changing commits are updating the configuration's revision when there is no need to do so.

The affected use case is when you have at least 2 pipelines, `A -> B` which depend on each other and both are defined in config-repo, which also happens to be a material in both pipelines.
The intentional changes to prevent pipeline from scheduling were added so that pipeline always schedules only with material matching revision of pipeline config.
Current solution works well when there is no pipeline B, because once server sees new material it schedules the pipeline which will till the end. 
When there is pipeline B, someone, *might* have pushed new commit which *may* contain changes of pipeline config or not. If it does contain config changes then it might have already changed pipeline B and newly scheduled pipeline will run with *old material* and new configuration. That can be bad in some scenarios.
So far GoCD server could not determine if changes in config-repo actually introduced changes in configuration, so we used safest approach - compare revision of source.
I think you got all above, but we need to be on the same page.

You decided to use `equalsIgnoringOrigin` to check if config is equal although revision have changed. I think that's the right approach.
But I have a couple of issues on the PR:
 * [This](https://github.com/gocd/gocd/pull/4811/files#diff-840c9113cfcc087f67174f9c169d64eaR132) looks bad, why do you need to set revision in `pipelineConfigAndMaterialRevisionMatch`? It is a method which should not modify the arguments, (given its name).
 * [This is bothering much](https://github.com/gocd/gocd/pull/4811/files#diff-753e1935d392c52926a40766599ac870R147). Rather than using date and this early `return` perhaps you could track a new field in origin, something like `lastConfigChangingRevision` and use in the `pipelineConfigAndMaterialRevisionMatch` to determine if it is OK to schedule a pipeline. My point is that if `equalsIgnoringOrigin` cannot detect some changes, config will be stale. I don't think `equalsIgnoringOrigin` was used anywhere outside of tests so far.

ibnc:
@tomzo 

> This looks [bad](https://github.com/gocd/gocd/pull/4811/files#diff-840c9113cfcc087f67174f9c169d64eaR132), why do you need to set revision in pipelineConfigAndMaterialRevisionMatch? It is a method which should not modify the arguments, (given its name).

That actually shouldn't have made it into the commit. It's unneeded, and weird as you said.

> [This is bothering much](https://github.com/gocd/gocd/pull/4811/files#diff-753e1935d392c52926a40766599ac870R147). Rather than using date and this early return perhaps you could track a new field in origin, something like lastConfigChangingRevision and use in the pipelineConfigAndMaterialRevisionMatch to determine if it is OK to schedule a pipeline. My point is that if equalsIgnoringOrigin cannot detect some changes, config will be stale. I don't think equalsIgnoringOrigin was used anywhere outside of tests so far.

I actually explored that path. The problem I ran into was that there is insufficient data to use the last config changing revision effectively. Here's what I mean:

Let's say we have 3 commits:
* sha: 3333 (doesn't change config) 
* sha: 2222 (doesn't change config)
* sha: 1111 (changes config)

If the dependent pipeline was triggered by the upstream pipeline using commit 2222, then it won't work. lastConfigChangingRevision will be 1111, and the origin's revision will be 3333, but the material's revision will actually be 2222. So the pipeline will not be scheduled even though the configuration hasn't changed.

The only other thing we have is the [DependencyMaterialRevision](https://github.com/gocd/gocd/blob/b44109e562b7f42fd86c0d4e27727eaf62ecd770/domain/src/main/java/com/thoughtworks/go/domain/materials/dependency/DependencyMaterialRevision.java#L29) which is essentially useless for our purposes.


With all of that said, It's possible I missed something. I'm currently looking into if I can get the revisions since the config changed to see if the current revision is in that list

arvindsv:
@ibnc Searching for "When configuration repository that defines the pipeline is the same as one of materials" in #1133 might be useful too.

tomzo:
> I guess I'm just curious why we need this specific check. Like, what are the cases where we need the check?

We absolutely need the check for the cases when all this is true:
 * pipeline is defined in config-repo
 * config-repo is one of the materials in pipeline

Because the system which is creating build causes, (which further leads to scheduling) is pretty much independent from material poling and config-repo parsing. Without the extra checks in `pipelineConfigAndMaterialRevisionMatch` that could lead to situation when user pushed some commit `123`, which had both pipeline configuration and other changes and those might not run together (it would be previous config + new material, or new config + old material). But he would naturally expect those to be scheduled together.
There is much more about discussion on this in #1133. You can also look at the PR which specifically addressed these cases https://github.com/gocd/gocd/pull/1536
Although I am aware of the problems you are facing when there are more dependent pipelines in same configuration repo. As I mentioned [earlier](https://github.com/gocd/gocd/pull/4811#issuecomment-395003906) I think this is a limitation, caused by gocd using only revision to compare if config has changed.

> it's possible to run revisions that happened before a configuration change happened. It's possible to rerun and trigger out of date stages because they're manual triggers.

Yes, that was intentionally made that way, so that you could "override" the automatic behavior.

If you are thinking about disabling the check for the affected cases with downstream pipelines, that can be done 2 ways:
 * move downstream pipelines to other repository
 * you can implement a configuration switch which will just disable this. There are switches to turn off fan-in feature or tune it, so I don't see a reason why this shouldn't be configurable if necessary for the user.

But please keep in mind, the check is necessary for the *"first pipeline", which has no upstream pipelines defined in same repository*.

ibnc:
@tomzo 

gotcha. Thank you for the explanation and links. 

It sounds like it would be reasonable to skip this check if the pipeline has an upstream repository?

ibnc:
er, upstream pipeline

tomzo:
For the record, we spoke on gitter and agreed on following:

In the case of 2 pipelines `A -> B` it is safe to allow pipeline B to run when both are defined in same config repository, regardless of revision of material in build cause of pipe B.

ibnc:
@tomzo 

Thoughts on my most recent commit?

config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java
```diff
@@ -367,6 +367,15 @@ public boolean isConfigDefinedRemotely() {
         return this.origin instanceof RepoConfigOrigin;
     }
 
+    public boolean hasSameConfigOrigin(PipelineConfig other) {
+        if (!(isConfigDefinedRemotely()))
+            return false;
+
+        RepoConfigOrigin otherOrigin = (RepoConfigOrigin) other.getOrigin();
+        RepoConfigOrigin repoConfigOrigin = (RepoConfigOrigin) this.origin;
+        return repoConfigOrigin == otherOrigin;
```
tomzo: `==` would check reference equality. You should call `equals`

tomzo:
It looks OK, except the equality check which I pointed.
I mean, I think this would work, but can you try to write integration tests or at least unit tests, which are failing before this change and passing when it's applied?

config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java
ibnc: @tomzo You're totally right. I'll push that up real quick

ibnc:
@tomzo That's what my plan was. I wanted to get your feedback on my work before investing the time

config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java
```diff
@@ -367,6 +367,15 @@ public boolean isConfigDefinedRemotely() {
         return this.origin instanceof RepoConfigOrigin;
     }
 
+    public boolean hasSameConfigOrigin(PipelineConfig other) {
+        if (!(isConfigDefinedRemotely()))
+            return false;
+
+        RepoConfigOrigin otherOrigin = (RepoConfigOrigin) other.getOrigin();
+        RepoConfigOrigin repoConfigOrigin = (RepoConfigOrigin) this.origin;
```
tomzo: The cast is unnecessary and may fail when origin on `other` is not remote.
All implementations of ConfigOrigin can handle equality. You can just use
```
this.origin.equals(otherOrigin)
```

Also, name of method is confusing. `hasSameConfigOrigin` suggests that origin is of any type. But you actually meant `hasSameRemoteConfigOrigin`

ibnc:
@tomzo I pushed up an integration test, and everything is passing on build.gocd.org.

This is good to merge if everything looks good to you

tomzo:
Looks fine to me, but please squash the commits into 1 with some meaningful message describing changes.

ibnc:
@tomzo done.

@arvindsv @ketan could one of you merge this?

PR #4820, opened 2018-06-06
Add ability to group pipelines by environment on dashboard

marques-work:
Must add `?ui=test` to the URL to reveal this feature.

@akshaydewan @bdpiparva hey there, @arvindsv asked me to inform you of this work. Maybe we can chat about it for context/review?

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
@@ -24,6 +24,7 @@ const PipelineWidget      = require('views/dashboard/pipeline_widget');
 const PersonalizeWidget   = require('views/dashboard/personalize_widget');
 const PipelineSelectionVM = require('views/dashboard/models/pipeline_selection_view_model');
 const PipelineSelection   = require('models/dashboard/pipeline_selection');
+const Routes              = require('gen/js-routes');
 
 const DashboardWidget = {
   oninit(vnode) {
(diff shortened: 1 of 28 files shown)


ketan:
You probably left out `GoDashboardActivityListener` — this guy is responsible for invalidating and recomputing the cached dashboard state in case environments change, or pipelines move between environments.

marques-work:
@ketan thanks for the heads up, I'll take a look at `GoDashboardActivityListener` and make the appropriate changes.

arvindsv:
@akshaydewan or @GaneshSPatil Can one of you take a quick look sometime and give @marques-work some feedback on this? There's still a bit of work on this around making it work well with the "Filter View" or "Personalize" option and switching of modes. But, if you can take a cursory look and have feedback, it might be useful.

There is also a question about whether caching of environments is needed, around [here](https://github.com/gocd/gocd/pull/4820/files#diff-4fc85196b93699b13af50b5fd6d8fb90R110). You can think about it later.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
```diff
@@ -92,6 +93,99 @@ const DashboardWidget = {
...
+      groupByEnvironment = <f.checkbox model={vm.groupByEnvironment} attrName="enabled" listener="onclick" label="Group by Environment?" />;
+    }
+
+    function renderEnvironments(pipelinesByEnvironment) {
+      const envs = Object.keys(pipelinesByEnvironment).sort();
+
+      if (envs.indexOf("*") > -1) {
+        envs.splice(envs.indexOf("*"), 1);
+        envs.push("*");
+      }
```
akshaydewan: The "*" stuff seems a bit hacky to me. Maybe it would be more readable if we had a separate list to maintain pipelines without an environment. Thoughts?

akshaydewan:
Could you add some tests in javascript around `getPipelinesByEnvironment`?
Not sure if this would be too much effort, but can we add a toggle in `available.toggles` and `Toggles.java` instead of `?ui=test`? Just for consistency.

ketan:
@marques-work — close this in favor of #3842?

marques-work:
@akshaydewan @ketan Sorry for the late reply, I've missed my GH notifications.

@akshaydewan I can certainly add tests, and will look at the `available.toggles` work.

@ketan not sure, #3842 mockups do include a "group by environment" button of some sort, so at least parts of this could be useful. Unsure if a merge would do more damage than just cherry-picking and adapting. I'll come back to this when I get to the part where we need to get environment grouping working.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
marques-work: It could also arguably be `null` instead of a magic string, which would be more accurate. I probably didn't want to null check while sorting (don't remember why I chose this). I sort of prefer keeping everything in the same collection rather than a separate list because then we can use the same code. I would rather not have to write a separate case to handle the list of ungrouped pipelines.

Changing the ungrouped pipelines to live under a `null` key in `pipelinesByEnvironment` could simplify this to make it more readable as well. I just tested in the console; sorting an array of environment names always puts `null` at the bottom, which is what I was going for anyway. Also, the below lines (e.g. setting `name` and `settingsIcon` can be simplified with something like `name = env || "(no environment)"`.

What do you think?

marques-work:
Closing this PR in favor of #5006 (which is a newer attempt at this same feature)

PR #4826, opened 2018-06-06
Remove plugin info api v3

ketan:


varshavaradarajan:
This is for 18.7 - the july release, right?

ketan:
> This is for 18.7 - the july release, right?

Yes

PR #4828, opened 2018-06-07
Ensure /go/home redirects to dashboard

ketan:


ketan:
@bdpiparva — merging this, but can you take a look to make sure I've not missed something.

Issue #3711, opened 2017-07-28
Copyright footer is misaligned for stage-->job page 


ankitsri11:
##### Summary
Copyright footer is misaligned for job tab available under stage page.

<img width="1440" alt="screen shot 2017-07-27 at 16 44 47" src="https://user-images.githubusercontent.com/6838313/28729149-45b56ce8-7380-11e7-8340-351f3f3a0c55.png">
 

###### Basic environment details

* Go Version: 17.8.0 (5275-95e3f80dc99a41d4a5f512b01842b30915b5eed5)

bdpiparva:
It is reproducible with 18.1.0 as well

rajiesh:
Verified on 18.7.0 (7117-1709149e6946f9cec95e4a8e914a47c7806f0672)

PR #4831, opened 2018-06-07
Fixed positioning of footer on stage settings page

pranavbansod:


bdpiparva:
This is to fix https://github.com/gocd/gocd/issues/3711

PR #4843, opened 2018-06-12
Data sharing APIs

jyotisingh:
Introduce data_sharing/usagedata api V1 (internal) - to get usage data to be shared, available for all authenticated users
Introduce data_sharing/settings api v1 (public)- to get and update if data sharing settings. Update authorization restricted to super-admins alone.
Introduce data_sharing/reporting API v1 (internal)- to get and patch details about when were the usage-data was last shared.


GaneshSPatil:
Merging as the relevant build passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/2943

PR #4847, opened 2018-06-14
Report Usage Data to GoCD Data Sharing Remote Server

GaneshSPatil:
* data_sharing_server_url is required by the browser clients to send the server usage information
* Pass along data_sharing_server_url along with the last_reported_time in data reporting API


base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -181,6 +181,8 @@
     public static GoSystemProperty<String> GO_UPDATE_SERVER_URL = new GoStringSystemProperty("go.update.server.url", "https://update.gocd.org/channels/supported/latest.json");
     public static GoSystemProperty<Boolean> GO_CHECK_UPDATES = new GoBooleanSystemProperty("go.check.updates", true);
 
+    public static GoSystemProperty<String> GO_DATA_SHARING_SERVER_URL = new GoStringSystemProperty("go.datasharing.server.url", "https://datasharing.gocd.org/v1");
```
GaneshSPatil: The Data Sharing Server URL will change.

@arvindsv @jyotisingh -- we can decide later on what should be the data sharing URL.

GaneshSPatil: There is no reason for `GO_DATA_SHARING_SERVER_URL` being a configurable system property.
Can it be a final string defined in `SystemEnvironment`?

If any user defines `go.datasharing.server.url` with a new value, the GoCD data collecting server will stop receiving data even if the data sharing consent is set to true.

@arvindsv @jyotisingh -- thoughts?

server/src/main/java/com/thoughtworks/go/server/service/DataSharingService.java
```diff
@@ -75,9 +77,13 @@ public void initialize() throws DuplicateMetricReporting, DuplicateDataSharingSe
         }
 
         if (existingUsageStatisticsReporting == null) {
-            UsageStatisticsReporting usageStatisticsReporting = new UsageStatisticsReporting(UUID.randomUUID().toString(), new Date());
+            UsageStatisticsReporting usageStatisticsReporting = new UsageStatisticsReporting(UUID.randomUUID().toString(), dataSharingServerUrl, new Date());
             updateUsageStatisticsReporting(usageStatisticsReporting, new HttpLocalizedOperationResult());
+        } else if (!dataSharingServerUrl.equals(existingUsageStatisticsReporting.getDataSharingServerUrl())) {
```
GaneshSPatil: This is required to make sure that if `GO_DATA_SHARING_SERVER_URL` ever changes, it is updated in the DB.

If we can ensure that  `GO_DATA_SHARING_SERVER_URL` is never going to change, we can remove this part of code.

cc: @jyotisingh

server/db/migrate/h2deltas/1807002_usage_data_reporting_table.sql
```diff
@@ -16,9 +16,10 @@
...
 
 CREATE TABLE UsageDataReporting (
-  id             BIGINT GENERATED BY DEFAULT AS IDENTITY ( START WITH 1 ) PRIMARY KEY,
-  serverId       VARCHAR(255) NOT NULL,
-  lastReportedAt TIMESTAMP
+  id                   BIGINT GENERATED BY DEFAULT AS IDENTITY ( START WITH 1 ) PRIMARY KEY,
+  serverId             VARCHAR(255) NOT NULL,
+  dataSharingServerUrl VARCHAR NOT NULL,
+  lastReportedAt       TIMESTAMP
 );
```
jyotisingh: This is not okay. Write a new delta since the original PR that introduced this table is merged already and potentially deployed at a few places.

jyotisingh: Also, I dont think we need to persist the url in db at all

server/src/main/java/com/thoughtworks/go/server/service/DataSharingService.java
jyotisingh: Why store the url in db? What benefit does it bring? Its just a config. If the `GO_DATA_SHARING_SERVER_URL ` is changed to something else, we should start reporting to that instead

base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
jyotisingh: Ofcourse, for now this value for url seems fine. 

I think its fine to leave it as a `SystemEnvironment` property. The only reason for a user to redefine `go.datasharing.server.url` would be to stop sending reports to our server, so in that sense, for us it's as good as them not giving a consent and there's nothing we can do about it. Leaving it as a `SystemEnvironment` gives us some flexibility. At a later point, if we think we do not need that flexibility, we can get rid of the property and make it a constant instead.

server/webapp/WEB-INF/rails.new/webpack/models/shared/usage_data_reporter.js
```diff
@@ -0,0 +1,80 @@
...
+    now.getFullYear() === lastReported.getFullYear());
+};
+
+const reportToGoCDDataSharingServer = function (url, data) {
+  return AjaxHelper.POST({url, data});
+};
+
+const canReportUsageData = () => {
+  const lastReportedTime = localStorage.getItem(USAGE_DATA_LAST_REPORTED_TIME_KEY);
+  return !isUsageDataReportedToday(lastReportedTime);
```
jyotisingh: This wont work! For a given browser, this checks if the last reporting wasn't done today, then it would initiate that. We need this check to be done for a given server once in a day. This would make it happen n times, where n is the number of browsers that accessed Go server in that day.

jyotisingh: Look at the implementation of Version-updater to see how it handles that

server/src/main/java/com/thoughtworks/go/domain/DataSharingSettings.java
```diff
@@ -72,4 +71,10 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(super.hashCode(), allowSharing, updatedBy, updatedOn);
     }
+
+    public void copy(DataSharingSettings from) {
```
jyotisingh: change to `copyFrom` and write a unit test

Issue #3775, opened 2017-08-23
graphical glitch: "pin" shows up incorrectly in big value stream maps


FredrikWendt:
##### Summary

On large value stream maps, the word "pin" appears where it should not. The "pin" text is an offset from pipelines which are positioned very far out to the right.

##### Environment

Appears in the same way on all devices tested (Windows 10, Ubuntu 17.04, Fedora) with Chrome, Chromium and Firefox.

###### Basic environment details

* Go Version: 17.7.0 (5147-53fdb1b15184f93966059a42429bf9ed0bfdee59). 

##### Steps to Reproduce

1. Create a pipeline with any material, such as git
2. Create a series of pipelines, forming a chain of at least 10 downstream pipelines from the one created in step 1.
3. Run the first pipeline.
4. Load the value stream map for the run of the pipeline.

##### Expected Results

It should look great.

##### Actual Results

The word "pin" appears over the first material.

##### Screenshots

Here's an example with 16 levels deep value stream map. As you can see, the "shape" of the right most part of the pipeline mesh, is the same shape that the "pin" labels make on the left side. I'm guessing there's a CSS offset of some magic -2000px, which is not enough now that the value stream map is more than 2000px (2000 is just an approximation).  

![large-vsm-pin-zoom-zero](https://user-images.githubusercontent.com/89504/29633962-5956f23a-8848-11e7-86c9-8849548b5766.png)
![large-value-stream-map-pin-text](https://user-images.githubusercontent.com/89504/29633961-5956b536-8848-11e7-9b99-d3e80b0c3420.png)

FredrikWendt:
Looks like this is a duplicate of #665.

varshavaradarajan:
This can be closed now with the milestone 18.7?

PR #4857, opened 2018-06-18
Fixed the vsm pin issue. The VSM pin is represented by an icon. (#3775)

varshavaradarajan:
There is no need for the button text.
* The css property text_indent is provided on unnecessary texts like stage name, material icon. This commit removes that.


varshavaradarajan:
Regression passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/2968

PR #4864, opened 2018-06-19
WIP (don't merge) - named personalization views

marques-work:
@ketan this is what I have so far: migration to add columns with opportunistic data migration (requires JSON serialization)

`PipelineSelections` now holds a `Filters` object, which is just a fancy holder for a list of `DashboardFilter` objects that either act as blacklist or whitelist filters.

The `Filters` object serializes to JSON and back.

`DashboardFilter` implements a `isPipelineVisible(pipelineName)` method to determine whether or not a given pipeline be shown. The idea here is that `PipelineSelections#includesPipeline()` will delegate to this method on the currently selected `DashboardFilter` (which is the analog to a named view tab in the UI).

Likely, the `DashboardControllerDelegate` will have to accept a parameter indicating the name of the view filter, and pass that along to the `PipelineSelections` object in order to return the filtered set of pipelines.

Currently, the filters JSON structure looks something like this:

```json
{
  filters: [
    {name: "my filter", pipelines: ["p1", "p2", ...], type: "blacklist" },
    ...
  ]
}
```

Anyhow, lmk what you think -- I'm still open to trying a different approach (e.g. a more conventional 1 row per named view like we discussed on slack).


ibnc:
@marques-work the mock ups show the default filter not having the down-karet. So I don't think we need to do the check during the tab deletion.... We also need to add the removal of the down-karet from the default view tab

marques-work:
@ibnc you're right, I don't see the downward chevron for the default.

@naveenbhaskar does the lack of the chevron mean that one cannot configure the default view at all? In other words, is the default view always a wildcard filter (i.e. always shows every pipeline)? If so, wouldn't a user want to be able to configure the default instead of having to click a tab each time they visit the page?

marques-work:
@ibnc cleared things up with @naveenbhaskar. The default tab should have a downward chevron and be editable, reorderable, but not deletable.

api/api-dashboard-v2/src/main/java/com/thoughtworks/go/apiv2/dashboard/representers/PipelineGroupsRepresenter.java
```diff
@@ -25,6 +25,7 @@ public static void toJSON(OutputWriter jsonOutputWriter, DashboardFor dashboardF
         jsonOutputWriter
             .addLinks(linksWriter -> linksWriter.addLink("self", Routes.Dashboard.SELF)
                 .addAbsoluteLink("doc", Routes.Dashboard.DOC))
+            .add("personalization", dashboardFor.getPersonalizationEtag())
```
GaneshSPatil: @marques-work @arvindsv @ketan -- **FYI**. We are modifying the Dashboard API v2 JSON response. Ideally, we should have created a new API version for the dashboard API. But as we haven't announced v2 so it should be OK to modify it.

GaneshSPatil:
@ibnc @marques-work @naveenbhaskar @maheshp  -- Ideally, the selected tab on the dashboard should always be persisted across page navigations.

Lets say a user is on the third tab, and (s)he navigates to job-details-page, going back to the dashboard by clicking on the `Pipelines` tab (from header) should take the user to the third tab on dashboard.

**Scnearios**:
1. If the user opens the dashboard for the first time, default tab should be shown.
2. If the user navigates to other GoCD pages from dashboard and comes back to the dashboard again, the last visited tab should be shown.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/pipeline_list_vm.js
```diff
@@ -0,0 +1,76 @@
+/*
```
GaneshSPatil: can we define the view model under `views/dashboard/models`?
Similar to [trigger with options VM](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/trigger_with_options_vm.js)?

marques-work: Isn’t that where this file is already?

`views/dashboard/models/pipeline_list_vm.js`

api/api-pipeline-selection-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineselection/PipelineSelectionControllerDelegate.java
```diff
@@ -85,13 +88,25 @@ public String show(Request request, Response response) {
...
         return PipelineSelectionsRepresenter.toJSON(pipelineSelectionResponse);
     }
 
     public String update(Request request, Response response) {
         String fromCookie = request.cookie(COOKIE_NAME);
+        final Long userId = currentUserId(request);
+
+        if (!Objects.equals(pipelineSelectionsService.load(fromCookie, userId).etag(), getIfMatch(request))) {
+            response.status(PRECONDITION_FAILED);
+            return "Update failed because client filter data is out-of-date";
```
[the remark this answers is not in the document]
marques-work: @GaneshSPatil sure, I’ll take a look. Didn’t know we had that :)

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/pipeline_list_vm.js
GaneshSPatil: Whoops! Yes you are right :)

server/src/main/java/com/thoughtworks/go/server/domain/user/WhitelistFilter.java
```diff
@@ -41,8 +44,18 @@ public String name() {
     }
 
     @Override
-    public boolean isPipelineVisible(CaseInsensitiveString pipeline) {
-        return null != pipelines && !pipelines.isEmpty() && pipelines.contains(pipeline);
+    public boolean isPipelineVisible(CaseInsensitiveString pipeline, StageInstanceModel latestStage) {
```
marques-work: @ibnc I think another valid option would be filtering on the client side, and it might be simpler?

marques-work:
Updated mockups:

**Tab reordering with confirmation/commit:**

778717d2-8096-49e4-b7f7-e6b96c968e17

**Updated colors and layout:**

316a64c9-ef2f-4780-9e70-8ab6c26a9542

ibnc:
@naveenbhaskar What do we want to do style wise for the edit dropdown? I figure we want the bg color to be the same as the tabs, but what should the colors be for the buttons? (what should the color be for a non-clickable button?)

GaneshSPatil:
Hello @marques-work,
Pressing the `ESC` button while `create new view` popup is open closes the view but the faded background styles stay.

Steps to reproduce:
1. Click on `create new view` popup.
2. Enter view name.
3. Press `ESC`.

GaneshSPatil:
@marques-work @naveenbhaskar -- do we need left scroll and right scroll buttons when there are not enough views to scroll?

GaneshSPatil:
Server Side error Messages are not rendered on the `create new view` modal.
Steps to reproduce:
1. Click on create a new view.
2. Enter view name longer than 64 characters.
3. Hit Save.

`Create new view` should render server-side errors too.

GaneshSPatil:
@naveenbhaskar @marques-work -- Just a thought. As we yank last few characters for a large view name (`max-length`) while rendering, should we also consider having a `min-length` for the viewname area ?

Here is a screenshot:
screen shot 2018-07-31 at 11 33 49 am

GaneshSPatil:
@marques-work @ibnc -- Create new view API renders a text response in case of errors with `Content-Type: application/vnd.go.cd.v1+json;charset=utf-8` Header. 
We follow a convention of sending `{"message": <ERROR_MESSAGE>}` response in case of errors. 

Definitely, this is not related with this PR but might affect while deserializing the create API view response in case of errors.
**Note:** We can fix this as part of separate PR too.

marques-work:
@GaneshSPatil thanks for your feedback; all good points.

**Back/Forward Buttons**:

We detect if we need paging and render the left/right nav buttons differently (faded). Perhaps not different enough though :). I can fade them out more, but I'd like to keep the buttons there to be sure layout calculations for the scrolling and such don't become more complicated.

**Min-length**:

I don't have much of an opinion here. What about you @naveenbhaskar? Current the `min-width` is whatever the width is when the name is a thin character, like `|`. If you think we need some larger clickable area, that's easy to do.

**Server error response**:

Regarding server side errors, I have a stashed change that does this, but I didn't commit it yet because I wasn't sure exactly where to display and didn't write any positioning logic for it. But it's on my radar, certainly. And yes, I can wrap it in the message format (still need to change it to use `HaltApiResponses`.

PR #4865, opened 2018-06-20
Improve handling of empty/null passwords and encrypted passwords

ketan:
- add some code to the `@PostConstruct` annotated methods to set
  `password` and `encryptedPassword` attributes to `null` if they are
  blank strings.
- add some code to deserialize methods to re-encrypt DES passwords as
  required.


config/config-api/src/main/java/com/thoughtworks/go/config/MailHost.java
```diff
@@ -35,8 +35,8 @@
 public class MailHost implements Validatable, PasswordEncrypter {
     @ConfigAttribute(value = "hostname", optional = false) private String hostName;
     @ConfigAttribute(value = "port", optional = false) private int port;
-    @ConfigAttribute(value = "username", optional = true, allowNull = true) private String username = "";
-    @ConfigAttribute(value = "password", optional = true, allowNull = true) private String password = "";//should never be used, will be converted to encrypted password on load, is used by magical-loader
+    @ConfigAttribute(value = "username", optional = true, allowNull = true) private String username;
+    @ConfigAttribute(value = "password", optional = true, allowNull = true) private String password;//should never be used, will be converted to encrypted password on load, is used by magical-loader
```
varshavaradarajan: Remove the comment

config/config-api/src/main/java/com/thoughtworks/go/config/MailHost.java
```diff
@@ -181,6 +181,8 @@ public void setAdminMail(String adminMail) {
 
     @PostConstruct
     public void ensureEncrypted() {
+        this.username = StringUtils.stripToNull(username);
```
varshavaradarajan: Rename the method as we are not just dealing with encryption at post-construct time?

config/config-api/src/main/java/com/thoughtworks/go/config/MailHost.java
```diff
@@ -221,20 +223,23 @@ private void resetPassword(String password) {
     }
 
     private void setPasswordIfNotBlank(String password) {
-        if (StringUtils.isBlank(password)) {
+        this.password = StringUtils.stripToNull(password);
+        this.encryptedPassword = StringUtils.stripToNull(encryptedPassword);
```
varshavaradarajan: Is this necessary? Since we are reassigning the encrypted value later?

config/config-api/src/main/java/com/thoughtworks/go/config/materials/perforce/P4MaterialConfig.java
```diff
@@ -318,6 +321,7 @@ private void setPasswordIfNotBlank(String password) {
 
     @PostConstruct
     public void ensureEncrypted() {
+        this.userName = StringUtils.stripToNull(this.userName);
         setPasswordIfNotBlank(password);
 
         if (encryptedPassword != null) {
```
varshavaradarajan: Shouldn't we do `StringUtils.stripToNull(encryptedPassword) != null`?

config/config-api/src/test/java/com/thoughtworks/go/config/materials/svn/SvnMaterialConfigTest.java
```diff
@@ -163,4 +163,22 @@ public void shouldHandleNullWhenSettingUrlForAMaterial() {
 
         assertNull(config.getUrl());
     }
+
+//    @Test
```
varshavaradarajan: Uncomment?

config/config-server/src/test/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoaderTest.java
```diff
@@ -4171,6 +4175,45 @@ public void shouldMigrateDESEncryptedEnvironmentVariables_XslMigrationFrom108To1
...
+        assertThat(XpathUtils.nodeExists(content, "//*[@password='']"), is(true));
+        assertThat(XpathUtils.nodeExists(content, "//*[@encryptedPassword='']"), is(true));
+
+
+        CruiseConfig config = ConfigMigrator.loadWithMigration(content).config;
+
+        String xmlPartial = xmlWriter.toXmlPartial(config);
+
+        assertThat(XpathUtils.nodeExists(xmlPartial, "//@password"), is(false));
+        assertThat(XpathUtils.nodeExists(xmlPartial, "//@encryptedPassword"), is(false));
```
varshavaradarajan: Also check if a non empty password or encryptedPassword exists, the nodeExists?

config/config-api/src/main/java/com/thoughtworks/go/config/MailHost.java
ketan: It's part of the interface, so unfortunately, I cannot.

config/config-api/src/main/java/com/thoughtworks/go/config/MailHost.java
ketan: Yes this is necessary.

For a situation where a user submits an empty string for `password` and `encryptedPassword`, you'd need to `null`ify both before doing anything.

config/config-server/src/test/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoaderTest.java
ketan: There is no non-empty `password` or `encryptedPassword` in this test. This is covered in other tests, so I let it be.

varshavaradarajan:
This PR takes care of 
- dropping empty/null passwords and encrypted password. We can't do the same for null/empty secure env var values, secure plugin properties as empty. 
- re-encrypting secure values to convert from des to aes at the time of deserialization of passwords, env vars,  plugin properties.

@ketan - do you want to squash some of the commits, may be?

PR #4866, opened 2018-06-20
[WIP] Fetch external artifact view

bdpiparva:
## Todo

- [ ] Show key-value pair view when not able to determine plugin id.


server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -1191,4 +1237,33 @@ public ConfigElementImplementationRegistry getRegistry() {
         return registry;
     }
 
+    public Map<String, Map> artifactIdToPluginIdForFetchPluggableArtifact(String stagePatent,
+                                                                          String currentPipelineName,
+                                                                          String currentStageName) {
+
+        final List<PipelineConfig> pipelineConfigs = stagePatent.equals("templates") ? getAllPipelineConfigs() : pipelinesForFetchArtifacts(currentPipelineName);
```
varshavaradarajan: Just a question: If, for the current pipeline, the dependency materials are not specified, then this returns empty, right? However, until the user clicks on save of fetch task, the validation about the pipeline not being a dependency material won't happen. Which means that, upon entering the artifactId in the fetch view, no plugin view shows up. At the time, we show the plugin dropdown, is it?

server/webapp/WEB-INF/rails.new/app/controllers/admin/tasks_controller.rb
```diff
@@ -67,17 +72,25 @@ def initialize params, user, security_service, task, pluggable_task_service
...
+          else
+            job.addTask(@task)
+          end
           @pluggable_task_service.validate(@task) if @task.instance_of? com.thoughtworks.go.config.pluggabletask.PluggableTask
           @pluggable_task_service.validate(@task.cancelTask()) if (!@task.cancelTask().nil?) && (@task.cancelTask().instance_of? com.thoughtworks.go.config.pluggabletask.PluggableTask)
         end
       end.new(params, current_user.getUsername(), security_service, @task, pluggable_task_service), create_failure_handler, {:controller => '/admin/tasks', :current_tab => params[:current_tab]}) do
         assert_load :job, @node
         assert_load :task, @subject
+        assert_load :artifact_ids, ["foo", "bar"].to_json
```
varshavaradarajan: This needs to populated all artifact ids that can be used, right? I think this is being used for autocomplete. But I might be wrong.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/_external_artifact_config.erb
```diff
@@ -0,0 +1,19 @@
...
+  <div class="contextual_help has_go_tip_right" title="The ID of the artifact."></div>
+  <%= error_message_on(@task, com.thoughtworks.go.config.FetchPluggableArtifactTask::ARTIFACT_ID, :css_class => "form_error") %>
+</div>
+
+<div class="form_item_block artifact_plugin_selection hidden">
+  <span>Could not determine the plugin from the store. Please choose a plugin to configure the plugin properties.</span>
+  <label>Plugin ID</label>
+  <%= select("FetchArtifact", "pluginId", scope[:artifact_plugin_to_fetch_view].collect {|p| [p.to_hash["name"], p.to_hash["id"]]}, {include_blank: true}) %>
+</div>
+<input type="hidden" name="task[pluginId]" id="task_pluginId"/>
```
varshavaradarajan: Shouldn't this make use of `scope[:var_form]` ? That will also generate the same input box with the same name and id.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/_external_artifact_config.erb
```diff
@@ -0,0 +1,19 @@
...
+<div class="form_item_block" id="fetch_pluggable_task_view">
+</div>
+
+<span id="fetch_pluggable_artifact_data" class="plugged_material_data" style="display: none">
```
varshavaradarajan: Change class name.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/_form.html.erb
```diff
@@ -1,50 +1,231 @@
...
+    <div class="contextual_help has_go_tip_right" title="The name of the stage to fetch artifacts from."></div>
+    <%= error_message_on(@task, com.thoughtworks.go.config.FetchTask::STAGE, :css_class => "form_error") %>
+  </div>
+  <div class="form_item_block">
+    <%= label_tag random_dom_id('pipeline_'), required_label_text('Job') -%>
+    <%= scope[:form].text_field com.thoughtworks.go.config.FetchTask::JOB %>
+    <div class="contextual_help has_go_tip_right" title="The name of the job to fetch artifacts from."></div>
+    <%= error_message_on(@task, com.thoughtworks.go.config.FetchTask::JOB, :css_class => "form_error") %>
+  </div>
+  <% if scope[:task].getAppropriateTask().getTaskType() == 'fetch_pluggable_artifact' %>
```
varshavaradarajan: There is already a boolean called `external_fetch_task `. Make use of that?

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/_form.html.erb
```diff
@@ -1,50 +1,231 @@
...
+  }
+
+  function renderViewForSelectedArtifactId() {
+    var artifactId = jQuery("#task_artifactId").val();
+    if (!artifactId || artifactId.trim().length === 0) {
+      hidePluginSelectionDropdown();
+      return;
+    }
+
+    if (artifactId && artifactId.startsWith("#{")) {
```
varshavaradarajan: Make it a contains check? Because I can specify the artifactId as `foo-#{bar}` too.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/_form.html.erb
```diff
@@ -1,50 +1,231 @@
...
+    return artifactInStage[jobName];
+  }
+
+  function isAnyParameterized() {
+    if (arguments.length === 0) {
+      return false;
+    }
+    for (var i in arguments) {
+      var value = jQuery("#" + arguments[i]).val();
+      if (value != null && value.startsWith("#{")) {
```
varshavaradarajan: Check for contains.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/edit.html.erb
```diff
@@ -1,9 +1,23 @@
-<%= render :partial => "admin/tasks/fetch/form.html", :locals => {:scope => scope} %>
+<%= render :partial => "admin/tasks/fetch/form.html", :locals => {:scope => {:task => scope[:task],
+                                                                             :artifact_id_to_plugin_id => @artifact_id_to_plugin_id,
+                                                                             :artifact_ids => @artifact_ids,
```
varshavaradarajan: Not being used, I think?

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/new.html.erb
```diff
@@ -1,9 +1,23 @@
-<%= render :partial => "admin/tasks/fetch/form.html", :locals => {:scope => scope} %>
+<%= render :partial => "admin/tasks/fetch/form.html", :locals => {:scope => {:task => scope[:task],
```
varshavaradarajan: new.html.erb and edit.html.erb seem to have the same content except for routeFetch and routeFetchPluggable. Consider moving to a partial and providing the values for routeFetch and routeFetchPluggable?

config/config-api/src/main/java/com/thoughtworks/go/config/FetchPluggableArtifactTask.java
```diff
@@ -88,7 +120,30 @@ protected void validateAttributes(ValidationContext validationContext) {
...
+            return;
+        }
+        String pluginId = (String) attributeMap.get("pluginId");
+        if (StringUtils.isBlank(pluginId)) {
+            return;
+        }
+
+        ArtifactPluginInfo artifactPluginInfo = getArtifactPluginInfo(pluginId);
+        if (artifactPluginInfo == null) {
+            return;
```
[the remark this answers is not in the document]
bdpiparva: As of now we are not supporting reload of plugin while GoCD server is running. If a plugin is not loaded we don't have a way to identify which values we have to encrypt.

server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
bdpiparva: Yes, the user will not get autosuggest unless they reload the page.

server/webapp/WEB-INF/rails.new/app/views/admin/tasks/fetch/new.html.erb
bdpiparva: I rather keep those `erbs`. I am fine with those duplicated lines.

config/config-api/src/main/java/com/thoughtworks/go/config/FetchPluggableArtifactTask.java
```diff
@@ -88,7 +120,30 @@ protected void validateAttributes(ValidationContext validationContext) {
...
+        }
+
+        for (PluginConfiguration pluginConfiguration : artifactPluginInfo.getFetchArtifactSettings().getConfigurations()) {
```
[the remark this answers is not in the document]
bdpiparva: Not needed

server/webapp/WEB-INF/rails.new/app/controllers/admin/tasks_controller.rb
bdpiparva: That line is not needed.

config/config-api/src/main/java/com/thoughtworks/go/config/FetchPluggableArtifactTask.java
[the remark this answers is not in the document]
varshavaradarajan: I am not saying encrypt it. I am saying deserialize the user given values. If its plaintext, leave it as plain text.

PR #4868, opened 2018-06-20
Fixed infinite redirect issue for authentications.

pranavbansod:


arvindsv:
@bdpiparva @pranavbansod Maybe comment on https://github.com/gocd/gocd/issues/4872, asking whether someone will be able to verify this?

PR #4870, opened 2018-06-21
Introduce usagedata encryption API

GaneshSPatil:
* Allow only admins to access unencrypted /data_sharing/usagedata API
* Introduce encrypted /data_sharing/usagedata/encrypted API for normal users
* Remove _links from encrypted API as its not required
* Use 'application/octet-stream' Content-Type for the APIs returning encrypted response

References: https://tools.ietf.org/html/rfc8188#section-3.1


GaneshSPatil:
Build Passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/3047

PR #4877, opened 2018-06-22
Update pluggable SCM checkout status on console log.

bdpiparva:
This is PR will fail the job if failed to check out material and logs the failure on the console. On success,  log the success message on console log and continue to run next task.

### Failure
screen shot 2018-06-22 at 2 34 35 pm

### Success
screen shot 2018-06-22 at 2 36 48 pm


arvindsv:
From what I saw, it looks reasonable.

Issue #4875, opened 2018-06-22
Handle encryption of plugin properties for full config save flows.


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

#4760 handles encryption of artifact related plugin properties for the pipeline config API. Encryption of plugin properties needs to be done after preprocess and before validation of the entire config. This is so that the parameters are resolved and we can determine which plugin metadata to use for encrypting the properties. Since there are many config save workflows, where we do the encryption is important so as to not cause any more perf issues with config save and handle all flows.

PR #4882, opened 2018-06-27
Handling encryption of pluggable publish/fetch artifact config proper…

jyotisingh:
…ties during full config saves - #4875

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
@@ -46,7 +46,6 @@ import org.apache.commons.lang3.StringUtils;
 import javax.annotation.PostConstruct;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
 
 import static com.thoughtworks.go.util.ExceptionUtils.bomb;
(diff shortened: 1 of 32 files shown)


server/src/test-integration/java/com/thoughtworks/go/config/CachedGoConfigIntegrationTest.java
```diff
@@ -1130,6 +1155,62 @@ public void shouldMarkAPreviousInvalidPartialAsValid_IfMainXMLSatisfiesTheDepend
...
+        String config = IOUtils.toString(getClass().getResourceAsStream("/data/pluggable_artifacts_with_params.xml"), UTF_8);
+        FileUtils.writeStringToFile(configFile, config, UTF_8);
+
+        cachedGoConfig.forceReload();
+
+        Configuration ancestorPluggablePublishAftifactConfigAfterEncryption = goConfigDao.loadConfigHolder()
+                .configForEdit.pipelineConfigByName(new CaseInsensitiveString("ancestor"))
+                .getExternalArtifactConfigs().get(0).getConfiguration();
+        assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getValue(), is("SECRET"));
+        assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getEncryptedValue(), startsWith("AES:"));
```
varshavaradarajan: change to `new GoCipher().encrypt(..)`

server/src/test-integration/java/com/thoughtworks/go/config/FullConfigSaveNormalFlowIntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = {
+        "classpath:WEB-INF/applicationContext-global.xml",
+        "classpath:WEB-INF/applicationContext-dataLocalAccess.xml",
+        "classpath:testPropertyConfigurer.xml"
+})
+public class FullConfigSaveNormalFlowIntegrationTest {
```
varshavaradarajan: Is there anyway to use a base class to execute the tests for FullConfigSaveNormalFlow and FullConfigSaveMergeFlow because the assertions are the same? Only change is which object is used to execute config save.

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -1555,4 +1552,13 @@ public int hashCode() {
         result = 31 * result + (agents != null ? agents.hashCode() : 0);
         return result;
     }
+
+    public class AllTemplatesWithAssociatedPipelines extends HashMap<CaseInsensitiveString, Map<CaseInsensitiveString, Authorization>> {
```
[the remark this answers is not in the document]
jyotisingh: Need to access them from `GoConfigCloner`, so can't make them `private`. `public` or `protected` works - so I left it to be `public`, no real reason for making them `protected`.

server/webapp/WEB-INF/rails.new/lib/java_imports.rb
```diff
@@ -15,7 +15,6 @@
 ##########################GO-LICENSE-END##################################
 
 module JavaImports
-  java_import com.rits.cloning.Cloner unless defined? Cloner
```
varshavaradarajan: I think the Cloner is used is other places to clone entities or just parts of the config.

varshavaradarajan:
LGTM. Waiting for regression to go green.

PR #4884, opened 2018-06-27
Get rid of interceptor injector and AOP from servlet configuration. (done)

akshaymayekar:
1. Changed default annotation handler to **RequestMappingHandlerMapping** from **DefaultAnnotationHandlerMapping**(It has been deprecated).
2. Moved **FlashInterceptor** to RequestMappingHandlerMapping interceptor and removed **InterceptorInjector** and **AOP**.


AksMKR:
To be merged after url-rewrite PR is merged.

server/webapp/WEB-INF/spring-all-servlet.xml
```diff
@@ -38,36 +37,20 @@
   <context:component-scan base-package="com.thoughtworks.go.server.web"/>
   <context:component-scan base-package="com.thoughtworks.go.server.newsecurity.controllers"/>
 
+  <mvc:annotation-driven/>
```
ketan: I believe this is the default behavior with spring.

server/src/main/java/com/thoughtworks/go/server/web/InterceptorInjector.java
```diff
@@ -1,50 +0,0 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *************************GO-LICENSE-END***********************************/
-
-package com.thoughtworks.go.server.web;
-
-import org.aspectj.lang.ProceedingJoinPoint;
```
ketan: Remove the aspectj dependency too?

server/webapp/WEB-INF/spring-all-servlet.xml
AksMKR: This is needed to instantiate RequestMappingHandlerMapping. For DefaultAnnotationHandlerMapping we don't need that tag.

server/src/main/java/com/thoughtworks/go/server/web/InterceptorInjector.java
AksMKR: removed **aspectj** dependency. Keeping the **aspectjweaver** as its used at few places.

maheshp:
Related to https://github.com/gocd/gocd/pull/4887, closing for now.

PR #4888, opened 2018-06-28
Config repos: Use type and origin attribute to determine the type of fetch task.

varshavaradarajan:
Earlier, fetch artifact task and fetch external artifact task were considered as 2 different types of tasks. In this PR, both are considered as the same type and we differentiate between the two based on the origin from which the artifact is to be fetched - gocd or external. This is in par with how we represent the two fetch tasks in config xml.


tomzo:
I understand that config-repo plugins now will need to submit `origin` field in fetch tasks. While this PR seems OK, I think there is an update needed in migration in https://github.com/gocd/gocd/pull/4779 to make sure that this field is added on older fetch tasks.

varshavaradarajan:
@tomzo - thanks for taking a look. Yes, #4779 will have to be updated. We are currently having a discussion as to whether this attribute should be called `origin` because origin has previously always been used in the context of config-repos.  We'll merge this after deciding on the attribute name.

varshavaradarajan:
@tomzo - we have decided on `aritfact_origin`. Once #4899 is merged, we will merge this.

PR #4891, opened 2018-06-29
Remove templates API v4 from rails.

varshavaradarajan:
* The template API v4 was moved to spark when adding support for external artifacts.


GaneshSPatil:
The changes LGTM, we can merge post the green PR build

PR #4895, opened 2018-06-29
Clean up CaseInsensitiveString

marques-work:
- rename static caseInsensitiveStrings() to list()
  - simpliy .equals() and .hashCode()


marques-work:
@ketan I'm merging this small PR because I think it's very low risk. It's not passing the integration tests for the same reasons/issues that master is having with the h2 upgrade. If this is problematic, I'm happy to reverse-merge it.

PR #4896, opened 2018-07-02
Revert "Upgrade to h2-1.4.197"

ketan:
This reverts commit 4929bf14cfe4032dcebf41a7a327740a2c67d222.


peterrosell:
According to the discussion from the link you provide the problem is that the database is corrupt and that version 1.4 detects these errors earlier now and throws an exception.

> That's **not a format change, that's a "we prefer to detect corruption earlier rather than later" change**​
>
>**Obviously your DB become corrupted at some point in the past**, but H2 didn't notice back then because it was not checking the range of such things.
>
>I suggest you revert to an earlier version, do a dump, and restore into the new version, after fixing up the data.

So that mean that you will never be able to upgrade unless you do the backup-and-restore procedure. I think I saw this error when I played around with an upgrade by myself. I didn't see this workaround back then and dropped my try. 

It might be interesting to do that workaround and verify that it works. Of course it will be a hard automate this in a simple gocd-server upgrade, but I think it's worth to use the newer version of h2. 

Might it be possible to configure which version to use? So a larger group can test the new h2 if they want and issues can be solved before everyone starts using it as default.

Issue #4905, opened 2018-07-03
[External Artifact publish view] any key operations on store id field removes off value already provided


rajiesh:
##### Issue Type

- Bug Report

##### Summary

If user use params in store_id and do Ctrl+C on that field it erases out all the value already provided and asks for the plugin id to be selected

###### Basic environment details

* Go Version: 18.7.0 (6990-a4e86591253b2b689af07cf6c98d9babb6d41b5b)
* Docker registry external artifact plugin - 0.0.1-41

##### Steps to Reproduce

Say user configured external publish artifact as here(using param for store ID),

<img width="983" alt="screen shot 2018-07-03 at 11 56 44 am" src="https://user-images.githubusercontent.com/6533924/42202271-6659de5e-7eb8-11e8-839b-d30afad606a1.png">

Now if user selects value in store ID field and do Ctrl+C the previous selection goes off as seen here,

<img width="908" alt="screen shot 2018-07-03 at 11 56 57 am" src="https://user-images.githubusercontent.com/6533924/42202342-a139b170-7eb8-11e8-88c3-0d20996354c2.png">
which you don't want to be shared.
Remember to surround them in 3 backticks like this —

```
long log lines go here
```
-->

rajiesh:
Verified on 18.7.0 (7107-ca37138fb0de89afb2f607cb57de22478550f8ac)

PR #4908, opened 2018-07-03
[CSS] Fix css for artifacts tab in job details.

varshavaradarajan:
* Inline the delete artifact icon so that it doesn't take up the left over space for the entire row.


varshavaradarajan:
#4905

PR #4911, opened 2018-07-04
Improve API related error messages.

varshavaradarajan:
If a mandatory property is not provided by a user, the error message currently reads

`"Json does not contain property 'origin'"`

Since the above message does not indicate which part of the API had to contain the mandatory property, adding that info makes sense. With this PR, the error message now reads

`"Json {\"pipeline\":\"\",\"stage\":\"defaultStage\",\"job\":\"defaultJob\",\"artifact_id\":\"helpless-quinoa\",\"configuration\":[{\"key\":\"fetch_property\",\"value\":\"foo\"}]} does not contain property 'origin'"`

@ketan - if this is alright, I'll do the same for all the other messages in `HaltApiMessages.java`.

### DIFF
--- a/api/api-base/src/main/java/com/thoughtworks/go/api/representers/JsonReader.java
+++ b/api/api-base/src/main/java/com/thoughtworks/go/api/representers/JsonReader.java
@@ -41,7 +41,7 @@ public class JsonReader {
 
     public String getString(String property) {
         return optString(property)
-            .orElseThrow(() -> haltBecauseMissingJsonProperty(property));
+            .orElseThrow(() -> haltBecauseMissingJsonProperty(property, jsonObject));
     }
 
     public Optional<Long> optLong(String property) {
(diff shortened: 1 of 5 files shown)


ketan:
This is definitely better than the old message, however one of the problems I see with this is when you have similar nodes in multiple places in the json document that a user submits.

I was thinking if there's a way to send the path to the particular json node that's being parsed and has an error. Kind of how json assert and json unit prints in an error message. If that is not doable, I think this solution is just fine as well.

varshavaradarajan:
@ketan - when we initialize json reader in the respective representers, we only have the json related to that part of the config as json object, which I used in the error message. So, not sure how to get the path to the json node.

ketan:
Like I said before, if it's not doable, this is still an improvement over before. So merge as you see fit.

Also — do we need similar methods for other halt messages around JSON parsing?

varshavaradarajan:
@ketan - yes. This was just a sample. Which halt method makes use of `property`, I am planning to pass along the json object to it.

Issue #3390, opened 2017-04-12
Deleting pipeline config via API don't notify scheduleService to cancel job


kulikov:
##### Issue Type

- Bug Report

##### Summary

When I remove pipeline config via API I've got some repeated errors in gocd server logs:

```
2017-04-12 15:49:25,519 ERROR [86@MessageListener for ScheduleCheckListener] JMSMessageListenerAdapter:77 - Exception thrown in message handling by listener com.thoughtworks.go.server.scheduling.ScheduleCheckListener@75b61bef
java.lang.RuntimeException: com.thoughtworks.go.config.PipelineNotFoundException: Pipeline 'front-office-table-rendering-fix' not found.
	at com.thoughtworks.go.server.scheduling.ScheduleCheckListener.onMessage(ScheduleCheckListener.java:51)
	at com.thoughtworks.go.server.scheduling.ScheduleCheckListener.onMessage(ScheduleCheckListener.java:24)
	at com.thoughtworks.go.server.messaging.activemq.JMSMessageListenerAdapter.runImpl(JMSMessageListenerAdapter.java:73)
	at com.thoughtworks.go.server.messaging.activemq.JMSMessageListenerAdapter.run(JMSMessageListenerAdapter.java:53)
```

This logs repeated each 10 seconds.

API request that I use for deleting:
```
DELETE https://localhost:8154/go/api/admin/pipelines/front-office-table-rendering-fix
200 OK
```

##### Environment
```plain
{
  "Timestamp": "2017-04-12T16:37:26Z",
  "Go Server Information": {
    "Version": "17.3.0 (4704-3c7d8f7e5d78e46f72ff1194bb322c6b5d72a90f)"
  },
  "Config Statistics": {
    "Valid Config": {
      "Number of pipelines": 8,
      "Number of agents": 2,
      "Number of environments": 1,
      "Number of unique materials": 7,
      "Number of schedulable materials": 7
    },
    "Security": {
      "LDAP": false,
      "Password": true
    }
  },
  "Config file locations": {
    "loc.config.dir": "/etc/go",
    "loc.log.root.0": "/var/lib/go-server/.",
    "loc.log.basename.0": "plugin-slack.notifier.log"
  },
  "OS Information": {
    "OS Name": "Linux",
    "OS Version": "4.4.0-64-generic",
    "System Architecture": "amd64",
    "Available Processors": 2,
    "Average System Load": 1.14
  },
  "Runtime Information": {
    "Name": "12@...",
[...]

PR #4913, opened 2018-07-04
Remove Jobs from JobPlans If the pipeline is deleted after it is scheduled (But the jobs are still unassigned) (#3390)

GaneshSPatil:


server/src/test-integration/java/com/thoughtworks/go/server/service/BuildAssignmentServiceIntegrationTest.java
```diff
@@ -702,6 +702,24 @@ public void shouldCancelAScheduledJobInCaseStageHasBeenRenamed() throws Exceptio
...
+        u.checkinInOrder(hgMaterial, hgRevs);
+
+        ScheduleTestUtil.AddedPipeline p1 = u.saveConfigWith("PIPELINE_WHICH_WILL_EVENTUALLY_WILL_BE_DELETED", u.m(hgMaterial));
+        BuildCause buildCause = BuildCause.createWithModifications(u.mrs(u.mr(u.m(hgMaterial).material, true, hgRevs)), "user");
+        Pipeline originalPipelineRun = scheduleService.schedulePipeline(p1.config.name(), buildCause);
+        pipelineConfigService.deletePipelineConfig(new Username("user"), p1.config, new HttpLocalizedOperationResult());
+
+        CruiseConfig cruiseConfig = configHelper.load();
+        buildAssignmentService.onTimer();   // To Reload Job Plans
+        buildAssignmentService.onConfigChange(cruiseConfig);
```
jyotisingh: `buildAssignmentService.onConfigChange` would have already taken care of deletion of jobPlans that once belonged to a deleted pipeline. The bug is about these not getting removed if pipeline was deleted via the API.  So, I assume this test will pass even if you remove the fix you have made.

Since you have already called `pipelineConfigService.deletePipelineConfig` from this test to simulate the API, which implicitly calls `BuildAssignmentService.pipelineConfigChangedListener` you should just be able to make your assertions right away.  Remove the call to  `buildAssignmentService.onConfigChange(cruiseConfig)`

server/src/test-fast/java/com/thoughtworks/go/server/service/BuildAssignmentServiceTest.java
```diff
@@ -162,4 +163,69 @@ public void shouldMatchNonElasticJobToNonElasticAgentIfResourcesMatch(){
...
+        JobPlan jobPlan3 = getJobPlan(irrelevantPipeline.getName(), irrelevantPipeline.get(0).name(), irrelevantPipeline.get(0).getJobs().first());
+
+        jobPlans.add(jobPlan1);
+        jobPlans.add(jobPlan2);
+        jobPlans.add(jobPlan3);
+
+        //delete a stage
+        pipeline.remove(1);
+
+        List<JobPlan> jobsToRemove = buildAssignmentService.getMismatchingJobPlansFromUpdatedPipeline(pipeline, jobPlans);
```
jyotisingh: Why dont you invoke `pipelineConfigChangedListener()` from the test instead of exposing these methods - both `getMismatchingJobPlansFromUpdatedPipeline` and `getAllJobPlansFromDeletedPipeline ` for the test. That would make the tests a bit more explicit

PR #4918, opened 2018-07-04
Fix styling on data sharing settings page

GaneshSPatil:
* Render consent-wrapper div and shared-data div with equal widths (50%)

screen shot 2018-07-04 at 4 10 09 pm
screen shot 2018-07-04 at 4 10 30 pm


GaneshSPatil:
\cc: @naveenbhaskar @jyotisingh

PR #4922, opened 2018-07-05
PipelineSelections now serialize to JSON

marques-work:
The main goal of this effort is to build support for managing multiple personalization settings per user entirely into the frontend. With the frontend managing most of the logic, the backend is basically a hard disk, metaphorically; its responsibility is reduced to the simple task of de/serializing JSON to and from the database, leaving all other logic to be contained in the UI.

The main benefits probably won't be fully appreciated until we build the ability to name and define multiple filters on the UI, which will be in PR #4864.

`PipelineSelections` now holds a `Filters` object, which is just a fancy holder for a list of `DashboardFilter` objects that either act as blacklist or whitelist filters.

The `Filters` object serializes to JSON and back.

`DashboardFilter` implements a `isPipelineVisible(pipelineName)` method to determine whether or not a given pipeline be shown. The idea here is that `PipelineSelections#includesPipeline()` will delegate to this method on the currently selected `DashboardFilter` (which is the analog to a named view tab in the UI).

The `DashboardControllerDelegate` accepts a `viewName` query parameter, indicating the name of the view filter, and selects the named filter from the user's filter set to generate the filtered set of pipelines. Currently, this bit is a dark launch, but maintains existing functionality in the absence of the parameter.

Currently, the filters JSON structure looks something like this:

```json
{
  "filters": [
    {"name": "my filter", "pipelines": ["p1", "p2"], "type": "whitelist" },
    {"name": null, "pipelines": [], "type": "blacklist" }
  ]
}
```

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/datamigration/M001.java
+++ b/server/src/main/java/com/thoughtworks/go/server/datamigration/M001.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.datamigration;
+
+import com.google.gson.Gson;
+import org.apache.commons.lang3.StringUtils;
+
+import java.sql.*;
+
+class M001 {
+    private static final String ID = "id";
+    private static final String SELECTIONS = "selections";
+    private static final String BLACKLIST = "isblacklist";
+    private static final int SCHEMA = 1;
+    private static final Gson GSON = new Gson();
+
+    static Migration convertPipelineSelectionsToFilters() {
+        return (cxn) -> {
+            if (!required(cxn)) return;
+
+            try (Statement s = cxn.createStatement()) {
+                final ResultSet rs = s.executeQuery("SELECT id, selections, isblacklist FROM pipelineselections WHERE version = 0");
+                while (rs.next()) {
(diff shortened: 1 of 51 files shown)


server/src/main/java/com/thoughtworks/go/server/datamigration/M001.java
```diff
@@ -0,0 +1,85 @@
...
+            }
+        };
+    }
+
+    static void perform(Connection cxn, long id, String selections, boolean isBlacklist) throws SQLException {
+        try (PreparedStatement ps = cxn.prepareStatement("UPDATE pipelineselections SET selections = NULL, version = ?, filters = ? WHERE id = ?")) {
+            ps.setInt(1, SCHEMA);
+            ps.setString(2, asJson(selections, isBlacklist));
+            ps.setLong(3, id);
+            ps.executeUpdate();
```
marques-work: The data migration will essentially convert the pipeline selections representation into a JSON document
with this structure:

```
{
  filters: [
    { name: "first filter", type: "whitelist", pipelines: [ "p1", "p2", ..., "pN" ] },
    { name: "second filter", type: "blacklist", pipelines: [ "p4", "p5" ] },
    ...
  ]
}
```

The implications here are that we can support multiple pipeline filters per user as a single row in the
table and that this will allow us to consolidate and shift more logic to the frontend. The backend can
be written in a way such that its only concerns are serialization, deserialization, and persistence.

server/src/main/java/com/thoughtworks/go/server/domain/user/PipelineSelections.java
```diff
@@ -20,159 +20,113 @@
...
+    public void update(Filters filters, Date date, Long userId) {
         this.userId = userId;
-        this.isBlacklist = isBlacklist;
-        updateSelections(selections);
         this.lastUpdate = date;
+        this.viewFilters = null == filters ? Filters.defaults() : filters;
+        this.version = CURRENT_SCHEMA_VERSION;
     }
 
+    @Deprecated // TODO: remove when removing old dashboard
```
marques-work: `includesGroup()`, `includesPipeline()`, and `isBlacklist()` are only used by the old dashboard page. These can be removed once we eliminate the old dashboard. This logic is encapsulated in `DashboardFilter` instances now.

https://github.com/marques-work/gocd/blob/92a6806b3eccb22c0ac7badc9b86dd9c7b613402/server/src/main/java/com/thoughtworks/go/server/domain/user/PipelineSelections.java#L83-L101

arvindsv:
Ketan and Marques had talked about this earlier. Marques has paired on this, and Ketan has also reviewed it. Was waiting for release to merge. There's another part to this, which makes the actual UI changes, which will need reviewing / pairing.

Issue #4923, opened 2018-07-06
Analytics Plugin doesn't works for GoCD windows instance


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary
Windows GoCD server fails to load any analytics chart to the path issue.
This issue is while creating an URI from a file path. The system-dependent default name-separator character needs to be used instead of hardcoding it to `/`.
On UNIX systems `/`, whereas, on windows systems `\\` is used as name-separator character.

Go Server Log:
```
2018-07-05 15:55:48,037 ERROR [qtp1697511312-49] AnalyticsDelegate:106 - Encountered error while fetching analytics 
java.lang.IllegalArgumentException: Illegal character in path at index 6: assets\plugins\com.thoughtworks.gocd.analytics\B31EA0A3366304AFA5C012D7250339F125C132BC6DA9599541318B40982EEB90/agent-with-highest-utilization-chart.html
```

###### Basic environment details

* Go Version: `v18.6.0`
* Plugin Version: `1.0.0-683`

##### Expected Results
GoCD Server should load all the analytics charts on windows operating systems.

##### Actual Results
GoCD Server fails to load all the analytics charts on windows operating systems.

##### Possible Fix
Replace [`/`](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/analytics/AnalyticsData.java#L46) with the operating specific name sepearator character.
Use [`File.separator`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator) which is returns system-dependent default name-separator character.

rajiesh:
Verified on 18.7.0 (7107-ca37138fb0de89afb2f607cb57de22478550f8ac

PR #4924, opened 2018-07-06
    Use FilenameUtils.separatorsToUnix to convert all the '\\' to '/' which is a valid name seperator character in the URI (#4923)

GaneshSPatil:
Use FilenameUtils.separatorsToUnix to convert all the '\\' to '/' which is a valid name seperator character in the URI (#4923)
    * The URI path creation fails on windows due to unsupported character '\\' being used in the URI.
    * Convert all the name separator character to the Unix name separator '/' before creating the URI from the file path.

### DIFF
--- a/plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/analytics/AnalyticsData.java
+++ b/plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/analytics/AnalyticsData.java
@@ -16,8 +16,11 @@
 
 package com.thoughtworks.go.plugin.domain.analytics;
 
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
+
 import java.net.URI;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/analytics/AnalyticsData.java
```diff
@@ -43,7 +45,7 @@ public String getFullViewPath() {
         if (StringUtils.isBlank(assetRoot)) {
             return viewPath;
         }
-        return URI.create(assetRoot + "/" + viewPath).normalize().toString();
+        return URI.create(assetRoot + File.separator + viewPath).normalize().toString();
```
ketan: Won't this use the wrong separator on windows?

GaneshSPatil: @ketan -- File.separator will give the right path separator based on the OS. References: https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator

But the problem is with `URI.create`.  Its failing if we provide `\\`. As the URI cannot contain backward slash.

Instead of URI we can use `Paths.get`.

\cc: @maheshp

PR #4928, opened 2018-07-09
Clear single tone stores before and after test.

bdpiparva:


server/src/test-fast/java/com/thoughtworks/go/server/service/plugins/builder/DefaultPluginInfoFinderTest.java
```diff
@@ -49,29 +49,23 @@
     private DefaultPluginInfoFinder finder;
     private PluggableTaskMetadataStore taskMetadataStore;
     private NotificationMetadataStore notificationMetadataStore;
-    private AuthorizationMetadataStore authorizationMetadataStore;
-    private ElasticAgentMetadataStore elasticAgentMetadataStore;
 
     @Before
     public void setUp() throws Exception {
+        com.thoughtworks.go.ClearSingleton.clearSingletons();
```
ketan: FYI. `ClearSingleton` is now a `@Rule`:

```
    @Rule
    public final ClearSingleton clearSingleton = new ClearSingleton();
```

bdpiparva: Thanks!

PR #4929, opened 2018-07-09
Do not get current revision from config.git within synchronized block.

varshavaradarajan:
There is no need for this to be a blocking call.

@jyotisingh - do you remember why this was within a lock?


jyotisingh:
I guess that's because all actions on the config.git repo are synchronized to prevent the commands from interfering from each other. Any specific reason for getting rid of it?

varshavaradarajan:
The actions that write something to the git repo, it makes sense to do it within a lock. This is called from configuration_controller for **getting** the cruise-config xml. Sometimes, the response time is slow. I  checked other usages as well, I didn't find any reason for the getCurrentRevision method to be inside the synchronized block.

varshavaradarajan:
@jyotisingh - closing this. Let me know if you want me to follow up on this.

Issue #4858, opened 2018-06-18
Artifact upload fails on file name that includes ".." ?


gunnarx:
##### Issue Type
- Bug Report

##### Summary

Upload of artifacts directory failed with IllegalPathException and "is outside extraction target directory".

###### Basic environment details

* Go Version: 17.1.0 (4511-c7c379ff500a5ae70c8cd7a49f936595c6bd2d69)
* JAVA Version: N/A
* OS: Ubuntu Linux
* Browser vendor and version (if relevant): N/A

(It's an old version, I know.  If you believe this to be fixed, just close the ticket and I'll try once upgraded)

###### Additional Environment Details

N/A

##### Steps to Reproduce

Define a job with an artifact defined to point to a directory containing many files, including the one that has a problem.

##### Expected Results
Successful upload

##### Actual Results
Upload failed, retries 3 times, then stops.
Go server logs indicate error as given in the long log below.

The file in question appears to be named exactly this:  
``` 
series.subdir..._.._
```
The file is just a plain text file.

##### Possible Fix

Possibly:
* Is it possible that the check for illegal path is giving a false error (i.e. it reacts to the ".." in the file)?
* Maybe, check paths only for directories, not for the actual file name... 

##### Log snippets

```
2018-06-18 12:38:49,843 ERROR [qtp1604839423-18] ArtifactsService:107 - Failed to save the file to: [/var/lib/go-server/artifacts/pipelines/Release-minnowboard/26/build/2/init_and_bitbake]
com.thoughtworks.go.util.IllegalPathException: File release/sources/x86_64-poky-linux/mozjs-17.0.0-r0/series.subdir..._.._ is outside extraction target directory
        at com.thoughtworks.go.util.ZipUtil.bombIfZipEntryPathContainsDirectoryTraversalCharacters(ZipUtil.java:117)
        at com.thoughtworks.go.util.ZipUtil.extractTo(ZipUtil.java:136)
        at com.thoughtworks.go.util.ZipUtil.unzip(ZipUtil.java:125)
        at com.thoughtworks.go.server.service.ArtifactsService.saveFile(ArtifactsService.java:89)
        at com.thoughtworks.go.server.controller.ArtifactsController.saveFile(ArtifactsController.java:204)
        at com.thoughtworks.go.server.controller.ArtifactsController.postArtifact(ArtifactsController.java:166)
        at sun.reflect.GeneratedMethodAccessor511.invoke(Unknown Source)
[...]

PR #4941, opened 2018-07-10
Allow artifact files with `..` in them

arvindsv:
It should be safe to allow artifact files with ".." in them, as long as the path doesn't contain a "/../"

Tested with a config such as this:

```
<pipeline name="Test">
  <materials>
    <git url="https://github.com/arvindsv/faketime.git" />
  </materials>
  <stage name="defaultStage">
    <jobs>
      <job name="defaultJob">
        <tasks>
          <exec command="mkdir">
            <arg>test</arg>
          </exec>
          <exec command="touch">
            <arg>test/series.subdir..._.._</arg>
            <runif status="passed" />
          </exec>
        </tasks>
        <artifacts>
          <artifact type="build" src="test" dest="test" />
        </artifacts>
      </job>
    </jobs>
  </stage>
</pipeline>
```

For #4858.

There are other parts such as [this](https://github.com/gocd/gocd/blob/07d8dc6d5aa5331cdbfff7b1812cffb571f884a2/server/src/main/java/com/thoughtworks/go/server/controller/ArtifactsController.java#L228) which will cause problems when trying to view that artifact in the artifact tab, though.


base/src/main/java/com/thoughtworks/go/util/ZipUtil.java
```diff
@@ -104,8 +104,8 @@ void addToZip(ZipPath path, File srcFile, ZipOutputStream zip, boolean excludeRo
         }
     }
 
-    private void bombIfZipEntryPathContainsDirectoryTraversalCharacters(String filepath) {
-        if (filepath.contains("..")) {
+    private void failIfZipEntryPathContainsDirectoryTraversalCharacters(String filepath) {
+        if (filepath.contains("/../") || filepath.startsWith("../") || filepath.endsWith("/..")) {
```
arvindsv: Didn't want to actually get the canonical path and compare, etc. Thought it might just add a bit of overhead.

ketan: What about a windows equivalent path `\..\` or `\\..\\` depending on whether it's escaped, or not?

`FileNameUtils#separatorsToUnix(filepath).split('/').contains("..")` is what I'd think we'd want?

varshavaradarajan: @arvindsv - why not make use of something like `FilenameUtil.isNormalizedPathOutsideWorkingDir` ? Its what we use to validate that the task working directory is not outside the agent sandbox.

arvindsv: @varshavaradarajan I did consider using something like:

```
filepath.contains("..") && FilenameUtil.isNormalizedDirectoryPathInsideNormalizedParentDirectory(dir, filepath)
```

but it seems unnecessary. A zip entry (an entry *inside* a zip file) with `../` in the name seems unnecessary to support. Also, `isNormalizedDirectoryPathInsideNormalizedParentDirectory` expects the second argument to be a directory - and changing/rewriting that seemed unnecessary. What do you think?

arvindsv: @ketan I think you're right. I'll think about it. In this case though, this is a zip entry. The backslash separator doesn't really apply. However, I'll need to see how such a zip entry shows up in Windows. Meaning, whether a zip file created in Windows uses backslash or it is standardized to a forward slash.

arvindsv: @ketan FYI:

- The [zip format specification](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) says that the slashes must be forward slashes only.
- .Net framework was also finally changed in May this year to reflect this.

I guess it is possible for zip files generated on Windows (and maybe by our Windows agents) to have backslash as the name separator. I'll verify and convert them to a forward slash before comparison, if that's the case.

GaneshSPatil:
@arvindsv @maheshp -- are we doing this PR for the upcoming release?

arvindsv:
No. Moving it.

adityasood:
@arvindsv @maheshp since this was moved to 18.11, do we merge this and test it for this release?

base/src/main/java/com/thoughtworks/go/util/ZipUtil.java
arvindsv: @adityasood @ketan I tried this with a Windows agent too and it works. The zip file is generated and uploaded (and downloaded on the server successfully).

Also, tried generating [a zip file](https://gist.github.com/arvindsv/89b334567e9ddebd43aaf70cca88e704) with Windows separators and a `\..\` and it uploaded to a server running on Mac OS X and it created a file with backslashes in it, which is valid on a Mac.

```
$ unzip -l windows_dots.zip
Archive:  windows_dots.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        9  11-19-2018 18:43   1.txt
        9  11-19-2018 18:43   \..\2.txt
---------                     -------
       18                     2 files

$ curl http://localhost:8153/go/files/Test-Zipping/2/defaultStage/2/defaultJob/test2 -H 'Confirm: true' -X POST -F 'zipfile=@windows_dots.zip'
File test2 was created successfully

$ ls -l
total 8,192
-rw-r--r-- 1 user staff 9 Nov 19 18:53  1.txt
-rw-r--r-- 1 user staff 9 Nov 19 18:53 '\..\2.txt'
```

Just to be on the safe side, I'll look at disallowing forward and backslashes, disregarding which OS this code is running on (since it could be a Windows agent + Linux server or Linux agent + Windows server ... etc.)

arvindsv:
@adityasood @ketan I've now made it reject `/../` and `\..\` irrespective of OS of the server or agent. Maybe that's enough? This code also passes all the tests:

```java
if (Arrays.asList(filepath.split("/")).contains("..") || Arrays.asList(filepath.split("\\\\")).contains("..")) {
    throw new IllegalPathException(String.format("File %s is outside extraction target directory", filepath));
}
```

I have no preference.

adityasood:
@arvindsv I think you have covered the scenarios.

GaneshSPatil:
@arvindsv @adityasood -- can we merge this? is there anything else we are waiting on?

GaneshSPatil:
@arvindsv -- ping!

ketan:
Closing, no activity

arvindsv:
Boo! But fair.

PR #4947, opened 2018-07-11
Pipeline history- allow periods in pipeline names while getting the pipeline name from the path.

varshavaradarajan:


varshavaradarajan:
Closing this since we are reverting the urlrewrite changes in #4952 .

PR #4953, opened 2018-07-12
Create short lived API sessions for some of the agent requests.

varshavaradarajan:
* /admin/agent-plugins.zip
* /admin/agent
* /admin/agent/token
* /admin/tfs-impl.jar
* /admin/latest-agent.status


base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -140,6 +140,7 @@
     public static GoSystemProperty<Integer> IDLE_TIMEOUT = new GoIntSystemProperty("idle.timeout", 30000);
     public static GoSystemProperty<Integer> RESPONSE_BUFFER_SIZE = new GoIntSystemProperty("response.buffer.size", 32768);
     public static final GoSystemProperty<Integer> API_REQUEST_IDLE_TIMEOUT_IN_SECONDS = new GoIntSystemProperty("api.request.idle.timeout.seconds", 300);
+    public static final GoSystemProperty<Integer> AGENT_REQUEST_IDLE_TIMEOUT_IN_SECONDS = new GoIntSystemProperty("agent.request.idle.timeout.seconds", 30);
```
maheshp: Do we know the Agent request timeout, I am just wondering if session time should be higher than that?

PR #4966, opened 2018-07-19
Prevent attempt to re-add a pre-registered agent to config

jyotisingh:
If a pre-registered agent with auto-register keys setup makes an attempt to register with the server (in case the keys on the agent side were deleted), the server would try to add such an agent to config all over again, and fail much later during the xsd validation stating that a duplicate entry was being inserted into the config file. Since the agent attempting registration has access to both the uuid and the right token that was issued to it, its alright to send over keys to this agent. Avoiding re-adding of this agent to config.xml


GaneshSPatil:
The PR LGTM, can be merged post green PR build

PR #4982, opened 2018-07-26
Pass along server id for the update version info check along with current version

GaneshSPatil:


server/webapp/WEB-INF/rails.new/app/controllers/api_v1/version_infos_controller.rb
```diff
@@ -48,10 +47,14 @@ def latest_version
 
     private
 
-    def to_json_hal(version_info)
+    def server_id
+      data_sharing_usage_statistics_reporting_service.get.getServerId
```
maheshp: `data_sharing_usage_statistics_reporting_service.get` makes a call to the DB to load the 'UsageStatisticsReporting' object, consider caching this?

PR #4986, opened 2018-07-27
Cache data sharing objects instead of loading them always from DB

GaneshSPatil:


server/src/main/java/com/thoughtworks/go/server/dao/DataSharingSettingsSqlMapDao.java
```diff
@@ -48,11 +55,32 @@ public void saveOrUpdate(DataSharingSettings dataSharingSettings) throws Duplica
...
+                });
+
+                sessionFactory.getCurrentSession().saveOrUpdate(toSave);
+            }
+        });
     }
 
     public DataSharingSettings load() {
-        return transactionTemplate.execute(status -> (DataSharingSettings) sessionFactory.getCurrentSession().getNamedQuery("load.datasharing.settings").uniqueResult());
+        if (cachedSettings == null) {
```
jyotisingh: You need to synchronize on `cachedSettings` while setting as well as invalidating the cache, add double checking too.

GaneshSPatil: Ok. will make that change

GaneshSPatil: @ketan -- Data Sharing Settings API is used by Consent SPA, It might not be used so frequently, but its good to cache. 
Usage Data Reporting API Information is accessed by **every browser client every half an hour** to check if it can perform data reporting, hence usage data information is necessary to cache!

server/src/main/java/com/thoughtworks/go/server/dao/DataSharingSettingsSqlMapDao.java
```diff
@@ -48,11 +55,32 @@ public void saveOrUpdate(DataSharingSettings dataSharingSettings) throws Duplica
...
+            cachedSettings = transactionTemplate.execute(status -> (DataSharingSettings) sessionFactory.getCurrentSession().getNamedQuery("load.datasharing.settings").uniqueResult());
+        }
+
+        return cachedSettings;
+    }
+
+    public void invalidateCache() {
```
jyotisingh: Also, this method need not be public.

GaneshSPatil: I know. but the integration Test setup is in a manner that `DBHelper.onTeardown()` will clear off the data from the DB but won't clear the cached object. To do so, we need it public.

Another workaround would be to put it under `GoCache`, but the cached object is accessed within that specific DAO instance, hence kept the cached object local and exposed a method to clear it off!

jyotisingh: I think we should stick to `GoCache`, its not a workaround.

GaneshSPatil: Ok

PR #4989, opened 2018-07-30
Usage data /encrypted POST API

GaneshSPatil:
* Fetch subordinate public key and signature from data sharing remote server
* update usage_data/encrypted call to be a POST API call, which returns encrypted usage data provided the signature and subordinate public key (which was fetched from remote data sharing server).


GaneshSPatil:
Only to be merged after usage-data-collector-app is updated to handle `/encryption_keys` GET API.

Issue #4971, opened 2018-07-20
Do not expose gocd server class-loader to a plugin


bdpiparva:
##### Issue Type
- Bug Report

##### Summary
GoCD server classloader is being exposed to the plugin and it creates an issue in certain scenarios. One such scenario is when plugins dependencies are using current thread classloader to load the class.

```java
public static Class loadClass(String className) throws ClassNotFoundException {
  try {
      ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
      return contextClassLoader.loadClass(className);
  } catch (Exception e) {
      return Class.forName(className);
  }
}
```

Above code will load the specified class using GoCD server classloader and from GoCD server dependencies which cause plugins to not work.

###### Basic environment details
* Go Version: `Go Version: 18.7.0 (7118-8494e0df4236380e6b1260f21bea2df13703b64c)`

##### Expected Results
It should load the class from dependencies defined in the plugin, not from the server.

##### Possible Fix

https://github.com/gocd/gocd/blob/9e996b9983ea65eee6dbca364ac02873b088ed53/plugin-infra/go-plugin-infra/src/main/java/com/thoughtworks/go/plugin/infra/DefaultPluginManager.java#L142-L148

Before submitting a request to plugin set current threads classloader to plugins classloader and revert it back to original classloader at the end.

maheshp:
Fixed as part of https://github.com/gocd/gocd/pull/5011

PR #4992, opened 2018-07-31
Setting thread context classloader to bundle's classloader before invoking the plugin - #4971

jyotisingh:
While osgi handles classloader isolation for the bundle in most cases, thread context classloader was always set to parent's classloader (webapp classloader in this case) causing the plugin to load up classes with Thread.currentThread().getContextClassLoader().loadClass(class-name) from dependencies loaded up by webapp instead of the ones provided by the plugin


akshaydewan:
Closing this PR and using https://github.com/gocd/gocd/pull/5011 instead

PR #4995, opened 2018-08-02
Invalidate cache aftercommit within the transaction started by the service

GaneshSPatil:


GaneshSPatil:
PR build passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/3204

PR #5000, opened 2018-08-05
Dashboard view tabs list dropdown

marques-work:
Adds a dropdown as an alternate way to navigate dashboard view tabs. Depends on #4864.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/tabs_list_dropdown.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/tabs_list_dropdown.js.msx
@@ -1,0 +1,37 @@
+const m = require("mithril");
+const _ = require("lodash");
+
+const TabsListDropdown = {
+  view(vnode) {
+    const vm = vnode.attrs.vm;
+    return <div class="tabs-list-dropdown-widget">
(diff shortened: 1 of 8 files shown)


marques-work:
@naveenbhaskar Hi Naveen, could you review the UI interactions, etc?

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/tabs_list_dropdown.js.msx
```diff
@@ -0,0 +1,37 @@
...
+  }
+};
+
+const ViewTabsDropdownToggle = {
+  view(vnode) {
+    const vm = vnode.attrs.vm;
+    function toggleTabsListDropdown(e) { e.stopPropagation(); vm.toggleTabsListDropdown(); }
+
+    return <button class="tabs-list-dropdown-toggle" onclick={toggleTabsListDropdown}>
+      <i class="fa fa-caret-down" />
```
akshaydewan: A small thing - we've been following the convention of using font-awesome in CSS rather than using the classes directly in HTML. (E.g. see `.bell` in `_notifications.scss`). There is a safety net in place which blows up asset compilation in case the class specified is wrong. The idea is to make it more noticeable if someone accidentally changes the class name, or if font-awesome breaks compatibility in a future release.

PR #5007, opened 2018-08-07
Anisharoofi adminsapi v1

anisharoofiTW:
Added Admins API endpoint


api/api-admins-config-v1/config/cipher.aes
```diff
@@ -0,0 +1 @@
+0e8e9da34b8c22ed6e0399b6e95670e2
```
GaneshSPatil: Do we need to commit this file?

PR #5012, opened 2018-08-07
Personalized Views Filtering by Pipeline State

ibnc:
Allows for personalized views to only show pipelines that match the selected state(s). For example, you can have a tab that only shows building pipelines, and another that shows pipelines that have failed.

Screenshot of the ui component:
screen shot 2018-08-07 at 1 04 34 pm

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
@@ -18,7 +18,8 @@ package com.thoughtworks.go.server.domain.user;
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.util.*;
+import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import static com.thoughtworks.go.server.domain.user.DashboardFilter.*;
@@ -54,10 +55,10 @@ class FilterValidator {
     }
 
     private static void validateState(Set<String> state) {
-        if (state != null) {
-            if (!VALID_STATES.containsAll(state))
-                throw new FilterValidationException(MSG_INVALID_STATES);
-        }
+        if (state == null)
+            throw new FilterValidationException("Filter state should never be NULL");
+        if (!state.isEmpty() && !VALID_STATES.containsAll(state))
+            throw new FilterValidationException(MSG_INVALID_STATES);
     }
 
     static void validateNameFormat(String name) {
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
@@ -18,10 +18,10 @@ package com.thoughtworks.go.server.domain.user;
 
 import org.apache.commons.lang3.StringUtils;
 
-import java.util.Map;
+import java.util.*;
 import java.util.regex.Pattern;
 
-import static com.thoughtworks.go.server.domain.user.DashboardFilter.DEFAULT_NAME;
+import static com.thoughtworks.go.server.domain.user.DashboardFilter.*;
 
 class FilterValidator {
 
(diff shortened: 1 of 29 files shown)


marques-work:
@ibnc I would still encourage doing this on the front end. it would likely be simpler, and have preferable behavior with respect to caching on both the frontend and backend. The only backend pieces needed might be serialization and validation before saving to DB, and a data migration.

ibnc:
@marques-work I generally agree, but I think the state filtering should be where the pipeline list filtering is located. Shifting the filtering behavior to the front end should probably be a separate PR

marques-work:
@ibnc I just wouldn't want to do the work twice - may as well put new stuff in the front end rather than putting it server side and then undoing it to move it to the front. That's what I'm thinking. After the environment work, it's probably what I'd pick up next.

server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
```diff
@@ -49,6 +54,18 @@ static void validateFilter(Map<String, DashboardFilter> current, DashboardFilter
         validateNamePresent(name);
         validateNameFormat(name);
         validateNameIsUnique(current, name);
+        validateState(filter.state());
+    }
+
+    private static void validateState(List<CaseInsensitiveString> state) {
+        if (state != null) {
+            if (state.size() >= 3)
```
marques-work: It might be better to use a `Set<String>` here instead, so there's no need to check `size()`, which feels like an anti-pattern; we shouldn't care about size, only about the validity of the values. Also, as the states are constants, no need to use `CaseInsensitiveString`.

ibnc:
I'm unsure why the Jasmine tests are failing. They work fine locally, and on the Windows build. Regardless, this is ready to merge. @GaneshSPatil @ketan @arvindsv Could one of y'all take a look?

akshaydewan:
Hi @ibnc, the jasmine tests seem to be passing in the PR build from what I can see.
If there's anything else I need to look at, do let me know.

ibnc:
@akshaydewan Nope. This PR is mergeable as far as I'm concerned

akshaydewan:
@marques-work since you've reviewed this PR, can you take a look at the latest changes (if you haven't already done that) and merge it?

One thing I see is that the styles for the view are a bit off, and we don't need that comma between the `Failed` and `Building` checkboxes. That can be fixed later, but IMO, it's a good idea to be in a release-ready state on master. We can add a toggle if we foresee breaking changes.

akshaydewan:
Why did we do a data migration for changing the `states` on the filters? Could we assume it to be absent or null and handle it in the service or UI?

ibnc:
@akshaydewan Totally didn't notice that comma in the UI. Just pushed a commit to remove it

marques-work:
@akshaydewan @arvindsv I understand your concern about the added data migration. My instinct is that it still has some value.

It does provide some overhead on startup, but it should be very small in most cases. If there are manny, many users with many filters each, then it could potentially be noticeable, but I don’t think that is representative of most installations.

As for the value of having a migration, my arguments are the following:

1. This is an actual schema change. The fact that we are using JSON means that the schema can be made to be tolerant to inconsistencies doesn’t mean we shouldn’t try to enforce this.

2. My instinct is that this yields cleaner code in general concept. Here, application logic need not perform extra checks in case a hydrated entity does not have the expected schema. Now in this case, it’s true that the change is relatively minor and it would be simple to null-check on serialization/deserialization as well as on programmatic filter construction. However, I also don’t want to set a precedent for future schemata changes, especially if the changes are more drastic than this.

If there is an issue with the visibility as to what a data migration is doing, let’s address that directly.

Thoughts?

marques-work:
@ibnc it looks like this PR is already included in #5026. I may just merge that one once it’s ready unless you feel otherwise.

ibnc:
@marques-work it might be better to merge this one, then the other one. Just so the history is cleaner/clearer? either way works tho

akshaydewan:
Can we add some log statements to the DB migration to see when it happens and how long it takes?

marques-work:
@akshaydewan we certainly can add timing logging to the migrations. I’ll push something to master.

ibnc:
@akshaydewan They only run when the schema version is bellow the new json schema version. It's in the required method on both M001 and M002. So, they only run once

marques-work:
Merging this PR

PR #5017, opened 2018-08-09
Remove encryption of secure plugin properties in setConfigAttributes.

varshavaradarajan:
* The encryptSecureProperties method that gets invoked for a full config save will take care of encrypting the secure values of properties. 

No need to perform this operation twice. Especially since the encryptSecureProperties will be called after preprocessing and takes care of params. The one that's been removed as part of this PR did not take care of preprocessing params.

Verified by entering a value for a secure property for a plugin and checking that the config xml stored the encrypted value. 

cc @bdpiparva


ketan:
@GaneshSPatil could you take a look at this?

GaneshSPatil:
PR build passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/3328

PR #5026, opened 2018-08-13
Personalization Filtering Moved to the Frontend

ibnc:
Still a WIP (This PR has commits from https://github.com/gocd/gocd/pull/5012 which hasn't been merged yet)

As the title states


ibnc:
@marques-work still a WIP, but thoughts?

server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_filter.js
```diff
@@ -0,0 +1,62 @@
...
+
+  this.byPipelines = (pipelineName) => {
+    const conditions = (this.pipelines !== [] && this.pipelines.includes(pipelineName));
+    if (this.isBlacklist()) {
+      return !conditions;
+    }
+    return conditions;
+  };
+
+  this.byState = (pipeline) => {
```
marques-work: Could we not invert this check? It might be simpler to do something like this:

```
if (stage.isBuilding()) return this.state.includes(“building”);
if (stage.isFailing()) return this.state.includes(“failing”);
```

Then you don’t need to permute your conditions.

server/webapp/WEB-INF/rails.new/webpack/models/dashboard/personalization.js
```diff
@@ -57,6 +61,10 @@ function Personalization(filters, pipelineGroups) {
       filterSet = filters; // only changes local copy when successful
     });
   };
+
+  this.isPipelineVisible = (filterName, pipeline) => {
```
marques-work: This feels like it belongs on DashboardFilter, not Personalization as far as responsibilities go.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/personalization_vm.js
```diff
@@ -98,6 +98,10 @@ function PersonalizationVM(currentView) {
 
     changeListeners.push(fn);
   };
+
+  this.isPipelineVisible = (pipeline) => {
```
marques-work: Again, the isPipelineVisible() probably does not belong here, but rather on the filter itself.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_widget.js.msx
```diff
@@ -81,7 +81,7 @@ const DashboardWidget = {
         <div class="dashboard" key="dashboard-content">
           <PersonalizedViewsWidget vm={personalizeVM} />
           {messageView}
-          {_.map(vnode.attrs.dashboard.getPipelineGroups(), (pipelineGroup) => {
+          {_.map(vnode.attrs.dashboard.getPipelineGroupsForView(personalizeVM), (pipelineGroup) => {
```
marques-work: It doesn’t feel quite right to pass the entire VM here. I would figure we pass an actual DashboardFilter object instead. That would break any weird cross dependencies or circular ones. That said I haven’t sat down to verify the resultant dependency graph, so I’m speaking purely from instinct here. Even if there are no dependency issues, it still feels more correct to pass in a filter rather than an object that will delegate a level or two down. If you feel strongly otherwise, we can talk about it.

server/webapp/WEB-INF/rails.new/webpack/models/dashboard/pipeline_groups.js
```diff
@@ -32,6 +32,16 @@ const PipelineGroup = function (name, path, editPath, canAdminister, pipelines)
       return new PipelineGroup(self.name, self.path, self.editPath, self.canAdminister, filteredPipelines);
     }
   };
+
+  this.applyViewFilter = (personalizeVM, pipes) => {
```
marques-work: See comment below regarding passing a filter directly.

server/webapp/WEB-INF/rails.new/webpack/models/dashboard/personalization.js
ibnc: The only issue I have with this is that Personalization is supposed to be the only interface to DashboardFilters and DashboardFilter. Removing this method from Personalization would mean the PersonalizationVM would be calling methods on the DashboardFilter.... Thoughts?

ibnc: My comment applies to the bellow comments too. Do we want to maintain Personalization as the only interface to interact with DashboardFilter(s), or are we okay with changing that

server/src/main/java/com/thoughtworks/go/server/service/GoDashboardService.java
```diff
@@ -82,8 +82,9 @@ private GoDashboardPipelineGroup dashboardPipelineGroupFor(PipelineConfigs pipel
 
         if (goDashboardPipelineGroup.hasPermissions() && goDashboardPipelineGroup.canBeViewedBy(user.getUsername().toString())) {
             pipelineGroup.accept(pipelineConfig -> {
-                if (filter.isPipelineVisible(pipelineConfig.name())) {
-                    goDashboardPipelineGroup.addPipeline(allPipelines.find(pipelineConfig.getName()));
+                GoDashboardPipeline pipeline = allPipelines.find(pipelineConfig.getName());
```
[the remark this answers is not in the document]
ibnc: @marques-work I actually didn't add this code. From what I can see in the commit history, it is needed. It's to ensure we rely on the xml config, and not the cache

marques-work:
@ibnc I merged #5012 and rebased this PR on top of master

ibnc:
@marques-work I think this PR is done if everything looks good to you. Things had to switch around a little because of upstream changes

marques-work:
I'm merging this PR - jasmine tests on windows failed (flaky test that's plaguing master), but otherwise ok.

PR #5032, opened 2018-08-16
Html and Css refactoring

akshaydewan:
On these pages:
- Elastic Profiles
- Artifact Stores
- Plugins
- Roles
- Authorization Config


server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/agents.scss
```diff
@@ -318,7 +318,7 @@ $arrow-color: #000;
 .agents-table-body {
   $foo: 44px;
 
-  margin-top: 148px - $foo;
+  margin-top: 148px - $foo - 70px;
```
akshaydewan: ¯\_(ツ)_/¯
@naveenbhaskar I made a dirty fix, we might have to take another look at this.

server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/components/_collapse.scss
```diff
@@ -0,0 +1,86 @@
+/*!
```
ketan: Please do not add a bang `!` it gets copied into the generated output.

akshaydewan: Fixed, thanks

ketan: I should add that this is the default behavior with intellij (and/or the copyright plugin we use), and I could not find a toggle to turn it off.

If there's a lint we can add to catch this, perhaps we should consider adding it.

PR #5036, opened 2018-08-17
Fix issue on Safari with checkboxes for personalization editing

kierarad:
- Safari and Chrome handle values for onchange differently
- Using onclick for consistent behavior between browsers

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/personalize_editor_vm.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/personalize_editor_vm.js
@@ -66,11 +66,11 @@ function arr(val) { // ensure value is an array
 // allows a boolean to control the presence of a value in
 // a list within a m.Stream
 function boolToList(model, stream, attr) {
-  model[attr] = function(boolPreviousValue) {
+  model[attr] = function(value) {
     const r = stream();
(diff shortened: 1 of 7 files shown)


kierarad:
> Safari and Chrome handle values for onchange differently

This is actually false. It appeared to be this way due to the redraw in the click handler on the body of the dashboard (that hides the dropdowns). It appeared Chrome and FF were inverting the value but it was actually the redraw that was clobbering the value update (only in these browsers). So the onchange handlers were receiving different values depending on the browser.

The onchange handlers now update the model based on the true "checked" value of the checkboxes because the redraw is now async and no longer interferes.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/personalize_editor_vm.js
```diff
@@ -39,7 +39,7 @@ function PersonalizeEditorVM(opts, pipelinesByGroup) { // opts is usually the cu
 
   this.includeNewPipelines = function (boolPreviousValue) {
     if (!arguments.length) { return inverted(); }
-    type(!boolPreviousValue ? "blacklist" : "whitelist");
+    type(boolPreviousValue ? "blacklist" : "whitelist");
```
marques-work: Should probably rename this to `value`

marques-work:
Fixed arg name. Merging.

PR #5043, opened 2018-08-20
Dashboard Personalization UI tweaks

akshaydewan:
Feedback from @naveenbhaskar was that the personalization tabs should behave like the browser tabs in Firefox. Changes made:
- [x]  Hide the arrows when the tabs fit on the screen
- [x]  Bugfix: The tabs dropdown went out of view when no pipelines were present
- [x]  The "+" button should be next to the tabs, similar to Firefox.
- [x] Prevent shrinking of arrow scroll buttons in the flex layout.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss
@@ -5,6 +5,7 @@ $dashboard-tab-action-button-color: #6b808c;
 $dashboard-tab-height: 50px;
 $dashboard-add-tab-width: 40px;
 $dashboard-tab-list-toggle-width: $dashboard-add-tab-width;
+$dashboard-tabs-padding: 30px;
 $dashboard-tabs-border-color: #e9eeef;
 $dashboard-tabs-border-radius: 0;
 $dashboard-tabs-pager-width: $dashboard-add-tab-width;
@@ -24,7 +25,7 @@ $dashboard-tabs-box-shadow: 0 0 10px rgba(0, 0, 0, 0.21);
   border-radius:    $dashboard-tabs-border-radius;
   line-height:      $dashboard-tab-height;
   height:           $dashboard-tab-height;
-  padding:          0;
+  padding:          0 $dashboard-tabs-padding;
   margin-bottom:    20px;
   box-shadow:       $dashboard-tabs-box-shadow;
 
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss
@@ -1,4 +1,3 @@
-$dashboard-tabs-padding: 30px;
 $dashboard-tab-bg: #fff;
 $dashboard-tab-primary-text-color: #943a9e;
 $dashboard-tab-secondary-text-color: #3c5d73;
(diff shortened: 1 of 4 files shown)


marques-work:
@akshaydewan You may also want to adjust the horizontal padding of the container depending on whether the `paged` class is set. See https://github.com/gocd/gocd/blob/79595c74bf0d200d37cba614ac8dbd681178b15c/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss#L168

marques-work:
@naveenbhaskar keeping the `+` icon next to the last tab until the scroll threshold (a la Firefox) might be very difficult without scripting; not sure if it's worth the complexity.

Also, from a UX perspective, isn't it better to keep such a generic action in a static place to help users build muscle memory?

naveenbhaskar:
@marques-work , 
 I would always prefer to have the plus icon next to the views. But if you guys think that is too much of a coding, then we can fix it on the right side of the bar always. 

@akshaydewan 
The down arrow also has to be hidden, until we have hidden views in the slider.

marques-work:
@naveenbhaskar Let me spike out some restructuring and possible scripting to see if it can be done more easily.

My speculation/concern arose out of the fact that the plus icon is managed by a component that isn't directly aware of the scrolling elements, and is 1 or 2 ancestral levels up from the tab elements. That makes it awkward to position at best, and likely impossible with pure CSS.

I'll spend maybe an hour to see if moving the "add tab" component down a couple of levels causes more harm than help; moving this further down the DOM tree would make it easier to style the way browsers design their tabs, but may introduce subtle complexities elsewhere. I'll let you know.

akshaydewan:
I tried using Flex layout to implement the changes. A couple of things I still need to look at:
- The arrow scroll buttons grow and shrink on window resize. I need to set them to a fixed width.
- The padding on the left side of the tabs is gone. Personally, I would like the padding to shrink as the tabs grow in size, otherwise it seems like wasted space. There are some flex options that can do this, I'll spike it out. Otherwise I might just have to toggle the padding based on the `paged` state.

akshaydewan:
Fixes https://github.com/gocd/gocd/issues/5047

akshaydewan:
Fixed an issue with IE and Edge that was causing the `paged` value to toggle on every `onupdate`.  Details of the issue [here](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11689089/) and [here](https://stackoverflow.com/questions/30900154/workaround-for-issue-with-ie-scrollwidth)

server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/pipeline_configs/_dashboard_view_tabs.scss
```diff
@@ -149,30 +150,21 @@ $dashboard-tabs-box-shadow: 0 0 10px rgba(0, 0, 0, 0.21);
...
   white-space:    nowrap;
 }
 
 .dashboard-tabs-sortable {
-  padding-bottom: 100px;
-  width:          100%;
-  overflow-x:     auto;
-  overflow-y:     hidden;
-  outline:        none;
+  overflow: hidden;
```
marques-work: Removing the padding-bottom and setting overflow-x to hidden disables the native horizontal scroll for hardware that is capable (e.g. Mac laptops and other touch devices that allow horizontal scroll, which forces people to use the nav buttons).

Could we restore the `overflow-x: auto` plus some arbitrary `padding-bottom` (e.g. `100px`) to ensure the horizontal scrollbars are rendered outside of the clipped region?

marques-work:
@akshaydewan for the issue with the buttons shrinking, could we not just set `flex-shrink: 0; flex-grow: 0;`? it's probably good to have static widths for the `<`, `>`, and `+` buttons.

marques-work:
Looks good to me. Merging.

PR #5053, opened 2018-08-21
Default personalization filter should not be special

marques-work:
Currently, the `Default` personalization filter/tab has special treatment. Really though, it's no better than the rest :).

Seriously speaking, the intended changes of this PR are as follows:

- [x] No longer validate that a filter named `Default` must exist
- [x] Instead, validate that at least 1 filter exists, regardless of its name
- [x] De/serialization should not try to normalize the name `default` (enforcing title-case)
- [x] Users can rename and delete this filter, just as with any other filter
- [x] Filter template for adding new filter should be a blank whitelist
- [x] The UI should prevent the last filter from being deleted by disabling the remove button. Add a tooltip indicating such behavior for the last tab.
- [x] The "default" filter (i.e. the one that is active upon first loading the dashboard without specifying a name in the `viewName` param) is the first (read: leftmost) filter.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
@@ -43,7 +43,7 @@ class FilterValidator {
     static final String MSG_INVALID_STATES = "An invalid filter state has been provided. Only " + BUILDING_STATE + " and " + FAILED_STATE + " are supported";
 
     static void validateDefaultIsPresent(Map<String, DashboardFilter> current) {
-        if (!current.containsKey(DEFAULT_NAME.toLowerCase())) throw new FilterValidationException(MSG_NO_DEFAULT_FILTER);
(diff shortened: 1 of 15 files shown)


ibnc:
@marques-work I'd think we just return an error to the delete modal if it's the last.

marques-work:
@ibnc changed the behavior criteria to just disable the delete button after talking with @arvindsv. Don't even let users try it.

marques-work:
@akshaydewan It's true that we send the entire json document representing all of a user's filters. Currently, it was done this way to keep the server-side controller "dumb" -- all of the critical logic for building filters was purposefully isolated in the frontend, rather than splitting that logic between the backend and frontend. The pipeline selections backend serves only to persist and validate. That's my "ideal", but I'm not necessarily married to that if this is a major concern and breaking this contract is the only way around it.

That said, there still may be ways to keep this separation of responsibilities while also optimizing `UPDATE` operations (e.g. perform a JSON merge), but I'll give it some thought.

PR #5054, opened 2018-08-22
Restore server side pipeline name filtering

marques-work:


marques-work:
Merging this squashed partial revert. PR tests passed on `127d721`, not going to re-run for the loading indicator.

PR #5056, opened 2018-08-23
Page-header styles and structure update

naveenbhaskar:
Page header structure  and styles update - This will change the headers for plugins, auth config, elastic profile, data sharing, artifact stores and roles pages


ketan:
Run lint task?

Issue #5015, opened 2018-08-09
Collect Extra metrics as part of GoCD Usage Data


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

* Know what plugins are installed on GoCD Instance.
	- We've recently built a lot of plugins. To know if the community finds them useful and are using it on their production instances.
	- To know the popular community-built plugins, which are used by GoCD instances and are not available on GoCD plugins website(if any).

* Usage of elastic agents.
	- To know how many instances prefer dynamic provisioning of agents.
	- To know what percentage of their workflow is moved to used elastic agents.
	- Also, to know the preferred platform for elastic agents from the type of elastic agents.
	  Eg: - ECS if instances use `com.thoughtworks.gocd.elastic-agent.ecs`, 
	        - Kubernetes if instances use `cd.go.contrib.elasticagent.kubernetes`

* Usage of config repo plugins.
	- Since GoCD `v17.8.0`, we've started shipping config repo plugins along with the core. To know if instances are moving towards version-controlled configuration.

JSON Format:
```json
{
    "server_id": "2ab54f99-5fb5-4a5b-850b-ac790aba52b4",
    "message_version": 2,
    "data": {
        "pipeline_count": 167,
        "agent_count": 37,
        "oldest_pipeline_execution_time": 1398304181981,
        "gocd_version": "18.8.0",
        "installed_plugins": [
        	{
        		"id": "cd.go.contrib.elasticagent.kubernetes",
        		"version": "1.0.1-107"
        	},
        	{
        		"id": "com.thoughtworks.gocd.elastic-agent.ecs",
        		"version": "4.1.0"
        	}
        ],
        "job_count": 200,
        "config_repo_pipeline_count": 10,
        "elastic_agent_jobs": [
        	{
        		"plugin_id":  "cd.go.contrib.elasticagent.kubernetes",
        		"job_count": 10
        	},
        	{
        		"plugin_id":  "com.thoughtworks.gocd.elastic-agent.ecs",
        		"job_count": 100
        	}
        ]
    }
}
```

**NOTE:** We are not collecting any private or personal information and we will always be transparent about what is being shared.

ketan:
Looks reasonable to me. In addition to elastic agents, are there metrics we could collect about usage of other plugins — package repo plugins, authorization, scm plugins?

GaneshSPatil:
@ketan -- Not that I've thought about. Will put some more thoughts and update the issue.

[...]

PR #5057, opened 2018-08-23
Usage data api v3

GaneshSPatil:
https://github.com/gocd/gocd/issues/5015#issuecomment-414952746


GaneshSPatil:
Closing this PR as we don't need seperate toggles for usagedata as of now.
Reference: https://github.com/gocd/gocd/issues/5015#issuecomment-415301390

PR #5058, opened 2018-08-23
Do not pause pipeline oncreate via api

GaneshSPatil:


GaneshSPatil:
PR Build Passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/3332

PR #5060, opened 2018-08-23
Add drilldown handler for vsm analytics charts

GaneshSPatil:


server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_analytics.js
```diff
@@ -88,14 +113,14 @@
     });
   };
 
-  var VSMAnalytics = function VSMAnalytics(data, graphRenderer, showAnalyticsPath, analyticsPanel, analyticsButton) {
+  var VSMAnalytics = function VSMAnalytics(data, graphRenderer, vsmAnalyticsChart, analyticsPanel, analyticsButton) {
```
marques-work: No need to declare a `var` once you've declared a named function.

```
function VSMAnalytics(...) {
  ...
}
```

Will do just fine :)

server/webapp/WEB-INF/rails.new/app/assets/javascripts/vsm_analytics.js
```diff
@@ -51,25 +51,50 @@
...
+          handler(r.data, null);
+        });
+      }
+    };
+
+    PluginEndpointRequestHandler.defineFetchAnalyticsHandler(models);
     PluginEndpoint.ensure("v1");
+  }
+
+  var showAnalytics = function showAnalytics(options) {
```
marques-work: No need to declare a `var` once you've declared a named `function`.

```
function showAnalytics(options) {
  ...
}
```

Is sufficient.

Issue #4997, opened 2018-08-03
Using SSO proxy with basic auth causes CSRF errors in the GUI


sandro-h:
##### Issue Type

- Bug Report

##### Summary

We have a reverse proxy in front of GoCD (and other apps) that provides Single Sign-On. In the case of GoCD, the proxy does this by simply appending a basic auth Authorization header to the requests forwarded to the GoCD server.

In previous versions this worked fine, but since upgrading from 18.3 to 18.6 we are experiencing issues when performing changes in the GUI.
We receive "The change you wanted was rejected." errors when editing the config xml or any pipeline, for example.
The server log (see below) shows that the CSRF token cannot be validated.

After some digging, I'm fairly certain this is due to a change in session management which will recreate the session when the Authorization header is set, even if there is still a valid ongoing session (see Possible Fix section).

###### Basic environment details

* Go Version: 18.6.0 (6883-1013a67ccf0cda9c87e6361286708b584be42338)
* JAVA Version: 1.8.0_111
* OS: Linux 3.10.0-514.6.1.el7.x86_64
* Browser vendor and version (if relevant): Firefox 61, Chrome 67 

##### Steps to Reproduce

In lieu of a full reverse proxy setup, we can reproduce the behavior of GoCD resetting the session using curl:

```
# Create a session using basic auth and store the cookie
curl "<host>:8154/go/pipelines" -k -v -o /dev/null -c cookies.txt -u $USER 

# Use the session cookie WITHOUT basic auth
curl "<host>:8154/go/pipelines" -k -v -o /dev/null -b cookies.txt
-> the session is maintained

# Use the session cookie AND basic auth
curl "<host>:8154/go/pipelines" -k -v -o /dev/null -b cookies.txt -u $USER
```

##### Expected Results

The session from the cookie is maintained even if the Authorization header is set.

##### Actual Results

The server creates a new session and returns a Set-Cookie header with a new session cookie.
This invalidates the CSRF token stored in the old session.

##### Possible Fix

I'm fairly certain this change in behavior is due to this change in the AbstractBasicAuthenticationFilter: https://github.com/gocd/gocd/commit/bf001a4fbe0c93cbdb359fc8bc37dd36a550c7c8#diff-a4fb18faf26d223eb093b3cf955a474b

I don't know if there was a good reason to remove the ```if (SessionUtils.hasAuthenticationToken(request))``` check, though.

##### Log snippets

```
[...]

PR #5067, opened 2018-08-27
Do not authenticate user if request is previously authenticated

bdpiparva:


arvindsv:
@bdpiparva Is this for #4997?

bdpiparva:
@arvindsv - Yes, PR is to fix issue #4997. It is still in WIP. I will update the issue once I finish it.

PR #5077, opened 2018-08-27
Tristate checkboxes for Personalize Modal

ibnc:
Screenshot:

screen shot 2018-08-27 at 12 54 14 pm


marques-work:
LGTM.

FYI: the existing `agents/tri_state_checkbox` component doesn't appear to fit the view model very well here, so I wouldn't advocate for using it in this widget. This change is a small attribute change, and is favorable to changing how these checkboxes work to fit `tri_state_checkbox`.

ibnc:
@marques-work I agree with your assessment of the `tri_state_checkbox` component

Issue #2404, opened 2016-07-08
Pipeline Group API


iamsumanth:
##### Issue Type

- Feature proposal
##### Summary

Add an API to create, read, update, delete on resource `pipeline group`.

#### Request:

```
GET /go/api/admin/pipeline_groups
Accept: application/vnd.go.cd.v1+json
```
#### Response:

```
Content-Type: application/vnd.go.cd.v1+json
{
    "_links": {},
    "_embedded": {
        "groups" : [
            {
                "_links": {}, 
                "name" : "group_name",
                "authorization" : {
                    "view" : {
                        "role" : [],
                        "user" : []
                    },
                    "operate" : {
                        "role" : [],
                        "user" : []
                    },
                    "admins" : {
                        "role" : [],
                        "user" : []
                    }
                },
                "pipelines" : [
                    {
                        "_links": {},
                        "name" : "pipeline_name"
                    }
                ]
            }
        ]
    }       
}
```
#### Request:

```
POST /go/api/admin/pipeline_groups
Accept: application/vnd.go.cd.v1+json
Content-Type: application/json
{
    "name" : "group_name",
    "pipelines" : [
        {
            "name" : "pipeline_name"
        }
    ],
    "authorization" : {
        "view" : {
            "role" : [],
            "user" : []
        },
        "operate" : {
            "role" : [],
            "user" : []
        },
        "admins" : {
            "role" : [],
            "user" : []
        }
    }
}
```
#### Response:

```
{
    "_links": {},
    "_embedded": {
        "name" : "group_name",
        "authorization" : {
            "view" : {
                "role" : [],
                "user" : []
            },
            "operate" : {
                "role" : [],
                "user" : []
            },
            "admins" : {
                "role" : [],
                "user" : []
            }
        },
        "pipelines" : [
            {
                "_links": {},
                "name" : "pipeline_name"
            }
        ]
    }
}
```
#### Request:

```
GET /go/api/admin/pipeline_groups/:group_name
Accept: application/vnd.go.cd.v1+json
```
#### Response:

```
{
    "_links": {},
    "_embedded": {
        "name" : "group_name",
        "authorization" : {
            "view" : {
                "role" : [],
                "user" : []
            },
            "operate" : {
                "role" : [],
                "user" : []
            },
            "admins" : {
                "role" : [],
                "user" : []
            }
        },
        "pipelines" : [
            {
                "_links": {},
                "name" : "pipeline_name"
            }
        ]
    }
}
```
#### Request:

[...]

PR #5078, opened 2018-08-27
Introducing API for managing Pipeline Groups

dtsato:
According to #2404, this PR implements the `#index` and `#show` actions to view all pipeline groups, and view a specific pipeline group. I'd like to get this reviewed before I work on the other actions, since I was able to build this on top of the existing `PipelineConfigsService`, which doesn't currently support creating, updating and destroying groups yet.


GaneshSPatil:
@dtsato -- Thanks for the PR.
can you provide the JSON representation of the Pipeline group entity, that will be used for CREATE, UPDATE and GET API Operations?

dtsato:
@ketan @GaneshSPatil I've just finished the `#update` endpoint, which required me to create a new command and extend the existing `PipelineConfigsService`. Since this is a new codebase for me, can you help review this to ensure I didn't miss anything important? I tested it locally and seems like it works (including the handling of E-Tags).

I'm inclined to call it done for now, since I don't anticipate needing the API to `#create` or `#destroy` yet.

Let me know if you have any feedback or anything else you need from me to have this merged (e.g. where are the developer documentation?)

Thanks!

dtsato:
Documentation PR submitted: https://github.com/gocd/api.go.cd/pull/243

ketan:
I'm reviewing the PR — mostly looks good, have a few nitpicks, and I can fix them while reviewing.

ketan:
@dtsato — I've squashed all your commits, along with some minor changes. You should be able to `git fetch origin && git diff master...origin/master` to see changes. You'll need to rebase your branch before you proceed further.

I'm going to merge this EOD today(or early tomorrow) after we've cut a 18.9 branch.

ketan:
@dtsato — do you have any questions before you can proceed further with the remaining create and delete calls?

ketan:
Since this is a very low risk change (no changes to server modules), I'm merging this for 18.8. But this will remain an "undocumented api" until next release.

dtsato:
Thanks @ketan! I wasn't looking at create/delete because I didn't need, but happy to spend some more time to iron this out. A few questions: for create, is it ok to create an empty group (without any pipelines), or should we create group with pipelines? In that case, my pipeline representation is not enough. Similar for delete: do we cascade delete all pipelines, or only delete if the group is empty?

ketan:
> A few questions: for create, is it ok to create an empty group (without any pipelines), or should we create group with pipelines? In that case, my pipeline representation is not enough.

That's perfectly acceptable, and I believe that's what we should be doing. Creating pipelines as part of the pipeline group API will cause other side-effects (having to bump api when something in the pipeline-config api change)

> Similar for delete: do we cascade delete all pipelines, or only delete if the group is empty?

Delete only if the group is empty, seems reasonable. We should be able to consider a "recurse/force" option later, if needed.

PR #5080, opened 2018-08-28
Validate that at least 1 pipeline is selected in the personalization modal

### LINKED ISSUE #5049: [Personalization View] Disable failed/building filters until a pipeline is selected in pipeline selection view
##### Issue Type
- Bug Report

##### Summary

Using the current pipeline selection filters (failed/building) the user can create a view with no pipelines selected. However, the filters need atleast one pipeline/pipeline-group to filter on the pipeline status.

This is misleading for the end-user and might lead to the creation of empty views

* Go Version: 18.8.0-7286
* JAVA Version: Java8
* OS: MacOS
* Browser vendor and version (if relevant): Chrome

##### Steps to Reproduce

1. Click to create a new personalization tab
2. Click on failed/building in the pipeline selection filters
3. Click on none in the pipeline selection filters
4. Save the tab
5. This tab will always be empty as the user is not prompted for selecting atleast one pipeline

##### Expected Results
either disable the filter until a pipeline is selected or make the filter dynamic to show all pipelines in the selected state ( until atleast one is selected)

##### Actual Results
View with no pipelines is created

marques-work:
Disables the Save button and sets a title attribute with the current error.
Need to create a tooltip from this and display it still.

Fixes #5049.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/filter_validations.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/filter_validations.js
@@ -20,9 +20,9 @@ const Validatable = require("models/mixins/validatable_mixin");
 const PADDED_SPACES = /^(?! ).*[\S]$/; // deny leading/trailing spaces
 const ASCII_PR_CHAR = /^[\x20-\x7E]+$/; // ASCII printable characters only (codes 32 - 126)
 
-const MSG_PAD_SPACE = "Filter name must not have leading or trailing whitespace";
-const MSG_BAD_CHARS = "Filter name is only allowed to contain letters, numbers, spaces, and punctuation marks";
-const MSG_DUPE_NAME = "Another filter with this name already exists";
+const MSG_PAD_SPACE = "View name must not have leading or trailing whitespace";
+const MSG_BAD_CHARS = "View name is only allowed to contain letters, numbers, spaces, and punctuation marks";
+const MSG_DUPE_NAME = "Another view with this name already exists";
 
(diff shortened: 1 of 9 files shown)


GaneshSPatil:
@marques-work -- The PR LGTM, didn't find any jasmine-specs for the same though. Is it ok if we write some `FilterValidations` model specs?

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/models/filter_validations.js
```diff
@@ -24,15 +24,33 @@ const MSG_PAD_SPACE = "Filter name must not have leading or trailing whitespace"
 const MSG_BAD_CHARS = "Filter name is only allowed to contain letters, numbers, spaces, and punctuation marks";
 const MSG_DUPE_NAME = "Another filter with this name already exists";
```
GaneshSPatil: Just a note, all the messages should change to use the word `view` instead of `filter`. does not need to happen as part of this PR though

marques-work:
@GaneshSPatil I think adding tests is a good idea. I thought we had some but maybe I'm thinking of the backend Filter tests that perform validations. I'll add this.

I may as well change the verbiage from "filter" to "view" as well.

ankitsri11:
@GaneshSPatil Hi there, I added tests for `FilterValidations` as well as `PipelineListVM.hasAnySelections()`. I also changed the validation messages to say "view". Let us know if there's any other outstanding issues.

GaneshSPatil:
LGTM

marques-work:
@bdpiparva I will rebase it

PR #5081, opened 2018-08-28
Minor refactoring around timer schedulers

ketan:


server/src/main/java/com/thoughtworks/go/server/service/TimerScheduler.java
```diff
@@ -54,32 +52,25 @@
...
-            quartzScheduler.start();
-            scheduleAllJobs(goConfigService.getAllPipelineConfigs());
-            goConfigService.register(this);
-            goConfigService.register(pipelineConfigChangedListener());
-        } catch (SchedulerException e) {
-            showGlobalError("Failed to initialize timer", e);
-        }
+        scheduleAllJobs(goConfigService.getAllPipelineConfigs());
+        goConfigService.register(this);
+        goConfigService.register(pipelineConfigChangedListener());
```
ketan: Just moved this initialization in the spring configuration, so what we inject now is a scheduler, and not a factory.

Issue #5095, opened 2018-08-30
Current Material is not highlighted in VSM


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary
When looking at the VSM for a particular entity (pipeline or material), the entity should be highlighted to indicate that the current VSM is rendered in the context of the pipeline or material.

Material tile is not highlighted, when the given VSM is rendered for a material.
**Note:** Works for pipeline though.

##### Expected Results
The material should be highlighted for which the VSM is rendered.

##### Actual Results
No material is highlighted.

GaneshSPatil:
This issue has a side-effect, which causes the source material not being highlighted in the VSM mode.

Required for release: `v18.8.0`

rajiesh:
Verified on 18.8.0 (7433-4bf750d09bf4dba76a2b7ce0d72e88ecdfdbd96a)

PR #5096, opened 2018-08-30
Highlight material when the VSM is rendered for a specific material (#5095)

GaneshSPatil:


GaneshSPatil:
Before fix:
screen shot 2018-08-30 at 12 22 52 pm

After fix:
screen shot 2018-08-30 at 12 22 33 pm

PR #5106, opened 2018-08-31
[Analytics Plugin] Hide multiple pipeline instances from the VSM pipeline tile in analytics mode

GaneshSPatil:


GaneshSPatil:
Jasmine Passed: https://build.gocd.org/go/pipelines/windows-jasmine/14/defaultStage/1

PR #5107, opened 2018-08-31
Fix config repo cloning

### LINKED ISSUE #5013: Unable to access git config repo over api
##### Issue Type

- Bug Report

##### Summary

Git clone of `go/api/config-repository.git` fails with a 404 error. It seems the error is because of a missing whitelist in `SparkPreFilter`.

```
git clone https://build.gocd.org/go/api/config-repository.git
Cloning into 'config-repository'...
fatal: repository 'https://build.gocd.org/go/api/config-repository.git/' not found
```

ketan:
This broke during 18.6, likely as part of the spring upgrade

Fixes #5013

### DIFF
--- a/server/webapp/WEB-INF/web.xml
+++ b/server/webapp/WEB-INF/web.xml
@@ -69,6 +69,11 @@
   </filter-mapping>
 
   <filter-mapping>
+    <filter-name>ConfigGitRepositoryFilter</filter-name>
+    <url-pattern>/api/config-repository.git/*</url-pattern>
+  </filter-mapping>
+
+  <filter-mapping>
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/web.xml
```diff
@@ -68,6 +68,11 @@
     <url-pattern>/*</url-pattern>
   </filter-mapping>
 
+  <filter-mapping>
```
ketan: Had to rejig the order of filters, to avoid this url from being processed by url-rewrite filter. This filter was sending the request over to rails.

arvindsv:
Looks good.

PR #5109, opened 2018-08-31
Remove animations completely for stacking modals 

kierarad:
The animations as part of the modal for Personalization were causing a "bounce" effect. This removes all the animations completely. 

I tried to remove just the animation on the modal open (which was experiencing the bounce) but that caused the modal to flicker when reopened (after cancelling).

I think effort would be better spent either fixing the bounce than trying to fix the flicker issue. We can revisit adding the animations back in the future.


marques-work:
Closing this as this was fixed in d693b19b54bdccf586826b451963fcb6e9c2901c

PR #5112, opened 2018-09-01
Add tooltips for new, building, and failed on personalization editor

kierarad:
@marques-work will you take a look at this when you have a minute and help see it through the build while I'm away?

I opted for click over hover due to the link in the first tooltip. I changed the positioning so the tooltip would overflow the modal.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_schmodal.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_schmodal.scss
@@ -122,7 +122,6 @@ $overlay-border-radius: 3px;
 
 .overlay-content {
   background: #fff;
-  position:   relative;
   padding:    20px 30px;
   min-height: $overlay-height;
   max-height: 730px;
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_schmodal.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_schmodal.scss
@@ -122,6 +122,7 @@ $overlay-border-radius: 3px;
 
 .overlay-content {
   background: #fff;
+  position:   relative;
   padding:    20px 30px;
   min-height: $overlay-height;
   max-height: 730px;
--- a/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
+++ b/server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
@@ -61,14 +61,15 @@ const BlanketSelector = {
   }
 };
 
-const StageStateCriteria = { //eslint-disable-line no-unused-vars
+const StageStateCriteria = {
   view(vnode) {
     const vm = vnode.attrs.vm;
+
     return <div class="stage-state-selector">
       <f.checkbox name="state" model={vm} attrName="failing" label="Failed" />
-      <ToolTip tooltipText="Show only pipelines whose last run stage has failed, among pipelines selected for this view"/>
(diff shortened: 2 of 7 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
```diff
@@ -121,6 +123,17 @@ const AlertContainer = {
   }
 };
 
+const ToolTip = {
+  view(vnode) {
+    return <span class="tooltip-outer-container">
+      <span class="tooltip-inner-container">
+        <i class="tooltip-hint" onclick={ (e) => { e.stopPropogation; e.target.nextSibling.classList.toggle("show"); }}/>
```
marques-work: You probably mean `e.stopPropagation()` here.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
```diff
@@ -121,6 +123,17 @@ const AlertContainer = {
...
+        <span class="tooltip" role="tooltip">{vnode.attrs.tooltipText} <a href={vnode.attrs.linkUrl} target="_blank">{vnode.attrs.linkText}</a></span>
```
marques-work: It would probably be a better pattern to just allow the content to pass through via `vnode.children`, rather than parsing special attrs.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
ketan: It's usually considered a risky thing to manipulate DOM outside of mithril, not to mention causes repaints outside of the mithril framework.

The right way to do this would be to remember the "toggle state" in the controller:

```
const Tooltip = {
  oninit(vnode) {
    vnode.state.showTooltip = Stream(false)
    vnode.state.toggleTooltip = function() {vnode.state.showTooltip(!(vnode.state.showTooltip))}
  },
  view(vnode) {
    <i class='...' onclick={vnode.state.toggleTooltip.bind(vnode.state)} />
    <span class=`tooltip ${vnode.state.showTooltip() ? 'show' : 'hide'}`>...</span>
  }
}
```

server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/shared/_schmodal.scss
```diff
@@ -122,7 +122,6 @@ $overlay-border-radius: 3px;
 
 .overlay-content {
   background: #fff;
-  position:   relative;
```
marques-work: This may have unintended consequences - is this really what you mean? it may be better to keep `relative` in place to establish a sensible offset parent, while being explicit about `overflow: visible`.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/personalization_modal_widget.js.msx
marques-work: Oddly, this is how the tooltips in `buttonWithTooltip` and `linkWithTooltip` work in `form_helper`. But I do agree as a general guideline that mithril likes to manage this state.

marques-work: But more importantly, this implementation for `onclick` means that it's possible to open more than one tooltip, which is strange. Also, clicking outside will not clear the tooltip. Clicking the link needs to clear the tolip as well. `:hover` is much simpler, but I get why you needed `onclick`.

PR #5131, opened 2018-09-05
Add a class to the checkbox label for styling

akshaydewan:
As discussed with @naveenbhaskar ,  we want to use classes in the CSS rather than binding to specific tags


marques-work:
@akshaydewan I think this is fine - later on, we should probably make this interface the same across form_helper functions. Doesn't have to be part of this PR.

marques-work:
I'll push through build and merge when green

PR #5135, opened 2018-09-05
Dashboard Groups tooltips refactoring

akshaydewan:
This started off with just improving the text on the tooltips, but ended up with a lot of `switch-case` statements in the view. Instead, I made the Group view model polymorphic.

When the Edit button on the Group is in a disabled state, it shows a custom tooltip with the reason for being disabled. If it also has a `title` attribute in this state, it overlaps the custom tooltip. But we do want the aria-label to be present for accessibility. That's what makes the logic a bit messy.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_groups.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_groups.js
@@ -132,6 +132,7 @@ DashboardGroups.fromEnvironmentsJSON = (json) => {
   return new DashboardGroups(_.map(json, (group) => new Environment(group)));
 };
 
+DashboardGroups.PipelineGroup = PipelineGroup;
 DashboardGroups.Environment = Environment;
 
 module.exports = DashboardGroups;
--- a/server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_groups.js
+++ b/server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_groups.js
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-const _ = require('lodash');
+const _      = require('lodash');
+const Routes = require("gen/js-routes");
 
 class Group {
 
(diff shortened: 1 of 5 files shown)


server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_groups_widget.js.msx
```diff
@@ -1,54 +1,18 @@
...
-    const tt = vm.canAdminister ? "" : "You do not have permission to edit.";
-    const paths = routeForScheme(scheme, vm.name);
-    const accessibleLabel = labelForScheme(scheme, vm.name);
+    const vm    = vnode.attrs.vm;
+    const paths = vm.routes();
 
     return <div class="dashboard-group_title">
-      <f.link disabled={!vm.canAdminister} href={paths.show}  class="pipeline_group_name" aria-label={accessibleLabel}>{vm.name || "Pipelines not in any Environment"}</f.link>
-      {vm.name && <f.link class="edit_config dashboard-group_edit-config" disabled={!vm.canAdminister} href={paths.edit} tooltipText={tt} aria-label={`Edit ${accessibleLabel}`} />}
+      <f.link disabled={!vm.canAdminister} href={paths.show} class="pipeline_group_name"
```
marques-work: I know you didn't add the class `pipeline_group_name` - but to be consistent, we should make rename the class to the more generic `dashboard_group_name` as it encompasses both environments and pipeline groups.

server/webapp/WEB-INF/rails.new/webpack/models/dashboard/dashboard_groups.js
```diff
@@ -46,10 +124,14 @@ function DashboardGroups(groups) {
...
-  return new DashboardGroups(_.map(json, (group) => new Group(group)));
+DashboardGroups.fromPipelineGroupsJSON = (json) => {
+  return new DashboardGroups(_.map(json, (group) => new PipelineGroup(group)));
+};
+
+DashboardGroups.fromEnvironmentsJSON = (json) => {
+  return new DashboardGroups(_.map(json, (group) => new Environment(group)));
 };
 
-DashboardGroups.Group = Group;
```
marques-work: Perhaps for symmetry, set `DashboardGroups.PipelineGroup` as well.

server/webapp/WEB-INF/rails.new/webpack/views/dashboard/dashboard_groups_widget.js.msx
arvindsv: @marques-work Does that affect functional specs?

marques-work:
Build finally went through. Merging.

Issue #775, opened 2014-12-12
API to trigger stage and job


jyotisingh:
If there are multiple stages in pipelines and if one if the later stages is manual, then the only way to trigger them right now is through Go dashboard. This API should be capable of triggering a stage if the previous stage has finished running (pass or fail).

moritz:
I'm very interested in such an API. The use case is so I can build a chat bot that triggers a production release of some software by triggering a stage with manual approval.

hveiga:
Is there any update on this?

kapowie:
Please prioritise this. We also want to integrate Go with an external system. We use JIRA for our release authorisations and we want to automate the manual approvals from our JIRA workflow changes.

ketan:
This feature is currently not on the roadmap. However, if someone is willing to contribute a patch (the gocd team can provide pointers) we are more than happy to merge it into an upcoming release.

samcgardner:
Alright then. @ketan if you can tell me what the outstanding issues with /go/run/<pipeline> are, I'm happy to sling in a PR to fix 'em

moritz:
@samcgardner the only issue with /go/run/<pipeline> that I'm aware of is the lack of API documentation. The endpoint itself seems to work (but I'm not on the go team, so take this with a grain of salt).

dcworldwide:
Any updates on this guys? Seems like a very useful addition. We need it to resume a stage that was setup as manual so that we could have a manual QA process take place. We want to call the api from another app so that non-goCD admins can resume the pipeline based on their business role.

cfontes:
I am also really interested in it...

I can try to implement it with my terrible Go skills, but no promises.... Can someone experienced in the code give me a couple pointers so I don't spend a week trying to get around ?

Cheers

bdpiparva:
@cfontes -- This is on our roadmap and we plan to get this out in the next couple of releases.

ketan:
@cfontes — I'd think there's a few things you may be able to help with before dipping your toes into code:

- hash out what specific APIs you'd like to see (trigger stage, trigger specific job, trigger all jobs, trigger failing jobs, trigger cancelled jobs,...)
- hash out how the various APIs should look like (http verbs, payloads, responses). See https://api.gocd.org for some inspiration.
[...]

PR #5140, opened 2018-09-06
Add a Stage Operations API

marques-work:
Currently, this only allows a user to trigger stages (i.e. manual approval or rerun).

Addresses #775


api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1Delegate.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private String run(Request req, Response res) {
+        String pipelineName = req.params("pipeline_name");
+        String pipelineCounter = req.params("pipeline_counter");
+        String stageName = req.params("stage_name");
+
+        try {
+            scheduleService.rerunStage(pipelineName, pipelineCounter, stageName);
+        } catch (Exception e) {
+            HaltApiResponses.haltBecauseOfReason(e.getMessage());
```
akshaydewan: Catching all exceptions and sending the message in the response seems wrong. But it looks like the service throws a bunch of RuntimeExceptions in scenarios that it checks for. Does anyone have any suggestions if this can be avoided? @ketan ?

api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1Delegate.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.http.HttpStatus;
+import spark.Request;
+import spark.Response;
+
+import java.util.Arrays;
+
+import static java.lang.String.format;
+import static spark.Spark.*;
+
+public class StageOperationsControllerV1Delegate extends ApiController {
```
akshaydewan: I think we don't need a separate Delegate class now that we've upgraded to Spring 4. Originally, the problem was that a Spring bean wouldn't load if we used lambdas in the class, and would crash at server startup with a cryptic error message.
On my local machine, I tried moving the delegate methods into the Spring class itself, and I was able to run the server and hit the API.

api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1Delegate.java
marques-work: @akshaydewan likely you can do one of two things without making big changes: 

1. Test for the message formats. Yuck.
2. Make specific exceptions that extend RuntimeException and test for those.

api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1Delegate.java
marques-work: @akshaydewan alright, I can move them into the controller and run it through the build. If all passes, that’d be great.

api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/spring/StageOperationsControllerV1.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private String run(Request req, Response res) {
+        String pipelineName = req.params("pipeline_name");
+        String pipelineCounter = req.params("pipeline_counter");
+        String stageName = req.params("stage_name");
+
+        try {
+            scheduleService.rerunStage(pipelineName, pipelineCounter, stageName);
+        } catch (Exception e) {
```
marques-work: Moving this comment from @akshaydewan as the delegate class was removed:

Catching all exceptions and sending the message in the response seems wrong. But it looks like the service throws a bunch of RuntimeExceptions in scenarios that it checks for. Does anyone have any suggestions if this can be avoided? @ketan ?

marques-work: Copying my reply as well:

@akshaydewan likely you can do one of two things without making big changes:

1. Test for the message formats. Yuck.
2. Make specific exceptions that extend RuntimeException and test for those.

akshaydewan: We should fix it the right way (point 2 above), but that should be a separate task.

ketan: > Catching all exceptions and sending the message in the response seems wrong. But it looks like the service throws a bunch of RuntimeExceptions in scenarios that it checks for. Does anyone have any suggestions if this can be avoided? @ketan ?

It's possible to provide an overload for `rerunStage` (or rename `internalRerun`) to inject a custom `ResultUpdatingErrorHandler` impl?

marques-work: @ketan I'll try this out

marques-work:
@ketan Could you take a quick glance again? I believe I've addressed the concerns you had.

ketan:
@marques-work could you rebase before I take a look at this? beware that I've pushed a few commits on top of yours so you'll need to get those changes before any force pushes.

marques-work:
@ketan ok I've rebased and pushed (and fetched your changes as well). Could you take a quick look? Thanks!

server/src/main/java/com/thoughtworks/go/server/service/ScheduleService.java
```diff
@@ -251,8 +241,43 @@ private boolean canSchedule(PipelineConfig pipelineConfig) {
...
+                result.accepted(String.format("Request to schedule stage %s accepted", identifier), "", healthStateType);
+            }
+        } catch (RuntimeException e) {
+            String message = String.format(
+                    "Stage rerun request for stage [%s] could not be completed " +
+                            "because of an unexpected failure. Cause: %s", identifier,
+                    e.getMessage()
+            );
+            LOGGER.error(message, e);
+            result.internalServerError(message, healthStateType);
```
GaneshSPatil: We're catching all the RuntimeExceptions and returning `500`. Is there something we can do about it? I mean rather than throwing/catching runtime exceptions, can we create specific exceptions that might occur while triggering a stage?

GaneshSPatil: Also, didn't see any unit and the integration tests for this code.

ketan: > We're catching all the RuntimeExceptions and returning `500`. Is there something we can do about it? I mean rather than throwing/catching runtime exceptions, can we create specific exceptions that might occur while triggering a stage?

I'm going to apply this patch and merge the PR

```patch
diff --git a/server/src/main/java/com/thoughtworks/go/server/service/ScheduleService.java b/server/src/main/java/com/thoughtworks/go/server/service/ScheduleService.java
index 48b8d12bc..df0cea198 100644
--- a/server/src/main/java/com/thoughtworks/go/server/service/ScheduleService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/ScheduleService.java
@@ -269,13 +269,17 @@ public class ScheduleService {
                 result.accepted(String.format("Request to schedule stage %s accepted", identifier), "", healthStateType);
             }
         } catch (RuntimeException e) {
-            String message = String.format(
-                    "Stage rerun request for stage [%s] could not be completed " +
-                            "because of an unexpected failure. Cause: %s", identifier,
-                    e.getMessage()
-            );
-            LOGGER.error(message, e);
-            result.internalServerError(message, healthStateType);
+            // if the result is successful, but there's still an exception, treat it as a 500
+            // else the result is assumed to contain the right status code and error message
+            if (result.isSuccess()) {
+                String message = String.format(
+                        "Stage rerun request for stage [%s] could not be completed " +
+                                "because of an unexpected failure. Cause: %s", identifier,
+                        e.getMessage()
+                );
+                LOGGER.error(message, e);
+                result.internalServerError(message, healthStateType);
+            }
         }
         return stage;
     }
```

ketan:
I'm merging this — should we additionally consider adding a `rerunFailed` and `rerunSelected` jobs API as well?

We already have the services methods in place and should be quick to implement. Avoids us having to bump this api in the future.

GaneshSPatil:
@ketan @marques-work -- have we documented this API? If not can we do this as part of this release?

maheshp:
@GaneshSPatil as @ketan has mentioned we were considering documenting this API once it supports `rerunFailed` and `rerunSelected` else would require us to bump up the version.

PR #5141, opened 2018-09-06
Run eslintold on old javascript test too

GaneshSPatil:
* IE11 allows ES6 binding feature (const, let) but none of the other ES6 features.
* Using const in the code gives a false impression of ES6 support, Hence, do not use any ES6 features in the old code.

ES6 compatibility map: https://kangax.github.io/compat-table/es6/


GaneshSPatil:
@ketan -- mind merging this PR?
I've verified the functionality, works as expected!

PR #5142, opened 2018-09-06
Add XML migration to create a blank `<backup/>` element with defaults

ketan:


ketan:
This will emit a `<backup emailOnSuccess="true" emailOnFailure="true" />` to retain defaults

PR #5144, opened 2018-09-06
Set search box width only for dashboard

akshaydewan:
The same style class is used in the search modal, which causes it to misalign.

### DIFF
--- a/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
+++ b/server/webapp/WEB-INF/rails.new/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
@@ -188,12 +188,15 @@ body {
   box-shadow:         none;
   border-radius:      3px;
   background:         #fff;
-  width:              350px;
 
   &::placeholder {
     font-size: 13px;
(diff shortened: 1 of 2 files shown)


marques-work:
@akshaydewan I would rather the elements not share a css class at all. IMO, I would have renamed the class that the modal search box uses (e.g. personalization-modal-search). Anything shared could be mixed in or put in explicitly generic classes to avoid accidental misuse.

akshaydewan:
I initially thought of making separate classes for each of the search boxes. The downside is that it could lead to an inconsistent look and feel. For instance, there is a search box used on the Agents page, which has a purple background when it is in focus, while the one on the dashboard has a grey background.

Some of us discussed the issue of inconsistent look-and-feel yesterday, and we thought that making components for everything would be a good way to go forward. In this case, it would mean extracting out a `<SearchBox/>` component and using it everywhere.

Easier said than done, however. We still would have the issue of specifying a width for the component depending on where it is used. And I'm sure we'll discover other issues along the way.

akshaydewan:
Spoke to @marques-work. We're merging this now for the release, but I'm working on extracting out a `<SearchBox/>` for a cleaner fix.

PR #5145, opened 2018-09-06
Call onchange() when a tab is activated

akshaydewan:
ensureCurrentTabIsInView will be triggered indirectly.

@marques-work Can you take a look? I hope this doesn't cause any unintended behaviour.


marques-work:
Sending through build. Will merge upon green.

PR #5146, opened 2018-09-07
Analytics - added support to handle request from plugin to link to vsm_page

maheshp:


GaneshSPatil:
@maheshp -- didn't see the new set of changes to support stage-details-page link support as part of this PR. are we doing it as part of separate PR?

PR #5161, opened 2018-09-11
Disable autocompletion, autocapitalization and spellchecking...

Vyax:
…when searching for pipelines


Vyax:
I'm on 18.5.0. Has the search bar been removed completely in the newer releases?

arvindsv:
@Vyax It has been replaced and should be better! Take a look at https://www.gocd.org/releases/#18-8-0

Vyax:
Okay, I haven't tried it yet.

ketan:
I just checked the newer dashboard, which has this:

```html
<input autocomplete="off" autocapitalize="off" autocorrect="off" spellcheck="false" type="text" placeholder="Search pipelines" data-prop-name="searchText" class="pipeline-search pipeline-search_dashboard">
```

Close this as "FIX_NOT_NEEDED" ;) ?

PR #5168, opened 2018-09-13
Default to port 443 in go agent

kosta:
I had the issue that setting in `/etc/default/go-agent` `GO_SERVER_URL=https://gocdhostname/go` did not work with the error "Failed to get non-loopback local ip address". Using `GO_SERVER_URL=https://gocdhostname:443/go` (note the port number) fixes that issue.

The reason is that if you don't supply a port number, `url.getPort()` returns `-1`, which you obviously cannot connect to, hence that URL was rejected.

This PR defaults to port 443 if no port is given. That is good enough for gocd agent, which can only use https if I observed correctly. Still, this code might want to look at the protocol and default to port 80 if the protocol is "http" for other use cases (and keep -1 for unknown protocols).

Let me know if this change seems acceptable.

I would like to add tests as well but am unsure how to proceed. This function is currently not tested directly. Please advise.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java
+++ b/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java
@@ -179,7 +179,11 @@ public class SystemUtil {
     public static String getClientIp(String serviceUrl) {
         try {
             URL url = new URL(serviceUrl);
-            try (Socket socket = new Socket(url.getHost(), url.getPort())) {
+            int port = url.getPort();
+            if (port == -1) {
+                port = url.getDefaultPort();
+            }
+            try (Socket socket = new Socket(url.getHost(), port)) {
                 return socket.getLocalAddress().getHostAddress();
             }
         } catch (Exception e) {


arvindsv:
@kosta The idea is fine. Can you please sign the CLA?

Also, since this is a generic location (`SystemUtil`), it's safer to use something like this (below) instead of assuming 443:

```diff
diff --git a/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java b/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java
index 1ac4a87d8f..0a57e233a7 100644
--- a/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java
+++ b/base/src/main/java/com/thoughtworks/go/util/SystemUtil.java
@@ -179,7 +179,7 @@ public class SystemUtil {
     public static String getClientIp(String serviceUrl) {
         try {
             URL url = new URL(serviceUrl);
-            try (Socket socket = new Socket(url.getHost(), url.getPort())) {
+            try (Socket socket = new Socket(url.getHost(), url.getPort() == -1 ? url.getDefaultPort() : url.getPort())) {
                 return socket.getLocalAddress().getHostAddress();
             }
         } catch (Exception e) {
```

I looked at the testing of it, and even without this change, anything I write passes. :/ So, we might have to deal with it at a different level, at functional tests or integration tests. I'm happy to accept this without that, at this time.

arvindsv:
Also, thank you so much @kosta for taking the time to look into this and contribute.

kosta:
I'll have to wait until next week until I can sign the CLA. Afterwards, I'll sign it and do the change you suggested.

arvindsv:
@kosta Ok.

kosta:
Ok I signed the CLA and am using `url.getDefaultPort()`. (Which is much better than defaulting to a hard-coded port)

I avoided using the trinary operator for better readability and to avoid two function calls (that are probably inlined anyway and not in a hot path, but it didn't smell right).

Hope this is ok, please let me know if you'd like any other modifications.

arvindsv:
Looks good. Thank you very much, @kosta! Merging.

Issue #5178, opened 2018-09-18
[Personalization View] Group pipelines by drop down has a background CSS attribute on select causing the icon to appear broken


adityasood:
##### Issue Type

- Bug Report

##### Summary

The new_dashboard.scss has an [import of commons.scss](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss#L22). This has a bakground css defined for all select [here](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/shared/_common.scss#L297) . This causes the drop down button to appear broken.

###### Basic environment details

* Go Version:  18.10.0-7490
* JAVA Version: Java10
* OS: MacOs and Windows
* Browser vendor and version (if relevant): Chrome (v:69,68,67), Firefox(v: 62,61)

##### Steps to Reproduce

1.  Start a GoCD server
2. Observe the `Group pipelines by` dropdown

##### Expected Results
The `select` should be overriden in the new_dashboard to not have a `background-position:` attribute

##### Actual Results
The `background-position:` attribute breaks the select drop down button on new dashboard page

![group_pipelines_by_drop_down](https://user-images.githubusercontent.com/17826464/45667139-8c853100-bb35-11e8-8ad2-d7e0a34b7082.gif)

PR #5187, opened 2018-09-20
Fix the groupBy dropdown on the dashboard

akshaydewan:
Fixes: #5178

The look and feel of the dropdown is changed, and extracted it out into a component.

### DIFF
--- a/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/shared/_dropdown.scss
+++ b/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/shared/_dropdown.scss
@@ -14,60 +14,94 @@
  * limitations under the License.
  */
 
+$height: 34px;
+$dropdown-border: #d6e0e2;
+$text-color: #000;
+
 .groupby {
-  position: relative;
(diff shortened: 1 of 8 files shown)


server/webapp/WEB-INF/rails/app/assets/new_stylesheets/shared/_dropdown.scss
```diff
@@ -0,0 +1,107 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+$height: 34px;
+$dropdown-border: #d6e0e2;
+$text-color: #000;
```
ketan: Use different variable names? I fear this may cause something else to break.

akshaydewan:
Removed the `select` style from `_common.scss`. This should hopefully fix all the other broken dropdowns.

PR #5199, opened 2018-09-24
Attempting to fix a few flaky tests

akshaydewan:
- Added a `lastRun` time to the `DefaultPluginJarMonitor` thread to ensure that the test continues on the next run of the monitor thread.

The code changes are only for tests, which is not ideal, but still better than having flaky tests IMO.

~~- Changed the contract of `StreamPumper.pump()` to return a Future~~
~~- Separated out instance creation into a `get` method~~
(Removed the StreamPumper commit, since it will need some more work)

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/domain/builder/pluggableTask/PluggableTaskConsoleTest.java
+++ b/common/src/test/java/com/thoughtworks/go/domain/builder/pluggableTask/PluggableTaskConsoleTest.java
@@ -91,8 +91,7 @@ public class PluggableTaskConsoleTest {
 
         doNothing().when(safeOutputStreamConsumer).stdOutput(anyString());
         console.readOutputOf(in);
-        Thread.sleep(100);// may become flaky!! Fingers crossed
-        verify(safeOutputStreamConsumer, times(7)).stdOutput(anyString());
+        verify(safeOutputStreamConsumer, timeout(10000).times(7)).stdOutput(anyString());
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


plugin-infra/go-plugin-api/src/main/java/com/thoughtworks/go/plugin/api/task/Console.java
```diff
@@ -41,8 +42,9 @@
      * where the "process" object is of type {@link java.lang.Process}.
      *
      * @param in The input stream to read as standard error.
+     * @return A future holding a boolean, which will always be true on completion. You can wait on the Future for testing
      */
-    void readErrorOf(InputStream in);
```
ketan: This change is in the plugin API jar, it may break plugins

util/src/main/java/com/thoughtworks/go/util/command/StreamPumper.java
```diff
@@ -89,10 +94,9 @@ public void readToEnd() {
         }
     }
 
-    public static StreamPumper pump(InputStream stream, StreamConsumer streamConsumer, String prefix, String encoding) {
-        StreamPumper pumper = new StreamPumper(stream, streamConsumer, prefix, encoding);
-        new Thread(pumper).start();
-        return pumper;
+    public Future<Boolean> pump() {
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        return executorService.submit(this);
```
ketan: Who shuts this down? The earlier thread would shutdown after the `StreamPumper` runnable returned from the `run` method.

PR #5200, opened 2018-09-24
Header panel component

GaneshSPatil:


GaneshSPatil:
@naveenbhaskar @akshaydewan  -- can you review this PR to verify if the component structure & code is written appropriately?

server/webapp/WEB-INF/rails/webpack/views/components/header_panel/css/index.scss
```diff
@@ -0,0 +1,48 @@
+/*!
```
ketan: Please no `/*!` that causes the comment to be included in the output CSS. Google for "loud css/scss comment". Preferably add a lint for it :)

server/webapp/WEB-INF/rails/webpack/views/components/header_panel/index.js.msx
```diff
@@ -0,0 +1,41 @@
...
+const m     = require('mithril');
+const _     = require('lodash');
+const style = require('./css/index.scss');
+
+const HeaderPanel = {
+  view(vnode) {
+    let buttons;
+    if (!_.isEmpty(vnode.attrs.buttons)) {
+      buttons = (
+        <div className={style["page-actions"] || "page-actions"}>
```
ketan: Should we consider `camelCase` class names now for ease of use? Having to type hyphens seems too much. @naveenbhaskar may have a different opinion about this :)

Issue #5194, opened 2018-09-22
Bug: Fetch external artifact - does not validate


arvindsv:
##### Issue Type

- Bug Report

##### Summary

If you have a fetch external artifact plugin and you've set it up to [validate configuration during fetch artifact](https://plugin-api.gocd.org/current/artifacts/#validate-fetch-artifact-config), it never makes that call, when creating or editing a fetch artifact task from the UI.

###### Basic environment details

* Go Version: 18.9.0
* JAVA Version: 1.8

##### Steps to Reproduce

1. Use this plugin tree: https://github.com/sheroy/docker-registry-artifact-plugin/tree/master
2. Use this tree for GoCD codebase: https://github.com/arvindsv/gocd/tree/allow-artifact-plugins-to-introduce-new-env-vars (or use your own, but have a plugin which validates fetch artifact view - the current, stable docker registry artifact plugin does not).
3. Create a fetch artifact task. Put in a space in the prefix field (of the custom plugin).
4. Save.
5. Edit the task.
6. Save.

##### Expected Results

No validation message shows up.

##### Possible Fix

It's a bug. The fix for update, at least, looks [something like this](https://github.com/arvindsv/gocd/commit/86085544f13c6d044e5cb034a9e1bc14c2cdbcf4#diff-769ff536bbe6d6eeca69644673260d6c), but I haven't been able to get the specs to pass. There were no specs testing this functionality.

##### Any other info

Config to use:
```xml
<?xml version="1.0" encoding="utf-8"?>
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="113">
  <server artifactsdir="artifacts" agentAutoRegisterKey="2b110569-cf92-40ff-af9d-cc89fc6f2d70" webhookSecret="627843e4-b257-469a-84ae-5b6dfffcc506" commandRepositoryLocation="default" serverId="bd42e53a-f5e1-4e07-9600-9df205a5b409" tokenGenerationKey="37ad75a4-0e64-47eb-a831-5f7b4558534c">
    <backup emailOnSuccess="true" emailOnFailure="true" />
  </server>
  <artifactStores>
    <artifactStore id="abc" pluginId="cd.go.artifact.docker.registry">
      <property>
        <key>RegistryURL</key>
        <value>https://index.docker.io/v1/</value>
      </property>
      <property>
        <key>Username</key>
        <value>your-username</value>
[...]

PR #5202, opened 2018-09-25
Fixed bug: Validate fetch artifact config before save.

bdpiparva:
This is to fix https://github.com/gocd/gocd/issues/5194


arvindsv:
Is there a spec for FetchPluggableArtifactTask? Something [like this](https://github.com/arvindsv/gocd/commit/86085544f13c6d044e5cb034a9e1bc14c2cdbcf4#diff-6f1b719b7d9aa871fd4669522720b541R37)?

bdpiparva:
Added e9ba458839a3a6d71f33a8a1678bbbc3612ebe85 for covering FetchPluggableArtifactTask.

PR #5205, opened 2018-09-25
Package installers with java 10.

ketan:


ketan:
I've removed support for 32 bit installers. See https://stackoverflow.com/questions/46356345/can-java-9-run-on-a-32-bit-os for context.

PR #5208, opened 2018-09-25
SystemEnvironment#configDir() should honor `cruise.config.dir`

marques-work:


marques-work:
won't fix - nuances here around `configDir()` and `getConfigDir()` that I don't entirely understand yet. this isn't important enough to push through.

PR #5216, opened 2018-09-30
Avoid ajax calls for tests

ketan:
* Setup constants file with sane defaults, when the dom is lacking the
  attributes. This is typically the case with tests.
* Create the ajax poller in the `oninit` of components. This avoids
  performing ajax calls on a `require` of the said js file.


ketan:
Assigning to @marques-work since he brought this up a few days ago.

akshaydewan:
Thanks for fixing this. In addition, we'll also try to import only the `spec` js files in `specRoot.js`.

@marques-work I'm going ahead and merging this to fix the builds

PR #5220, opened 2018-10-01
Do not schedule timer triggered pipelines on an inactive server

maheshp:
* Currently on an inactive server MDU, Scheduling etc are turned off.
  All non-GET requests are not allowed. This commit ensures even the
  timer triggered pipelines are not scheduled.


ketan:
My understanding was that this was managed by the [`ModeAwareThreadPoolTaskScheduler`](https://github.com/gocd/gocd/blob/57d544a8797828299e43a1def622df70e23681e8/server/src/main/java/com/thoughtworks/go/server/service/ModeAwareThreadPoolTaskScheduler.java). Did we accidentally mess this up during the spring upgrade?

maheshp:
> My understanding was that this was managed by the ModeAwareThreadPoolTaskScheduler. Did we accidentally mess this up during the spring upgrade?

I don't think this was changed during the Spring upgrade. AFAIK the TimerScheduler was never scheduled using the ModeAwareThreadPoolTaskScheduler as all the jobs are scheduled as part of the `initialize` method.

PR #5228, opened 2018-10-04
Do not run tests on java 9

GaneshSPatil:
* Remove java 9 from supported JVM

### DIFF
--- a/.ci/with-java
+++ b/.ci/with-java
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-# 4, because that's the number of JVMs we test on
-mod="$((${GO_PIPELINE_COUNTER} % 4))"
+# 3, because that's the number of JVMs we test on
+mod="$((${GO_PIPELINE_COUNTER} % 3))"
 
 function use_jdk() {
   source "${HOME}/.jabba/jabba.sh"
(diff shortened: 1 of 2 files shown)


.ci/with-java
```diff
@@ -13,10 +13,8 @@ if [ "$mod" = "0" ]; then
   # use the system jvm
   true
 elif [ "$mod" = "1" ]; then
-  use_jdk "openjdk@1.9.0-4" "tgz+https://nexus.gocd.io/repository/s3-mirrors/local/jdk/openjdk-9.0.4_linux-x64_bin.tar.gz"
-elif [ "$mod" = "2" ]; then
   use_jdk "openjdk@1.10.0-2" "tgz+https://nexus.gocd.io/repository/s3-mirrors/local/jdk/openjdk-10.0.2_linux-x64_bin.tar.gz"
-elif [ "$mod" = "3" ]; then
+elif [ "$mod" = "2" ]; then
   use_jdk "openjdk@1.11.0-28" "tgz+https://nexus.gocd.io/repository/s3-mirrors/local/jdk/openjdk-11-28_linux-x64_bin.tar.gz"
```
bdpiparva: Can we use `jabba use openjdk@1.11.0-28` instead? That will fail the build if java is not installed on container.

GaneshSPatil: See method `use_jdk`.

ketan:
There's also a powershell script that you've missed out on?

GaneshSPatil:
> There's also a powershell script that you've missed out on?

FIxed!

GaneshSPatil:
Do not merge! We should remove jdk9 and add jdk12 to the support metrix

bdpiparva:
@GaneshSPatil - Sure, we can target support for java11 in next release and also give it a try to java12.

PR #5231, opened 2018-10-05
Resolve annoying Sass deprecation messages.

marques-work:
These fixes ought to be safe, but if anyone has issues, LMK.


marques-work:
Spoke to @ketan about merging this after cutting 18.10. Merging this now that 18.10 happened.

PR #5113, opened 2018-09-01
Environment Index Page - Change view to show the environment config for each environment.

varshavaradarajan:
* Add id attribute to environment name on the config index page to fix the link from the dashboard.

@arvindsv , @maheshp , @ketan  - Now that the environments page is a config only page, I am yet to move the environments tab under the admin dropdown as per what we had discussed: https://github.com/gocd/gocd/pull/4157#issuecomment-355881484.  The url now to this page should be `/go/admin/environments` or `/go/environments`? And if its going to be `/go/admin/environments` should we set up redirection from `/go/environments` to `/go/admin/environments`?

Also, because the group by environments is available on the dashboard for view users, this page will be accessible only to super admins. Right? 

@ketan, @maheshp  - I have not deleted `shared/_pipeline.html.erb` because `/go/old_dashboard` is still available at this time. When will it be removed?

PR #5235, opened 2018-10-05
Introduce environments index page to list all the environments.

GaneshSPatil:
* List all the environments in expand/collapsible manner on the environments list page
* Show information about number of pipelines and number of agents
* Retain environment specific show page for the redirection from the dashboard
* Move environment next to pipelines under admin's dropdown
* Add redirects from /go/environments to /go/admin/environments

This PR addresses all the features from: https://github.com/gocd/gocd/pull/5113 PR


GaneshSPatil:
screen shot 2018-10-05 at 5 35 48 pm
screen shot 2018-10-05 at 5 35 51 pm
screen shot 2018-10-05 at 5 36 14 pm

server/webapp/WEB-INF/rails/app/assets/stylesheets/new-theme.scss
```diff
@@ -83,6 +83,58 @@ h4 {
...
+  cursor:        pointer;
+  padding:       10px 50px 10px 30px;
+  margin:        0 15px;
+  background:    #fff;
+  border:        1px solid #ddd !important;
+  border-radius: 5px;
+
+  &:after {
+    content:     "\f105";
+    font-family: 'FontAwesome';
```
akshaydewan: Why not use the `icon-after` mixin?

GaneshSPatil:
Old code.. can we use mixin here too? If so i can change this code

akshaydewan:
> Old code.. can we use mixin here too? If so i can change this code
Ok, I got thrown off my the name `new-theme.scss`

server/webapp/WEB-INF/rails/config/routes.rb
```diff
@@ -187,19 +187,22 @@
     get "templates/:name" => "config_view/templates#show", as: :config_view_templates_show, constraints: {name: TEMPLATE_NAME_FORMAT}
   end
 
-  scope 'environments' do
+  get 'environments', to: redirect('admin/environments'), as: :environment_redirect
```
bdpiparva: Currently `/go/environments` is redirected to => `/go/admin/environments`.

Need to consider redirects for the following routes -

- `go/environments/foo/show` =>`go/admin/environments/foo/show`
- `go/environments/new` => `go/admin/environments/new`
- POST  to `/go/environments/:env_name` => `/go/admin/environments/:env_name`

bdpiparva: This is to make sure that old links for env page will continue to work. We will put deprecation of old routes in the release note.

bdpiparva:
Changed HTML structure and css to make the UI consistent with other pages

screen shot 2018-10-09 at 6 12 46 pm

Issue #4861, opened 2018-06-19
Email-notification filter setting does not acknowledge  "Only if it contains my check-ins" option


adityasood:
##### Issue Type
- Bug Report

##### Summary
While setting email notification filters, if the user checks "Only if it contains my check-ins" option, the filter created still shows "All" in the "Check-ins Matcher" column

###### Basic environment details

* Go Version: : 18.6.0 (6883-1013a67ccf0cda9c87e6361286708b584be42338)
* JAVA Version: Java8
* OS: MacOS (generic zip installer)
* Browser vendor and version (if relevant): Chrome and Firefox

##### Steps to Reproduce

1.  Login to GoCD ( authentication needs to be enabled)
2. Go to user preferences section and create an email-notification filter
3. Enable "Only if it contains my check-ins"

##### Expected Results
The "Check-ins Matcher" column should show "MIne" as the column value.

##### Actual Results
The "Check-ins Matcher" column shows "All" as the column value.

<img width="1110" alt="screen shot 2018-06-19 at 5 30 55 pm" src="https://user-images.githubusercontent.com/17826464/41596403-e15d8d20-73e7-11e8-87a6-802f566b712d.png">

jyotisingh:
For anyone looking at fixing this, the issue is with the javascript which is invoking the API with incorrect payload. 
Posts the payload
```
{
  "pipeline": "[Any Pipeline]",
  "stage": "[Any Stage]",
  "event": "Fixed",
  "myCheckin": "on"
}
```
The [API](https://api.gocd.org/current/#update-current-user) expects a property by the name `match_commits` with a `true|false` value instead of `myCheckin` with `on|off` value.

pavelkostyurin:
So, is there a way to make it work?

adityasood:
Verified on 18.10.0-7703.

PR #5238, opened 2018-10-08
Fixing #4861

rvrn22:
Changed the "myCheckin": "on" in the API call to match_commits: true|false as mentioned by jyotisingh


ketan:
@adityasood — could you take a look and test this so someone can merge?

adityasood:
@ketan verified. Can be merged.

Issue #4836, opened 2018-06-08
Error scheduling pipelines on change of material


maheshp:
##### Issue Type

- Bug Report

##### Summary

There could be instances/reasons wherein a pipeline scm material could be renamed by adding or removing a `/` to the material URL. A rename in these cases would change the fingerprint of the material. 

Material equality check should ideally consider the fingerprint of the material, but looks like its ignored, e.g [GitMaterial#equals ](https://github.com/gocd/gocd/blob/master/domain/src/main/java/com/thoughtworks/go/config/materials/git/GitMaterial.java#L314). The materials use UrlArgument.java to represent the URL, the [UrlArgument#equal()](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/util/command/UrlArgument.java#L118) ignores the trailing slash when comparing URL strings, hence ```https://github.com/gocd/gocd``` and ```https://github.com/gocd/gocd/``` are considered to be same which is incorrect.

This behaviour leads to errors/inconsistency while scheduling a pipeline since a check for [materialConfigurationChanged](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/scheduling/BuildCauseProducerService.java#L167) is false.

PR #5239, opened 2018-10-08
Materials should be same if fingerprints match

maheshp:
* Fixes https://github.com/gocd/gocd/issues/4836
* E.g  Consider two Git materials with URL - 'https://github.com/gocd/gocd' and
  'https://github.com/gocd/gocd/'. The two materials are considered different since
  fingerprint differs as one of them has a trailing slash.
  On the other hand GitMaterial#equals considers both to be same as the
  fingerprint was not used for comparision, rather the 'url' of the type
  'UrlArgument' is used to compare. 'UrlArgument' ignores the trailing
  slash while comparing the URL's

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/config/materials/git/GitMaterial.java
+++ b/domain/src/main/java/com/thoughtworks/go/config/materials/git/GitMaterial.java
@@ -324,13 +324,10 @@ public class GitMaterial extends ScmMaterial {
 
         GitMaterial that = (GitMaterial) o;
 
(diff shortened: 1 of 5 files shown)


ketan:
I don't feel that equals and hashcode should be a function of the material fingerprint. A fingerprint is a domain concept and I don't feel equals and hashcode should consume that.

The bug seems to be with how `UrlArgument` treats the trailing slash, and we should address it there.

maheshp:
This needs more thought, I am just wondering bout the repercussion of changing `UrlArgument`. Closing this for now, will submit a new PR once we decide on a fix.

PR #5242, opened 2018-10-09
Webpack improvements

akshaydewan:
- Enable camelCase option plugin for accessing styles easily in JS
- Use HappyPack where possible
- Refactor webpack config for readability


server/webapp/WEB-INF/rails/config/webpack-base.config.js
```diff
@@ -236,7 +188,61 @@ module.exports = function (env) {
...
+          }
+        }
+      ],
+      threadPool: happyThreadPool
+    }),
+    new HappyPack({
+      id:         'scss',
+      loaders:    [
+        {loader: 'style-loader'},
+        {loader: "css-loader?modules&camelCase&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]"}, // translates CSS into CommonJS
```
ketan: Use the options hash, like in line 211 and 225?

akshaydewan: Ok, will do that 👍

Issue #4469, opened 2018-03-01
Agent Error occurred during loop NullPointerException: Cannot invoke "org.apache.http.Header.getValue()" because "md5Header" is null


mario-fuentes:
##### Issue Type

- Bug Report

##### Summary

The GoCD Server inform that our two agents are idle but lot of task are pending waiting for an agent.  The agent docker log of both ends with a serie of NullPointerException, more than twenty continuous errors of the fellow type:

`
2018-03-01 11:55:55,365 [stdout] - 2018-03-01 11:55:55,364 ERROR [loopThread] AgentController:91 - [Agent Loop] Error occurred during loop: 

2018-03-01 11:55:55,365 [stdout] - java.lang.NullPointerException: null
`

##### Environment

[gocd-support.txt](https://github.com/gocd/gocd/files/1771224/gocd-support.txt)

###### Basic environment details

Go Server Version:	18.1.0 (5937-8a847b96ee8d38173f80178ed1285f0e53a970e0)
JVM version:	1.8.0_151
OS Information:	Linux 4.9.76-3.78.amzn1.x86_64
Usable space in artifacts repository:	2.72 GB
Database schema version:	1801007
Pipelines Count:	3

###### Additional Environment Details

We are using the official docker images.

ketan:
Do you have the complete stacktrace on the agent? This error does not
contain enough information to diagnose the issue.

mario-fuentes:
Hi Ketan, I know that the log isn't sufficient to diagnose, but our agents aren't running in debug mode.  I restarted the agent's containers and are working again.  If the issue persists, we should active the debug mode.

ketan:
For now I'll close this issue, when you have more information, please post it here and we can reopen this issue.

mario-fuentes:
Here more details

```
2018-03-01 13:56:32,068 [stdout] - 2018-03-01 13:56:32,067 ERROR [loopThread] AgentController:91 - [Agent Loop] Error occurred during loop: 
2018-03-01 13:56:32,068 [stdout] - java.lang.NullPointerException: null
2018-03-01 13:56:32,068 [stdout] - 	at com.thoughtworks.go.agent.service.AgentUpgradeService.validateMd5(AgentUpgradeService.java:97)
2018-03-01 13:56:32,068 [stdout] - 	at com.thoughtworks.go.agent.service.AgentUpgradeService.checkForUpgrade(AgentUpgradeService.java:80)
[...]

PR #5243, opened 2018-10-09
Change log level to debug of logs around latest-agent.status request, to avoid too much noise in server logs

adityasood:


GaneshSPatil:
This log was added to debug https://github.com/gocd/gocd/issues/4469 issue further.
GoCD experimental build [7575](https://hub.docker.com/r/gocdexperimental/gocd-server/tags/) contains these changes.
As we haven't identified the issue, can we keep the log and change the level to `debug`?

PR #5250, opened 2018-10-10
Fix a bunch of deprecation warnings

ketan:


util/src/main/java/com/thoughtworks/go/util/command/StreamPumper.java
```diff
@@ -46,24 +50,19 @@ private StreamPumper(InputStream in, StreamConsumer streamConsumer, String prefi
         this.clock = clock;
         this.lastHeard = System.currentTimeMillis();
         try {
-            this.in = new LineNumberReader(new InputStreamReader(in, encoding));
+            this.in = new InputStreamReader(in, encoding);
```
akshaydewan: Is the change from `BufferedReader` to `InputStreamReader` deliberate?

ketan: Just wanted to use the minimal interface. The `LineNumberReader` seemed like something we're not really using, and the streams are all "in-memory", so buffering was not quite needed.

PR #5252, opened 2018-10-10
Pipeline label regex fix

marques-work:
Simplify PipelineLabel validation and interpolation to prevent long-running regex issues

  - Do not validate in XSD
  - Break up gnarly regex into simple string operations + simpler regexes
    - match COUNT
    - match env:
    - match materials + truncation spec
  - Explicitly interpolate COUNT instead of pretending it's a special material
  - Split out logic for environment variable resolution vs material resolution


marques-work:
Merged this after checking with @ketan (we paired on big chunks of this).

PR #5256, opened 2018-10-11
Making a bunch of APIs to support pipeline-counter instead of both counter and label

akshaydewan:
Original PR:  #4660
Rebased with master, fixed conflicts. Pending PR run.
(Pushed on my own branch, because I don't have access to push to Jyoti's branch)


rahulpargaonkar:
Tested the APIs  which has pipeline_counter  in request URL, all API works fine but only a API, 'Get Stages' from 'Feed' is having both 'pipeline_counter' and 'pipeline_label'. And, currently, if 'pipeline_label' is used instead of counter then its gives 404. So, either we have this particular API with 'pipeline_label' (Does it requires?) or we need to remove it from API doc.

rahulpargaonkar:
verified API doc, now there is no alternate API which uses `pipeline label`

PR #5258, opened 2018-10-11
Move elastic profile api to spark

bdpiparva:


GaneshSPatil:
@bdpiparva -- Can we keep the rails API around too for a while? We can use `SparkOrRailsToggle` at the url rewrite level to route to either sparkAPI or railsAPI based on the toggle?

bdpiparva:
@GaneshSPatil - I don't think so we should keep it around. If there are issues in this we have a month to fix it.

GaneshSPatil:
> @GaneshSPatil - I don't think so we should keep it around. If there are issues in this we have a month to fix it.

If we've fair confidence in the changeset, i am OK with it :)

PR #5260, opened 2018-10-11
Reimplement AjaxPoller with plain setTimeout()

marques-work:
- The `repeat` npm module is no longer maintained or developed (since 2012)
  - Repeat objects don't seem to always `stop()` properly (?!?!)
  - Interface is the same, should be a transparent change


marques-work:
Merging. Windows jasmine tests refuse to check out this revision, but I've tested this locally in windows, and it's working.

PR #5267, opened 2018-10-15
Remove pipeline config API v5

GaneshSPatil:
* Pipeline config API v5 has been deprecated as part for GoCD release v18.7.0


GaneshSPatil:
@maheshp @ketan -- Ideally, this should have been merged for current release `v18.10.0`. But, as we are releasing soon, marking it for next release `v18.11.0`.

ketan:
Can simplify the `urlrewrite.xml` file to not check the header anymore.

PR #5268, opened 2018-10-15
Remove template config API v3

GaneshSPatil:
* Template config API v3 has been deprecated as part for GoCD release v18.7.0


GaneshSPatil:
@maheshp @ketan -- Ideally, this should have been merged for current release `v18.10.0`. But, as we are releasing soon, marking it for next release `v18.11.0`.

PR #5274, opened 2018-10-16
Fix rspec failures pipeline config v5

GaneshSPatil:
* Replace all the occurrences of apiv5_admin_pipeline_url with spark pipeline config v6 url.
* Do not refer to representers from other modules.

### DIFF
--- a/server/webapp/WEB-INF/rails/app/presenters/api_v1/admin/merged_environments/pipeline_config_summary_representer.rb
+++ b/server/webapp/WEB-INF/rails/app/presenters/api_v1/admin/merged_environments/pipeline_config_summary_representer.rb
@@ -26,7 +26,7 @@ module ApiV1
         end
 
         link :self do |opts|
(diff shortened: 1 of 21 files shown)


server/webapp/WEB-INF/rails/app/presenters/api_v1/admin/merged_environments/pipeline_config_summary_representer.rb
```diff
@@ -26,15 +26,15 @@ def initialize(options)
         end
 
         link :self do |opts|
-          opts[:url_builder].apiv5_admin_pipeline_url(@pipeline.getName())
+          "#{opts[:url_builder].default_url_options[:protocol]}://#{opts[:url_builder].default_url_options[:host]}/api/admin/pipelines/#{@pipeline.getName()}"
```
ketan: `default_url_options` are just that — defaults used when the incoming request does not contain the required data. As such `default_url_options` can be blank. See `ApiV1::Dashboard::PipelineRepresenter` for how to do this right.

PR #5281, opened 2018-10-17
Add Expand Collapsible component

GaneshSPatil:
#### Pending:
- [ ] Styles
- [ ] Test for expand collapse behaviour


server/webapp/WEB-INF/rails/webpack/views/components/expand_collapsible/index.scss
```diff
@@ -0,0 +1,66 @@
...
+  position:      relative;
+  border:        1px solid $border-color;
+  border-radius: $global-border-radius;
+  margin-bottom: 10px;
+  &.collapsed {
+    background: $page-header-bg;
+  }
+  &.expanded {
+    box-shadow: 0 0 10px rgba(0, 0, 0, 0.21);
+    .c-collapse_header {
```
ketan: FYI: we're moving to `camelCase` classnames. Talk to @naveenbhaskar and @akshaydewan on how to best name classes now.

GaneshSPatil: @ketan -- This was just a copy paste from the existing code. @naveenbhaskar is going to work on the styles bit and probably will remove this class.

akshaydewan:
The names `expand_collapsible` and `<ExpandCollapse>` seem strange. Can we just call it a CollapsiblePanel ?

server/webapp/WEB-INF/rails/webpack/views/components/expand_collapsible/index.js.msx
```diff
@@ -0,0 +1,52 @@
...
+    self.showState = () => {
+      return self.expanded() ? 'show' : 'hide';
+    }
+  },
+
+  view(vnode) {
+    const self = this;
+
+    return (
+      <div class={`${styles.collapse} ${self.expanded() ? styles.expanded : styles.collapsed}`}>
```
akshaydewan: We've added a module called added [classNames](https://github.com/JedWatson/classnames). Some of these conditional styles can be written like
```javascript
const classes = classNames({
  styles.collapse,
  styles.expanded: self.expanded()
  styles.collapsed: !self.expanded()
})
```
or something similar

server/webapp/WEB-INF/rails/webpack/views/components/expand_collapsible/spec/index_spec.js
```diff
@@ -0,0 +1,97 @@
...
+    m.redraw();
+  }
+
+  function find(id) {
+    return $root.find(`[data-test-id='${id}']`);
+  }
+
+  function containsClass(classList, expectedClass) {
+    return _.some(classList, (c) => _.includes(c, expectedClass))
+  }
```
akshaydewan: Are we checking for a substring that has the class name? That doesn't seem like a good idea to me. I would rather import the scss file in the test and do an exact match.

PR #5282, opened 2018-10-17
Add sass lint for webpack. Fix some lint errors.

akshaydewan:


akshaydewan:
@ketan I've removed some rules like camelCase and vendor-prefix as of now.

ketan:
> @ketan I've removed some rules like camelCase and vendor-prefix as of now.

Is this because there were too many to fix?

akshaydewan:
For camelCase, yes. Also, the css-loader exposes names in camelCase format, so I wasn't sure if we want to enforce that rule.
For vendor-prefix, I spoke to @naveenbhaskar and we add them in some places for additional safety. So I just kept them.

ketan:
> For camelCase, yes. Also, the css-loader exposes names in camelCase format, so I wasn't sure if we want to enforce that rule.

That actually makes it convenient to use in JS, so I prefer to keep it that way, unless @naveenbhaskar (or someone really dislikes it :)

> For vendor-prefix, I spoke to @naveenbhaskar and we add them in some places for additional safety. So I just kept them.

That can be added by a post-processor like [autoprefixer](https://autoprefixer.github.io/), so I intentionally dropped it.

ketan:
I've rebased and force pushed a build to fix lint (apparently something got fixed with master, with the agents.scss).

PR #5295, opened 2018-10-19
Should get stage from identifier of case in which pipeline and stage name is specified.

vishaldevgire:
The `pipelineName` and `stageName` should be treated as case-insensitive strings when used in caching and retrieving data from the DB. 

Following code always looks up pipeline  from db in specified case which can lead to pipeline not found if specified name is in a different case then the one in db.  

https://github.com/gocd/gocd/blob/f0d1a6e6b46f9c342a698b9e9c0fdd8e8d21e47e/server/src/main/resources/com/thoughtworks/go/server/dao/maps/Stage.xml#L597

To fix above issue we removed casting of pipeline name during the query execution
https://github.com/gocd/gocd/pull/5295/files#diff-eee1825c5ee96e91bc228702bffb2ae9R597

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java
@@ -357,7 +357,7 @@ public class StageSqlMapDao extends SqlMapClientDaoSupport implements StageDao, 
     }
 
     String cacheKeyForAllStageOfPipeline(String pipelineName, Integer pipelineCounter, String stageName) {
-        return cacheKeyGenerator.generate("allStageOfPipeline", pipelineName, pipelineCounter, stageName);
+        return cacheKeyGenerator.generate("allStageOfPipeline", pipelineName.toLowerCase(), pipelineCounter, stageName.toLowerCase());
     }
 
     public List<Stage> findStageHistoryForChart(String pipelineName, String stageName, int pageSize, int offset) {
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java
```diff
@@ -475,7 +475,7 @@ String mutexForStageHistory(String pipelineName, String stageName) {
     }
 
     String cacheKeyForStageHistories(String pipelineName, String stageName) {
-        return cacheKeyGenerator.generate("stageHistories", pipelineName, stageName);
+        return cacheKeyGenerator.generate("stageHistories", pipelineName.toLowerCase(), stageName.toLowerCase());
```
ketan: Perhaps consider pushing this in `CacheKeyGenerator`:

```
public String generate(String identifier, Object... args) {
        final List<Object> allArgs = Arrays.stream(args).map(arg -> {
            if (isAllowed(arg)) {
                return arg;
            }
            throw new IllegalArgumentException("Type " + arg.getClass() + " is not allowed here!");
        }).map(arg -> {
            if (arg instanceof CaseInsensitiveString) {
                return ((CaseInsensitiveString) arg).toLower();
            } else {
                return arg;
            }
        }).collect(Collectors.toList());

        allArgs.add(0, clazz.getName());
        allArgs.add(1, identifier);

        return StringUtils.join(allArgs, DELIMITER).intern();
    }
```

PR #5301, opened 2018-10-22
Upgrade Highcharts to v6.2.0

GaneshSPatil:
\cc @maheshp @arvindsv @ketan 

We've Highcharts license for `v6`
Upgrading GoCD core to `v6.x.x` of Highcharts should be fine?


arvindsv:
@GaneshSPatil Not really. :( We have a limited license. Not an unlimited license like the one for 2.x.

GaneshSPatil:
@arvindsv -- IIRC, we've Highchart's OEM and Enterprise License, which allows us to use Highcharts on 100 instances/applications.

Can you check and let me know?

If we're limited by the usage number, we can close this PR.

arvindsv:
@GaneshSPatil Yes, I think the "100 instances" part will become a problem. Each installation will be considered an "instance". If we add it to the installer (18.10, 18.11, etc) it'll be an instance, is my understanding.

GaneshSPatil:
Ahh. Thanks for the information. :)
Closing this PR.

Issue #5232, opened 2018-10-05
UI componentization and refactoring


akshaydewan:
This issue is to track the UI work around componentization and CSS refactoring

## Infrastructure 
- [x] Upgrade to webpack 4
- [ ] Setup Mini-CSS-Extract plugin to remove `<style>` tags from pages and pull them out into a separate file
- [ ] Move specs into a folder along with the code
  - [ ] Verify that specs are not packaged in production and have no side effects on the main code
  - [ ] Move all specs with the code for consistency
- [ ] Minify CSS and strip out comments
- [ ] Fix FontAwesome in components (cannot use `.erb` file)
- [ ] Spike using Typescript
- [ ] Spike changing theme from javascript
- [ ] Check for better IDE support (cmd+click for links, navigate to tests etc)
- [ ] Try to make the webpack build faster, move things into HappyPack if possible
- [ ] Optimize the chunks created in webpack

### Plugins page
- [ ] Build a basic css framework without Foundation
- [ ] Extract the site menu
- [ ] Extract the site footer
- [ ] Fix styles existing components, since Foundation isn't included
- [ ] Extract existing widgets and componentize them properly
### Elastic Agent Profiles page
- [ ] Use components extracted from Plugins page, and make necessary changes

maheshp:
@rahulpargaonkar to lead the testing effort for this.

PR #5308, opened 2018-10-23
Create Page Spinner component (#5232)

GaneshSPatil:
##### Normal
```
<PageSpinner />
```

##### Small
```
<PageSpinner small={true}/>
```

**Note:** PageSpinner component is not mounted on the `/go/kitchen-sink` page.


GaneshSPatil:
\cc: @naveenbhaskar

PR #5310, opened 2018-10-23
Improve performance of the config XML pages on admin section

ketan:


adityasood:
moving it to QA done based on some numbers we looked at when this PR was merged.

PR #5311, opened 2018-10-23
Introduce ConfigRepo Internal API

marques-work:
### Note

Up for debate as to whether this will be an internal API or a V2 public API

### Context

In the config-repos management page, we will need to display the last parsed revision + result for each repo, preferably on page load as it will be more useful. For example, if an admin can quickly identify which config-repos have errors, that adds value. In order to do this with the current facilities, we would have to iterate through each repo and make separate XHRs to fetch the status. This is not performant and doesn't scale when one has multiple repos.

This PR will allow us to list all repos and the associated results together in a single XHR. Any refreshes to individual repos on the page will only need a single request to get both structure and result.

The question now is whether this should be internal or public (currently built as public API). If the config-repos CRUD API is meant to represent the objects as they exist in the config XML, then this PR violates that guideline. I don't disagree with this assessment, but I do see a need to build what we've built.

@kierarad and I would propose the following:

  - Change this API to be internal, keeping the V1 API as-is
  - Change the config-repos-operations API to be a public API (these endpoints are useful as one can force-update and fetch status + parse results)

Thoughts? @ketan @maheshp @GaneshSPatil @akshaydewan 

Differences from V1

  - No create, update, or delete
  - Only implements `all()` and `get()` equivalents
  - Responds with config-repo structure AND last parsed result as a single JSON entity
  - Introduce a new set of Java representers for materials and config-repos
    - These could be used in the future to support porting the V1 API from Rails to Spark


marques-work:
## TODO:

- [ ] calculate etags for caching (this will be based on both structure and parse result)

ketan:
If I understand right, there is an addition of the `last_parse` attribute on the response JSON. This is then shown in the UI. After we're sure the page is good (in terms of what relevant information needs to be shown), we can call this as V2.

ketan:
Needless to say, the rest of api v1 will need to be ported over :)

marques-work:
@ketan yes, that is exactly what these endpoints do. Thanks for taking a look!

Eventually, we will port V1 over. Hopefully the representers introduced here will make it much easier.

One change here that I’ve discussed with @maheshp - we’re going to instead make this API internal for now until the SPA is done. Thereafter we can reevaluate to see which endpoints should become public.

marques-work:
This API is now completely internal for the time being.

PR #5312, opened 2018-10-24
Import all components

GaneshSPatil:
* Many of the times, its often required to import multiple components,
  instead of writing individual imports, we can import entire components dir

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/index.ts
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/index.ts
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
@akshaydewan @ketan -- WDYT?
* The IDE navigations take directly to components (Thanks to typescript).
* Importing `c` as `components` gives autocompletion and param suggestions too (as earlier).
* This just helps to know all the components in one page (instead of looking into the entire components dir).
* And, we still have a choice of not importing the entire components dir and importing individual required components.

ketan:
> WDYT?

There's now two ways of importing the same component :) And particularly makes reading hard `c.Something.SomethingElse`

GaneshSPatil:
> > WDYT?
> 
> There's now two ways of importing the same component :) And particularly makes reading hard `c.Something.SomethingElse`

Makes sense. Closing this PR.

PR #5318, opened 2018-10-24
Do not show the "Tests" tab when shine is disabled.

ketan:
Also disable the `/tests` url if shine is disabled.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
+++ b/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
@@ -592,8 +592,8 @@ public class SystemEnvironment implements Serializable, ConfigDirProvider {
     }
 
 
-    public String getEnvironmentVariable(String key) {
-        return System.getenv(key);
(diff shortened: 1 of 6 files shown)


base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -605,7 +605,7 @@ public String getBaseUrlForShine() {
     }
 
     public Boolean isShineEnabled() {
-        String shineEnabled = getEnvironmentVariable("SHINE_ENABLED");
+        String shineEnabled = getEnvironmentVariable("SHINE_ENABLED", "false");
         return shineEnabled == null ? false : !"false".equalsIgnoreCase(shineEnabled); //should return true for shine_enabled set to anything but false.
```
varshavaradarajan: Change this line to return ` !"false".equalsIgnoreCase(shineEnabled);` ?

ketan: This code is going away soon :) I don't want to give it too much love ;)

PR #5324, opened 2018-10-25
Create a modal component

akshaydewan:
Epic: #5232 

Implementation of Modal component in TypeScript.
- Used the modal for the errors and warnings popup.
- Styles are pending and theme needs to be standardized
- The `SampleModal` and `SampleModal2` classes are used only in kitchensink
- Handling of the `Enter` key isn't implemented yet
- Modals opening other modals have some issues, this can be fixed later.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
@@ -23,7 +23,8 @@ const classnames = require('classnames/bind').bind(styles);
 
 export interface Attrs {
   small?: boolean,
-  disabled?: boolean
+  disabled?: boolean;
+  onclick?: Function;
 }
 
 class Button extends MithrilComponent<Attrs> {
--- a/server/webapp/WEB-INF/rails/webpack/views/components/modal/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/modal/index.scss
@@ -28,7 +28,7 @@ $white: #fff;
   border:        none;
   border-radius: $overlay-border-radius;
   position:      absolute;
-  z-index:       101;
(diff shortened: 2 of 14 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
```diff
@@ -23,6 +23,7 @@ const classnames = require('classnames/bind').bind(styles);
 
 export interface Attrs {
   small?: boolean
+  onclick?: () => any;
```
GaneshSPatil: How about `Function`?

server/webapp/WEB-INF/rails/webpack/views/components/modal/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  render() {
+    ModalManager.render(this);
+  }
+
+  close() {
+    ModalManager.close(this);
+  }
+
+  handleEscapeKey() {
+    $('body').on(`keydown.${this.id}`, (e) => {
```
ketan: Can we consider weaning off jquery? This may mean holding a reference to the function, which we can later uninstall ourselves?

server/webapp/WEB-INF/rails/webpack/views/components/modal/index.scss
```diff
@@ -0,0 +1,130 @@
...
+$overlay-border-color: #ccc;
+$overlay-border-radius: 3px;
+$white: #fff;
+
+.overlay {
+  display:       none;
+  border:        none;
+  border-radius: $overlay-border-radius;
+  position:      absolute;
+  z-index:       101;
```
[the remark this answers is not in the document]
akshaydewan: Fixed and added a sass lint rule as well. Thanks for pointing it out.

PR #5330, opened 2018-10-25
Export Pipeline Config to Repo Config Format

ibnc:
This PR converts the Config model (PipelineConfig, StageConfig, etc) to the CR model (CRPipeline, CRStage, etc), and sends it as JSON to the specified plugin. The idea being that the plugin will return the configuration in the plugins data format.

So, the yaml plugin will convert the JSON into YAML, and send it back to the server. The JSON plugin will just give back the received JSON.

This PR only implements the extension side of things, and is currently unused. We still need to figure out what the interaction will look like from the user's perspective, and we need to wait until  @marques-work and @kierarad are finished reworking the config repo api endpoints right now. 

I'll be opening PRs on the JSON and YAML plugins here soon to implement this functionality. 

@tomzo would you be able to review this PR?

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
+++ b/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
@@ -29,6 +29,7 @@ import com.thoughtworks.go.config.pluggabletask.PluggableTask;
 import com.thoughtworks.go.config.remote.PartialConfig;
 import com.thoughtworks.go.domain.EnvironmentVariable;
 import com.thoughtworks.go.domain.RunIfConfigs;
+import com.thoughtworks.go.domain.Task;
 import com.thoughtworks.go.domain.config.Arguments;
 import com.thoughtworks.go.domain.config.Configuration;
 import com.thoughtworks.go.domain.config.ConfigurationProperty;
@@ -46,10 +47,13 @@ import com.thoughtworks.go.security.GoCipher;
 import com.thoughtworks.go.util.command.HgUrlArgument;
 import com.thoughtworks.go.util.command.UrlArgument;
 import org.apache.commons.lang3.StringUtils;
+import org.eclipse.rdf4j.query.algebra.Str;
+import org.hibernate.annotations.Fetch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * Helper to transform config repo classes to config-api classes
@@ -250,7 +254,6 @@ public class ConfigConverter {
         buildTask.target = crBuildTask.getTarget();
         buildTask.workingDirectory = crBuildTask.getWorkingDirectory();
     }
-
 
     private Configuration toConfiguration(Collection<CRConfigurationProperty> properties) {
         Configuration configuration = new Configuration();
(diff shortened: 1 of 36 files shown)


tomzo:
Hi @ibnc 

> we need to wait until @marques-work and @kierarad are finished reworking the config repo api endpoints right now.

Sounds right, this might change a bit.

> I'll be opening PRs on the JSON and YAML plugins here soon to implement this functionality.

Greatly appreciated :+1: .

> @tomzo would you be able to review this PR?

I've read through this. It is quite simple, but quite large domain to transform. Yet you still written the zillion tests on each conversion so I have nothing to add to that :smile: .

One thing that might be an issue in the end is that exported pipeline will have no group. Although it is specific to json and yaml plugins. Both define pipeline with a group in it's body. So for example the test you wrote in json plugin - https://github.com/tomzo/gocd-json-config-plugin/pull/19/files#diff-15dcca2244fe260b6fee615c316b9082R195 is actually returning pipeline which is invalid. 
I think plugin might want to have group in context of export. So for example `String pipelineExport(String pluginId, final CRPipeline pipelineConfig)` would rather be `String pipelineExport(String pluginId, final CRPipelineWithGroup pipelineConfig)`.
It depends on how the final feature will look like. But if this is to be used for exporting 100+ pipelines, then I guess each of them should be valid, ready to use when server spits it out.

arvindsv:
@ibnc This can't be merged till a new version of the ConfigRepoExtension is created. [This](https://github.com/gocd/gocd/pull/5330/files#diff-99e9528bf2fe4053efe82c35da436b52R42) has to change to support "1.0" and "2.0". Take a look at other extensions to get a sense of how versioning is handled, or I can help too. Thanks!

arvindsv:
Just to be clear, without that change: An upgrade of GoCD server will make the (older) plugins stop working.

arvindsv:
@ibnc I've opened an issue #5338, for any discussions about the feature, not related to this PR directly. Thank you!

tomzo:
One more thing... I think there should be an API call that queries plugin for its capabilities, similar to the one in elastic agents endpoint. So that some plugin may not support exporting.

ibnc:
> Hi @ibnc
> 
> > we need to wait until @marques-work and @kierarad are finished reworking the config repo api endpoints right now.
> 
> Sounds right, this might change a bit.
> 
> > I'll be opening PRs on the JSON and YAML plugins here soon to implement this functionality.
> 
> Greatly appreciated 👍 .
> 
> > @tomzo would you be able to review this PR?
> 
> I've read through this. It is quite simple, but quite large domain to transform. Yet you still written the zillion tests on each conversion so I have nothing to add to that 😄 .
> 
> One thing that might be an issue in the end is that exported pipeline will have no group. Although it is specific to json and yaml plugins. Both define pipeline with a group in it's body. So for example the test you wrote in json plugin - https://github.com/tomzo/gocd-json-config-plugin/pull/19/files#diff-15dcca2244fe260b6fee615c316b9082R195 is actually returning pipeline which is invalid.
> I think plugin might want to have group in context of export. So for example `String pipelineExport(String pluginId, final CRPipeline pipelineConfig)` would rather be `String pipelineExport(String pluginId, final CRPipelineWithGroup pipelineConfig)`.
> It depends on how the final feature will look like. But if this is to be used for exporting 100+ pipelines, then I guess each of them should be valid, ready to use when server spits it out.

You're totally right. Just added the group name.

lol, thanks for noticing the absurd amount of tests need to actually do unit testing for this... It took a while to do

ibnc:
@tomzo @arvindsv should the capabilities request be v2 or v1?

arvindsv:
@ibnc Usually, I consider existing versions "done" / unchanging. But ... there's more:

Typically, the server makes a call to the plugin to understand its capabilities. This becomes tricky with existing plugins. There are some cases we should think about:

- GoCD 18.10.0 doesn't know about capabilities for config-repo extension point / plugins.
- Assume GoCD 18.11.0 _does_ know about capabilities for config-repo extension point / plugins.
- Existing plugin (v1) has no knowledge of capabilities
- New plugin (v2) has knowledge of capabilities

1. Existing plugin on GoCD 18.10.0:
    - Doesn't get asked for capabilities. Nothing to do.

2. New plugin on GoCD 18.10.0:
    - Server doesn't know about v2 and will probably fail.
    - Server doesn't know about capabilities anyway.
    - I feel this case doesn't need to be supported.

3. Existing plugin on GoCD 18.11.0:
    - If server makes a call for capabilities, plugin fails that request.
    - Server will need to assume that the plugin is an existing v1 plugin. Depending on the order, the server might already know that the plugin's version is v1, and it can just skip the capabilities call. I think that's what is done elsewhere.

4. New plugin on GoCD 18.11.0:
    - Server makes a call for capabilities and agent responds.
    - Can negotiate capabilities.

In this particular case, if we want, we can assume that we know of all (three) implementations of this endpoint. And that we are in a position to change them, so that they're existing plugins (v1) with capabilities. This obviously adds a few cases, but doesn't fundamentally remove case 3 above.

tl; dr - I think that the server should assume that capabilities call is not handled by v1 clients and treat them appropriately. For v2 clients, it can negotiate capabilities.

ibnc:
@arvindsv What do you think about my approach (in my last commit)? We make the request to v1 plugins, but ignore the response and return Capabilities where everything is false.

Also, general feedback would be appreciated on my last commit

arvindsv:
@ibnc  If the (v1) plugin fails that request, `onSuccess` won't be called, right? I would consider making a version check before making that call. We've had cases where we've made those calls, without checking version and we've ended up confusing people who see those errors in logs.

Generally, the rest looks ok to me, based on a  quick look, some comments / thoughts:

1. If [this](https://github.com/gocd/gocd/pull/5330/commits/a5ab1841b51058219a61f010f345c008cc7f6f4d#diff-b0493718c4fc0ca29fc3ef4caf757862R119) is not supposed to be called at all, I'd just throw rather than return null. I'd assume we wouldn't even show `export` as an option, based on the capabilities.

2. [This method name](https://github.com/gocd/gocd/pull/5330/commits/a5ab1841b51058219a61f010f345c008cc7f6f4d#diff-7eae5e5e9054a3e3b063417c279f738aR37) needs to change.

3. It's not obvious that `new Capabilities()` means all false. I'd consider either `new Capabilities(false)` or `Capabilities.NONE` or something of that sort.

ibnc:
@arvindsv you're totally right. How does my most recent commit look?

arvindsv:
@ibnc Did you mean to comment out the [previous line instead](https://github.com/gocd/gocd/pull/5330/commits/3549eae2fce9ad40f2fb24ab4931a2aff8fe4406#diff-44b97bb4968d2df2b6339115e41a118fR128)? I'd also throw [here](https://github.com/gocd/gocd/pull/5330/commits/3549eae2fce9ad40f2fb24ab4931a2aff8fe4406#diff-b0493718c4fc0ca29fc3ef4caf757862R73).

But, yes - it looks good to me. In a more extreme case, in elastic agents extension, the extension point itself is versioned. That's a little cleaner, but is overkill for this change. Your approach seems alright to me.

ibnc:
@arvindsv Then this is mergeable (assuming the build passes)?

arvindsv:
@ibnc I'll take a look again later and merge it.

ibnc:
This is failing because of Windows Jasmine tests. They appear to failing for everything right now

arvindsv:
I guess it comes down to, will the capabilities check be done earlier and these methods won't be called or will we wait till we get the (`null`) response from these methods in `ConfigRepoExtension#pipelineExport` to decide. I'd assume, the former.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+        );
+
+        commonCRTaskMembers(crExecTask, task);
+        return crExecTask;
+    }
+
+    public CRBuildTask buildTaskToCRBuildTask(BuildTask buildTask) {
+        CRBuildTask crBuildTask;
+        if (buildTask instanceof RakeTask) {
+            RakeTask rake = (RakeTask) buildTask;
```
arvindsv: This line/var is unused.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+                    String.format("unknown run if condition '%s'", runIf));
+        }
+    }
+
+    private List<CRConfigurationProperty> configurationToCRConfiguration(Configuration config) {
+        ArrayList<CRConfigurationProperty> properties = new ArrayList();
+        if (config != null) {
+            for (ConfigurationProperty p : config) {
+                CRConfigurationProperty crProp = new CRConfigurationProperty(p.getKey().getName());
+                if (p.getValue() != null)
```
arvindsv: Should check `isSecure`, instead of `getValue`. `getValue` decrypts on the fly and is not often `null`. `getConfigValue` might work, but it is probably more readable to check `isSecure`?

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+        crExecTask.setArgs(task.getArgList().stream()
+                .map(Argument::getValue)
+                .collect(Collectors.toList())
+        );
+
+        commonCRTaskMembers(crExecTask, task);
+        return crExecTask;
+    }
+
+    public CRBuildTask buildTaskToCRBuildTask(BuildTask buildTask) {
```
arvindsv: Can you make all of these methods private/package-private if they don't need to be public? IntelliJ shows that as a warning.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+
+    public CRFetchArtifactTask fetchTaskToCRFetchTask(FetchTask task) {
+        CRFetchArtifactTask fetchTask = new CRFetchArtifactTask(
+                task.getStage().toString(),
+                task.getJob().toString(),
+                task.getSrc());
+
+        fetchTask.setDestination(task.getDest());
+        fetchTask.setPipelineName(task.getPipelineName().toString());
+        fetchTask.setSourceIsDirectory(task.isSourceAFile());
```
arvindsv: This _seems_ wrong.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+    public CRDependencyMaterial dependencyMaterialConfigToCRDependencyMaterial(DependencyMaterialConfig dependencyMaterialConfig) {
+        CRDependencyMaterial crDependencyMaterial = new CRDependencyMaterial(
+                dependencyMaterialConfig.getPipelineName().toString(),
+                dependencyMaterialConfig.getStageName().toString());
+        if (dependencyMaterialConfig.getName() != null)
+            crDependencyMaterial.setName(dependencyMaterialConfig.getName().toString());
+        return crDependencyMaterial;
+    }
+
+    private CRScmMaterial scmMaterialToCRScmMaterial(ScmMaterialConfig scmMaterialConfig) {
```
arvindsv: Can you break this method up a bit, at least? It's hard to follow. I know the original was similar.  :) But, I think this can be rewritten a bit, even if it's a little longer.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+        String materialName = null;
+        if (scmMaterialConfig.getName() != null) {
+            materialName = scmMaterialConfig.getName().toString();
+        }
+        if (scmMaterialConfig instanceof GitMaterialConfig) {
+            GitMaterialConfig git = (GitMaterialConfig) scmMaterialConfig;
+            return new CRGitMaterial(materialName, git.getFolder(), git.isAutoUpdate(), git.isShallowClone(), git.getUrl(), git.getBranch(), git.isInvertFilter(), git.getFilterAsString());
```
arvindsv: Does `git.getFilterAsString()` really work? I'm not completely sure, but it's unlikely to be correct. The filters will have to be separate. Same with other materials. If it works in one place, I guess it works elsewhere.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+
+            return crP4Material;
+        } else if (scmMaterialConfig instanceof SvnMaterialConfig) {
+            SvnMaterialConfig svnMaterial = (SvnMaterialConfig) scmMaterialConfig;
+            CRSvnMaterial crSvnMaterial = new CRSvnMaterial(materialName, svnMaterial.getFolder(), svnMaterial.isAutoUpdate(),
+                    svnMaterial.getUrl(), svnMaterial.getUserName(), null, svnMaterial.isCheckExternals(), svnMaterial.isInvertFilter(), svnMaterial.getFilterAsString());
+            if (svnMaterial.getEncryptedPassword() != null) {
+                crSvnMaterial.setEncryptedPassword(svnMaterial.getEncryptedPassword());
+            } else {
+                crSvnMaterial.setPassword(svnMaterial.getPassword());
```
arvindsv: Shouldn't we just use encrypted password here as well? The `SvnMaterialConfig#ensureEncrypted` method should ensure that the password is encrypted, when it is in this form. Right?

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+        SCMs scms = getSCMs();
+        String id = pluggableScmMaterialConfig.getScmId();
+        SCM scmConfig = scms.find(id);
+        if (scmConfig == null)
+            throw new ConfigConvertionException(
+                    String.format("Failed to find referenced scm '%s'", id));
+
+        return new CRPluggableScmMaterial(pluggableScmMaterialConfig.getName().toString(),
+                id, pluggableScmMaterialConfig.getFolder(),
+                pluggableScmMaterialConfig.getFilterAsString());
```
arvindsv: `getFilterAsString` might have the same problem mentioned elsewhere.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: How so?

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: I'm unsure. I did it this way because users are able to specify a plain text password? @tomzo thoughts?

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -639,4 +641,395 @@ private TrackingTool toTrackingTool(CRTrackingTool crTrackingTool) {
...
+    }
+
+    public CRApproval approvalToCRApproval(Approval approval) {
+        CRApproval crApproval = new CRApproval();
+        for(AdminUser user: approval.getAuthConfig().getUsers()) {
+           crApproval.addAuthorizedUser(user.getName().toString());
+        }
+
+        for(AdminRole role: approval.getAuthConfig().getRoles()) {
+            crApproval.addAuthorizedUser(role.getName().toString());
```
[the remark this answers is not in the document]
ibnc: You're totally right. This work was tedious :/

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: Yeah... It's pretty cumbersome. I'll break it up a bit

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
arvindsv: I was looking mostly at the names, but shoudn't it be:

```
fetchTask.setSourceIsDirectory(!task.isSourceAFile());
```
?

The parameter's name is `srcIsDirectory`.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
tomzo: Yes users can have plain text passwords, if they want to. This applies to svn, perforce and TFS.
But what @arvindsv  had in mind was probably that now we can enforce encrypted passwords on user - by default exported pipeline would have encrypted password.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
tomzo: Should be `fetchTask.setSourceIsDirectory(!task.isSourceAFile());`

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: So, @arvindsv should I be using the encrypted password for all of the materials?

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: 👍

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
arvindsv: @ibnc Yes, we should. No reason to export with a plain-text password. What @tomzo said. Thank you!

server/src/test-fast/java/com/thoughtworks/go/config/ConfigConverterTest.java
```diff
@@ -1631,7 +1631,7 @@ public void shouldConvertFetchArtifactTaskWhenSourceIsDirectoryToCR() {
                new CaseInsensitiveString("upstream"),
                new CaseInsensitiveString("stage"),
                new CaseInsensitiveString("job"),
-               "src",
```
arvindsv: @ibnc Shouldn't this be verified by adding a line:

```
fetchTask.setSrcdir("src")
```

Otherwise, the test name doesn't reflect what's happening.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
arvindsv: @ibnc Looks alright to me.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
arvindsv: Just to be clear, I believe it's already done for other materials.

ibnc: 👍

server/src/test-fast/java/com/thoughtworks/go/config/ConfigConverterTest.java
ibnc: nvm. Changed the wrong test :/

ibnc:
> I guess it comes down to, will the capabilities check be done earlier and these methods won't be called or will we wait till we get the (`null`) response from these methods in `ConfigRepoExtension#pipelineExport` to decide. I'd assume, the former.

@arvindsv I was planning on checking the Capabilities before listing the pipeline for the export, and just not show plugins that don't support exporting

PR #5332, opened 2018-10-26
Do not send ambiguous `buildId` query string for artifact APIs

ketan:


ketan:
Closing in favor of #5348

PR #5344, opened 2018-10-29
Render component-pages using entrypoints defined under single_page_apps/ (#5232)

GaneshSPatil:


server/webapp/WEB-INF/rails/webpack/single_page_apps/plugins.tsx
```diff
@@ -14,20 +14,35 @@
...
+}
 
 $(() => {
   const pluginElement = $('#plugins');
   const isUserAnAdmin = pluginElement.attr('is-user-an-admin');
 
-  const onSuccess = (pluginInfos) => {
+  //todo: This check determines whether the page needs to be rendered using old spa style or using new components styles
+  //pluginsElement div will only be available while rendering using old style.
+  if (pluginElement.get().length === 0) {
```
GaneshSPatil: There are two implementations of plugins page. One - old SPA and other - the new components style.
Depending on the toggle value, we need to render old/new page. To make it work (with minimal changes), this hack is required.

On the componentized plugins page, there will be no div with `plugins` id on it, hence it will render the components-plugins page.
Whereas, the old SPA will have a container div with `plugins` id.

This code will be removed soon once plugins SPA is componentized.
\cc: @akshaydewan @ketan

PR #5352, opened 2018-10-31
Configure tsconfig.json to treat <rails.root>/webpack as the module root

marques-work:
- This lets us do things like `import SparkRoutes from "helpers/spark_routes"`
    instead of `import SparkRoutes from "../../../helpers/spark_routes"` in your
    editor (e.g., VS Code) without it yelling at you.
  - Changes <rails.root>/tsconfig.json => <rails.root>/tsconfig.common.json
  - spec/ and webpack/ now have their own tsconfig.json files, which extend
    tsconfig.common.json to enable this behavior
  - Webpack config to awesome-typescript-loader still uses the cononical config (just
    renamed to tsconfig.common.json) so there should be no change in behavior for
    `webpack-dev` or `webpack-prod` yarn tasks; this should be a safe change.

### DIFF
--- a/server/webapp/WEB-INF/rails/config/webpack-base.config.js
+++ b/server/webapp/WEB-INF/rails/config/webpack-base.config.js
@@ -101,7 +101,7 @@ module.exports = function (env) {
               {
                 loader:  'awesome-typescript-loader',
                 options: {
-                  configFile:  path.join(__dirname, '..', 'tsconfig.json')
+                  configFileName:  path.join(__dirname, '..', 'tsconfig.common.json')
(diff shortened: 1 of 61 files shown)


marques-work:
This looks like a lot of files (ok, well it is), but the main change is c9c2e1e.

The other commit is just removing the `../` from a lot of import statements. Again, webpack-dev/prod tasks aren't affected by this change. The "nice" paths were already resolvable by webpack before this PR, but this just makes our IDEs more friendly :).

Care to look @ketan or @GaneshSPatil?

PR #5360, opened 2018-11-01
Use the plugin_infos model in the plugins page.

akshaydewan:
Pending: The `angular_plugin_new` can possibly be converted to typescript as well

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/models/shared/plugin_infos_new/extensions.ts
+++ b/server/webapp/WEB-INF/rails/webpack/models/shared/plugin_infos_new/extensions.ts
@@ -14,10 +14,13 @@
  * limitations under the License.
  */
 
+import {Stream} from "mithril/stream";
 import {AnalyticsCapabilities} from "./analytics_plugin_capabilities";
 import {AuthCapabilities} from "./authorization_plugin_capabilities";
 import {ElasticPluginCapabilities} from "./elastic_plugin_capabilities";
 import {ExtensionType} from "./extension_type";
+
+const stream = require("mithril/stream");
 
 class Configuration {
   readonly key: string;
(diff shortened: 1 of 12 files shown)


server/webapp/WEB-INF/rails/webpack/models/shared/plugin_infos_new/extensions.ts
```diff
@@ -14,11 +14,14 @@
  * limitations under the License.
  */
 
+import {Stream} from "mithril/stream";
 import {AnalyticsCapabilities} from "./analytics_plugin_capabilities";
 import {AuthCapabilities} from "./authorization_plugin_capabilities";
 import {ElasticPluginCapabilities} from "./elastic_plugin_capabilities";
 import {ExtensionType} from "./extension_type";
 
+const Stream = require("mithril/stream");
```
GaneshSPatil: isn't this variable definition conflicting?
As per [this](https://github.com/MithrilJS/mithril.d.ts#stream-example) example, can we use `stream` for the real stream? and `Stream` for the type?

akshaydewan: The import as per the example wasn't working, it give a `no default export` error. But I've changed the require statement's stream to a small 's'

server/webapp/WEB-INF/rails/webpack/views/pages/new_plugins/spec/plugins_widget_spec.tsx
```diff
@@ -15,16 +15,17 @@
...
 import * as collapsiblePanelStyles from "../../../components/collapsible_panel/index.scss";
 import * as keyValuePairStyles from "../../../components/key_value_pair/index.scss";
 
 describe("New Plugins Widget", () => {
   const simulateEvent = require("simulate-event");
-  const PluginInfos   = require("models/shared/plugin_infos");
 
-  const pluginInfos = PluginInfos.fromJSON([getEAPluginInfo(), getNotificationPluginInfo()]);
+  const data        = [getEAPluginInfo(), getNotificationPluginInfo()];
+  const pluginInfos = PluginInfos.fromJSON({_embedded: {plugin_info: data}});
```
akshaydewan: @GaneshSPatil @ketan Somehow, the `fromJSON` method in the older PluginInfos was able to accept data without an `_embedded` object in the json. Do we need to support the same behaviour in the new code? I think not, so I have just created this object with an `_embedded`, as it should be received from the API.

Issue #4797, opened 2018-05-29
Allow javascript to be executed in the plugin settings modal


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

On the old plugins page, we had explicitly rendered the modal contents without escaping (https://github.com/gocd/gocd/pull/4226/files#diff-768a61bc4d59d2c699d5f45a6da3dd03L15). This meant that any script that was present in the plugin view template could be executed. After upgrading the mithril 1.1.6, this behaviour was broken even though we use `m.trust` for rendering the plugin view template.  This is due to a breaking change introduced when we moved from mithril 0.2.3 to 1.1.6. It is being followed in https://github.com/MithrilJS/mithril.js/issues/2168. Once there is a resolution upstream, we can upgrade.

###### Basic environment details

* Go Version: `18.5.0`
* Browser vendor and version (if relevant): `Google Chrome 66.0.3359.139, Firefox 59.0.2`

##### Steps to Reproduce

1. Refer to https://github.com/MithrilJS/mithril.js/issues/2168 

@GaneshSPatil , @arvindsv - please add/edit this if I have missed any details. Thank you.

arvindsv:
@GaneshSPatil At some point soon, we should see if [this workaround](https://github.com/MithrilJS/mithril.js/issues/2168#issuecomment-434457083) works ...

/cc @maheshp

adityasood:
fixed by #5361

PR #5361, opened 2018-11-01
Allow script tags to be defined in plugin settings (#4797)

GaneshSPatil:
Add the following script tag in any plugin settings to test:
```
<script>window.onload = alert("Script works");</script>
```

**NOTE:** Script tag needs to be within the main container.

Eg:
```html
<div data-plugin-style-id="kubernetes-plugin">
	<style></styles>
        <div class="row"> plugin settings</div>
        <script>window.onload = alert("Script works");</script>
</div>
```

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/shared/angular_plugin.js.msx
+++ b/server/webapp/WEB-INF/rails/webpack/views/shared/angular_plugin.js.msx
@@ -110,24 +110,38 @@ function onUpdate(scope, args) {
   };
 }
 
+const oncreate = function (vnode) {
+  vnode.children.forEach((elem) => {
+    if (!(elem.dom instanceof Element)) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/rails/webpack/views/shared/angular_plugin.js.msx
```diff
@@ -110,24 +110,38 @@ function onUpdate(scope, args) {
   };
 }
 
+const oncreate = function (vnode) {
+  vnode.children.forEach((elem) => {
+    if (!(elem.dom instanceof Element)) {
+      return;
+    }
+
+    elem.dom.querySelectorAll("script").forEach((script) => {
```
GaneshSPatil: [Do Not Merge] Expecting a failure on IE11 due to these lines.

GaneshSPatil:
Moving it to `18.12.0`

PR #5363, opened 2018-11-02
Move config-repo v1 API to Spark

marques-work:
This PR officially replaces the rails `config_repos_controller.rb`. It also includes some refactoring/cleanup around the `CrudController` interface. Hopefully, an improvement.

- [x] index
- [x] show
- [x] create
- [x] update
- [x] delete


kierarad:
One more thing:

- [ ] return 304 on show if etag matches
- [ ] return 422 on GoConfigInvalidExceptions (already there?)

config/config-api/src/main/java/com/thoughtworks/go/config/Cacheable.java
```diff
@@ -0,0 +1,29 @@
...
+package com.thoughtworks.go.config;
+
+import static com.thoughtworks.go.util.CachedDigestUtils.sha256Hex;
+
+public interface Cacheable {
+    default String etag() {
+        return sha256(Integer.toString(hashCode()));
+    }
+
+    default String sha256(String subject) {
```
akshaydewan: We end up allowing subclasses to override this method, which doesn't seem right to me. Maybe just inline it?

spark/spark-base/src/test/java/com/thoughtworks/go/api/base/JsonUtils.java
```diff
@@ -34,6 +34,14 @@ public static String toObjectString(Consumer<OutputWriter> consumer) {
...
     }
 
+    public static String toObjectStringWithoutLinks(Consumer<OutputWriter> consumer) {
+        return new JsonOutputWriter(new StringWriter(1024), null).forTopLevelObject(consumer).writer.toString();
+    }
+
+    public static Map<String, ?> toObjectWithoutLinks(Consumer<OutputWriter> consumer) throws IOException {
+        return JsonOutputWriter.OBJECT_MAPPER.readValue(toObjectStringWithoutLinks(consumer), Map.class);
+    }
+
```
akshaydewan: I noticed that we're not sending links anymore in error responses for ArtifactStoreConfig and Roles. Is this intentional? Seems minor, but I wonder if it would affect any 3rd parties consuming the API. Even if it doesn't warrant a version bump, we should probably mention it in the release notes.

akshaydewan:
LGTM other than the two minor comments I added.

config/config-api/src/main/java/com/thoughtworks/go/config/Cacheable.java
marques-work: @akshaydewan are you referring to making this not default (i.e., abstract)? We could do that, but I think a sha256 of the hashcode is a sensible default, so long as you implement `hashCode()`. Many of our models do.

If you’ve strong objections, I’m happy to inline that and leave the method abstract in the `Cacheable` interface.

spark/spark-base/src/test/java/com/thoughtworks/go/api/base/JsonUtils.java
marques-work: @akshaydewan

We do include links still AFAIK. The place where we remove links is when errors happen and we return an error message with a data key, holding the payload. I don’t think there’s any documentation for that behavior? It doesn’t make sense to send links back within an error message IMO.

But do let me know if I’m mistaken and the links are *not* being returned on GET and INDEX. That would be unintentional.

marques-work: To be clear, I’m referring to these error messages:
```
{
  message: “error occurred!”,
  data: {
    // request payload goes here, no need to include links because it’s just dumping what the user sent
  }
}
```

server/webapp/WEB-INF/rails/app/presenters/api_v1/shared/config_origin/config_repo_summary_representer.rb
```diff
@@ -20,16 +20,22 @@ module ConfigOrigin
       class ConfigRepoSummaryRepresenter < BaseRepresenter
         alias_method :config_repo, :represented
 
+        def urlFor(url_builder, path)
```
GaneshSPatil: I see this block of code has been repeated at many places, ([Pipeline Representer](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/presenters/api_v1/dashboard/pipeline_representer.rb#L35-L55), [Pipeline Config Summary Representer](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/presenters/api_v2/config/pipeline_config_summary_representer.rb#L23-L35)) can this method be pushed under `BaseRepresenter`? so that we can reuse it rather than defining at each representer?

config/config-api/src/main/java/com/thoughtworks/go/config/Cacheable.java
akshaydewan: I'll rephrase - From an interface point of view, having a `Cacheable` object implement/override `etag()` seems reasonable, but having  it implement/override `sha256` seems like a smell to me. It looks as though we're using `sha256` more like a utility function in the interface, in which case, we can call `com.thoughtworks.go.util.CachedDigestUtils.sha256Hex` directly. Or am I missing something?

spark/spark-base/src/test/java/com/thoughtworks/go/api/base/JsonUtils.java
akshaydewan: It's more a matter of consistency IMO. And in case of errors,  users may still find the `doc` link helpful.

server/webapp/WEB-INF/rails/app/presenters/api_v1/shared/config_origin/config_repo_summary_representer.rb
marques-work: That seems like a good improvement. I’ll take care and if it

akshaydewan:
Hey folks, I've squashed and merged this PR, since there was SPA work dependent on it. We can address the comments in subsequent commits.

config/config-api/src/main/java/com/thoughtworks/go/config/Cacheable.java
marques-work: Ah I get what you meant. Yeah I can certainly see your point. I mostly put it there for convenience for any other class that wants to implement cacheable. But I probably should have Inlined it instead since it’s already a utility method.

spark/spark-base/src/test/java/com/thoughtworks/go/api/base/JsonUtils.java
marques-work: We can certainly document this change if the old behavior is documented at all. I’ll look to see if we do.

marques-work:
@akshaydewan I will push another PR to address the issues here.

PR #5377, opened 2018-11-10
Migrate `AjaxPoller` to TS

ketan:


server/webapp/WEB-INF/rails/webpack/helpers/ajax_poller.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import * as CONSTANTS from "helpers/constants";
+import * as _ from "lodash";
+import * as m from "mithril";
+import {Stream} from "mithril/stream";
+import * as stream from "mithril/stream";
+
+export interface Options<T> {
+  intervalSeconds: number;
+  inSeconds: number;
```
akshaydewan: Is this the initial interval? Can we rename it to `initialIntervalSeconds` to make it clear?

server/webapp/WEB-INF/rails/webpack/helpers/api_poller_spec.ts
```diff
@@ -0,0 +1,171 @@
+/*
```
akshaydewan: File name should be `ajax_poller_spec`

PR #5383, opened 2018-11-12
System Notifications in typescript

GaneshSPatil:
Tasks:
  - [x] Migrate System Notifications to typescript
    - [x] System Notification
    - [x] Data Sharing Notification
  - [x] Migrate System Notification Component widget to Typescript
  - [ ] ~~Move Version Updater to Typescript~~ (will be done in a separate PR)
  - [x] Remove JQuery usages
    - [x]  Use javascript promise instead of `$.Deferred`
  - [ ] ~~Create Model_mixin helper class which can be extended by models instead of dynamically defining methods.~~ (will be done in a separate PR)
  - [x] Migrate specs


akshaydewan:
Just a thought - It took me a while to understand how the data was flowing and how the notifications were getting created/added. In most of our views, we have the models do an AJAX call and the widgets display data (nice and simple), but in cases like these, there's clearly the need of an orchestrator - something like a service, if you will.

I don't have a clear design in mind, but I feel like there are definitely some improvements that can be made in terms of ease of reading/understanding this kind code. Thoughts/suggestions are most welcome.

GaneshSPatil:
> Just a thought - It took me a while to understand how the data was flowing and how the notifications were getting created/added. In most of our views, we have the models do an AJAX call and the widgets display data (nice and simple), but in cases like these, there's clearly the need of an orchestrator - something like a service, if you will.
> 
> I don't have a clear design in mind, but I feel like there are definitely some improvements that can be made in terms of ease of reading/understanding this kind code. Thoughts/suggestions are most welcome.

@akshaydewan -- agree on this with you. I just ported the code over as is to typescript. Didn't think about changing the models. Will work on it.

maheshp:
Moved this to 18.12.0 as we are close to a release.

PR #5388, opened 2018-11-13
Ignore disable/delete agent request from plugin in absence of corresp…

maheshp:
…onding agents

* With the introduction of job_completion notification for the elastic
  agent plugins, there could be a scenario wherein a plugin can make a
  request to delete an agent through both the ping as well as the
  job_completion call. This check ensures only available agents are
  disabled/deleted.


rahulpargaonkar:
Tested a this issue for docker elastic agent. Its works fine as no exception message when agent delete request is created.

PR #5392, opened 2018-11-14
Introduce SparkUrlAware module to abstract url building for Spark controllers

marques-work:
Usage: the following method is available in all api presenters

`spark_url_for(opts, SparkRoutes::ConfigRepos::id("foo"))`

Followup to #5363.


marques-work:
Merging, jasmine failures seem to be IE related and not related to these changes.

PR #5395, opened 2018-11-14
Change contract of API calls for error handling

akshaydewan:
Epic #5232 

This changes the contract of the API calls to always return a `Promise<ApiResult<T>>`. The ApiResult can either be a success or error. If the Promise is rejected, the `catch` block in the `ApiRequestBuilder` will handle it, and as such, consumers of the `ApiRequestBuilder` can assume that they don't need to add a `catch` of their own in most circumstances.

The `ApiResult` class exposes a functional interface, similar to `Either` and `Result` used in FP.


akshaydewan:
I've merged this, since it's likely to create conflicts the longer is waiting. And I've already resolved conflicts twice on this branch :)

PR #5402, opened 2018-11-15
Adding simple panel component

bdpiparva:
- This is added for elastic profiles SPA as it groups profiles based on
  elastic plugin.

screen shot 2018-11-15 at 11 56 02 am

/cc @naveenbhaskar


ketan:
Can you please make sure the alignment is proper :)

ketan:
To be more specific:

- ensure that the text "This is the panel {header|body}" moves towards the left.
- the buttons on the right move towards the right.
- the "primary button" vertically center — see the plugins page for "status report" icon if you want "inspiration"

GaneshSPatil:
@bdpiparva @vishaldevgire -- We should be reusing the existing `CollapsiblePanel` component, with an additional class to support `headerPanel` styles.

Also, the component can be tweaked to support `neverCollapse` attribute for `HeaderPanel` which are always **expanded**.

Note: On the `Elastic Profiles SPA`, the header panel should be expand-collapsible, it will help to see all the profiles under one plugin.

\cc: @akshaydewan @ketan @naveenbhaskar

PR #5404, opened 2018-11-15
[Don't merge] Use the old menu and footer on new pages for consistency.

akshaydewan:
This brings in foundation css and js in the new pages, along with some (temporary) style changes. Revert this commit to bring back the new menu.


akshaydewan:
I'm working on a change to use the new header and footer styles on all the old pages [on this branch](https://github.com/akshaydewan/gocd/tree/components/new-header-everywhere). Leaving this PR as it as, so that we have the option to go either way.

maheshp:
Not required, we are using the new menu across all the pages.

Issue #5391, opened 2018-11-13
Backup message never goes away from dashboard


arvindsv:
##### Issue Type

- Bug Report

##### Summary

When a backup happens, the dashboard page shows a message "Server is under maintenance mode, please try later.", which never goes away after the backup is complete.

###### Basic environment details

* Go Version: 18.11.0 (7876-cda55765809491e150a03bbd7492f3447b38b60a)
* JAVA Version: 10.0.2
* OS: Mac OS X 10.14.1
* Browser vendor and version (if relevant): Tried on Chrome and Firefox.

##### Steps to Reproduce

1. Go to dashboard.
2. Run a backup (which takes some time). I did this:

   ```
   echo -n 20 >/tmp/time
   ```

   and then used this patch:

    ```diff
    diff --git a/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java b/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
    index 1ccd22146..99aead777 100644
    --- a/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
    +++ b/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
    @@ -31,6 +31,7 @@ import com.thoughtworks.go.server.service.result.HttpLocalizedOperationResult;
     import com.thoughtworks.go.server.web.BackupStatusProvider;
     import com.thoughtworks.go.serverhealth.HealthStateType;
     import com.thoughtworks.go.service.ConfigRepository;
    +import com.thoughtworks.go.util.FileUtil;
     import com.thoughtworks.go.util.SystemEnvironment;
     import com.thoughtworks.go.util.TimeProvider;
     import com.thoughtworks.go.util.VoidThrowingFn;
    @@ -45,6 +46,7 @@ import org.springframework.beans.factory.annotation.Autowired;
     import org.springframework.stereotype.Service;
     
     import java.io.*;
    +import java.nio.charset.StandardCharsets;
     import java.util.ArrayList;
     import java.util.Collection;
     import java.util.Date;
    @@ -236,6 +238,14 @@ public class BackupService implements BackupStatusProvider {
         }
     
         private void backupDb(File backupDir) {
    +        try {
    +            Integer s = Integer.valueOf(FileUtils.readFileToString(new File("/tmp/time"), StandardCharsets.UTF_8));
    +            System.err.println(new Date() + ": Sleeping for " + s + " seconds ...");
    +            Thread.sleep(s * 1000);
[...]

PR #5407, opened 2018-11-16
Render the dashboard again after server comes out of maintainance mode (#5391)

GaneshSPatil:
* Forcefully clear the dashboard cache if the server sends 503,
  so that on the next successful dashboard api response, the page can be re-rendered.


GaneshSPatil:
@vishaldevgire -- can you review this PR?

vishaldevgire:
@GaneshSPatil  Looks good to me.

Issue #5408, opened 2018-11-16
Migrate versions api from rails to spark


sanaj7:
##### Issue Type

- Feature enhancement

##### Summary

Migrate the version api from rails to spark.

###### Basic environment details

* Go Version: `18.11.0 (7941-d152a7802acb9c7c75ed6b845ba907dcbb550725)`
* JAVA Version: `JDK 8`

maheshp:
Done as part of https://github.com/gocd/gocd/pull/5409

adityasood:
@rajiesh @maheshp I guess this can be closed?

maheshp:
@adityasood yes can be closed if verified the changes.

adityasood:
verified on 18.12.0-8158

PR #5409, opened 2018-11-16
Version api in spark (v1) (#5408)

sanaj7:


api/api-version-v1/src/main/java/com/thoughtworks/go/apiv1/version/VersionControllerV1.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void setupRoutes() {
+        path(controllerBasePath(), () -> {
+            Spark.get("", mimeType, this::get);
+        });
+    }
+
+    public String get(Request req, Response res) throws IOException {
```
ketan: Should be called `show` — which is what the action is doing. `GET` is an http verb.

api/api-version-v1/src/main/java/com/thoughtworks/go/apiv1/version/VersionRepresenter.java
```diff
@@ -0,0 +1,39 @@
...
+public class VersionRepresenter {
+
+
+    public static void toJSON(OutputWriter jsonWriter, CurrentGoCDVersion currentGoCDVersion){
+
+        jsonWriter.addLinks(outputLinkWriter -> outputLinkWriter
+                        .addLink("self",Routes.Version.BASE)
+                        .addAbsoluteLink("doc",Routes.Version.DOC));
+
+        jsonWriter.add("version",currentGoCDVersion.goVersion());
```
ketan: Please indent code :) `CMD+OPTION+L` in intellij

api/api-version-v1/src/test/groovy/com/thoughtworks/go/apiv1/version/VersionRepresenterTest.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+  @BeforeEach
+  void setUp() {
+    currentGoCDVersion = new CurrentGoCDVersion()
+  }
+
+  @Test
+  void 'should render a version with hal representation'() {
+    def actualJson = toObjectString({ VersionRepresenter.toJSON(it, currentGoCDVersion) })
+    print(actualJson)
```
ketan: No print statements.

server/webapp/WEB-INF/rails/config/routes.rb
```diff
@@ -256,8 +256,6 @@
 
       get 'dashboard', controller: :dashboard, action: :dashboard, as: :show_dashboard
 
-      match 'version', controller: :version, action: :show, as: :version, via: %w(get head)
```
ketan: I don't think the `head` was intentional. I suppose this support was removed, which is OK.

ketan: s/intentional/unintentional/

sanaj7: intentional : we don't need the meta information in this case in the http head

sanaj7:
> LGTM. Can you make sure to squash commits and sign the CLA?

Should I squash all the commits into one ?

api/api-version-v1/src/test/groovy/com/thoughtworks/go/apiv1/version/VersionRepresenterTest.groovy
```diff
@@ -0,0 +1,41 @@
...
+import static com.thoughtworks.go.api.base.JsonUtils.toObjectString
+import static net.javacrumbs.jsonunit.fluent.JsonFluentAssert.assertThatJson
+
+class VersionRepresenterTest {
+
+  private CurrentGoCDVersion currentGoCDVersion;
+
+  @BeforeEach
+  void setUp() {
+    currentGoCDVersion = new CurrentGoCDVersion()
```
GaneshSPatil: This field should be mocked. This is calling the real `CurrentGoCDVersion` object, which value will keep changing with every new commit.

Issue #5394, opened 2018-11-14
Cannot update ConfigRepo through API


maheshp:
##### Issue Type

- Bug Report

##### Summary

Updates of config repo through [API](https://api.gocd.org/current/#update-config-repo) does not work. The json returned is same as before update.

###### Basic environment details

* Go Version: `18.11.0 - latest experimental`

rajiesh:
Issues in config repo to be fixed for 18.11 release:

- Appropriate error message not provided in response JSON if same material url is used in multiple config repo,

Create config `repo1` using json,

```{
    "id": "repo1",
    "plugin_id": "json.config.plugin",
    "material": {
      "type": "git",
      "attributes": {
        "url": "http://github.com/test-project",
        "auto_update": true,
        "branch": "master"
      }
    }
  }
```

Create config `repo2` using json(the material url is same as repo1)

```{
    "id": "repo2",
    "plugin_id": "json.config.plugin",
    "material": {
      "type": "git",
      "attributes": {
        "url": "http://github.com/test-project",
        "auto_update": true,
        "branch": "master"
      }
    }
  }
```

Second request fails with response as below, response do not contain the appropriate error message

```{
    "message": "Validations failed for config-repo 'repo2'. Error(s): [Validation failed.]. Please correct and resubmit.",
    "data": {
        "id": "repo2",
        "plugin_id": "json.config.plugin",
        "material": {
            "type": "git",
            "attributes": {
                "name": null,
                "auto_update": true,
                "url": "http://github.com/test-project",
                "branch": "master"
            }
        },
        "configuration": []
    }
}
```

- Invalid material type returns 500 server error,

Submit a create config repo request as below,

```
{
    "id": "repo2",
    "plugin_id": "json.config.plugin",
    "material": {
      "type": "GIT",
      "attributes": {
        "url": "http://github.com/test-project",
        "auto_update": true,
        "branch": "master"
      }
    }
  }
```

API returns 500 server error. Error in logs,

```
2018-11-14 16:24:13,091 ERROR [qtp84739718-14] GeneralError:54 -
java.lang.IllegalArgumentException: Unsupported material type: GIT
	at com.thoughtworks.go.apiv1.configrepos.representers.MaterialRepresenter.fromJSON(MaterialRepresenter.java:114)
	at com.thoughtworks.go.apiv1.configrepos.representers.ConfigRepoConfigRepresenterV1.fromJSON(ConfigRepoConfigRepresenterV1.java:37)
[...]

PR #5410, opened 2018-11-19
Do not fail parsing of JSON if a json attribute is missing

ketan:


maheshp:
Fixes issue around the validation errors not showing up - https://github.com/gocd/gocd/issues/5394

PR #5411, opened 2018-11-19
Align plugins widget row element items (#5232)

GaneshSPatil:
screen shot 2018-11-19 at 1 17 32 pm

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
@@ -26,6 +26,7 @@ export interface Attrs {
   // pass in an object if you do not care about order, pass an array of key-value pairs if you care about order
   data: { [key: string]: m.Children } | m.Children[];
   inline?: boolean;
+  keyValuePairItemClass?: string;
 }
 
 export class KeyValuePair extends MithrilViewComponent<Attrs> {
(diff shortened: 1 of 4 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
```diff
@@ -26,6 +26,7 @@ export interface Attrs {
   // pass in an object if you do not care about order, pass an array of key-value pairs if you care about order
   data: { [key: string]: m.Children } | m.Children[];
   inline?: boolean;
+  keyValuePairItemClass?: string;
```
GaneshSPatil: @akshaydewan @ketan -- Instead of passing `keyValuePairItemClass`, we can change the data structure of `data` from `json` to  `array of json`,
where, each data element will look like:
```json
{
  "key": "plugin id",
  "value": "my.kubernetes.plugin",
  "classes": ["plugin-row-element", "plugin-id"]
}
```
Changing the data structure to `array` will always enforce the order. and also will give an handle to configure individual key-value-pair item.

I will be sending another PR to fix this, as it was a little bigger change affecting other pages.

ketan: I thought we agreeed to not be able to inject classnames like this?

GaneshSPatil: @ketan -- Sometimes, we will need to style the specific component, and I think classname is the easy way to do it, rather than passing in the styled element.

If there is a better of doing this, we can change this code.

akshaydewan: I think we could expose a semantic attribute from the component, e.g. `isHeader` or something in this case, and then the component internally takes care of the styling.

akshaydewan: If we reach a point where we have too much customization needed, then it wouldn't be a good candidate for componentization, or we'd need to re-think the approach to compose components in that scenario.

I've taken a stab at aligning the items in this PR: #5420. Please take a look and let me know what you think. I think @naveenbhaskar had made a grid layout which we haven't yet incorporated into the project. If/when that is in place, we can possibly make further improvements to the layout.

PR #5418, opened 2018-11-20
Allow an XHR setter to be passed 

akshaydewan:
To get a handle on the current request, so that it can be cancelled from AjaxPoller


ketan:
LGTM.

PR #5419, opened 2018-11-20
Move agents API v4 to spark

bdpiparva:


api/api-agents-v4/src/main/java/com/thoughtworks/go/apiv4/agents/AgentsControllerV4.java
```diff
@@ -0,0 +1,208 @@
...
+        return null; // to be implemented
+    }
+
+    @Override
+    public Consumer<OutputWriter> jsonWriter(AgentInstance agentConfigs) {
+        return null;
+    }
+
+    private void checkSecurityOr403(Request request, Response response) {
+        if (Arrays.asList("PATCH", "PUT", "DELETE").contains(request.requestMethod().toUpperCase())) {
```
ketan: Can we invert this, so GET and HEAD are allowed, everything else is denied, prevents accidentally opening up anything else?

bdpiparva: Sure, I will do it

server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
```diff
@@ -125,4 +125,12 @@ export default class {
       return "/go/api/elastic/profiles";
     }
   }
+
+  static agentsPath(uuid: string): string {
```
ketan: the signature should be `agentsPath(uuid?: string): string`

PR #5425, opened 2018-11-20
Add required data-test attributes on the plugins spa (#5232)

GaneshSPatil:


server/webapp/WEB-INF/rails/webpack/views/components/collapsible_panel/index.tsx
```diff
@@ -26,6 +26,7 @@ const classnames = bind(styles);
 type AttributeType = m.Children;
 
 export interface Attrs<Header, Actions> {
+  dataTestId?: string;
```
GaneshSPatil: @akshaydewan -- this needs to explicitly added, if we do `{...vnode.attrs}`, that is adding other `attrs` as well such as `header`, `actions`, `error` etc.

server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
```diff
@@ -30,11 +30,16 @@ export interface Attrs {
 export class KeyValuePair extends MithrilViewComponent<Attrs> {
   view(vnode: m.Vnode<Attrs>) {
     const isInline = vnode.attrs.inline;
+
     const elements: JSX.Element[] = [];
     vnode.attrs.data.forEach((value, key) => {
+
+      const dataTestIdForKey   = `key-value-key-${key.replace(/ /g, "-").toLowerCase() as string}`;
```
GaneshSPatil: This is required for keys like `Supported operating systems`, etc

Issue #5190, opened 2018-09-20
Thoughts about "drain mode" for GoCD


arvindsv:
I've been asked about this, a few times: A good way to know when a GoCD server (the whole system actually) is ready and in a good place to be brought down for either a restart or an upgrade.

Considerations:

  - Don't be busy checking for material updates
  - Agents should all have finished work (should be idle or missing or lost contact)
  - There should be no "create-agent" calls waiting (elastic agents)
  - No new jobs should be assigned to agents
  - No new requests for triggering pipelines and approval of manual gates should be accepted (UI / API)
  - Should ideally show a message to all users

@gocd/committers What is missing? Is this a bad idea for some reason?

ketan:
> - There should be no "create-agent" calls waiting (elastic agents)

No way to know that. No agent is currently building seems like a more sensible option to me.

> - Should ideally show a message to all users

Likely in a more prominent location where users acknowledge seeing this. Aka not server health messages :)

> Is this a bad idea for some reason?

I don't see it as a bad idea.

arvindsv:
> Likely in a more prominent location where users acknowledge seeing this. Aka not server health messages :)

Yes, yes. Banner, etc. :)

> > There should be no "create-agent" calls waiting (elastic agents)
> No way to know that. No agent is currently building seems like a more sensible option to me.

I mean that the scheduling subsystem should be turned off. And we would show that there are jobs waiting for scheduling. If they have been picked up (in progress of scheduling), we could decide it's not ready.

I mentioned elastic agents explicitly to say that we shouldn't forget them.

ketan:
For clarity, here's what I'd think means "ready to shutdown" after admin clicks some "maintenance mode" button:

- show a visible message (Q: persist to db, in case someone shuts down the server midway?)
- stop all MDU
- stop all scheduling
- any agents that ping for work receive nothing (even if there are scheduled jobs that that can possibly be assigned). If create agent calls were made and elastic agents are booting up, they'll just sit idle, until the plugin decides to terminate them. This termination may as well happen after the server comes back online.

Once any agents that are currently building complete their jobs, we can notify the admins that the server is now drained and can be shut down. We may possibly even shutdown automatically on behalf of an admin, if they prefer so, although I personally don't like this option.

maheshp:
[...]

PR #5458, opened 2018-11-27
Do not assign scheduled jobs to agents during server drain mode (#5190)

GaneshSPatil:
* Disable buildAssignmentService during server drain mode.
* Do not send createAgent requests to elastic agent plugins for elastic jobs.

---
Spoke to @bdpiparva regarding the implementation.

Spring scheduled tasks don't have a way to conditionally run them based on a flag. 
##### Two ways to conditionally run tasks:
  - Check for the condition inside the task function. 
  - Use task schedulers to stop and reinitialize timer tasks based on some condition and listeners.

Stopping and starting spring timer tasks might affect the GoCD services as many tasks have state initialization. Also, it's not a good idea to stop timers and start them again.

Going ahead with checking condition inside the task function.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
@@ -68,6 +68,7 @@ public class BuildAssignmentService implements ConfigChangedListener {
     private final UpstreamPipelineResolver resolver;
     private final BuilderFactory builderFactory;
     private AgentRemoteHandler agentRemoteHandler;
+    private DrainModeService drainModeService;
     private final ElasticAgentPluginService elasticAgentPluginService;
     private final SystemEnvironment systemEnvironment;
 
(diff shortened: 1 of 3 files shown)


ketan:
> Spring scheduled tasks don't have a way to conditionally run them based on a flag.

There's a way to setup a factory for scheduled tasks. I believe we've done this in the past. `ModeAwareThreadPoolTaskScheduler` seems to do just that.

GaneshSPatil:
> > Spring scheduled tasks don't have a way to conditionally run them based on a flag.
> 
> There's a way to setup a factory for scheduled tasks. I believe we've done this in the past. `ModeAwareThreadPoolTaskScheduler` seems to do just that.

@ketan -- Yes, A scheduler factory can be set up to achieve the same. But the spring scheduler don't have an interface to conditionally `run` the `runnable`.

If we look at `modeAwareRunnable.run` method, it first checks for the condition and calls `runnable.run();`

Also, `ModeAwareRunnable` is used to disable all the timers and subsystems of GoCD. Whereas, during drain mode, we need to disable specific subsystems (allowing other tasks to run). Hence, individual tasks can be aware of the `drain-mode` and run conditionally.

\cc: @bdpiparva

PR #5467, opened 2018-11-28
ModeAwareFilter Fixes

GaneshSPatil:
* Consider '/remoting', '/agent-websocket' and 'add-on' request as whitelisted API requests
* Allow admins to toggle drain mode setttings during drain mode


server/src/main/java/com/thoughtworks/go/server/newsecurity/filters/ModeAwareFilter.java
```diff
@@ -40,6 +43,11 @@
     private final SystemEnvironment systemEnvironment;
     private DrainModeService drainModeService;
 
+    private static final OrRequestMatcher ALLOWED_DRAIN_MODE_REQUEST_MATCHER = new OrRequestMatcher(
+            new RegexRequestMatcher("/api/stages/[0-9]*/cancel", "POST", true),
```
ketan: Add the new stage operate api for cancel as well?

Issue #5374, opened 2018-11-08
Expose CR contract to plugins


ibnc:
Epic: #5175 (Thoughts about improving pipelines as code (config-repo))

Proposal/idea for improving pipeline as code. See https://github.com/gocd/gocd/issues/5338#issuecomment-437125598 and https://github.com/gocd/gocd/issues/5338#issuecomment-437133247

## summary
I'd like to propose that we expose the config repo contract (aka the [CR models](https://github.com/gocd/gocd/tree/master/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/configrepo/contract)) to the the plugins.  

There are a few possible benefits to doing this that I can see: 
*  It'd make serialization/deserialization easier on the plugin side ([example where it would help](https://github.com/tomzo/gocd-yaml-config-plugin/blob/master/src/main/java/cd/go/plugin/config/yaml/YamlConfigPlugin.java#L82))
*  It'd make it easier to create new repo config plugins. 
*  It'd help ensure the data structure is consistent/correct across the plugins.
   * along these same lines make it easier to test/validate  (instead of doing this: [test](https://github.com/tomzo/gocd-yaml-config-plugin/blob/master/src/test/java/cd/go/plugin/config/yaml/transforms/PipelineTransformTest.java) and [test files](https://github.com/tomzo/gocd-yaml-config-plugin/tree/master/src/test/resources/parts) seeing if it can be serialized to the CR model would likely be sufficient to validate)
* it'd help simplify exporting gocd structures on the plugin side

@ketan @arvindsv @marques-work @kierarad @tomzo

tomzo:
I am buying all the benefits for having the classes available as public library.
IMO the best solution is to create a separate module with CR classes, which has its own release cycle, meaning independent from GoCD core.
Then it is a reference for server or plugins or any tool that needs it.

I wouldn't fall back to using class-based plugin API, if that's what you are proposing.

ibnc:
@tomzo 

At this point, I don't care about the specifics until we decide to do this work. I'm just proposing that the classes are available to the plugins.

arvindsv:
Sorry to be caring only about the specifics. I don't see any harm in moving them out to their own module, which can be spun off as a separate module in https://search.maven.org/search?q=g:cd.go.plugin.

I don't know about it being outside of GoCD core, but if it is versioned properly, then GoCD core can treat it as a dependency. That should be ok too.

ibnc:
[...]

PR #5470, opened 2018-11-28
Expose Config Repo Contract to Plugins

ibnc:
This is extracting out the config repo contract and codecs into a separate module so that both the server and plugins will have access to them.

See #5374 for more details/discussion 

@arvindsv @ketan @maheshp I'm unsure what my next step should be as far as building and publishing this module...


ketan:
LGTM. As far as building and publishing this as a jar is concerned, we do that here:

https://github.com/gocd/build_utilities/blob/master/upload-to-maven#L71-L90

Basically, the installer pipeline publishes the api jar, which is then pulled down by the task above in `UploadInstallers` pipeline to publish to maven central.

For this PR:

You seem to have checked in the .class files from idea :)

Apart from this — I believe we need some discipline in maintaining the classes in this package, primarily because these are "API".

Is there some tooling available that'll break the build in case there's an API breakage? I'm familiar with http://javadiff.sourceforge.net/ and https://github.com/melix/japicmp-gradle-plugin, I'm sure there are other tools that do the job. These tools usually check the current jar against a known "reference" api jar and generate a api diff report.

ibnc:
@ketan lol, you're totally right. I just pushed a commit to remove those .class files.

ibnc:
@ketan I suspect that the [ConfigConverterTest](https://github.com/gocd/gocd/blob/master/server/src/test-fast/java/com/thoughtworks/go/config/ConfigConverterTest.java) will provide enough coverage to ensure that the module will be correct... What do you think?

ibnc:
What do you think about bundling go-plugin-api and go-plugin-config-repo into a single build artifact to make the changes on the build_utilities side simpler? It'd allow for a single download

ketan:
> @ketan I suspect that the [ConfigConverterTest](https://github.com/gocd/gocd/blob/master/server/src/test-fast/java/com/thoughtworks/go/config/ConfigConverterTest.java) will provide enough coverage to ensure that the module will be correct... What do you think?

Renaming something in the IDE will silently change the tests without us realizing that API has been broken.

ketan:
> What do you think about bundling go-plugin-api and go-plugin-config-repo into a single build artifact to make the changes on the build_utilities side simpler? It'd allow for a single download

There's a case to be made about similar interfaces for other plugin types. I feel having separate jars is probably better (so plugins can depend on the smallest possible subset of things).

ketan:
If you're looking for help publishing this module to maven central, I can help you walk through the code. I believe @arvindsv should be able to do that as well.

ibnc:
@ketan What is the Jar that I should be comparing to?

Also, I'm unsure how this will help ensure the api is working properly in the future. I can see how it'd ensure things are correct right now, but I don't see the benefit of adding a break point in the build for this check. 

Maybe I'm not understanding things. Do you have an example I could look at to help illuminate things?

arvindsv:
@ibnc - Ketan's on vacation. I can probably work with you on this tomorrow or day after, if you want.

ibnc:
@arvindsv Then I guess my question would be, is this required work before the PR can be merged? I think this can be follow up work that I can do after having a conversation with @ketan when he's back from vacation.

arvindsv:
I don't have enough context to merge it anyway, without working with you on it. So, I guess we can wait till Ketan's back.

ibnc:
@ketan @arvindsv Another thing to note is that the ConfigConverterTest hasn't changed beyond the imports changing. So the tests are still valid. The only renaming that was done is the package change. The actual class names are all the same. I don't think we need to worry about this for this PR.

@arvindsv if you're not comfortable reviewing/merging this PR, then would you mind taking a look @marques-work?

ketan:
LGTM. I'll merge this today (after talking to @arvindsv)

PR #5475, opened 2018-11-29
Added switch component.

bdpiparva:
### Off state

screen shot 2018-11-29 at 2 22 08 pm

### On state

screen shot 2018-11-29 at 2 22 16 pm


server/webapp/WEB-INF/rails/webpack/views/components/spinner/index.tsx
```diff
@@ -1,4 +1,4 @@
-/*
+  /*
```
GaneshSPatil: whitespace change?

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+import {MithrilComponent} from "jsx/mithril-component";
+import {Stream} from "mithril/stream";
+
+import {bind} from "classnames/bind";
+import * as styles from "./index.scss";
+
+const uuid4      = require("uuid/v4");
+const classnames = bind(styles);
+
+export interface Attrs {
```
ketan: Can we use (or extend) the `LabelAttrs` interface as in `input_fields.tsx`? If it works for checkboxes, it should work for switches too.

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+        {label}
+        <input id={switchId} type="checkbox"
+               {...vnode.attrs}
+               checked={vnode.attrs.field()}
+               onclick={vnode.state.onclick.bind(this)}
+               className={styles.switchInput}
+               data-test-id="switch-checkbox"/>
+        <label for={switchId} className={styles.switchPaddle} data-test-id="switch-paddle"/>
+      </div>);
+  }
```
ketan: From what I can see, this is the same as the existing checkbox widget, with a different classname. Why redo the entire component?

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx
bdpiparva: Sure.

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx
bdpiparva: Mostly, this has differrent style then the checkbox component. We are using the switch in data sharing SPA as well(to keep ui consistent with it.)

bdpiparva: @ketan - I am seeing following problems with using `checkbox` component from `input_fields.tsx` -

1. The `id` attribute is not exposed in checkbox component, In switch component we need the reference of id to provide `labelFor`
2. Exposing classname will cause the inconsistance in UI as there is a way to modify the style

Not changing anything in the PR. any concerns?

server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
```diff
@@ -158,3 +158,90 @@ input[type="checkbox"].form-control { //sass-lint:disable-line no-qualifying-ele
...
+$switch-large-paddle-size: 1.5rem;
+$switch-large-space-between: 0.25rem;
+$switch-large-translate-x: 2rem;
+$switch-large-width: 4rem;
+$switch-large-height: 2rem;
+$switch-small-width: 2rem;
+$switch-small-height: 1rem;
+$switch-small-paddle-size: 0.75rem;
+$switch-small-space-between: 0.1255rem;
+$switch-small-translate-x: 1rem;
```
ketan: Move this to top of the file.

server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
```diff
@@ -158,3 +158,90 @@ input[type="checkbox"].form-control { //sass-lint:disable-line no-qualifying-ele
...
+  height:             $switch-large-height;
+  cursor:             pointer;
+  display:            flex;
+  border-radius:      35px;
+  background-color:   $switch-off-color;
+  -webkit-transition: 0.4s;
+  transition:         0.4s;
+}
+
+.switch-paddle:after {
```
ketan: Move this to inside the `.switch-paddle` class above?

server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
```diff
@@ -158,3 +158,90 @@ input[type="checkbox"].form-control { //sass-lint:disable-line no-qualifying-ele
...
+  align-self:         center;
+  width:              $switch-large-width;
+  height:             $switch-large-height;
+  cursor:             pointer;
+  display:            flex;
+  border-radius:      35px;
+  background-color:   $switch-off-color;
+  -webkit-transition: 0.4s;
+  transition:         0.4s;
+}
```
ketan: Can be written as (I think)

```scss
.switch-paddle {
  ...
  &:after {
    ...
  }
  input:checked ~ &:after {
    ...
  }
}

```

server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
```diff
@@ -158,3 +158,90 @@ input[type="checkbox"].form-control { //sass-lint:disable-line no-qualifying-ele
...
+
+input:checked ~ .switch-paddle {
+  background: $switch-on-color;
+}
+
+.switch-small {
+  .switch-paddle {
+    width:  $switch-small-width;
+    height: $switch-small-height;
+  }
```
ketan: See comment above.

PR #5493, opened 2018-12-04
Show "unsupported browser" banner in the footer when using IE.

ketan:
z5f74d1f631cf855e5f1


ketan:
Currently tests only IE. I've skipped testing for a bunch of other browsers (opera, baidu, samsung, ...) But if we feel we need to change this implementation, it should be straightforward to change.

PR #5495, opened 2018-12-04
Drain mode info api (#5190)

GaneshSPatil:
* Create a drain mode info API to provide information about:
	- Running MDUs
	- is Server completely drained


GaneshSPatil:
@bdpiparva -- the PR build is green. Here: https://build.gocd.org/go/pipelines/value_stream_map/trigger/3905. Merging it.

PR #5496, opened 2018-12-04
Ui fixes

naveenbhaskar:


server/webapp/WEB-INF/rails/webpack/views/components/header_icon/index.scss
```diff
@@ -33,8 +35,11 @@ $collapse-icon-size: 30px;
 }
 
 .unknown-icon {
-  @include icon-before($type: $fa-var-question, $color: $icon-light-color, $margin: 5px, $size: 30px);
+  @include icon-before($type: $fa-var-question, $color: $icon-light-color, $size: 25px);
```
ketan: I think it's possible to pass `$margin:0` and remove the lines below.

Issue #5464, opened 2018-11-28
UI Issues for 18.12


akshaydewan:
This issue to track all the UI issues that need be fixed for the 18.12 release, so that we can prioritize things easily, and so that it's easier for @naveenbhaskar to keep track.

- [x] Config Repo forms layout needs to be fixed
- [x] Size of modals needs to be tweaked (there is too much white space either at the bottom or right side)
- [x] Add Data Sharing to the new Menu
- [x] Make Config Repo delete confirmation modal Small
- [x] Fix for long text in key-value-pair component
- [x] Feedback for the new Menu - the Security section seems odd.
- [x] For Internet Explorer 11 users, we need to show a message saying that the browser is not supported
- [x] UI for notifying users that Drain Mode is active
- [x] Revisit "danger button" color.
- [x] Move plugin ID from top into the details section on plugins SPA page.
- [x] Fix broken headers in history and Job details page
- [x] Inconsistent highlighting for menu items under admin (https://github.com/gocd/gocd/issues/5464#issuecomment-445298019)
- [x] 'Admin' menu is not clickable or hoverable  on IE edge browser.
- [x] 'GoCD YAML files pattern:' field is not displaying on expanding config repo even though it has a values.
- [x] On config repo spa , after couple of seconds, a same config repo displaying multiple times.

rahulpargaonkar:
Adding some UI issues on plugins page & config repo spa page
1. on plugins page, irregular spacing between plugin names & ids
<img width="1425" alt="screen shot 2018-11-28 at 10 29 08 pm" src="https://user-images.githubusercontent.com/33080845/49171172-f2770b80-f363-11e8-816f-eaf9c1d61048.png">

2. on config_repo SPA page, error symbol is miss-aligned on plugins page, it looks incorrect as compared to green tick symbol.
<img width="1430" alt="screen shot 2018-11-28 at 11 19 04 pm" src="https://user-images.githubusercontent.com/33080845/49171222-14708e00-f364-11e8-86ce-416ae30d34fc.png">

3. on config repo add/edit pop-up window, a text box sized window is highlighted when a checkbox is selected or not- selected.

4. when config repo expanded the the username is diplayed as null if not provided. Instead of null, can we specify either 'not specified'? or hide it? Or may be we can make username and password field interdependent?

ibnc:
[...]

PR #5498, opened 2018-12-04
Remove admin link for non admins

maheshp:
Fixes UI issue 8 & 9 in #5464


server/webapp/WEB-INF/vm/shared/_header.vm
```diff
@@ -53,7 +53,20 @@
     #if(!$current_tab && $currentTab.name)
         #set($current_tab = $currentTab.name)
     #end
-<body>
+    #set ($showAnalyticsDashboard = ($supportsAnalyticsDashboard && $userHasAdministratorRights))
```
akshaydewan: Maybe we can move this in the VelocityContext itself?

maheshp: Done

PR #5501, opened 2018-12-04
BugFix: KeyValuePair component should render number value properly

bdpiparva:
- The component was rendering number as '(Not specified)'. Ideally it
  should render the value of type number as it is.

### Before the fix
screen shot 2018-12-04 at 8 33 03 pm

### After the fix
screen shot 2018-12-04 at 8 32 21 pm

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
@@ -58,6 +58,10 @@ export class KeyValuePair extends MithrilViewComponent<Attrs> {
       return (<pre>{value.toString()}</pre>);
     }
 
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.tsx
```diff
@@ -67,7 +71,7 @@ export class KeyValuePair extends MithrilViewComponent<Attrs> {
     }
 
     // performat some "primitive" types
-    if (_.isString(value) || _.isNumber(value)) {
+    if (_.isString(value)) {
```
ketan: I continue to be amazed at the clusterfuck of a language that is javascript.

```
_.isEmpty(2)
=> true
```

PR #5505, opened 2018-12-05
Changed config_xml link for group admins #000

maheshp:
* Group Admins can view a snippet of the config_xml over the entire
  config. The config_xml was incorrectly linked to show the entire config,
  changed to pipeline snippets.


arvindsv:
@maheshp When you saw this, the group admin got a 403, right? It was just that the link was broken, but the permissions / security aspects were fine?

Issue #5370, opened 2018-11-06
email notification not sent when pipeline is cancelled due to reaching the job timeout


aqneves:
##### Issue Type

- Bug Report

##### Summary

Email notification not sent when pipeline is cancelled due to reaching the job timeout despite having subscribed all email notifications for that pipeline and having an email associated with the account. All other notifications are properly received.

##### Environment

Any environment.

###### Basic environment details

* Go Version: 18.3.0
* JAVA Version: 1.8.0_181
* OS: RHEL 7.5

###### Additional Environment Details

n/a

##### Steps to Reproduce

1. create pipeline X
2. subscribe to all notifications of pipeline X
3. configure a job for pipeline X that is a sleep of 2 minutes 
4. configure the job timeout to expire after 1 minute
5. run the pipeline

##### Expected Results
We should be notified that the pipeline has started and then we should be notified that the pipeline was cancelled

##### Actual Results
Email notification for pipeline starting was received.
Email notification for pipeline cancelling was not received.

##### Possible Fix
n/a

##### Log snippets
n/a

##### Code snippets/Screenshots
n/a

##### Any other info
n/a

rajiesh:
Verified on 18.12.0 (8203-bbeebad6c03497cf3f5bf0bcddbe13bbc08a11c8)

PR #5506, opened 2018-12-05
Send an email notification when job is cancelled due to timeout (#5370)

GaneshSPatil:
### Proposed Solution 1 (has problems)

~~Fix:~~
~~* Post a StageStatusTopic message when a job is completed regardless of the
 job state (passed, failed or cancelled).~~

~~Reason:~~
~~* StageStatusTopic was NOT invoked if the stage was not completed normally.
  A normal completion of the job involved only PASSED or FAILED stage state.~~
~~* Cancelled Stage due to timeout cancellation of a job scenario was missed
  while story implementation.~~

~~Older commit references (from cruise repo):~~
~~* 99dd349eefea4fb080b2c5146062207bfdcbda29~~
~~* 3ee6dc7cc125626565f46900cb51ffc4e0eaee66~~

---

### Proposed Solution 2 (has problems)

~~**Above mentioned implemention has a bug. When a stage is cancelled, the email notification would be sent multiple times, once for the stage-cancel event, and other for each job cancelled event.**~~

~~__When a job is cancelled, there is no cancelled cause (either cancelled by a user or by GoCD) added on the JobInstance. hence there is no way to stop bubbling up the job-cacelled events.__~~

~~#### New fix:~~
~~Send a stage cancelled event immediately once a job is cancelled due to inactivity. ~~

~~This implementation causes a Cancelled Stage notification to be sent even when other jobs in the stage are still building.~~

~~This is the similar case as of user cancelling the stage. Lets say a stage is cancelled by admin, the notification is sent even when jobs are performing oncancel taks.~~

---

#### Fix:
* Post a StageStatusTopic message when a job is completed regardless of the
     job state (passed, failed or cancelled).
* Do not explicitly post a StageStatusMessage when the job is cancelled by
      user. When a stage is cancelled, all jobs will be cancelled and
      JobStatusListener will send a StageStatusMessage on job completion.
      Eventually, a StageResultMessage will be sent when final job in the stage
      is cancelled. EmailNotificationListener listens to StageResultMessage,
      which will send the email when job is completely cancelled.

#### Reason:
* StageStatusTopic was NOT invoked if the stage was not completed normally.
      A normal completion of the job involved only PASSED or FAILED stage state.
* Cancelled Stage due to timeout cancellation of a job scenario was missed
      while story implementation.

#### Older commit references (from cruise repo):
* 99dd349eefea4fb080b2c5146062207bfdcbda29
* 3ee6dc7cc125626565f46900cb51ffc4e0eaee66

\cc: @maheshp


GaneshSPatil:
@maheshp @rajiesh -- Do Not Merge Until Functional specs are green.

maheshp:
@GaneshSPatil the third approach would not work :( . 

Consider a stage scheduled where none of the jobs have an agent assigned yet, on stage cancel users will not get a notification now since we no longer send a notification on stage cancel - https://github.com/gocd/gocd/pull/5506/files#diff-2ca83558267023df1078778e5f450f62L184 .

GaneshSPatil:
@arvindsv -- Yes it works.
In this scenario, the `StageService.cancelStage` method won't be called. 
This is handled by `JobStatusListener.onMessage` handler.

When the first job is finished and `JobStatusListener.onMessage` handler is called. The state of the stage is still `unknown` and hence it won't post a topic on the `stageStatusTopic`. [here](https://github.com/gocd/gocd/pull/5506/files#diff-ddd7dadf1000e8ce2203431c938db2b9R51)

On the second job completion due to timeout, The state of the job is `cancelled` and it will post a topic on the `stageStatusTopic`.

This will lead to sending the email only once, even if multiple jobs during the stage are cancelled.

arvindsv:
Ok, cool. I don't know if there are any cases around job reschedule, maybe not.

maheshp:
@GaneshSPatil wondering if the below scenario is covered.

1. Consider a Stage with a single job
2. The job is scheduled and assigned to an agent.
3. The agent dies while running a job.
4. Before the unresponsive job is cancelled by GoCD user cancels the stage.

Since the Stage has a job assigned to an agent with the current changes a email notification is not sent.

GaneshSPatil:
[DNM] @maheshp -- above-mentioned scenario does not work. 
\cc: @rajiesh @adityasood

GaneshSPatil:
@maheshp -- I've created a separate issue for the above-mentioned scenario.
Merging this PR. we will fix this issue in the upcoming releases of GoCD.

GaneshSPatil:
PR build - https://build.gocd.org/go/pipelines/value_stream_map/trigger/4006

PR #5524, opened 2018-12-07
Add `name` attribute to Header Icon and use it for `data-test-id`

akshaydewan:
I've only added the `alt` attribute for accessibility, since screen-readers parse that correctly and `aria-label` would be redundant.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/header_icon/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/header_icon/index.tsx
@@ -17,19 +17,23 @@
 import {MithrilComponent} from "jsx/mithril-component";
 import * as _ from "lodash";
 import * as m from "mithril";
+import * as s from "underscore.string";
 import * as styles from "./index.scss";
 
 interface Attrs {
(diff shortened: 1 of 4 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/header_icon/index.tsx
```diff
@@ -17,19 +17,23 @@
 import {MithrilComponent} from "jsx/mithril-component";
 import * as _ from "lodash";
 import * as m from "mithril";
+import * as s from "underscore.string";
 import * as styles from "./index.scss";
 
 interface Attrs {
   imageUrl?: string; //an image URL of the icon
-  dataTestId?: string;
+  name?: string;
```
ketan: I'd think accessibility should not be "optional" :)

akshaydewan: In this case, I think it is ok to be optional. The icon is usually a plugin icon which can be skipped on screenreaders. Only in case of Config Repos, the icon actually conveys some information.

PR #5528, opened 2018-12-07
Change menu order

akshaydewan:
Wait for feedback on the order before merging


arvindsv:
Isn't "User summary" part of "Security"?

maheshp:
> Isn't "User summary" part of "Security"?

Guess we were looking at Security just from the point of view of the `<security/>` tag which is wrong. Moving "User Summary" to Security makes sense. @akshaydewan WDYT?

PR #5532, opened 2018-12-07
Fix Dashboard rendering bug

GaneshSPatil:


server/webapp/WEB-INF/rails/webpack/single_page_apps/new_dashboard.js
```diff
@@ -224,6 +222,7 @@ $(() => {
     });
 
     renderView();
+    repeater().start();
```
ketan: Seem to be creating a new repeater every time. Is that intentional?

GaneshSPatil: @ketan -- this is moved under `onPluginInfosResponse` which will be called only once after the dashboard page is loaded.

PR #5533, opened 2018-12-07
Relayout the config repos page

ketan:
- show the plugin icon first
- move the indicator to the end (Naveen needs to make sure it renders OK)


ketan:
Adding screenshot:

screenshot_2018-12-07 config repos - go 1

@naveenbhaskar — can I get some help aligning the indicator icon on the right?

Issue #5512, opened 2018-12-06
Personalization tab(the left most tab) shows all pipelines before updating it to show relevant pipelines


rajiesh:
##### Issue Type

- Bug Report

##### Summary

This is an issue reported by @ibnc on https://github.com/gocd/gocd/issues/5464. Moving it here since this issue will be worked on separately.

**Note:** This issue is not reproducible in a local setup with few pipelines. In build.gocd.org we can observer it but intermittently. If anyone can reproduce it consistently please add more info.

Personalization tab(the left most tab) shows all pipelines before updating it to show relevant pipelines.

When user click the Go icon on the top left, it takes me back to the furthest left tab, but the content of the page is incorrect.

<img width="1438" alt="isabelle-issue-5464" src="https://user-images.githubusercontent.com/6533924/49566488-c73a8080-f950-11e8-8b87-f5b55aa9cc2d.png">

<img width="778" alt="issue-54640-ss-2" src="https://user-images.githubusercontent.com/6533924/49566495-cb669e00-f950-11e8-9ba7-723cd51ac490.png">

A GIF which shows the issue,

![issue-5464-gif-3](https://user-images.githubusercontent.com/6533924/49566502-d3264280-f950-11e8-8f79-18010cc4d791.gif)

adityasood:
verified on 18.12.0-8222

PR #5534, opened 2018-12-08
Revert a2b17483d188cc6470f3d90174cbdc920752d651

arvindsv:
Revert "Show the first personalization tab pipelines on dashboard load (#5512) (#5527)"

This reverts commit a2b17483d188cc6470f3d90174cbdc920752d651.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/single_page_apps/new_dashboard.js
+++ b/server/webapp/WEB-INF/rails/webpack/single_page_apps/new_dashboard.js
@@ -23,7 +23,6 @@ const DashboardWidget = require('views/dashboard/dashboard_widget');
 const PluginInfos     = require('models/shared/plugin_infos');
 const AjaxPoller      = require('helpers/ajax_poller').AjaxPoller;
 const PageLoadError   = require('views/shared/page_load_error');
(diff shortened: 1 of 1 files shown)


arvindsv:
@GaneshSPatil Reverting this, because it breaks [this spec](https://build.gocd.org/go/tab/build/detail/regression-SPAs/1075/Firefox/1/regression-runInstance-6#tab-gaugereport).

What's happening is that, with this fix, if you go to a view which is not "Default" and refresh, you'll go to "Default":

Refreshing on https://my-gocd-server/go/pipelines?viewName=MyView1#!/
...
takes you to: https://my-gocd-server/go/pipelines?viewName=Default#!/ (or the left-most view).

And this part of the functional specs is doing [a reload](https://github.com/gocd/ruby-functional-tests/blob/599e3b8491bd01b7693eb3302abd22afd4cb64fa/lib/pages/new_pipeline_dashboard_page.rb#L84) and then was verifying that that view was unchanged ([here](https://github.com/gocd/ruby-functional-tests/blob/599e3b8491bd01b7693eb3302abd22afd4cb64fa/specs/PersonalizationView.spec#L65)).

PR #5536, opened 2018-12-10
Remove unnecessary IE8 check from VSM controller

GaneshSPatil:
* VSM page redirects to pipeline-dependency page if loaded on IE8
* Remove explicit IE8 check because:
	- IE8 is an unsupported browser.
	- A generic unsupported browser message is shown on all pages
	  if application is loaded on IE.


GaneshSPatil:
\cc: @maheshp @ketan

PR #5547, opened 2018-12-11
admin link subnav fix

naveenbhaskar:


server/webapp/WEB-INF/rails/webpack/views/components/site_menu/index.scss
```diff
@@ -75,8 +73,21 @@
     }
   }
 
-  &:hover, &:focus, &:focus-within {
-    // background: $sub-navigation-bg;
+  &:hover {
+    @media(min-width: $screen-md) {
```
ketan: consider making it a mixin with a very obvious name (`because-ms-edge-has-some-bug`) to avoid someone changing one block of code and breaking something?

PR #5549, opened 2018-12-11
Drain Mode enable/disable API (#5190)

GaneshSPatil:
* Change drain mode API from PATCH to POST


GaneshSPatil:
@bdpiparva -- I will be sending another PR to merge the settings response into info API response and get rid of settings API. As part of that PR, I will fix the javascript.

Issue #5010, opened 2018-08-07
Loading many Config Repo Pipelines takes a long time


dziemba:
##### Issue Type

- Feature enhancement

##### Summary

When restarting a GoCD instance with many config repo pipelines, it takes a long time for all of them to show up.

This is confusing to users ("where did my pipeline go?") and breaks dependencies (pipeline A depends on B, which is not available yet).

###### Basic environment details

* Go Version: 18.7.0
* JAVA Version: 1.8
* OS: Debian

###### Additional Environment Details

- 315 config repos
- 470 pipelines from these config repos
- 300 XML pipelines

##### Steps to Reproduce

1. Create a GoCD instance with many config repos
2. Wait until all of them show up properly
3. Restart GoCD
4. Look at pipeline overview, to see which pipelines show up

##### Expected Results
I expect all pipelines to show up immediately.

##### Actual Results
XML pipelines show up immediately. Config repo pipelines only show up after some time has passed. For many pipelines/repos, this is significant: it takes 20min for everything to show up again.

##### Possible Fix

I assume git materials are refreshed before parsing config pipelines on a fresh boot. Maybe cache pipeline definitions and then asynchronously refresh from git?

I'm happy to provide more details if needed.

arvindsv:
I remember discussions with @tomzo (the author of this) a long time ago - and we decided not to go down that path. Maybe @tomzo remembers more details? The flyweights do exist across restarts.

Given that we've discussed it before, I'm going to mark this as wont-fix for now. Happy to reopen if @tomzo thinks this should be reconsidered.

arvindsv:
Actually, I'll leave it open till we decide. Thanks.

tomzo:
@dziemba I can confirm this becomes an issue with 100+ config repos. For start I suggest you play around with following system properties:
 * `material.check.threads` which is the number of concurrent material checks which are not defined by config repos.
 * `material.config.check.threads` -  which is the number of concurrent material checks which are defined by config repos. The default is `2` which means your 350 config repos are handed by only 2 workers on start unless you already changed this.

You can set system properties with something like this:
```
export GO_SERVER_SYSTEM_PROPERTIES="-Dmaterial.check.threads=2 -Dmaterial.config.check.threads=8"
```

By changing above values I could keep our start time under 4 minutes. @dziemba please tell me if you have experimented with these settings, if not then please try and report back.

> Maybe cache pipeline definitions and then asynchronously refresh from git?

[...]

PR #5550, opened 2018-12-11
Config repo performance

ibnc:
This PR multithreads the config repo processing. It will now checkout the repo, parse the config, and merge the config in parallel (Merging still has a lock, so that bit will be in serial).

This PR addresses https://github.com/gocd/gocd/issues/5010

@ketan @tomzo

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/materials/ConfigMaterialUpdateCompletedMessage.java
+++ b/server/src/main/java/com/thoughtworks/go/server/materials/ConfigMaterialUpdateCompletedMessage.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.server.materials;
+
+import com.thoughtworks.go.domain.materials.Material;
+
+public class ConfigMaterialUpdateCompletedMessage extends MaterialUpdateCompletedMessage {
+    public ConfigMaterialUpdateCompletedMessage(Material material, long trackingId) {
+        super(material, trackingId);
+    }
+}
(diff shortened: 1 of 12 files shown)


arvindsv:
@ibnc Can you explain the problem too? Previously, there were multiple listeners (configured, at least) as well. Was there a lock at the wrong place?

ibnc:
> @ibnc Can you explain the problem too? Previously, there were multiple listeners (configured, at least) as well. Was there a lock at the wrong place?

Previously, there were multiple listeners that updated the materials, but only one listener for checking out, parsing, and merging the config repo updates from the materials. The multiple listeners for config repo materials were dedicated material update listeners that would only update the material in the DB. So increasing the number of those listeners would only have a small effect on the over all performance. The actual checking out, parsing, and merging were done in serial.

ibnc:
@tomzo 

You'd be the best person to ask about the initial check out and parsing work. 

I'm currently looking through the epic https://github.com/gocd/gocd/issues/1133 but there's a lot to parse through there.

tomzo:
The PR which added second queue for config-repos and the config repo updater - https://github.com/gocd/gocd/pull/1333 
The design was described here https://github.com/gocd/gocd/issues/1133#issuecomment-116826923

This was over 3 years ago so I don't recall all of it, and code might have a changed since then. But from what I written there:
> ConfigMaterialUpdater - new component
Added new component - ConfigMaterialUpdater which listens on config-material-update-
completed topic. So when MDU is done then ConfigMaterialUpdater gets its chance to work with material being updated:
 -It uses MaterialRepository to check if there were any changes
 -It uses existing pollers code to checkout material to directory
 -It calls GoRepoConfigDataSource (where parsing happens) and when done
 -It posts to material-update-completed which is picked up by MaterialUpdateService using standard procedure as if this was an old-school material. This removes material from inProgress status.

So you are making (previously singleton) `ConfigMaterialUpdater` into multiple listeners, and then rely on  locking in merge mechanism.
OK, I am starting to get your point.

tomzo:
Ping me once you've fixed the failing tests. 
I'd like to check if this is not breaking anything by starting a test server with all my config-repos and see if it will get to valid config after parsing all of them.

ibnc:
> Ping me once you've fixed the failing tests.
> I'd like to check if this is not breaking anything by starting a test server with all my config-repos and see if it will get to valid config after parsing all of them.

That'd be a great help :)

ibnc:
@tomzo Fixed the tests

ibnc:
@tomzo :D 

Thank you for helping!

tomzo:
Can you rebase and squash the commits into 1?

ibnc:
Can do

arvindsv:
@maheshp and I will probably take a look at the whole of MDU and see what can be done to make it simpler, if possible. Even before this change, it was a little too complicated and convoluted.

ibnc:
@arvindsv it's def too complicated/convoluted

@maheshp technically, yes it will reduce the number of listeners for material updates. That being said, ConfigMaterialUpdateListener will not affect the rate that the MaterialUpdateListener listeners will process. 

Perhaps we should raise the default number of MaterialUpdateListeners to 3 or 4?

arvindsv:
@ibnc Do you mean _by_ 3 or 4? The default number of listeners is 10. Config update listeners were 2, for a total of 12. Now, the total is 10, as far as I understand.

PR #5551, opened 2018-12-12
Consolidate ConfigRepoOperationsControllerV1 endpoints into ConfigReposInternalControllerV1

kierarad:
This should help with #5562 so that the `trigger_update` and `status` endpoints are available. Remember that `trigger_update` is a `POST` request when using it.

Note: We removed the `last_parsed_result` endpoint when consolidating because it is unused and redundant now that we have endpoints for getting the config repo config with the parse results.


marques-work:
@akshaydewan would you have time to review this?

marques-work:
@akshaydewan just so we’re on the same page, this PR only consolidates the endpoint under one controller; it doesn’t actually fix the UI component to use the right endpoint.

akshaydewan:
This looks fine, and I'm going to merge it. I'm not sure what we should call the `operations` endpoint (`/config_repo_ops` seems a bit weird), but that can be changed later.

Thanks for making this change. It was my mistake that I missed the `POST` method in my request, but like you said, it brings more clarity to the API.

PR #5557, opened 2018-12-12
Set HSTS Header based on toggle.

akshaydewan:
Moved `chain.doFilter` below operations in `DefaultHeaderFilters`, because response is always committed otherwise.
When using HTTP, the header will be present, but ignored by the browser.

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/server/web/DefaultHeadersFilterTest.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/server/web/DefaultHeadersFilterTest.java
@@ -17,7 +17,8 @@
 package com.thoughtworks.go.server.web;
 
 import com.thoughtworks.go.server.service.support.toggle.FeatureToggleService;
-import com.thoughtworks.go.server.service.support.toggle.Toggles;
+import com.thoughtworks.go.util.SystemEnvironment;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
@@ -34,6 +35,7 @@ import static org.mockito.MockitoAnnotations.initMocks;
 
 public class DefaultHeadersFilterTest {
 
+    private static final String ENABLE_HSTS_HEADER = "gocd.enable.hsts.header";
     @Mock
     private HttpServletResponse response;
     @Mock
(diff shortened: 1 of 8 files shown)


GaneshSPatil:
@akshaydewan -- the changes LGTM, just a thought. Instead of putting it behind a feature toggle, can we use `SystemProperty`?

akshaydewan:
Yes, I think we can use a `SystemProperty` instead. There are many other SSL related configurations in the `SystemEnvironment` class so it makes sense to put it there.

akshaydewan:
Moved the toggle to SystemEnvironment. Also removed default headers from rails, since we would otherwise end up with double headers in the responses.

server/src/test-fast/java/com/thoughtworks/go/server/web/DefaultHeadersFilterTest.java
```diff
@@ -16,44 +16,73 @@
...
     public void shouldAddDefaultHeaders() throws Exception {
+        systemEnvironment.setProperty("gocd.enable.hsts.header", "true");
+        filter.doFilter(request, response, chain);
+
+        verify(response).isCommitted();
+        verify(response).setHeader("X-XSS-Protection", "1; mode=block");
+        verify(response).setHeader("X-Content-Type-Options", "nosniff");
+        verify(response).setHeader("X-Frame-Options", "SAMEORIGIN");
+        verify(response).setHeader("X-UA-Compatible", "chrome=1");
+        verify(response).setHeader("Strict-Transport-Security", "31536000");
```
bdpiparva: Is it `Strict-Transport-Security: 31536000` or `Strict-Transport-Security: max-age=31536000 `?

ketan: From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#Syntax

> Syntax:
> 
> ```
> Strict-Transport-Security: max-age=<expire-time>
> Strict-Transport-Security: max-age=<expire-time>; includeSubDomains
> Strict-Transport-Security: max-age=<expire-time>; preload
> ```

ketan: Also — should we allow configuring the params?

akshaydewan: We're going with four options: `-Dgocd.enable.hsts.header=true`, `-Dgocd.hsts.header.maxage=[value]` (defaults to 1 year), `-Dgocd.hsts.header.includesubdomains=true` and `-Dgocd.hsts.header.preload=true`. The user needs to use just the first one to enable the header, and the rest can be used for further customization

PR #5571, opened 2018-12-13
Moved authorisation config spa to component based spa

kritika-singh3:
To component based spa on, 
```
curl http://localhost:8153/go/api/admin/feature_toggles/use_new_auth_config_spa -d toggle_value=on -H 'Confirm:true' -u username:password
```

### Will merge after 18.12.0 release

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/models/auth_configs/spec/auth_configs_crud_spec.ts
+++ b/server/webapp/WEB-INF/rails/webpack/models/auth_configs/spec/auth_configs_crud_spec.ts
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2018 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/rails/webpack/views/pages/spec/auth_configs_spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+  afterEach(window.destroyDomElementForTest);
+
+  it("should render page with page header", (done) => {
+    jasmine.Ajax.stubRequest("/go/api/admin/plugin_info?type=authorization")
+           .andReturn(pluginInfoResponse(PluginInfoTestData.authorization()));
+
+    jasmine.Ajax.stubRequest("/go/api/admin/security/auth_configs").andReturn(authConfigResponse());
+
+    const originalFunction = authConfigsPage.fetchData;
+    done();
```
akshaydewan: @bdpiparva I think the `done()` should be after the expectations, so that the test waits for the promise

PR #5572, opened 2018-12-13
Introduce a `ButtonGroup` component to add spacing between buttons

vrushaliwaykole:
- The old `ButtonGroup` is renamed to `IconGroup` because it applies
  to `Icon` widgets


GaneshSPatil:
@naveenbhaskar had a concern with naming `ButtonGroup` as `IconGroup`. Not sure if that's still the case.

\cc: @naveenbhaskar @ketan

Issue #3234, opened 2017-03-07
Improve audit trail - Save/show user who cancelled a stage


arvindsv:
##### Issue Type

- Feature enhancement

##### Summary

GoCD does not log save some audit information, such as the user who cancelled a stage. Every stage should have more detailed and trackable audit information.

maheshp:
Done as part of https://github.com/gocd/gocd/pull/5578

PR #5578, opened 2018-12-14
Store and show username of user who cancelled a stage.

varshavaradarajan:
There are 3 places where this information should be shown
- [x] View - stage details page
- [x] Stage feed entry - access by /go/api/pipelines/pipeline_name/stages.xml
- [ ] API - https://api.gocd.org/current/#get-stage-instance and https://api.gocd.org/current/#get-stage-history - since these are old APIs, I haven't modified them as its a breaking change. Should I do that? Or should stage instance API and stage history API be converted to spark?

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/domain/Stage.java
+++ b/domain/src/main/java/com/thoughtworks/go/domain/Stage.java
@@ -34,6 +34,7 @@ public class Stage extends PersistentObject {
     private String name;
     private JobInstances jobInstances = new JobInstances();
     private String approvedBy;
+    private String cancelledBy;
     private int orderId;
     private Timestamp createdTime;
     private Timestamp lastTransitionedTime;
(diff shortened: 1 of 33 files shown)


domain/src/main/java/com/thoughtworks/go/domain/Stage.java
```diff
@@ -35,6 +35,7 @@
     private String name;
     private JobInstances jobInstances = new JobInstances();
     private String approvedBy;
+    private String cancelledBy;
```
GaneshSPatil: Just a question - `String`? Can this be `Username`?
We might require to compare usernames (to know who canceled). 
Instead of operating on primitives, it is better to wrap it in an object.

GaneshSPatil:
@varshavaradarajan -- a couple of thoughts:

* The canceled by information is required for auditing purposes. Currently on the dashboard, even if the pipeline instance (stage) is canceled, it shows `Triggered by changes` but no information about who canceled. Can we also show canceled by information? 

* GoCD cancels the job if the job hasn't printed the output for `X` amount of time. A stage will eventually be considered as canceled as one of the jobs was canceled by GoCD. Can we add a GoCD user there? As GoCD has canceled the job, which eventually caused the stage to be in cancelled state.
Check: [ConsoleActivityMonitor](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/ConsoleActivityMonitor.java#L84-L88)

* Also, Good to have, not must have. Tweak the email message to include stage canceled by username. Usages of: [StageEvent](https://github.com/gocd/gocd/blob/master/domain/src/main/java/com/thoughtworks/go/domain/StageEvent.java#L40)

* List of affected APIs:
  - Get Stage Instance
  - Get Stage History
  - Dashboard (If we choose to show this information on dashbord)
  - Notification Extension (If we choose to send this information to plugins)

GaneshSPatil:
Issue link: https://github.com/gocd/gocd/issues/3234

varshavaradarajan:
@GaneshSPatil - thank you! I forgot about the Dashboard API. Makes sense to show it there. Since API v2 hasn't been announced, no need to introduce another version? Unless you guys don't merge this until after announcing v2. I haven't worked on any of the APIs yet. Will do so this week. 

About the notification extension, does this need another version?

I think the message `Go cancelled this job as it has not generated any console output for more than %s minute(s)` is just fine. I don't see the need to introduce a GoCD user for this actually. Because:
1. If stage result is cancelled and cancelled by is null, then Go cancelled the stage.
2. The "GoCD _user_" if introduced would mean that there is a GoCD user on the server that people should be able to access (use this to login)/assign roles etc, right? 
3. There are other places where go does things on its own (like automatic triggering of pipelines - its go that's triggering the pipeline, but we show this as triggered by changes, instead of triggered by GoCD user because of changes. I would think that it would be inconsistent to introduce the GoCD user just for stage cancel then. 
 
But its up to you all to decide. 

Anyway, I don't think I have handled this case in the view or stage feed entry. Should I just check for stage result being cancelled and cancelled by being null to show that it was cancelled by GoCD for now?  We talked about wrapping the cancelledBy string in Username after querying the string from the stages table. I can do that in the `Stage.java` I think.

GaneshSPatil:
> I don't see the need to introduce a GoCD user for this actually. 

I was thinking of introducing GoCD user and use it at all the operations, such as pipeline trigger, pause, cancel, etc. But it brings in another set of questions, what if the GoCD username conflicts with real user's name. Also brings in the complexity of managing GoCD user internally (only).

Let's go ahead with adding `null` if GoCD has canceled the stage.

> Since API v2 hasn't been announced, no need to introduce another version?

We are announcing Dashboard API v2 as part of the release `v18.12.0` (happening this week). I don't see a need of holding onto it. Lets announce v2 for now, and along with these changes, we will create v3.

> Should I just check for stage result being cancelled and cancelled by being null to show that it was cancelled by GoCD for now?

I would say, lets explicitly check for `null` and add a comment explaning a null user check is added to check if the stage is being cancelled by GoCD due to job-timeout.

> wrapping the cancelledBy string in Username after querying the string from the stages table

👍

varshavaradarajan:
So, should stage instance and history API be moved to spark or not?

varshavaradarajan:
@GaneshSPatil - adding more routes to StageOperationsV1 for stage instance and history. That's alright? I don't want to create a separate module for it. Once the stage instance and history are moved, it will be versioned (v1). That would leave only `cancel` in rails. I won't move that as part of this PR.

GaneshSPatil:
@varshavaradarajan -- can we do the APIs as part of separate PR?

Reasons:
  - Changeset on the current PR is getting bigger, which causes it harder to review.
  - We might need to change the API responses a bit. 
             Example: the pagination information from the stage history. Its meta information and should be sent as part of headers or even if we keep it in response_body, it can be moved under `_metadata` field.
   It will be more manageble, if we've API specific conversations on a separate PR.

GaneshSPatil:
@rahulpargaonkar @adityasood -- We should modify the existing Functional test to verify the cancelled by user's name is shown on the stage_details_page.

\cc: @rajiesh

PR #5579, opened 2018-12-15
WIP - Config Repo Preflight

kierarad:
This adds an endpoint for checking if able to merge config repo definition


tomzo:
Hi @kierarad @marques-work ,

I've read the PR, I have only one suggestion. Don't you think API should accept `multipart/form-data` so that multiple files with their **names** could be proposed?
There are 2 reasons for this:
 * (this is minor) working in JSON plugin, the file extension has a meaning, pipeline or environment
 * in yaml, if multiple files are changed within the same repository, and then we have validation errors, then how location of the error will be shown? It might be impossible to guess in which file the error is really, because this location context was lost in the request to begin with.

Also it would be nice to add a test which sort of showcases the error location in the returned message.

arvindsv:
I agree with all of that. I think we should definitely send the whole directory, since there can be dependencies between files.

marques-work:
@tomzo @arvindsv Multipart upload should work now; just need to update the plugin PRs with the contract change.

Here's what the request looks like:

```
curl -u user:pass -H 'Accept: application/vnd.go.cd.v1+json' -F 'files[]=@path/to/file1.gocd.yaml' -F 'files[]=@path/to/file2.gocd.yaml' "http://localhost:8153/go/api/admin/config_repo_ops/preflight?pluginId=yaml.config.plugin"

# existing repo also passes repoId=xxx
curl -u user:pass -H 'Accept: application/vnd.go.cd.v1+json' -F 'files[]=@path/to/file1.gocd.yaml' -F 'files[]=@path/to/file2.gocd.yaml' "http://localhost:8153/go/api/admin/config_repo_ops/preflight?pluginId=yaml.config.plugin&repoId=myExistingRepo"
```

tomzo:
I think this is OK now, but a few commits could be squashed and rebased before merge.

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
```diff
@@ -34,6 +35,7 @@
 
 public abstract class ApiController implements ControllerMethods, SparkController {
     private static final Set<String> UPDATE_HTTP_METHODS = new HashSet<>(Arrays.asList("PUT", "POST", "PATCH"));
+    private static final String MULTIPART_REQ_ATTR = "org.eclipse.jetty.multipartConfig";
```
arvindsv: Can we push this to somewhere closer to Jetty? My worry is, if a Jetty upgrade changes this property, we won't know to change it here.

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
```diff
@@ -76,6 +78,10 @@ protected void verifyContentType(Request request, Response response) throws IOEx
         }
     }
 
+    protected void setMultpipartUpload(Request req, Response res) {
```
arvindsv: Typo in name.

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
```diff
@@ -76,6 +78,10 @@ protected void verifyContentType(Request request, Response response) throws IOEx
...
+        req.raw().setAttribute(MULTIPART_REQ_ATTR, new MultipartConfigElement(System.getProperty("java.io.tmpdir", "tmp")));
```
arvindsv: Does this mean that, once the request is served, this directory will not be cleaned up?

config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
```diff
@@ -47,35 +47,52 @@
 import static org.apache.commons.io.IOUtils.toInputStream;
 
 public class MagicalGoConfigXmlLoader {
-    private static final Logger LOGGER = LoggerFactory.getLogger(MagicalGoConfigXmlLoader.class);
```
arvindsv: Most (all?) of the changes in this file are moving stuff around, right?

server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java
```diff
@@ -0,0 +1,57 @@
...
+    public PartialConfigUpdateCommand(final PartialConfig partial, final String fingerprint, PartialConfigResolver resolver) {
+        this.partial = partial;
+        this.fingerprint = fingerprint;
+        this.resolver = resolver;
+    }
+
+    @Override
+    public CruiseConfig update(CruiseConfig cruiseConfig) {
+        if (partial != null && fingerprint != null) {
+            cruiseConfig.getPartials().remove(resolver.findPartialByFingerprint(cruiseConfig, fingerprint));
```
arvindsv: `findPartialByFingerprint` can be null. Is that going to be a problem? In some race condition maybe?

server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -1043,10 +1042,68 @@ public ArtifactStores artifactStores() {
         return cruiseConfig().getArtifactStores();
     }
 
+    public CruiseConfig validateCruiseConfig(CruiseConfig cruiseConfig) throws Exception {
```
arvindsv: Except for this method, just moves in this file?

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
marques-work: @arvindsv I had the same question in my head, but I _empirically_ found (at least in my basic trials) that nothing was being written to `java.io.tmpdir` after a multipart request.

FWIW, there's a method on `MultiPartInputStreamParser` that can explicitly delete the parts from disk, but we don't have a handle on that object. There's also a flag for "delete on exit" but same lack of handle problem. So...shrug?

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
marques-work: :) oops

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
marques-work: I will move it into `GoServer`, as I don't want to add any more module dependencies (mostly because I'm unsure of consequences). This might be close enough to the `jetty9` module?

server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java
marques-work: Not sure - this was a refactor (move method, rename) so this was likely a problem before. I can add a null check just in case anyway.

server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
marques-work: Exactly correct :)

config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
marques-work: Most. Just these minor changes largely due to changing the constructor signature of certain exceptions:

```diff
diff --git a/config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java b/config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
index cbdb1dc2b..d28cec426 100644
--- a/config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
+++ b/config/config-server/src/main/java/com/thoughtworks/go/config/MagicalGoConfigXmlLoader.java
@@ -111,11 +111,11 @@ public class MagicalGoConfigXmlLoader {
     }
 
     public CruiseConfig preprocessAndValidate(CruiseConfig config) throws Exception {
-        LOGGER.debug("[Config Save] In preprocessAndValidate: Cloning.");
+        LOGGER.debug("[Config Validation] In preprocessAndValidate: Cloning.");
         CruiseConfig cloned = CLONER.deepClone(config);
-        LOGGER.debug("[Config Save] In preprocessAndValidate: Validating.");
+        LOGGER.debug("[Config Validation] In preprocessAndValidate: Validating.");
         validateCruiseConfig(cloned);
-        LOGGER.debug("[Config Save] In preprocessAndValidate: Done.");
+        LOGGER.debug("[Config Validation] In preprocessAndValidate: Done.");
         config.encryptSecureProperties(cloned);
         return cloned;
     }
@@ -133,14 +133,14 @@ public class MagicalGoConfigXmlLoader {
         }
     }
 
-    private CruiseConfig validateCruiseConfig(CruiseConfig config) throws Exception {
+    public CruiseConfig validateCruiseConfig(CruiseConfig config) throws Exception {
         LOGGER.debug("[Config Save] In validateCruiseConfig: Starting.");
         List<ConfigErrors> allErrors = validate(config);
         if (!allErrors.isEmpty()) {
             if (config.isLocal())
-                throw new GoConfigInvalidException(config, allErrors.get(0).asString());
+                throw new GoConfigInvalidException(config, allErrors);
             else
-                throw new GoConfigInvalidMergeException("Merged validation failed", config, config.getMergedPartials(), allErrors);
+                throw new GoConfigInvalidMergeException(config, config.getMergedPartials(), allErrors);
         }
 
         LOGGER.debug("[Config Save] In validateCruiseConfig: Running validate.");

```

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
marques-work: Better yet, I found the constant in `org.eclipse.jetty.server.Request.__MULTIPART_CONFIG_ELEMENT` and extracted this into the `server` module as `RequestUtils.configureMultipart()`.

api/api-base/src/main/java/com/thoughtworks/go/api/ApiController.java
arvindsv: Maybe it uses it only if the payload is bigger than its buffer?

marques-work: It’s certainly possible— I’ll try with a larger payload (100mb?). If I see a file not being cleaned, I’ll have an explicit cleanup at the end.

PR #5581, opened 2018-12-17
Fix UI issues (#5464)

GaneshSPatil:
* Fix elastic profiles modal width.

#### References:
* https://github.com/gocd/gocd/issues/5464#issuecomment-447135559

#### Screenshot
screen shot 2018-12-17 at 10 56 00 am
screen shot 2018-12-17 at 10 56 08 am

\cc: @akshaydewan @naveenbhaskar @varshavaradarajan

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/pages/elastic_profiles/modals.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/pages/elastic_profiles/modals.tsx
@@ -52,7 +52,7 @@ abstract class BaseElasticProfileModal extends Modal {
   protected constructor(elasticProfile: ElasticProfile,
                         pluginInfos: Array<PluginInfo<Extension>>,
                         type: ModalType) {
-    super(Size.medium);
+    super(Size.large);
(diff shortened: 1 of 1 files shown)


bdpiparva:
@GaneshSPatil  - Also change the width of config_repo modal

GaneshSPatil:
@bdpiparva -- The internal styles of config repo modals will break if we change the modal width.
Here is a screenshot what I mean:

screen shot 2018-12-17 at 12 28 20 pm
screen shot 2018-12-17 at 12 28 12 pm

bdpiparva:
@GaneshSPatil - I see, but we should be able to adjust it for a large modal

akshaydewan:
I had worked with @naveenbhaskar on the config repo modals and we decided to go with the medium sized modal. The view isn't rendered by a plugin, and we had checked all material types to make sure it looks alright.

I think we should keep the Config Repos as it is, unless there's a good reason to change it.

rahulpargaonkar:
verified the issue on build.gocd. seems fixed

PR #5583, opened 2018-12-17
Move elastic profiles /usages api to internal path

vishaldevgire:
Changes:

- Created a new controller for `api/internal/elastic/profiles/profile_id/usages` API (this is required since the base path has changed)
- Moved `ElasticProfileUsageRepresenter` class to new package
- Changed ElasticProfileSPA to use internal path for `/usages` API


GaneshSPatil:
This is not necessary for the current release. Marking it for  `19.1.0`

PR #5598, opened 2018-12-19
Do not run pipeline config jasmine specs

GaneshSPatil:
* Pipeline config SPA has been abandoned.
  Pipeline Config SPA needs to be rewritten with new coding conventions

### DIFF
--- a/server/webapp/WEB-INF/rails/spec/webpack/specRoot.js
+++ b/server/webapp/WEB-INF/rails/spec/webpack/specRoot.js
@@ -19,7 +19,7 @@
 require('babel-polyfill');
 
 function importAll(r) {
-  r.keys().forEach(r);
+  r.keys().filter((filename) => !filename.includes('pipeline_configs')).forEach(r);
 }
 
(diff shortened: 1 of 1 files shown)


GaneshSPatil:
I would like to delete the code as well. But while rewriting SPA we might use it for reference purpose.

bdpiparva:
@GaneshSPatil - instead of modifying `specRoot.js` please can you disable the test form the file by adding `x` on the top level describe

GaneshSPatil:
> @GaneshSPatil - instead of modifying `specRoot.js` please can you disable the test form the file by adding `x` on the top level describe

@bdpiparva -- I thought of doing it. There are many files (~25) and I felt its too much to modify.

bdpiparva:
@GaneshSPatil - Agree, need to change many files, but modifying `specRoot.js` to ignore  `pipeline_config` may also ignore unintended files as well.

GaneshSPatil:
Hmm, agree!
I will close this PR then. :)

PR #5606, opened 2018-12-20
Remove page specific vm layouts of componentized pages

GaneshSPatil:


spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/ElasticProfilesController.java
```diff
@@ -55,6 +55,7 @@ public ModelAndView index(Request request, Response response) {
         HashMap<Object, Object> object = new HashMap<Object, Object>() {{
             put("viewTitle", "Elastic Agent Profiles");
         }};
-        return new ModelAndView(object, "elastic_profiles/index.vm");
+
+        return new ModelAndView(object, null);
```
bdpiparva: This is needed in case of if a user wants to switch to the old SPA.

Issue #5427, opened 2018-11-20
TFS checkout fails when material points to a single file


jsmythsci:
##### Issue Type

- Bug Report

##### Summary

TFS checkout fails when material points to a single file

##### Environment

GoCD is running on a Linux server and the affected agents are running Windows Server 2012R2.

###### Basic environment details

* Go Version: `18.10.0 (7703-42d1cbe661161b5400289ead86c0447c84af8c0a)`
* JAVA Version: `1.8.0_191`
* OS: `Linux 2.6.32-754.6.3.el6.x86_64`
* Browser vendor and version (if relevant): `N/A`

###### Additional Environment Details

This issue has been encountered on a copy of our production system that we created to test upgrading to the latest version of GoCD. The affected pipelines all work as intended on the following system:

* Go Version: `15.2.0(2248-c1f90a6aa3192a)`
* JAVA Version: `1.8.0_181`
* OS: `Linux 2.6.32-754.6.3.el6.x86_64`

Agents in the working environment are also running Windows Server 2012R2 but would be using an older version of Java since they are running an older version of the agent.

##### Steps to Reproduce

1. Create a new pipeline.
2. Add a TFS material.
3. Set the Project Path to a single file (e.g. $/TeamProjectName/AssemblyInfo.cs)
4. Set the Destination Directory to a single file (e.g. tfs/AssemblyInfo.cs)
5. Trigger the pipeline

##### Expected Results
Pipeline should check the file out of TFS and launch its first task.

##### Actual Results
Pipeline fails with an error like the following:

```Failed while checking out into Working Folder: pipelines\MyPipeline\tfs\AssemblyInfo.cs, Project Path: $/TeamProjectName/AssemblyInfo.cs, Workspace: bb18a58c6abfa36db84643de09258c86feaa48bc711174334eaea61594b9623f, Username: BuildUser, Domain: mydomain, Root Cause: java.io.FileNotFoundException: E:\GoCD\Go Agent\pipelines\MyPipeline\tfs\AssemblyInfo.cs\$tf\5\377de3f5-a6f8-41ed-970a-9b0bbf2c1a50.gz (The system cannot find the path specified)```

##### Possible Fix
This appears to be an extension of #183 (TFS does not create destination directory if project path is a file). It seems that TFS is trying to create a local workspace and is failing because:

1. It creates a file object instead of a directory.
2. It tries to create the local $tf directory inside the file object.

[...]

PR #5608, opened 2018-12-21
Proposal: Agent side toggles controlled by the server

arvindsv:
[Not complete. This is just a proposal, before I spend more time on it. If you all say "No! It's a bad idea", I will not spend any more time on it]

Wanted to think about https://github.com/gocd/gocd/issues/5427. Not sure whether it is the right thing to do. However, there is no easy way to put in a toggle, which is controlled from the server side, which affects the agent side.

The benefit of this kind of a toggle is that we can experiment more freely on the agent side. Currently, if we have an agent-side toggle, turning it on and off is a pain, since it has to be done individually. With this toggle framework (or at least a framework to pass in properties to the agent, from the server side), it's easier to toggle it.

1. Server sets the header with a set of values: "GoCD-Agent-Extra-Properties: Key1=Value1 Key2=Value2" etc.

2. Agent sees it the next time it starts or checks for upgrade. Sets its own properties.

Wonder if it would have been easier to do things like the TFS upgrade, if we had this. Maybe.

### DIFF
--- a/agent-common/src/main/java/com/thoughtworks/go/agent/common/util/HeaderUtil.java
+++ b/agent-common/src/main/java/com/thoughtworks/go/agent/common/util/HeaderUtil.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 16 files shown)


arvindsv:
Closing. Replaced by https://github.com/gocd/gocd/pull/5666.

Issue #1353, opened 2015-07-16
Pagination is broken.


juhijariwala:
**Problem:**
Pagination in stage history panel shows 10 records per page.
In the below example, you can see **Page1** has records from **5-14** while **Page2** has records from **1-4**. 

**Page1**
<img width="268" alt="screen shot 2015-07-15 at 10 06 09 am" src="https://cloud.githubusercontent.com/assets/4953510/8698068/28329aea-2b18-11e5-8fe1-dd8bec39c39c.png">

**Page2**
<img width="243" alt="screen shot 2015-07-15 at 10 06 22 am" src="https://cloud.githubusercontent.com/assets/4953510/8698070/2b8a0232-2b18-11e5-9fae-55e2016728ee.png">

**Note:** (Page 1 holds the latest records)

Let's suppose more records get added to the history. If I went to page 2, I would expect to only see records 1-4, but instead I see records 1-9 which is incorrect. The content of page 2 should not change. If you compare the first and second set of images you can see that records from 5-9 are displayed on both of the pages, page 1 and page 2. For more explanation refer http://www.sitepoint.com/paginating-real-time-data-cursor-based-pagination/ .

**page 1**
<img width="247" alt="screen shot 2015-07-15 at 2 10 40 pm" src="https://cloud.githubusercontent.com/assets/4953510/8698228/c4368fcc-2b19-11e5-9b44-ad5e247cb32c.png">

**page 2**
<img width="247" alt="screen shot 2015-07-15 at 2 10 49 pm" src="https://cloud.githubusercontent.com/assets/4953510/8698234/c72c4aaa-2b19-11e5-8376-7f6cecff9c5d.png">
​
For this problem, I would like to propose a solution where there are just two buttons to navigate between pages (previous and next).
The previous button will retrieve a fixed number of records before the top record in the page. The next button will retrieve a fixed number of records after the last record in the page.

| Stage History |
| --- |
| Record 11 |
| Record 12 |
| Record 13 |
| Record 14 |
| Record 15 |
| Record 16 |
| Record 17 |
| Record 18 |
| Record 19 |
| Record 20 |

Previous will retrieve the records before Record10.
Next will retrieve the records after Record20.

**Result of prev**

| Stage History |
| --- |
| Record 1 |
| Record 2 |
| Record 3 |
| Record 4 |
| Record 5 |
| Record 6 |
| Record 7 |
| Record 8 |
| Record 9 |
| Record 10 |

**Result of next**

| Stage History |
[...]

PR #5620, opened 2018-12-25
Stage history api V1

varshavaradarajan:
@GaneshSPatil - Review only the last commit. Since I am basing this branch off of the main branch in #5578, other commits are shown. They will disappear after the other PRs are merged. Okay?

### DIFF
--- a/api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1.java
+++ b/api/api-stage-operations-v1/src/main/java/com/thoughtworks/go/apiv1/stageoperations/StageOperationsControllerV1.java
@@ -22,15 +22,18 @@ import com.thoughtworks.go.api.representers.JsonReader;
 import com.thoughtworks.go.api.spring.ApiAuthenticationHelper;
 import com.thoughtworks.go.api.util.GsonTransformer;
 import com.thoughtworks.go.api.util.HaltApiResponses;
+import com.thoughtworks.go.apiv1.stageoperations.representers.StageInstancesRepresenter;
 import com.thoughtworks.go.apiv1.stageoperations.representers.StageRepresenter;
 import com.thoughtworks.go.config.exceptions.RecordNotFoundException;
 import com.thoughtworks.go.domain.JobInstance;
 import com.thoughtworks.go.domain.NullStage;
 import com.thoughtworks.go.domain.Stage;
+import com.thoughtworks.go.presentation.pipelinehistory.StageInstanceModels;
 import com.thoughtworks.go.server.service.PipelineService;
 import com.thoughtworks.go.server.service.ScheduleService;
 import com.thoughtworks.go.server.service.StageService;
 import com.thoughtworks.go.server.service.result.HttpOperationResult;
+import com.thoughtworks.go.server.util.Pagination;
 import com.thoughtworks.go.serverhealth.HealthStateScope;
 import com.thoughtworks.go.serverhealth.HealthStateType;
 import com.thoughtworks.go.spark.Routes;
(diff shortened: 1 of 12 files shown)


api/api-dashboard-v3/src/main/java/com/thoughtworks/go/apiv3/dashboard/representers/StageRepresenter.java
```diff
@@ -0,0 +1,46 @@
...
+                linkWriter.addLink("self", Routes.Stage.self(pipelineName, pipelineCounter, model.getName(), model.getCounter()));
+            })
+            .add("name", model.getName())
+            .add("counter", model.getCounter())
+            .add("status", model.getState().name())
+            .add("approved_by", model.getApprovedBy())
+            .add("scheduled_at", model.getScheduledDate());
+
+        if (model.getState().stageResult() == StageResult.Cancelled) {
+            jsonOutputWriter.add("cancelled_by", model.getCancelledBy() == null ? "GoCD" : model.getCancelledBy());
```
GaneshSPatil: @varshavaradarajan -- I think, this is the only change from Dashboard API `v2` and `v3`.

Just a nitpick (info), if a system has a user with `GoCD`, it will conflict with our approach. Though, I am OK with this conflict.

\cc: @maheshp @varshavaradarajan @arvindsv @ketan

arvindsv:
@GaneshSPatil 

> This PR contains other changes as well apart from Stage history api V1.

The description says only the last commit should be reviewed.

api/api-dashboard-v3/src/main/java/com/thoughtworks/go/apiv3/dashboard/representers/StageRepresenter.java
arvindsv: I think that's ok. We also treat "changes" as special, in a way ("Triggered by changes").

varshavaradarajan:
What about:
```json
"_meta": {
  "total": 3,
  "offset": 1,
  "page_size": 10,
  "next_page": "http://localhost:8153/go/api/stages/up42/up42_stage/history/2"
}
```
`next_page` will be `null` if its the last page.

ketan:
[This blog from slack](https://slack.engineering/evolving-api-pagination-at-slack-1c1f644f8e12) describes why the current pagination we use is completely broken, in addition to high performance cost of retrieval of data, it's also inaccurate when you're paginating through results while writing items to the dataset. The blog mentions some tradeoffs of cursor based pagination, but that's possibly something we can address in other ways.

See #1353 and #1100.

ketan:
This is a good example of the complexities involved with the current pagination. A method to get a page of stage history from cache: https://github.com/gocd/gocd/blob/49ddcb8a0a7c3c8fe85025b8ca2b443a5dfdbd62/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L442-L462

Then because there's a cache involved, something to purge the cache:  https://github.com/gocd/gocd/blob/49ddcb8a0a7c3c8fe85025b8ca2b443a5dfdbd62/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L117-L130

Which is invoked when a single stage is updated: https://github.com/gocd/gocd/blob/49ddcb8a0a7c3c8fe85025b8ca2b443a5dfdbd62/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L284-L301 or created: https://github.com/gocd/gocd/blob/49ddcb8a0a7c3c8fe85025b8ca2b443a5dfdbd62/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L91-L115

varshavaradarajan:
@ketan, @GaneshSPatil  - I have an initial patch for cursor pagination for stage history (api only).  However, I don't think it should be a part of this PR. Because, changes to pagination involves changes to more than just the stage history API. If you want to hold off announcing stage history v1 because of the proposed changes to pagination, that's fine. 

The patch still needs to take care of the following things:
- db perf
- currently uses Pagination.offset as the next cursor. But that needs to be updated.
- The old rails API will be broken because of this change. Needs fixing
- refactoring Stage.xml 

I didn't fully get what's been proposed as part of https://github.com/gocd/gocd/issues/1353. But the stage details page uses another query to get the information, which also needs to change to remove offset and use cursor pagination instead. 

Here's what the API response will contain. 
```
"meta" : {
    "page_size" : 10,
    "next_cursor" : 19,
    "_links" : {
      "next_page" : {
        "href" : "http://localhost:8153/go/api/stages/p3/s3/history/19"
      }
    }
  }
```
Commit: https://github.com/varshavaradarajan/gocd/commit/df3063c676084efb9cc31fed6967eee6b717c0b1

varshavaradarajan:
@ketan - updates? I haven't included the pagination changes as part of this PR as mentioned above. I think this can be merged and the discussion around pagination continued.

Issue #5605, opened 2018-12-20
Move artifact store SPA to typescript


bdpiparva:
- [x] Move existing `ArtifactStore` spa to typescript(#5623)
- [x] Add a toggle to switch to old page(`use_old_artifact_stores_spa `)
   ```bash
   curl -H 'Confirm:true' \
          -u 'username:password' \ 
          http://example.ci.com/go/api/admin/feature_toggles/use_old_artifact_stores_spa \
          -d toggle_value=on
  ```
- [x] Update functional tests for the same(https://github.com/gocd/ruby-functional-tests/pull/256)

PR #5623, opened 2018-12-26
Moving artifact store spa to typescript - #5605

bdpiparva:


.idea/codeStyles/Project.xml
```diff
@@ -15,11 +15,11 @@
     <HTMLCodeStyleSettings>
       <option name="HTML_ENFORCE_QUOTES" value="true" />
     </HTMLCodeStyleSettings>
-    <JSCodeStyleSettings version="0">
+    <JSCodeStyleSettings>
```
GaneshSPatil: These settings keep changing based on the IntelliJ version that we use.
Not sure whether we should commit these changes or ignore.

\cc: @ketan

PR #5628, opened 2018-12-27
Users api v3

GaneshSPatil:
Introduce Users API v3 to include `is_admin` and `roles` information


api/api-users-v3/src/main/java/com/thoughtworks/go/apiv3/users/UsersControllerV3.java
```diff
@@ -0,0 +1,198 @@
...
+        if (!deletionFailureResult.isEmpty()) {
+            res.status(result.httpCode());
+            return writerForTopLevelObject(req, res, outputWriter -> BulkDeletionFailureResultRepresenter.toJSON(outputWriter, deletionFailureResult, result));
+        }
+
+        return renderHTTPOperationResult(result, req, res);
+    }
+
+    private String saveUserAndRenderResult(Request req, Response res, HttpLocalizedOperationResult result, User user, String username) throws IOException {
+        userService.save(user, TriState.from(user.isEnabled()), TriState.from(user.isEmailMe()), user.getEmail(), user.getMatcher(), result);
```
ketan: This is problematic. In the earlier API, if the boolean flag was not set, it implied not changing that property. The new implementation, it appears may possibly cause a `PATCH` api with no `enabled/email_me` option set to actually unset the value on the user record.

If the current implementation does handle that, then perhaps a test is required.

PR #5639, opened 2018-12-28
Users api v3 remaining endpoints

GaneshSPatil:


ketan:
Commit message inspired from https://whatthecommit.com/ ?

GaneshSPatil:
> Commit message inspired from https://whatthecommit.com/ ?

@ketan -- Sorry about that. I will squash and amend (write a better) commit message :).

Issue #5554, opened 2018-12-12
Remove support for DES


maheshp:
##### Summary

Support for AES encryption/decryption was introduced in release 18.7.0, with that, support for DES was deprecated. Support for DES was retained to support DES cipher-texts configured in config-repos. 

Support for DES needs to removed for 19.1.0

akshaydewan:
Since we still need to support the upgrade path from pre 18.7 to 19.1+, we'll keep the migration code and `DESEncrypter` in place. 
We can remove the System property `gocd.des.cipher.enabled` and remove support for encrypting and decrypting using DES, except in the upgrade scenario.

PR #5646, opened 2019-01-02
Remove support for DES

akshaydewan:
Issue: #5554
- remove the system property that enables DES
- remove decryption support for DES.

The DES to AES migration code is still in place.


ketan:
LGTM: I've run the integration tests stage despite failures, there are likely a lot of hardcoded values that may need fixing.

ketan:
There are a bunch of `cipher` files checked into the repo. Perhaps remove that too? For tests that do require a specific (DES) cipher to be set, there are helper methods to allow setting the cipher. See usages of `ResetCipher`.

akshaydewan:
Merging this change. Will try to remove the cipher files and send a separate PR.

PR #5648, opened 2019-01-03
Remove cipher files

akshaydewan:
AES cipher will be generated at runtime. They aren't being used in tests or anywhere else.

Related to #5554


ketan:
@akshaydewan — mind taking a look at the test failures when you have time (low priority)

Issue #5651, opened 2019-01-03
Show config repo material clone errors in server health messages


GaneshSPatil:
##### Issue Type
- Bug Report

##### Summary
When an invalid (non-existing) material URL is provided for config repo materials, even after the MDU finishes with an error, no error is shown in the server health messages.

###### Basic environment details

* Go Version: `GoCD Version: 19.1.0 (8275-d3e8eed2a25e5171c152f8d14c0ac679c0827912).`
* JAVA Version: `jdk11`

##### Steps to Reproduce

1. Create a config repo pipeline with invalid URL.
2. Wait for few minutes for MDU to happen.
3. An MDU will happen and a message will be populated in the server health messages for a few seconds.
4. The message is cleared after few seconds (~[0-5] secs) from server health messages.

##### Expected Results
A material clone error should be shown under "server health messages".

##### Actual Results
No error is shown. Or error is shown just for few seconds (~[0-5] secs)

##### Possible Fix
`ServerHealthService.purgeStaleHealthMessages` is triggered via spring timer every 5 secs.
Which goes over all messages and check if it needs to be removed, based on the scope.

Initially, MDU happens for the invalid config repo material and a server health message is populated accordingly for the same.
Later, on the spring timer, `ServerHealthService.purgeStaleHealthMessages` will be called, which will try to remove the error based on if the material isRemovedFromConfig.

As for checking if the material is removed from the config, we only fetch materials belonging to pipelines and don't go over config repo materials. Which will result into config repo material not belonging in the config. And the server health message for the material is removed.

We should fetch all materials (pipeline materials + configrepomaterials) to check if the material belongs to the config.

arvindsv:
@GaneshSPatil Does that mean that in 18.12, config repo errors are not shown at all?

GaneshSPatil:
@arvindsv -- No. Only config repo material clone errors are not shown.
There are multiple types of `ServerHealthScope`, `MATERIAL`, `MATERIAL_UPDATE`, `CONFIG_PARTIAL`, etc.

If MDU fails for any material, a message is added to either `MATERIAL` or `MATERIAL_UPDATE` scope.
If a config repo parsing fails, a message is added to `CONFIG_PARTIAL` scope.

Until now, if an MDU fails for the config repo materials, those messages were lost. 
All the messages related to config repo are shown in server health messages, as they belongs to `CONFIG_PARTIAL` scope.

GaneshSPatil:
[...]

PR #5652, opened 2019-01-03
Do not remove MaterialHealthStateScope for config repo materials (#5651)

GaneshSPatil:


common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java
```diff
@@ -232,7 +232,7 @@ public boolean isRemovedFromConfig(CruiseConfig cruiseConfig, String group) {
         },
         MATERIAL {
```
GaneshSPatil: This is only done for the material scope. It also needs to be done for `material_update` scope.
As a config repo material can also fail while updating. :)

PR #5661, opened 2019-01-04
Bulk update system admins

akshaydewan:
Bumped up Admins Config API to v2
Added `PATCH` endpoint to update System Admins
There's no validation performed right now -- any string can can be passed, even if it doesn't exist as a user. I can implement this subsequently

Example:
```bash
curl -X PATCH \
  http://localhost:8153/go/api/admin/security/system_admins \
  -d '{
     "operations": {
     	"users": {
     		"add": ["meeow", "foo"],
     		"remove": ["admin"]
     	},
     	"roles": {
     		"add": ["xyz"]
     	}
     }
}'
```

```json
Response:
200 OK
{
    "_links": {
        "doc": {
            "href": "https://api.gocd.org/#system_admins"
        },
        "self": {
            "href": "http://localhost:8153/go/api/admin/security/system_admins"
        }
    },
    "roles": [
        "xyz"
    ],
    "users": [
        "foo",
        "meeow"
    ]
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,9 @@ agent/target/
 api/api-admins-config-v1/logs/
 api/api-admins-config-v1/out/
 api/api-admins-config-v1/target/
+api/api-admins-config-v2/logs/
+api/api-admins-config-v2/out/
+api/api-admins-config-v2/target/
 api/api-artifact-store-config-v1/config/
 api/api-artifact-store-config-v1/logs/
 api/api-artifact-store-config-v1/out/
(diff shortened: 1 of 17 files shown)


akshaydewan:
Updated request structure to allow add/remove for users and roles in the same request

GaneshSPatil:
@akshaydewan -- The PR LGTM.

As you've already mentioned in the issue description, we should add validation for update operations.
Scenarios:
- Remove user/role to verify the existing user/role is a super admin

Not sure if we can validate the admin while adding the role/user as the user might be coming from the plugin.

akshaydewan:
Merged this. I'll add the validations in another PR, so that @vrushaliwaykole and @ketan can start using it if needed.

PR #5663, opened 2019-01-04
WIP - MDU changes to support parallel processing of config repo materials

maheshp:
** Do Not Merge **


maheshp:
@ibnc thanks for your inputs. I am trying to list  few reasons for reverting #5550 

1.  I believe the behaviour and responsibilities of `MaterialUpdateListener` is different from `ConfigMaterialUpdater`(renamed to `ConfigMaterialUpdateListener`). While the `MaterialUpdateListener` is responsible to update a given material and add relevant modifications into the db(basically responsible for MDU), the `ConfigMaterialUpdater` is responsible to parse and merge the config repo post the MDU. 
2. I think the `MaterialUpdateListenerFactory` should be used to create only `MaterialUpdateListeners` based on the material types rather than `ConfigMaterialUpdaters` which should be handled separately through a different SystemEnvironment which I have done through - `config.material.post.update.threads`
3. Also, I thought the `MaterialUpdateService` being notified with `MaterialUpdateCompletedMessage` upon completion of parsing and merging the config_repo by `ConfigMaterialUpdater` seemed a bit more simpler rather than the `MaterialUpdateService` notifying the `ConfigMaterialUpdater` to parse and merge a config repo.

If required I can get on a call and we can discuss this further.

ibnc:
> @ibnc thanks for your inputs. I am trying to list few reasons for reverting #5550
> 
> 1. I believe the behaviour and responsibilities of `MaterialUpdateListener` is different from `ConfigMaterialUpdater`(renamed to `ConfigMaterialUpdateListener`). While the `MaterialUpdateListener` is responsible to update a given material and add relevant modifications into the db(basically responsible for MDU), the `ConfigMaterialUpdater` is responsible to parse and merge the config repo post the MDU.

I'm not sure I agree. My thinking is that `GoRepoConfigDataSource` is the one responsible for parsing and merging the configuration. `ConfigMaterialUpdater` or `ConfigMaterialUpdateListerner` is responsible for setup and checkout. This gets into the naming convention comments I left as well.

> 2. I think the `MaterialUpdateListenerFactory` should be used to create only `MaterialUpdateListeners` based on the material types rather than `ConfigMaterialUpdaters` which should be handled separately through a different SystemEnvironment which I have done through - `config.material.post.update.threads`

Agreed

> 3. Also, I thought the `MaterialUpdateService` being notified with `MaterialUpdateCompletedMessage` upon completion of parsing and merging the config_repo by `ConfigMaterialUpdater` seemed a bit more simpler rather than the `MaterialUpdateService` notifying the `ConfigMaterialUpdater` to parse and merge a config repo.

Makes sense.

> 
> If required I can get on a call and we can discuss this further.

I don't think we need to do that. I'd be curious to see what the performance boost of the PR would be. @tomzo Do you think you could test this out on the same instances you tested out #5550 on, and report back?

arvindsv:
@ibnc 

> I'd be curious to see what the performance boost of the PR would be

I'd expect to see almost no performance boost from this PR. And, in my opinion, that's ok.

The biggest benefit I see is that we can increase/decrease the number of threads for the _merging_ config-repo materials, independently of the number of MDU threads for non-config-repo materials. I believe this was part of your PR and from what I know, that behavior continues in this one. That's where we saw the most performance boost.

In this PR, the number of config material update listeners can be changed and are independent of the threads for non-config-repo materials. That's what I like about this one.

Overall, I think your PR set the standard for perf. This one makes it more configurable (adds that old property back) and changes the messaging (or listener framework) so that it doesn't come back to the material update service more than once for a single material.

maheshp:
> Edit: If I understand correctly, the number of listener threads for MDU was a combination of normal material and config-repo material, after the earlier PR. The number of merging threads was introduced and made configurable. In this PR, the number of config-repo listener threads is independent of the number of non-config-repo listener threads.

Yes that's right, listing the system properties used for configuring the threads around MDU and ConifgRepo parse/merge.

| System Property                  | Default Value | Description  |
|-------------------------------- |---------------|-------------| 
| material.check.threads               | 10       | No of threads used to update SCM Materials|
|dependency.material.check.threads| 3   | No of threads used to update Dependency Materials |
|material.config.check.threads |2| No of threads used to update ConfigRepo Materials |
|config.material.post.update.threads|2| No of threads used to parse & merge ConfigRepo post material update |

maheshp:
@tomzo your any inputs for this PR, when you get time.

base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -214,6 +214,8 @@
 
     public static GoIntSystemProperty DEPENDENCY_MATERIAL_UPDATE_LISTENERS = new GoIntSystemProperty("dependency.material.check.threads", 3);
 
+    public static GoIntSystemProperty CONFIG_MATERIAL_POST_UPDATE_LISTENERS = new GoIntSystemProperty("config.material.post.update.threads", 3);
```
tomzo: @maheshp in the summary, you written this default was supposed to be `2`. But here we have 3.

server/src/main/java/com/thoughtworks/go/server/materials/ConfigMaterialUpdater.java
```diff
@@ -27,34 +27,33 @@
...
+import org.springframework.stereotype.Service;
 
 import java.io.File;
 
 /**
  * Updates configuration from repositories.
  */
-public class ConfigMaterialUpdateListener implements GoMessageListener<MaterialUpdateCompletedMessage> {
-    private static final Logger LOGGER = LoggerFactory.getLogger(ConfigMaterialUpdateListener.class);
+public class ConfigMaterialUpdater implements GoMessageListener<MaterialUpdateCompletedMessage> {
```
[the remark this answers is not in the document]
tomzo: I agree on that part, would be better to have `ConfigMaterialUpdateListener`

base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
maheshp: Oops, will change it to 2.

tomzo:
>  I'd be curious to see what the performance boost of the PR would be. @tomzo Do you think you could test this out on the same instances you tested out #5550 on, and report back?

@ibnc I did build this PR and started server with all production repositories. But 19.1 removes DES support and we still have some repos with secrets encrypted with it. So I cannot really give you any number now, sorry.

ibnc:
@tomzo no worries :)

arvindsv:
Alright, you three. You _have_ to decide on the names for these. :) It looks like everyone is fine with the change, but not happy with the names. Here's what I think it looks like:

2019_01_11_mdu_config_repo

Who wants to provide suggestions for names? The candidates for change are:

- ConfigMaterialUpdater
- ConfigMaterialPostUpdateListenersFactory (this is the thing that current makes ConfigMaterialUpdater objects)
- ConfigMaterialPostUpdateQueue

arvindsv:
The suggestions I've heard are:

- ConfigMaterialUpdater => ConfigMaterialUpdateListener
- ConfigMaterialPostUpdateListenersFactory => ConfigMaterialUpdateListenersFactory
- ConfigMaterialPostUpdateQueue => "The queue name should also be consistent with the other two corresponding classes"

ibnc:
> The suggestions I've heard are:
> 
> * ConfigMaterialUpdater => ConfigMaterialUpdateListener
> * ConfigMaterialPostUpdateListenersFactory => ConfigMaterialUpdateListenerFactory
> * ConfigMaterialPostUpdateQueue => "The queue name should also be consistent with the other two corresponding classes"
> 
> ConfigMaterialUpdateQueue is already taken.

I'm in favor of going the route of `ConfigMateterialUpdateListener` and `ConfigMaterialUpdateListenerFactory`  

I'm actually okay with the queue name not changing

PR #5666, opened 2019-01-04
Agent side toggles controlled by the server

arvindsv:
There is no easy way to put in a toggle, which is controlled from the server side, which affects the agent side.

The benefit of this kind of a toggle is that we can experiment more freely on the agent side. Currently, if we have an agent-side toggle, turning it on and off is a pain, since it has to be done individually. With this toggle framework (or at least a framework to pass in properties to the agent, from the server side), it's easier to toggle it.

1. Server sets the header with a set of values: "GoCD-Agent-Extra-Properties: Key1=Value1 Key2=Value2" etc.

2. Agent sees it the next time it starts or checks for upgrade. Sets its own properties.

I've verified this by printing out the properties by adding a line [here](https://github.com/gocd/gocd/pull/5666/files#diff-c571d5ed8575f158cc12fdc2f23f4f80R102). For this verification, I used:

```
GO_SERVER_SYSTEM_PROPERTIES="'-Dgocd.agent.extra.properties=extra.property=value1%20with%20space extra%20property%20with%20space=value2%20with%20space'" ./server.sh
```

Couple of considerations with this PR:

1. I don't like the name `HeaderUtil`. But, couldn't think of anything. It helps reduce some duplication.

2. The header is added to almost all calls in AgentRegistrationController (like `/admin/agent-plugins.zip`). That's not really needed. I can put it in a more specific place than [here](https://github.com/gocd/gocd/pull/5666/files#diff-9c099507cd71dda2017a601c4e4f3fc7R185), so that it only affects the two calls `/admin/latest-agent.status` and `/admin/agent`. If you want.

Originally: https://github.com/gocd/gocd/pull/5608

### DIFF
--- a/agent/src/main/java/com/thoughtworks/go/agent/service/AgentUpgradeService.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/service/AgentUpgradeService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 ThoughtWorks, Inc.
+ * Copyright 2019 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,11 +30,15 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.io.IOException;
+import java.util.Map;
+
+import static com.thoughtworks.go.agent.common.util.HeaderUtil.parseExtraProperties;
+import static com.thoughtworks.go.util.SystemEnvironment.*;
 
 @Service
 public class AgentUpgradeService {
     private static final Logger LOGGER = LoggerFactory.getLogger(AgentUpgradeService.class);
-    static final Marker FATAL = MarkerFactory.getMarker("FATAL");
+    private static final Marker FATAL = MarkerFactory.getMarker("FATAL");
     private final GoAgentServerHttpClient httpClient;
     private final SystemEnvironment systemEnvironment;
     private URLService urlService;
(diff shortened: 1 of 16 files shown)


agent/src/main/java/com/thoughtworks/go/agent/service/AgentUpgradeService.java
```diff
@@ -91,6 +97,11 @@ void checkForUpgrade(String agentMd5, String launcherMd5, String agentPluginsMd5
         }
     }
 
+    private void updateExtraProperties(Header extraPropertiesHeader) {
+        final Map<String, String> extraProperties = parseExtraProperties(extraPropertiesHeader);
+        extraProperties.forEach(System::setProperty);
```
ketan: All of this looks good to me. Can merge this as is.

If any of the properties change server side, can/should we force a restart? Some JVM specific system properties cannot be changed at runtime (encoding, ssl debug, networking).

arvindsv: @ketan I considered that. I thought it would be too disruptive since it will restart all agents immediately if that happens. I'd prefer to not do it unless we find a need for it.

ketan:
>
> @ketan <https://github.com/ketan> I considered that. I thought it would
> be too disruptive since it will restart all agents immediately if that
> happens. I'd prefer to not do it unless we find a need for it.
>
I could argue the same about this PR :)

arvindsv:
@ketan :)

I need it for https://github.com/gocd/gocd/issues/5427 - that's the whole reason this exists.

arvindsv:
Verified it and it works ok, with the Base64, etc. @ketan said I can merge it after verification. So, doing that.

Issue #5650, opened 2019-01-03
Config Repositories SPA Improvements (Part 2)


GaneshSPatil:
epic: https://github.com/gocd/gocd/issues/5175

##### Issue Type

- Feature enhancement

#### New features (ordered by priority):

- [x] Show config repo parsed information
	- [x] last known commit
	- [x] last good commit
	- [x] metadata of the last good commit (author, commit_msg, sha)

- [x] Show errors related to each config repository
	- [x] config repo parsing errors
	- [x] Config entity validation errors

- [x] Check connection for config repo materials

- [x] Internally force update a config repo upon creation (either via UI or via API). (#5679)

- [x] Show MDU for config repo material in progress information on the SPA.
	- MDU can be scheduled externally by forcing an update from the SPA/API or timer based MDU within GoCD. 
	A user must know if the MDU is in-progress for the material so that the user can make a decision of scheduling another update or not.
	Also today, a user forces an update, there is no feedback around the progress of the update apart from just a message saying `An update was scheduled for this config repository.`

- [x] Show real-time data on the SPA. (#5655)
	Users shouldn't need to refresh the page. Maybe introduce ajax_polling

- [ ] Improve config repo validation error messaging.
    - [ ] Perform schema validation to check the existence of mandatory fields instead of NullPointerExceptions.
      Example: For a config repo pipeline git material, if URL field is not specified, the following error is shown in the server health messages:
       `runtime; - Unexpected error when handling plugin response java.lang.NullPointerException`

#### Bugs:
- [x] Show error if invalid repo url is provided. (#5651) (#5652)
	  Currently, the config repo material clone error is populated in the server health messages only for a fraction of time (5 secs) and then removed.

- [x] Improve page operation messages
	- [x] `An update was scheduled for this config repository.`
		Instead of saying this, change it use the config_repo_id

#### Tech Debt
- [ ] Get rid of hardcoded plugin-images and make use of image_ref from the plugin_infos. 
       Requires: # to be done first.

#### UX Improvements
- [ ] Smart Expand 
    - [ ] Do not expand anything if all config repos are green
    - [ ] If a user creates a new config repo, expand that config repo (only for the current page)
    - [ ] Expand a config repo, if it has errors

- [ ] Smart Highlights/Indicators
    - [ ] Indicator for config repo never parsed
    - [ ]  Indicator for config repo MDU in progress
    - [ ]  Indicator for latest known commit is bad
    - [ ] Highlight config repo if it has errors

[...]

PR #5676, opened 2019-01-08
Schedule material update for config repo materials upon creation or update (#5650)

GaneshSPatil:


ketan:
What happens if a config repo is defined in a config repo as in here https://github.com/tomzo/gocd-yaml-config-plugin#configrepo

server/src/main/java/com/thoughtworks/go/config/update/CreateConfigRepoCommand.java
```diff
@@ -20,20 +20,24 @@
...
+                                   HttpLocalizedOperationResult result, ConfigRepoExtension configRepoExtension,
+                                   MaterialUpdateService materialUpdateService, MaterialConfigConverter converter) {
+        super(securityService, configRepo, username, result, configRepoExtension, materialUpdateService, converter);
         this.configRepo = configRepo;
     }
 
     @Override
     public void update(CruiseConfig preprocessedConfig) {
         preprocessedConfig.getConfigRepos().add(configRepo);
+        scheduleMaterialUpdate(configRepo);
```
maheshp: @GaneshSPatil I am not sure if this is the right place to schedule a material update. What happens if the configRepo is invalid or if the entity is not saved, we will still end up scheduling a material update.

GaneshSPatil: @maheshp -- I see what you're saying.
Currently, the EntityConfigUpdateCommand does not expose a post update method to perform post update validations/operations.
[This](https://github.com/gocd/gocd/issues/2780) issue talks about it.

`getPreprocessedEntityConfig` is the only method called post update. I am not sure whether the EntityConfigUpdateCommand is the right place to perform such operations.

How about moving it outside of the command and let configRepoService deal with it?

GaneshSPatil:
> What happens if a config repo is defined in a config repo as in here https://github.com/tomzo/gocd-yaml-config-plugin#configrepo

We are not scheduling a material update immediately for config-repos defined in config repos. Such materials will be updated after 30 secs (material.update.interval).

server/src/main/java/com/thoughtworks/go/config/update/CreateConfigRepoCommand.java
maheshp: Wondering if this can be done through a EntityConfigChangedListener?

GaneshSPatil: @maheshp -- That's a good option. I will give it a try.

arvindsv:
@GaneshSPatil @ketan It's not a config-repo defined in a config-repo. It's not possible to do that. It's just a way of saying: "The pipeline's material is the same as that of the config-repo that it is defined in". Not sure if it changes anything.

ketan:
>
> @GaneshSPatil <https://github.com/GaneshSPatil> @ketan
> <https://github.com/ketan> It's not a config-repo defined in a
> config-repo. It's not possible to do that. It's just a way of saying: "The
> pipeline's material is the same as that of the config-repo that it is
> defined in". Not sure if it changes anything.
>

Thanks for confirming

Issue #5665, opened 2019-01-04
Api to add/remove users to roles


akshaydewan:
Related to: #5539
We need the ability to add or remove users to a GoCD Role.

Bump the Roles API to v2 and introduce:
`PATCH /go/api/admin/security/roles`
```json
{
  "operations": [
    {
      "role": "role1",
      "users": {
        "add": [
          "user1",
          "user2"
        ],
        "remove": [
          "user3",
          "user4"
        ]
      }
    },
    {
      "role": "role2",
      "users": {
        "add": [
          "user2"
        ]
      }
    }
  ]
}
```
```

adityasood:
verified on 19.1.0-8424

PR #5680, opened 2019-01-08
Bulk Update Users in Roles

akshaydewan:
For Issue: #5665

- Bumped up Roles API to v2 - copied from v1
- Added a PATCH endpoint
- Added `RolesConfigBulkUpdateCommand`. The request data model is in the server module so that data can be passed from the API to the command
- The response is the list of all **GoCD roles** in case of success, or an error message in case of failure.
- The case where role doesn't exist is validated. I don't think any other validations are needed.

Thoughts:
- I've copied over the `Command` pattern used in other classes, but I think some refactoring is needed. The `EntityConfigUpdateCommand` interface itself is kinda smelly IMO 
  - weird temporal coupling - it looks like isValid should be called before update, but it happens the other way round. 
  - The `result` object is mutated all over the place.
  - The passing around of preProcessedConfig is confusing. A bit of immutability would have helped. But easier said than done.


akshaydewan:
@maheshp With this change, do we need to add `RolesConfig.class` to `SecurityConfigChangeListener` ?

maheshp:
> @maheshp With this change, do we need to add RolesConfig.class to SecurityConfigChangeListener ?

Yes, we need to add that. I will merge this PR and send a PR separately for this.

PR #5683, opened 2019-01-09
Show cancelled by on the dashboard on hovering over a stage.

varshavaradarajan:
@GaneshSPatil - we had spoken about showing the cancelled by information instead of `triggered by` on the pipeline. I thought that the title on hover where we show the stage status was more appropriate. Let me know what you and @naveenbhaskar think.

screen shot 2019-01-08 at 10 26 35 pm

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
+++ b/server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
@@ -70,6 +70,10 @@ export default class {
     }
   }
 
+  static showDashboardPath(): string {
+    return "/go/api/dashboard";
+  }
+
   static DataSharingSettingsPath(): string {
     return "/go/api/data_sharing/settings";
   }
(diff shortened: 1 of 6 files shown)


GaneshSPatil:
I am OK with showing this information on hover of a cancelled stage. Initially, it used to say `up42_stage cancelled`. Not it will say `up42_stage cancelled by Bob`.

\cc: @naveenbhaskar @maheshp

server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
```diff
@@ -70,6 +70,10 @@ export default class {
     }
   }
 
+  static showDashboardPath(): string {
+    return "/go/api/dashboard";
```
GaneshSPatil: @varshavaradarajan -- This should accept a view name and pass it along as a query param.

varshavaradarajan: Oh okay. Will fix it. Sorry, didn't check with the PR run.

Issue #5682, opened 2019-01-09
Add validation for PATCH endpoint in System Admins API


akshaydewan:
Related to #5558
Validations to be added:
- When trying to remove a user/role from the list of admins, verify that the user/role is actually an admin. This applies only to the API, and is not relevant when doing a config xml save.
- Verify that the role exists when trying to add it to the list of admins. This validation should already be implemented on the config-xml save, and so it should automatically work for the API as well.

Since we don't validate usernames in the config-xml while adding or removing admins, we won't validate them in the API either.

adityasood:
Verified on 19.1.0-8424. The messaging needs to be improved for the following

1. When trying to add a non-existent role as an admin, today the API does not give back a meaningful message. Issue to track this is https://github.com/gocd/gocd/issues/5742

PR #5684, opened 2019-01-09
Validation to check if admins to remove are actually admins

akshaydewan:
Issue: #5682

Example:
```bash
curl -X PATCH \
  http://localhost:8153/go/api/admin/security/system_admins \
  -u 'admin:badger'
  -d '{
     "operations": {
     	"users": {
     		"remove": ["fearless rat"]
     	},
     	"roles": {
     		"remove": ["venomous pool frog"]
     	}
     	
     }
}'
```

Response:
~~404 not found~~
```
422 unprocessable entity
{
    "message": "Update failed because some users or roles do not exist under super admins.",
    non_existent_users: ['fearless rat'],
}
```


GaneshSPatil:
> Would 422 be a better response code instead of 404?

@akshaydewan -- Yes, I too think 422 will be a better response code instead of 404.

referring to the same example from the description,

If we provide, both non-existing user and role to remove. It will only tell either users or roles. But not both.
We've faced a similar problem as part of bulkUpdateUsers command. The response was:
```
{
      message           : "Deletion failed because some users were either enabled or do not exist.",
      non_existent_users: ['Bob'],
      enabled_users     : ['John']
}
```

How about keeping something similar this as well?
```
{
      message           : "Deletion failed because some users or roles do not exist under super admins.",
      non_existent_users: ['Bob'],
      non_existent_roles  : ['committers']
}
```

akshaydewan:
Updated the response code to 422 and added `non_existent_users` and `non_existent_roles`

PR #5686, opened 2019-01-09
Adding Server version information in v2 ServerInfoProcessor

SanjanaB:
The issue aims to expose Go Server version information to the api consumers.
 Fixes GoCD api changes required to address the [issue](https://github.com/gocd-private/azure-elastic-agent-plugin/issues/10)

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/server/service/plugins/processor/serverinfo/v2/MessageHandlerForServerInfoRequestProcessor2_0Test.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/server/service/plugins/processor/serverinfo/v2/MessageHandlerForServerInfoRequestProcessor2_0Test.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


server/src/test-fast/java/com/thoughtworks/go/server/service/plugins/processor/serverinfo/v2/MessageHandlerForServerInfoRequestProcessor2_0Test.java
```diff
@@ -0,0 +1,52 @@
...
+    public void shouldSerializeServerConfigToJSON() {
+        ServerConfig serverConfig = new ServerConfig();
+        serverConfig.ensureServerIdExists();
+        serverConfig.setSecureSiteUrl("https://example.com:8154/go");
+        serverConfig.setSiteUrl("http://example.com:8153/go");
+        CurrentGoCDVersion goCDVersion = CurrentGoCDVersion.getInstance();
+
+        MessageHandlerForServerInfoRequestProcessor2_0 processor = new MessageHandlerForServerInfoRequestProcessor2_0();
+
+        assertThat(processor.serverInfoToJSON(serverConfig),
```
ketan: Use `assertThatJson` instead of matching JSON strings. There's no guarantee of ordering of the different keys in JSON. This is likely to cause flaky tests.

SanjanaB: @ketan assertThatJson expects a json object. but the return type here is String. We could otherwise use `JSONAssert.assertEquals`.

PR #5687, opened 2019-01-09
Change to inline function definition over function hoisting

rajiesh:
Needed this change for sahi functional tests which run on old firefox where XhrPromise fails to load. This has caused pipeline history, edit pages broken on old firefox.

Tested it locally, pipeline export feature is not affected by this change. Changes were as per @GaneshSPatil's suggestion

@ketan could you please have a look and merge it if no issues

### DIFF
--- a/server/webapp/WEB-INF/rails/app/assets/javascripts/application.js
+++ b/server/webapp/WEB-INF/rails/app/assets/javascripts/application.js
@@ -53,7 +53,6 @@
 //= require "lib/often-0.3.2.js"
 //= require "lib/component-emitter-1.2.1.js"
 //= require "plugin-endpoint.js"
-//= require "xhr_promise.js"
 //= require "gocd-link-support.js"
 //= require "plugin-endpoint-request-handler.js"
 //= require "ansi_up.js"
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
@ketan @rajiesh @marques-work -- a bit of context.

`xhr_promise.js` was added into `application.js`. 
In production mode. Just one file is created for the entire javascript, known as `application.js`.
In `xhr_promise.js`, we've defined methods by using style `function FN_NAME(){}`. These functions will be hoisted at the top of the file. If any error happens while defining such methods, the program would quit, without evaluating the rest of the javascript code.

In production, as there is only one file (`application.js`), any error happens while loading the javascript will cause rest of the javascript not be evaluated.

We've defined `NotSupported` method under an if statement in `strict mode`. which will cause following error: `function statement not at top level of a program or function is prohibited`.

As this error happened in the hoisted method block. The rest of the code from `application.js` did not get evaluated, resulting in not defining `Util`, `JQuery` modules. Which was the reason all the old pages started failing for `Util is not defined` error.

Here is what we've done:
* Change the function definition from hoisted function style to scoped function definition style. So even if anything fails, it will fail for the current code and anything below it. will not fail from the beginning.
* And, moved the require of `xhr_promise.js` at the bottom of all the requires, which will make sure that all the other libraries are loaded as it was loading previously. Even if anything fails while importing this new library, it will affect only itself.

PR #5704, opened 2019-01-14
Fix config repo spa (#5650)

GaneshSPatil:
* Fix config repo SPA to work with newer config repo API contract.


GaneshSPatil:
[Jasmine job](https://build.gocd.org/go/tab/build/detail/build-windows-PR/4207/build-non-server/2/jasmine) is green.

PR #5706, opened 2019-01-14
Drain mode spa ui fixes

GaneshSPatil:
* Learn more link:
screen shot 2019-01-14 at 4 33 24 pm
* Scheduled Stages Tooltip:
screen shot 2019-01-14 at 4 33 47 pm


server/webapp/WEB-INF/rails/webpack/views/components/tooltip/index.scss
```diff
@@ -14,40 +14,49 @@
  * limitations under the License.
  */
 @import "../../global/common";
-@import "../icons/index";
```
GaneshSPatil: @ketan -- removed explicitly importing icons styles, which was causing styles duplication.
Instead, we've changed the HTML structure of the tooltip component, so it just uses icon component.

PR #5707, opened 2019-01-14
Remove the pipeline config old implementation

vrushaliwaykole:
The page was throwing a 500 server because a rails route was broken.

Since we are considering re-building the page, it seems better to
throw this code away.


GaneshSPatil:
Not sure if anything from `shared/` could also be deleted. 
I remember, `shared/plugin_infos.js` was only used on the pipeline config SPA, which can be deleted.

ketan:
`shared/plugin_infos` seems to be used in a few places. I've added (a long time ago) a plugin that emits warnings when running webpack, when unused files are found:

```
WARNING in 
UnusedFilesWebpackPlugin found some unused files:
webpack/helpers/server_health_messages_helper.js
Child html-webpack-plugin for "_specRunner.html":
     1 asset
    Entrypoint undefined = _specRunner.html
    [./node_modules/html-webpack-plugin/lib/loader.js!./spec/webpack/_specRunner.html.ejs] 2.03 KiB {0} [built]
    [./node_modules/webpack/buildin/global.js] (webpack)/buildin/global.js 472 bytes {0} [built]
    [./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 497 bytes {0} [built]
        + 1 hidden module
Done in 43.84s.
```

GaneshSPatil:
@ketan -- Thanks for confirming. :)

@vrushaliwaykole -- can we change the commit message to say `Remove Pipeline config SPA` instead of saying `Remove the config repos page.`?

maheshp:
@ketan @GaneshSPatil I am wondering if any of the code could be used when plan to re-write e.g there are a bunch of models.

Issue #5703, opened 2019-01-14
Bump up bundled ConfigRepo plugin 


maheshp:
Bump up the version of the bundled config_repo plugins to support pipeline export for GoCD 19.1.0.

maheshp:
@marques-work can you please take care of this please. You might have to work with @tomzo to get a stable release of both the ConfigRepo plugins and bundle the appropriate version with GoCD.

tomzo:
I can make the PR, just tell me if the latest versions were tested enough.

marques-work:
@tomzo I just pushed up a couple of tiny PRs for more consistent CLI output. After merging those, I would say let's version bump and make a new release. In terms of testing, these have not been run on build.gocd.org, and I've only run them locally. Let's build new releases, and then deploy them on build.gocd.org and run it for a day or so to make sure things still work as intended. After that, if there are no issues, you can mark those releases as stable. Sound ok?

marques-work:
@maheshp I merged @tomzo’s updates - next deploy should ensure build.gocd.org is running the latest versions and lets just make sure things run smoothly. Then I will ask @tomzo to mark them as stable if all goes well

PR #5710, opened 2019-01-15
bump versions of config-repo plugins #5703

tomzo:
@marques-work I've merged and released your last PRs in both repos. 
You can merge this, unless something new has come up.


marques-work:
@tomzo Thanks for updating the plugins! Merging this so we can dogfood it.

PR #5713, opened 2019-01-15
Update versions of bundled auth plugins to 1.0.3

arvindsv:
@ankitsri11 has tried these out. Bundling the new version with less verbose logging.

/cc @rajiesh @adityasood @ankitsri11


arvindsv:
Also related: https://github.com/gocd/gocd/commit/41eac8e12151e0bf735c53e4463484f23870185d - which fixed what was wrong with this PR.

PR #5721, opened 2019-01-17
Remove unnecessary `CAST` on the a sql statement

ketan:
Unsure why this was added, but it appears that the cast is redundant.


arvindsv:
@ketan These things ... somehow have a tendency to bite us, especially on old, big instances. Please consider a system property around this (defaulting to no casting) temporarily? For only a few releases.

ketan:
Merging after putting in a guard with a system property.

Issue #5564, opened 2018-12-12
User summary page epic


vrushaliwaykole:
Epic story to track progress of the new user summary page:

- [x] Show list of users (#5553)
- [x] Search API for user search (#5560)
- [x] API changes (#5558)
- [x] Search modal for users page (#5563)
- [x] Ability to add users using interface, (or just rely on search to be able to add users)?
- [ ] Ability to perform bulk operations from UI (enable/disable/delete, change roles, make admin...)

GaneshSPatil:
related issues: https://github.com/gocd/gocd/issues/71

rajiesh:
Verified on 19.2.0 (8639-9ba73d561b80d9d4e20dbf903aaeb3d29fcd3ab4)

PR #5722, opened 2019-01-17
(#5564) Added banner on user summary page if there is no admin tag (no specifically added admins) in config file.

vrushaliwaykole:


server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
```diff
@@ -154,6 +164,10 @@ export class UsersPage extends Page<null, State> {
     });
   }
 
+  private getMeta() {
+    return document.body.dataset.meta ? JSON.parse(document.body.dataset.meta) : {};
```
ketan: HTMLElement.dataset. TIL.

Maybe you want to improve this: https://github.com/gocd/gocd/blob/048a28e840df8d255c1552c430312925963a581c/server/webapp/WEB-INF/rails/webpack/helpers/spa_base.tsx#L72-L84

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/UsersController.java
```diff
@@ -52,8 +56,12 @@ public void setupRoutes() {
     }
 
     public ModelAndView index(Request request, Response response) {
+        boolean isSecurityEnabled = this.securityService.isSecurityEnabled();
```
ketan: This does not look right. I think we ought to be explicitly checking if the `<admin/>` tag is present. See if you can delegate to `SecurityConfig#noAdminsRequired` through the security service?

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/spring/VelocityTemplateEngine.java
```diff
@@ -51,6 +52,8 @@ public String render(ModelAndView modelAndView) {
             VelocityContext context = initialContextProvider.getVelocityContext((Map) model, controller, modelAndView.getViewName());
             StringWriter writer = new StringWriter();
             template.merge(context, writer);
+            Object meta = context.get("meta");
+            context.put("meta", new Gson().toJson(meta));
```
ketan: `Gson` instances are slightly heavy object to construct. Consider putting it in a static variable.

spark/spark-spa/src/main/resources/velocity/layouts/component_layout.vm
```diff
@@ -50,7 +50,9 @@
       data-version-copyright-year="${currentVersion.copyrightYear()}"
       data-version-formatted="${currentVersion.formatted()}"
       data-version-full="${currentVersion.fullVersion()}"
-      data-is-server-in-drain-mode="${drainModeService.isDrainMode()}">
+      data-is-server-in-drain-mode="${drainModeService.isDrainMode()}"
+      data-meta="${meta}"
```
ketan: this might need some escaping, esp if `meta` contains quotes, or other HTML chars that can possibly break.

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/UsersController.java
maheshp: Yes, this doesn't seem right. All users are considered admins even if Security is enabled but if no admins are defined. Check - https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/SecurityConfig.java#L91

server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
vrushaliwaykole: Done

PR #5726, opened 2019-01-17
Bring back template support pipeline export

marques-work:
1. Reverts the commit that blocked export on pipelines using templates
2. Export pipeline configs based on `configForEditing` so we don't denormalize the config we are exporting

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
@@ -248,12 +248,12 @@ public class GoConfigService implements Initializer, CruiseConfigProvider {
         return getCurrentConfig().hasPipelineNamed(pipelineName);
     }
 
-    public PipelineConfig pipelineConfigNamed(final CaseInsensitiveString name) {
-        return getCurrentConfig().pipelineConfigByName(name);
+    public PipelineConfig editablePipelineConfigNamed(final String name) {
+        return getConfigForEditing().pipelineConfigByName(new CaseInsensitiveString(name));
     }
 
-    public PipelineConfig pipelineConfigNamed(final String name) {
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/thoughtworks/go/server/service/GoConfigService.java
```diff
@@ -248,12 +248,12 @@ public boolean hasPipelineNamed(final CaseInsensitiveString pipelineName) {
         return getCurrentConfig().hasPipelineNamed(pipelineName);
     }
 
-    public PipelineConfig pipelineConfigNamed(final CaseInsensitiveString name) {
-        return getCurrentConfig().pipelineConfigByName(name);
+    public PipelineConfig editablePipelineConfigNamed(final String name) {
+        return getMergedConfigForEditing().pipelineConfigByName(new CaseInsensitiveString(name));
```
maheshp: @marques-work I am little paranoid changing this method as it is used in multiple places and as of now I don't know the repercussion of changing it.

I understand why you had to make this change as the `getCurrentConfig()` returns a preprocessed config which would have a pipeline using templates to show all the stage details.

 How bout having a new method for fetching the pipeline from `GoConfigService#getConfigForEditing()`?

maheshp: @marques-work sorry my bad, I think you have changed only the method used by `ExportControllerV1` ignore my previous comment on being used in multiple places.

arvindsv:
@maheshp - I was supposed to merge this, after renaming the method  `editablePipelineConfigNamed` to `pipelineConfigForEdit`. :( We should rename it sometime later.

maheshp:
@arvindsv no problem :). I should have renamed it as well, will do it.

PR #5728, opened 2019-01-18
Remove dependence on some rails gems

ketan:
`rails` was an "empty" gem that merely dependended on a set of gems to
offer a bunch of functionality for webapps. Some of this was unsed by
gocd, and was continued to be packaged along with the installer, along
with their transitive dependencies.

This PR gets rid of `rails`, specifically to remove the following unused
gems:

- actioncable
- actionmailer
- activejobs
- activerecord
- activestorage


server/webapp/WEB-INF/rails/Gemfile.lock
```diff
@@ -110,19 +80,6 @@ GEM
...
-      actionpack (= 5.2.1)
-      actionview (= 5.2.1)
-      activejob (= 5.2.1)
-      activemodel (= 5.2.1)
-      activerecord (= 5.2.1)
-      activestorage (= 5.2.1)
-      activesupport (= 5.2.1)
-      bundler (>= 1.3.0)
-      railties (= 5.2.1)
-      sprockets-rails (>= 2.0.0)
```
ketan: This is the list of gems that `rails` depended on. We only use the subset that's `require`d in `application.rb` https://github.com/gocd/gocd/blob/814b9f8f98fe90e09ffe277cf4b3e1752ae45c9f/server/webapp/WEB-INF/rails/config/application.rb#L3-L8

`require 'rails'` is actually provided by railtie gem.

prateekbaheti:
@ketan changes make sense considering the gems that are being loaded and used by the app. Some checks seem to be failing (jasmine build).

ketan:
Closing, not yak-shaving jasmine anymore.

PR #5731, opened 2019-01-18
Test connection support for ConfigRepo create/edit modal

vishaldevgire:
check_connection


server/webapp/WEB-INF/rails/webpack/views/components/flash_message/index.scss
```diff
@@ -23,11 +23,16 @@
...
   margin:        0 0 10px 0;
   position:      relative;
-
+  overflow:      auto;
+  max-height:    200px;
   p {
     margin: 0;
   }
-
+  pre {
```
vishaldevgire: `margin: 0` to reset <pre> element's default margin

server/webapp/WEB-INF/rails/webpack/views/components/flash_message/index.scss
```diff
@@ -23,11 +23,16 @@
...
+    margin: 0;
+    white-space: pre-wrap;
+    word-wrap: break-word;
```
vishaldevgire: Styles required to avoid a horizontal scroll.

GaneshSPatil:
@vishaldevgire -- Do we need to show `Checking Connection` as a message?
IIRC, it used to be a spinner next to the button on older styles.

Also, do we need a vertical scroll? As Without fixing this error, the user can't proceed. Does it make sense to show the entire error (without clipping)?

ketan:
>  IIRC, it used to be a spinner next to the button on older styles.

Yes — button with spinner. The spinner would turn to a check mark, and the button would go green.

In case of an error, the button would have an `!` mark on it, with the color turning red/orange. And an error message, like we show now.

server/webapp/WEB-INF/rails/webpack/views/components/flash_message/index.scss
ketan: Do not use global styles like these `p` and `pre` will apply globally. Only apply styles on classes.

server/webapp/WEB-INF/rails/webpack/models/config_repos/config_repos_crud.ts
```diff
@@ -17,14 +17,18 @@
...
-  const configurations = o.createConfigurationsFromText();
+function toSnakeCaseJSON(o: object) {
   const text           = JSON.stringify(o, s.snakeCaser);
-  const json           = JSON.parse(text);
+  return JSON.parse(text);
+}
+
+export function configRepoToSnakeCaseJSON(o: ConfigRepo) {
+  const configurations = o.createConfigurationsFromText();
+  const json           = toSnakeCaseJSON(o);
```
ketan: I can see why this has been done. Maybe add a comment why?

server/webapp/WEB-INF/rails/webpack/views/components/flash_message/index.scss
vishaldevgire: @ketan But these styles are under class `.callout`, so they would not be applied globally.

server/webapp/WEB-INF/rails/webpack/models/config_repos/config_repos_crud.ts
vishaldevgire: The intention was to separate the logic of converting object to JSON (`toSnakeCaseJSON ()`) and config repo specific logic. So that `toSnakeCaseJSON()` can be reused for other objects like `Material`

vishaldevgire:
@GaneshSPatil @ketan 

Old check connection:
old_check_connection

The new behaviour is quite similar to old test connection interface,  as you can see in old test connection, there is no spinner just a message `checking connection...` also the button doesn't change color. 

@GaneshSPatil Showing a message `checking connection...` is good idea than keeping the user guessing if there is any activity going on from the time test button is pressed till result is returned.

Question for: @ketan @GaneshSPatil @naveenbhaskar 

Should I go ahead with the following suggested changes:

1. A spinner when button is pressed? or just simple test connection text (instead of `info` FlashMessage)
2. On success, the spinner/text would turn to a check mark
3. In case of an error, the button would have an ! mark on it

I am assuming 
 - that there are no concerns about success/error message styling
 - button color change is not required

vishaldevgire:
> Also, do we need a vertical scroll? As Without fixing this error, the user can't proceed. Does it make sense to show the entire error (without clipping)?

We added vertical scroll because the error message is not of fixed size and it can be of any length.

But I am torn on this one.

naveenbhaskar:
For check connection, i think we can use a button with spinner inside  and changes to green on success. and exclamation mark on red color if failed. @vishaldevgire .. Pls make this change.

vishaldevgire:
Here is how it looks after incorporating suggested changes:

really_new_demo

GaneshSPatil:
@vishaldevgire -- I didn't find the tests for the API Operations neither for the newly added UI component.

Can we please add a few tests to verify the API request and the behavior of the UI component when Checking connection passes, fails or is in-progress.

Issue #5688, opened 2019-01-09
Functional issues in pipeline export feature


rajiesh:
##### Issue Type

- Bug Report

##### Summary

This issue captures all the issues related to pipeline export, it can also include regression issues caused by the pipeline export related code changes. Please add the issue description as the task list below, more description to the issue can be provided in comments and its  link added to task list below,

##### Issues

- [x] On `go/admin/pipelines` page `move to` option to move pipelines across groups is missing. @ketan pointed out that [this](https://github.com/gocd/gocd/pull/5509/files#diff-28e552efa7cda9715ac6567d9ec79858L2054) css change could be the issue. This was test issue, Kiera fixed it by [this](https://github.com/gocd/ruby-functional-tests/pull/282) commit 

- [x] Currently group admins cannot export a pipeline, API return 403 response. Shouldn't group admins have access to export pipelines on the group they are admins? IMO group admins should have access to export pipelines from the group they are admins.

- [ ] Exported json has `timeout` at task level. Timeout is applicable only at the job level settings. Also Job timeout value for `Never` should be integer `0` not float `0.0`. However timeout at task level or the float value for job timeout does not affect the pipeline, pipeline still works as expected. Please refer to the json [here](https://github.com/gocd/gocd/issues/5688#issuecomment-452983983)

- [x] ~~Export as option is available for pipeline coming from config repo too. It doesn't cause any issue but doesn't make sense to have export feature for config repo pipelines. Screenshot [here](https://github.com/gocd/gocd/issues/5688#issuecomment-452985332).~~ After watching the demo video by Marques could see whats the intention behind providing export option for config repo pipelines, striking this point off

- [x] Pipeline export should list only plugins with export capabilities. More details [here](https://github.com/gocd/gocd/issues/5730)

rajiesh:
A exported pipeline in JSON,

```
{
  "group": "defaultGroup",
  "name": "test-export",
  "label_template": "${COUNT}",
  "lock_behavior": "none",
  "environment_variables": [],
  "parameters": [],
  "materials": [
    {
      "url": "https://github.com/gocd/gocd",
      "branch": "master",
      "shallow_clone": true,
      "filter": {
        "ignore": [],
        "whitelist": []
      },
      "auto_update": true,
      "type": "git"
    }
  ],
  "stages": [
    {
      "name": "defaultStage",
      "fetch_materials": true,
      "never_cleanup_artifacts": false,
      "clean_working_directory": false,
      "approval": {
[...]

PR #5735, opened 2019-01-19
Allow pipeline group admins to export pipelines in groups they see

kierarad:
This should fix the second issue described here - https://github.com/gocd/gocd/issues/5688

Previously the filter chain required that a user be an admin to access "/api/admin/export/**" so group admin requests never reached the API control. We whitelisted users (ROLE_USER) access so that the permissions could be handled at the controller level. 

Refactored a little so that the groupName query param is no longer required and can be determined by the pipeline name in the request url.


ketan:
Closing in favor of #5748

PR #5744, opened 2019-01-21
Pull out user authorization check into its own function

akshaydewan:
Related to #2780 

Just pulling out `isUserAuthorized` into its own function. This will be followed by a clean up of `canContinue` and the validation functions in subsequent PRs.


server/src/main/java/com/thoughtworks/go/config/GoConfigDao.java
```diff
@@ -89,7 +89,7 @@ public void updateConfig(EntityConfigUpdateCommand command, Username currentUser
         synchronized (GoConfigWriteLock.class) {
             try {
                 LOGGER.info("Config update for entity request by {} is being processed", currentUser);
-                if (!command.canContinue(cachedConfigService.currentConfig())) {
+                if (!command.canContinue(cachedConfigService.currentConfig()) || !command.isUserAuthorized()) {
```
ketan: Do the (cheaper) authorization check earlier?

PR #5745, opened 2019-01-21
Implement the `get-icon` call for config repo plugins.

ketan:
This is identical to the elastic agent plugin, except the request name.


ketan:
Not sure how, but the icons are showing up on the plugins page already without any other changes:

screen shot 2019-01-21 at 9 09 00 pm

/cc @tomzo

PR #5752, opened 2019-01-22
Close the modal when clicking on the overlay outside it

akshaydewan:
The new modal component didn't close when we click the area outside it. This PR adds that functionality.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/modal/modal_manager.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/modal/modal_manager.tsx
@@ -44,7 +44,10 @@ export namespace ModalManager {
       view() {
         return (
           _.map(allModals, (entry) => {
-            return <div key={entry.key} data-test-id={entry.key} class={styles.overlayBg}>{m(entry.value)}</div>;
+            return <div key={entry.key}
+                        data-test-id={entry.key}
+                        class={styles.overlayBg}
+                        onclick={(event) => overlayBackgroundClicked(event, entry.value)}>{m(entry.value)}</div>;
           })
         );
       }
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/modal/modal_manager.tsx
```diff
@@ -56,17 +56,36 @@ export namespace ModalManager {
...
+    return (event: Event) => {
+      const overlayBg = document.getElementsByClassName(styles.overlayBg).item(0);
+      if (event.target === overlayBg) {
+        modal.close();
+      }
+    };
+  }
+
+  function addOutsideClickListener(modal: Modal) {
+    modalContainer.addEventListener("click", overlayBackgroundClicked(modal));
```
ketan: Possibility of multiple handlers being installed on the same container?

server/webapp/WEB-INF/rails/webpack/views/components/modal/modal_manager.tsx
```diff
@@ -56,17 +56,36 @@ export namespace ModalManager {
...
+  }
+
+  function removeOutsideClickListener(modal: Modal) {
+    modalContainer.removeEventListener("click", overlayBackgroundClicked(modal));
```
ketan: this will likely cause "leaks" because each invocation of `overlayBackgroundClicked` returns a different function. So technically a listener you add is not the same as the one you remove.

It's likely that the leaked listeners end up doing nothing, because the modal is closed. However they are leaking listeners, and possibly holding a copy of the modal instance itself (along with anything else that the modal hold onto)

akshaydewan: Ok, I'll look into it and fix it

akshaydewan:
> What happens with nested modals, do we need a test around it?
In the new components, nested modals are already broken. (Good thing we aren't using them anywhere). There are double overlays and the Escape key closes all the modals.

It will need some more work to fix and we can do that separately IMO.

akshaydewan:
@ketan I've changed the approach and added `onclick` directly to the `<div>`. I think that should fix the listener leak. LMKWYT

Issue #5694, opened 2019-01-10
Define global/plugin SCMs as a part of the pipeline JSON/YAML


ankitsri11:
EPIC: https://github.com/gocd/gocd/issues/5175

Currently, global/plugin SCMs configuration is defined as part of cruise-config.xml. User defining pluggableSCM in config-repo depends on this definition. 

For example:

SCM configuration in cruise-config.xml

```
<scms>
    <scm id="4937e2fe-be38-46ae-a8cc-65e38c68c15e" name="pull-request">
      <pluginConfiguration id="github.pr" version="1" />
      <configuration>
        <property>
          <key>url</key>
          <value>https://github.com/gocd/gocd.git</value>
        </property>
      </configuration>
    </scm>
  </scms>
```
PluggableSCM definition in config-repo pipeline

```
{
  "scm_id": "4937e2fe-be38-46ae-a8cc-65e38c68c15e",
  "destination": "destinationDir",
  "filter": {
    "ignore": [
      "dir1",
      "dir2"
    ]
  },
  "name": "github.pr",
  "type": "plugin"
}
```

User Experience:

- No need to define global/plugin SCMs configuration in cruise-config.xml for config-repo pipelines
- Allow user to define global/plugin SCMs configuration as part of config-repo pipelines

For example:

```
{
  "materials": [
    {
      "type": "scm-plugin",
      "pluginID": "github.pr",
      "configurations": {
        "url": "https://github.com/gocd/gocd.git"
      }
    },
    {
      "url": "http://my.git.repository.com",
      "branch": "feature",
      "filter": {
        "ignore": [
          "externals",
          "tools"
        ]
      },
      "destination": "dir1",
      "auto_update": false,
      "name": "gitMaterial1",
      "type": "git",
      "shallow_clone": true
    }
  ]
}
```

Note - GoCD needs to handle if two pipelines have the same SCM configuration. Should error out something like below:

> Cannot save SCM, found duplicate SCMs. pull-request, feature;; Cannot save SCM, found duplicate SCMs. pull-request, feature;;

/cc: @arvindsv @ibnc

tomzo:
> This mostly involves a change to https://plugin-api.gocd.org/current/config-repo/#parse-directory and a format_version change

And adding new CR classes and merging part in config domain, which will handle same/similar material definitions, also among the ones defined in xml. (Have a look at how environments are handled, it is a similar case).

[...]

PR #5756, opened 2019-01-22
Adding the ability to add SCMs through config repo

ibnc:
https://github.com/gocd/gocd/issues/5694


ibnc:
@tomzo Would you mind taking a look at what I've got thus far?

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -335,6 +339,14 @@ private EnvironmentsConfig mergeEnvironmentConfigs() {
             return environments;
         }
 
+        private SCMs mergeScms() {
+            SCMs scms = BasicCruiseConfig.this.scms;
```
tomzo: I am not sure it is safe to modify the `scms` instance taken from xml part, we probably should clone it and then append to a new list.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -344,9 +346,18 @@ private PackageDefinition getPackageDefinition(String packageId) {
     }
 
     private PluggableSCMMaterialConfig toPluggableScmMaterialConfig(CRPluggableScmMaterial crPluggableScmMaterial) {
-        SCMs scms = getSCMs();
         String id = crPluggableScmMaterial.getScmId();
```
tomzo: `id` can be null now, right?
It is `scm_id`, which in [this example](https://github.com/gocd/gocd/issues/5694#issuecomment-454978548) can be skipped.

tomzo: On a second thought, if we allow `scm_id` to be not specified by user then we have a problem, because it is still needed to identify the scm at global level. It is used for example in api responses.
Do we generate scm_id then? Must be globally unique and I guess it should not change on each merge.

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
```diff
@@ -335,6 +339,14 @@ private EnvironmentsConfig mergeEnvironmentConfigs() {
...
+            for (PartialConfig part : this.parts) {
+                scms.addAll(part.getScms());
```
tomzo: You'll also have to take care of deduplication here. If user specified 2 scms which have same definition, we should make it into 1 at global level.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
arvindsv: Correct. It should be generated based on a hash of the fingerprint or something of that sort, I think. Maybe it's not too much of a problem if it changes on each merge, since it's only used for a ref - but it depends on the implementation. Ideally, it doesn't change - just makes it simpler.

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
arvindsv: Agree. Will also need to take care of the case where an existing global SCM (in the config) is re-defined in the config repo.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
ibnc: Yes, but only if it's creating a new SCM. Plugin configuration is required to create a new SCM

If they're creating a new SCM, then they can choose to provide an `scm_id` if they want. (See [here](https://github.com/gocd/gocd/pull/5756/files#diff-5236c1c9c0b120c4ff0f543e908fd396R356)). If they don't provide an `scm_id` then the post constructor callback will set it for them ([post constructor](https://github.com/gocd/gocd/blob/abe03ce0f013ff18b766e2378de711f22cbaa2d0/config/config-api/src/main/java/com/thoughtworks/go/domain/scm/SCM.java#L328))

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
ibnc: Doesn't this already cover it? https://github.com/gocd/gocd/blob/e2b72aa28a5204e3947212b8e0af1686c939f788/server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java#L49-L53

tomzo: @ibnc if user specified 2 SCMs which have equal definitions but skipped `scm_id`, then post constructor callback would have assigned 2 random uuids and this would not prevent the duplicated instances from being inserted here.

config/config-api/src/main/java/com/thoughtworks/go/config/BasicCruiseConfig.java
ibnc: I'm having trouble understanding the logic behind some of the merge code. I'm specifically confused about the purpose of the `MergeEnvironmentConfig` and `MergePipelineConfigs`

ibnc:
@tomzo would you mind taking another look at my PR now that I've made some changes related to your feedback?

tomzo:
@ibnc looks OK, I left 2 small suggestions.
From my experience, it is good to test manually a few cases on development server. The most valuable would be:
 - add some SCMs in xml, then one in configrepo. Try to remove/add SCM in xml. Also through API.
 - push invalid SCM so that merge fails, Try to remove/add SCM in xml. Also through API. It should still work because of fallback merge.

arvindsv:
@ibnc Not sure, but I don't think this change should need a schema bump. Ideally, this shouldn't change the existing tags at all. I know that pipeline groups and even environments end up putting an extra tag in the config. I don't remember why.

Is that the reason the schema bump is needed? If so, can you explain the reason (and I'll try to remember this time)? Thank you!

ibnc:
@arvindsv It's so we can show something in the xml that isn't the full SCM config. I suspect environments and pipeline groups did this to create a visual distinction between locally and remotely configured items. This won't be needed after we remove the config xml page.

here's an example of what the schema change allows for:
```
<scms>
  <scm id="8e27f393-97e2-4572-983b-6219d9cf718a" name="wubba_lubba_dub_dub" />
</scms>
```

tomzo:
For environments and pipeline groups, there exists an empty tag to represent the main xml into which we are merging partials. This was added by @jyotisingh when fixing merge algorithm. Basically, when there is no entity defined in xml, we still need a kind of stub to represent the empty environment/group when merging partials.

ibnc:
Do we want to remove the schema bump then? And display either the full remote scm config or not display any of the remote configs?

tomzo:
> Do we want to remove the schema bump then? And display either the full remote scm config or not display any of the remote configs?

If the direction is to remove xml page in the future, then I guess there is no point in extending it now. I would not display any of the remote config. It only matters to return the entities via API.

ibnc:
@arvindsv thoughts?

ibnc:
@tomzo @arvindsv This is the PR that added the environment and pipeline group stubs to the xml https://github.com/gocd/gocd/pull/2256

ibnc:
It doesn't look like we need the scm stubs in the xml. I'll remove them unless there are any objections

@tomzo @arvindsv

ibnc:
@tomzo @arvindsv if everything looks good to y'all, then would one of you mind merging this?

Issue #5742, opened 2019-01-21
[User Summary] Improve messaging for the System Admins API Patch endpoint when a non-existent group is added


adityasood:
##### Issue Type
- Feature enhancement

##### Summary

Improve the messaging when trying to add a non-existent role as an admin, today the API does not give back a meaningful message.

```
curl -kv https://localhost:8154/go/api/admin/security/system_admins \
  -H 'Accept: application/vnd.go.cd.v2+json' \
  -H 'Content-Type: application/json' \
  -H 'X-GoCD-Confirm: true' \
  -X PATCH \
  -u 'user:password' \
  -d '{
     "operations": {
      "users": {
        "add": ["maryy"]
      },
      "roles": {
        "add": ["created-via-api1"] \\ non existen role
      }

     }
}'
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8154 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Request CERT (13):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: CN=in-adityas-2.local; OU=Cruise server webserver certificate
*  start date: Jan  1 00:00:00 1970 GMT
*  expire date: Jan 18 11:29:33 2029 GMT
[...]

PR #5760, opened 2019-01-23
Add validation error details to command result

akshaydewan:
Issue #5742

Example:
```
curl -v http://localhost:8153/go/api/admin/security/system_admins \
  -H 'X-GoCD-Confirm: true' \
  -X PATCH \
  -u 'admin:badger' \
  -d '{
     "operations": {
      "roles": {
        "add": ["created-via-api1"]
      }
     }
}'

< HTTP/1.1 422 Unprocessable Entity
{
  "message" : "Validations failed for admins. Error(s): [Role \"created-via-api1\" does not exist.]. Please correct and resubmit.",
  "data" : {
    "roles" : [ "created-via-api1" ],
    "users" : [ "admin" ],
    "errors" : {
      "roles" : [ "Role \"created-via-api1\" does not exist." ]
    }
  }
}
```

I've removed the `non_existent_users` and `non_existent_roles` fields from the json in this case because it can be confusing. Populating these fields was difficult in this case because the service/command doesn't know what was the exact validation failure in the entity.
In other cases where we try to revoke an admin who's not actually an admin, `non_existent_users` and `non_existent_roles` fields will continue to be rendered.

I added a hack to de-duplicate the error messages. They are duplicated because of the these lines in `AdminRole` (similar in `AdminUser`)

```java
public void addError(String message) {
        errors().add(NAME, message); // Do not remove this - The old view for editing group authorization, template authorization makes use of it.
        errors().add(ADMIN, message);
    }
```

Does anyone know the first line is still needed?

### DIFF
--- a/api/api-admins-config-v2/src/main/java/com/thoughtworks/go/apiv2/adminsconfig/representers/AdminsConfigRepresenter.java
+++ b/api/api-admins-config-v2/src/main/java/com/thoughtworks/go/apiv2/adminsconfig/representers/AdminsConfigRepresenter.java
@@ -34,6 +34,10 @@ public class AdminsConfigRepresenter {
         jsonWriter.addLinks(
                 outputLinkWriter -> outputLinkWriter.addAbsoluteLink("doc", Routes.SystemAdmins.DOC)
                         .addLink("self", Routes.SystemAdmins.BASE));
+        toJSONWithoutLinks(jsonWriter, admin);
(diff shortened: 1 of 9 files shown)


akshaydewan:
Discussed with @adityasood, and we're thinking of adding the `AdminsConfig` entity to the response, which will contain the errors, in addition to this `message`

PR #5765, opened 2019-01-24
footer UI fix

naveenbhaskar:


GaneshSPatil:
@ketan -- Are we going to address [the issue](https://github.com/gocd/gocd/pull/5765#issuecomment-458131814) mentioned by @rajiesh  in a separate PR?  or we wont fix it?

ketan:
It was not a regression, so I merged this. We can address that issue separately.

naveenbhaskar:
I will sent another PR for this

PR #5766, opened 2019-01-24
Add Toggle Admin Privileges Switch on Users Managemenr Page

GaneshSPatil:
* Introduce toggle button at user row level to change the admin privileges
	- Show Whether the current user is an admin by YES, NO or Not
	  Specified.
	- Show optional tooltip explaining making current user an
	  admin will cause all other users to loose admin privileges,
	  when no system admins are configured.
* Show and remove 'No admin users configured banner' on the fly,
  without required users to refresh the page.
* Remove 'Make System Admin' and 'Revoke System Admin' buttons.
* Remove 'meta' (noAdminsConfigured) sent from the server as part
  of the body attributes. Make use of System-Admins API to determine
  whether admins are configured or not.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/models/admins/admin_crud.ts
+++ b/server/webapp/WEB-INF/rails/webpack/models/admins/admin_crud.ts
@@ -29,6 +29,10 @@ export interface BulkUpdateSystemAdminJSON {
 export class AdminsCRUD {
   static API_VERSION_HEADER = ApiVersion.v2;
 
+  static all() {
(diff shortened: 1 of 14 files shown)


GaneshSPatil:
screen shot 2019-01-24 at 3 56 42 pm
screen shot 2019-01-24 at 3 56 52 pm
screen shot 2019-01-24 at 3 57 18 pm

GaneshSPatil:
Following more improvements can be done the Users Management SPA:
- Spinners for Page Operations
- Better Operation Messages
- Table styles could be improved more.
- Filter count for the individual filter element.

We can address these changes as part of next release.
\cc: @naveenbhaskar @ketan @maheshp

PR #5784, opened 2019-01-29
Fix styling for maintenance mode info on the SPA

GaneshSPatil:


GaneshSPatil:
### Before fix:
screen shot 2019-01-29 at 12 04 15 pm

### After fix:
#### For Desktops having the width more than 1366 (show it half width of the page):
screen shot 2019-01-29 at 12 03 06 pm

#### For Desktops having the width less than 1366 (show it full width of the page):
screen shot 2019-01-29 at 12 03 36 pm

PR #5785, opened 2019-01-29
Use helpers to build all documentation URLs

ketan:
There was inconsistency in how a lot of the doc URLs were generated:

- some URLs used `/current/`
- others urls used the current version number to build a URL
- some others did not use either

The new implementation exposes `docsUrl()` and `apiDocsUrl()` helper
methods in Java, JavaScript and TypeScript to be able to consistently
generate URLs everywhere.

The typescript helper is auto-generated at build time and exposes the
version numbers that may be useful in some places (in footer, for e.g.)


server/webapp/WEB-INF/rails/lib/tasks/gen_webpack_js_routes.rake
```diff
@@ -39,4 +39,37 @@ task generated_js: :environment do
     f.puts("// This file is automatically generated by `generated_js` task. Any changes will be lost")
     f.puts TsRoutes.generate(include: JsRoutes.configuration.include)
   end
+
+
+  open(Rails.root.join("#{ENV['OUTPUT_DIR']}/gocd_version.ts"), 'w') do |f|
```
ketan: This is where the typescript helper is at. Exposes identical data as in the `CurrentGoCDVersion` class.

PR #5755, opened 2019-01-22
Display the time at which a pipeline was paused (#5719)

varshavaradarajan:
fixes #5719

PR #5790, opened 2019-01-30
Rename migration from 1902002_add_paused_at to 1902001_add_paused.

varshavaradarajan:
@ketan - sorry, just noticed the filename in https://github.com/gocd/gocd/pull/5755/files. Sorry. Hopefully, build.go.cd has not been upgraded since the merge.


varshavaradarajan:
@ketan - please check if everything comes up alright before merging. I have. But, not sure if I need to do anything else here.

Issue #5761, opened 2019-01-23
Environments page: Not showing agents through config-repos?


arvindsv:
##### Issue Type

- Bug Report

##### Summary

See: https://github.com/tomzo/gocd-json-config-plugin/issues/36

When agents are assigned to environments through a config-repo, they don't seem to be showing in the environments page: `http://gocd-server/go/admin/environments`

###### Basic environment details

* Go Version: 18.12

/cc @mbotmcc @tomzo @ketan

akshaydewan:
We probably need to think a bit about the interaction between Agents, Environments and Pipelines between config-repos and those configured from the UI. (Apologies if this has already been discussed)

Let's consider that we have:
2 pipelines: one configured from config-repo (named `firstpipe`), the other configured from the UI (named `up42`).
2 environments: one configured from config-repo (named `dev`), the other configured from the UI (named `test`). `firstpipe` belongs to `dev` environment and `up42` belongs to `test`
1 agent: Shared between both the environments/pipelines.

What I see on my box is that the sharing of agent does not work - As long as the agent has the `test` environment assigned from the UI, the `firstpipe` pipeline doesn't get assigned to the agent. When I remove the `test` environment from the agent, the job gets picked up without re-triggering it. (I didn't debug it, so I don't know what's causing this behaviour).

From the UI point of view, should the `dev` environment defined from the config-repo show up in the dropdown for Envrionments assignment on the Agents page?

ketan:
> From the UI point of view, should the `dev` environment defined from the config-repo show up in the dropdown for Envrionments assignment on the Agents page?

Doing that would affect the config XML (by defining a "new" `dev` environment in the config xml), if that's what you're asking/implying. This is already doable by editing the config xml.

I'm not sure if it will be obvious to the end user that this assignment may actually end up creating a new environment. So IMO, the questions really should be:

- do we want to make it obvious that assignment will end up creating a new `<environment/>` in the config xml?
- do users really care where the environments are defined, or do they just care about the end result — that the agent is assigned to the environment?

tomzo:
In older UI the environments page was showing environments, pipelines and agents which are defined via config-repo in grayscale and user could not change those. But then it got lost at some point when UI was updated.
I think user does not care unless something is impossible to change. I guess we should be displaying which elements of the config are read-only due to config-repo definition.

arvindsv:
### Current state:

[...]

PR #5796, opened 2019-01-31
Show Agents assigned to Environments from Config Repos

akshaydewan:
Issue #5761

This change is to fetch all the agents (instead of just local agents) so that agents associated via Config Repo can be displayed on the Environments page.

When a user tries to Edit agents on this page, the ones configured via Config Repo will be checked and disabled (similar to Pipelines and Environment Variables configured via the Config Repo).

screenshot 2019-01-31 at 12 29 35
screenshot 2019-01-31 at 12 29 49

### DIFF
--- a/server/webapp/WEB-INF/rails/app/helpers/agents_helper.rb
+++ b/server/webapp/WEB-INF/rails/app/helpers/agents_helper.rb
@@ -18,15 +18,15 @@ module AgentsHelper
 
   AgentStatus = com.thoughtworks.go.domain.AgentStatus
 
-  def agent_selector uuid, selector_name, selected_uuids, is_remote_agent=false
-    "<td class='selector'>#{agent_selector_without_cell(uuid, selector_name, selected_uuids, is_remote_agent)}</td>".html_safe
+  def agent_selector uuid, selector_name, selected_uuids, is_associated_through_config_repo=false
+    "<td class='selector'>#{agent_selector_without_cell(uuid, selector_name, selected_uuids, is_associated_through_config_repo)}</td>".html_safe
   end
 
-  def agent_selector_without_cell uuid, selector_name, selected_uuids, is_remote_agent
+  def agent_selector_without_cell uuid, selector_name, selected_uuids, is_associated_through_config_repo
     disabled_attr, title_attr = "", ""
-    if is_remote_agent
+    if is_associated_through_config_repo
       disabled_attr = "disabled"
-      title_attr = "title='Agent is associated in a Config Repository'"
+      title_attr = "title='Agent is associated in a Config Repository, cannot edit'"
     end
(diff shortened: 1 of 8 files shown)


maheshp:
@akshaydewan does it make sense to have some indication why the checkbox is disabled, maybe a tooltip?

akshaydewan:
Yes, a tooltip would be easiest to implement. Another option could be to split the list items similar to the pipelines modal:
screenshot 2019-01-31 at 19 38 22

arvindsv:
Yes, please. Either one of those options should be ok.

akshaydewan:
@maheshp I've added the tooltip and fixed the tests that were failing.

server/webapp/WEB-INF/rails/app/helpers/agents_helper.rb
```diff
@@ -18,12 +18,17 @@ module AgentsHelper
 
   AgentStatus = com.thoughtworks.go.domain.AgentStatus
 
-  def agent_selector uuid, selector_name, selected_uuids
-    "<td class='selector'>#{agent_selector_without_cell(uuid, selector_name, selected_uuids)}</td>".html_safe
+  def agent_selector uuid, selector_name, selected_uuids, is_remote_agent=false
```
maheshp: I was confused when I saw `is_remote_agent`. Should we consider renaming to something like `is_associated_through_config_repo`.

 I guess the reason for using remote is because of the method `MergeEnvironmentConfig#containsAgentRemotely`

server/webapp/WEB-INF/rails/app/helpers/agents_helper.rb
```diff
@@ -18,12 +18,17 @@ module AgentsHelper
...
+    "<td class='selector'>#{agent_selector_without_cell(uuid, selector_name, selected_uuids, is_remote_agent)}</td>".html_safe
   end
 
-  def agent_selector_without_cell uuid, selector_name, selected_uuids
-    "<input type='checkbox' name='#{ERB::Util.h(selector_name)}' value='#{ERB::Util.h(uuid)}' class='agent_select' #{(selected_uuids && selected_uuids.include?(uuid)) ? "checked='true'" : ""}/>".html_safe
+  def agent_selector_without_cell uuid, selector_name, selected_uuids, is_remote_agent
+    disabled_attr, title_attr = "", ""
+    if is_remote_agent
+      disabled_attr = "disabled"
+      title_attr = "title='Agent is associated in a Config Repository'"
```
maheshp: How bout `Agent is associated in a Config Repository, cannot edit.`?

maheshp:
@akshaydewan we will have to put a similar fix on the agents page as well. Do you want to do that as part of a separate PR?

server/webapp/WEB-INF/rails/app/helpers/agents_helper.rb
akshaydewan: Yeah, there are many usages of the `remote` and `local` terminology in the code. But I also find it confusing. So I've changed the name as you suggested.

akshaydewan:
@maheshp Yeah, I can make the agents page change in a separate PR.

akshaydewan:
Updated tooltip design:
screenshot 2019-02-11 at 15 46 28

PR #5800, opened 2019-02-01
Slow down when listener starts getting JMSException

arvindsv:
JMSException is unexpected and can happen when the consumer is somehow closed. This leads to the while loop in the [run method](https://github.com/gocd/gocd/blob/ba495470c154cf46aadf3ac45c67544735391c61/server/src/main/java/com/thoughtworks/go/server/messaging/activemq/JMSMessageListenerAdapter.java#L51-L57) running a lot, very quickly, filling up logs and overwhelming everything till the server is restarted.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/messaging/activemq/ActiveMqMessagingService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/messaging/activemq/ActiveMqMessagingService.java
@@ -1,18 +1,18 @@
-/*************************GO-LICENSE-START*********************************
- * Copyright 2014 ThoughtWorks, Inc.
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *************************GO-LICENSE-END***********************************/
+ */
 
 package com.thoughtworks.go.server.messaging.activemq;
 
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/thoughtworks/go/server/messaging/activemq/ActiveMqMessagingService.java
```diff
@@ -51,11 +51,12 @@
     private ActiveMQConnection connection;
     public ActiveMQConnectionFactory factory;
     private BrokerService broker;
+    private final SystemEnvironment systemEnvironment;
 
     @Autowired
     public ActiveMqMessagingService(DaemonThreadStatsCollector daemonThreadStatsCollector) throws Exception {
         this.daemonThreadStatsCollector = daemonThreadStatsCollector;
-        SystemEnvironment systemEnvironment = new SystemEnvironment();
+        systemEnvironment = new SystemEnvironment();
```
arvindsv: Don't know why it uses `new` in an autowired constructor.

server/src/main/java/com/thoughtworks/go/server/messaging/activemq/JMSMessageListenerAdapter.java
```diff
@@ -33,13 +34,14 @@
...
 
-    private JMSMessageListenerAdapter(MessageConsumer consumer, GoMessageListener listener, DaemonThreadStatsCollector daemonThreadStatsCollector) {
+    private JMSMessageListenerAdapter(MessageConsumer consumer, GoMessageListener listener, DaemonThreadStatsCollector daemonThreadStatsCollector, SystemEnvironment systemEnvironment) {
         this.consumer = consumer;
         this.listener = listener;
         this.daemonThreadStatsCollector = daemonThreadStatsCollector;
-
+        this.systemEnvironment = systemEnvironment;
 
         thread = new Thread(this);
```
ketan: Since we're just spawning a background thread. I wonder if we need to handle situations where the thread dies (I believe I've seen this situation a few times, thread going OOM)

arvindsv: @ketan Yes. Would be nice. But, I'll leave that to one of you. It probably needs to be managed in one place, and across other important threads too.

arvindsv: Also, trying to handle OOMs is not easy. It's possible that further attempts at bringing up the thread will again cause OOMs. @maheshp suggested adding a server health message to recommend restarting GoCD.

arvindsv: @maheshp I've added the server health message. I've verified that there's only one shown even if multiple queues have died.



I can change the message, if needed. I don't think most users will see it. So, it should be ok.

arvindsv:
It's all green.

Issue #5614, opened 2018-12-24
Show a warning if installed plugin is not supported by the current GoCD version


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Each GoCD plugin defines, the minimum required GoCD version. There is no visual indication, that the installed plugin is not supported by the current GoCD version

###### Basic environment details

* Go Version: `18.12.0`

##### Steps to Reproduce

1. Start GoCD server
2. Install any incompatible version plugin

##### Expected Results
GoCD should show a visual indication (warning or gray out) of the installed plugin is not supported by the current version.

##### Actual Results
No visual distinction between a supported plugin and unsupported plugin

##### Possible Fix
Gray out the plugin box.

rajiesh:
Verified on 19.2.0 (8610-4713200bc880ce17014d826e20a30ed032e3ac26)

PR #5801, opened 2019-02-04
[#5614] If plugin is incompatible with current gocd version, added error message stating the same.

vrushaliwaykole:
#### Required GoCD version by plugin is higher then installed GoCD version

screen shot 2019-02-04 at 4 28 52 pm

#### Plugin uses unknow extension
screen shot 2019-02-08 at 4 35 31 pm

#### Extension versions are not supported by GoCD
screen shot 2019-02-08 at 5 12 06 pm

---

screen shot 2019-02-08 at 5 14 04 pm


GaneshSPatil:
@vrushaliwaykole -- Not sure if this check is good enough.

Currently, we are checking if the `minimum required version of GoCD` specified in plugin description is equal to or greater than the current version of GoCD.

Though, For any GoCD plugin, the compatibility with the server version lies between a range and not the minimum server version. The plugin is compatible with the server until the versions of GoCD supports that plugin extension.

Let's say, There is a plugin `Foo-plugin` with version `v1.0.0` which uses `Credential extension v1` as part of GoCD `v19.1.0`. Let say, as part GoCD version `v19.5.0` we introduce `Credential extension v2` and remove `v1`. So, the `Foo-plugin v1.0.0`  will not work with GoCD version `>= v19.5.0`.
But with our logic, `Foo-plugin` is compatible with any version of GoCD above `v19.1.0`.

\cc: @bdpiparva @ketan

vrushaliwaykole:
As discussed with @GaneshSPatil @ketan, We are going to merge current implementation. However, checking compatibility remains unresolved. Created the issue for the same: #5804

bdpiparva:
## Fixes following issues

- Mark plugin invalid when target version specified in plugin.xml is incompatible with current GoCD version
- Mark plugin invalid when required extensions versions are incompatible with GoCD supported extensions versions

	e.g GoCD supports (Authorization: 1.0, 2.0, Elastic: 3.0, 4.0.. so on) 
	
	- Plugin A requiring (Authorization: 2.0 and Elastic 3.0) is valid but
	- Plugin B requiring (Authorization: 3.0) is invalid.

- `PluginChangeListener` won't be called for the invalid plugins

## Pending issue

- If the plugin is implemented with extension name which is not present in installed GoCD server then the plugin is still considered as a valid plugin. This is little complicated as getting required extensions by the plugin is tricky to get.

Issue #5730, opened 2019-01-18
Pipeline export should list only plugins with export capabilities


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Currently on pipelines page `/go/admin/pipelines` the `Export using` lists all the pipeline as code plugins which are loaded by the server, which will include the plugins which do not yet support pipeline exports. It should be changed to list only plugins which support pipeline export

###### Basic environment details

* Go Version: `19.1.0 (8417-bc5413918eb9c792352710a4b6b79c64cea4f439)`

##### Steps to Reproduce

1. This can be observed on build.gocd.org. Groovy config repo plugin do not support pipeline export still its listed under `Export using` list

GaneshSPatil:
As of now, we've fixed config-repo groovy plugin as well to support export capability. Which makes all the publicly available config repo plugin supports export functionality.

If there is any other config-repo plugin available which does not support export capability, It will show up in the `export using` plugin list dropdown. But, clicking on it will fail the export call with an error banner on the page saying `Plugin "cd.go.contrib.plugins.configrepo.fancy-plugin" does not support pipeline config export.`

Fix for this issue requires checking the config-repo plugin capabilities (from plugin infos) to verify the plugin supports export functionality. And the plugin-infos should cache the plugin capabilities (fetched only once on plugin load).

---

Moving this issue to next release as its not affecting wider audiance.

\cc: @rajiesh @ketan @maheshp @marques-work

maheshp:
Reopening for testing.

rajiesh:
Verified on 19.3.0 (8959-259d1f79e361875995072e5c1fcc9cbda7ccc600)

PR #5806, opened 2019-02-06
Fix export list

kierarad:
This is a fix for issue #5730 

In order to do this I've done 2 things: 
- I've added the capabilities to the plugin info for config repos so we can check if the plugin supports export
- I've filtered the dropdown on the pipeline config page to only show the plugins that support pipeline export


plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/configrepo/ConfigRepoPluginInfoBuilder.java
```diff
@@ -35,11 +36,14 @@ public ConfigRepoPluginInfoBuilder(ConfigRepoExtension extension) {
...
+        return new ConfigRepoPluginInfo(descriptor, image(descriptor.id()), pluggableInstanceSettings, capabilities(descriptor.id()));
     }
 
     private com.thoughtworks.go.plugin.domain.common.Image image(String pluginId) {
         return extension.getIcon(pluginId);
     }
 
+    private Capabilities capabilities(String pluginId) {
+        return extension.getCapabilities(pluginId);
+    }
```
ketan: 👍 

The intention behind this is to cache the calls made to the plugin, that always return the same data. Capabilities, get view, get icon, etc. I noticed a few days ago that `GoConfigPluginService` directly calls `getCapabilities` on the extension, instead of looking up the metadata store. https://github.com/gocd/gocd/blob/84ae8699545ee85320fbaeb0702acc4009d74032/server/src/main/java/com/thoughtworks/go/config/GoConfigPluginService.java#L62-L64

Would you mind changing that reference to use the metadata store instead (in a separate commit)?

server/dev/config/cipher.aes
```diff
@@ -0,0 +1 @@
+e83d7343445b54ecdf4c5ca2afc31f47
```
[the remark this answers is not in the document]
kierarad: @ibnc yes, good catch. I quickly pushed this up yesterday so that @ketan could see what I was referring to in regards to my versioning question. I definitely didn't mean to include this file. I should've noted that this PR is a WIP

plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/configrepo/ConfigRepoPluginInfoBuilder.java
kierarad: Of course. Good catch

kierarad:
@ketan I think this is ready for review (again). When I bumped the version I went ahead and removed v4 since it sounded like we are the only ones who use the API and we don't maintain any other previous versions of the plugin infos API. Let me know if you or @maheshp think this is a bad idea and I will add v4 back.

maheshp:
@kierarad I took a look at the PR, the changes around adding the capabilities to `ConfigRepoPluginInfo` seems fine and changes to use the cached capabilities in `GoConfigPluginService.java` is right.

Bumping up the PluginInfo needs to change. Sorry, I should have mentioned this earlier that we are moving away from rails, hence whenever we write a new API or bump up the version of an existing API we write it in Java/Spark. I am afraid we will have to do the same for PluginInfo as well.

Also, whenever we introduce a  new version of an existing API we deprecate and retain the current version for around 3 months and later remove it. Your changes actually removes the v4 of the endpoint and introduces v5 that would break for existing users.

For now I would suggest we split this PR into 2 parts, have one PR with changes to capabilities and get it merged and open a new PR for the PluginInfo V5 in Spark.

kierarad:
Thanks @maheshp! 

> Bumping up the PluginInfo needs to change. Sorry, I should have mentioned this earlier that we are moving away from rails, hence whenever we write a new API or bump up the version of an existing API we write it in Java/Spark. I am afraid we will have to do the same for PluginInfo as well.
>
Don't be sorry. This was my mistake. I should've known. That makes perfect sense and it's silly that I didn't consider that. 
 
> Also, whenever we introduce a new version of an existing API we deprecate and retain the current version for around 3 months and later remove it. Your changes actually removes the v4 of the endpoint and introduces v5 that would break for existing users.
> 
Ok this makes sense. I was unsure, so I should of asked. Again this is on me. Lesson learned, I was questioning my approach (obviously for good reason) so I should've asked instead of just doing. 

> For now I would suggest we split this PR into 2 parts, have one PR with changes to capabilities and get it merged and open a new PR for the PluginInfo V5 in Spark.
>
Will do. For now I've just added the V4 back to rails. I'll create a separate PR for the rewrite of the V5 in Spark.

maheshp:
Thanks @kierarad!

kierarad:
@maheshp is this one ok to merge?

arvindsv:
@kierarad Since you're moving this to Spark, why don't we just close this and just have this merged once it is moved to Spark? That way, the testing / reviewing effort of it can be reduced.

Since there are only 3 config repo plugins currently, and they all support the export feature, I think it's ok that it shows them all. Once the API in Spark is done, it'll be doing the right thing.

arvindsv:
Also: That way, you won't have to get rid of this, when moving to Spark.

kierarad:
@arvindsv There is still quite a bit of effort that will need to go into moving the PluginInfos API to Spark. Given the shift in priorities to improving the trial experience I don't foresee it will be completed for quite some time. Should go ahead and close this still?

arvindsv:
@kierarad Yes, I'd still do that. When someone gets a chance, they can pick it up again.

@maheshp We'll need to add this to the list and decide when to pick it back up.

ketan:
Can bring this back if needed.

Issue #5321, opened 2018-10-25
Epic: Access Token based authentication support for GoCD APIs


vishaldevgire:
##### Issue Type
- Feature enhancement

##### Summary
Currently GoCD APIs support basic authentication, this issue is about adding support for access token-based authentication to GoCD APIs. 

### For 19.2.0
- [x] Add API endpoints to generate, get, list and delete token (#5354) (PR #5797)
- [x] Modify existing API request flow to allow token-based authentication (#5357) (PR #5764). 
- [x] Add a separate authorization call in authorization extension to fetch roles for a given username.(PR: #5849, #5845)
- [x] Allow specifying the reason for revoking access token. (Issue: #5821)
- [x] Disallow users to use Access Token APIs in case plugin does not have capabilities. (Issue: #5822)
- [x] Add a cache store for access token related plugin calls (Issue: #5823)
- [x] Deletion of User should revoke access token. (Issue: #5818) (PR #5848)
- [x] Update last used time for access token (Issue: #5824) (PR #5870)
- [x] Add a toggle for the access token SPA(#5875)
- [x] Document new authentication method and update API documentation 
   - [x] API docs(https://github.com/gocd/api.go.cd/pull/289)
   - [x] Authorization extension v2(https://github.com/gocd/plugin-api.go.cd/pull/98)

#### Bugs
 - [x] Revoke, should return 422 if token is already revoked #5859

### Can be done post-release
- [ ] Affected plugins which require an upgrade for access token feature (Issue: #5825)

### Next release
- [ ] Add SPA (page in UI) to manage, generate, list, delete token(#5803)

### Won't Fix
- [ ] Deletion of auth config should revoke access token (Issue: #5819)

arvindsv:
If you're thinking about the API, I'd consider the POST / PUT taking a request body in which it has something like:

```javascript
"permissions": "all"
```

or

```javascript
"permissions": [
    {
        "/go/api/*": true
    }
]
```

... something like that. To make it easier to get people used to possible changes around granular auth tokens. However, since we will have versioning, maybe it's ok not to. Thinking out loud here.

arvindsv:
Probably best to not bother about it now, and let versioning help us out later?

maheshp:
@rahulpargaonkar to test this feature.

arvindsv:
@maheshp, @rajiesh and I had a discussion about this and here's what we came up with:

### Assumptions:

Just to be clear. Feel free to tell us our assumptions are wrong.

[...]

PR #5809, opened 2019-02-06
Validate User belonging to the access token exists before proceeding with AccessTokenFilter (#5321)

GaneshSPatil:
AccessTokenBasedPluginAuthenticationProvider flow:
1. Verify provided access token is valid.
2. Verify access token is not revoked.
3. Check is the current user still exists with authorization plugins,
   against the belonging auth config id.
4. If user exists and plugin has fetching roles for user capability, fetch roles.
5. Mark user authenticated and proceed with the requested operation.

**NOTE:**
This change will require an upgrade in the authorization plugins to support
'doesUserExists' plugin-extension call. Without this implementations, GoCD
users will not be able to use access-token feature.

### DIFF
--- a/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationExtension.java
+++ b/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationExtension.java
@@ -224,6 +224,25 @@ public class AuthorizationExtension extends AbstractExtension {
         });
     }
 
+    public boolean isValidUser(String pluginId, String username, SecurityAuthConfig authConfig) {
+        try {
+            return pluginRequestHelper.submitRequest(pluginId, IS_VALID_USER, new DefaultPluginInteractionCallback<Boolean>() {
+                @Override
+                public String requestBody(String resolvedExtensionVersion) {
+                    return getMessageConverter(resolvedExtensionVersion).isValidUserRequestBody(username, authConfig);
+                }
+
+                @Override
+                public Boolean onSuccess(String responseBody, Map<String, String> responseHeaders, String resolvedExtensionVersion) {
+                    return true;
+                }
+            });
+        } catch (Exception e) { //any error happened while verifying the user existence will lead to assume user does not exists.
+            return false;
+        }
+
+    }
+
     public AuthorizationMessageConverter getMessageConverter(String version) {
         return messageHandlerMap.get(version);
     }
--- a/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverter.java
+++ b/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationMessageConverter.java
@@ -70,5 +70,5 @@ public interface AuthorizationMessageConverter {
 
     String authorizationServerUrlRequestBody(String pluginId, List<SecurityAuthConfig> authConfigs, String siteUrl);
 
-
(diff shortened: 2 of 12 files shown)


GaneshSPatil:
Password file changes to verify this behaviour: https://github.com/gocd/gocd-filebased-authentication-plugin/pull/36

plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationExtension.java
```diff
@@ -224,6 +224,25 @@ public AuthenticationResponse onSuccess(String responseBody, Map<String, String>
         });
     }
 
+    public boolean doesUserExists(String pluginId, String username, SecurityAuthConfig authConfig) {
```
bdpiparva: Maybe `isValidUser` instead of `doesUserExists`. 
- The user account can be disabled by admin
- The user account can be temporarily locked

bdpiparva: This call should not be made for the plugin using extension version v1.

bdpiparva:
@GaneshSPatil - Since we have `getUserRoles` call now do you think we should split the `authenticateUser` call into two separate calls?

Currently `authenticateUser` is doing authentication and authorization both. With `getUserRoles` it can be confusing for the plugin developer why to return roles information twice.  I am thinking we should do authentication only as part of `authenticateUser` call and return the user. To get roles information we should use the `getUserRoles` call.

plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/authorization/AuthorizationExtension.java
GaneshSPatil: @bdpiparva -- we are not making this call for `v1` extension. It should throw an `UnsupportedOperationException` if we try to make a call for `v1` extension.

I will rename the plugin-extension call to `is-valid-user` and change the relevant code too?

GaneshSPatil:
> @GaneshSPatil - Since we have `getUserRoles` call now do you think we should split the `authenticateUser` call into two separate calls?
> 
> Currently `authenticateUser` is doing authentication and authorization both. With `getUserRoles` it can be confusing for the plugin developer why to return roles information twice. I am thinking we should do authentication only as part of `authenticateUser` call and return the user. To get roles information we should use the `getUserRoles` call.

@bdpiparva -- If a user is authenticated, I would expect we to get a user with the permission (roles in our case) as we need the user and the roles to operate on.

I do agree that we are adding a new endpoint to explicitly fetch the roles. But the use-cases for both the endpoints are different.  The `authenticate` endpoint will be used for web-based login, whereas, the `get-roles` endpoint will be used for fetching the permissions for a valid user.

Also, was wondering if there are any advantages in breaking the existing workflow and splitting it up into two? Maybe we will separate out endpoint responsibilities (authenticate just doing authentication, get-roles just fetching roles) at an overhead of changing the contract.

**Note:** If we plan to do this, I will send a separate PR as the current PR does not need to address this change.

GaneshSPatil:
PR build passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/4271

GaneshSPatil:
@rahulpargaonkar @rajiesh @adityasood -- with this change the access token flow will be broken until we start bundling the newer versions of plugins (password file).

To test the changed behavior. We can use plugins build from the PR branches. 
For password-file-based plugin use https://github.com/gocd/gocd-filebased-authentication-plugin/pull/36
For ldap-authorization plugin use https://github.com/gocd-private/ldap-authorization-plugin/pull/35

\cc: @bdpiparva

PR #5811, opened 2019-02-06
Allow plugin settings API to accept booleans

arvindsv:
If plugin settings UI has a checkbox with code such as:

```
<input type="checkbox" ng-model="abc1" id="abc1" ng-true-value="true" ng-false-value="false" />
```

then, upon save, it fails with an exception in the GoCD server logs:

```
ERROR [qtp1275035040-38] Rails:-2 -
ERROR [qtp1275035040-38] Rails:-2 - TypeError (cannot convert instance of class org.jruby.RubyBoolean$False to class java.lang.String):
ERROR [qtp1275035040-38] Rails:-2 -
ERROR [qtp1275035040-38] Rails:-2 - app/presenters/api_v1/config/plugin_configuration_property_representer.rb:47:in `from_hash'
app/presenters/api_v1/base_representer.rb:105:in `from_hash'
app/controllers/api_v1/admin/plugin_settings_controller.rb:31:in `create'
```

This PR allows it to accept: `[ { "key": "some.key.1", "value": true } ]` in the JSON and convert it to `[ { "key": "some.key.1", "value": "true" } ]` and use it.

To test this, you can use the two commands below. The second one will fail, without this change.

```
$ curl -v 'http://localhost:8153/go/api/admin/plugin_settings' \
    -H 'Accept: application/vnd.go.cd.v1+json' \
    -X POST \
    -d '{ "plugin_id": "json.config.plugin", "configuration": [ { "key": "some.key.1", "value": "true" } ] }'
...
< HTTP/1.1 200 OK
< Content-Type: application/vnd.go.cd.v1+json; charset=utf-8

{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/admin/plugin_settings/json.config.plugin"
    },
    "doc": {
      "href": "https://api.gocd.org/19.2.0/#plugin-settings"
    },
    "find": {
      "href": "http://localhost:8153/go/api/admin/plugin_settings/:plugin_id"
    }
  },
  "plugin_id": "json.config.plugin",
  "configuration": [
    {
      "key": "some.key.1",
      "value": "true"
    }
  ]
}

$ curl -v 'http://localhost:8153/go/api/admin/plugin_settings' \
    -H 'Accept: application/vnd.go.cd.v1+json' \
    -X POST \
    -d '{ "plugin_id": "json.config.plugin", "configuration": [ { "key": "some.key.1", "value": true } ] }'

< HTTP/1.1 500 Server Error
< Content-Type: text/html;charset=utf-8

<!DOCTYPE html>
<html>
<head>
  <title>We're sorry, but something went wrong (500)</title>

...

<body class="rails-default-error-page">
  <!-- This file lives in public/500.html -->
  <div class="dialog">
    <div>
      <h1>We're sorry, but something went wrong.</h1>
    </div>
    <p>If you are the application owner check the logs for more information.</p>
  </div>
</body>
</html>
```

To use this from a plugin, a recommended pattern in the UI should be:

```
<input type="checkbox" ng-model="abc1" id="abc1" ng-init="abc1 = abc1 || false" ng-true-value="true" ng-false-value="false" />
```

### DIFF
--- a/server/webapp/WEB-INF/rails/app/presenters/api_v1/config/plugin_configuration_property_representer.rb
+++ b/server/webapp/WEB-INF/rails/app/presenters/api_v1/config/plugin_configuration_property_representer.rb
@@ -1,5 +1,5 @@
 ##########################################################################
-# Copyright 2015 ThoughtWorks, Inc.
+# Copyright 2019 ThoughtWorks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -44,7 +44,7 @@ module ApiV1
 
       def from_hash(data, options={})
         data = data.with_indifferent_access
-        configuration_property.deserialize(data[:key], data[:value], data[:encrypted_value])
+        configuration_property.deserialize(data[:key], data[:value].to_s, data[:encrypted_value])
         configuration_property
       end
     end
--- a/server/webapp/WEB-INF/rails/spec/presenters/api_v1/config/plugin_configuration_property_representer_spec.rb
+++ b/server/webapp/WEB-INF/rails/spec/presenters/api_v1/config/plugin_configuration_property_representer_spec.rb
@@ -1,0 +1,101 @@
+##########################################################################
+# Copyright 2019 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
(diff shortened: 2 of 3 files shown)


server/webapp/WEB-INF/rails/spec/controllers/api_v1/admin/plugin_settings_controller_spec.rb
```diff
@@ -190,19 +190,35 @@
           ])))
       end
 
-      it 'should deserialize plugin settings from given object' do
+      it 'should deserialize plugin settings from given object with sane defaults for boolean and null' do
```
varshavaradarajan: Consider moving this up to plugin_configuration_property_representer_spec as plugin_configuration_property_representer.rb is shared among representers used for scm, package-repo, auth-config and plugin settings API?

arvindsv: Yes, makes sense. I'll do that.

arvindsv: How about this?

varshavaradarajan: LGTM

Issue #5814, opened 2019-02-07
Pipeline export should consider exec tasks with args as attribute


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Currently we still allow a task to be created as below by directly editing the cruise-config.xml - 
```
<tasks>
    <exec command="ls" args="-l" />
</tasks>
```
When we export pipeline with this task the resulting JSON is -

```
"tasks": [
            {
              "command": "ls",
              "timeout": -1.0,
              "arguments": [],
              "run_if": "passed",
              "type": "exec"
            }
          ]
```

Arguments is empty as its defined in the attribute of exec task. Similar result on YAML config too.

Currently from UI we cannot create a config with args in attribute, but from cruise-config.xml we can. This issue was encountered on one of our internal GoCD server instance which is running for years. So this can be faced by old installation of GoCD server when they come to use the export feature. Need to find a way to handle it.

###### Basic environment details

* Go Version: `19.1.0-8469`
* JSON config repo plugin: `0.3.6`
* YAML Config repo plugin: `0.8.6`

ibnc:
Looking into this

ibnc:
@rajiesh I'm closing this issue. If you notice any issues please reopen it

PR #5815, opened 2019-02-07
Exec args edgecase fix

ibnc:
See #5814

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/ExecTask.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/ExecTask.java
@@ -298,6 +298,18 @@ public class ExecTask extends AbstractTask implements CommandTask {
                 '}';
     }
 
+    public Arguments getArguments() {
+       if (args.isEmpty()) {
+           return argList;
+       }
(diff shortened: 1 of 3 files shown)


config/config-api/src/main/java/com/thoughtworks/go/config/ExecTask.java
```diff
@@ -298,6 +298,18 @@ public String toString() {
                 '}';
     }
 
+    public Arguments getArguments() {
+       if (args.isEmpty()) {
+           return argList;
+       }
+       Arguments arguments = new Arguments();
+       for (String arg : args.split(" ")) {
```
arvindsv: @ibnc [This is](https://github.com/gocd/gocd/blob/68f8bafb4fea5d132a9363c3fe39b6eea1628466/common/src/main/java/com/thoughtworks/go/domain/builder/CommandBuilder.java#L53) what GoCD does with arguments provided as "args".

Splitting by space will cause problems when using a task such as:

```
<exec command="ls" args="'file with space'" />
```

arvindsv:
Looks good.

Issue #2, opened 2014-02-13
Make Go more accessible to color-blind users


arvindsv:
Here's feedback from a user of Go, who is color-blind:

> I am color blind and I find the Go UI very hard to use. It might be pretty obvious to use green and red as indicators on the Go UI, but at times I find it a little hard to differentiate between the two and this makes the Go UI extremely unusable to me. I can almost never tell if a stage has failed unless I click into the stage. If I stare hard enough, I can figure out, but it's not as obvious to me as it is to you. 
> 
> The simple fix for this is to use labels instead of color for build statuses. Though I understand that color blindness is a relatively uncommon condition (not very uncommon though, around 10% of all men are red-green color blind), if you do have this condition, you can imagine the GO UI as a bunch of grey lines with no indication of where the failures are.

Needs discussion. Should this be an option? Should we change something more basic?

grahamc:
A small amount of CSS might fix this:

```css
.Passed:after {content: "✓";}

.Failed:after {content: "✗";}
  
.stage_bar {
  color: white; padding-left: 5px;
}
```

![pipelines_-_go](https://cloud.githubusercontent.com/assets/76716/9287205/b2573004-42cf-11e5-906d-1985d19d5f6f.jpg)

protanopia:
![simulate](https://cloud.githubusercontent.com/assets/76716/9287206/c2e44434-42cf-11e5-8c29-d26071fe2c1e.jpg)

deuteranopia:
![simulate-2](https://cloud.githubusercontent.com/assets/76716/9287220/e8f93440-42cf-11e5-823f-2958a530e629.jpg)

tritanopia:
![simulate-4](https://cloud.githubusercontent.com/assets/76716/9287221/fb47d610-42cf-11e5-9e45-252771eef094.jpg)

ibnc:
Can this issue be closed now since #3233 was merged?

arvindsv:
I think so. Closing this.

Issue #5492, opened 2018-12-04
Changing plugin ID of config repo seems to have no effect


ketan:
##### Issue Type

- Bug Report

##### Steps to reproduce

 1. Create a config repo plugin.
 2. wait for few seconds & go to pipeline dashboard to check whether the pipeline is displayed.
 3. If pipeline displayed on dashboard, then go to config repo page again 
 5. Change the plugin type for the config repo (choose any other plugin), so that the plugin is unable to now parse the config repo files in the repo
 6. when user go back to dashboard page then the same pipeline is still displayed even though the wrong plugin data is provided.

akshaydewan:
Just documenting what we've found out so far -- 
After an MDU happens, `ConfigMaterialUpdater:80` runs this code:

```
if (lastParseRevision == null) {
    //never parsed
    updateConfigurationFromCheckout(folder, revision, material);
} else if (latestModification.findRevisionFor(material.config())
        .hasChangedSince(lastParseRevision)) {
    // revision has changed. the config files might have been updated
    updateConfigurationFromCheckout(folder, revision, material);
} else {
    // revision is the same as last time, no need to parse again
}
```

Since the revision is the same as last time, it doesn't parse the config repo, even if the plugin ID has changed.

ankitsri11:
Fixed as part of this PR: https://github.com/gocd/gocd/pull/5816

Verified on GoCD Version: 19.2.0 (8559-38e68d0ca9dfce367dd1d9a380e824595cd87d34)

Issue #5792, opened 2019-01-30
Improvements to Config Repo Implementations


adityasood:
##### Issue Type

- Feature enhancement

##### Summary

Following are some scenarios where we can improve the user experience

1. Today if there is a merge conflict occurs due to 
   * Resources not available in GoCD - e.g environments, elastic profiles, this leads to an error. Once these are created in the cruise-config.xml, the user would still have to do a dummy commit on the config repo, for us to parse it. While we might not be able to do a validation while each resource is created, it would make sense to provide an option to force a re-validation of the config repo for a known commit.

   * Pipeline Dependency not fulfilled

2. Change in plugin configuration: For e.g, if the user adds a file pattern on the plugin settings and this leads to no pipelines. If the user changes this configuration, they are still required to make a dummy commit to the repo. This causes a not so good getting started to experience and restricts users to try out configuration options.

* NOTE will add to this issue and keep it as a placeholder for improvements. Related epic https://github.com/gocd/gocd/issues/5175

ankitsri11:
@adityasood I tried adding this config-repo https://github.com/tomzo/gocd-json-config-example without adding `up42` pipeline. I got below merge error:

![screenshot 2019-02-05 at 13 32 04](https://user-images.githubusercontent.com/6838313/52305788-d76bdd80-294a-11e9-8c16-750f6511f3c0.png)

But once I added `up42` pipeline, I could see all the config-repo pipelines on the dashboard and started to build. However, the merge error didn't go away. We might need to look into that.

Another thing, I noticed as per the config-repo environment file (https://github.com/tomzo/gocd-json-config-example/blob/master/dev.goenvironment.json), `up42` pipeline needs to belong to `dev ` environment but adding `up42` pipeline after adding config-repo, I get below error.

![config-repo](https://user-images.githubusercontent.com/6838313/52306062-9de7a200-294b-11e9-8545-afcd1eb33e14.gif)

Error in the go-server.log:

```
2019-02-05 13:40:17,138 ERROR [qtp632249781-26] Rails:-2 -
2019-02-05 13:40:17,139 ERROR [qtp632249781-26] Rails:-2 - Java::JavaLang::NullPointerException ():
2019-02-05 13:40:17,139 ERROR [qtp632249781-26] Rails:-2 -
[...]

PR #5816, opened 2019-02-08
Parse and merge ConfigRepos on change to ConfigRepoConfig #5792

maheshp:
Fixes issue #2 in https://github.com/gocd/gocd/issues/5792

* ConfigRepos are not parsed and merged if there is no change in the
  ConfigRepo material since last parse. A change to the ConfigRepoConfig
  should force a parse and merge even if the material has not changed.


kierarad:
this also fixes #5492

Issue #5705, opened 2019-01-14
High cpu


greuelpirat:
##### Issue Type

- Bug Report

##### Summary

Viewing Tests-Tab results in High CPU Usage with low respondence in Chrome.

###### Basic environment details

Go Server Version: | 18.12.0 (8222-e6778f1cc84bf91e323d337c876046167da36985)
-- | --
JVM version: | 11.0.1
OS Information: | Windows Server 2016 10.0
Usable space in artifacts repository: | 948 GB
Database schema version: | 1808002
Pipelines Count: | 14

Chrome 71.0.3578.98 (Windows)

##### Steps to Reproduce

1.  Open Tests-Tab with failed Unittests

##### Expected Results
Viewing Unittests. CPU free

##### Actual Results
Viewing Unittests. CPU under full load

arvindsv:
@greuelpirat When the CPU is under load, can you save the output of: http://your-server:8153/go/api/support ... every 10 or 20 seconds, 3 or 4 times and zip it up and provide it in a [gist](https://gist.github.com)?

That might help see what the server is doing. Without that, it's hard to understand what is happening - since it obviously doesn't happen to everyone (for instance, it's working fine with me).

Thank you!

greuelpirat:
The CPU-Load happens not on the server. It is in the browser while viewing the Tests-Tab.

Here is a screenshot with Chrome.

![image](https://user-images.githubusercontent.com/16866222/51120123-b402b680-1814-11e9-8a23-5f3b3de859c4.png)

Can I send you the corresponding test-artifact?

greuelpirat:
Here are the files. Hope you have everything.

[data.zip](https://github.com/gocd/gocd/files/2755789/data.zip)

Thank you for your help

ketan:
@greuelpirat — Are you able to reproduce this in incognito mode, with all extensions disabled?

greuelpirat:
The cpu load also occurs in incognito mode (with all extensions disabled)

I have also tried the docker-image, but there seems the test-pipeline to be missing:

    2019-01-28 08:55:25,552 ERROR [qtp2028371466-16] JobController:176 - Job detail page error:
    java.lang.RuntimeException: Job Test/2/defaultStage/1/defaultJob not found
        at com.thoughtworks.go.util.ExceptionUtils.bomb(ExceptionUtils.java:30)
        at com.thoughtworks.go.server.controller.JobController.jobDetail(JobController.java:126)

[...]

PR #5820, opened 2019-02-08
Fix for High CPU usage on Console Logs/Tests page

akshaydewan:
Issue #5705

I was able to replicate the issue and concur with @arvindsv's findings - the `fixElement` function being called via `window.setInterval` was the culprit.

I tried increasing the interval value, and the CPU usage went down to 50% instead of 100%. However, it still seemed silly that the page should use that much CPU while doing nothing - no scrolling, no resizing, just sitting idle and eating CPU.

So I tried removing the `setInterval` function and the pages continued working normally. I couldn't find anything broken. The scroll and resize event handlers defined below it continue to pin the header and footer as required. I *think* the `setInterval` function should be safe to remove, but it would be good if someone else tests it too. (After all, it was originally added for some reason, right?)

### DIFF
--- a/server/webapp/WEB-INF/rails/app/assets/javascripts/lib/jquery-pinOnScroll.js
+++ b/server/webapp/WEB-INF/rails/app/assets/javascripts/lib/jquery-pinOnScroll.js
@@ -87,12 +87,6 @@
       var throttledFixElement   = _.throttle(fixElement, 100);
       var throttledUnFixElement = _.throttle(unFixElement, 100);
 
-      window.setInterval(function () {
-        maybeCallUsingRequestAnimationFrame(function () {
-          throttledFixElement(elem, opts);
-        });
-      }, 100);
-
       $(window).on('scroll.pinOnScroll', function () {
         maybeCallUsingRequestAnimationFrame(function () {
           throttledFixElement(elem, opts);
@@ -104,7 +98,7 @@
           throttledUnFixElement(elem, opts);
           throttledFixElement(elem, opts);
         });
-      })
+      });
 
       $(window).on('resetPinOnScroll', function (e, eParams) {
             if (eParams.calcRequiredScroll) {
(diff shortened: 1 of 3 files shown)


arvindsv:
Looks alright to me.

server/webapp/WEB-INF/rails/app/assets/javascripts/lib/jquery-pinOnScroll.js
```diff
@@ -87,12 +87,6 @@
       var throttledFixElement   = _.throttle(fixElement, 100);
       var throttledUnFixElement = _.throttle(unFixElement, 100);
 
-      window.setInterval(function () {
-        maybeCallUsingRequestAnimationFrame(function () {
```
ketan: I believe I'd added this because the pinning would otherwise not work when new content was added to the page (by way of fetching new console logs via ajax).

If it's possible to pin the button by removing the code, seems much better to me.

arvindsv:
There's a weird bug around here. If you have that button ("Back to top") and you switch tabs, the button still shows, even if you don't need it in that tab.

server/webapp/WEB-INF/rails/app/assets/javascripts/lib/jquery-pinOnScroll.js
akshaydewan: Ah! I see. I'll test out this scenario and see if we can find a way around it.

akshaydewan: Listening to some existing [events](https://github.com/gocd/gocd/pull/5820/commits/a02fb9ca81c9b3405af07894cb472e0cbcc41c4d) seems to fix this without the need for polling.

akshaydewan:
> There's a weird bug around here. If you have that button ("Back to top") and you switch tabs, the button still shows, even if you don't need it in that tab.

I noticed it too. It seems to be an existing bug which we can fix separately.

PR #5828, opened 2019-02-11
configrepo - allow empty string secure variables

tomzo:
Original issue - https://github.com/tomzo/gocd-yaml-config-plugin/issues/84

This change allows config-repo users to set secure environment variables to empty strings.
Technically the following definition is not correct because empty string `""` should have been encrypted with cipher text and look something like `AES: 3a23gfh`
```
    secure_variables:
      MYSECRET: ""
```

There is one reasonable use case to allow above definitions - when user actually does not want to store secret in the config-repo, but rather provide it at runtime when triggering pipeline with overridden
 environment variables.

empty_secure_var

Above screenshot from following pipeline:
```yaml
  "yaml-example": # definition of yaml-example pipeline
    group: yaml-example
    label_template: "${mygit[:8]}"
    locking: off
    materials:
      mygit: # this is the name of material
        # keyword git says about type of material and url at once
        git: http://my.example.org/mygit.git
        branch: ci
      upstream:
        # type is optional here, material type is implied based on presence of pipeline and stage fields
        # type: dependency
        pipeline: yamlpipe1
        stage: build
    environment_variables:
      DUMMY: "Tigers"
    secure_variables:
      MYSECRET: ""
    stages: # list of stages in order
      - build: # name of stage
          clean_workspace: true
          jobs:
            csharp: # name of the job
              resources:
               - net45
              artifacts:
               - build:
                   source: bin/
                   destination: build
               - test:
                   source: tests/
                   destination: test-reports/
              tabs:
                report: test-reports/index.html
              tasks: # ordered list of tasks to execute in job csharp
               - fetch:
                   pipeline: yamlpipe1
                   stage: build
                   job: build
                   source: test-bin/
                   destination: bin/
               - exec: # indicates type of task
                   command: bash
                   arguments:
                    - -c
                    - "echo ${MYSECRET}"
```

## Implementation

When `ConfigConverter` sees that `encrypted value` is an empty string, it will be encrypted using current cipher and returned as secure variable to the config server.
Previously it would lead to situation when secure variable is treated as normal one. So no way to implement the use case of providing secrets at run time.

@ibnc can you take a look at this?

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
+++ b/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
@@ -136,8 +136,7 @@ public class ConfigConverter {
                 throw new RuntimeException("Encryption of empty secure variable failed", e);
             }
             return new EnvironmentVariableConfig(cipher, crEnvironmentVariable.getName(), encryptedValue);
-        }
-        else {
+        } else {
             String value = crEnvironmentVariable.getValue();
             if(StringUtils.isBlank(value))
                 value = "";
--- a/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
+++ b/server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
@@ -41,6 +41,7 @@ import com.thoughtworks.go.domain.scm.SCMs;
 import com.thoughtworks.go.plugin.configrepo.contract.*;
 import com.thoughtworks.go.plugin.configrepo.contract.material.*;
 import com.thoughtworks.go.plugin.configrepo.contract.tasks.*;
+import com.thoughtworks.go.security.CryptoException;
 import com.thoughtworks.go.security.GoCipher;
 import com.thoughtworks.go.util.command.CommandLine;
 import com.thoughtworks.go.util.command.HgUrlArgument;
@@ -124,8 +125,19 @@ public class ConfigConverter {
 
     public EnvironmentVariableConfig toEnvironmentVariableConfig(CREnvironmentVariable crEnvironmentVariable) {
         if (crEnvironmentVariable.hasEncryptedValue()) {
+            // encrypted value is not null or empty string
             return new EnvironmentVariableConfig(cipher, crEnvironmentVariable.getName(), crEnvironmentVariable.getEncryptedValue());
-        } else {
+        } else if(!crEnvironmentVariable.hasValue() && "".equals(crEnvironmentVariable.getEncryptedValue())) {
+            // encrypted value is an empty string - user wants an empty, but secure value, possibly to override at trigger-time
+            String encryptedValue = null;
+            try {
+                encryptedValue = cipher.encrypt("");
+            } catch (CryptoException e) {
+                throw new RuntimeException("Encryption of empty secure variable failed", e);
+            }
+            return new EnvironmentVariableConfig(cipher, crEnvironmentVariable.getName(), encryptedValue);
+        }
+        else {
             String value = crEnvironmentVariable.getValue();
             if(StringUtils.isBlank(value))
                 value = "";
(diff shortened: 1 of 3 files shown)


arvindsv:
@tomzo Makes sense to me.

/cc @ankitsri11

arvindsv:
I think @ibnc can take a look and merge it later today.

ketan:
Should we be explicit and set to `null` instead of a blank string?

arvindsv:
@ketan I'd treat both similarly and document one. In this way of using it, to me, it seems clear what the intention of the user is. They want to set it to an empty value (and not be bothered by an exception that the empty value should be encrypted).

tomzo:
@ketan 

> Should we be explicit and set to `null` instead of a blank string?

The problem is that structure returned by config-repo plugin is
```json
{
   "key" : "name",
   "value" : "normalstring",
   "encrypted_value" : "cipher text"
}
```
Then config converter decides if this is a normal or secure variable based on content of those 2 strings. 
But in this case we want both to empty `empty/null` and that can be ambiguous if it was supposed to be secure or not:
```
{
   "key" : "name",
   "value" : null,
   "encrypted_value" : null
}
```
So I made it this way:
If value is not `Blank` (null or empty) and `encrypted_value` is an empty string, then we have the corner case mentioned.

ketan:
Thanks for the clarification. Lets go with blank.

server/src/main/java/com/thoughtworks/go/config/ConfigConverter.java
```diff
@@ -124,8 +125,19 @@ public BasicEnvironmentConfig toEnvironmentConfig(CREnvironment crEnvironment) {
...
+        } else if(!crEnvironmentVariable.hasValue() && "".equals(crEnvironmentVariable.getEncryptedValue())) {
+            // encrypted value is an empty string - user wants an empty, but secure value, possibly to override at trigger-time
+            String encryptedValue = null;
+            try {
+                encryptedValue = cipher.encrypt("");
+            } catch (CryptoException e) {
+                throw new RuntimeException("Encryption of empty secure variable failed", e);
+            }
+            return new EnvironmentVariableConfig(cipher, crEnvironmentVariable.getName(), encryptedValue);
+        }
```
ibnc: Remove newline to keep syntax consistency 

should be 
```
} else {
```

server/src/test-fast/java/com/thoughtworks/go/config/ConfigConverterTest.java
```diff
@@ -162,6 +164,24 @@ public void shouldConvertEnvironmentVariableWhenSecure() {
...
+    public void shouldConvertEnvironmentVariableToSecureWhenEncryptedValueIsEmptyString() {
+        CREnvironmentVariable crEnvironmentVariable = new CREnvironmentVariable("key1", null, "");
+        EnvironmentVariableConfig result = configConverter.toEnvironmentVariableConfig(crEnvironmentVariable);
+        assertThat(result.isSecure(), is(true));
+        assertThat(result.getValue(), is(""));
+        assertThat(result.getName(), is("key1"));
+    }
+
+    @Test
+    public void shouldConvertEnvironmentVariableToInsecureWhenEncryptedValueIsNull() {
```
ibnc: Just a thought that def doesn't need to be done, but maybe this test can be merged with `#shouldConvertNullEnvironmentVariableWhenNotSecure` ?

They're basically doing the same thing, and it'd be one less test?

ibnc:
@tomzo This PR looks good to me. There's only a minor change that I'd like to see you change (see above). And to be safe we should probably run it through the build

tomzo:
@ibnc thanks. 
I removed the test and renamed the other related tests to be more specific about the behavior we want.

Issue #5823, opened 2019-02-11
Add a cache store for access token related plugin calls


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Currently, when an access token is used, the GoCD server makes a call to the plugin to check the user belonging to the access token is still valid (the `is-valid-user` call) and if the user is valid, it makes a call to fetch the roles belonging to the user (the `get-roles` call). Making these calls to the plugin on each use of access token is expensive, as the plugin needs to talk to the remote provider(LDAP or Github) to fetch this information.

Making this request on every API call might cause rate limiting or performance problems. Hence, we should cache this information for some interval (expecting it to be `30 mins`).

rajiesh:
Verified on 19.2.0 (8610-4713200bc880ce17014d826e20a30ed032e3ac26)

PR #5839, opened 2019-02-12
Cache Access token related authorization plugin api calls for 30 minutes (#5823)

GaneshSPatil:
* Cache 'is-user-valid' and 'get-user-oles' plugin API calls for 30 minutes.
* After 30 mins, the cache will be invalidated and a new call will be made to
  the plugin to fetch the new response.
* Use com.google.common.cache.CacheBuilder to implement the time-based cache.

References:
https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheBuilder.java

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
@@ -23,6 +23,7 @@ import com.thoughtworks.go.config.PluginRoleConfig;
 import com.thoughtworks.go.config.SecurityAuthConfig;
 import com.thoughtworks.go.plugin.access.authorization.AuthorizationExtension;
 import com.thoughtworks.go.plugin.domain.authorization.AuthenticationResponse;
+import com.thoughtworks.go.util.SystemEnvironment;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -32,12 +33,11 @@ import java.util.stream.Collectors;
 
 @Component
 public class AuthorizationExtensionCacheService {
-    private final int CACHE_EXPIRY_IN_MINUTES = 30;
+    private final int CACHE_EXPIRY_IN_MINUTES = SystemEnvironment.getGoServerAuthorizationExtensionCallsCacheTimeoutInSeconds() / 60;
 
-    private Cache<String, Boolean> isValidUserCache;
-    private Cache<String, AuthenticationResponse> getUserRolesCache;
+    private final Cache<String, Boolean> isValidUserCache;
+    private final Cache<String, AuthenticationResponse> getUserRolesCache;
     private final AuthorizationExtension authorizationExtension;
-    private final String IS_VALID_USER_CACHE_KEY = "is_valid_user_request_for";
 
     public AuthorizationExtensionCacheService(AuthorizationExtension authorizationExtension, Ticker ticker) {
         this.authorizationExtension = authorizationExtension;
@@ -53,7 +53,7 @@ public class AuthorizationExtensionCacheService {
     }
 
     public boolean isValidUser(String pluginId, String username, SecurityAuthConfig authConfig) {
-        String cacheKey = cacheKeyFor(IS_VALID_USER_CACHE_KEY, pluginId, username, authConfig);
+        String cacheKey = cacheKeyFor(pluginId, username, authConfig);
         Boolean fromCache = isValidUserCache.getIfPresent(cacheKey);
 
         if (fromCache == null) {
@@ -65,7 +65,7 @@ public class AuthorizationExtensionCacheService {
     }
 
     public AuthenticationResponse getUserRoles(String pluginId, String username, SecurityAuthConfig authConfig, List<PluginRoleConfig> pluginRoleConfigs) {
-        String cacheKey = cacheKeyFor(IS_VALID_USER_CACHE_KEY, pluginId, username, authConfig, pluginRoleConfigs);
+        String cacheKey = cacheKeyFor(pluginId, username, authConfig, pluginRoleConfigs);
         AuthenticationResponse fromCache = getUserRolesCache.getIfPresent(cacheKey);
 
         if (fromCache == null) {
@@ -76,13 +76,12 @@ public class AuthorizationExtensionCacheService {
         return fromCache;
     }
 
-    private String cacheKeyFor(String key, String pluginId, String username, SecurityAuthConfig authConfig, List<PluginRoleConfig> pluginRoleConfigs) {
-        String roleConfigNames = pluginRoleConfigs.stream().map(role -> role.getName().toString()).collect(Collectors.joining("__"));
-
-        return String.format("%s__%s__%s_auth_config_%s_role_configs_%s", key, pluginId, username, authConfig.getId(), roleConfigNames);
+    private String cacheKeyFor(String pluginId, String username, SecurityAuthConfig authConfig, List<PluginRoleConfig> pluginRoleConfigs) {
+        String roleConfigNames = pluginRoleConfigs.stream().map(role -> role.getName().toString()).sorted().collect(Collectors.joining("&&"));
+        return String.format("%s##%s", cacheKeyFor(pluginId, username, authConfig), roleConfigNames);
     }
 
-    private String cacheKeyFor(String key, String pluginId, String username, SecurityAuthConfig authConfig) {
-        return String.format("%s__%s__%s__%s", key, pluginId, username, authConfig.getId());
+    private String cacheKeyFor(String pluginId, String username, SecurityAuthConfig authConfig) {
+        return String.format("%s##%s##%s", pluginId, username, authConfig.getId());
     }
 }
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+@Component
+public class AuthorizationExtensionCacheService {
+    private final int CACHE_EXPIRY_IN_MINUTES = 30;
```
ketan: Move to system property?

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+    private Cache<String, AuthenticationResponse> getUserRolesCache;
+    private final AuthorizationExtension authorizationExtension;
+    private final String IS_VALID_USER_CACHE_KEY = "is_valid_user_request_for";
+
+    public AuthorizationExtensionCacheService(AuthorizationExtension authorizationExtension, Ticker ticker) {
+        this.authorizationExtension = authorizationExtension;
+        isValidUserCache = CacheBuilder.newBuilder()
+                .ticker(ticker).expireAfterWrite(CACHE_EXPIRY_IN_MINUTES, TimeUnit.MINUTES).build();
+        getUserRolesCache = CacheBuilder.newBuilder()
+                .ticker(ticker).expireAfterWrite(CACHE_EXPIRY_IN_MINUTES, TimeUnit.MINUTES).build();
```
ketan: What is the default max capacity of the cache? Should we consider setting a max size on the number of items in cache?

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+
+        if (fromCache == null) {
+            fromCache = authorizationExtension.getUserRoles(pluginId, username, authConfigs, pluginRoleConfigs);
+            getUserRolesCache.put(cacheKey, fromCache);
+        }
+
+        return fromCache;
+    }
+
+    private String cacheKeyFor(String key, String pluginId, String username, List<SecurityAuthConfig> authConfigs, List<PluginRoleConfig> pluginRoleConfigs) {
```
ketan: First arg seems redundant?

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+        String authConfigIds = authConfigs.stream().map(PluginProfile::getId).collect(Collectors.joining("__"));
```
ketan: Consider sorting the IDs to get a consistent key.

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+        String roleConfigNames = pluginRoleConfigs.stream().map(role -> role.getName().toString()).collect(Collectors.joining("__"));
```
ketan: Consider sorting again. Should we consider the type of the role, in addition to the name?

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+
+        return String.format("%s__%s__%s_auth_configs_%s_role_configs_%s", key, pluginId, username, authConfigIds, roleConfigNames);
```
ketan: `__` are valid identifiers for all inputs that form a part of the cache key, which may allow an attacker some things to cause key conflicts to cause the cache to respond with something it should not.

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Autowired
+    public AuthorizationExtensionCacheService(AuthorizationExtension authorizationExtension) {
+        this(authorizationExtension, Ticker.systemTicker());
+    }
+
+    public boolean isValidUser(String pluginId, String username, SecurityAuthConfig authConfig) {
+        String cacheKey = cacheKeyFor(IS_VALID_USER_CACHE_KEY, pluginId, username, authConfig);
+        Boolean fromCache = isValidUserCache.getIfPresent(cacheKey);
```
ketan: This can possibly be simplified by doing:

```java
return cache.get(key, () -> {
  return expensiveComputation();
})
```

Or even:

```java
cache = CacheBuilder.build((key) -> {
   return expensiveComputation();
});

...

return cache.get(thing);
```

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
GaneshSPatil: @ketan -- there is no maximum size on the cache. Not sure if we want to add a max cache size. As the number of elements in the cache can never be greater than the total number of access tokens.
And the cache will be invalidated after 30 mins, so no eviction too is required.

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
GaneshSPatil: @ketan -- I too thought about this approach. But the `expensiveComputation` requires a lot more information than just the cache key.

For eg: `IsValidUserCache`, the cache key contains `plugin-id, username, auth-config-id` but to compute if the user is valid, we need plugin-id, username and entire auth config, which can not be extracted from the key.

To use the inline cache builder, we need a way to get all the information from the cache key and then compute the operation. Which seemed an overhead, hence used this approach.

GaneshSPatil:
[DO NOT MERGE] This implementation has a functional bug.
Currently, all the calls are cached based on the auth-config-id and role-config names.

Imagine a situation where auth config id is same, but the auth config configuration has changed. In this scenario, unauthorized users will be able to use the system for a time window of 30 mins (cache timeout).

Long version explained using an example:
###### We have the following 2 organizations on GitHub:
1. `gocd`
2. `gocd-private`

##### User `bob` is part of `gocd` org, but is **NOT** in `gocd-private` org.

Follow the below steps to reproduce the bug:
1. Create an auth config for GitHub auth plugin and specify `gocd` organization.
2. Login as `bob` github user and create a personal access token.
3. Use the token once to access GoCD APIs
4. Edit auth config and change organization to `gocd-private`. Now bob user is unable to login to GoCD web console as he is not part of `gocd-private` organization.
5. But, Bob user can use his personal access token to access GoCD APIs (at max for next 30 mins).

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,100 @@
...
+    private final AuthorizationExtension authorizationExtension;
+
+    public AuthorizationExtensionCacheService(GoConfigService goConfigService, AuthorizationExtension authorizationExtension, Ticker ticker) {
+        this.authorizationExtension = authorizationExtension;
+        isValidUserCache = CacheBuilder.newBuilder()
+                .ticker(ticker).expireAfterWrite(CACHE_EXPIRY_IN_MINUTES, TimeUnit.MINUTES).build();
+        getUserRolesCache = CacheBuilder.newBuilder()
+                .ticker(ticker).expireAfterWrite(CACHE_EXPIRY_IN_MINUTES, TimeUnit.MINUTES).build();
+
+        goConfigService.register(this.securityConfigChangeListener());
```
maheshp: @GaneshSPatil we should be handling changes to Security through config.xml as well right?

GaneshSPatil: @maheshp -- fixed as part of https://github.com/gocd/gocd/pull/5852

server/src/main/java/com/thoughtworks/go/server/service/AuthorizationExtensionCacheService.java
```diff
@@ -0,0 +1,100 @@
...
+
+        if (fromCache == null) {
+            fromCache = authorizationExtension.getUserRoles(pluginId, username, authConfig, pluginRoleConfigs);
+            getUserRolesCache.put(cacheKey, fromCache);
+        }
+
+        return fromCache;
+    }
+
+    private String cacheKeyFor(String pluginId, String username, SecurityAuthConfig authConfig, List<PluginRoleConfig> pluginRoleConfigs) {
```
bdpiparva: Use `https://github.com/gocd/gocd/blob/c85904bba350615ad5f16b7a39b1d6ba19dad4df/server/src/main/java/com/thoughtworks/go/server/cache/CacheKeyGenerator.java#L27` for cache key generation?

PR #5845, opened 2019-02-13
Authorisation extension: `get-user-roles` request should be made with only one auth_config.

vrushaliwaykole:


GaneshSPatil:
Fixes: https://github.com/gocd/gocd/issues/5846

PR #5848, opened 2019-02-13
Revoke tokens for user when user is deleted.

### LINKED ISSUE #5818: Deletion of User should revoke access token.
##### Issue Type

- Feature proposal

##### Summary
Upon deletion of the user, all the access token belonging to the user should be revoked.

ketan:
Fixes #5818


bdpiparva:
@ketan -- Need to add same for the Postgres addon.

Issue #5842, opened 2019-02-12
ETag of plugin settings changes as config-repos are parsed


tomzo:
##### Issue Type

- Bug Report

##### Summary

ETag on `go/api/admin/plugin_settings/<plugin-id>` is changing way too often. In fact it changes each time when any of the config-repos parses a commit.

How I found this and reproduction:
1. GoCD server is starting with a bunch of config-repos, so there is expected a few minutes when config-repos are parsing materials.
2. I query `curl -i -u **** -H 'Accept:application/vnd.go.cd.v1+json' 'http://go.ai-traders.com/go/api/admin/plugin_settings/cd.go.contrib.elastic-agent.openstack' ` to get current configuration of the plugin.
3. Observed that on each response the `ETag` header is different, while plugin configuration is unchanged.

##### Environment

Example response:
```
curl -i -u **** -H 'Accept:application/vnd.go.cd.v1+json' 'myserver/go/api/admin/plugin_settings/cd.go.contrib.elastic-agent.openstack' 
HTTP/1.1 200 OK
Server: nginx/1.15.6
Date: Tue, 12 Feb 2019 17:18:35 GMT
Content-Type: application/vnd.go.cd.v1+json; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Set-Cookie: JSESSIONID=node0ak438u3bb3mfe186qi9z1fkn662.node0;Path=/go;Expires=Tue, 26-Feb-2019 17:18:35 GMT;Max-Age=1209600;HttpOnly
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Set-Cookie: JSESSIONID=node019crqdj1745z01m7ia1kgaj66o663.node0;Path=/go;Expires=Tue, 26-Feb-2019 17:18:35 GMT;Max-Age=1209600;HttpOnly
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Referrer-Policy: strict-origin-when-cross-origin
ETag: W/"19d9dec6ce1452a5744b89b8053bd25c"
Cache-Control: max-age=0, private, must-revalidate
X-Request-Id: a9a40727e2e04d87b7ca86134eed0ffb
X-Runtime: 0.105705
Vary: Accept-Encoding, User-Agent

{ ... plugin settings ... }
```

###### Basic environment details

[...]

PR #5854, opened 2019-02-14
Ensuring Etag is consistent when plugin settings does not change

prateekbaheti:
Fix for issue #5842
 -  Using PluginSettings HashCode to generate the entity md5 instead of json since json has encrypted secure environment variables whose value changes one each request due to new salt.
- Fixing usage of entity md5 in plugins_controller by passing the md5 of plugin setting fetched from db. This worked earlier since the md5 was different for each call and hence the stale check in PluginService.updatePluginSettings would never be true.

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
@@ -76,12 +76,13 @@ public class EntityHashingServiceTest {
 
     @Test
     public void shouldUseObjectHashCodeForPluginSettings() {
-        PluginSettings pluginSettings = new PluginSettings();
-        String expectedMd5 = CachedDigestUtils.md5Hex(String.valueOf(pluginSettings.hashCode()));
+        PluginSettings pluginSettings = new PluginSettings("com.foo.plugin");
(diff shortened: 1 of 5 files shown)


server/src/test-fast/java/com/thoughtworks/go/server/service/EntityHashingServiceTest.java
```diff
@@ -73,6 +74,16 @@ public void shouldRegisterToListenForConfigChange() {
         verify(goConfigService).register(entityHashingService);
     }
 
+    @Test
+    public void shouldUseObjectHashCodeForPluginSettings() {
+        PluginSettings pluginSettings = new PluginSettings();
+        String expectedMd5 = CachedDigestUtils.md5Hex(String.valueOf(pluginSettings.hashCode()));
```
arvindsv: @prateekbaheti Would it be possible to write a test with a known hash, so that if something gets added, the test would fail and the person making the change can decide whether it should affect the hash?

prateekbaheti: Makes sense, I will change that.

prateekbaheti:
@arvindsv It got removed accidentally. I have fixed it and ensured a test fails if that is missed.

arvindsv:
Ok, thanks. I'm assuming someone else will review this. Mine was just a drive-by, but looks ok to me.

PR #5865, opened 2019-02-18
Split access token API into 2 to simplify URLs

ketan:
- /go/api/current_user/access_tokens: for the current user
- /go/api/admins/access_tokens: for admins to list tokens for all users

PR #5867, opened 2019-02-18
Updated the access token SPA after the model changes

kritika-singh3:
1. Updated the AccessTokenJSON and AccessToken Model in types.ts
2. Updated the crud and spark routes to utilize the new URL for personal tokens


bdpiparva:
This is done as Access token API response changed from server side in PR #5865

Issue #5786, opened 2019-01-29
[Maintenance Mode] Add time stamp for the last triggered Maintenance Mode along with date


adityasood:
##### Issue Type
- Feature enhancement

##### Summary

It would be ideal to add the timestamp along with the date and name of the last user who initialized the server maintenance mode.

GaneshSPatil:
@adityasood -- we currently show the maintenance mode changed user and time.

<img width="828" alt="screen shot 2019-02-19 at 10 12 17 am" src="https://user-images.githubusercontent.com/15275847/52990596-f7150280-342e-11e9-8d0e-4f14823181e7.png">
<img width="787" alt="screen shot 2019-02-19 at 10 12 28 am" src="https://user-images.githubusercontent.com/15275847/52990597-f7150280-342e-11e9-9a25-fe7629311871.png">


is there anything else that we want to address as part of this issue?

GaneshSPatil:
@adityasood -- sorry for the above response. Currently, it just shows the date and username. Whereas it should also show time along with the date.

rajiesh:
Verified on 19.2.0 (8576-004ea5ec2257f2ef7bfbddfbf74ffa952ecb6dca)

PR #5873, opened 2019-02-19
Show maintenance mode updated by timestamp along with date (#5786)

GaneshSPatil:
#### Before
screen shot 2019-02-19 at 11 58 19 am

#### After
screen shot 2019-02-19 at 12 04 56 pm


maheshp:
Fixes #5786

Issue #5037, opened 2018-08-17
Console Log unable to parse "\n" 


adityasood:
##### Issue Type
- Bug Report
##### Summary

Currently, if a user has a script which uses "\n" character set in their task execution, the console log fails to parse and does not show up from this task onwards. This can easily be reproduced with the YAML file attached with the issue

###### Basic environment details

* Go Version: 18.7.0
* JAVA Version: Java8
* OS: MacOs
* Browser vendor and version (if relevant): Chrome

##### Steps to Reproduce

1. Create a YAML pipeline using the attached YAML file
2. Run the task. The console logs does not show up
3. Inspect using the browser developer tools, you will observe parse failure error (attached)
4. Edit the YAML file to remove "\n" from the task. The console log will now show up

##### Expected Results
The console logs should show up.

##### Actual Results
The console logs do not show up. There is a parse error

<img width="1376" alt="screen shot 2018-08-17 at 11 11 38 am" src="https://user-images.githubusercontent.com/17826464/44249705-b681c580-a20e-11e8-8c5c-6563dcafa09d.png">
[sample-yaml-file.txt](https://github.com/gocd/gocd/files/2296105/sample-yaml-file.txt)
[console.log](https://github.com/gocd/gocd/files/2296107/console.log)
<img width="1440" alt="screen shot 2018-08-17 at 11 04 25 am" src="https://user-images.githubusercontent.com/17826464/44249747-e92bbe00-a20e-11e8-991f-1fac608ab29f.png">

GaneshSPatil:
For users facing this issue, Please use the `console.log` artifact to see the job console.

Navigate to `Artifacts Tab`  -> `cruise-output/console.log`

GaneshSPatil:
@adityasood -- @vrushaliwaykole  and I will be taking a look at this issue.

ankitsri11:
Fixed as part of this PR: https://github.com/gocd/gocd/pull/5877

Verified on GoCD Version: 19.2.0 (8559-38e68d0ca9dfce367dd1d9a380e824595cd87d34)

PR #5877, opened 2019-02-19
Render job console log containing new line characters in executable task (#5037)

GaneshSPatil:
#### Raw Console Log for exec task containing new line:
```task
##|12:33:01.795 [go] Start to build yamlpipe123/1/build/1/build on in-ganeshpl-2.local [/Users/ganeshp/projects/gocd/gocd/agent]
!!|12:33:01.817 [go] Task: /bin/bash -c "echo -e 'this is 
!!|12:33:01.817 theBestUse 
!!|12:33:01.817 ofOurplugin
!!|12:33:01.817 '>testFile.txt"
?0|12:33:01.988 [go] Task status: passed (168 ms)
j0|12:33:02.017 [go] Current job status: passed
```

#### Why did it fail?
Console Log Appender was expecting the exec task in the format `!!|{DATE} [go] Task: {COMMAND} \n`, but as the task itself was split into multiple lines, it was assuming each line as a separate task.

As the subsequent lines from the exec task command werent satisfying the exec task format, it was failing to render such tasks.

---

#### Fix:

Each line from the console is processed via Console log output transformer which formats and prettifies the log content. (eg: printing red, green, yellow, etc).

Before sending the logs to transformer, inline multiline executable tasks into a single line.

With this fix, all the existing logs should continue to work as is. This fix is concats multi-line exec task command into a single line and then sends it to log transformer for processing.

### DIFF
--- a/server/webapp/WEB-INF/rails/app/assets/javascripts/log_output_transformer.js
+++ b/server/webapp/WEB-INF/rails/app/assets/javascripts/log_output_transformer.js
@@ -158,8 +158,45 @@
       flushToDOM();
     }
 
+    function isTaskLine(line) {
+      return line.startsWith('!!');
+    }
+
+    function getOnlyCommand(line) {
+      return line.match(/\ (.*?)\n/)[1];
+    }
+
+    function appendNewLineIfDoesNotExists(line) {
+      return line.endsWith('\n') ? line : line + '\n';
+    }
+
+    function doesContainGoMarker(lineToProcess) {
+      return lineToProcess.indexOf('[go]') !== -1;
+    }
+
+    function inlineMultilineExecutableTask(logLines) {
+      var logLinesToRemove         = [];
+      var currentExecutableTaskIndex = -1;
+      for (var i = 0; i < logLines.length; i++) {
+        var lineToProcess = logLines[i];
+        if (isTaskLine(lineToProcess) && doesContainGoMarker(lineToProcess)) {
+          currentExecutableTaskIndex = i;
+        }
+
+        if (isTaskLine(lineToProcess) && !doesContainGoMarker(lineToProcess)) {
+          logLines[currentExecutableTaskIndex] = appendNewLineIfDoesNotExists(logLines[currentExecutableTaskIndex]).replace("\n", "\\n") + getOnlyCommand(lineToProcess);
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
@adityasood -- can you help us verifying whether the changes on this PR fixes the issue?

ketan:
Shouldn't the fix be for whoever generates the line (agent side) to escape
the "\n"? Otherwise, you'd need to consider "\r" and "\r\n" :)

- Ketan

GaneshSPatil:
@ketan -- Yes, ideally the fix should have been on the agent side (as it produces the logs).
But changing it at the agent side was a high-risk changeset as:
1. We required to change the separator or treat `\n` separator from the exec task specially. Changing the producer would have caused the renderer to change too and would have resulted in a much bigger change.
2. Changing console-log formatter at agent might cause bugs which result in losing logs. (Critical bug).

Hence, we've avoided changing it at either the console-log-producer (agent) or the console-log-appender (server). This fix has specifically been added for multiline exec tasks, which is a very low-risk changeset.

ankitsri11:
I tried with below configuration and it works fine. 

```
format_version: 3
pipelines:
  Sample:
    group: defaultGroup
    label_template: ${COUNT}
    lock_behavior: none
    materials:
      git:
        git: https://github.com/adityasood/licensemgmt.git
        shallow_clone: false
        auto_update: true
        branch: master
    stages:
    - defaultStage:
        fetch_materials: true
        keep_artifacts: false
        clean_workspace: false
        approval:
          type: success
        jobs:
          defaultJob:
            timeout: 0
            tasks:
            - exec:
                arguments:
                - -c
                - echo -e 'this is \ntheBestUse \nofOurplugin\n'>testFile.txt
                command: /bin/bash
                run_if: passed
            - exec:
                arguments:
                - -c
                - cat testFile.txt
                command: /bin/bash
                run_if: passed
```

Console output:

```
[go] Task: /bin/bash -c "echo -e 'this is \ntheBestUse \nofOurplugin\n'>testFile.txt"took: 0.111s
[go] Task status: passed, took: 0.111s
[go] Task: /bin/bash -c "cat testFile.txt"took: 0.112s
this is 
theBestUse 
ofOurplugin

[go] Task status: passed, took: 0.112s
```

PR #5843, opened 2019-02-12
Reparse previously failed Config Repo parses if Server side issues have been resolved

ibnc:
see the first issue in #5792 

Here's how to reproduce the issue this PR is fixing:
1. Add a config repo with the following yaml
```
format_version: 3 
pipelines:
  blah:
    group: first
    label_template: ${COUNT}
    lock_behavior: none
    materials:
      git:
        git: git@github.com:tomzo/gocd-yaml-config-plugin.git
        shallow_clone: false
        auto_update: true
        branch: master
      dependency:
        pipeline: aincrad
        stage: defaultStage
    stages:
    - '2':
        fetch_materials: true
        keep_artifacts: false
        clean_workspace: false
        approval:
          type: success
        jobs:
          '2':
            run_instances: '0'
            tasks:
            - exec:
                command: ls
                run_if: passed
```
2. Wait until that revision is parsed, and the error appears on the config repo page and in the server health messages
3. Create the pipeline `aincrad` with the stage `defaultStage`
4. The server health message will disappear, but it's impossible to have the config repo config merged without a fake commit or server restart

When testing if this works, be sure to either delete the pipeline aincrad or change the yaml before merging the code and starting the server. 

This PR will remove the need to restart the server or make a fake commit. It'll be possible for the failed parse to be retried after the blocking issue is resolved (in the case of the above example, once the pipeline is created). 

@maheshp @GaneshSPatil Would y'all be up for reviewing at this PR?

PR #5879, opened 2019-02-19
Fix for integration test random failure

ibnc:
This PR contains a fix for a randomly failing integration test introduced in #5843

See [here](https://build.gocd.org/go/tab/build/detail/build-linux/3479/build-server/2/ServerIntegrationTests-runInstance-2#) for the specific failure


ibnc:
Merging since it's a small change to fix an integration test that has been randomly failing

arvindsv:
Do we know why it was randomly failing?

ibnc:
The reason behind the random failure is that the changes introduced in my previous PR invalidated the check we were doing in the test. We essentially weren't looking at the correct source of truth on whether or not the parsing had failed (i.e. we weren't looking at the server health messages)

Issue #5715, opened 2019-01-16
Dashboard UI - provide option to create a pipeline under a pipeline group


varshavaradarajan:
##### Issue Type

- Feature enhancement

##### Summary

If the pipelines are grouped by pipeline groups on the dashboard, then provide an option to create a pipeline within that group from the dashboard for admins and group admins. This will avoid users having to go to `/go/admin/pipelines`, search for the specific group and then clicking on `Create a pipeline within this group.`

GaneshSPatil:
Done as part of PR: https://github.com/gocd/gocd/pull/5885 in GoCD `v19.3.0`

PR #5885, opened 2019-02-20
Add a link to create pipeline from dashboard when grouped by pipeline groups. (#5715)

varshavaradarajan:
#5715 

screen shot 2019-02-19 at 11 57 44 pm

Um, I don't know if there is a better way to show the add icon. 
cc @naveenbhaskar

### DIFF
--- a/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
+++ b/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss
@@ -174,7 +174,7 @@ body {
 .dashboard-group_title {
   background:    #e7eef0;
   margin:        0;
-  padding:       10px 30px;
+  padding:       10px 10px 10px 30px;
   position:      relative;
 
   border-radius: $global-border-radius $global-border-radius 0 0;
@@ -253,6 +253,17 @@ body {
 .dashboard-group_title {
   .tooltip {
     top: 120%;
+  }
+
+  .dashboard-group_add_pipeline {
+    float:      right;
+    background: #fff;
+    padding:    2px 10px;
+    .icon_add {
+      color:        $icon-color;
+      font-weight:  600;
+      line-height:  25px;
+    }
   }
 }
 
--- a/server/webapp/WEB-INF/rails/config/initializers/jsroutes.rb
+++ b/server/webapp/WEB-INF/rails/config/initializers/jsroutes.rb
@@ -2,6 +2,7 @@ JsRoutes.setup do |config|
   config.prefix = com.thoughtworks.go.util.SystemEnvironment.new.getWebappContextPath
   config.camel_case = true
   config.include = [
+    /^pipeline_new/,
     /analytics/,
     /^api_internal/,
     /^apiv\d/,
--- a/server/webapp/WEB-INF/rails/spec/webpack/models/dashboard/dashboard_groups_spec.js
+++ b/server/webapp/WEB-INF/rails/spec/webpack/models/dashboard/dashboard_groups_spec.js
@@ -44,6 +44,9 @@ describe("DashboardGroups", () => {
       expect(pipelineGroups.groups[0].tooltipForEdit()).toBe("");
       expect(pipelineGroups.groups[0].titleForEdit()).toBe("Edit Pipeline Group 'first'");
       expect(pipelineGroups.groups[0].ariaLabelForEdit()).toBe("Edit Pipeline Group 'first'");
+      expect(pipelineGroups.groups[0].tooltipForNewPipeline()).toBe("");
+      expect(pipelineGroups.groups[0].titleForNewPipeline()).toBe("Create a new pipeline within this group");
+      expect(pipelineGroups.groups[0].ariaLabelForNewPipeline()).toBe("Create a new pipeline within this group");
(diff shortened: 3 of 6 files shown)


GaneshSPatil:
@varshavaradarajan -- 
Wondering if the '+' is the best icon to indicate `create pipeline within the group` operation.

Few questions:
1. `+` icon shouldn't be confusing. What does it mean? Create a pipeline within the group? Or Create a new pipeline group itself? 
2. Allowing to create new pipeline within a group, but not for environments. Will it be consistent?

\cc: @naveenbhaskar

varshavaradarajan:
> `+` icon shouldn't be confusing. What does it mean? Create a pipeline within the group? Or Create a new pipeline group itself?

If its specified along with the group name, it doesn't mean creating a new pipeline group, right?

> Allowing to create new pipeline within a group, but not for environments. Will it be consistent?

Why are we aiming for consistency in this matter? Pipeline group and environment are 2 different entities. A pipeline must always belong to a pipeline group while a pipeline can be optionally part of some environment. Why is it not okay to display options that make sense for a pipeline group if it means improving the UX? If you think that doing this is not necessary, feel free to close the PR, though.

maheshp:
>  If you think that doing this is not necessary, feel free to close the PR, though.

@varshavaradarajan IMO, having an ability to create a pipeline on dashboard would be useful over going to a different page to do the same. We just need to decide on the right UX for this.

arvindsv:
I always came up with this in my head, when I thought of the ability to add a pipeline to a group:

ketan:
I like Aravind's idea, given the confusion about what the add button on the
group name header means.

- Ketan

varshavaradarajan:
@arvindsv - should that add box be shown only for admins or be shown to everyone with the add button disabled for view users? 

Also, are there plans to make a dashboard group expand/collapse? Because, having the add box in that case, may not be of much use as it is still the same number of interactions as clicking on the group name from the dashboard and then clicking on "create a pipeline within this group" in `/go/admin/pipelines`?

arvindsv:
> @arvindsv - should that add box be shown only for admins or be shown to everyone with the add button disabled for view users?

I think it should show only for admins. /cc @naveenbhaskar Thoughts?

> Also, are there plans to make a dashboard group expand/collapse? Because, having the add box in that case may not be of much use as it is still the same number of interactions as clicking on the group name from the dashboard and then clicking on "create a pipeline within this group" in `/go/admin/pipelines`?

True, but it's still an extra page load in that case. I think this box is still useful.

naveenbhaskar:
if we add a  blank pipeline box at the end of pipeline group, the problems i see is 
1. If there are a lot of pipelines in a group, user has to scroll a lot to reach the end and click on it.
2. There was a proposal for expand-collapse the pipeline groups, so this is not easily reachable.

The best place for adding a pipeline to the group is pipeline group header itself. it is  easily accessible  and noticeable and it should only show for people who can add pipelines. 

image


This design will be consistent with design for admin-> pipelines  

image

ibnc:
Would the possibility of combining @arvindsv @naveenbhaskar's designs be out of the question?

@arvindsv`s design would be really helpful/intuitive for new users, and @naveenbhaskar's design would add benefit for instances with a great number of pipelines.

thoughts?

ketan:
+1 to what @ibnc said.

arvindsv:
I like that suggestion, if it makes sense from a design perspective.

varshavaradarajan:
@naveenbhaskar - can you please review this PR when you're free? Thank you! 

screen shot 2019-03-01 at 6 58 22 pm

varshavaradarajan:
@arvindsv, @ibnc - by the way, just wanted to point out that a pipeline group which has no pipelines is not shown on the dashboard. So, this is only for making it easier to add pipelines to groups which already has at least one pipeline. Not sure if you want to change that as part of improving the ux for new users.

streisguth:
Hi all, I know I'm a little late to this issue but I do not think that having both buttons is a good design solution. I am anticipating that having the same action in two similar places will be confusing to a user, as I would expect each of those buttons to have an action _different from one another_.

My preference would be to have the button in the top of the pipeline group *only*, as it is the design that covers the use cases of having either a small number of pipelines OR a large number of pipelines.

GaneshSPatil:
+1 to @streisguth 
Keeping 2 buttons to perform the same operation is confusing. 

Can we do smart positioning?
If there are more than X pipelines (lets say 15), use the button at the top of the pipeline group.
And if there are less than X pipelines, use the '+' pipeline box?

Note:
Also, with this approach, the sudden change in the feature usability might cause confusion.

naveenbhaskar:
+1 to @streisguth .. There should not be two action element for the same action.

varshavaradarajan:
@naveenbhaskar - I've already made changes suggested by @streisguth . The `New Pipeline` button shows up in the dashboard group header like your screenshot. Do you mind reviewing the changes so we can get this merged?

naveenbhaskar:
sure @varshavaradarajan

naveenbhaskar:
@varshavaradarajan .. I have verified it and it is good to go from UI point of view.

GaneshSPatil:
The final UI screenshot:
Screen Shot 2019-03-25 at 12 40 35 PM

\cc: @naveenbhaskar @varshavaradarajan @rajiesh @adityasood

Issue #5861, opened 2019-02-15
Agents Page should not allow users to modify Environments associated through Config Repo


akshaydewan:
Related to #5761

### Summary
If an Agent is associated with an Environment via a Config Repo, it should not be editable (meaning, a user shouldn't be able to update the agent and un-select the Envrionment). The current UI makes this possible, but results in a no-op with a success message.

### Implementation
The Agents API returns a model that contains only the associated environment names as strings. We will need to add a flag saying whether the environment was associated from a Config Repo.

Current response example:
```json
{
  "_links" : { //..snipped
  },
  "_embedded" : {
    "agents" : [ {
      "_links" : { //..snipped
      },
      "uuid" : "1c7982c8-6615-492f-871b-1a98f70c08e5",
      "hostname" : "in-akshaydewan.local",
      "ip_address" : "127.0.0.1",
      "sandbox" : "",
      "operating_system" : "",
      "agent_config_state" : "Enabled",
      "agent_state" : "LostContact",
      "environments" : [ "dev" ],
      "build_state" : "Unknown",
      "free_space" : "unknown",
      "resources" : [ ]
    }, {
      "_links" : { //..snipped
      },
      "uuid" : "f7ba19f9-21f6-4120-aec1-db6d4564a115",
      "hostname" : "in-akshaydewan.local",
      "ip_address" : "127.0.0.1",
      "sandbox" : "",
      "operating_system" : "",
      "agent_config_state" : "Enabled",
      "agent_state" : "LostContact",
      "environments" : [ ],
      "build_state" : "Unknown",
      "free_space" : "unknown",
      "resources" : [ ]
    } ]
  }
}
```

The change in the environments object can be:
```json
{
  "environments": [
    {
      "name": "dev",
      "origin": {
        "type": "config-repo",
        "_links": {
          "self": {
            "href": "http://localhost:8153/go/api/admin/config_repos/foo"
          },
          "doc": {
            "href": "https://api.gocd.org/#config-repos"
          },
          "find": {
            "href": "http://localhost:8153/go/api/admin/config_repos/:id"
          }
        }
      }
    }
  ]
}
```

It will require the Agents API version to be bumped up.

There is another call which is made to an internal Environments endpoint `/go/api/admin/internal/environments`, but we don't send any Agents' information here, and IMO, it makes more sense to add the information to the Agent.

### Question
[...]

PR #5886, opened 2019-02-21
Update Agents page to render disabled checkbox for Config Repo Environments

akshaydewan:
Issue #5861 

- Bumped up Agents API to v5 to include "associated_from_config_repo" flag
- Updated the UI to render (disabled) checkbox and tooltip where applicable

screenshot 2019-02-21 at 13 38 58


api/api-agents-v5/src/main/java/com/thoughtworks/go/apiv5/agents/representers/EnvironmentsRepresenter.java
```diff
@@ -0,0 +1,49 @@
...
+public class EnvironmentsRepresenter {
+    public static void toJSON(OutputListWriter writer, Collection<EnvironmentConfig> environments) {
+        List<EnvironmentConfig> sortedEnvironmentConfigs = environments.stream()
+                .sorted(new EnvironmentConfigComparator())
+                .collect(Collectors.toList());
+
+        for (EnvironmentConfig environment : sortedEnvironmentConfigs) {
+            writer.addChild((childWriter) -> {
+                childWriter.add("name", environment.name());
+                childWriter.add("associated_from_config_repo", !environment.isLocal());
```
GaneshSPatil: @akshaydewan -- On dashboard API response, we've defined `from_config_repo` field. can this also be called `from_config_repo`?

Secondly, we were thinking of sending the `origin` object as part of the API response, stating where the config entity defined.
The benefit of sending origin object is, we can specifically say, this `environment -> to -> agent` association has been defined in `example-foo` config repo. And have a link to the config-repo-spa as well for further details.

A similar issue for dashboard pipelines defined in config repo is listed [here](https://github.com/gocd/gocd/issues/4403).

**Note:** This doesn't need to happen now. can be done later too (at an overhead of bumping up API version) :)

akshaydewan: The tricky bit here is that we want to know *association* between an agent and an Environment. The Environment may be defined in a Config Repo, but it's association with an Agent can be defined from the UI (which will end up in XML). [See [EnvironmentConfig.isLocal()](https://github.com/gocd/gocd/blob/58886bb2b85294bb94e805651bb90ca75fae330f/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentConfig.java#L95) vs [EnvironmentConfig.containsAgentRemotely(uuid)](https://github.com/gocd/gocd/blob/58886bb2b85294bb94e805651bb90ca75fae330f/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentConfig.java#L105) ).

To avoid this ambiguity, I would rather include the word "associated". 

Regarding `origin`, we have the same ambiguity - is the Environment's origin, or the Environment-Agent relationship?

In case of the Dashboard, it is simpler because we need to know the origin of the Pipeline.

arvindsv: If the association is in the XML, won't the environment tag be in the XML and won't the environment then be considered "local" and not remote?

akshaydewan: @arvindsv In such cases, the Environment becomes a `MergeEnvironmentConfig` object and it still returns `isLocal` as false.

arvindsv: @akshaydewan I see. Ok.

GaneshSPatil: @akshaydewan @arvindsv  -- Thanks for the explainations. 

Though, on the enviornments, to resolve the ambiguity of entity-association and entity-definition, we were thinking of something like https://github.com/gocd/gocd/issues/3301#issuecomment-341654482

Wouldn't this apply here too?

We still can name the field `"association"` but can we make it return origin? If not, I am happy with whatever we have :)

akshaydewan: @GaneshSPatil and I had a chat about this. The way we reason about Environments is that the configuration can be `partial` - with some parts defined in the Config Repo, and other parts in XML. So each piece of the  environment configuration can have an `origin`. Ultimately the Environment config object becomes a `MergeEnvironmentConfig`.

With this reasoning, it would make sense to add an `origin` field in the API response. I can make the change in this PR itself to avoid another version bump.

api/api-agents-v4/src/main/java/com/thoughtworks/go/apiv4/agents/AgentsControllerV4.java
```diff
@@ -80,8 +80,8 @@ public String controllerBasePath() {
     @Override
     public void setupRoutes() {
         path(controllerBasePath(), () -> {
-            before("", this::setContentType);
-            before("/*", this::setContentType);
+            before("", mimeType, this::setContentType);
+            before("/*", mimeType, this::setContentType);
```
ketan: I just realized that this is a copy-paste error. All controllers use this particular method, without the `mimeType` param.

PR #5895, opened 2019-02-22
Add an Id validator and use it in Config Repo SPA

akshaydewan:
There was no client side validation and the server returns a 500 when the Id format is invalid, so the user was getting a cryptic error message.

@arvindsv, I've copied the error message from the older code and it's a bit long. Let me know if we want to change it - `Invalid ${attr}. This must be alphanumeric and can contain underscores and periods (however, it cannot start with a period). The maximum allowed length is 255 characters.`

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/models/config_repos/types.ts
+++ b/server/webapp/WEB-INF/rails/webpack/models/config_repos/types.ts
@@ -106,6 +106,7 @@ export class ConfigRepo implements ValidatableMixin {
     }
     ValidatableMixin.call(this);
     this.validatePresenceOf("id");
+    this.validateIdFormat("id");
     this.validatePresenceOf("pluginId");
     this.validateAssociated("material");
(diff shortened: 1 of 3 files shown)


arvindsv:
@akshaydewan As long as it looks ok (wrapping, etc) when that message shows, I think it's fine. It provides exact information about what is expected, without resorting to showing a regex. :) So, trying to shorten that might be hard.

I suspect it'll still fail with a cryptic error message if it is over 255 chars. No?

akshaydewan:
You're right about the scenario with length over 255 chars. I've pushed a fix for that.

Here's how the error looks - I think it looks ok
screenshot 2019-02-22 at 17 32 10

arvindsv:
It looks a bit crowded. But, it is what it is. I think you need to add: "hyphens, " before the word "underscores".

arvindsv:
This looks good to me.

PR #5896, opened 2019-02-22
Access token API fixes

kritika-singh3:
### Summary

1. Returning CONFLICT (409) on revoke access token api when the token in question has already been revoked. 
Note: This needs to be updated in #5869 
2. Made `description` property mandatory in create access token API - in case description is not provided, 422 status is received.
3. Fixed the issue wherein access token APIs were called using existing access tokens.

### DIFF
--- a/api/api-access-token-v1/src/main/java/com/thoughtworks/go/apiv1/accessToken/CurrentUserAccessTokenControllerV1.java
+++ b/api/api-access-token-v1/src/main/java/com/thoughtworks/go/apiv1/accessToken/CurrentUserAccessTokenControllerV1.java
@@ -85,7 +85,7 @@ public class CurrentUserAccessTokenControllerV1 extends AbstractUserAccessTokenC
 
         final JsonReader reader = GsonTransformer.getInstance().jsonReaderFrom(request.body());
 
-        String tokenDescription = reader.optString("description").orElse(null);
+        String tokenDescription = reader.getString("description");
 
         AccessToken created = accessTokenService.create(tokenDescription, currentUsernameString(), currentUserAuthConfigId(request));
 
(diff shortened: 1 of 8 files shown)


api/api-access-token-v1/src/main/java/com/thoughtworks/go/apiv1/accessToken/CurrentUserAccessTokenControllerV1.java
```diff
@@ -85,7 +88,7 @@ public String createAccessToken(Request request, Response response) throws Excep
 
         final JsonReader reader = GsonTransformer.getInstance().jsonReaderFrom(request.body());
 
-        String tokenDescription = reader.optString("description").orElse(null);
+        String tokenDescription = reader.getString("description");
```
ketan: Shouldn't this be a validation error than a 400 response? What if a token is created another way, and somehow a `null` ends up in the DB?

bdpiparva: @ketan - `reader.getString(...)` returns `422` if `description` is not provided.

```bash
< HTTP/1.1 422 Unprocessable Entity
< Date: Mon, 25 Feb 2019 08:40:42 GMT
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< X-UA-Compatible: chrome=1
< Set-Cookie: JSESSIONID=node01lpygaoc8bl43qcv59nuljf1e18.node0;Path=/go;Expires=Mon, 11-Mar-2019 08:40:42 GMT;Max-Age=1209600;HttpOnly
< Expires: Thu, 01 Jan 1970 00:00:00 GMT
< Set-Cookie: JSESSIONID=node011l991wrxtapvtah4mqowdmey19.node0;Path=/go;Expires=Mon, 11-Mar-2019 08:40:42 GMT;Max-Age=1209600;HttpOnly
< Cache-Control: max-age=0, private, must-revalidate
< Content-Type: application/vnd.go.cd.v1+json;charset=utf-8
< X-Runtime: 0
< Transfer-Encoding: chunked
<
* Connection #0 to host localhost left intact
{
  "message": "Json `{\"description1\":\"my first token\"}` does not contain property 'description'"
}
```

ketan: I'm suggesting that we instead make this a model validation, rather than a controller enforced validation.

bdpiparva: Make sense. I will revert the change.

Issue #5858, opened 2019-02-14
/etc/init.d/go-agent stop on Linux does not use PID and kills all agents run by go user with the same service name


adityasood:
##### Issue Type

- Bug Report

##### Summary

If you have the following setup

1. Start an agent on a linux box using the /etc/init.d/go-agent start command. This will run as go user
2. Start a gocd docker agent as a go user.
3. Run ps -aux | grep go-agent. This should list both the agents and check they are run by the same user
4. Kill the agent using the /etc/init.d/go-agent stop command. This will kill both the agents since the killproc function does not honor PID as per [this](https://github.com/gocd/gocd/blob/71a87e5b6a923a45e9b1c0606ee9d93b59ee9205/installers/go-agent/release/go-agent.init#L39)

aqneves:
Hi,

The fix that it's being planned won't fully fix the issue.

The problem is that /etc/init.d/go-agent considers that the pid file is on one specific location:
PID_FILE="/var/run/go-agent/${SERVICE_NAME}.pid"

But, the start_agent.sh can place it in one of two locations:
if [ "$1" == "service_mode" ] && [ -d "/var/run/go-agent" ]; then
PID_FILE="/var/run/go-agent/${SERVICE_NAME}.pid"
else
PID_FILE="$AGENT_WORK_DIR/go-agent.pid"
fi

In our scenario, the file is placed on /var/lib/go-agent/go-agent.pid which means that the stop of the agent will not work.

Regards

ibnc:
> 1. Start an agent on a linux box using the /etc/init.d/go-agent start command. This will run as go user
> 2. Start a gocd docker agent as a go user.
> 3. Run ps -aux | grep go-agent. This should list both the agents and check they are run by the same user
> 4. Kill the agent using the /etc/init.d/go-agent stop command. This will kill both the agents since the killproc function does not honor PID as per [this](https://github.com/gocd/gocd/blob/71a87e5b6a923a45e9b1c0606ee9d93b59ee9205/installers/go-agent/release/go-agent.init#L39)

[...]

PR #5897, opened 2019-02-22
/etc/init.d/go-agent stop stops the agent using pid from PID_FILE

ibnc:
see #5858 for details on the issue this addresses

I would have ideally used `pkill -u go -F $PID_FILE` but pkill on centos 6 does not have a supported way to use the pid file to stop a process. This approach should ideally work on all of the linux distros we support.

I'm also unsure why we were passing arguments we weren't using to `killproc` This line is the one I'm referring to [here](https://github.com/gocd/gocd/compare/master...ibnc:agent_linux_issue?expand=1#diff-48385395d62c8d5e49e9612b048e274dL99)

### DIFF
--- a/installers/go-agent/release/go-agent.init
+++ b/installers/go-agent/release/go-agent.init
@@ -31,12 +31,11 @@ SERVICE_NAME=${0##*/}
 # Strips initV leading chars if the script is located in rcX.d (automatically executed on runlevel change)
 INIT_DIR=${0%/*}; echo ${INIT_DIR##*/} | grep -e '^rc[0-9]\.d$' >/dev/null && SERVICE_NAME=$(echo "${SERVICE_NAME}" | sed 's/^[SK][0-9][0-9]//');
 
-PID_FILE="/var/run/go-agent/${SERVICE_NAME}.pid"
 CUR_USER=$(whoami)
 
 # LSB implimentation is not standard across distros, so we have to roll our own...
 killproc() {
-    kill $(cat $PID_FILE) >/dev/null
+    eval "$@" >/dev/null
 }
 
 start_daemon() {
@@ -96,7 +95,7 @@ stop_go_agent() {
     check_proc
 
     if [ $? -eq 0 ]; then
-        killproc
+        killproc /usr/share/${SERVICE_NAME}/stop-agent.sh ${SERVICE_NAME} service_mode
 
         # Make sure it's dead before we return
         until [ $? -ne 0 ]; do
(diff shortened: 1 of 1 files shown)


installers/go-agent/release/go-agent.init
```diff
@@ -96,7 +96,7 @@ stop_go_agent() {
     check_proc
 
     if [ $? -eq 0 ]; then
-        killproc -p $PID_FILE /usr/share/${SERVICE_NAME}/agent.sh >/dev/null
+        killproc
```
ketan: Would you not pass an executable to killproc? Am I missing something?

arvindsv: `killproc` is not the standard one. It's just a function in this file.

ibnc:
@ketan @arvindsv any idea why we weren't just using `stop-agent.sh` to stop agents? It looks like it does everything we want (including fixing the issue of pid location mentioned in the issue)

installers/go-agent/release/go-agent.init
```diff
@@ -96,7 +95,7 @@ stop_go_agent() {
     check_proc
 
     if [ $? -eq 0 ]; then
-        killproc -p $PID_FILE /usr/share/${SERVICE_NAME}/agent.sh >/dev/null
+        killproc /usr/share/${SERVICE_NAME}/stop-agent.sh ${SERVICE_NAME} service_mode
```
marques-work: My vote would be to just explicitly invoke `stop-agent.sh` and remove `killproc()` altogether since it's now a useless use of `eval`.

Suggesting to change this line to:

```
>/dev/null /usr/share/${SERVICE_NAME}/stop-agent.sh ${SERVICE_NAME} service_mode
```

marques-work: My vote would be to just explicitly invoke `stop-agent.sh` and remove `killproc()` altogether since it's now a useless use of `eval`.

Suggesting to change this line to:

```
>/dev/null /usr/share/${SERVICE_NAME}/stop-agent.sh ${SERVICE_NAME} service_mode
```

arvindsv: @ibnc No, I'm not sure why. @marques-work's suggestion makes sense to me. I wouldn't do `>/dev/null` though. My preference would be to leave it be a little verbose, especially if it fails to kill. If, on failure to kill, it shows a message on stderr, then `>/dev/null` should be fine.

marques-work: @arvindsv better to be paranoid and remove the `>/dev/null` :)

ketan:
Can we change the cat `$PID_FILE | xargs kill` to instead `kill $(cat $PID_FILE)`. Reduces the need for having xargs as a dependency, because containers :)

ibnc:
> one see if that's allright?

@ketan I see no issues with the change. It works as it should from what I can see

PR #5900, opened 2019-02-25
User management page UI update

vrushaliwaykole:
- Updated table view
- Added icons to display status of  make / remove superadmin operation.
- Make / Remove superadmin previlege works only for explicitely added users and not for group admin access. So restricted the same on UI.
- Specs for newly added code


api/api-users-v3/src/main/java/com/thoughtworks/go/apiv3/users/representers/UserRepresenter.java
```diff
@@ -33,6 +33,7 @@ public static void represent(OutputWriter childWriter, UserToRepresent user) {
                 .add("email", user.getEmail())
                 .add("email_me", user.isEmailMe())
                 .add("is_admin", user.isAdmin())
+                .add("is_individual_admin", user.isIndividualAdmin())
```
GaneshSPatil: @vrushaliwaykole -- From the implementation perspective of the SPA, it might make sense to add this field to the user response, but not sure if from the Users API consumer perspective if it adds any value.

A consumer of the API will want to know whether the user is an admin or not, regardless whether the user is admin directly or via a role.

---

To check if the user is an admin directly or via a role, we can rely upon the `super admin`'s  API at the client (browser) side.


##### Implementation details:

On the page load, we can make a call to users API and roles API and then construct the Users Model based on the responses of both.

server/webapp/WEB-INF/rails/webpack/models/users/users.ts
```diff
@@ -108,6 +118,45 @@ export class User {
...
+
+  markUpdateUnsuccessful() {
+    this.updateOperationStatus(UpdateOperationStatus.ERROR);
+  }
+
+  clearUpdateStatus() {
+    this.updateOperationStatus(null);
+  }
+
+  updateFromJSON(json: UserJSON) {
```
ketan: Lets not mutate user objects, replacing is a better option. If a new attribute is added to a user, then someone needs to remember to update this method as well.

server/webapp/WEB-INF/rails/webpack/models/users/users.ts
```diff
@@ -64,16 +69,20 @@ export class User {
   enabled: Stream<boolean>;
   checkinAliases: Stream<string[]>;
   roles: Stream<UserRoleJSON[]>;
+  updateOperationStatus: Stream<UpdateOperationStatus | null> = stream();
```
ketan: This seems like a view state to me. Perhaps store this in a view model, or directly on the individual component if it works better.

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.scss
```diff
@@ -105,3 +105,9 @@ input:checked ~ .switch-paddle:after {
     transform:         translateX($switch-small-translate-x);
   }
 }
+
+
+.disabled{
```
GaneshSPatil: indent?

server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx
```diff
@@ -54,7 +54,9 @@ export class SwitchBtn extends MithrilComponent<Attrs, State> {
     const classNames = classnames({[styles.switchSmall]: isSmall}, styles.switchBtn);
     const switchId   = `switch-${uuid4()}`;
     const label      = vnode.attrs.label ? (
-      <label className={styles.switchLabel} data-test-id="switch-label">{vnode.attrs.label}</label>) : null;
+      <label
```
GaneshSPatil: indent?

server/webapp/WEB-INF/rails/webpack/views/global/_variables.scss
```diff
@@ -104,3 +104,5 @@ $btn-text-color: #fff;
 $btn-reset: #666;
 $btn-danger: $go-danger;
 
+// tables
+$selected-table-row: #e2f7fa;
```
GaneshSPatil: where are other styles for the table?
How do we know what is the table row color? and upon selection this color will go well with the other unselected table rows?

server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
```diff
@@ -240,18 +273,42 @@ export class UsersPage extends Page<null, State> {
...
+                                            json: BulkUpdateSystemAdminJSON,
+                                            user: User): void {
     AdminsCRUD.bulkUpdate(json)
               .then((apiResult) => {
                 apiResult.do((successResponse) => {
+                  vnode.state.userViewStates[user.loginName()] = {
+                    updateOperationStatus: UpdateOperationStatus.SUCCESS
+                  };
+
+                  setTimeout(() => {
```
GaneshSPatil: can we remember the last timeout, and clear it if required before creating the new timeout?

This is required if the user performs multiple updates within `CLEAR_USER_UPDATE_STATUS * 1000` interval.

server/webapp/WEB-INF/rails/webpack/views/global/_variables.scss
vrushaliwaykole: The other table styles are in the table component. However, we are not using the table component for users page. So added variable in default variables.

GaneshSPatil:
Also,
Not sure if the following error styles are as per the design

screen shot 2019-02-27 at 5 33 29 pm

PR #5905, opened 2019-02-25
Move SPAs to freemarker templates

ketan:


ketan:
Closing in favor of #5920.

Issue #5899, opened 2019-02-25
Dashboard Pipeline Paused at sometimes shows "Invalid Date"


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Some of the dashboard pipelines show paused-at time as "Invalid Date"

###### Basic environment details

* Go Version: GoCD Version: 19.2.0 (8584-d9d3c0aa402137988ded4e00595162c678f9a8bd).

###### Additional Environment Details

Look for following pipelines on build.gocd.org (Before they are unpaused :) ):
* go-cd-contrib-bhkwan
* Performance-with-h2
* vishal-test
* akshay-test

GaneshSPatil:
For anyone who is looking to contribute to GoCD, this is a good first issue to start off with :)

#### Context around the issue:

GoCD started [storing the time at which pipeline was paused](https://github.com/gocd/gocd/pull/5755) from release `19.2.0`.

Let's say, the currently installed GoCD version is `19.1.0`, where we pause a pipeline. In `19.1.0` GoCD did not have the capability to store the paused_at time.
Later, we upgrade our GoCD instance to `19.2.0`. As the dashboard will try rendering the paused-at time for paused pipelines. 
The pipelines which were paused prior to `19.2.0`, will have paused-at time `null`. Converting null to Date will result in `Invalid Date`.

#### How to fix:

Modify the pipeline pause message to not include paused at, when time not available.

So if paused-at time available, the output should be:
```
Paused by SanjanaB (Pausing it currently. Will resume later) on 22 Feb, 2019 at 15:45:09 Local Time
```

And, if the paused-at time is not available, the output should be:
```
Paused by SanjanaB (Pausing it currently. Will resume later)
```

Change [this](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/webpack/views/dashboard/pipeline_operations_widget.js.msx#L194-L199) section of the code to optionally include paused-at time.

rajiesh:
Verified on 19.2.0 (8610-4713200bc880ce17014d826e20a30ed032e3ac26).

PR #5906, opened 2019-02-25
Dashboard - pipeline pause message fix

kritika-singh3:
In case of a paused pipeline, if the pause time is not present - it won't get rendered


maheshp:
Fixes - https://github.com/gocd/gocd/issues/5899

PR #5909, opened 2019-02-26
[Do Not Merge] Do not use helpers to build documentation URLs in ruby

GaneshSPatil:
Fix GoCD performance issue (v19.2.0)

As part of commit acfab936a802b5a3dbe3594b29fc3011c12f1bb3, we introduced
'docsUrl()' and 'apiDocsUrl()' static helper methods in Java to generate
appropriate urls based on the current GoCD version.

These methods were called from the ruby code as well (for some of our APIs
and erb pages). JRuby does not have the concept of static methods. When a
static method from Java is invoked, JRuby internally defines a new class
in metaspace with the static method definition and then creates a new object
of the newly defined class and accesses it.
Upon method invocation, these classes are freed and can be GCed.

In GoCD's context, for each 'docsUrl()' and 'apiDocsUrl()' method invocation,
a new class was defined in heap and thrown away immediately.
Each GoCD API entity has at least 3 links (self, find and doc), Invoking rails
based GoCD APIs and ERB based UI Pages was causing performance issues as on
each request metaspace was bloating up because of dynamic prototype classes
creation for static method references.

As of now, i have reverted the rails references to these static methods.
Which has brought in the following inconsistencies:
1. Rails based APIs will return '/current' in the docs urls instead of the
   current GoCD version, whereas Spark based API urls will return the current
   GoCD version.
2. If we're referring to the docs pages from old rails UI pages, it will point
   to '/current' instead of the current GoCD version.

**NOTE:** These (above mentioned) issues has always been there with GoCD.
With acfab936a802b5a3dbe3594b29fc3011c12f1bb3 commit, we fixed it everywhere,
but because of perf issues, the current commit reverts the fix for rails based
APIs and Pages.


GaneshSPatil:
Do not merge this until we get a green signal from perf-run.

ketan:
> These methods were called from the ruby code as well (for some of our APIs
> and erb pages). JRuby does not have the concept of static methods. When a
> static method from Java is invoked, JRuby internally defines a new class
> in metaspace with the static method definition and then creates a new object
> of the newly defined class and accesses it.
> Upon method invocation, these classes are freed and can be GCed.

This is news to me. Where can I read more about this to understand it better?

ketan:
Superseeded by #5916

maheshp:
@GaneshSPatil closing this PR as this issue is fixed by https://github.com/gocd/gocd/pull/5916

PR #5910, opened 2019-02-26
Attempt to fix flaky Agents Spec

GaneshSPatil:


GaneshSPatil:
@akshaydewan

#### Seed used
`88276`

#### Issue with the test
The test was expecting the widget to be mounted to the test element, but for some reason one of the tests explicitly removes it and then again it doesn't get mounted (from beforeEach)

#### Fix
Explicitly mount the widget for the flaky test.

**NOTE:** I haven't spent time analyzing why it is not getting mounted. As of now, it has fixed the issue for the given seed (and other random seeds too), if required I can dig deeper to see the problem and right fix for it.

PR #5911, opened 2019-02-27
Reparse failed config-repos on config update

marques-work:
- Reverts changes to ConfigReposMaterialParseResultManager
    from https://github.com/gocd/gocd/pull/5843
  - This should resolve the issue where a config-repo is not
    reparsed after fixing a missing upstream pipeline (or similar
    missing dependency)
  - More reliable than consulting ServerHealthStates to determine
    when or whether to reparse
  - Also resolves issue noted in https://github.com/gocd/gocd/issues/5792#issuecomment-465845347

This essentially adds a ConfigChangedListener that clears the last parsed modification of a PartialConfigParseResult on config update. This allows the config repo to be reparsed.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/ConfigReposMaterialParseResultManager.java
+++ b/server/src/main/java/com/thoughtworks/go/config/ConfigReposMaterialParseResultManager.java
@@ -16,13 +16,10 @@
 
 package com.thoughtworks.go.config;
 
-/*  @description
- * Keeps track all config repo parsing states
- */
-
-
 import com.thoughtworks.go.config.remote.PartialConfig;
+import com.thoughtworks.go.domain.materials.MaterialConfig;
 import com.thoughtworks.go.domain.materials.Modification;
+import com.thoughtworks.go.listener.ConfigChangedListener;
 import com.thoughtworks.go.server.service.ConfigRepoService;
 import com.thoughtworks.go.serverhealth.HealthStateScope;
 import com.thoughtworks.go.serverhealth.ServerHealthService;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/thoughtworks/go/config/ConfigReposMaterialParseResultManager.java
```diff
@@ -33,81 +30,90 @@
...
-        return null;
+    /**
+     * After a successful update of the server config, we should reparse any failed config-repos in case
+     * any upstream dependency issues have been resolved.
+     *
+     * @return a listener that clears the last parse states and errors related to config-repos so as to allow reparsing
+     */
+    ConfigChangedListener onConfigUpdateDoReparse() {
+        return newCruiseConfig -> {
+            for (Map.Entry<String, PartialConfigParseResult> entry : fingerprintOfPartialToParseResultMap.entrySet()) {
```
ketan: This seems to clear the map on every config save, even ones where the config update does not involve changes to pipelines/config repos/scm materials. Is this intentional?

Can we possibly also use the `EntityConfigChangedListener` to improve this a bit so that config changes to pipelines/config repos/scm materials will selectively update just the result that needs updating?

The current code will have to remain as-is for handling a full config save when someone edits the XML.

marques-work: I was not confident that I knew all the ways merging a partial config could fail; missing template, pipeline, materials but might be others? This I made the criteria very broad.

I suppose at scale this could have performance implications that I do not know yet.

Given the different ways a merge can fail dependency checks, is there a good way to find which partialconfig result matches a config entity in order to selectively update affected results? Other than parsing the exception message, that is.

server/src/main/java/com/thoughtworks/go/config/ConfigReposMaterialParseResultManager.java
```diff
@@ -33,81 +34,152 @@
...
-        HealthStateScope healthStateScope = HealthStateScope.forMaterialConfig(configRepoService.findByFingerprint(fingerprint).getMaterialConfig());
-        List<ServerHealthState> serverHealthStates = serverHealthService.filterByScope(healthStateScope);
-        if (!serverHealthStates.isEmpty()) {
-            return PartialConfigParseResult.parseFailed(null, represent(serverHealthStates.get(0)));
+    /**
+     * Registers the various ConfigChangedListener instances to mark config-repos for reparse on config updates.
+     *
+     * @param configService the {@link GoConfigService} to register events
+     */
+    void attachConfigUpdateListeners(GoConfigService configService) {
```
ketan: @bdpiparva told me there's a better way to do this. Since inject multiple listener instances, we'll end up with multiple callbacks, once for the entity update and another for the full save.

He'll push a fix, and point you to the right way to do this (for next time you implement a listener for multiple entities :))

bdpiparva: Commit b30a4b3513f539261c980272ab828b45be0659e1 is to combine all entity change listeners into one.

marques-work: @ketan @bdpiparva awesome, thanks!

PR #5916, opened 2019-02-27
Memoize ruby invocations for com.thoughtworks.go.CurrentGoCDVersion.

ketan:
Introduces a wrapper ruby class that wraps around
`com.thoughtworks.go.CurrentGoCDVersion` and memoizes the value for
further usage.


ketan:
@rajiesh confirmed that the fix worked. Merging.

PR #5920, opened 2019-02-27
Migrate login page to freemarker

ketan:
Merge after #5905


ketan:
I'm attaching some screenshots that @naveenbhaskar sent across for a few login possibilities as initial points for discussing improvements to the current login screen:

---

### When there are auth configs that use password plugins and web based plugins

image

---

### When there are only auth configs that use password plugins, no auth configs corresponding to web based plugins

image

---

### When there are auth configs that only use web based plugins, no password plugins

image

--- 

### When there is exactly one auth config that uses a web based plugin

image

arvindsv:
@ketan @naveenbhaskar If we're thinking about the user experience for this, it might be useful to think about a case where there is a default login mechanism. For instance, there could be the possibility of setting GitHub as the "default" and it'll show something similar to the fourth screenshot above. Going specifically to `/login` could leave you at the login screen.

The reason is: We always recommend that instances should be setup with password auth as a backup. So, not enough instances will see the fourth screenshot, which is a shame.

rajiesh:
@ketan the sahi tests are fixed. We have this PR for the webdriver tests - https://github.com/gocd/ruby-functional-tests/pull/345

PR #5921, opened 2019-02-27
Disable access token creation if plugin doesn't support it.

bdpiparva:
screen shot 2019-02-27 at 6 01 38 pm

### DIFF
--- a/spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/AccessTokensController.java
+++ b/spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/AccessTokensController.java
@@ -1,20 +1,26 @@
 /*
-* Copyright 2019 ThoughtWorks, Inc.
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-*     http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


ketan:
Would you also not change the API endpoint/controller to disallow creation?

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/AccessTokensController.java
```diff
@@ -1,55 +1,86 @@
...
+        final AuthenticationToken<?> authenticationToken = SessionUtils.getAuthenticationToken(request.raw());
+        final Map<String, Object> meta = new HashMap<>();
+        meta.put("pluginId", authenticationToken.getPluginId());
+        meta.put("supportsAccessToken", supportsAccessToken(authenticationToken));
+        return meta;
+    }
+
+    private boolean supportsAccessToken(AuthenticationToken<?> authenticationToken) {
+        final SecurityAuthConfig profile = securityAuthConfigService.findProfile(authenticationToken.getAuthConfigId());
+        return authorizationExtensionCacheService.isValidUser(authenticationToken.getPluginId(), authenticationToken.getUser().getUsername(), profile);
```
GaneshSPatil: An FYI edge case

A Github user logs into GoCD and has a session. (valid for 30 mins)
During this time, the user is removed from the Github team (by an admin),  When the user lands on the access token SPA, it will say the plugin `github` does not support access token creation, where the reality is:
The plugin supports access token creation but the user is invalid.

bdpiparva: @GaneshSPatil - User will not get that message as long as we have cached user data, beyond that yes user, will get that message. This implementation is for the current release only. Going forward we will have a specific capability.

PR #5930, opened 2019-03-01
Remove Elastic Agent Extension v3

GaneshSPatil:


GaneshSPatil:
@arvindsv -- can we decide upon removal of plugin extensions?

Depending upon whether we choose to remove the extensions or to keep them, we can either merge this PR or close it.

GaneshSPatil:
@arvindsv and @bdpiparva --
We can support the latest few versions of all the extensions.

In this particular scenario,
Elastic agent v4 extension was introduced for `job-completion` call. So, plugins implementing `v3` extension just needs to change the supported version to `v4` and the plugin will continue to work without immediate agent cleanup on job completion.

Also, as part of the upcoming release, we will introduce EA extension v5 to support the multi-cluster feature.

Hence, I think v3 can be removed in this release.

PR #5934, opened 2019-03-01
Simplified Pipeline Creation Flow SPA

marques-work:
We're going to build a new UI flow to create pipelines to support improving the trial experience. We'll be working in this branch.

This and related sub-items are part of this Trello card: https://trello.com/c/5QDrNrdc


base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java
```diff
@@ -219,6 +219,7 @@
     public static GoIntSystemProperty CONFIG_MATERIAL_POST_UPDATE_LISTENERS = new GoIntSystemProperty("config.material.post.update.threads", 2);
 
     public static GoSystemProperty<String> GO_SERVER_MODE = new GoStringSystemProperty("go.server.mode", "production");
+    public static GoSystemProperty<Boolean> GO_TEST_DRIVE = new GoBooleanSystemProperty("go.server.test.drive", false);
```
marques-work: @ibnc could we use the feature toggle mechanism instead of java props? I'd like to keep things consistent.

Specifically, I'm referring to the `FeatureToggles` class. It would entail you adding an entry to `server/out/production/resources/available.toggles` so that we can override these values with those set in `$configDir/go.feature.toggles`.

Then to test for the toggle in code, one can do something like: `FeatureToggleService.isToggleOn("toggle-name")`

PR #5941, opened 2019-03-04
Move `/go/cctray.xml` to spark

ketan:


akshaydewan:
The code LGTM. Did we have any caching in place with Rails? Because the response returned an ETag header previously, and in Spark it doesn't. I tried sending the If-None-Match header with Rails, but didn't get back a 304, so I'm not sure.

ketan:
Rails default etag is a digest of the response content. I will add some
etag implementation tomorrow.

ketan:
I've added the etag implementation and merge it in. When you have time, can you take a look at the last commit?

server/src/main/java/com/thoughtworks/go/server/service/CcTrayService.java
```diff
@@ -38,24 +42,43 @@ public CcTrayService(CcTrayCache ccTrayCache, GoConfigService goConfigService) {
...
+
+    public Appendable renderCCTrayXML(String siteUrlPrefix, String userName, Appendable appendable, Consumer<String> etagConsumer) {
+        boolean isSecurityEnabled = goConfigService.isSecurityEnabled();
         List<ProjectStatus> statuses = ccTrayCache.allEntriesInOrder();
 
-        StringBuilder xml = new StringBuilder("<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<Projects>\n");
-        for (ProjectStatus status : statuses) {
-            if (!isSecurityEnabled || status.canBeViewedBy(userName)) {
+        String hashCodes = statuses.stream().map(ProjectStatus::hashCode).map(Object::toString).collect(Collectors.joining("/"));
+        String etag = DigestUtils.sha256Hex(ProjectStatus.SITE_URL_PREFIX + "/" + hashCodes);
```
akshaydewan: Did you mean to use `siteUrlPrefix` instead of `ProjectStatus.SITE_URL_PREFIX`?

server/src/test-fast/java/com/thoughtworks/go/server/service/CcTrayServiceTest.java
```diff
@@ -70,11 +71,11 @@ public void shouldGenerateCcTrayXMLForCurrentUserWhenSecurityIsEnabled() throws
         when(ccTrayCache.allEntriesInOrder()).thenReturn(asList(statusFor("proj1", "user1"), statusFor("proj2", "user2")));
 
         loginAs("USER1");
-        String xml = ccTrayService.getCcTrayXml("some-prefix");
+        String xml = ccTrayService.renderCCTrayXML("some-prefix", "USER1", new StringBuilder(), etag -> {}).toString();
         assertCcTrayXmlFor(xml, "some-prefix", "proj1");
 
         loginAs("uSEr2");
-        xml = ccTrayService.getCcTrayXml("some-prefix");
+        xml = ccTrayService.renderCCTrayXML("some-prefix", "uSEr2", new StringBuilder(), etag -> {}).toString();
```
akshaydewan: Might be a good idea to add an assertion for the etag as well.

PR #5942, opened 2019-03-04
Fix for stop agent infinite loop with docker agent edge case

ibnc:
This PR fixes the second issue raised in #5858 by @adityasood 

I've tested this locally with a ubuntu vm with an ubuntu docker image inside with the agent installed manually. I was able to reproduce the issue, and my changes fixed the issue.

### DIFF
--- a/installers/go-agent/release/go-agent.init
+++ b/installers/go-agent/release/go-agent.init
@@ -31,6 +31,8 @@ SERVICE_NAME=${0##*/}
 # Strips initV leading chars if the script is located in rcX.d (automatically executed on runlevel change)
 INIT_DIR=${0%/*}; echo ${INIT_DIR##*/} | grep -e '^rc[0-9]\.d$' >/dev/null && SERVICE_NAME=$(echo "${SERVICE_NAME}" | sed 's/^[SK][0-9][0-9]//');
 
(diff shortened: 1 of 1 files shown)


rajiesh:
@ibnc Tested the fix on latest master it fixes the issue @adityasood had raised.

An observation, on a fresh installation of go-agent when user start agent like `/etc/init.d/go-agent start` below message is thrown. Think this is because of the `check_proc` is called once before agent startup [here](https://github.com/ibnc/gocd/blob/0c860c08534a39edd681ff38165d32b1976e153e/installers/go-agent/release/go-agent.init#L67) and pid file is not present

```
# /etc/init.d/go-agent start
cat: /var/run/go-agent/go-agent.pid: No such file or directory
error: list of process IDs must follow -p

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Started Go Agent.
```

arvindsv:
I'm also very suspicious of `ps`. It varies across installations. Is `pgrep -F` an option?

ibnc:
> I'm also very suspicious of `ps`. It varies across installations. Is `pgrep -F` an option?

The `-F` isn't available for all versions of `pgrep` 

For example, It the `-F` option isn't available on centos 6 I believe

PR #5952, opened 2019-03-07
Remove cctray from rails

ketan:


ketan:
To be merged for 19.4 (or later in 19.3 if @rajiesh or @adityasood can confirm that the new cctray is good)

PR #5953, opened 2019-03-07
Add previous and next version to version.json

GaneshSPatil:
```console
$ cat installers/target/distributions/meta/version.json
{
    "go_version": "19.3.0",
    "previous_go_version": "19.2.0",
    "next_go_version": "19.4.0",
    "go_build_number": "8651",
    "go_full_version": "19.3.0-8651",
    "git_sha": "6d6ea20b4093156424d513793246ef6ea233486b",
    "pipeline_name": null,
    "pipeline_counter": null,
    "pipeline_label": null,
    "stage_name": null,
    "stage_counter": null,
    "job_name": null
}
```


GaneshSPatil:
This information is required for the automating downstream pipelines. Many of the post-release pipelines are manual and requires human intervention for passing these extra environment variables.

Also, the previous-version, next-version means different for different pipelines, with this change, there will be a consistency in these variable references.

\cc: @bdpiparva @ketan

PR #5955, opened 2019-03-07
Ensure that we specify mimetype on all routes

ketan:


ketan:
Do not merge. Will add tests, just wanted folks to see this.

Issue #5856, opened 2019-02-14
Update how GoCD fetches git materials


arvindsv:
##### Issue Type

- Feature enhancement

##### Summary

It's been a while. `git` has changed over time and provides more features. This is a placeholder to take a look at the git material and see what can be changed. For instance, there's usage of `git submodule foreach` and reading the `.gitconfig`, etc. which might be replaced with `git submodule sync`.

When doing this, we should also consider other issues brought up around git, such as:
  - [x] Shallow clone doesn't apply to submodules: https://github.com/gocd/gocd/issues/2123
  - [ ] ~~Use the `--jobs` flag when updating submodules to parallelize the operation~~
  - [ ] Maybe look at whether `GIT_SSH_COMMAND` can be incorporated?
  - [x] Possibly do a `git clean -dffx`? [Might be a breaking change. Be careful]
  - [x] GoCD fails to update git submodules: https://github.com/gocd/gocd/issues/3635
  - [ ] ~~Skip git gc, resetting, cleaning if clean working directory is used: https://github.com/gocd/gocd/issues/2124~~
  - [x] Check connection should check if branch name is valid: https://github.com/gocd/gocd/issues/2093
  - [x] Add git material fails, with color options in the .gitconfig file: https://github.com/gocd/gocd/issues/326

Issue opened after a conversation with @GaneshSPatil.

ibnc:
@kierarad or @arvindsv would either of you be up for giving this PR one last review, and then merging if everything looks good?

akshaydewan:
Here's the current process that happens when we reset the working directory before running a job: (`GitCommand.resetWorkingDir`)

``

akshaydewan:
Another possible optimization that we can do while implementing **Shallow clone doesn't apply to submodules** - if we run `git clone` with additional args `--recurse-submodules` and `--shallow-submodules`, then we can potentially save time by not resetting the working directory as described in #2124 , because re-initializing submodules would be immediately after cloning is not needed and takes up time.

arvindsv:
I wonder if it makes sense to say that if you use submodules, you'll need version X. If there is no `.gitmodules`, maybe we don't need to run those commands at all?

It would be nice to not force a new version unnecessarily. Having said all that 2.13 is already considered "old". So, it wouldn't be wrong for us to drop support for others.

akshaydewan:
Here's what I found in Git's release notes:

1.7.4
 * "git submodule sync" updates metainformation for all submodules,
[...]

PR #5964, opened 2019-03-08
Update the `resetWorkingDirectory` process

akshaydewan:
Issue #5856
- Use submodule deinit instead of manually removing submodules from .git/config
- Update few other commands using newer submodule command arguments like --recursive and --foreach
- Do a git clean -dffx instead of -dff

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/domain/materials/git/GitCommandTest.java
+++ b/common/src/test/java/com/thoughtworks/go/domain/materials/git/GitCommandTest.java
@@ -25,7 +25,6 @@ import com.thoughtworks.go.domain.materials.mercurial.StringRevision;
 import com.thoughtworks.go.helper.GitSubmoduleRepos;
 import com.thoughtworks.go.helper.TestRepo;
 import com.thoughtworks.go.mail.SysOutStreamConsumer;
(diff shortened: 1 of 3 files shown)


ketan:
Not so sure about the change to git clean. There was a long conversation about it #1059

akshaydewan:
Based on that thread, I can add an [agent side toggle](https://github.com/gocd/gocd/pull/5666) for reverting to the old behaviour for git clean.

akshaydewan:
Currently the linux build is failing with the error log showing:
```
org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
	Error performing command: --- Command ---
git submodule deinit -f --all
--- Environment ---
{}
--- INPUT ----


--- EXIT CODE (1) ---
--- STANDARD OUT ---

--- STANDARD ERR ---
STDERR: You need to run this command from the toplevel of the working tree.
---
```

I'm not sure if there's some parent working directory that is under version control, or if the path is somehow wrong when running the test on CI.

Moreover, the git version on CentOS 6 is `1.7.1`, which doesn't support `submodule deinit`. We will also need to update git to at least 1.9 on the build agents.

ketan:
Tests run in a temporary directory which is under `project-name/tmp`

Issue #5956, opened 2019-03-07
Functional Issues with Spark based Environments API v2


GaneshSPatil:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: `GoCD Version: 19.2.0 (8641-db17f5840f9d1c6838add93af050b812bbdf5c68).`
* JAVA Version: `1.8`

###### Bugs:
  - [x] Secure environment variables configured for the environment are returned as `******` instead of returning `encrypted_value`. (Fixed as part of #5957)
 - [x] API should accept both `value` and `encrypted_value` in the environment variables. Details [here](https://github.com/gocd/gocd/issues/5956#issuecomment-470489060) (Fixed with https://github.com/gocd/gocd/pull/5967)

rajiesh:
API does not accept `encrypted_value` as a key in Environments variables JSON, it always looks for `value`. So create/update request like this 
```
{
    "name": "from-api",
    "pipelines": [
        {
            "name": "pipeline-in-environment543d8f06"
        }
    ],
    "environment_variables": [
        {
            "name": "secured",
            "encrypted_value": "AES:Nla+0y+uOInW4ZVsHzLZBg==:qP+V9XpV53UvzgcieGWpKyEiP5uHudFuArTsnETozo8=",
            "secure": true
        }
    ]
}
```
will fail [here](https://github.com/gocd/gocd/blob/master/api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentRepresenter.java#L64) with NullPointerException since there is no `value` in the request

PR #5967, opened 2019-03-11
Environments api should accept both `encrypted_value` and `value` 

vishaldevgire:
- Allow `create/update/patch` APIs to accept both `encrypted_value` and `value` in JSON request
- Added a validation to check for existence of either `encrypted_value` or `value`
- While serializing errors in environment variable configuration are rendered


api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
```diff
@@ -16,27 +16,62 @@
...
+            });
+        });
+        return variables;
+    }
+
+    public static EnvironmentVariableConfig fromJSON(JsonReader jsonReader) {
+        String name = jsonReader.getString("name");
+        Boolean secure = jsonReader.optBoolean("secure").orElse(false);
+
+        if (!jsonReader.optString("value").isPresent()
```
bdpiparva: Read only once and store it in a variable

api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
```diff
@@ -16,27 +16,62 @@
...
+import com.thoughtworks.go.apiv2.environments.exception.InvalidGoCipherTextRuntimeException;
 import com.thoughtworks.go.config.EnvironmentVariableConfig;
-import com.thoughtworks.go.security.GoCipher;
+import com.thoughtworks.go.config.EnvironmentVariablesConfig;
+import com.thoughtworks.go.security.CryptoException;
+
+import java.util.HashMap;
 
 
 public class EnvironmentVariableRepresenter {
```
GaneshSPatil: can this be moved to `shared-v2`?

vishaldevgire: It can be, I will make that change.

api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
```diff
@@ -16,27 +16,62 @@
...
+        return variables;
+    }
+
+    public static EnvironmentVariableConfig fromJSON(JsonReader jsonReader) {
+        String name = jsonReader.getString("name");
+        Boolean secure = jsonReader.optBoolean("secure").orElse(false);
+
+        if (!jsonReader.optString("value").isPresent()
+                && !jsonReader.optString("encrypted_value").isPresent()) {
+            HaltApiResponses.haltBecauseOfReason("Environment variable must contain either 'value' or 'encrypted_value'");
```
bdpiparva: What should be the behaviour when`encrypted_value` and `values` both are specified?

vishaldevgire: Its covered in the call to `environmentVariableConfig.deserialize(name, value, secure, encryptedValue);`, this line adds an error stating both the values in one request are not allowed. The request fails with a 422 saying the same.

api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
vishaldevgire: Moved.

api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
vishaldevgire: Done

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
```diff
@@ -0,0 +1,86 @@
...
+        return variables;
+    }
+
+    public static EnvironmentVariableConfig fromJSON(JsonReader jsonReader) {
+        String name = jsonReader.getString("name");
+        Boolean secure = jsonReader.optBoolean("secure").orElse(false);
+        Optional<String> optValue = jsonReader.optString("value");
+        Optional<String> optEncValue = jsonReader.optString("encrypted_value");
+
+        if (!optValue.isPresent() && !optEncValue.isPresent()) {
```
varshavaradarajan: Why perform this check when `environmentVariableConfig.deserialize` handles it? https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L302

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
```diff
@@ -0,0 +1,86 @@
...
+            HaltApiResponses.haltBecauseOfReason("Environment variable must contain either 'value' or 'encrypted_value'");
+        }
+
+        String value = secure ? optValue.orElse(null) : jsonReader.getString("value");
```
varshavaradarajan: Doesn't matter if it is secure or not at this point, deserialize method checks for it, all that is needed to be done is pass the string for "value" if present, else null. Why check for secure?

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
vishaldevgire: `if (!optValue.isPresent() && !optEncValue.isPresent())` is testing if none of the two(encrypted_value, value) are provided, while the test [environmentVariableConfig.deserialize](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L302) is testing if both values are provided. These tests are checking the opposite.

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
vishaldevgire: You are right this check is there in deserialize, I will remove it from here.

vishaldevgire: I think this check will have to be there, because if incoming request is something like this 
```
{
  "secure": false,
  "encrypted_value": "..."
}
```
then it should fail with 422 (`jsonReader.getString("value")`), because a `secure: false` request  should not have `encrypted_value`, it should have `value`.  This is similar to what we are doing in other places, when we are de-serializing environment variable.

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
varshavaradarajan: Its okay to not provide either. There are instances on build.go.cd where just the name exists and no value, so that it can be overridden at the time of scheduling the pipeline - https://build.gocd.org/go/admin/pipelines/PublishStableRelease/environment_variables. I don't remember what happens if `{"name": "foo", "secure": true}` is provided. Maybe that's something to check and handle in `environmentVariableConfig.deserialize`.

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
varshavaradarajan: This is handled by https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L298

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
varshavaradarajan: I think for both pipeline config and environments, environment variable representer should be consistent. And, providing both value=nil and encrypted_value=nil (through API or XML) is not accepted currently (results in https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L198 when saved via XML, and results in https://github.com/gocd/gocd/blob/master/api/api-base/src/main/java/com/thoughtworks/go/api/util/HaltApiMessages.java#L93 via API, for pipelines). This check is useful. However, if you're doing it here, then the apiv6 EnvironmentVariableRepresenter must reflect the same. Its better to push this check to the EnvironmentVariableConfig#deserialize method then, don't you think?

maheshp:
Fixes https://github.com/gocd/gocd/issues/5956

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
vishaldevgire: > This is handled by https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java#L298 ?

This code is adds it as one of the errors but doesn't cause the request to fail with 422. 

I am keeping following line as the request should fail with 422.

`String value = secure ? optValue.orElse(null) : jsonReader.getString("value");`

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/representers/EnvironmentVariableRepresenter.java
vishaldevgire: Currently, all APIs accepting environment variables in request body fail with 500 when you don't provide any of the two (`encrypted_value` or `value`) properties, I agree that the fix to this should be consistent across all APIs. I have created an [issue](https://github.com/gocd/gocd/issues/6017) to target this. 

For this PR, I think we following check:

`if (!optValue.isPresent() && !optEncValue.isPresent()) `

api/api-shared-v2/src/main/java/com/thoughtworks/go/apiv2/shared/exception/InvalidGoCipherTextRuntimeException.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.apiv2.shared.exception;
+
+public class InvalidGoCipherTextRuntimeException extends RuntimeException {
```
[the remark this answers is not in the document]
vishaldevgire: Done.

PR #5972, opened 2019-03-12
Environment api accepts both value and encrypted_value for env vars

ibnc:
see #5956 bug number two. @rajiesh provides more details [here](https://github.com/gocd/gocd/issues/5956#issuecomment-470489060)

### DIFF
--- a/api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
+++ b/api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
@@ -19,6 +19,7 @@ package com.thoughtworks.go.apiv2.environments.representers;
 import com.google.gson.JsonObject;
 import com.thoughtworks.go.api.base.OutputWriter;
 import com.thoughtworks.go.config.EnvironmentVariableConfig;
+import com.thoughtworks.go.config.VariableValueConfig;
 import com.thoughtworks.go.security.GoCipher;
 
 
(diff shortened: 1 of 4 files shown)


api/api-environments-v2/src/main/java/com/thoughtworks/go/apiv2/environments/representers/EnvironmentVariableRepresenter.java
```diff
@@ -31,12 +32,16 @@ public static void toJSON(OutputWriter outputWriter, EnvironmentVariableConfig e
...
-
-        EnvironmentVariableConfig environmentConfig = new EnvironmentVariableConfig(new GoCipher(), name, value, secure);
-        return environmentConfig;
+        EnvironmentVariableConfig environmentVariableConfig = new EnvironmentVariableConfig();
+        environmentVariableConfig.setName(jsonReader.get("name").getAsString());
+        if (secure) {
+            environmentVariableConfig.setEncryptedValue(jsonReader.get("encrypted_value").getAsString());
+            environmentVariableConfig.setIsSecure(secure);
+        } else {
+            environmentVariableConfig.setValue(new VariableValueConfig(jsonReader.get("value").getAsString()));
```
varshavaradarajan: Can't someone provide:
```
{
"name": "e",
"secure": "true",
"value": "to-be-encrypted-by-go"
}
```
and this should return
```
{
"name": "e",
"secure": "true",
"encrypted_value": "AES:encrypted"
}
```
because its a pain to use the encryption API to get the encrypted value and then use it in the environments API? 

Also, isn't https://github.com/gocd/gocd/pull/5967 trying to do the same?

ibnc: @varshavaradarajan didn't see that there was already a PR for this. I'll close this PR

Issue #4241, opened 2018-01-25
Webhook support for hosted BitBucket server


ankitsri11:
##### Issue Type

- Bug Report

##### Summary

Currently, below is configurable for the BitBucket SaaS version (Bitbucket.org) but not for the hosted instance.

https://{webhookSecret}:x@gocd.example.com:8154/go/api/webhooks/bitbucket/notify

For hosted instance, GoCD expects (refer [this](https://github.com/gocd/gocd/blob/cba9fd72860ec1858d63ea27a8251275da6b1fcd/server/webapp/WEB-INF/rails.new/app/controllers/api/web_hooks/bit_bucket_controller.rb)) event-key, body with payload in a json format and user agent which is specific to bitbucket when a post request is received from Bitbucket. If any of the element is missing, you will see an error something like below.

###### Basic environment details

* Go Version: All versions

###### Workaround 

Use below curl command to notify request:

curl -H 'Content-Type: application/json' -H 'X-Event-Key: repo:push' -v --user SECRET:x -d @payload.json  -X POST -A 'Bitbucket-Webhooks/2.0' https://gocd.example.com:8154/go/api/webhooks/bitbucket/notify

Attached payload.txt file can be used as reference. 

[payload.txt](https://github.com/gocd/gocd/files/1665614/payload.txt)

renaldrozario:
Do you have webhooks in pipeline for hosted bitbucket?

joedai:
Do we have any roadmap items created to support hosted bitbucket in the future? If so will that be in a recent release?

bashazaid:
Any support for the Hosted Bitbucket server in latest GoCD versions?
We are using GoCD v20.9 and trying to configure webhooks for pipeline from hosted bitbucket server but the end points are not working.
Endpoints tried are :
https://user:passwd@gocd.example.com:8154/go/api/webhooks/bitbucket/notify/

https://user:passwd@gocd.example.com:8154/go/api/material/notify/bitbucket/?repository_url=repo.url.git

both are not working for me, giving the 400 error, not supported.

Any other ways we can do the webhook configuration with bitbucket server, I also tried with SCM and SCM plugin, not working for bitbucket server.

piyushsh:
Any update on this issue, we are also facing similar issue on our side. Any support would be really appreciated.

sriramnrn:
Per the documentation at https://api.gocd.org/current/#hosting-on-bitbucket

one should use https://{
[...]

PR #5978, opened 2019-03-13
Add webhook endpoint for hosted bitbucket servers (#4241)

sandro-h:
Hosted bitbucket server sends different webhook requests than the
cloud version.

Rather than trying to detect which type of server it is in the existing endpoint,
use a new endpoint 'webhooks/hosted_bitbucket/notify'.

Main differences to cloud bitbucket, based on hosted version 5.16.1:

- Finding the correct material url is different, because the
  hosted bitbucket server simply sends a list of clone urls
  that can be used directly (minus credentials). So there's no need
  to guess from a list of possible urls. This way we can also trivially
  support ssh urls.

- Hosted bitbucket sends a standard HMAC signature header using the webhook secret.
  No need to check basic auth headers.

- The request body is differently structured.

### DIFF
--- a/server/webapp/WEB-INF/rails/app/controllers/api/web_hooks/hosted_bit_bucket_controller.rb
+++ b/server/webapp/WEB-INF/rails/app/controllers/api/web_hooks/hosted_bit_bucket_controller.rb
@@ -1,0 +1,109 @@
+##########################################################################
+# Copyright 2017 ThoughtWorks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+##########################################################################
+
+module Api
+  module WebHooks
+    class HostedBitBucketController < WebHookController
+      before_action :prempt_ping_call
+      before_action :verify_content_origin
+      before_action :allow_only_push_event
+      before_action :allow_only_git_scm
+      before_action :verify_payload
+
+      protected
+
+      def repo_branch
+        payload['changes'].find {|change| change['ref']['type'] == 'BRANCH'}['ref']['displayId']
+      rescue
+        nil
+      end
+
+      def possible_urls
+        payload['repository']['links']['clone'].collect {|l| without_credentials(l['href'])}
+      end
+
+      def without_credentials (str)
+        uri = URI.parse(str)
+        uri.user = nil
+        uri.password = nil
+        uri.to_s()
+      rescue
+        str
+      end
+
+      def repo_host_name
+        # Return the full repo url for logging purposes only
+        payload['repository']['links']['self'][0]['href']
+      end
+
+      def repo_full_name
+        # Already returning the full repo url in repo_host_name.
+        ''
+      end
+
+      def prempt_ping_call
+        if request.headers['X-Event-Key'] == 'diagnostics:ping'
+          render plain: nil, status: :accepted, layout: nil
+        end
+      end
+
+      def allow_only_push_event
+        unless request.headers['X-Event-Key'] == 'repo:refs_changed'
+          render plain: "Ignoring event of type `#{request.headers['X-Event-Key']}'", status: :bad_request, layout: nil
+        end
+      end
+
+      def verify_payload
+        if payload.blank?
+          render plain: 'Could not understand the payload!', status: :bad_request, layout: nil
+        end
+        if repo_branch.blank?
+          render plain: 'No branch present in payload, ignoring.', status: :bad_request, layout: nil
+        end
+      rescue => e
+        Rails.logger.warn('Could not understand bitbucket webhook payload:')
+        Rails.logger.warn(e)
+        render plain: 'Could not understand the payload!', status: :bad_request, layout: nil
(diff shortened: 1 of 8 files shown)


sandro-h:
We signed the CLA for our organization, so I'm not sure if I need to also sign it individually. I'll check with the person who organized it.

arvindsv:
@sandro-h No, you shouldn't need to. I checked and your name is mentioned in the org-level signature. One of us will need to tell the CLA bot that it's ok. 

@gocd/committers Please assume the CLA is signed for this PR.

arvindsv:
@sandro-h I took a look at why this was still pending, and I found that you need to sign the CLA as well. :/ Without it, there is no "link" from your name provided as a part of the org-level CLA signature and your GitHub account. Sorry.

It'll look like this, once you click on the CLA link above:

cla

server/webapp/WEB-INF/rails/app/controllers/api/web_hooks/hosted_bit_bucket_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+      protected
+
+      def repo_branch
+        payload['changes'].find {|change| change['ref']['type'] == 'BRANCH'}['ref']['displayId']
+      rescue
+        nil
+      end
+
+      def possible_urls
+        payload['repository']['links']['clone'].collect {|l| without_credentials(l['href'])}
```
bdpiparva: I think it is better to follow the contract instead of overriding `possible_urls`. Please consider implementing `repo_host_name`, `repo_full_name` and `repo_branch` over `possible_urls`.

I already went through the PR description about hosted version sending clone URLs with credentials. Please can you include he `Webhook` payload json` post reducting sensitive information.

sandro-h: I'll check again if I can fit it into the existing contract - on my first attempt to write this, it didn't fit very smoothly with using a list of possible urls to construct the final urls. But I recognize that the current approach of stuffing it into the contract is not very robust.

Regarding a sample payload: the one in the `should call the material update service upon receiving a good request and respond with 202 [accepted]` test is exactly such a redacted payload. If that's not sufficient, please let me know.

sandro-h: Checked the details again, and the main issue is that our bitbucket server returns _varying_ urls, so I cannot extract a _single_ `repo_host_name` that fits. To give you an example, this is what it returns in the clone URL list:
```
"clone": [
  {
    "href": "https://sso.my-company.com/bitbucket/scm/my-project/my-repo.git",
    "name": "http"
  },
  {
    "href": "ssh://git@git.my-company.com:7999/my-project/my-repo.git",
    "name": "ssh"
  }
],
```

So the full host domain differs:
- sso.my-company.com
- git.my-company.com:7999

I can cover that by extracting just the top domain (my-company.com) and adding specific templates to the `possible_urls` list, but that would be way too specific to my company's use case, in my opinion.

So I'm inclined to adapt the contract/hierarchy to support hooks that send a list of ready-made clone urls.
Something like:
```
WebHookController
  \- HostedBitBucketController
  \- GuessUrlWebHookController
       \- GitHubController
       \- BitBucketController
       \- GitLabController
```

Where `WebHookController` defers `possible_urls`, `repo_branch`, and `repo_name` (the latter for logging only) to its subclasses,
and `GuessUrlWebHookController` implements `possible_urls` as before with a list of possible urls and defers the necessary fields in the url templates to its subclasses.

Alternatively we could adapt just the resolution of the url templates in `possible_urls`, to allow for a _list_ returned in `repo_host_name`, which then generates a potentially very large cross-product of all host name / url combinations.

What do you think?

bdpiparva: > ```
> WebHookController
> \- HostedBitBucketController
>  \- GuessUrlWebHookController
>       \- GitHubController
>       \- BitBucketController
>       \- GitLabController
>```

@sandro-h - Sorry for the late response. I think it is a good idea to follow the approach that you mentioned above.

sandro-h: I've pushed the discussed changes. I can squash the commits afterwards, if you wish.

> Where WebHookController defers possible_urls, repo_branch, and repo_name (the latter for logging only) to its subclasses,

I renamed the intended `repo_name` method to `repo_log_name`, to more clearly mark its purpose because it cannot be used for any real logic besides logging (since its content is somewhat arbitrary depending on the webhook type).

One small annoyance for the hosted Bitbucket webhook is that there's no robust way to extract a `repo_log_name` containing the project (e.g. my-project). Hosted Bitbucket only returns the repo name in a separate field and parsing the URL for the project name doesn't sound very robust.
Therefore the `repo_log_name` for the HostedBitBucket hook only contains a host name and repo name.

bdpiparva: @sandro-h - I will take a look at your changes next week.

bdpiparva:
@adityasood - Please can you give it a try to the new webhook endpoint.

adityasood:
verified- 19.3.0-8865.

bashazaid:
Can someone please give the end point what we can use for Bitbucket server.
we are trying to configure webhooks for gocd pipelines  with bitbucket server.
gocd version v20.9

sandro-h:
> Can someone please give the end point what we can use for Bitbucket server.
> we are trying to configure webhooks for gocd pipelines with bitbucket server.
> gocd version v20.9

Hi @bashazaid, it should be `go/api/webhooks/hosted_bitbucket/notify` (checked with gocd 20.7).

Issue #5538, opened 2018-12-10
Multi-cluster Support for Elastic Agents


GaneshSPatil:
##### Issue Type

- Feature enhancement

#### Tasks:
  - [x] Define Elastic Agent Extension v5 (#5937)
  - [x] Migrate existing Elastic Agent plugins to implement v5 extension
  - [x] Support for cluster config entity in GoCD (Migration + CRUD) (#5938)
  - [x] Bump Up Elastic Agent Profiles API to support clusterProfileId (#5983)
  - [x] Migrate Existing EA profiles to point to cluster config using `migrate-config` plugin api call.
  - [x] Fix Elastic Agent Profiles UI to work with cluster config.
  - [x] Remove ability to configure plugin settings for plugins implementing EA extension v5
  - [x] Write a cluster profiles SPA to manage cluster profiles
  - [x] document elastic agent extension v5
  - [x] SPA (#6045)
  - [ ] Show a server health message when v4 plugins are used.
  - [x] Store cluster profile information on plugin infos API

GaneshSPatil:
Current Problems with EA plugins:

* No Multi-Cluster support
* Overriding plugin-setting fields at each individual profile causes a lot of duplication and makes it harder to maintain the profiles, as, change in the field value will require changing it at all profiles.

GaneshSPatil:
Proposed solution:

* Allow users to define named-clusters at the plugin-settings level and then refer to these clusters from elastic profiles level. Users can define any number of named clusters.
Even if just a small field needs to be overridden, users should create a new cluster. 
  Eg: On kubernetes, if a user wants to use two namespaces of the same cluster. A user needs to define two different clusters with same cluster URL but different namespaces.

bdpiparva:
- Approach 1:
   - A new endpoint for the cluster configuration. This allows a user to provide cluster configuration globally(credentials, region etc..)
   - Refer the cluster settings from an elastic profile.
   - Plugin settings will contain only global configuration no credential related.
   - `Create agent`, `validate-profile` and `should assign work` call will get profile and cluster credential in the request

- Approach 2:
  - Modify the existing plugin settings an take a culster settings as JSON, YML or Groovy
  - Validate configuration on save of the settings for `uniqueness` of cluster config name and `mandatory` fields
  - Rest of the stuff will remain the same. This is simple but implementation may vary from plugin to plugin

arvindsv:
[...]

PR #5979, opened 2019-03-13
[Do Not Merge]Add XML Migration to support clusterProfileId attribute on profile tag (#5538)

GaneshSPatil:
* Add clusterProfileId attribute on 'profile' tag
* Define a default clusterProfile under elastic/clusterProfiles

### DIFF
--- a/api/api-elastic-profile-v1/src/main/java/com/thoughtworks/go/apiv1/elasticprofile/representers/ElasticProfileRepresenter.java
+++ b/api/api-elastic-profile-v1/src/main/java/com/thoughtworks/go/apiv1/elasticprofile/representers/ElasticProfileRepresenter.java
@@ -48,7 +48,8 @@ public class ElasticProfileRepresenter {
     public static ElasticProfile fromJSON(JsonReader jsonReader) {
         ElasticProfile elasticProfile = new ElasticProfile(
                 jsonReader.getString("id"),
-                jsonReader.getString("plugin_id"));
+                jsonReader.getString("plugin_id"),
+                null);
         elasticProfile.addConfigurations(ConfigurationPropertyRepresenter.fromJSONArray(jsonReader, "properties"));
         return elasticProfile;
     }
--- a/base/src/main/java/com/thoughtworks/go/util/GoConstants.java
+++ b/base/src/main/java/com/thoughtworks/go/util/GoConstants.java
@@ -54,7 +54,7 @@ public class GoConstants {
 
     public static final String PRODUCT_NAME = "go";
 
-    public static final int CONFIG_SCHEMA_VERSION = 117;
+    public static final int CONFIG_SCHEMA_VERSION = 118;
 
     public static final String APPROVAL_SUCCESS = "success";
     public static final String APPROVAL_MANUAL = "manual";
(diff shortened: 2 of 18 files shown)


GaneshSPatil:
[Do not merge]
---
Had a conversation with @maheshp, based on which following implementation improvements can be done:

1. Make `clusterProfileId` an optional attribute on elastic agent profile.
  - For EA plugins implemented using 'v4' extension, no need of specifying 'clusterProfileId'.
  - For EA plugins implemented using `v5` extension, `clusterProfileId` is mandatory.
  - Instead of adding an XSD validation for `clusterProfileId` being the mandatory field, we can add this check at the Java Config Entity validation level.

2. As the `clusterProfileId` is an optional field on EA profiles. the EA profiles will continue to work as is, with EA extension v4. **No need for config-migration** (defining `default` cluster config and `default` clusterProfileId on all profiles).

3. Introduce `migrate-plugin-settings-to-cluster-profile` call as part of EA extension v5
  - This call accepts **plugin settings** and returns the **cluster profile**. For many plugins, this call will return the exact same configuration back as of plugin-settings. For plugins which have a different configuration for the cluster profile, will need to take care of migration.
  - Upon plugin upgrade, the server will make a call (once) to migrate plugin-settings to cluster-profile, and define a new cluster will the plugin returned profile. Also, all the profiles belonging that plugin will start referring to the newly created cluster.

---

Few Advantages with above approach:
1. Nothing changes for EA plugins implemented using v4 extension. Neither the elastic profile config entity.
2. The task of config migration of plugin-settings to cluster profile is delegated to plugin. As there will be change in the configurations (different/new settings), plugins will take care of config migrations.

\cc: @bdpiparva @maheshp

GaneshSPatil:
Closing this PR.
Will send another PR to support optional clusterProfileId attribute on elastic agent profiles.
AND
Once we decide upon the plugin settings migration strategy, we can either add the config migration or the new plugin-extension call (`migrate-plugin-settings-to-cluster-profile`).

arvindsv:
@GaneshSPatil @maheshp If someone edits the config directly and removes the clusterProfileId for a v5 plugin, what happens? This might work somehow, but my preference would be to keep the config consistent and not dependent on the version of a plugin.

GaneshSPatil:
@arvindsv -- 

> If someone edits the config directly and removes the clusterProfileId for a v5 plugin, what happens?

The validation of cluster_profile_id can happen at the elastic agent profile entity validation level, where if the plugin used for defining elastic agent profile is on v5, the profile needs to have a cluster_profile_id reference.

> This might work somehow, but my preference would be to keep the config consistent and not dependent on the version of a plugin.

Defining an unused attribute on elastic agents profile might cause confusion. As it refers to a dummy cluster which will not be used anywhere.

bdpiparva:
@GaneshSPatil - I think `migrate-plugin-settings-to-cluster-profile` is the good approach as plugin author can define how the migration should happen.

Instead of calling `migrate-plugin-settings-to-cluster-profile` just make it `migrate(pluginSetting, clusterProfiles, elasticProfiles)` and let the plugin author decide what to do.


> Before Migration

```
requestBodyForMigration = {
	"plugin_settings": {"ACCESS_KEY":"ADSDFDGDGSd", "SECRET_KEY":"dsadasddsgdflkghjkdfg"},
	"cluster_profiles" :[],
	"elastic_profiles" : [...],
        "extension_version": "v5"
}
```

> Migration response from the plugin

```
{
  "plugin_settings": {},
  "cluster_profiles": [
    {
      "id": "default",
      "configuration": {
        "ACCESS_KEY": "ADSDFDGDGSd",
        "SECRET_KEY": "dsadasddsgdflkghjkdfg"
      }
    }
  ],
  "elastic_profiles": [...]
}
```

I think we should also start considering versioning the plugin config as well. That will allow plugin author to migrate same config to different version without breaking plugin compatibility.

WDYT?

PR #5991, opened 2019-03-18
Support for secret param in material

bdpiparva:
link to issue #6014

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/util/command/UrlArgument.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/util/command/UrlArgument.java
@@ -17,6 +17,7 @@
 package com.thoughtworks.go.util.command;
 
 import com.thoughtworks.go.config.ConfigAttributeValue;
+import com.thoughtworks.go.config.SecretParamAware;
 import com.thoughtworks.go.config.SecretParams;
 
 import java.net.URI;
@@ -25,7 +26,7 @@ import java.net.URISyntaxException;
 import static com.thoughtworks.go.util.ExceptionUtils.bombIfNull;
 
 @ConfigAttributeValue(fieldName = "url")
-public class UrlArgument extends CommandArgument {
+public class UrlArgument extends CommandArgument implements SecretParamAware {
     protected String url;
     private SecretParams secretParams;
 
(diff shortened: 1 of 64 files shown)


domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java
```diff
@@ -140,7 +140,7 @@ private BuildCommand cleanWorkingDir(String workDir) {
 
     private BuildCommand isRepoUrlChanged(String workDir) {
         return test("-neq",
-                material.getUrlArgument().forCommandline(),
+                material.getUrlArgument().rawUrl(),
```
maheshp: This should be changed back to `forCommandline()`

domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java
```diff
@@ -159,7 +159,7 @@ private BuildCommand cmdClone(String workDir, int cloneDepth) {
         if (material.isShallowClone()) {
             cloneArgs.add(format("--depth=%s", String.valueOf(cloneDepth)));
         }
-        cloneArgs.add(material.getUrlArgument().forCommandline());
+        cloneArgs.add(material.getUrlArgument().rawUrl());
```
maheshp: This should be changed back to forCommandline()

domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java
bdpiparva: done

domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitMaterialUpdater.java
bdpiparva: done

config/config-api/src/main/java/com/thoughtworks/go/util/command/UrlArgument.java
```diff
@@ -35,26 +36,42 @@ public UrlArgument(String url) {
...
+    }
+
+    @Override
+    public String originalArgument() {
         return url;
     }
 
+    @Override
     public String forDisplay() {
+        if (hasSecretParams()) {
```
maheshp: We might have to change this, e.g `https://#{SECRET[id][key]}:password@foo.com/bar.git` how will this be handled?

PR #5993, opened 2019-03-18
Fix version parsing for Git and Hg

akshaydewan:
- Introduced a library to do Semantic Version parsing and comparisons so that we don't have to compare floats. https://github.com/skuzzle/semantic-version (MIT licensed)

## For Git
The earlier regex did not parse the minor version correctly. E.g. for the string `git version 2.17.2 (Apple Git-113)` we would get `group(1)` as `2.1`.
Since we are interested only `major.minor.patch`, the regex group now looks for that only.

## For Hg
The earlier regex did not parse the major version correctly for double digits. E.g. for the string `Mercurial Distributed SCM (version 10.19.40)`, the match would fail. Also attaching a patch version of zero if not present, so that the semver library can parse it properly.

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitVersion.java
+++ b/domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitVersion.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


akshaydewan:
Related to Issue #5856

domain/src/main/java/com/thoughtworks/go/domain/materials/git/GitVersion.java
```diff
@@ -0,0 +1,56 @@
...
+import de.skuzzle.semantic.Version;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
+
+public class GitVersion {
+
+    private static final Pattern GIT_VERSION_PATTERN = Pattern.compile(".*\\s+(\\d+.\\d+.\\d+).*");
```
ketan: This will fail parsing the proper version if given `git version x.y.z (ubuntu 18.04.1)`?

ketan: This probably works much better, I think: https://github.com/JetBrains/intellij-community/blob/master/plugins/git4idea/src/git4idea/config/GitVersion.java#L64-L65

akshaydewan: Thanks! I've made the change.

PR #5994, opened 2019-03-18
Remove unused controller.

ketan:


maheshp:
The `validate` endpoint from PluggableTaskController was used from the old Quick Edit page which has been deleted now, therefore this controller was redundant.

PR #5996, opened 2019-03-19
Store Cluster Profile as part of Job Agent Metadata (#5538)

GaneshSPatil:


server/src/main/java/com/thoughtworks/go/domain/JobAgentMetadata.java
```diff
@@ -25,21 +26,23 @@
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 public class JobAgentMetadata extends PersistentObject {
     private Long jobId;
     private String metadata;
```
bdpiparva: Since metadata is there with metatadataVersion we should make use of it - 

e.g.  `metadataVersion=2.0` and the content can be modeled as 

```json
{ 
    "profile" : {...},
    "cluster_profile": {...}
}
```

Now with the current implementation column name `metadata` does not make sense. It is better to use metadata or rename the column to `elasticprofilemetadata`

server/src/main/java/com/thoughtworks/go/domain/JobAgentMetadata.java
```diff
@@ -55,7 +58,23 @@ public ElasticProfile elasticProfile() {
         return new ElasticProfile(id, pluginId, clusterProfileId, configProperties);
     }
 
-    private static String toJSON(ElasticProfile elasticProfile) {
+    public ClusterProfile clusterProfile() {
+        Gson gson = new Gson();
```
bdpiparva: Make this a constant? and also set naming policy to lower case with underscore

server/src/main/java/com/thoughtworks/go/domain/JobAgentMetadata.java
GaneshSPatil: how about `elasticAgentProfileMetadata`?

Issue #2123, opened 2016-04-20
Shallow clone doesn't apply to submodules


JorritSalverda:
##### Issue Type
- Bug Report
##### Summary

Shallow clone applies --depth 3 to the repository in the material, however the git submodule update doesn't get the --depth parameter added and thus clones the whole history of each submodule. This takes up a lot of time.
###### Basic environment details

- Go Version: `16.3.0`
- JAVA Version: `java-8-openjdk-amd64 jre`
- OS: `debian jessie`
##### Steps to Reproduce
1. set git repository with submodules as material and check shallow clone
2. set clean working directory on stage
3. set git log for git repo as task
4. set git log for submodule - by using it's path as executing directory as task
5. run the pipeline
##### Expected Results

The main git repo shows only 3 log items at most; I would expect the same for each submodule
##### Actual Results

The submodule has full history
##### Possible Fix

Pass --depth n to git submodule update if shallow clone is checked

heaten:
Hi, 
Any idea that when this patched will be merged?

arvindsv:
@heaten The PR was not merged, according to @GaneshSPatil. I've opened a new issue #5856, targeted at GoCD release 19.3, to take a look at some pending issues around git. So, I'll close this issue in favor of that. Thank you for bringing this up.

PR #6010, opened 2019-03-21
Shallow clone submodules

akshaydewan:
Issues: #5856 and #2123

The test for this is kinda hacky and checks only whether the required arguments are present or absent depending on the git version. If there's anything better that can be done, let me know.

I see that sometimes we want to "unshallow" a working copy. Not sure of this scenario (perhaps during a force-push?). Is there any similar scenario where an unshallow may be required for a submodule?

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/FileUtil.java
+++ b/base/src/main/java/com/thoughtworks/go/util/FileUtil.java
@@ -107,6 +107,10 @@ public class FileUtil {
         return uriString.replaceAll("^file:/", "file:///");
     }
 
+    public static String toFileURI(String path) {
+        return toFileURI(new File(path));
+    }
+
     public static String filesystemSafeFileHash(File folder) {
         String hash = Base64.getEncoder().encodeToString(DigestUtils.sha1(folder.getAbsolutePath().getBytes()));
         hash = hash.replaceAll("[^0-9a-zA-Z\\.\\-]", "");
--- a/build.gradle
+++ b/build.gradle
@@ -594,6 +594,22 @@ task gitRelatedTests { thisTask ->
     })
     thisTask.dependsOn eachTestTask
   }
+  finalizedBy 'gitRelatedTestsJunitHtmlReport'
+}
+
+task gitRelatedTestsJunitHtmlReport(type: TestReport) { TestReport thisTask ->
+  group LifecycleBasePlugin.VERIFICATION_GROUP
+  description = "Run all Git related tests"
+
+  forAllTask { Test eachTestTask ->
+    eachTestTask.filter({ TestFilter testFilter ->
+      testFilter.setFailOnNoMatchingTests(false)
+      testFilter.setIncludePatterns("*Git*Test*")
+    })
+    thisTask.reportOn eachTestTask.binResultsDir
+    eachTestTask.finalizedBy(thisTask)
+  }
+  destinationDir = file("${project.buildDir}/reports/tests/gitRelatedTests")
 }
 
 task sparkTest { thisTask ->
--- a/build.gradle
+++ b/build.gradle
@@ -584,6 +584,18 @@ task compileAll { compileAllTask ->
 
 task prepare(dependsOn: compileAll)
 
+task gitRelatedTests { thisTask ->
+  group LifecycleBasePlugin.VERIFICATION_GROUP
+  description = "Run all Git related tests"
+  forAllTask { Test eachTestTask ->
+    eachTestTask.filter({ TestFilter testFilter ->
+      testFilter.setFailOnNoMatchingTests(false)
+      testFilter.setIncludePatterns("*Git*Test*")
+    })
+    thisTask.dependsOn eachTestTask
+  }
+}
+
 task sparkTest { thisTask ->
   group LifecycleBasePlugin.VERIFICATION_GROUP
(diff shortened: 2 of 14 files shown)


arvindsv:
Unshallow was probably during a rerun. I remember the depth being used was something like 1, 2, 100 and then INTEGER_MAX (full clone). See: https://github.com/gocd/gocd/pull/1887

ketan:
Unshallow was used to ensure that the revision pointed to by `*_FROM_REVISION` was present in the current checked out copy. I'm unsure if that'll be needed for submodules.

akshaydewan:
Found one scenario that fails in this PR:
1. Create a git repo with a submodule
2. Checkout the submodule to a previous commit instead of latest
3. Use this as a material in a pipeline and run it. (Apply the shallow clone setting)
4. The submodule fails to fetch the commit since it was cloned with depth=1. `error: Server does not allow request for unadvertised object 1ce4a6...`

I'll work on fixing this, so don't merge this yet. I'm sure there are other scenarios as well.

akshaydewan:
Thinking about this further, this seems risky to implement. In order to figure out whether the submodule contains the required SHA and unshallow it if needed is not trivial, IMO.

One way to do it would be to run `git submodule summary` like so:
```bash
git submodule summary foo
* foo 8250263...4f173fd:
  Warn: foo doesn't contain commit 8250263b19c2db477bb9d39a7e8b6bb5edf19e75
```
Then we would have to parse the output using a regex, unshallow, and check again. I couldn't find any other way to figure out what the expected SHA of the of the submodule should be.

Add nested submodules to the mix, and it opens up a can of worms.

Instead, I propose that users can use a key in [.gitmodules](https://book.git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow), saying `shallow=true` (available since git 2.10), which GoCD will honour automatically. And if there's a case where users want to refer to an older commit, they can remove the entry from .gitmodules.

The drawback would be that it is counter-intuitive if a user has already selected the Shallow Clone checkbox on the materials page.

Let me know what you think @arvindsv @ketan

arvindsv:
@akshaydewan Have you checked whether [shallow](https://book.git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow) can be set in `.git/config` like [other properties](https://book.git-scm.com/docs/git-config#Documentation/git-config.txt-submoduleltnamegtfetchRecurseSubmodules) can?

If that works, that might be an option too. Right?

akshaydewan:
Based on the discussion we had, we're going for a "best effort" approach - We will use the `git submodule update` command with a `--depth=1` option if shallow clone is selected. If the command fails (for whatever reason), we will try again with `--depth=100`. And if that also fails, we will use `Integer.MaxValue` as the depth.

akshaydewan:
> since the --depth option is supported since git 1.9, we don't need to add a check for the version.

Found out the hard way (and luckily) that the behaviour of submodule update with --depth isn't the same on git 2.9 and few other lower versions, although it is supported since 1.8.4. On 2.9 and 2.8 (and probably previous versions?), the submodule refuses to unshallow even with depth=Integer.MaxValue.

I have added a condition to enable --depth only if the user has git 2.10 or above.

I would feel a lot safer if we could somehow run the battery of Git related tests on all versions of Git from 1.9 to the latest :)

Another thing we could consider in the future is allowing users to configure the exact commands for cloning, resetting the working directory, updating submodules etc. on the server level, in a template or such that we can parse execute. There would be some sensible defaults to begin with, and users could tweak and optimize things to their needs. This could apply to other SCMs too. WDYT?

arvindsv:
> I would feel a lot safer if we could somehow run the battery of Git related tests on all versions of Git from 1.9 to the latest :)

Yes. Would be nice. The problematic part of it is getting older versions, I suppose. I checked and there are no good ways of getting a docker image with older versions of git or even nixos packages.

> Another thing we could consider in the future is allowing users to configure the exact commands for cloning, resetting the working directory, updating submodules etc. on the server level, in a template or such that we can parse execute. There would be some sensible defaults to begin with, and users could tweak and optimize things to their needs. This could apply to other SCMs too. WDYT?

Till now, we've had quite strict definitions for what needs to happen. We could look into loosening that, but it might be a nightmare to configure.

ketan:
I think I may have build scripts that spit out git RPMs that can sit side by side, from snap-ci days.

I'll dig them out.

ketan:
https://github.com/snap-ci/git-build/blob/master/README.md

Just need to parametrize this to have it output a bunch of RPMs/DEBs: https://github.com/snap-ci/git-build/commit/7b613d171028e46afa911f260b434c0618928a2a

arvindsv:
Ooh. We should build it once and keep it. :) But, the RPMs might have dependencies to glibc etc. Might have to build and store docker images.

ketan:
You'd need specific binaries for centos 6 and 7. Might be easier to build RPMs than run git via docker.

akshaydewan:
I have the rpms and the build pipeline checked in here: https://github.com/akshaydewan/git-tests

The build is [green](https://build.gocd.org/go/pipelines/git-tests/2/test/2/overview). @ketan Mind taking a look at the code before I merge this?

akshaydewan:
I'm merging this because I need the gradle file changes

PR #6027, opened 2019-03-27
Update trial loading page based on feedback

kierarad:
- Added space and delimiter between server loading message and GoCD concepts
- Replaced spinner with ellipsis
- Updated copy for GoCD concepts

@marques-work can you merge?


marques-work:
@kierarad LGTM - merging

PR #6036, opened 2019-03-28
User management

naveenbhaskar:
1. sticky action panel for user table while scrolling
2. focus to the flash messages after an action
3. confirmation dialog box for delete user action.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
@@ -23,7 +23,7 @@ import {RolesCRUD} from "models/roles/roles_crud";
 import {TriStateCheckbox, TristateState} from "models/tri_state_checkbox";
 import {computeBulkUpdateRolesJSON, computeRolesSelection} from "models/users/role_selection";
 import {UserFilters} from "models/users/user_filters";
-import {BulkUserOperationJSON, BulkUserUpdateJSON, User, Users} from "models/users/users";
+import {BulkUserUpdateJSON, User, Users} from "models/users/users";
 import {UsersCRUD} from "models/users/users_crud";
 import * as Buttons from "views/components/buttons";
 import {FlashMessage, MessageType} from "views/components/flash_message";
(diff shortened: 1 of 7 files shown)


arvindsv:
@naveenbhaskar Can you add an image or an animated gif to show this?

naveenbhaskar:
Mar-29-2019 12-57-25

server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
```diff
@@ -94,7 +95,19 @@ export class UsersPage extends Page<null, State> {
         users: usersToDelete.userNamesOfSelectedUsers()
       };
 
-      this.bulkUserDelete(vnode, json);
+      new DeleteUserConfirmModal(json,
+                                 (msg) => {
```
vrushaliwaykole: Instead of passing callbacks, can we just call `onDelete` in `ondelete` callback of `DeleteConfirmModal`?

server/webapp/WEB-INF/rails/webpack/views/pages/users/delete_user_confirmation_modal.tsx
```diff
@@ -0,0 +1,58 @@
...
+    super(DeleteUserConfirmModal.getMessage(usersToBeDeleted.users), () => this.delete(usersToBeDeleted));
+    this.onSuccessfulSave = onSuccessfulSave;
+    this.onOperationError = onOperationError;
+
+  }
+
+  private static getMessage(usersToBeDeleted: string[]) {
+    return (<div>
+      <p>Are you sure you want to delete these users:
+        <span className={styles.deletedUsers}>
```
vrushaliwaykole: Can we rename style as they are not deleted users but users to be deleted?

kritika-singh3: Fixed

server/webapp/WEB-INF/rails/webpack/views/pages/users.tsx
```diff
@@ -147,8 +160,7 @@ export class UsersPage extends Page<null, State> {
     return (
       <div>
         {bannerToDisplay}
-        <FlashMessage type={this.flashMessage.type} message={this.flashMessage.message}/>
```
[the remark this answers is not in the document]
kritika-singh3: Refactored to maintain consistency as suggested.

adityasood:
@naveenbhaskar @kritika-singh3 the header with `username` etc. should ideally be fixed and should not scroll. See agents page for similar interaction.

adityasood:
@kritika-singh3 @naveenbhaskar while deleting a non-disabled user, I think the delete pop-up does not make sense. We should directly let the user know that the user can not be deleted before being disabled. WDYT?

Issue #5667, opened 2019-01-05
Do not validate plugin properties while deleting a plugin profile entity


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

ArtifactStore, ElasticAgentProfile and SecurityAuthConfig all implement PluginProfile which validates the plugin properties by making a request to the plugin during deletion when it shouldn't - https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/PluginProfilesService.java#L91.

The update method is called from:
- https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/ElasticProfileService.java#L60
- https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/SecurityAuthConfigService.java#L70
- https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/ArtifactStoreService.java#L62

###### Basic environment details

* Go Version: `18.12.0`

maheshp:
Done as part of https://github.com/gocd/gocd/pull/6053

PR #6053, opened 2019-04-02
Do not perform plugin validations when deleting entities. (#5667)

varshavaradarajan:


server/src/main/java/com/thoughtworks/go/server/service/ArtifactStoreService.java
```diff
@@ -51,15 +51,15 @@ public ArtifactStores getPluginProfiles() {
     }
 
     public void create(Username currentUser, ArtifactStore newArtifactStore, LocalizedOperationResult result) {
-        update(currentUser, newArtifactStore, result, new CreateArtifactStoreConfigCommand(goConfigService, newArtifactStore, artifactExtension, currentUser, result));
+        update(currentUser, newArtifactStore, result, new CreateArtifactStoreConfigCommand(goConfigService, newArtifactStore, artifactExtension, currentUser, result), true);
```
bdpiparva: I think command should tell does it requires validation or on instead of passing it to `update(...)`

varshavaradarajan:
@bdpiparva - why? The entityCommand isn't the one performing the plugin validations, its the PluginProfileService. Since the actual validation itself is not performed at the Command, why should this boolean come from the command?

bdpiparva:
@varshavaradarajan - Actually command is doing validation. 

```java
command.validateUsingExtension(newPluginProfile.getPluginId(), newPluginProfile.getConfigurationAsMap(true));
            addErrorsToConfiguration(result, newPluginProfile);
```

bdpiparva:
@varshavaradarajan - any updates?

varshavaradarajan:
@bdpiparva - https://github.com/gocd/gocd/pull/5743#discussion_r251278625

bdpiparva:
> I am wondering if the Command should decide bout the plugin validation. This is more a behaviour of the PluginProfileService, how bout update taking a flag validatePluginProperties and consider validation based on it.

Agreed, that the command is `EntityConfigUpdateCommand` and it should only take care of updating an entity in config.xml, I was confused as the command was internally performing `isValid` which does the config validation.

In this case, the plugin related validation should move out of the command and It should happen from service.

PS: @varshavaradarajan  - Sorry for the delayed response. Please, can you rebase the PR ;) ?

varshavaradarajan:
@bdpiparva - rebased the PR. It is failing for material related errors on master. Once the main build is fixed, I'll rerun the PR pipelines.  Meanwhile, can you please take another look at the PR, especially the changes around elastic profile? Thank you!

bdpiparva:
@varshavaradarajan - PR LGTM. I will merge it post 19.4.0 release.

PR #6056, opened 2019-04-02
Reduce the number of background ajax calls when tests run.

ketan:


server/webapp/WEB-INF/rails/webpack/views/components/server_health_summary/server_health_summary.tsx
```diff
@@ -15,44 +15,47 @@
...
+  private createRepeater(vnode: m.Vnode<{}, State>) {
+    const options = {
+      repeaterFn: () => {
+        return ServerHealthMessages.all()
+                                   .then((result) => {
+                                     result.do(
+                                       (successResponse) => vnode.state.serverHealthMessages(successResponse.body),
+                                       (errorResponse) => {
+                                         // tslint:disable-next-line:no-console
+                                         console.log("There was an error fetching server health messages!");
```
GaneshSPatil: `console.error`? Also, how about printing the `errorResponse` too?

ketan: `console.error` because we don't have a better place to show this error :)

GaneshSPatil: @ketan -- Apology for not being clear in the response.
I was wondering if we could use `console.error` instead of `console.log`.

Issue #5937, opened 2019-03-04
Define Elastic Agent Extension v5


GaneshSPatil:
##### Issue Type

- Feature proposal

##### Summary

Introduce elastic agent extension v5 to support cluster configuration related calls.

### Introduce following new API calls related to cluster profile:
  - [x] Get Cluster Configuration Metadata
  - [x] Get Cluster Configuration View
  - [x] Validate Cluster Configuration
  - [x] Get Cluster Status Report
  - [x] Cluster Profile Changed
### Introduce following new API calls related GoCD config migration:
  - [x] migrate-config

### Modify following API calls to send along associated cluster configuration:
  - [x] Create Agent 
  - [x] Job Completion 
  - [x] Agent Status Report
  - [x] Should Assign Work

### Modify following API calls to send along all cluster configurations:
  - [x] Server Ping
  - [x] Plugin status report page

### Following API calls remain unchanged:
  - [x] Validate Profile
  - [x] Get Profile View
  - [x] Get Profile Metadata
  - [x] Get Plugin Icon

### Capabilities API call changes have following capabilities:
  - [x] supports_plugin_status_report
        Indicates supports getting a status report at the plugin level. (An aggregated view of all the clusters)
  - [x] supports_cluster_status_report
        Indicates supports getting status report for a cluster.
  - [x] supports_agent_status_report
        Indicates supports getting a status report for a single Elastic agent.

### Following messages will not be sent to the plugins implementing v5 extension:
  - [x] Get Settings View
  - [x] Get Plugin Settings Configuration
  - [x] Validate Plugin Settings Configuration

kritika-singh3:
For adding support for cluster-status-report, we are thinking of the following URLs for all status reports:

1. Plugin status report: 
    `status_reports/:plugin_id`
2. Cluster status report:
     `status_reports/:plugin_id/cluster/:cluster_profile_id`
3. Agent status report:
     old: `status_reports/:plugin_id/:elastic_agent_id`
     new: `status_reports/:plugin_id/agent/:elastic_agent_id`

Is it fine or someone has any other suggestion?

zerolaser:
@GaneshSPatil @ketan   Not sure if this is the right place to discuss about elastic agents or not. 
With GoCD Elastic Agent Plugin for Amazon ECS 

[...]

PR #6057, opened 2019-04-02
Added cluster status report api to elastic agent extension v5

kritika-singh3:
The API to access the cluster status if supported will be:
`/go/admin/status_reports/:plugin_id/cluster/:cluster_profile_id`
Link to the issue: #5937

### DIFF
--- a/server/webapp/WEB-INF/rails/app/controllers/admin/status_reports_controller.rb
+++ b/server/webapp/WEB-INF/rails/app/controllers/admin/status_reports_controller.rb
@@ -47,7 +47,23 @@ module Admin
       render_plugin_error e
     end
 
+    def cluster_status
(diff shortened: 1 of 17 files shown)


server/webapp/WEB-INF/rails/app/controllers/admin/status_reports_controller.rb
```diff
@@ -47,7 +47,23 @@ def agent_status
       render_plugin_error e
     end
 
+    def cluster_status
```
GaneshSPatil: How about adding some specs for this controller method?

kritika-singh3: Done

Issue #6058, opened 2019-04-02
Improve OSX installers


ketan:
##### Summary

Since java 9, there's been an introduction of [Platform-Specific Desktop Features](http://openjdk.java.net/jeps/272), internal APIs such as those in the Mac OS X `com.apple.eawt` package will no longer be accessible and is handled by `java.awt.Desktop` class instead.

This makes it difficult to be able to write version independent code that works on both java 8 and anything above.

Considering the difficulty of maintaining and testing a rather primitive desktop app, I'd like for us to remove the desktop app, and move to a simple zip installer for osx (with a JRE packaged alongside).

This will also involve modifying, ir wrapping existing scripts to be able to detect a JRE present alongside the extracted code and use that JRE for running GoCD.

**Backward compatibility notes**:

- allow for [reading environment variables from known locations](https://github.com/gocd/gocd/blob/b41c08c0cef95b925adbd2446ecd6072dd1e3f38/installers/osx-shared/JavaApplicationStub64#L754-L762)
- allow for running out of the correct `WorkingDirectory` (I think `~/Library/{Go Agent,Go Server}/...`
- the agent window app would store the [server URL](https://github.com/gocd/gocd/blob/b41c08c0cef95b925adbd2446ecd6072dd1e3f38/agent-bootstrapper/src/main/java/com/thoughtworks/go/agent/bootstrapper/osx/AgentMacWindow.java#L37), and use that URL to actually launch the agent process.

adityasood:
verified on  19.3.0-8914

PR #6059, opened 2019-04-02
Rejig OSX installers (#6058)

ketan:
* Do not package as an `.app` anymore
* The OSX distribution is now the same as the generic zip installers,
  with an (AdoptOpenJDK) JRE packaged alongside.
* Remove all OSX window apps.


marques-work:
@ketan you could also build out the installer with JRE 11, much like the windows build does. Seems to work well enough for the new trial installers.

ketan:
> @ketan you could also build out the installer with JRE 11, (correct me if I'm wrong, but I recalled the OS X package being bundled with JRE 8) much like the windows build does. Seems to work well enough for the new trial installers.

@marques-work: once I've sorted out all the docker images, the plan is to centralize JVM versions in a single location. As of now, every installer, every docker image seems to pull a slightly different version of the JRE. This is on my list of things to do.

ketan:
Closing in favor of #6122

PR #6069, opened 2019-04-04
Fix data sharing

GaneshSPatil:
* Do not send 'X-Requested-With' header for Non-GoCD API requests

Note:
* This header was introduced as part of PR: #5981
* Sending this header while requesting to datasharing.gocd.org
  will fail for CORS.

\cc: @akshaydewan


GaneshSPatil:
This failure was caught as part of https://build.gocd.org/go/tab/pipeline/history/verify-usage-data-reporting pipeline.

\cc: @rajiesh @adityasood

PR #6084, opened 2019-04-05
Add config repo capabilities to plugin infos

### LINKED ISSUE #5730: Pipeline export should list only plugins with export capabilities

kierarad:
In order to fix #5730 we need to be able to filter the list of plugins displayed to only include those that support pipeline export. 

This adds capabilities to the config repo extension of the plugin infos so that it can be accessed. 
It caches the capabilities in the metastore. 
It filters the list to include plugins that support pipeline export.

### DIFF
--- a/api/api-plugin-infos-v5/src/main/java/com/thoughtworks/go/apiv5/plugininfos/representers/extensions/ConfigRepoExtensionRepresenter.java
+++ b/api/api-plugin-infos-v5/src/main/java/com/thoughtworks/go/apiv5/plugininfos/representers/extensions/ConfigRepoExtensionRepresenter.java
@@ -27,11 +27,10 @@ public class ConfigRepoExtensionRepresenter extends ExtensionRepresenter {
 
         ConfigRepoPluginInfo configRepoPluginInfo = (ConfigRepoPluginInfo) extension;
 
-        if (configRepoPluginInfo.getCapabilities() != null) {
-            extensionWriter.addChild("capabilities", capabilitiesWriter ->
-                    capabilitiesWriter.add("supports_pipeline_export", configRepoPluginInfo.getCapabilities().isSupportsPipelineExport())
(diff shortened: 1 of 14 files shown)


api/api-plugin-infos-v5/src/main/java/com/thoughtworks/go/apiv5/plugininfos/representers/extensions/ConfigRepoExtensionRepresenter.java
```diff
@@ -16,5 +16,22 @@
...
+import com.thoughtworks.go.plugin.domain.configrepo.ConfigRepoPluginInfo;
+
 public class ConfigRepoExtensionRepresenter extends ExtensionRepresenter {
+    @Override
+    public void toJSON(OutputWriter extensionWriter, PluginInfo extension) {
+        super.toJSON(extensionWriter, extension);
+
+        ConfigRepoPluginInfo configRepoPluginInfo = (ConfigRepoPluginInfo) extension;
+
+        if (configRepoPluginInfo.getCapabilities() != null) {
```
bdpiparva: No need to add the `null` check. We always ask the plugin to send capabilities. In absence of capabilities, it should fail much earlier while building plugin info.

kierarad: Ok that makes sense. I'll remove it.

Issue #6087, opened 2019-04-05
Elastic Agents Profiles API allows creating profiles when plugin is absent


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Elastic agent profiles API allows users to create or update an elastic agent profile when plugin is absent.
This will cause the elastic agent profiles configurations to not be validated as the plugin is missing resulting in possibly invalid configurations.

###### Basic environment details

* Go Version: `GoCD Version: 19.3.0 (8785-34d4f423859709a883d3403b1d6c47ac4c5516be).`

###### Additional Environment Details

Intially, Elastic Agents Profiles API wasn't allowing creation or updates to elastic agent profiles in case of plugin absense.
Later, to support artifact stores, we've made changes to `PluginProfilesService` as part of [`7f86768ef69554960e7727c6285ff21db87d95e2`](https://github.com/gocd/gocd/commit/7f86768ef69554960e7727c6285ff21db87d95e2) commit, which caused this to go through.

older stacktrace it was failing for:
```
2019-04-05 11:13:40,084 ERROR [156@MessageListener for ServerPingListener] JMSMessageListenerAdapter:77 - Exception thrown in message handling by listener com.thoughtworks.go.server.messaging.elasticagents.ServerPingListener@5bece3c
com.thoughtworks.go.plugin.access.PluginNotFoundException: Did not find 'elastic-agent' plugin with id 'cd.go.contrib.elastic-agent.docker'. Looks like plugin is missing
	at com.thoughtworks.go.plugin.access.PluginRequestHelper.submitRequest(PluginRequestHelper.java:40)
	at com.thoughtworks.go.plugin.access.elastic.ElasticAgentExtension.serverPing(ElasticAgentExtension.java:80)
	at com.thoughtworks.go.plugin.access.elastic.ElasticAgentPluginRegistry.serverPing(ElasticAgentPluginRegistry.java:50)
	at com.thoughtworks.go.server.messaging.elasticagents.ServerPingListener.onMessage(ServerPingListener.java:32)
	at com.thoughtworks.go.server.messaging.elasticagents.ServerPingListener.onMessage(ServerPingListener.java:22)
	at com.thoughtworks.go.server.messaging.activemq.JMSMessageListenerAdapter.runImpl(JMSMessageListenerAdapter.java:73)
	at com.thoughtworks.go.server.messaging.activemq.JMSMessageListenerAdapter.run(JMSMessageListenerAdapter.java:53)
[...]

PR #6088, opened 2019-04-05
Validate Elastic Agent Plugin exists for create and update elastic profiles API (#6087)

GaneshSPatil:


server/src/main/java/com/thoughtworks/go/server/service/PluginProfilesService.java
```diff
@@ -59,13 +60,15 @@ public M findProfile(String id) {
         return result;
     }
 
-    private void validatePluginProperties(PluginProfileCommand command, PluginProfile newPluginProfile) {
+    void validatePluginProperties(PluginProfileCommand command, PluginProfile newPluginProfile) {
         try {
             ValidationResult result = command.validateUsingExtension(newPluginProfile.getPluginId(), newPluginProfile.getConfigurationAsMap(true));
             addErrorsToConfiguration(result, newPluginProfile);
-        }catch (RecordNotFoundException e) {
+        } catch (RecordNotFoundException e) {
```
varshavaradarajan: If this exception's no longer for #6087 , then the message should change or remove the catch block for RecordNotFoundException?

GaneshSPatil: @varshavaradarajan -- it is used [here](https://github.com/gocd/gocd/blob/master/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/PluginRequestHelper.java#L40-L42)

PR #6122, opened 2019-04-10
Updated OSX installers

kritika-singh3:
### Highlights
1. No `.app`
2. JRE packaged alongside
3. No OSX windows app

### Commands
server:  `bin/go-server [start | stop]`
agent: `bin/go-agent [start | stop] [-serverUrl https://gocd-server/go]`

Link to issue: #6058


adityasood:
This was verified, there was a small change. We removed passThrough from server to not allow any JavaAppArgs being passed, since server does not take any arguments. It was done as part of https://github.com/gocd/gocd/commit/35c13cd8a6b31e16323350a50086806536fee958. For any additional params, users can alter wrapper-properties.conf file

arvindsv:
@adityasood @kritika-singh3 The documentation needs to be updated: https://docs.gocd.org/19.3.0/installation/install/server/osx.html

Issue #6030, opened 2019-03-27
Modification check failed for material for pipelines that polling is disabled and or are paused .


childsc:
##### Issue Type
- Bug Report

##### Summary

We in our team manage about 800 pipelines for different teams within the company I work for.

Unfortunately, we are receiving a lot of warning and errors with the "message" starting "Modification check failed for material ..."

All these pipelines are for teams where the branch which they were using is no longer available and they have not configured the new branch in Go (possibly because there is no new branch to work from yet)

I was hoping that if you were to either (or both) Pause a pipeline and/or NOT select polling in the pipeline configuration it would stop this behavior.

The teams do not want to keep deleting the materials and adding dummy materials to pipelines and then needing to recreate the materials when needed as this would mean additional extra work.

###### Basic environment details

* Go Version: GoCD Version: 18.12.0 (8222-e6778f1cc84bf91e323d337c876046167da36985).
* JAVA Version:  OpenJDK 11.0.1
* OS: Linux 3.13.0-147-generic
* Browser vendor and version (if relevant): NA

###### Additional Environment Details
Example of error:

message : Modification check failed for material: URL: http://USERNAME:******@[URL PATH TO REPO] Branch: testing
detail  : Failed to run git clone command STDERR: Cloning into '/var/lib/go-server/pipelines/flyweight/6b10e850-5b6a-4bd2-acc5-9972543d87ad'...
          STDERR: fatal: Authentication failed for 'http://USERNAME:******@[URL PATH TO REPO]'
level   : ERROR
time    : 3/27/2019 10:45:25 AM

message : Modification check failed for material: URL: http://USERNAME:******@[URL PATH TO REPO], Branch: development
detail  : Working directory: pipelines/flyweight/10667df4-940b-4fdd-b534-2a80e07028d4
          STDERR: fatal: Authentication failed for 'http://USERNAME:******@[URL PATH TO REPO]'
level   : ERROR
time    : 3/27/2019 10:44:36 AM

##### Steps to Reproduce

1.  Create Test Branch for repo
2.  Configure Go to use above created branch as a material
3.  Disable polling for the material 
4.  Remove Branch from repo
5.  Receive warning message

##### Expected Results
Disable polling of material should mean repo is not verified.

##### Actual Results
We get a "Modification check failed for material:" warning 

##### Possible Fix
When disabling polling materials are not actually checked for existence. I would hope this should also boost the server performance?

##### Log snippets
[...]

PR #6124, opened 2019-04-10
Adding auto_update to material health state scope

ibnc:
* Addresses #6030

### DIFF
--- a/common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java
+++ b/common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java
@@ -58,7 +58,7 @@ public class HealthStateScope implements Comparable<HealthStateScope> {
     }
 
     public static HealthStateScope forMaterial(Material material) {
-        return new HealthStateScope(ScopeType.MATERIAL, material.getSqlCriteria().toString());
+        return new HealthStateScope(ScopeType.MATERIAL, material.getAttributesForScope().toString());
     }
 
     public static HealthStateScope forMaterialUpdate(Material material) {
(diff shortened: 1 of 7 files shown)


arvindsv:
I think I see where this is going, but can you please explain exactly what was happening and how this fixes it? Thank you!

ibnc:
@arvindsv sure.

Here's a run down of the issue:
1. An error occurs for a material update
2. A health state message is created for that material
3. "Poll for new changes" is turned off for that material
4. After the config has been updated, we go through the materials, and remove any health state messages that are no longer relevant because the config changed, i.e. stale messages (see [here](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/materials/MaterialUpdateService.java#L235))
5. The health state message for the material in question is not removed because, from the health state scoping perspective, nothing has changed. We scope materials [here](https://github.com/gocd/gocd/blob/master/common/src/main/java/com/thoughtworks/go/serverhealth/HealthStateScope.java#L61) using the material's [sql criteria](https://github.com/gocd/gocd/blob/master/domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java#L64). While the sql criteria is slightly different for each material type, it never includes `autoUpdate`. So, the health state message never gets considered as stale, and thus is never removed.

The changes here add the `autoUpdate` field to the `HealthStateScope` for materials. So with these changes, when polling is disabled for a material, it's health state messages will be cleaned up.


@arvindsv Let me know if I need to clarify anything I've written here.

arvindsv:
Makes sense to me. I'd prefer if someone else looked at the code, though. Others probably have a more up-to-date view of changes.

One case you should probably consider and verify is updating a material config through the [pipeline config API](https://api.gocd.org/current/#edit-pipeline-config).

A test analogous to [this one](https://github.com/gocd/gocd/blob/add0c661f6f975d3a604e1141658f87c8f076671/server/src/test-integration/java/com/thoughtworks/go/server/materials/MaterialUpdateServiceIntegrationTest.java#L57) might also help make it clear. Or maybe something in MaterialUpdateServiceTest.

arvindsv:
@ibnc Does this work when the pipeline config API is used?

ibnc:
@arvindsv I can confirm that this works even if you update the pipeline config through the pipeline config api

PR #6126, opened 2019-04-11
Remove support for agent backward compatibility all the way to 16.4

ketan:


ketan:
Removing some bootstrapper/launcher/agent side code that managed to understand the pre 16.4 invocation of the agent  (`java -jar agent-bootstrapper.jar example.com 8153`). This invocation connected to port 8153, to determine what the SSL port of the server is and was a possible vector for someone to MITM the connection at the initial connection phase of the agent.

PR #6129, opened 2019-04-11
Backup Page UI updates

akshaydewan:
Code changes for incorporating the new backups UI.

This required `backupProgressStatus` to be saved in the DB, since we need to know at what stage a backup errored out. Otherwise, the would assume that the error occurred on the last step which was in progress when it polled the server, which may be incorrect.

**NOTE:** The UI isn't finalized yet. I'll work with @naveenbhaskar to get that done. It currently looks like this, and I think this PR should be mergeable in this state:

Screenshot 2019-04-12 at 11 31 43


akshaydewan:
Paired with @naveenbhaskar and implemented the remaining UI changes:

Screenshot 2019-04-12 at 16 15 26

arvindsv:
Can you change all "Go" to "GoCD" please? Thank you for sharing an image. Makes it so much easier to take a look at. :)

arvindsv:
Also: Consistently calling it "Config" or "config". It alternates between upper-case and lower-case.

Could we call "Config repo" something else. In this case, it's config.git, isn't it? Maybe it can be called "Configuration History" or something? Oh. In the legend, it's called: "XML Configuration Version Repository". I'd just call it history.

akshaydewan:
@arvindsv I've updated the text and used headline-capitalization on the backup steps. I went with *Configuration History*. That made more sense to me.

Screenshot 2019-04-15 at 11 16 50

adityasood:
@akshaydewan can the right side where we explain what is backed up have "Config" Instead of "Configuration" as the first step or vice-versa to be consistent? It is confusing to me right now if both of them are same

akshaydewan:
@adityasood Done, and updated the screenshot in the previous comment.

bdpiparva:
Currently, we are printing information about when a backup is triggered. It is late for me to ask but can we also consider adding the total time it took to complete the backup(if possible).

This will help in the following situations -
- Admins to predict how much time server will be in the unusable state if I trigger the backup now.
- What if the sudden marginal change is backup time?
    - Did I miss anything to include in a backup?
    - I am backing up things that not suppose to be part of my backup?

@akshaydewan, @arvindsv  - WDYT?

akshaydewan:
Yes, I think we can show the total backup time, and possibly even the time that each of the steps took. But I'm not sure about the benefit it would offer.

In any case, that should be a separate PR, so I've merged this one.

arvindsv:
I'd just leave it at one time, rather than having it for each step.

PR #6130, opened 2019-04-11
Normalize how URL arguments are split and persisted in SCM configuration

kritika-singh3:
Issue Link:  #6132

 - Introduces `username`, `password`, `encryptedPassword` attributes on SCM materials that did not support it (git, hg).
- XSL and config repo migrations that grab the credentials from the URL and convert it to relevant username/password attributes for git and hg materials.
- Added `username` and `password` fields on (rails) UI to allow users to enter the respective values.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/materials/ScmMaterialConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/materials/ScmMaterialConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 ThoughtWorks, Inc.
+ * Copyright 2019 ThoughtWorks, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,20 +17,22 @@
 package com.thoughtworks.go.config.materials;
 
 import com.thoughtworks.go.config.*;
-import com.thoughtworks.go.config.elastic.ElasticProfile;
+import com.thoughtworks.go.config.preprocessor.SkipParameterResolution;
 import com.thoughtworks.go.config.validation.FilePathTypeValidator;
 import com.thoughtworks.go.domain.ConfigErrors;
-import com.thoughtworks.go.domain.PipelineGroups;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
+import com.thoughtworks.go.security.GoCipher;
 import com.thoughtworks.go.util.FilenameUtil;
 import com.thoughtworks.go.util.command.UrlArgument;
 import org.apache.commons.lang3.StringUtils;
 
+import javax.annotation.PostConstruct;
 import java.io.File;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static com.thoughtworks.go.util.ExceptionUtils.bomb;
 import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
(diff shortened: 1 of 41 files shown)


config/config-api/src/main/java/com/thoughtworks/go/config/materials/ScmMaterialConfig.java
```diff
@@ -55,6 +60,16 @@
     @ConfigAttribute(value = "autoUpdate", optional = true)
     private boolean autoUpdate = true;
 
+    @ConfigAttribute(value = "username", allowNull = true)
+    protected String userName;
```
bdpiparva: `allowNull` on username attribute for `TfsMaterialconfig` was set to `false` not it is `true`. This must be handled in the entity validation

config/config-api/src/main/java/com/thoughtworks/go/config/materials/mercurial/HgMaterialConfig.java
```diff
@@ -170,5 +157,12 @@ public void setConfigAttributes(Object attributes) {
         if (map.containsKey(URL)) {
             this.url = new HgUrlArgument((String) map.get(URL));
         }
+        if (map.containsKey("userName")) {
```
bdpiparva: This can be moved to `ScmMaterialConfig` as this is now supported by all materials

config/config-api/src/main/java/com/thoughtworks/go/config/materials/git/GitMaterialConfig.java
```diff
@@ -229,6 +214,13 @@ public void setConfigAttributes(Object attributes) {
             String branchName = (String) map.get(BRANCH);
             this.branch = StringUtils.isBlank(branchName) ? DEFAULT_BRANCH : branchName;
         }
+        if (map.containsKey("userName")) {
```
bdpiparva: This can be moved to `ScmMaterialConfig`

config/config-api/src/main/java/com/thoughtworks/go/config/migration/UrlDenormalizerXSLTMigration120.java
```diff
@@ -0,0 +1,105 @@
...
+        } catch (URISyntaxException e) {
+            return originalUrl;
+        }
+    }
+
+    public static String getUsername(String originalUrl) {
+        try {
+            if (isSupportedUrl(originalUrl)) {
+                String[] credentials = getCredentials(originalUrl);
+                if (credentials != null) {
```
bdpiparva: If `getCredentials(originalUrl)` returns one argument then we are considering it as `username`. It will be displayed in plain text in API calls.

config/config-server/src/main/resources/cruise-config.xsd
```diff
@@ -220,11 +220,17 @@
   </xsd:complexType>
   <xsd:complexType name="configRepoGitType">
     <xsd:attribute name="url" type="xsd:string" use="required"/>
+    <xsd:attribute name="username" type="xsd:string"/>
```
bdpiparva: Is it possible to define common attributes as an abstract type?

config/config-api/src/main/java/com/thoughtworks/go/config/materials/mercurial/HgMaterialConfig.java
ketan: The case of for SVN material, the key is `username`, so left it as is. Also: this method is only used by the rails based clicky admin pages, so I did not feel it appropriate to do any major changes.

config/config-api/src/main/java/com/thoughtworks/go/config/materials/git/GitMaterialConfig.java
ketan: Same as https://github.com/gocd/gocd/pull/6130#discussion_r275265469

config/config-server/src/main/resources/cruise-config.xsd
ketan: It should be, but not sure if it was worth doing in xsd :-/

PR #6131, opened 2019-04-11
Show the affected pipelines when material update fails.

varshavaradarajan:
* This shows all pipelines - local and config repo.

Showing the pipeline names for pipelines for which a user may not have access to view should be alright as its just the pipeline name, imo. 

Screen Shot 2019-04-11 at 9 38 48 PM

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/server_health_summary/server_health_messages_modal.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/server_health_summary/server_health_messages_modal.tsx
@@ -58,7 +58,7 @@ export class ServerHealthMessagesModal extends Modal {
     return <li data-test-id={messageId}
                data-test-message-level={message.level.toLowerCase()}
                className={classnames(styles.serverHealthStatus, message.level.toLowerCase())}>
-      <span data-test-class="server-health-message_message" className={styles.message}>{message.message}</span>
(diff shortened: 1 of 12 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/server_health_summary/server_health_messages_modal.tsx
```diff
@@ -58,7 +58,7 @@ export class ServerHealthMessagesModal extends Modal {
     return <li data-test-id={messageId}
                data-test-message-level={message.level.toLowerCase()}
                className={classnames(styles.serverHealthStatus, message.level.toLowerCase())}>
-      <span data-test-class="server-health-message_message" className={styles.message}>{message.message}</span>
+      <span data-test-class="server-health-message_message" className={styles.message}>{m.trust(message.message)}</span>
```
varshavaradarajan: Since most messages are already escaped in ServerHealthState and the messages where we want to show html are crafted by us, m.trust here should be fine.

GaneshSPatil:
@varshavaradarajan -- Showing the pipelines affected by material related errors in the `errors and warnings` will be an improvement over what we have today. As of today, we don't know at how many places this pipeline is configured.

But, is there an easy way for us to show this information on the dashboard?

\cc: @naveenbhaskar

PR #6139, opened 2019-04-15
Remove elastic agent profiles API v1 (#5538)

GaneshSPatil:


GaneshSPatil:
\cc: @maheshp @arvindsv

arvindsv:
Does it make sense to show a good message in the response body, when API versions are deprecated?

maheshp:
> Does it make sense to show a good message in the response body, when API versions are deprecated?
I would prefer not changing the response body, probably consider an appropriate response code.

GaneshSPatil:
@maheshp @arvindsv -- can we add a custom response header for deprecated APIs?

```
X-GoCD-Deprecated-API: true
```

Though, I am not sure how many people will pay attention to the response headers.

arvindsv:
Depends on what the current response is. If it's easy to understand that the version is no longer supported, then it's ok.

PR #6149, opened 2019-04-16
Ensure that all classes in test module is executed

bdpiparva:
- Making sure that the task should not fail if the concrete test class does not have any test.
- Always use `Abstract` in class name if the test class is marked abstract
- Lint will ignore the classes from validation if it does not follow *Test.(java|groovy) convention.


build.gradle
```diff
@@ -697,6 +708,21 @@ static def toCamelCase(String input) {
   return input.split("[^a-zA-Z0-9]").collect { it.capitalize() }.join("")
 }
 
+def static getTestClassesForTask(subProject, testTask) {
+  def testDirRoot = "/${testTask.name}/java/"
```
ketan: Use `sourceSets[${task.name}].srcDirs`? Reason being that you can possibly have tests under `groovy` (or `kotlin`) as well.

PR #6157, opened 2019-04-17
Exclude secrets extension in PluginInfo API response

vishaldevgire:
This is to hide secrets extension from current release.


api/api-plugin-infos-v5/src/main/java/com/thoughtworks/go/apiv5/plugininfos/representers/ExtensionRepresenterResolver.java
```diff
@@ -35,7 +35,7 @@
         extensionRepresenter.put("notification", new NotificationPluginInfoRepresenter());
         extensionRepresenter.put("analytics", new AnalyticsPluginInfoRepresenter());
         extensionRepresenter.put("artifact", new ArtifactPluginInfoRepresenter());
-        extensionRepresenter.put("secrets", new SecretsExtensionRepresenter());
+//        extensionRepresenter.put("secrets", new SecretsExtensionRepresenter());
```
bdpiparva: 🤔There were not tests?

vishaldevgire: There is a test for representer [here](https://github.com/gocd/gocd/blob/adf0f16777702c01c50d55bb003bdac89b910d12/api/api-plugin-infos-v5/src/test/groovy/com/thoughtworks/go/apiv5/plugininfos/representers/extensions/SecretConfigExtensionRepresenterTest.groovy#L25) and other logic is generic for which tests existed.

Issue #6142, opened 2019-04-16
Clear/Reset local storage on GoCD server upgrade to avoid stale notifications


adityasood:
##### Issue Type
- Bug Report

##### Summary

The notifications for the latest GoCD server version are stored in the local storage of the browser. When a GoCD server is upgraded from a previous version, while you were on the previous version the system notifications were not read. We currently do not clear/update the local storage post upgrade. This is the reason the user can see these messages post the upgrade as well. Clearing local storage should remove the messages.

## GoCD server on version 18.8.0

![Screen Shot 2019-04-16 at 11 10 18 AM](https://user-images.githubusercontent.com/17826464/56195946-8ae00500-6053-11e9-985a-ca6361454457.png)

## GoCD server on version 19.2.0

![Screen Shot 2019-04-16 at 11 16 22 AM](https://user-images.githubusercontent.com/17826464/56196068-c2e74800-6053-11e9-804e-dcfc19c3dbce.png)

### Discussed Solution

Options, we either should clear out the local storage or read the notification on an upgrade.

The best win is to clear out the local storage on an upgrade. All the information will be populated again on the first-page load.

We can store the current_gocd_version as part of local storage, and can check the current GoCD version with the local storage GoCD version,
and if those are different, we clear out the entire local storage.

We can perform this cleanup on the login page, as no background tasks such as version_updater, usage_data_reporter, system_notifications (which requires local storage data) get triggered on the login page.

@GaneshSPatil @ketan

arvindsv:
Tentatively marking this for 19.3. Not sure of the effort, etc. I think clearing it during an upgrade makes sense to me.

GaneshSPatil:
@arvindsv @ketan @adityasood -- Clearing local storage on the login page wasn't an option because:
1. When security is disabled, the login page won't load up, and this bug will be not be fixed for such systems.
2. Clearing the entire local storage will result in showing the data-sharing notification on every upgrade. (which would be annoying)

Hence, we've taken the following approach:
Remember the GoCD version for which the update-check was performed (in local storage). On an upgrade (check if the GoCD version has changed), remove the currently populated update-check notification. This will cause the old update check notification to be removed.
So, 
1. When the upgraded version is the latest GoCD version, no new notification will be shown.
2. But, if the upgraded version is not the latest GoCD version, the new update notification will be shown.

adityasood:
verified: 19.3.0-8940

PR #6166, opened 2019-04-18
Fix version updater check to show notification only when newer version is avalialble (#6142)

GaneshSPatil:
* GoCD server stores user notifications as part of local storage.
* When a newer gocd version is available, a new system notification
  for newer version is added into the local storage, which gets
  displayed on the header (bell icon)
* But, if the user doesnt read the notification and upgrades GoCD
  instance, the stale system notification from the local storage will
  be diplayed.

* The system notification for an upgrade is added based on comparision
  between currently installed GoCD version and newly available GoCD version.
* If the currently installed GoCD version and the version for which update
  notification was added are different, then its a stale notification and
  such notification is removed.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/CurrentGoCDVersion.java
+++ b/base/src/main/java/com/thoughtworks/go/CurrentGoCDVersion.java
@@ -30,6 +30,7 @@ public class CurrentGoCDVersion {
     private final String copyrightYear;
     private final String baseDocsUrl;
     private final String baseApiDocsUrl;
+    private final String gocdDistVersion;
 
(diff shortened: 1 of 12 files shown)


arvindsv:
I assume other notifications are unaffected.

GaneshSPatil:
@arvindsv -- yes. As we're dealing with on "UpdateCheck" notification, other notifications are unaffected.
Implemented specifics are documented [here](https://github.com/gocd/gocd/issues/6142#issuecomment-484497282) and [here](https://github.com/gocd/gocd/pull/6166/commits/383293ae7047f978d68ed85e41954da7a545790f)

PR #6171, opened 2019-04-19
Initial cut of replacing X509 with simpler HTTP authentication.

ketan:
An agent would now be expected to send the following headers:

```
X-Agent-GUID: some-guid
Authorization: token
```

The `token` is the HMAC based token returned to the agent as
part of the agent registration process.


ketan:
No changes have been made on the agent end. I'm thinking about adding some agent side feature toggle that we can possibly dog food with.

ketan:
@rajiesh @adityasood: this PR is good to merge. Since the agent now uses a combination of the guid+token to authenticate (instead of SSL certs), you should now be able to tunnel a connection through a reverse proxy. The agent will still require that you use an https url, so I recommend using ngrok (or something similar) to proxy your connection.

To force an agent to use token based authentication, set the environment variable `GO_USE_TOKEN_AUTH=true`.

Issue #6173, opened 2019-04-19
Remove pluginId field from profile tag


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Remove `pluginId` attribute from the profile tag. Each elastic agent profile points to a cluster profile. PluginId for a profile is implicitly referenced from cluster profile.

\cc: @maheshp

rajiesh:
Verified on 19.3.0 (8953-b8f0069eae5a58849cd7e0835b5f2ff421be02bc)

GaneshSPatil:
Reopening this issue.

As of now, we have fixed API only to not render plugin id, the `agentProfile` tag under `cruise-config.xml` still contains `pluginId` attribute.

rajiesh:
Verified on 19.4.0-9155

PR #6176, opened 2019-04-22
Remove plugin id from elastic agent profiles API (#5538)

GaneshSPatil:


GaneshSPatil:
Fixed as part of #6173

PR #6186, opened 2019-04-22
Extract materials models and test connection component

marques-work:
This PR is some work that has come out of developing the new "Add New Pipeline" page. Specifically, this makes the following changes:

1.  Moves the `materials` models/types from `config_repos` (a sub-model of this) to a top-level `materials` model. Also moves `checkConnection()` to be a `Material` method, rather than part of `ConfigRepoCrud`.
2. Extracts the "test connection button" into a proper component that can be shared between pages.

Apologies in advance for bundling these two changes into one PR. I felt it was easier for the team working on the Add Pipeline page to not have a PR that depends on just 1 additional PR rather than 2. If anyone feels strongly about this (or if it becomes significantly harder to understand), I'd be happy to split this into distinct PRs.


marques-work:
Merging now that the 19.3.0 was released. @ketan reviewed this some days ago.

PR #6190, opened 2019-04-24
login screen  - icon and text size increased

naveenbhaskar:
image

image

image


GaneshSPatil:
@naveenbhaskar -- The changes looks Okay.
Though, on the page load, the icon is rendered outside and then its placed at the center.
This causes a jumping effect on the icon. is this intentional?

ezgif com-video-to-gif

naveenbhaskar:
Fixed the jump

PR #6191, opened 2019-04-24
cluster profiles header panel alignment and buttons alignment

naveenbhaskar:
1. cluster profiles panel header alignment on 1366 resolution fix
2. spaces between buttons and icons on the collapsible header panel in various pages were not consistent. fixed  it

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.scss
@@ -81,7 +81,7 @@
 
 .icon-doc {
   @include icon-before($type: $fa-var-file);
-  margin-right: 10px;
+  margin-left: 20px;
 
   &:before {
     margin: 5px 10px 5px -5px;
--- a/server/webapp/WEB-INF/rails/webpack/views/components/icons/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/icons/index.scss
@@ -29,6 +29,7 @@ $btn-group-border-radius: $global-border-radius;
   color:         $icon-color;
   transition:    $transition;
   background:    $icon-bg;
+  margin-left:   20px;
 
   &:hover {
     color:      lighten($icon-color, 100%);
@@ -127,11 +128,13 @@ $btn-group-border-radius: $global-border-radius;
   border:        1px solid $border-color;
   border-radius: $btn-group-border-radius;
   display:       inline-flex;
+  margin-left:   20px;
 
   .btn-icon {
     border:        none;
     border-left:   1px solid $border-color;
     border-radius: 0;
+    margin-left:   0;
 
     &:first-child {
       border-left:   none;
--- a/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.scss
@@ -16,7 +16,6 @@
 @import "../../global/common";
 
 $key-value-pair-width: 200px;
-$title-element-width-xl: 280px;
 $title-element-width: 360px;
 
 .key-value-pair {
@@ -121,11 +120,7 @@ $title-element-width: 360px;
     font-size: 12px;
 
   }
-  @media (min-width: $screen-xl) {
-    font-size: 15px;
-    width:     $title-element-width-xl;
-  }
-  @media (min-width: $screen-xxl) {
+  @media (min-width: $screen-lg) {
     font-size: 15px;
     width:     $title-element-width;
   }
--- a/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/key_value_pair/index.scss
@@ -16,6 +16,7 @@
 @import "../../global/common";
 
 $key-value-pair-width: 200px;
+$title-element-width-xl: 280px;
 $title-element-width: 360px;
 
 .key-value-pair {
@@ -120,7 +121,11 @@ $title-element-width: 360px;
     font-size: 12px;
 
   }
-  @media (min-width: $screen-lg) {
+  @media (min-width: $screen-xl) {
+    font-size: 15px;
+    width:     $title-element-width-xl;
+  }
+  @media (min-width: $screen-xxl) {
     font-size: 15px;
     width:     $title-element-width;
   }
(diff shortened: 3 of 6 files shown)


GaneshSPatil:
@naveenbhaskar -- Not sure if these changes are required or fixes the mentioned issue.
For ECS Plugin, the cluster-collapsible-header is misaligned.

Screen Shot 2019-04-24 at 12 47 23 PM


---
### Note:
**As with a long `id` and `plugin-id` in the header, both `id` and `plugin-id` will be stripped. Hence we've removed the `plugin-id` from the header in the latest GoCD master.**


### The updated Elastic Profiles Page without plugin Id looks like:

Screen Shot 2019-04-24 at 12 52 01 PM

naveenbhaskar:
reverted the panel title width.

GaneshSPatil:
@naveenbhaskar -- I see 3 commits as part of the PR whereas few of the commits are already on GoCD master. Can we remove those unnecessary commits?

server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.scss
```diff
@@ -81,7 +81,7 @@
 
 .icon-doc {
   @include icon-before($type: $fa-var-file);
-  margin-right: 10px;
+  margin-left: 20px;
```
GaneshSPatil: @naveenbhaskar -- I am not sure why we need to add `margin-left` to an icon.
An icon simply should have styles to render an icon. 
The spacing styles of the icon are dependent on the place where the icon is used. I would say for margin and padding, we should wrap the icon in a  different container and style that particular container.

server/webapp/WEB-INF/rails/webpack/views/components/icons/index.scss
```diff
@@ -29,6 +29,7 @@ $btn-group-border-radius: $global-border-radius;
   color:         $icon-color;
   transition:    $transition;
   background:    $icon-bg;
+  margin-left:   20px;
```
GaneshSPatil: @naveenbhaskar -- I am not sure why we need to add `margin-left` to an icon.
An icon simply should have styles to render an icon. 
The spacing styles of the icon are dependent on the place where the icon is used. I would say for margin and padding, we should wrap the icon in a  different container and style that particular container.

GaneshSPatil:
Marking this for `19.4.0` as we're closer to the release.

GaneshSPatil:
@naveenbhaskar -- ping!

server/webapp/WEB-INF/rails/webpack/views/components/icons/index.scss
naveenbhaskar: This particular component is used in buttons which has an icon with it. Where ever we use this, we need consistent margins and that the only use case this can be used. we dont have any other scenarios where we need to use this. 
wrapping  the buttons in another div just for margin is not a good standard to follow. 
In normal scenarios, we use a second class for this purpose and we don't have that functionality in our components. I would recommend to have that rather than adding an extra div for margin.

server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.scss
naveenbhaskar: This particular component is used in buttons which has an icon with it. Where ever we use this, we need consistent margins and that the only use case this can be used. we dont have any other scenarios where we need to use this. 
wrapping  the buttons in another div just for margin is not a good standard to follow. 
In normal scenarios, we use a second class for this purpose and we don't have that functionality in our components. I would recommend to have that rather than adding an extra div for margin.

server/webapp/WEB-INF/rails/webpack/views/components/icons/index.scss
GaneshSPatil: Depending upon the place where the component is used, the margin might change.
Does that mean we need to add multiple secondary classes?

naveenbhaskar: Adding secondary classes is the right solution for this. 
Any styles related to the placement of the component should come from a secondary class and that should go in the page specific stylesheet.

GaneshSPatil: @naveen -- That might open the door for discrepancies in our components.

As there is no **enforcement** of secondary class to only hold placement related styles, the consumers of the component can style the component elements differently as per the convenience. And on each page, the same component might/will look drastically different.

Moreover, if the component styles are split across the component-styles and page-specific-styles, then the extracted piece of code is not truly a component.

naveenbhaskar: We can only define the look and feel of a component with the base class. and placement is based on where we put it. The whole concept of component is to reuse it with base class and extend it with secondary class. 
we should not change the look and feel of the component by the secondary class. it should used only for the placement.

GaneshSPatil: With components, the smallest unit of work is a component, classes and styles are internal to component. If we expose (secondary) classes outside, it might break the component encapsulation.

Not sure if thinking in terms classes is the right approach for the same.
--
@akshaydewan @ketan -- Thoughts? Opinions?

akshaydewan: Hmm, this has been an issue with components, and there are two ways (that I'm aware of) that we've worked around it.

1. Expose a semantic attribute for the component - e.g. the [SwitchBtn](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/webpack/views/components/switch/index.tsx#L37) component has an attribute called `small` which sets the margins and widths appropriately.
However, this may not be possible in all cases.

2. @marques-work had come up with an idea of [injecting a css object in a component](https://github.com/gocd/gocd/pull/6260/files#diff-5e3ba5f5b4060e26ff7e2949086521baR142). This css object will be strongly typed, allowing us to inject or override our own stylesheet in a typesafe manner.

IMO, injecting a secondary css class is too "low-level" for components, and we should avoid it if possible.

akshaydewan: To elaborate a bit on what I mean by "low-level", let's say we take the example of a semantic attribute like `small` or `size` or whatever. The actual implementation of this can be done using a secondary class, but that shouldn't be exposed from the interface of the component, since that can lead to breakages if we modify the original component, without us finding out.

OTOH, the approach of injecting the css, being typesafe, would give us hints of potential breakages if we were to modify the original component's styles. (Well, I hope. I haven't actually tried it out.)

PR #6192, opened 2019-04-24
Remove 'save' calls from 'find' functions

akshaydewan:
The save calls seem to be redundant and would cause problems if a user wants to view the dashboard in readonly mode on a standby server


adityasood:
this works and helps with the current issue on BC. Can be merged

arvindsv:
@marques-work Since this seems to have been added (and not left behind when something else was removed), can you take a slightly deeper look and see if there could be any reason for this?

marques-work:
@arvindsv I'll scrutinize the history more. It's possible that I introduced it to do an opportunistic migration but then forgot to remove it when introducing a formal data migration.

PR #6195, opened 2019-04-24
Restrict stages config_change endpoint to admins only  #000

maheshp:


maheshp:
> Wonder why it's not in admin/stages_controller.rb.

You are right, it should be under admin/stages_controller.rb. Will move it post the release.

PR #6197, opened 2019-04-25
Trying to fix UpdateTemplateConfigCommandTest #000

maheshp:
* Not exactly sure of the reason for this test to fail, it passes
  locally. Removing some reduntant mocks to see if it helps.

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/config/update/UpdateTemplateConfigCommandTest.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/config/update/UpdateTemplateConfigCommandTest.java
@@ -200,14 +200,9 @@ public class UpdateTemplateConfigCommandTest {
 
     @Test
     public void updateTemplateConfigShouldValidateAllExternalArtifacts() {
-        PluggableArtifactConfig s3 = mock(PluggableArtifactConfig.class);
-        PluggableArtifactConfig docker = mock(PluggableArtifactConfig.class);
-        when(s3.getStoreId()).thenReturn("s3");
-        when(docker.getStoreId()).thenReturn("docker");
+        PluggableArtifactConfig s3 = new PluggableArtifactConfig("id1", "s3");
+        PluggableArtifactConfig docker = new PluggableArtifactConfig("id2", "docker");
 
-        ConfigErrors configErrors = new ConfigErrors();
-        when(s3.errors()).thenReturn(configErrors);
-        when(docker.errors()).thenReturn(configErrors);
         JobConfig job1 = JobConfigMother.jobWithNoResourceRequirement();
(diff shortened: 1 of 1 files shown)


bdpiparva:
The pipeline is failing when it runs the test using Java 12 - 

From https://build.gocd.org/go/tab/pipeline/history/build-windows see the runs from 3578 to 3584

3578 - ran with java 12
3579 - ran with java 11
3580 - ran with system JVM which is Java 11 now and so on ...

### Why only this test, not others?

The test is using mocks and those are [cloned](https://github.com/gocd/gocd/blob/a19b3e302f9cd9a65b2ed927d098ca4ef903a7b6/config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfig.java#L685) using the library. Before java 12 deep clone was copying the mocks as it is hence test was working fine. With Java 12 we have noticed that the mock objects are cloned as well and lost the [setup](https://github.com/gocd/gocd/blob/master/server/src/test-fast/java/com/thoughtworks/go/config/update/UpdateTemplateConfigCommandTest.java#L205) done on the object.

PS: we will investigate it further to see how cloning is affected by Java 12.

bdpiparva:
Accessing members(methods and fields) of few classes are restricted in jdk 12. As a result, mock objects are not cloned completely. By looking at the changes that are done in commit(http://hg.openjdk.java.net/jdk/jdk12/rev/f55a4bc91ef4), it will not affect any production code. But all the clone operations on any mocked object will start behaving differently i.e. any stub done on methods will not be cloned hence test might fail.
```
PluggableArtifactConfig s3 = mock(PluggableArtifactConfig.class);
when(s3.getStoreId()).thenReturn("s3"); //  the stub will be lost on deep clone object
```

#### before jdk 12 filter list was -
```
Map<Class<?>,String[]> map = new HashMap<Class<?>,String[]>();
        map.put(Reflection.class,
            new String[] {"fieldFilterMap", "methodFilterMap"});
        map.put(System.class, new String[] {"security"});
        map.put(Class.class, new String[] {"classLoader"});
        fieldFilterMap = map;
```

#### In jdk 12 it changed to -
```
fieldFilterMap = Map.of(
            Reflection.class, ALL_MEMBERS,
            AccessibleObject.class, ALL_MEMBERS,
            Class.class, Set.of("classLoader"),
            ClassLoader.class, ALL_MEMBERS,
            Constructor.class, ALL_MEMBERS,
            Field.class, ALL_MEMBERS,
            Method.class, ALL_MEMBERS,
            Module.class, ALL_MEMBERS,
            System.class, Set.of("security")
        );
```

See issue: https://bugs.openjdk.java.net/browse/JDK-8210496 for more details.

PR #6203, opened 2019-04-25
Migrate Pluggable SCM API to spark

vishaldevgire:


GaneshSPatil:
@bdpiparva will be taking care of this PR.

\cc: @vishaldevgire @maheshp

PR #6218, opened 2019-04-29
Pipeline config api v7: Allow credentials only through attributes

bdpiparva:


adityasood:
@bdpiparva @maheshp would we need to add the branch attribute for material type hg as part of this change

bdpiparva:
@adityasood - Yes, we should add branch attribute and also keep branch name in URL for v6.

Issue #6221, opened 2019-04-29
user management page - roles squashed together


mbotmcc:
##### Issue Type

- Bug Report

##### Summary

On the user management page, if a user is in multiple roles, they're all squashed together making it impossible to read 

##### Environment

gocd 19.3

![image](https://user-images.githubusercontent.com/35934878/56885174-b3a8c700-6a63-11e9-884a-ef1ded7ae9f4.png)

maheshp:
\cc @naveenbhaskar can you please take a look.

naveenbhaskar:
fixed
![image](https://user-images.githubusercontent.com/1627961/56941543-baead600-6b32-11e9-81c5-b0719d3b716c.png)

adityasood:
verified-  19.4.0-9105

PR #6230, opened 2019-04-30
Roles name alignment fix on user management page

naveenbhaskar:
fix for https://github.com/gocd/gocd/issues/6221

image

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/pages/users/index.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/pages/users/index.scss
@@ -206,7 +206,7 @@ $selected-table-row: #e2f7fa;
 
 .table-row {
   border-bottom: 1px solid $border-color;
-  height:        45px;
+  min-height:        45px;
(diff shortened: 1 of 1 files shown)


server/webapp/WEB-INF/rails/webpack/views/pages/users/index.scss
```diff
@@ -206,7 +206,7 @@ $selected-table-row: #e2f7fa;
 
 .table-row {
   border-bottom: 1px solid $border-color;
-  height:        45px;
+  min-height:        45px;
```
GaneshSPatil: Not sure if the indentation is right.
#### Currently, it is:
Screen Shot 2019-05-06 at 10 40 26 AM


#### How about:
Screen Shot 2019-05-06 at 10 40 37 AM

?

---

Setting indentation settings of scss to align on values (in IntelliJ) will do the trick :)
Screen Shot 2019-05-06 at 10 42 04 AM

PR #6232, opened 2019-04-30
[Reattempt] Encrypt publish-artifact secure-plugin-properties when configured from admin/cruise-config.xml tab

GaneshSPatil:
* Make cruise-config representation consistent for config on disk and GoConfigHolder.

Problem:
* As part of full-config-save from UI, GoCD converts user submitted XML into the string
  and then sends the original config object for preprocessing and validation.
* As the converting XML to string happens before preprocessing is done, the string
  XML will not have plugin properties encrypted.
* Though, upon preprocessing, the config object with encrypted plugin properties
  is populated on GoConfigHolder. This causes the XML on disk and GoConfigHolder
  to be out of sync.

Fix:
* Convert XML to string after validation and preprocessing of cruise-config.xml

Fixes issue: https://hackerone.com/bugs?report_id=548755

### DIFF
--- a/server/src/test-integration/java/com/thoughtworks/go/config/FullConfigSaveFlowTestBase.java
+++ b/server/src/test-integration/java/com/thoughtworks/go/config/FullConfigSaveFlowTestBase.java
@@ -42,9 +42,7 @@ import java.util.ArrayList;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.asList;
-import static org.hamcrest.Matchers.nullValue;
-import static org.hamcrest.Matchers.startsWith;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.assertThat;
 
 @RunWith(SpringJUnit4ClassRunner.class)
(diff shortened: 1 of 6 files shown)


server/src/test-integration/java/com/thoughtworks/go/config/FullConfigSaveFlowTestBase.java
```diff
@@ -91,17 +89,23 @@ public void shouldEncryptPluginPropertiesOfPublishTask() throws Exception {
...
-        assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getValue(), is("SECRET"));
+
+        assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getValue(), is("IMAGE_SECRET"));
         assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getEncryptedValue(), startsWith("AES:"));
         assertThat(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getConfigValue(), is(nullValue()));
+
+        //verify xml on disk contains encrypted Image plugin property
+        assertThat(configHelper.getCurrentXml(), containsString(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getEncryptedValue()));
+        //verify xml from GoConfigHolder contains encrypted Image plugin property
+        assertThat(getImplementer().toXmlString(configHolder.configForEdit), containsString(ancestorPluggablePublishAftifactConfigAfterEncryption.getProperty("Image").getEncryptedValue()));
```
varshavaradarajan: Add similar tests for fetch?

GaneshSPatil: @varshavaradarajan -- This particular test verifies whether the preprocessed and the configForEdit both are updated with the changeset.

And verifying one of the plugin property encryption on preprocessed and the configForEdit is good enough.

We can rely upon the entity level encryption specs and `config.encryptSecureProperties` encryption specs for the fetch plugin property encryptions.

---

Also,
If we add the test of fetch plugin properties here. Later, whenever the new GoCD entity requires plugin property encryption, we will need to add it here too.
Eg: for role-config, elastic-agent-profiles, etc. and

varshavaradarajan: The very next text is for fetch. :)

PR #6235, opened 2019-04-30
Secret configuration SPA

kritika-singh3:
Added toggle for Secret Config SPA. The default toggle value is `false`.
To turn it on:
```bash
curl http://localhost:8153/go/api/admin/feature_toggles/show_secret_config_spa -d toggle_value=on -H 'Confirm:true' -u admin:badger
```


kritika-singh3:
Preview:
secret_config_spa

arvindsv:
@kritika-singh3 Can you add a placeholder for the "Id" field? Also, should the file path itself be encrypted? What do we gain from that? It would end up making it confusing, won't it?

kritika-singh3:
> Can you add a placeholder for the "Id" field? As in `<input placeholder="..."/>`. 

Will do.

> Also, should the file path itself be encrypted? What do we gain from that? It would end up making it confusing, won't it?

I agree. Will set  `secure=false` in the file plugin.

kritika-singh3:
Updated Preview:
secret_config_spa

PR #6249, opened 2019-05-06
Introduce a new class to store naturalOrderPmm and scheduleOrderPmm per pipeline

varshavaradarajan:
* This is done so that all PTEs (PipelineTimelineEntries) need not be unnecessarily loaded in memory at startup.
* Earlier, PTEs for even deleted pipelines would be loaded from the db.
* This loads PTEs for specified pipeline on demand.

To test:

- [ ] Server startup time
- [ ] Performance of the new query (pg and h2)
- [ ] Performance while scheduling a pipeline with loads of for the first time (pg and h2)
- [ ] Memory profile (is eviction needed ?)
- [ ] Does this break fan-in? (shouldn't as this is mainly refactoring)
- [ ]  Ask @arvindsv what else to test for


GaneshSPatil:
@varshavaradarajan @ketan @arvindsv -- ping!!

varshavaradarajan:
@GaneshSPatil - will work on it this month. Been busy

Issue #6095, opened 2019-04-08
Support for Rules in SecretConfig


maheshp:
Part of #5830 

GoCD System admins can defines rules for a SecretConfig. Rules define where the secrets configuration can be used. e.g. 

> Only pipeline group `teamA_pipelines` can access following secret config

```xml
<secretConfig id="teamA_secrets" pluginId="vault_based_plugin">
    <description>All secrets for env1</description>
     <configuration>
        <property>
            <key>path</key>
            <value>secret/dev/teamA</value>
        </property>
        <property>
            <key>vault_token</key>
            <value>token for team A</value>
        </property>
    </configuration>
    <rules>
        <allow action="refer" type="pipeline_group">teamA_pipelines</allow>  
    </rules>
</secretConfig>
```

In the above example, consider a GoCD server shared by multiple teams in an organization. A secret config can be defined to lookup for secrets for teamA from an external Secrets Manager like a vault. The SecretConfig can also be configured to have rules using which Admins can restrict the usage of the secrets config to pipeline_groups specific to the teamA. 

### Types of rule

GoCD will have two types of rule for now - 

1. Allow
   - The rule will allow the access of the current entity based on configured type, action and resource

2. Deny
    - The rule will deny the access of the current entity based on configured type, action and resource

### Rule syntax

> Allow rule example

```xml
<allow type="pipeline_group" action="refer">resource_identifier</allow>
```

> Deny example

```xml
<deny type="environment" action="refer">resource_identifier</allow>
```

|          | Description | Example |
| -------- | ----------- | ------- |
| type     | The attribute represents the type of GoCD config entity.                     | wildcard `*` or `pipeline_group`, `environment`, `template`, `agent` etc... |
| action   | The attribute represents that what type of action is allowed on the current entity. | wildcard `*` or Action `refer` on secret config means that it can be referred by the entities with matching `type` and `resource` |
| resource | A resource name or pattern with a wildcard(`*` or `?`) in it                                                  | `pipeline_test`, `environment_prod` or a patter like `pipeline_*` |

- `action` and `type` can have a wildcard(`*`) but it will not support pattern matching e.g. `ref*` 

**Note:** *`type` and `action` can be extended later on need basis to different entity types.*

### Rule evalution

1. Rules are denied by default
    - If rules are not defined on the entity(e.g. on secret config) cannot be referred anywhere.

> The secret config defined in below example cannot be referred anywhere

```xml
[...]

PR #6253, opened 2019-05-06
Update secret config UI to consume rules

kritika-singh3:
- Added support to define rules while add/update/clone secret configs
- Showcased the existing rules
- Added the information about wildcard support in resources for rules

### TODO

- [x] support for autocomplete in `resource` based on `type`
- [x] drag and drop feature for rearranging rules order


kritika-singh3:
Preview:
secret-config-spa

server/webapp/WEB-INF/rails/webpack/views/pages/secret_configs/rules_widget.tsx
```diff
@@ -0,0 +1,189 @@
...
+          dataTestId="rule-type"
+          property={rule().type}
+          required={true}
+          onchange={(e) => provider().update()}
+          errorText={rule().errors().errorsForDisplay("type")}>
+          <SelectFieldOptions selected={rule().type()}
+                              items={RulesWidget.types()}/>
+        </SelectField>,
+        <AutocompleteField
+          key={rule().type()}
```
marques-work: I don't think `key` here is meaningful in this context; it's not wrong, but it also doesn't affect the end result, as it really only matters on the outer reorderable set of children. I think you told me you attached a `key` attribute to the outer `<tr/>`, which seems like the right choice.

marques-work: With regards to the `<Table/>` component, I did see that `key` on the `<tr/>` was set to the `index` of the row. I may be mistaken as I'm basing this off a thought-experiment, but `index` may not be an effective key because it does not "follow" the data it is associated with; that means the `key` will always be reset to its current position rather than being associated with a specific piece of data, *I think*. `key` (to my understanding) is supposed to be the link between the underlying model and a given DOM node when elements are reorderable. Tying `key` to the `index` of a row might defeat the purpose.

I know I'm seeing some unintended behavior that *might* be related to this. For instance, when reordering rows while the autocomplete popup is open, the value of the `<input/>` reflects the reordering, but the popup selection does not, meaning mithril did not actually reorder the DOM, but just replaced the `<input/>` value, which sounds like it's related to a bad `key`, but that's just a guess.

kritika-singh3: > I don't think `key` here is meaningful in this context; it's not wrong, but it also doesn't affect the end result, as it really only matters on the outer reorderable set of children. I think you told me you attached a `key` attribute to the outer `<tr/>`, which seems like the right choice.

This key attached over here is basically to force `mithril` to re-render the autocomplete text field when the provider updates.

kritika-singh3: > With regards to the `<Table/>` component, I did see that `key` on the `<tr/>` was set to the `index` of the row. I may be mistaken as I'm basing this off a thought-experiment, but `index` may not be an effective key because it does not "follow" the data it is associated with; that means the `key` will always be reset to its current position rather than being associated with a specific piece of data, _I think_. `key` (to my understanding) is supposed to be the link between the underlying model and a given DOM node when elements are reorderable. Tying `key` to the `index` of a row might defeat the purpose.

I know that `index` isn't the right context for key, but there isn't any `identifier` associated with the data in the component as the input is `m.Child[][]`.

kritika-singh3:
> I did notice some strange behavior on reordering the rows while the autocomplete popup is open. I suspect this might be `key` related (see comments on using `index` for the `key` value in the `<tr/>`).
> 
> Also, when you open an existing secretsconfig for edit, I think it's strange that all the autocomplete popups are open for each rule row. I don't think that should be the case -- the autocomplete popups should start closed when you're editing an existing secretsconfig, IMO. I know this is an artifact of how Awesomplete works with `minChar`. Still, we should probably find a way to start with the popups in a closed state for a better user experience.

I also tried to look at this issue but will look again.

kritika-singh3:
Updated Preview:

secret-config-rules-ui

arvindsv:
Thank you for sharing this! :)

Thoughts:

1. Are the values glob patterns or regexs?

2. Can we change the text: "Configuring Rules is required to utilize this Secret Configuration. In absence of any rules, the secret configuration is denied access of any GoCD entities." to "The default rule is to deny access to this secret configuration to all GoCD entities. Configure rules below to override that behavior"? We should also point to a documentation page which explains rules and that order matters.

3. Can we change the placeholder of the ID to: "Enter any unique identifier"?

kritika-singh3:
@arvindsv 
> 1. Are the values glob patterns or regexs?

The values are not regexs for sure. The value can be the name of the entity with a wildcard or just a wildcard. Not sure it comes under glob patterns or not.

> 2.  "The default rule is to deny access to this secret configuration to all GoCD entities. Configure rules below to override that behavior"?

Should it be `The default rule is to deny access to this secret configuration for all GoCD entities. Configure rules below to override that behavior.`?

bdpiparva:
@arvindsv, @kritika-singh3  - Allowed patterns in rule is documented in [issue](https://github.com/gocd/gocd/issues/6095) under `Rule syntax`

kritika-singh3:
Update preview:
secret-config-rules-ui

kritika-singh3:
Updated Preview:
secret-config-rules-ui1

arvindsv:
In this image, the secrets shown will still not be usable, since `deny * *` is the first. Right? I know it's just an example to show the design. :) But, when we add it to release notes, we should add an animation with a secret config that works (as a reminder).

Issue #6132, opened 2019-04-12
Normalised URL for material configuration in GoCD


kritika-singh3:
### Issue Type
- Feature proposal
- Feature enhancement

### Issue

Currently, there are different ways in which the URL for a material can be specified. In case of Git and Mercurial - the credentials are provided in the URL itself. But for P4, SVN and TFS - we have additional attributes of `username` and `password` for the supplying the credentials.

This can be pretty confusing as to why the different configuration support URL format differently.

### Solution

Adding the attributes of `username` and `password` for Git and Mercurial configurations will help bring about the consistency among them. In order to maintain backward compatibility, the existing URLs for the mentioned configs will be normalised.

By normalised, we mean that the URL will be parsed and the credential part(userInfo) of the URL extracted out and stored in the attributes during config migration. 

#### git
   * User info in git url will be migrated to `username` and `password` attributes
       E.g.
        Original: `<git url="https://username:password@github.com" />`
        Updated: `<git url="https://github.com" username="username" password="password" />`

#### hg
   * User info in git URL will be migrated to` username` and `password` attributes
   * Additionally, if URL contains branch(specified after ##) will be migrated to branch attribute on the material config
       E.g.
        Original: `<hg url="https://username:password@domain.com##feature" />`
        Updated: `<hg url="https://domain.com" username="username" password="password" branch="feature" />`

### Normalization of Materials in DB
For consistency, we will have to normalize material URL's in db as well. Therefore the noramalization applied to material URL's in config will be applied to the db materials as well.

### Validation
Post-normalization, 
  - userInfo will not be allowed in URL's of both git and hg materials.
  - Branch in hg material URL will not be supported

#### git 
Prior to normalization, the fingerprint generation is a function of 
 - url
 - branch

Post-normalization
 - url (without userInfo)
 - username
 - branch

#### hg 
Prior to normalization, the fingerprint generation is a function of 
 - URL (which can have branch specified)

Post normalization
 - url (without userInfo and branch)
 - username
 - branch

### Config repo

Going forward, if the user specifies the credentials in the URL attribute (for the case of config-repo), the same normalization technique will be applied during save of the said configuration and the credentials will be stored in the separate attributes. 

### Breaking change

- Will be updated soon

arvindsv:
Remember that: `C:/abc/def.git` and `C:\abc\def.git` are both valid URLs, as are `/abc/def.git` and `file:///abc/def.git`.

maheshp:
[...]

PR #6257, opened 2019-05-07
Support for username, password and encrypted password in config repo material contract.

vrushaliwaykole:
#6132


arvindsv:
@vrushaliwaykole Wasn't there any migration needed for this? What happens to existing repos with the username and password in the URL?

vrushaliwaykole:
> @vrushaliwaykole Wasn't there any migration needed for this? What happens to existing repos with the username and password in the URL?

It is handled in config repo migration using `format_version` so I don't think explicit migration is required.

PR #6258, opened 2019-05-07
Add autocomplete text field

kritika-singh3:
Added component - AutocompleteTextField which takes in an array of possible values

- values shown will be filtered on the basis of the value in the text box


GaneshSPatil:
Just was wondering if it is same as of https://github.com/gocd/gocd/commit/a324baf0c94d5ef9d67131f09c91a221d9073e7c

kritika-singh3:
Closing as a duplicate.

PR #6263, opened 2019-05-08
Allow the ability to programmatically change autocomplete suggestions in AutocompleteField

marques-work:
SuggestionProvider now allows a programmatic update/refresh of autocomplete suggestions

  - Changed getData() signature to return a promise
  - AutocompleteField registers handlers on provider
  - SuggestionProvider implementations expose an `update()` method that will invoke `getData()`
    perform updates to the suggestion list when used in AutocompleteField. As long as one has
    a handle on the provider object, one can trigger `update()` from, say, an event handler or
    timer


marques-work:
@kritika-singh3 may be interested in this; do let me know if you have implementation questions or otherwise. This should allow you to implement a dynamic suggestion provider if I understood correctly.

kritika-singh3:
@marques-work Yes, it accepts a dynamic suggestion provider. Thanks!

PR #6272, opened 2019-05-08
Yet even more UI updates for the new Add Pipeline page

marques-work:
Mostly textual changes and minor styling.

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
@@ -54,6 +54,7 @@ $form-failed-field-background-color: lighten($form-failed-color, 35%);
   height:           $form-element-height;
   line-height:      $form-element-height;
   padding:          0 10px;
+  margin-right:     5px;
   font-size:        14px;
   background-color: $form-control-background-color;
   background-clip:  padding-box;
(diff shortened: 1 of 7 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/forms/forms.scss
```diff
@@ -113,7 +114,8 @@ input[type="checkbox"].form-control { //sass-lint:disable-line no-qualifying-ele
 }
 
 .form-input-reset {
-  margin: 5px 0 0 0;
+  margin: 5px 0 0;
+  display: inline-block;
```
marques-work: Doesn't alter its appearance/layout on the Config Repos modal (due to the width of the modal), so no breaking change here.

But it looks more intuitive on the Add Pipelines page.

server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
```diff
@@ -36,7 +36,7 @@ export interface RequiredFieldAttr {
 }
 
 interface LabelAttr extends RequiredFieldAttr {
-  label?: string;
+  label?: m.Children;
```
marques-work: We should be able to embed tooltip components and such in our labels. This should not break any existing code as `string` satisfies `m.Children`

PR #6274, opened 2019-05-08
[test drive] Add env vars to job adv settings on new add pipeline page 

ibnc:


server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/job.ts
```diff
@@ -15,31 +15,41 @@
...
 import {Task} from "models/pipeline_configs/task";
+import * as s from "underscore.string";
 
 export class Job extends ValidatableMixin {
   name: Stream<string>;
+  environmentVariables: Stream<EnvironmentVariableConfig[]>;
   tasks: Stream<Task[]>;
 
-  constructor(name: string, tasks: Task[]) {
+  constructor(name: string, tasks: Task[], envVars: EnvironmentVariableConfig[]) {
```
marques-work: Make optional? i.e., `envVars?: EnvironmentVariableConfig[]`. Perhaps consider storing as `NameableSet` since these should be unique by name.

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/job.ts
```diff
@@ -15,31 +15,41 @@
...
     this.validatePresenceOf("tasks");
     this.validateNonEmptyCollection("tasks", {message: "A job must have at least one task"});
     this.validateEach("tasks");
+    this.validateEach("environmentVariables");
   }
 
   toApiPayload() {
-    return JsonUtils.toSnakeCasedObject(this);
+    const raw = JsonUtils.toSnakeCasedObject(this);
+    raw.environment_variables = _.filter(this.environmentVariables(), (variable) => {
```
marques-work: Instead, shouldn't we rely on the validation that an `EnvironmentVariableConfig` cannot have a blank name? We _shouldn't_ have to do this, unless I'm missing something nuanced.

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/spec/job_spec.ts
```diff
@@ -63,4 +65,21 @@ describe("Job model", () => {
     });
   });
 
+  it("should exclude env vars with no name when serializing", () => {
```
marques-work: We shouldn't allow this case for a valid job. Validate on `EnvironmentVariableConfig`.

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/environment_variable_config.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor(secure?: boolean, name?: string, value?: string) {
+    super();
+    this.name = stream(name);
+    this.value = stream(value);
+    this.secure = stream(secure);
+    ValidatableMixin.call(this);
+  }
+
+  validate(key?: string) {
```
marques-work: Rather than overriding `validate()`, aren't we able to compose a set of rules with the methods provided by `ValidatableMixin`? For fields that are dependent on the value of others (perhaps `secure + (encryptedValue ^ value)`), could we not write a validator for this and use `this.validateWith()`? It also seems odd that we would allow a blank `name`. That doesn't seem meaningful for an environment variable.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/environment_variables_editor.tsx
```diff
@@ -0,0 +1,89 @@
...
+                );
+              })
+            }
+          </div>
+        </Form>
+      </FormBody>
+    );
+  }
+
+  addVar(secure: boolean, variables: EnvironmentVariableConfig[], event: Event) {
```
marques-work: What about removing entries?

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/job.ts
ibnc: Yeah, I'm not a fan of doing this either. The reason it's necessary is because when have 2 env vars that are on the page when it loads (one empty secure and one empty plain). They don't have to fill them in to submit the form.

ibnc: Those vars would be considered invalid if we validate the presence of the name. I couldn't think of a good way to have those who vars be on the page on load, and have the validations be a bit more standard. 

Does that makes sense?

marques-work: @ibnc I know we chatted on slack, but going to copy responses here for documentation.

I understand the reasoning now.

What I would opt for is to create a `stream` to hold unsaved values and have the component convert them into `EnvironmentVariableConfig` objects `onchange`.

That way you can have the right validations on environment variables, and not have to worry about this case where we leave the form unfilled while validating `EnvironmentVariableConfig` instances.

Kind of analogous to how the task editor starts with an empty task array and only creates Task instances after typing some text and pressing `enter`.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/environment_variables_editor.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Form last={true} compactForm={true}>
+          <div class={css.envVars}>
+            <label>
+              Environment Variables
+              <Buttons.Cancel onclick={this.addVar.bind(this, false, vnode.attrs.variables)} small={true} icon={Buttons.ButtonIcon.ADD} />
+            </label>
+            {
+              _.map(plainVars, (variable) => {
+                return (
+                  <div class={css.envVar}>
```
marques-work: To me, a `<table/>` seems like the appropriate semantic element. If you had `Name` and `Value` columns, you wouldn't need to have labels repeating on each row. Tidies things up a bit. Ditto with the secure variables.

marques-work: To me, a `<table/>` seems like the appropriate semantic element. If you had `Name` and `Value` columns, you wouldn't need to have labels repeating on each row. Tidies things up a bit. Ditto with the secure variables.

marques-work:
@ibnc sorry I didn't get a chance to look at this yet. I'll review maybe tonight, or definitely tomorrow.

marques-work:
I'm going to merge this because it poses low (no?) risk to the upcoming release; all of the files are JS for the new add pipeline experience, which is behind a toggle anyway and is of no consequence to anything else.

PR #6277, opened 2019-05-09
Modified table component to allow rows to be rearranged via drag-n-drop

kritika-singh3:
- table component will take in an additional attribute `draggable`
- when set to `true`, a drag icon will allow the user to drag and drop the rows


kritika-singh3:
Preview:

draggable-table-component

kritika-singh3:
Issue: the drag and drop events are handled outside of the mithril lifecycle. 
Hence, the view is not up-to-date with the model.

Solution:
Updated table component for the same purpose utilizing html5 attributes of `draggable` and drag events.
PR: https://github.com/gocd/gocd/pull/6285

PR #6283, opened 2019-05-10
#5427 - Change TFS workspace default to be SERVER rather than null

arvindsv:
See: https://github.com/gocd/gocd/issues/5427#issuecomment-468120891

For #5427, the plan was to change the default value of the workspace location to be SERVER, after a while. This change does that.


arvindsv:
Looks like all PR pipelines ran and suceeded. Merging.

PR #6290, opened 2019-05-15
Use AdoptOpenJDK's JRE for windows installers.

ketan:


adityasood:
verified with installer go-agent-19.4.0-9111 and go-server-19.4.0-9111.

Scenarios:

1. Fresh installation with packaged JRE for both server agent
2. Upgrade on an existing installation.

Issue #6295, opened 2019-05-16
Improve errors UI in Config Repo


akshaydewan:
The Config Repo UI shows exception details (from Git) which may not make sense to users. It also contains text like `<br/>` which is not parsed properly.

![Screenshot 2019-05-16 at 11 42 01](https://user-images.githubusercontent.com/6024038/57830595-35248700-77d0-11e9-816a-cb0e4673889d.png)

It may be better to show more error details somewhere where the user can drill down, because this can get rather large and unwieldy.

PR #6297, opened 2019-05-16
Use m.trust for showing config repo errors

akshaydewan:
MDU errors may contain tags like `<br/>` which are otherwise not rendered.

related to #6295


ketan:
Should we not fix the MDU to have this message? It's a potential for someone to inject code potential XSS by altering error messages which may contain input strings.

akshaydewan:
> Should we not fix the MDU to have this message? It's a potential for someone to inject code potential XSS by altering error messages which may contain input strings.

That would be ideal. We are also using `m.trust` in the server health messages modal. If it's a potential attack vector, then we should remove `m.trust` there as well.

akshaydewan:
Closing this. We'll figure out a better way to fix it.

PR #6307, opened 2019-05-21
Restyled the new pipeline buttons to be primary buttons 

kierarad:
Behind toggle ```add_pipeline_button_styling```

TBD whether to change in production, but want a more compelling CTA for trial

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
@@ -30,6 +30,8 @@ export enum ButtonIcon {
   FILTER,
 }
 
+export type Alignment = "right" | "left";
+
 export interface OnClickHandler {
   onclick?: (e: MouseEvent) => void;
 }
@@ -43,6 +45,7 @@ export interface Attrs extends OnClickHandler, Restyleable<Styles> {
   small?: boolean;
   disabled?: boolean;
   dropdown?: boolean;
+  align?: Alignment;
 }
 
 abstract class Button extends MithrilViewComponent<Attrs> {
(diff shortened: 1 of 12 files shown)


kierarad:
[Style new pipeline buttons to be more of a CTA](https://trello.com/c/g6MNhWNe/105-style-new-pipeline-buttons-to-be-more-of-a-cta)

arvindsv:
@kierarad Can you add an image / animation to this PR to show how it looks?

server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
```diff
@@ -30,6 +30,11 @@ export enum ButtonIcon {
   FILTER,
 }
 
+export enum Alignment {
```
marques-work: @kierarad a simpler alternative to using a TypeScript `enum` is to just do:

```typescript
export type Alignment = "right" | "left";
```

Then, you can just pass in the string literals `"right"` and `"left"` to the `vnode.attrs` object instead of `Alignment.right`, etc. TypeScript will still validate (and, in fact, fail to compile) if the values are not one of the options.

This becomes:

```jsx
<Buttons.Primary align="right" ... />
```

And the `alignClass()` implementation becomes:

```typescript
private static alignClass(align?: Alignment, css?: Styles) {
  css = css || defaultStyles;

  switch (align) {
    case "left":
      return css.leftAlign;
    case "right":
      return css.rightAlign;
  }
  return "";
}
```

marques-work:
@kierarad Implementation looks fine - I made a suggestion in a comment above; totally optional as using an `enum` is also a valid approach, so I'll leave it up to you on which approach you decide. The only thing I'd request is to maybe squash the commits, and then I'll merge this.

server/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
kierarad: @marques-work thanks! I actually prefer this as well. I'll change

marques-work: @kierarad cool :)

kierarad:
For reference here is a pic of the new styles
Screen Shot 2019-05-22 at 3 24 22 PM

PR #6308, opened 2019-05-21
[Experiment] Add universal add pipeline button to dashboard

kierarad:
- No group is passed, so an option for a group to be selected needs to be added to the new flow in order for the add pipeline to succeed

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/support/toggle/Toggles.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/support/toggle/Toggles.java
@@ -22,6 +22,7 @@ public class Toggles {
     public static String NEW_ADD_PIPELINE_FLOW = "new_add_pipeline_flow";
     public static String ADD_PIPELINE_BUTTON_STYLING = "add_pipeline_button_styling";
     public static String ALLOW_EMPTY_PIPELINE_GROUPS_DASHBOARD = "allow_empty_pipeline_groups_dashboard";
+    public static String UNIVERSAL_ADD_PIPELINE_BUTTON = "universal_add_pipeline_button";
     public static String SHOW_SECRET_CONFIG_SPA = "show_secret_config_spa";
     public static String PLUGIN_SETTINGS_API_USING_RAILS = "plugin_settings_api_using_rails";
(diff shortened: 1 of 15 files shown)


kierarad:
[Can we have a universal add pipeline button? (vs an add pipeline button per group)](https://trello.com/c/zXPOe6Ux/40-can-we-have-a-universal-add-pipeline-button-vs-an-add-pipeline-button-per-group)

arvindsv:
@kierarad Can you add an image / animation to this PR to show how it looks?

arvindsv:
I do like the add pipeline button per group ... but I'm willing to have my mind changed. For existing users, it'll feel more natural to have the add pipeline button at the group level (my hypothesis).

streisguth:
I think if we have an 'Add Pipeline' action outside of the context of a group, then I imagine we'll just have to make the Pipeline Group a required field, likely somewhere in the context of 'Pipeline Name' on the 'Add Pipeline' screen.

arvindsv:
Yes, that's what we do today. To me, it's less about what we will need to do if we put a single button than about what feels more natural to users. Whether it's more natural to look at a bunch of pipelines and say, "I need a new pipeline here. Let me click that button. Let me go select the group of the pipeline in a dropdown. Now, I can do the rest" vs. "I need a new pipeline here. Let me click the add pipeline button for that group. Now I can do the rest".

marques-work:
@kierarad I would think that enabling this would hide the "add" buttons on the groups. Are you planning to implement that? I'm assuming this is still WIP.

image

@arvindsv also, that's what it looks like :) ☝️

marques-work:
@streisguth @arvindsv @kierarad do we want to also expose pipeline group at the top-level of the add form? Currently, it's an advanced option, so the user won't see it unless they know what they are looking for. Bringing it up top would force the user to consider which group they want to add a pipeline to, though for new users, this may be a new concept.

arvindsv:
@marques-work If it's a global button, then it might make sense to bring the group out of advanced. If it's a group-level one, I think it's fine to leave it in advanced.

kierarad:
Closing this as some user research indicates adding pipelines within a specific group seems to be more natural for users

Issue #3924, opened 2017-10-08
Add Versioning to a new Material endpoint


drewsonne:
##### Issue Type

Improvemet

- Feature enhancement

##### Summary

To address #3880 the materials resource needs to support `If-Match` on write and `ETag` on read methods.

* Go Version: `> 17.10`
* JAVA Version: `Any`
* OS: `Any`
* Browser vendor and version (if relevant): `API, so not applicable`

##### Possible Fix

A new resource will be added with the following requests:

 - `GET /go/api/materials/:fingerprint/`

This will have an ETag hash added to the response, which will be based on the contents of the [material object](https://api.gocd.org/current/#the-pipeline-material-object), but will exclude any attributes which affect the integrity of the `:fingerprint`. Any changes to `:fingerprint` will be considered a new resource.

As there is no central guidance for the implementation of additions or modifications to the GoCD API (see "WIP" in https://developer.gocd.org/current/4/4.2.html#4211-apis-cctray--feeds), a part of this feature will be an addition of information for future development on the API. In other words, add some documentation to the above link.

In the absence of said central guidance, the [environment resource](https://api.gocd.org/current/#environment-config) will be used as a guideline for the extension of this API.

##### Notes

As the external driver for this issue is https://github.com/drewsonne/terraform-provider-gocd/issues/32 there is a case for 

 - `PUT /go/api/materials/:fingerprint/`
 - `PATCH /go/api/materials/:fingerprint/`
 - `POST /go/api/materials/`

but I am not sure of the viability of these at the moment as there may be internal integrity restrictions on this.

PR #6324, opened 2019-05-23
Spark implementation of Materials API

akshaydewan:
~~- Get materials moved from `/go/api/config/materials` to `/go/api/materials`, because it lists all materials, including the ones from config repo. (The documentation would have you believe it just lists the ones from cruise-config.xml)~~
Edit: Kept the endpoint `/go/api/config/materials`, because we also have material instances.
- Moved `PaginationRepresenter` to api-base
- Notify urls changed to `/go/api/admin/materials/svn/notify` instead of `/go/api/admin/material/notify/svn` to make it consistent with webooks
- Using the plural `/materials` everywhere to make it consistent with other endpoints implemented in spark
- `/notify` payload and response changed to json instead of POST data
- Show all the material properties (similar to pipeline config API) in the GET endpoint for materials
- Notify svn using URL to make it consistent


api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/MaterialsAdminControllerV1.java
```diff
@@ -0,0 +1,117 @@
...
+            post(Routes.Materials.NOTIFY.GIT, mimeType, this::gitNotify);
+            post(Routes.Materials.NOTIFY.HG, mimeType, this::hgNotify);
+            post(Routes.Materials.NOTIFY.SCM, mimeType, this::scmNotify);
+            exception(HttpException.class, this::httpException);
+        });
+    }
+
+    public String svnNotify(Request req, Response res) throws IOException {
+        JsonReader jsonReader = GsonTransformer.getInstance().jsonReaderFrom(req.body());
+        String uuid = jsonReader.optString("uuid")
```
bdpiparva: Any specific reason to use `optString` over `getString`?

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/MaterialsAdminControllerV1.java
```diff
@@ -0,0 +1,117 @@
...
+import spark.Response;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static spark.Spark.*;
+
+@Component
+public class MaterialsAdminControllerV1 extends ApiController implements SparkSpringController {
```
bdpiparva: Maybe a different controller name?

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/MaterialsAdminControllerV1.java
```diff
@@ -0,0 +1,117 @@
...
+        JsonReader jsonReader = GsonTransformer.getInstance().jsonReaderFrom(req.body());
+        String uuid = jsonReader.optString("uuid")
+                .orElseThrow(() -> new BadRequestException("Could not find the UUID in the payload"));
+        return notify(req, res, "svn", "uuid", uuid);
+    }
+
+    public String gitNotify(Request req, Response res) throws IOException {
+        JsonReader jsonReader = GsonTransformer.getInstance().jsonReaderFrom(req.body());
+        String repoUrl = jsonReader.optString("repository_url")
+                .orElseThrow(() -> new BadRequestException("Could not find the repository_url in the payload"));
```
bdpiparva: same as above

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/representers/ModificationsRepresenter.java
```diff
@@ -0,0 +1,56 @@
...
+            };
+        }
+        return modificationsWriter -> modifications.forEach(modification ->
+                modificationsWriter.addChild(ModificationRepresenter.toJSON(modification)));
+    }
+
+    private static Consumer<OutputLinkWriter> links(String fingerprint) {
+        return outputLinkWriter -> outputLinkWriter.addLink("self", Routes.Materials.modification(fingerprint))
+                .addAbsoluteLink("doc", Routes.Materials.DOC)
+                .addLink("find", Routes.Materials.FIND_MODIFICATIONS);
```
bdpiparva: Find link does not have offset when specified. Need to verify it with other entity having offset.

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/MaterialsAdminControllerV1.java
akshaydewan: Changed occurrences to `getString`

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/representers/ModificationsRepresenter.java
akshaydewan: I couldn't find any APIs that have a `find` link with an offset. We have pipeline and stage history APIs, but they are old, and don't have the `find` link. There is a recently updated StageRepresenter in `dashboard-v4`, but it doesn't have a `find` link.

We can take a call on whether to include it, and how to represent an optional parameter in the link if we do (e.g `/api/materials/:fingerprint/modifications/[:offset]`?)

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/MaterialsAdminControllerV1.java
akshaydewan: Does `MaterialsNotifyControllerV1` sound better? The base path is `/api/admin/materials`, so there's a possibility that endpoints other than `/notify` may end up in this controller.

bdpiparva: `MaterialsNotifyControllerV1` sounds better. Agreed that there is a possibility that the controller can have other endpoints, I think whenever we add new endpoint at that time we can change it to something else. - WDYT?

bdpiparva:
Related issue: https://github.com/gocd/gocd/issues/3924

api/api-materials-v1/src/main/java/com/thoughtworks/go/apiv1/materials/representers/ModificationsRepresenter.java
akshaydewan: Looking at some examples in the [spec](https://tools.ietf.org/html/draft-kelly-json-hal-08), it seems that optional params are represented in `{}`, so the link would look like `/api/materials/:fingerprint/modifications/{offset}`

akshaydewan:
Further changes done on this API:

## GET materials
```bash
curl -u admin:badger -H Accept:application/vnd.go.cd.v1+json http://localhost:8153/go/api/config/materials         
```
```json
{
  "_links" : {
    "doc" : {
      "href" : "https://api.gocd.org/19.5.0/#materials"
    },
    "self" : {
      "href" : "http://localhost:8153/go/api/config/materials"
    }
  },
  "_embedded" : {
    "materials" : [ {
      "type" : "git",
      "fingerprint" : "458ce82419bdbe41a4007ad713a730e5e467ed4fbefb421bf58aeff3e9c36e4b",
      "attributes" : {
        "url" : "https://github.com/akshaydewan/arduino",
        "destination" : null,
        "filter" : null,
        "invert_filter" : false,
        "name" : null,
        "auto_update" : false,
        "branch" : "master",
        "submodule_folder" : null,
        "shallow_clone" : false
      }
    }, {
      "type" : "svn",
      "fingerprint" : "6b0cd6b9181434866c555f3c3bb780e950389a456764c876d804c848efbad554",
      "attributes" : {
        "url" : "https://github.com/akshaydewan/arduino",
        "destination" : "ads",
        "filter" : null,
        "invert_filter" : false,
        "name" : null,
        "auto_update" : false,
        "username" : "admin",
        "encrypted_password" : "AES:lTFKwi5NvrlqmQ3LOQ5UQA==:Ebggz5N27w54NrhSXKIbng==",
        "check_externals" : false
      }
    }, {
      "type" : "hg",
      "fingerprint" : "f6b61bc6b33e524c2a94c5be4a4661e5f1d2b74ca089418de20de10b282e39e9",
      "attributes" : {
        "url" : "ssh://hg@bitbucket.org/akshaydewan86/gocd-ssh-hg",
        "destination" : "sadasd",
        "filter" : null,
        "invert_filter" : false,
        "name" : null,
        "auto_update" : false,
        "branch" : ""
      }
    }, {
      "type" : "dependency",
      "fingerprint" : "21f7963a8eca6cced5b4f20b57b4e9cb8edaed29d683c83621a77dfb499c553d",
      "attributes" : {
        "pipeline" : "up42",
        "stage" : "up42_stage",
        "name" : "up42_material",
        "auto_update" : true
      }
    } ]
  }
}
```

## POST to notify a material

```bash
curl -u admin:badger -H Content-Type:application/json -H Accept:application/vnd.go.cd.v1+json http://localhost:8153/go/api/admin/materials/hg/notify -d '{"repository_url": "ssh://hg@bitbucket.org/akshaydewan86/gocd-ssh-hg"}'
```
```json
{
  "message" : "The material is now scheduled for an update. Please check relevant pipeline(s) for status."
}
```

## GET material modifications
```bash
curl -u admin:badger -H Accept:application/vnd.go.cd.v1+json http://localhost:8153/go/api/materials/458ce82419bdbe41a4007ad713a730e5e467ed4fbefb421bf58aeff3e9c36e4b/modifications
```
```json
{
  "_links" : {
    "self" : {
      "href" : "http://localhost:8153/go/api/materials/458ce82419bdbe41a4007ad713a730e5e467ed4fbefb421bf58aeff3e9c36e4b/modifications"
    },
    "doc" : {
      "href" : "https://api.gocd.org/19.5.0/#materials"
    },
    "find" : {
      "href" : "http://localhost:8153/go/api/materials/:fingerprint/modifications"
    }
  },
  "_embedded" : {
    "modifications" : [ {
      "id" : 2,
      "user_name" : "Akshay Dewan <akudewan@gmail.com>",
      "revision" : "92385c780e7d89db9a3aaeae573b1756fbe6c6b2",
      "modified_time" : 1404828986000,
      "comment" : "Adding visualizer and knock sensor"
    } ],
    "pagination" : {
      "page_size" : 10,
      "offset" : 0,
      "total" : 1
    }
  }
}
```

I observed that the Modifications endpoint returns a server error if the pipeline hasn't been run yet. This is because no `MaterialInstance` has been created. This will need to be handled better.

adityasood:
The PR was validated on 19.6.0-9340

PR #6332, opened 2019-05-24
Errors-to-model-fields binding from a server API response

marques-work:
## Problems this is trying to solve:

JS models tend to make API calls for CRUD (and other) operations. Validating on the client side is sometimes insufficient for detecting errors, and inevitably, the user will encounter errors that might only be validated server-side and which prevent a user-initiated action from succeeding.

1) In such cases, the errors embedded in the server response should bind back to the respective fields in the model so that the user can correct the errors and resubmit the API request until success.

2) There may also be errors on fields not accounted/implemented by a JS model which prevent the API submission from succeeding. As there are no such fields in the model, errors would otherwise not be visible to the user, thus giving the impression that the page is broken and, honestly, a horrible user experience.

This implementation of `consumeErrorsResponse()` addresses both of these concerns by a) **binding errors in the server response back to the models** and b) **returning a set of errors that could not be matched to a field in any model or sub-model in the hierarchy**. Implementors are responsible for presenting any unbound errors to the user.

## Implementation notes:

Complete reimplementation of `consumeErrorsResponse()` inverts the data direction:

  - Instead of traversing the models and looking for errors on known fields in the server response, traverse the server response and record errors on model fields while keeping track of errors on unknown fields. Tracking unknown field errors was not possible in the previous implementation.
  - Encapsulates this related behavior into a class/mixin (and interface) `ErrorsConsumer`.
  - The class/mixin `ValidatableMixin` now extends `ErrorsConsumer`, so all `ValidatableMixin` subclasses also satisfy the `ErrorsConsumer` contract and thus get `consumeErrorsResponse()` for free.
  - Moves the `errors()` member from `ValidatableMixin` to `ErrorsConsumer`.
  - Sometimes the model hierarchy on the client-side does not exactly match the hierarchy on the server-side (and thus the server response object). Subclasses can override the `errorContainerFor()` function on an `ErrorsConsumer` to assign the appropriate descendant to which to bind the field error.

## Notes for Reviewers

The majority of the changes are:

  - [`server/webapp/WEB-INF/rails/webpack/models/mixins/errors_consumer.ts`](https://github.com/marques-work/gocd/blob/bind-pipeline-config-server-errors-back-to-fields/server/webapp/WEB-INF/rails/webpack/models/mixins/errors_consumer.ts) ([diff](https://github.com/gocd/gocd/pull/6332/files#diff-c0e66717724df0a05041b7903b204ed4))
  - [`server/webapp/WEB-INF/rails/webpack/models/mixins/new_validatable_mixin.ts`](https://github.com/marques-work/gocd/blob/bind-pipeline-config-server-errors-back-to-fields/server/webapp/WEB-INF/rails/webpack/models/mixins/new_validatable_mixin.ts) ([diff](https://github.com/gocd/gocd/pull/6332/files#diff-615297577147e487e6910589d12bc259))

The remaining changes mostly support the above.


server/webapp/WEB-INF/rails/webpack/models/mixins/errors_consumer.ts
```diff
@@ -0,0 +1,196 @@
...
+ * This class can either be used as a base class or a mixin. Currently, this is the base class
+ * for `ValidatableMixin`.
+ *
+ * The concepts encapsulated here are error storage and consuming errors from a server API
+ * response. This is important for any model handling CRUD via the API as there may be errors
+ * that only the server can return because they cannot be validated on the client side. This
+ * allows one to bind server errors back to the respective JS model fields so that the user
+ * can be made aware of errors that prevent the success of user-initiated operations.
+ */
+export class ErrorsConsumer implements ErrorsConsumer {
```
akshaydewan: Is the `implements` needed? Is it some weird compiler trick?

marques-work: Hmm, that’s a refactoring mistake, though syntactically correct. I will rename the class so it is distinct from the interface. I know that typescript classes are also interfaces but I don’t consider it a good practice to have the concrete class be the same as the interface because it partly removes the benefit of using interfaces, IMO. I’ll fix this, then merge. Thanks for looking at it :)

marques-work:
rebased onto master to resolve conflicts and merging now.

PR #6351, opened 2019-05-28
Add template support for the new add pipeline page

ibnc:


ibnc:
Screenshot not using template:

Screen Shot 2019-05-28 at 15 54 22



Screenshot using template:

Screen Shot 2019-05-28 at 15 54 34

ibnc:
Screenshot when there are no templates defined:

Screen Shot 2019-05-28 at 15 58 02

ibnc:
@marques-work when you've got a chance, would you mind looking this over?

marques-work:
> @marques-work when you've got a chance, would you mind looking this over?

Sure @ibnc, looking at this now.

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
```diff
@@ -27,6 +27,7 @@ import {Stage} from "./stage";
 export class PipelineConfig extends ValidatableMixin {
   group: Stream<string> = stream("defaultGroup");
   name: Stream<string>;
+  template: Stream<any> = stream();
```
marques-work: Why not `Stream<string>`? I believe the `template` value is just that?

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
```diff
@@ -45,10 +46,24 @@ export class PipelineConfig extends ValidatableMixin {
     this.validateAssociated("materials");
 
     this.stages = stream(new NameableSet(stages));
-    this.validateNonEmptyCollection("stages", {message: `A pipeline must have at least one stage`});
     this.validateAssociated("stages");
   }
 
+  isUsingTemplate() {
```
marques-work: To me, `isUsingTemplate()` feels like a view concern rather than a `model` concern. My suggestion would be to add a `boolean` flag to either the `vnode.state` or `vnode.attrs` of the view component responsible for the toggle switch.

It's probably better to be explicit here, rather than rely on the value of the attribute on the pipeline model when implementing `isUsingTemplate()` anyway - kind of feels like a data-abstraction violation, but I could be convinced otherwise if you feel strongly about it.

While we're on the topic, I kind of see the concepts in this dichotomy (totally open for discussion, btw):

### View concerns (i.e., stuff to implement on the component):

1. `isUsingTemplate()` - the state of the toggle switch
2. `useTemplate()` - changes the model as a consequence of toggling
3. Changes the visible view components as a consequence of toggling (in an MVVM paradigm, the view model would mutate the underlying data model)

### Model concerns (i.e., stuff to implement on the `PipelineConfig` model):

1. *validation* that if `template` is set, `stages` should not be, and vice-versa

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
ibnc: @marques-work `template()` can have one of two types of values: `undefined` or `string`

Basically, if template's value is a string, then stages must be empty. If stages is not empty, then template must be undefined. The reason being is that the server returns a validation error if `template` is in the json with any amount of `stages`

Essentially, stages and templates are mutually exclusive

marques-work: `Stream<string>` can hold both `undefined` and `string` :).

marques-work: It's admittedly a minor quibble, but it doesn't hurt to take advantage of typescript's typechecking.

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
ibnc: @marques-work I don't know if we should have a validation for the mutual exclusivity between `template` and `stages` because it's not an error the user could correct nor should they ever see. If seems like something that should be confirmed in the tests which it currently is being tested.

marques-work: @ibnc You’re right in that the user shouldn’t be able to see it because of how the view is rendered.

Do consider that conceptually, the model should *still validate* this because a model **doesn’t (and shouldn't) know/care about the presentation**. Its concerns are managing the data in its structure, asserting validity on said data, and handling CRUD & serialization.

Even if we choose not to validate on the client-side, if somehow it happens that we get into an invalid state, the server would prevent the request from succeeding and include error keys somewhere in the response for that field. This error won't be at the top level, so it won't be visible on the page in its current state. This makes the page appear broken if it were to ever occur.

We should figure out how to present this error, even though we don't believe it would happen under normal circumstances. My guess is that the error should live on the template field. I bet the server error response puts the error at that level in the hierarchy there anyway.

marques-work:
@ibnc rebased and squashed. Added a small change to include a link to the templates page in the flash message when the set of templates is empty. Will merge after green tests for non-server builds.

PR #6355, opened 2019-05-30
Update autocomplete component

kritika-singh3:
Refactored autocomplete component
  - set the text color of suggestions to black
  - added `awesomplete` class to parent dom outside initialization - so that the same will persist even on update/error
  - ~not set the property value directly, utilizing a method redirection to do so~
  - removed the event listener for `awesomplete-selectcomplete` as it seemed redundant
  - added `autoEvaluate` attribute to show suggestions on `provider.update`
  - decouple all state from the component class in favor of `vnode.state`


marques-work:
@kritika-singh3 I made a modification before merging to decouple all state from the `AutocompleteField` component and instead manage state locally on `vnode.state`. Probably makes no practical difference to the end-user, but I felt it was cleaner or a more "pure" component.

PR #6359, opened 2019-05-31
Migrate /go/about server information page to spark

GaneshSPatil:
Screen Shot 2019-05-31 at 10 53 04 AM


GaneshSPatil:
@rajiesh @maheshp @ketan -- As this page is read-only information page, I am not planning to write a feature_toggle to go back to the rails page.

I will be sending another PR to get rid of the rails based page.

viraj2712:
LGTM

\cc: @GaneshSPatil

PR #6360, opened 2019-05-31
Display pipelines defined by a config-repo in the SPA

marques-work:
# What are we trying to achieve?

We believe that users want to see the set of pipelines (at least) in a config repo they created. Particularly after adding a config repo for the first time, user research indicated that users would want some kind of confirmation of the fruits of their efforts when creating pipelines via config repo.

## Dev Notes:

- [x] Add endpoint to show pipelines defined by a specific config-repo
- [x] Create components to display hierarchy
- [x] Removed stages
- [x] Added support for environments

### Here's what it looks like. Will need help from @streisguth with regards to colors and styles:

[~outdated screenshot~](https://user-images.githubusercontent.com/723642/59706541-83ed8400-91b5-11e9-9a0e-c482e162dd94.png)

image


marques-work:
Related to #6471.

marques-work:
Also questionable: Is there a use-case for displaying stages here? Or are pipeline names sufficient for all cases?

marques-work:
@streisguth any thoughts on a better presentation of the hierarchy?

GaneshSPatil:
@marques-work -- As a user, I would like to know all the GoCD entities defined by a config repo. Eg: Pipelines, Environments, etc.

Talking about `Is there a use-case for displaying stages here?`. 
A user doesn't get much information from showing stage names, eg: number and kind jobs performed by the stage, other pipeline configuration(s). 

As we are working on pipeline config rewrite and we would consider showing read-only pipeline configurations, does it make sense to show the pipeline name here and link to the read-only config of the same pipeline?

Also, IMO, we should only show the identifier of the entities that can be defined in config repo. eg: pipeline-names, env-names. Any nested/sub entities should be avoided.

bdpiparva:
@marques-work - I like the idea of showing entities coming from the config repo on success. As @GaneshSPatil mentioned showing all the level of information can make the list look huge even with a couple of pipelines.

What about listing the top level entity and have like to the page where user can get more information about it?

marques-work:
@bdpiparva @GaneshSPatil Thanks for your input. Looks like there is consensus around only showing top-level entities, so I will remove stages. I didn't bother printing environments as I was working in the context of introducing PaC to new users who may not get what an environment is, but I will add that.

I did know about the idea of a read-only config view of the pipelines and thought that would be the ideal page. As that work is not complete yet, I've linked them to the pipeline history page for now -- unless either of you has a better suggestion?

ibnc:
@marques-work @GaneshSPatil 

On the question of stages. What about making what's displayed collapsible at each level? The stages could be collapsed by default, and the groups could be expanded by default. 

Regardless of whether or not we make it collapsable at every level, there should be a way to hide some or all of this information. Displaying too much info without the option to hide it will make the page difficult to navigate. 

Also, do we want to have the group link to a dashboard page with only pipelines in that group showing?

ibnc:
We should probably display something when there is nothing defined. Even a message saying there's nothing.

Screenshot of what currently displays:

Screen Shot 2019-06-20 at 11 09 15

marques-work:
> Also, do we want to have the group link to a dashboard page with only pipelines in that group showing?

@ibnc that would be tricky as we cannot guarantee that the pipeline is viewable according to filters. That was a consideration though.

marques-work:
> We should probably display something when there is nothing defined. Even a message saying there's nothing.

You're right, I should add that. I do question as to whether or not it should be allowable to have nothing defined. That feels like a bug.

GaneshSPatil:
> As that work is not complete yet, I've linked them to the pipeline history page for now -- unless either of you has a better suggestion?

@marques-work -- works for now, as we don't have any other page with shows only pipeline details.

---

> On the question of stages. What about making what's displayed collapsible at each level? The stages could be collapsed by default, and the groups could be expanded by default.

Make sense if we are planning to show stages. The UI shouldn't look too cluttered.

maheshp:
Sorry, for the late comments.

Instead of listing the pipelines/environments on the Config Repo page did we consider adding a filter to show pipelines for a given a config repo on the dashboard. Once we have the filter, we can link to dashboard from the config repo page. Personalization can complicate some of this though.

arvindsv:
I believe @marques-work considered this. I remember him mentioning personalization as the main problem in doing this.

marques-work:
@arvindsv @maheshp I did think about that - and thought perhaps we could create a view to include all defined pipelines after parsing as an easy impl. I thought it might be complicated to keep in sync, and also didn't want to pollute the user's personalization space. But maybe I'm wrong or overthinking it. We should discover what users want here.

marques-work:
@GaneshSPatil @maheshp @bdpiparva @ibnc @arvindsv @kierarad @streisguth:

Should I split out environments and groups into their own section? Or is it OK for them to live under the same hierarchy, so long as they are ordered by type as is the case now? What makes the most sense to you?

image

arvindsv:
I think what you've shown is ok. It's an obvious improvement over what we (don't) have currently. :) Maybe @naveenbhaskar or @streisguth have some ideas. But, I'd say get whatever you have now in, and we can always continue changing it if needed or if it's confusing.

arvindsv:
> I did think about that - and thought perhaps we could create a view to include all defined pipelines after parsing as an easy impl. I thought it might be complicated to keep in sync, and also didn't want to pollute the user's personalization space.

@marques-work Yes. Polluting the user's personalization space is not something we should do.

_Maybe_ (just maybe) we can think of having a dynamic view such as: `/go/pipelines?configRepoView=repo-01` but I wouldn't change a user's personalization view. Of course, this means adding new support in the APIs, etc - which seems like a lot of unnecessary work. So, we should tread carefully and be sure it'll be useful. I'm not yet convinced it will be.

kierarad:
I think the way you've presented the hierarchy looks good. Based on the screenshot it's hard to tell, but are they all links? If so, what do the groups link to?

marques-work:
@kierarad yes, they are all links:

* Environments => environment show page
* Groups => group show page (note: not `/go/admin/pipelines`)
* Pipelines => pipeline history page (for now)

marques-work:
@arvindsv yeah, that seems like a good amount of work that I'd hesitate to put in without a clear demand for it from our users

Issue #6136, opened 2019-04-15
Support for Secert Params in Materials - Enhancement


maheshp:
Part of #5830

When support for `SecretParams` in materials was implemented as part of #6014, all material was not supporting credentials as attributes(#6298) and was not yet considered, hence secretParams were allowed to be defined in both URL as well as in the password field.

Now all materials will have support to define credentials seperately over having them as part of the URL. With this, support for secretParams in materials will be restricted only to the password field.

- [x] 1. Remove support for secretParams in URL for all materials.

- [x] 2. All material will support secret params in the password field.

- [x] 3. Material password is encrypted when saved to config_xml, if a secretParam is defined as password the same will be encrypted and saved.

- [x] 4. UX - SecretParams are supported only in the password field. Since text entered in a password field are masked, entering a secret param which gets masked might not be an optimal 
         UX. Consider having an option to hide/show passwords for materials. That way users can see the entered secretParam, test connection and save the config. The option to show/hide a password should be enabled only while creating the material and not when editing existing material. (#6371)

- [x] Should use the resolved secret param value for the password if specified as secret params
	- MDU
	- Job assignment
	- Check connection

- [x] Make sure that secret params value is masked -
	- Server health messages
	- Logging
	- In UI form, error should not display secret params value(need to think about this)
	- Check connection

- [x] Post commit webhooks and notify material api should continue to work as is when materials are configured with secret_params.

- [x] Secret param resolution
	a. Material urls should be resolved before MDU - failure of resolution should be logged to go-server.log as well as should show up in server health message.
	b. Material urls should be resolved before assigning a job to an agent. Failure of resolution should fail job assignment and an appropriate message should be logged in the job console.
	c. Material urls should be resolved before an MDU on the manual trigger of a pipeline. Failure should be logged to the server health message.
	d. Material urls should be resolved before a test connection.
	e. Material urls should be resolved before MDU on notification to material webhook endpoints.
	f. SVN material expansion should work as is if material URL has secrets.
	g. Post-commit hook API should work as is when the material is configured with secret params

### Validation

- [x] Ensure secret params in the password has a valid secret config id(#6050).
- [x] Make sure that material can refer to the specified secret config.

### Fingerprint

- Fingerprint of existing material should remain as it is.

rajiesh:
Verified on 19.7.0-9567

PR #6371, opened 2019-06-03
 Add toggle for password show or hide

kritika-singh3:
Issue: #6136 point 4
- needed as secret param is allowed in password field which is masked by default
  making it difficult to edit/create


kritika-singh3:
Preview:
For new pipeline materials:
new-pipeline-password-toggle

For edit pipeline materials:
edit-pipeline-password-toggle

Issue #6369, opened 2019-06-03
Placeholder issue for Secrets management


rajiesh:
##### Issue Type

- Bug Report

##### Summary

This is placeholder issue for secrets management feature(SPA, APIs, usage of secrets)

- [x] Moving/Cloning a pipeline using secret from pipeline group which allowed usage of secrets to group which denied usage of secret returns generic error. Error message should be more specific. More details [here](https://github.com/gocd/gocd/issues/6369#issuecomment-498132581)

- [x] Editing a secret config rule should validate if the modified config confirms to the rule. More details [here](https://github.com/gocd/gocd/issues/6369#issuecomment-498348913)

- [x] Secrets not accepted as material password. Check connection fails with error `encryptedPassword: Encrypted password value for <material_type> with url <material_url> is invalid. This usually happens when the cipher text is modified to have an invalid value`

rajiesh:
Create secret config as,

```
<secretConfig id="test" pluginId="cd.go.secrets.file-based-plugin">
      <configuration>
        <property>
          <key>SecretsFilePath</key>
          <value>secret.db</value>
        </property>
      </configuration>
      <rules>
        <allow action="refer" type="pipeline_group">Group1</allow>
        <deny action="refer" type="pipeline_group">Group2</deny>
      </rules>
      <description />
    </secretConfig>
```

1. Create a pipeline, say P1, in Group1 and use the secret configures as an environment variable
2. Now move/clone the pipeline P1 to Group2
3. Error message shown,

<img width="688" alt="Screen Shot 2019-06-03 at 12 12 24 PM" src="https://user-images.githubusercontent.com/6533924/58781209-eb71d400-85f8-11e9-9520-bfb69e3ad7f0.png">

User unable to make out reason for failure

rajiesh:
Say admin user first create secret config with rule defined as this,

```
 <rules>
        <allow action="refer" type="*">*</allow>
      </rules>
```

Then other users started using the secret config freely in their pipelines. Now if admin user updates the secret config rule to say,

```
      <rules>
        <allow action="refer" type="*">PROD_*</allow>
      </rules>
```

The update to secret config rules gets saved successfully, but it leaves the config xml invalid. Every config save will fail since there is a reference for this secret config in resources not matching `PROD_*`

PR #6379, opened 2019-06-04
Fixing check connection error when secret config is refered

kritika-singh3:
Issue: #6369
 - Passing in the pipeline group name while validating scm material config in material_test controller
 - In check_connection command, utilizing resolved password i.e. secret resolved.
    - did for `git`, `hg`
    - need to test for `p4`

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/config/materials/mercurial/HgMaterialTest.java
+++ b/common/src/test/java/com/thoughtworks/go/config/materials/mercurial/HgMaterialTest.java
@@ -17,6 +17,7 @@
 package com.thoughtworks.go.config.materials.mercurial;
 
 import com.thoughtworks.go.config.CaseInsensitiveString;
+import com.thoughtworks.go.config.SecretParam;
 import com.thoughtworks.go.config.materials.Filter;
 import com.thoughtworks.go.config.materials.PasswordAwareMaterial;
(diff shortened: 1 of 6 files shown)


arvindsv:
@kritika-singh3 @bdpiparva @rajiesh I wonder if this will be a backdoor for any group admin to get any credentials. What happens if:

1. I'm a group admin who does not have permissions to administer a protected pipeline group which has a secret config associated with it. I'll assume for now that we print out the `{{SECRET:[verysecure][id]}}` in job console somewhere.

2. I call the check connection API and pass in the name of the protected group if needed and set the material to: `https://endpoint.i.control.example.com` with password set to: `{{SECRET:[verysecure][id]}}`.

Even though I don't have access to create a pipeline in that group, will the secret resolution work and provide that information to me?

bdpiparva:
@arvindsv - Let's say a user `bob` is a group admin of pipeline group `BobIsAdminHere` and protected pipelines are defined in the pipeline group `Protected`. 

> `bob` can create a pipeline in the pipeline group `BobIsAdminHere` which can refer the secrets from secret config defined for the pipeline group `Protected`.

@maheshp is working on the runtime rule evaluation which will solve the issue for us. Just before resolving the secret param check if a current entity can refer to the secret param used in it.

PR #6384, opened 2019-06-04
Show errors on clone/move of pipeline

kritika-singh3:
Issue: #6369
If the pipeline contains a secret config, which is accessible in one group but not in another,
errors need to be shown as to why:

 - [x] clone
 - [x] move


dhanasp:
Secret config related error 
 - while cloning :
Screenshot 2019-06-04 at 5 57 22 PM

 - while moving :
Screenshot 2019-06-04 at 5 51 08 PM

arvindsv:
@dhanasp Can you please change that to "... is not allowed to be used in ..."?

kritika-singh3:
Closing this as the error was picked very specific to the environment variable errors. The same is not present now as the validation is now at runtime. Will address this issue later on.

Issue #6350, opened 2019-05-28
Raw console logs of a building job are not available


arvindsv:
https://github.com/gocd/gocd/issues/6325#issuecomment-495350446:
> If we're looking at this, have you noticed that you can't get the raw console logs of a job which is in progress?

https://github.com/gocd/gocd/issues/6325#issuecomment-495595603:
> I'd expect that it shows the same log that's being shown in the scrolling logs when that is clicked. Even though it's being updated. We will need to check that it's building and do that, right?

@arvindsv For the raw console logs, we point to the final console.log file, not the temporary file. As a result, we can't see the raw logs until the build is finished. Not sure if this is existing behaviour or if it was changed accidentally. Nonetheless, it seems unrelated to this issue. We can fix it separately.

_Originally posted by @akshaydewan in https://github.com/gocd/gocd/issues/6325#issuecomment-496398992_

rajiesh:
Verified on 19.6.0 (9478-5f827efbc8001e16417fe13503f5e240d7ed060d)

PR #6392, opened 2019-06-06
Read the temp console log file when build is in progress

akshaydewan:
Issue #6350. The Raw Console logs will point to the temporary console log file when the build is in progress.
If the temp file hasn't been created yet (e.g. before the build is assigned to an agent), the user will still see a file not found message.


akshaydewan:
Need to fix the failing test on Windows - Don't merge it yet.

PR #6394, opened 2019-06-06
Fix check connection during new pipeline creation

viraj2712:
Issue: #6369
The password for material can be configured using a secret config. Validation for the same requires a group name. Since the pipeline hasn't been created yet, the config won't have the reference to the same.

Modified the material test API to accept a `pipeline_group` field.
On the UI end, fetched the group name from query params.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
@@ -109,7 +109,7 @@ public class RulesService {
         validateSecretConfigReferences(secretParams, group.getClass(), group.getGroup(), errorMessagePrefix);
     }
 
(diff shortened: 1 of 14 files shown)


arvindsv:
I asked a question here: https://github.com/gocd/gocd/pull/6379#issuecomment-498720377

Seeing how this is related, can someone take a look? Should we verify whether the current user is an admin of that group, at least?

kritika-singh3:
> Should we verify whether the current user is an admin of that group, at least?

Verified that the user is an admin or group admin: https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/controllers/api_v1/admin/internal/material_test_controller.rb#L25

kritika-singh3:
Updated the PR to check the material provided with the group name selected. Refactored `RulesService`. 

cc: @bdpiparva @maheshp

Issue #6287, opened 2019-05-14
ability to specify pause behaviour for a pipeline created through the pipeline config API


aqneves:
##### Issue Type

- Feature enhancement

##### Summary

I would be nice to be able to specify in the pipeline config api if a newly created pipeline should be paused by default or not. Currently, when we create a pipeline through the User Interface, due to the fact we can't specify for example multiple materials in the wizard, it's created paused by default, but when using the API (same goes for pipeline as code through config repository) it's created unpaused by default and it runs as soon as it's created.

There might be scenarios where a customer on-boarding is being prepared and although we are preparing the pipelines for this customer, we want to control from when it can actually be executed.

##### Environment

https://api.gocd.org/19.2.0/#create-a-pipeline

###### Basic environment details

GoCD 19.2

##### Steps to Reproduce

1. prepare a new pipeline creation file for a pipeline with a name that was never used before 
2. call the GoCD API to create that pipeline
3. verify that it runs automatically upon creation

arvindsv:
This was brought up in https://github.com/tomzo/gocd-yaml-config-plugin/issues/64 too.

msaurabh1978:
###  Design Considerations

The preference `Pause or do-not-pause` at the time of pipeline creation through API is not a pipeline configuration detail - it is more of metadata about the pipeline, hence this preference is not going to be added to the create pipeline config request payload (body). 

Other ways we have considered to support this preference is using

- Path Parameter
- Query Parameter or
- Request Header

***Path Parameters*** primarily contributes to uniquely identifying a resource (or a sub-resource) on the server. Something like,

/issues/**{issueId}**/comments/**{commentId}**

Since the preference `Pause or do-not-pause` does not constitute unique resource information, we are not considering the path parameter to support this preference. 

***Query Parameters*** are used mainly for filtering (narrowing down) the resource related details. Things like *searching based on filter criteria*, *pagination*, etc. Since the preference `Pause or do-not-pause` doesn't fit into this description, we are not considering *Query Parameters* for this.

***Standard Request Headers*** are used for exchanging crossing cutting concerns that are applicable across resources, things like *Authorisation*, *Content-negotiation*, and ***custom request headers*** are used for specifying *meta-data* about the resource (e.g state information). 

[...]

PR #6395, opened 2019-06-06
Fixed 6287 - Added support for pause/no-pause in pipeline config API

msaurabh1978:
Fixed Issue: #6287 (please refer to this issue for any design related details)

Implementation details are,

- [x] Updated create pipeline config API to honor request header `X-pause-pipeline-after-create` to pause or un-pause the pipeline after creation

- [x] Added unit tests for pipeline config controller v7


api/api-pipeline-config-v7/src/main/java/com/thoughtworks/go/apiv7/admin/pipelineconfig/PipelineConfigControllerV7.java
```diff
@@ -131,6 +139,10 @@ public String create(Request req, Response res) {
         Username userName = SessionUtils.currentUsername();
         pipelineConfigService.createPipelineConfig(userName, pipelineConfigFromRequest, result, group);
 
+        if(shouldPausePipelineOnCreate(req)){
+            pipelinePauseService.pause(pipelineConfigFromRequest.name().toString(), null, userName);
```
maheshp: Should we add a pause message for this?

ketan: And default to a sensible message, if one is not specified.

ketan:
Question: should this be part of a new API version? Specifying `X-pause-pipeline-after-create` on api v7 but with an older version of the gocd server will have no effect.

api/api-pipeline-config-v7/src/main/java/com/thoughtworks/go/apiv7/admin/pipelineconfig/PipelineConfigControllerV7.java
msaurabh1978: @ketan - Are u talking about making changes in <code>PipelinePauseService.pause()</code> method? wherein if caller of that method, does not specify the pause cause, then the sensible default message is used instead. That way pause message is never null in the database after this change. Is that right?

msaurabh1978:
> Question: should this be part of a new API version? Specifying `X-pause-pipeline-after-create` on api v7 but with an older version of the gocd server will have no effect.

Thinking aloud - in what case(s) then would we ever make changes to existing API version? because based on what you said (older version of GoCD server will have no effect) would be valid for any kind of change?

maheshp:
> Thinking aloud - in what case(s) then would we ever make changes to existing API version? because based on what you said (older version of GoCD server will have no effect) would be valid for any kind of change?

I don't remember we changing an existing API. A change to the API(contract) has always lead to a new version of the API.

arvindsv:
I'd expect us to change existing APIs only when fixing bugs in the API. If there's a new feature added or old feature removed, we should bump it up.

msaurabh1978:
Can u pls merge this pull request? I have incorporated all your review comments.

msaurabh1978:
Below is how the pause message shows up for a paused pipeline through pipeline config API.

GaneshSPatil:
@msaurabh1978 -- When we create a pipeline from UI, it shows pause message as `Under Construction`.

Can we use the same message:
1. As it is short and descriptive.
2. Consistency with the message that gets shown when we create a pipeline from UI.

msaurabh1978:
> @msaurabh1978 -- When we create a pipeline from UI, it shows pause message as `Under Construction`.
> 
> Can we use the same message:
> 
> 1. As it is short and descriptive.
> 2. Consistency with the message that gets shown when we create a pipeline from UI.

Yes, we can.

The way I am seeing it is `pause-cause` should reflect the actual reason for pausing the pipeline at any point in time (whether paused on creating or paused at a later time). `Under Construction` totally makes sense in case of pipeline getting created from UI because, at that point, it is really under construction. In this particular case, it is paused to have more control over when it can be started after reviewing it.

Another point is, whenever pipeline is paused at a later time after its creation, then the user can specify any message while pausing (could be a longer message as well).

So I think we need not be consistent with pause message on creating (as the context for pausing is different). Does this make sense?

maheshp:
There could be various reasons for pausing a pipeline post creation( to start later, to review and edit further, etc...) I think we should have a generic message that its paused post creation, how bout saying `This pipeline has been paused after creating it through API.`

ketan:
> how bout saying `This pipeline has been paused after creating it through API.`

Can I suggest passing in another header with a pause cause, and if that header is blank, then fallback to a sensible default?

ketan:
Discussed with @msaurabh1978 about the following options for default pause message, if one is not specified:

- @GaneshSPatil's suggestion "Under construction" does not clearly spell out why
- @maheshp's suggestion "This pipeline has been paused after creating it through API" makes it look like that's the default behavior, and gocd will pause all pipelines going forward
- @msaurabh1978's suggestion "intentionally paused" also makes it look like it's a decision that gocd made on behalf of the user.

We decided to instead use `No pause cause was specified when pipeline was created via API`. This seems to clarify exactly what has happened, without making it look like the GoCD server is making any decisions.

MRajasekhar:
Hi All, 

we are in situation to create pipelines with pause state from config repository code. the way we create pipelines from yaml file and config repo plugin reads -> creates/modify/deletes pipeline. 
When we are creating new pipelines - it is becoming active and running. we need to have it pause state so that we can release when it is ready to avoid unnecessary failures. 

Regards,
Rajasekhar

PR #6396, opened 2019-06-06
Use tanuki wrapper for all installers

ketan:


adityasood:
verified: 19.6.0-9439

PR #6405, opened 2019-06-10
Apply rules at runtime

bdpiparva:
- Validate if, environment variables defined in environment config can refer to a secret config
- Validate if environment variables defined in the pipeline config can refer to  a secret config
- Validate if the material in the pipeline config can refer to a secret config
- Fail build if any of the above-mentioned configs cannot refer to secret config and update the job console log

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
@@ -21,13 +21,13 @@ import com.thoughtworks.go.config.exceptions.RecordNotFoundException;
 import com.thoughtworks.go.domain.*;
 import com.thoughtworks.go.domain.builder.Builder;
 import com.thoughtworks.go.domain.exception.IllegalArtifactLocationException;
-import com.thoughtworks.go.domain.materials.Material;
 import com.thoughtworks.go.listener.ConfigChangedListener;
 import com.thoughtworks.go.listener.EntityConfigChangedListener;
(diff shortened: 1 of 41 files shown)


server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -357,46 +356,52 @@ private Work createWork(final AgentInstance agent, final JobPlan job) {
...
-        final SecretParams allSecretParams = SecretParams.union(
-                secretParamsInMaterials(buildAssignment),
-                secretParamsInEnvironmentVariables(buildAssignment)
-        );
+        secretParamResolver.resolve(environmentConfig);
 
-        secretParamResolver.resolve(allSecretParams);
+        EnvironmentVariableContext environmentVariableContext = new EnvironmentVariableContext();
+        environmentConfig.getVariables().forEach(variable -> {
+            String resolvedValue = variable.getSecretParams().substitute(variable.getValue());
```
maheshp: Consider EnvironmentVariableConfig return a resolved value rather than the substituting here?

bdpiparva: Thought about it while we were implementing. The `getValue()` is used at multiple places in a different context and not all of the requires the resolved value. I can add one more method `getResolvedValue()` which returns the resolved value, and error out if has an unresolved secret param.

maheshp: Ok, or `valueForCommandline`.  guess we are using `...ForCommandline` in other places.

PR #6413, opened 2019-06-11
Fixing styling for validation errors on add pipeline page

ibnc:
Need to wrap the text without throwing off everything else

before:
Screenshot_2019-06-04_at_11 46 26

after:

Screen Shot 2019-06-11 at 14 56 18

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/components.scss
+++ b/server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/components.scss
@@ -36,6 +36,7 @@ $error-container-width: 450px;
 .actions {
   .save-btns {
     white-space: nowrap;
+    height: 35px;
   }
 
   .error-response {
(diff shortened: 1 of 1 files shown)


ibnc:
@marques-work @kierarad would one of you mind reviewing this, and merging?

marques-work:
@ibnc I'll look at this now

marques-work:
@ibnc sorry, I did look at this and it looks ok on the surface, but there are some edge cases I wanted to validate. I'll do that today after I clean up my workspace and commit -- then I can build and play around with it.

marques-work:
@ibnc I'm going to remove the `height` property and see if this fixes the edge case where we get a bunch of server validation errors that we couldn't map. This ought to allow the actions section to expand to accommodate.

image

marques-work:
@ibnc ok removing `height: 35px` definitely fixes this.

image

PR #6552, opened 2019-06-19
Agents need to connect over secure port

kritika-singh3:


development-utility/development-agent/src/main/java/com/thoughtworks/go/agent/DevelopmentAgent.java
```diff
@@ -32,7 +32,7 @@ public static void main(String[] args) throws Exception {
         new ProcessRunner().command("curl", "http://localhost:8153/go/admin/tfs-impl.jar", "--fail", "--silent", "--output", "tfs-impl.jar").failOnError(false).run();
         new SystemEnvironment().set(SystemEnvironment.PLUGIN_ACTIVATOR_JAR_PATH, "go-plugin-activator.jar");
         assertActivationJarPresent();
-        AgentMain.main("-serverUrl", "http://localhost:8153/go");
+        AgentMain.main("-serverUrl", "https://localhost:8154/go");
```
bdpiparva: This is done in support of `GO_USE_TOKEN_AUTH`. I would suggest extracting https://github.com/gocd/gocd/blob/0effe2ebca923be4d36e48d3218ff576a3bbc6ac/agent/src/main/java/com/thoughtworks/go/agent/GoHttpClientHttpInvokerRequestExecutor.java#L84 as a system variable.

Then make the development agent to use token base auth
```
new SystemEnvironment().set(SystemEnvironment.GO_USE_TOKEN_AUTH, true);
```

bdpiparva: @kritika-singh3 - Ignore what I said before as it is going to be breaking change. As of now, we can use `https`

Issue #5830, opened 2019-02-12
Secrets Management [EPIC]


maheshp:
##### Issue Type

- Feature proposal

##### Summary

Currently secrets in GoCD are managed by encrypting them as secure environment variables or as secure configuration properties. While the secrets are encrypted there is scope for managing them better. Some of current pain points with managing secrets are,

1. Secrets are all over the place. The same secret can be defined across multiple environments, pipelines etc. Not centralized.
2. Secrets visibility. There is no good place to view all the secrets defined and where they are used.
3. Any time a secret is rotated, things would break in GoCD till the secrets are not updated manually in GoCD.
4. Access control to secrets is blurred and manual.

**Proposed solution**
1. Use an external secrets manager e.g Vault, AWS Secrets Manager etc to define secrets and manage in a centralized location. The external secrets manager would provide the following,
    - Centralized location for all secrets
    - Ability to rotate secrets
    - ACL
    - Auditing
2. GoCD to provide new Secrets Management plugin endpoints. Using these endpoints plugin authors can support an external secret manager of their choice. #5831
  a. Enhance Secrets Management Plugin Endpoint https://github.com/gocd/gocd/issues/6018
3. GoCD provides an ability to define and configure secrets_config. The secrets configuration would be specific to a plugin to would be used to lookup for secrets. #5832
4. How to use external secrets in GoCD? https://github.com/gocd/gocd/issues/5834
GoCD to introduce a concept of secret_params. The secret_params are resolved at runtime. e.g If a param is defined in a job, just before assigning the job to an agent, the server makes a request to the plugin to lookup for the secrets and the appropriate value is passed to the agent.
~~- #{SECRET[secret_identity_id][lookup_string]} (this can be changed appropriately)~~
Agreed secret param format `{{SECRET:[secret_config_id][lookup_string]}}`
5. Support secret params in Materials - #6014 
6. Support secret params in Materials enhancements - #6136
7. Support for Rules in SecretConfig #6095
8. API which supports CRUD operations for secrets_config. https://github.com/gocd/gocd/issues/5835
9. SPA which supports CRUD operations for secrets_config. https://github.com/gocd/gocd/issues/5837
10. Secret Params configuration and UX https://github.com/gocd/gocd/issues/5836
11. File based secrets management plugin https://github.com/gocd/gocd/issues/5838
12. Implement support to return plugin info for secret extension plugins https://github.com/gocd/gocd/issues/6079
13. Documentation

rajiesh:
[...]

PR #6555, opened 2019-06-20
Updated rules validation on MDU

kritika-singh3:
Epic: #5830 

 - MDU will go through if there exists at least one material which can refer to the secret config
    in question
 - MDU will throw `RulesViolationException` if all the pipeline which utilizes this material violates the 
   rules defined.

### Scenarios

 1. A material URL utilized in 2 different pipeline groups referring to the same secret config.
     - if the secret config can be accessed by both, the MDU will go through.
     - if the secret config can be accessed by only one of them, the MDU will go through.
     - if the secret config can't be accessed by either of them, the MDU will fail.

 2. A material URL utilized in 2 different pipeline groups referring to different secret config. In this case, the MDU going through will be dependent on which material went through the process. 
     - if the secret param present in the material can access either of the pipeline group, it will go through, otherwise not.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
@@ -20,12 +20,16 @@ import com.thoughtworks.go.config.*;
 import com.thoughtworks.go.config.materials.ScmMaterial;
 import com.thoughtworks.go.domain.JobIdentifier;
 import com.thoughtworks.go.remote.work.BuildAssignment;
+import com.thoughtworks.go.server.exceptions.RulesViolationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static com.thoughtworks.go.server.exceptions.RulesViolationException.throwCannotRefer;
 import static com.thoughtworks.go.server.exceptions.RulesViolationException.throwSecretConfigNotFound;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
```diff
@@ -44,13 +48,42 @@ public RulesService(GoConfigService goConfigService) {
...
+                            LOGGER.error("[Material Update] Failure", format("Pipeline '%s' does not have permission to refer to secrets using SecretConfig: '%s'", pipelineName.toString(), secretConfig.getId()));
+                            pipelinesWithViolatedSecretConfigIds.get(pipelineName).add(secretConfig.getId());
+
+                        }
+                    });
+                });
+
+        if (pipelinesWithViolatedSecretConfigIds.size() == pipelines.size()) {
+            String secretConfigId = secretParams.get(0).getSecretConfigId();
+            String errorMessage = format("Material with url: '%s' does not have permissions to refer to secret config with id: '%s'. The pipelines which use this material are: '%s'", scmMaterial.getUriForDisplay(), secretConfigId, pipelinesWithViolatedSecretConfigIds.keySet().toString());
```
bdpiparva: Since the material password is not part of the fingerprint. Different pipeline with the same material can have a password configured using a different secret config ID. In that case, the message will always print the first secret config ID and might be misleading.

What about the following the message?
```
Material with URL: 'some-url' does not have permissions to refer to secret config 'aws' in pipelines "prod", "pre-prod" and secret config 'file' in pipelines "test", "build"
```
/cc @maheshp

rajiesh:
@bdpiparva @kritika-singh3 the material is not the one which do not have permission, but the pipeline group. So the message was bit misleading when we say `Material with URL: 'some-url' does not have permissions to refer`, better would be `All Pipelines using material URL : <url> do not have permissions to refer secret config <config_id>. Material credential is using secret param from this config. Pipelines referring to the secret config are 'P1', 'P2'`

bdpiparva:
@rajiesh -  Agreed,

```
Pipelines "P1" and "P2" are using the material with URL: <url>, and do not have permissions to refer secret config <config_id>.
```

FYI: A material password can have multiple secret param combination in it. e.g. "{{SECRET:[foo][firstpart]}}_{{SECRET:[bar][secondpart]}}". but this is a very unusual scenario.

kritika-singh3:
This PR was incomplete, hence was reverted. Refer to #6564 for improved implementation and logging.

PR #6557, opened 2019-06-20
If server URL is http, then use client certs, irrespective of mutual tls flag

ketan:


arvindsv:
One consequence though, is if someone makes a mistake in the URL when configuring it, they send their token out in the open. There’s no indication anywhere that the connection happened / is happening over http. Ideally, reverse proxies are configured properly to reject these. If not, they’ll transparently “upgrade” to `https` and send a request to the server using `https`, but the connection from agent to proxy will be over http, with a token.

I think this PR still makes sense, since, without this, if an http GoCD server URL is provided, there will be no authentication information passed in at all.

At some point in the (near) future, we will probably throw away mutual SSL anyway, and be left with something like what's happening in this PR.

PR #6564, opened 2019-06-21
Updated rules validation on MDU

kritika-singh3:
Original PR: #6555  
 - MDU will go through if there exists at least one material which can refer to the secret config specified
 - MDU will throw `RulesViolationException` if all the pipeline which utilizes the material with the same fingerprint violates the secret config rules.

#### Preview

Screen Shot 2019-06-21 at 3 02 24 PM

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
@@ -46,57 +46,6 @@ public class RulesService {
         this.goConfigService = goConfigService;
     }
 
-    public boolean validateSecretConfigReferences1(ScmMaterial scmMaterial) {
-        List<CaseInsensitiveString> pipelines = goConfigService.pipelinesWithMaterial(scmMaterial.getFingerprint());
-        SecretParams secretParams = scmMaterial.getSecretParams();
-
-        List<String> missingSecretConfigs = secretParams.stream()
-                .filter(secretParam -> goConfigService.cruiseConfig().getSecretConfigs().find(secretParam.getSecretConfigId()) == null)
-                .map(SecretParam::getSecretConfigId)
-                .collect(Collectors.toList());
-
-        if (!missingSecretConfigs.isEmpty()) {
-            throwSecretConfigNotFound("ScmMaterial", scmMaterial.getUriForDisplay(), String.join(", ", missingSecretConfigs));
-        }
-
-        HashMap<CaseInsensitiveString, List<String>> pipelinesWithViolatedSecretConfigIds = new HashMap<>();
-        secretParams.stream()
-                .map(secretParam -> goConfigService.cruiseConfig().getSecretConfigs().find(secretParam.getSecretConfigId()))
-                .forEach(secretConfig -> {
-                    pipelines.forEach(pipelineName -> {
-                        MaterialConfig materialConfig = goConfigService
-                                .findPipelineByName(pipelineName)
-                                .materialConfigs()
-                                .getByMaterialFingerPrint(scmMaterial.getFingerprint());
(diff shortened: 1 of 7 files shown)


kritika-singh3:
Log snippet:

Screen Shot 2019-06-21 at 3 36 05 PM

server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
```diff
@@ -41,18 +46,103 @@ public RulesService(GoConfigService goConfigService) {
         this.goConfigService = goConfigService;
     }
 
-    public boolean validateSecretConfigReferences(ScmMaterial scmMaterial) {
+    public boolean validateSecretConfigReferences1(ScmMaterial scmMaterial) {
```
bdpiparva: Should this be removed?

server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
```diff
@@ -41,18 +46,103 @@ public RulesService(GoConfigService goConfigService) {
...
+    public boolean validateSecretConfigReferences(ScmMaterial scmMaterial) {
+        List<CaseInsensitiveString> pipelines = goConfigService.pipelinesWithMaterial(scmMaterial.getFingerprint());
+
+        HashMap<CaseInsensitiveString, StringBuilder> pipelinesWithErrors = new HashMap<>();
+        pipelines.forEach(pipelineName -> {
+            MaterialConfig materialConfig = goConfigService
+                    .findPipelineByName(pipelineName)
+                    .materialConfigs()
+                    .getByMaterialFingerPrint(scmMaterial.getFingerprint());
+            if (materialConfig instanceof ScmMaterialConfig) {
```
bdpiparva: I don't think this is required as the method is only taking ScmMaterial as an input

server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
```diff
@@ -41,18 +46,103 @@ public RulesService(GoConfigService goConfigService) {
...
+                PipelineConfigs group = goConfigService.findGroupByPipeline(pipelineName);
+                ScmMaterialConfig scmMaterialConfig = (ScmMaterialConfig) materialConfig;
+                SecretParams secretParams = SecretParams.parse(scmMaterialConfig.getPassword());
+                secretParams.forEach(secretParam -> {
+                    String secretConfigId = secretParam.getSecretConfigId();
+                    SecretConfig secretConfig = goConfigService.getSecretConfigById(secretConfigId);
+                    if (secretConfig == null) {
+                        addError(pipelinesWithErrors, pipelineName, format("\nPipeline '%s' is referring to none-existent secret config '%s'.", pipelineName, secretConfigId));
```
bdpiparva: Do we need  `\n` here?

server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
```diff
@@ -41,18 +46,103 @@ public RulesService(GoConfigService goConfigService) {
...
+                    } else if (!secretConfig.canRefer(group.getClass(), group.getGroup())) {
+                        addError(pipelinesWithErrors, pipelineName, format("\nPipeline '%s' does not have permission to refer to secrets using secret config '%s'", pipelineName, secretConfigId));
```
bdpiparva: same as above

server/src/main/java/com/thoughtworks/go/server/service/RulesService.java
kritika-singh3: No, we don't. We are using `StringUtils.join(list, '\n')`. I will update it.

PR #6565, opened 2019-06-21
Show spinner in modal when performing save/update

bdpiparva:
#### Before
before

#### After
after


GaneshSPatil:
@bdpiparva @kritika-singh3 -- When the spinner appears in the modal, the modal shrinks and expands again, based on the height of the modal.

Can we show the modal on the content itself with a blurred background?
\cc: @naveenbhaskar

naveenbhaskar:
@GaneshSPatil  agree with that.. The height of the modal should not change, just add a translucent white layer at the top of the content and add the spinner .

PR #6569, opened 2019-06-24
Show spinner on modal body with overlay on the content.

bdpiparva:
## Show spinner on the content with an overlay
after-1

Implements: https://github.com/gocd/gocd/pull/6565#issuecomment-504852948


akshaydewan:
Another pattern I've seen in some screens is to disable the Save button and show a spinner next to it on the button itself. IIRC, the Dashboard tabs modal does this.

bdpiparva:
@akshaydewan - IMO, Spinner next to a button makes sense when it is an inline edit, or check connection kind of stuff. @naveenbhaskar will have a better idea about it.

Issue #4189, opened 2018-01-16
Server configuration API endpoint


anishvenkat:
##### Issue Type

- Feature proposal

##### Summary

Currently, a user has to manually login to the dashboard to update the `security` configuration attributes or edit it from the cruise-config.xml file locally. Creating a placeholder for having an API endpoint to update these attributes via an API call.

inge4pres:
Hi :) the API is already there. 
See https://api.gocd.org/current/#authorization-configuration 

The way I'm using it for example is to configure password-file auth after the server boots in a container, mounting the file as a volume and running

```
curl 'http://localhost:8153/go/api/admin/security/auth_configs' \
      -H 'Accept: application/vnd.go.cd.v1+json' \
      -H 'Content-Type: application/json' \
      -X POST -d '{
    "id": "password-file",
    "plugin_id": "cd.go.authentication.passwordfile",
    "properties": [
      {
        "key": "PasswordFilePath",
        "value": "/godata/credentials.txt"
      }]
  }'
```

GaneshSPatil:
Hello @inge4pres, the authorization-configuration API allows admin users to manage authorization configuration. It is used to manage authConfigs under `<server>` tag.

This is the placeholder issue to create an API which will allow admin users to manage attributes of the `<server>` tag. For Example: `artifactsdir`, `agentAutoRegisterKey`, `webhookSecret`, etc

akshaydewan:
Wondering what would be the right endpoint for this. `/api/config/server` would imply that everything under `server` will be configurable from this API.

Maybe we can group them into entities. From what I see, we have this structure:
- Server
  - Urls
    - siteUrl
    - secureSiteUrl
  - Artifacts
    - Purge Settings
      - purgeStart
      - purgeUpto
    - ArtifactsDir
  - DefaultJobTimeout
  - CommandRepository
    - location
    - reload cache (action)

And then have enpoints for:
`/api/config/server/urls`
`/api/config/server/artifacts`
`/api/config/server/defaultJobTimeout`
`/api/config/server/commandRepository`

Although everything just goes in as attributes for the `<server>` tag in config-xml. And the current UI just has one Save button. So this approach may be more *purist* than *practical*

Thoughts?

ketan:
> So this approach may be more _purist_ than _practical_

Should we re-configure how the config XML looks in a manner that makes sense and then see how the API fits on top of that?

akshaydewan:
[...]

PR #6570, opened 2019-06-24
Introduce Mail host configuration API v1

GaneshSPatil:
The endpoint is:

```
/api/config/mailserver
```

The JSON response looks like:

```json
{
  "_links" : {
    "self" : {
      "href" : "http://test.host/go/api/config/mailserver"
    },
    "doc" : {
      "href" : "https://api.gocd.org/19.8.0/#mailserver"
    }
  },
  "hostname" : "ghost.name",
  "port" : 25,
  "username" : "loser",
  "encrypted_password" : "AES:lzcCuNSe4vUx+CsWgN11Uw==:Q2OlnqIf9S++yMPqSCx8qg==",
  "tls" : false,
  "sender_email" : "go@foo.mail.com",
  "admin_email" : "admin@foo.mail.com"
}
```


api/api-mailhost-v1/src/main/java/com/thoughtworks/go/apiv1/mailhost/representers/MailhostRepresenter.java
```diff
@@ -0,0 +1,50 @@
...
+                .add("username", mailhost.getUserName())
+                .add("encrypted_password", mailhost.getEncryptedPassword())
+                .add("tls", mailhost.getTls())
+                .add("from", mailhost.getFrom())
+                .add("admin", mailhost.getAdminMail());
+    }
+
+    public static MailHost fromJSON(JsonReader jsonReader) {
+        String hostname = jsonReader.getString("hostname");
+        int port = jsonReader.getInt("port");
```
ketan: Should be optional int, and throw a validation error from the modal. Not specifying a port will otherwise cause a 422 server error.

ketan: s/modal/model/

api/api-mailhost-v1/src/main/java/com/thoughtworks/go/apiv1/mailhost/representers/MailhostRepresenter.java
```diff
@@ -0,0 +1,50 @@
...
+        String username = jsonReader.optString("username").orElse(null);
+        String password = jsonReader.optString("password").orElse(null);
+        String encryptedPassword = jsonReader.optString("encrypted_password").orElse("");
```
ketan: See `PasswordDeserializer#deserialize`. Might be able to change the signature from:

`public String deserialize(String password, String encryptedPassword, AbstractMaterialConfig materialConfig)` to:

`public String deserialize(String password, String encryptedPassword, Validatable entity)`

server/src/main/java/com/thoughtworks/go/server/service/ServerConfigService.java
```diff
@@ -160,4 +188,19 @@ public String getWebhookSecret() {
...
+            if (e instanceof GoConfigInvalidException) {
+                if (!result.hasMessage()) {
+                    result.unprocessableEntity(entityConfigValidationFailed(mailHost.getClass().getAnnotation(ConfigTag.class).value(), e.getMessage()));
+                }
+            } else if (!(e instanceof ConfigUpdateCheckFailedException)) {
+                LOGGER.error(e.getMessage(), e);
+                result.internalServerError(saveFailedWithReason(e.getMessage()));
+            }
+        }
+    }
```
akshaydewan: How about:
```java
        catch (GoConfigInvalidException e) {
            if (!result.hasMessage()) {
                result.unprocessableEntity(entityConfigValidationFailed(mailHost.getClass().getAnnotation(ConfigTag.class).value(), e.getMessage()));
            }
        } catch (ConfigUpdateCheckFailedException e) {
            //nothing
        } catch (Exception e) {
            LOGGER.error(e.getMessage(), e);
            result.internalServerError(saveFailedWithReason(e.getMessage()));
        }
```

spark/spark-base/src/main/java/com/thoughtworks/go/spark/Routes.java
```diff
@@ -138,6 +138,11 @@ public static String vsm(String materialFingerprint, String revision) {
         }
     }
 
+    public static class MailHost {
+        public static final String SELF = "/api/admin/mailhost";
+        public static final String DOC = "https://api.go.cd/current/#mailhost";
```
akshaydewan: Do we want to the endpoint to be `/api/admin/mailhost` or `/api/admin/config/server/mailhost` or `/api/admin/server/mailhost`?

Whatever we choose should be consistent with https://github.com/gocd/gocd/issues/4189# IMO

ketan: How about `/api/admin/config/{smtp,emailserver,mailserver}`. I don't know why it was called `mailhost`, but my preferene is smtp, followed by emailserver and mailserver.

api/api-mailhost-v1/src/main/java/com/thoughtworks/go/apiv1/mailhost/representers/MailhostRepresenter.java
GaneshSPatil: @ketan -- Hmm. isn't that applicable for all the APIs?
We perform `get<TYPE>` on mandatory fields, to fail the API when the required fields are not specified.

Server returning 422 in case of not specifying required fields was intentional.

spark/spark-base/src/main/java/com/thoughtworks/go/spark/Routes.java
GaneshSPatil: @ketan @akshaydewan -- If we decide to call this API endpoint `smtp` or `emailserver`, does it make sense to change the xml entity from `mailhost` to whatever we decide?

ketan:
Anything I can help with to get this PR moving?

GaneshSPatil:
> Anything I can help with to get this PR moving?

@ketan sure.

Pending things on the PR:
- [ ] Comment Fixes
- [ ] API Documentation

spark/spark-base/src/main/java/com/thoughtworks/go/spark/Routes.java
ketan: Does anyone have opinions on this question?

GaneshSPatil:
@ketan -- we also need to rename the config tag from `mailhost` to `smtp`. 
But that can be done as part of a separate config-migration PR.

spark/spark-base/src/main/java/com/thoughtworks/go/spark/Routes.java
arvindsv: I think `mailserver` makes sense. I wouldn't bother changing it in the XML. It's likely quite an underused property. But, if it makes you happy to keep it consistent, @GaneshSPatil , I can understand. :)

GaneshSPatil: @arvindsv -- we are adding an XML migration to move the server attribute into a separate child tag for configurations such as: `Server URLs`, `Pipeline artifacts`, `Job Timeout`, `Command Repository`, etc. Hence thought of renaming this too as part of same migration. :)

PR #6575, opened 2019-06-25
Fix arguments for `git submodule foreach --recursive`

akshaydewan:
Parsing the command following `git submodule foreach --recursive` has changed in Git 2.22.
This was causing the command `git submodule foreach --recursive git clean -dffx` to fail, because git thinks that `-d` is part of the arguments for `foreach`


akshaydewan:
@ketan I'm merging this to unblock the pipeline, but if there are any changes to be done, let me know.

PR #6576, opened 2019-06-25
Fix installer upgrade on ubuntu and centos

GaneshSPatil:
* Prior to GoCD v19.6.0, use '/etc/init.d/SCRIPT status' command to
  determine whether GoCD service is running.

* From GoCD v19.6.0, use 'invoke-rc.d SCRIPT status' command to
  determine whether GoCD service is running.

---

Instructions followed:
```shell
# Install 19.5.0 GoCD agent
echo "deb https://download.gocd.org /" | sudo tee /etc/apt/sources.list.d/gocd.list
curl https://download.gocd.org/GOCD-GPG-KEY.asc | sudo apt-key add -
apt-get update
apt-get install default-jre
apt-get install -y go-agent

# Start 19.5.0 GoCD agent
/etc/init.d/go-agent start

# Verify status of Running GoCD agent
/etc/init.d/go-agent status

# Upgrade to 19.6.0 GoCD agent
dpkg -i go-agent_19.6.0-9353_all.deb

# Verify status of Running GoCD agent
service go-agent status
invoke-rc.d go-agent status
```


GaneshSPatil:
To be merged post successful verification of upgrade scenario on all platforms.
\cc: @adityasood @ketan

adityasood:
Couple of observations:

1. While go user is able to query the status of the go-server and go-agent service, it is unable to start/stop/restart the service. Root user is only allowed to do that currently.

2. The message on a fresh install of both server and agent still points to 

`Once that is done start the GoCD Agent with `/etc/init.d/go-agent start'`

This needs to change to something like `service go-agent start` or `/usr/share/go-agent/bin/go-agent start`

ketan:
> This needs to change to something like `service go-agent start` or `/usr/share/go-agent/bin/go-agent start`

It's a bit difficult to say for sure what option was used to perform an install. Perhaps checking for `/etc/init.d/SERVICE` `/etc/init/SERVICE.conf` and `/usr/lib/systemd/system/SERVICE.service` and display the message. @GaneshSPatil ?

installers/linux/deb/before-upgrade.sh.erb
```diff
@@ -19,8 +19,22 @@
...
+      mkdir -p /var/run/<%= name %>-running-check-from-deb-package
+      touch /var/run/<%= name %>-running-check-from-deb-package/running
+      invoke-rc.d <%= name %> stop || true
+    }
+
+    # Prior to GoCD v19.6.0 Check if go-agent is running using /etc/init.d
+    if [ -f "/etc/init.d/<%= name %>" ]; then
+      # Use /etc/init.d/SCRIPT status to determine service is running or not.
+      if /etc/init.d/<%= name %> status > /dev/null 2>&1; then
+        stopGoAgentServiceOnUpgrade
```
ketan: Should we not be using `/etc/init.d` to stop instead of `invoke-rc.d`?

GaneshSPatil: > a bit difficult to say for sure what option was used to perform an install. Perhaps checking for `/etc/init.d/SERVICE`

@ketan -- Not sure if I understand the problem with using `/usr/share/go-agent/bin/go-agent start`.

GaneshSPatil:
> a bit difficult to say for sure what option was used to perform an install. Perhaps checking for `/etc/init.d/SERVICE`

@ketan -- Not sure if I understand the problem with using `/usr/share/go-agent/bin/go-agent start`.

installers/linux/shared/partials/_go-agent_post_install_message.sh.erb
```diff
@@ -20,4 +20,10 @@
     echo "Now please edit /usr/share/go-agent/wrapper-config/wrapper-properties.conf and set the server to the url address of your GoCD Server (https://example.com:8154/go)."
     echo "To ensure full end-to-end transport security between the agent and server"
     echo "see https://docs.gocd.org/current/installation/ssl_tls/end_to_end_transport_security.html"
-    echo "Once that is done start the GoCD Agent with '/etc/init.d/go-agent start'"
+    if [ -f "/etc/init.d/go-agent" ]; then
+      echo "Once that is done start the GoCD Agent with '/etc/init.d/go-agent start'"
+    elif [ -f "/etc/systemd/system/go-agent.service" ]; then
+      echo "Once that is done start the GoCD Agent with 'systemctl start go-agent.service'"
```
ketan: I think it's just supposed to be `systemctl start go-{agent,server}`.

GaneshSPatil: @ketan -- yes will change!

GaneshSPatil: Just a question, as this message only shows up on the agent installation (similar one on server installation). Is there a need to specify both `{agent,server}` ?

ketan: I meant to say `systemctl start go-agent` in the agent post install, and `systemctl start go-server` in the server post install. Sorry, should have been clearer :)

GaneshSPatil:
Merging this PR using merge commit (all commits) instead of squash and merge (squashing into single commit). This will help to revert a particular commit if in case something goes wrong!

\cc: @ketan @adityasood

Issue #6567, opened 2019-06-21
Order pipeline group names in alphabetical order when cloning a pipeline


ryudice:
##### Issue Type

- Bug Report

##### Summary

Order Pipeline group names in alphabetical order when cloning a new pipeline.

We have a lot of grups and it seems they are shown in order of creation at the moment.

![image](https://user-images.githubusercontent.com/125211/59928980-1b90e380-93fd-11e9-81a9-fba60c839760.png)

###### Basic environment details

* Go Version: `19.3.0`

##### Steps to Reproduce

1. Clone a pipeline
2. 
3. 
4. 

##### Expected Results

The dropdown in the picture should order group names by alphabetical order so that they are easy to find

![image](https://user-images.githubusercontent.com/125211/59929087-5135cc80-93fd-11e9-8b0e-b3b2af37c367.png)

##### Actual Results
They seem to be order by creation time.

viraj2712:
As we are ordering pipeline groups alphabetically while cloning the pipeline, should we make it consistent for `Move to` as well?

![Screenshot 2019-06-27 at 4 42 06 PM](https://user-images.githubusercontent.com/34642373/60261882-86a95100-98fa-11e9-91ba-3b02ff496ebb.png)

\cc: @maheshp @bdpiparva @ketan

ketan:
Makes sense to me. Please make sure it is ordered in a case insensitive
manner.

kritika-singh3:
Fixed in https://github.com/gocd/gocd/pull/6577 as a part of GoCD version 19.6.0

chadlwilson:
Should this have changed the pipeline group ordering when viewing on the Dashboard, or just in config drop-down lists?

The ordering on the dashboard seems to be creation time, which is largely beyond one's control when using config repositories to import an existing set. You don't even seem to be able to change it in a custom View :(

PR #6577, opened 2019-06-26
Order pipeline groups alphabetically

viraj2712:
Issue: #6567

Order pipeline group names case-insensitively and alphabetically while cloning the pipeline as well as moving pipeline to another group.

### DIFF
--- a/server/webapp/WEB-INF/rails/app/controllers/admin/pipelines_controller.rb
+++ b/server/webapp/WEB-INF/rails/app/controllers/admin/pipelines_controller.rb
@@ -286,6 +286,10 @@ module Admin
         group_array.push({:group => group})
         group_list.push(group)
       end
+
+      group_array = group_array.sort_by { |group| group[:group].downcase }
+      group_list = group_list.sort_by(&:downcase)
+
       assert_load :groups_json, group_array.to_json
       assert_load :groups_list, group_list
     end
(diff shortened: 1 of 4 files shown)


viraj2712:
Sorted pipeline groups while moving the pipeline to another group:

Screenshot 2019-06-28 at 4 59 43 PM

Sorted pipeline groups while cloning the pipeline:

Screenshot 2019-06-28 at 4 59 25 PM

\cc: @ketan @bdpiparva

server/webapp/WEB-INF/rails/app/controllers/admin/pipelines_controller.rb
```diff
@@ -286,6 +286,10 @@ def load_group_list
         group_array.push({:group => group})
         group_list.push(group)
       end
+
+      group_array = group_array.sort_by { |group| group[:group].downcase }
+      group_list = group_list.sort_by { |group| group.downcase }
```
ketan: The ruby way would be to do: `group_list.sort_by(&:downcase)`, but what you have works too.

viraj2712: makes sense, will update it

viraj2712: updated

Issue #6468, opened 2019-06-12
[Bug] - Shouldn't allow to save pipeline config with empty template


gocd-ci-user:
> Migrated from [Trello Card 133](https:&#x002f;&#x002f;trello.com&#x002f;c&#x002f;7lrlBvKY)

Basically, we should not allow the user to submit the form when there are no templates and they enabled `Use Template`.

Perhaps we should even show the message as a warning box (yellow?) instead of the informational light blue box.

Either way, we should show a validation error that no template is selected (or something like that) prevents the user from submitting the form.

## Attachments

* ![Screenshot 2019-06-10 at 15.16.28.png](https://raw.githubusercontent.com/gocd/gocd/c65082a0b39ec9e0e5fa5565b0e7a2421487e15d/.github/trello-attachments/bd10562403e40fd8e36ecc962f605e1ec12e45d23da728bbbebab4d704413a6a/Screenshot_2019-06-10_at_15.16.28.png)

marques-work:
@ibnc Note that you should set the `forceOpen` attribute on the `<AdvancedSettings/>` component if an error is present so that the panel is force-expanded to reveal the validation error.

ankitsri11:
For the new workflow, it works fine. Verified on GoCD Version: 19.6.0 (9359-ea6a0afe0b8a34c617a47b70bde258a383797b1c). 

![Screenshot 2019-06-26 at 13 51 31](https://user-images.githubusercontent.com/6838313/60214019-89804500-9819-11e9-8564-defec8684dd3.png)

@marques-work @ibnc However, for the old workflow, it is still broken. See below:

Error in the go-server.log file:

```
2019-06-26 13:52:48,371 ERROR [qtp208645117-33] Rails:-2 - Exception:
NoMethodError: undefined method `getJobs' for nil:NilClass
        /private/tmp/go-server-19.6.0/work/jetty-0.0.0.0-8153-cruise.war-_go-any-/webapp/WEB-INF/rails/app/controllers/admin/pipelines_controller.rb:103:in `block in create'
        /private/tmp/go-server-19.6.0/work/jetty-0.0.0.0-8153-cruise.war-_go-any-/webapp/WEB-INF/rails/app/controllers/admin_controller.rb:116:in `save'
        /private/tmp/go-server-19.6.0/work/jetty-0.0.0.0-8153-cruise.war-_go-any-/webapp/WEB-INF/rails/app/controllers/admin_controller.rb:40:in `save_page'
[...]

PR #6583, opened 2019-06-27
Prevent NPE when trying to add a pipeline via templates when no templates exist

marques-work:
Fixes https://github.com/gocd/gocd/issues/6468 (well, a peripheral error discovered here)

  - Remember radio button state over POST (previously always reset to "use stages")
  - Display an error indicating that a template must be selected
  - Render the "no templates exist" message as a warning when an error on `templateName` is present


marques-work:
Going to merge this now that the build checks have passed. @kierarad has reviewed this work as well.

Issue #6475, opened 2019-06-12
[Bug] - Group admin gets 403 when adding pipeline


gocd-ci-user:
> Migrated from [Trello Card 130](https:&#x002f;&#x002f;trello.com&#x002f;c&#x002f;XBzDHkE4)

Steps to recreate:
1 - enable security
2 - create normal user
3 - give admin access to pipeline group
4 - follow link to "Add pipeline" for that group
5 - get 403 Forbidden

ankitsri11:
Verified on GoCD Version: 19.6.0 (9370-8e1b89e40fbf882e547888ed7c3ac46226eacc95)

@kierarad @marques-work however, pipeline group admin user creating a pipeline with empty template process is inconsistent with creating via the goadmin user. See below:

![pipeline-grp-admin](https://user-images.githubusercontent.com/6838313/60216822-eda60780-981f-11e9-9e70-063bde53e207.gif)

Should allow pipeline group admin user to enable template option similar to below:

![pipeline-grp-admin2](https://user-images.githubusercontent.com/6838313/60217201-08c54700-9821-11e9-8ea1-3564adcb2cc5.gif)

kierarad:
@ankitsri11 ok, I'll take a look

kierarad:
- [ ] Fix Template Toggle - group admin should not receive 403
- [ ] Fix Pipeline Group Dropdown - admin should not receive 403

ankitsri11:
Verified on GoCD Version: 19.6.0 (9407-ee11f624e4a6cbb4e4979d32ea512dcfe2d1777a)

PR #6589, opened 2019-07-01
[Bug] Template option & pipeline group dropdown broken for group admin (new add pipeline flow)

kierarad:
Fixes 2 issues found as part of #6475. 

- This fixes the template toggle on the new add pipeline page in the scenario where the call for available templates fails(ie if get a 403 when trying to retrieve templates because don't have template access).
- This also gives group admin's access to the pipeline groups index so we can populate the pipeline group dropdown on the new add pipeline page

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/template_editor_spec.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/template_editor_spec.tsx
@@ -25,7 +25,7 @@ import {TemplateEditor} from "../template_editor";
 
 describe("AddPipeline: TemplateEditor", () => {
   const helper = new TestHelper();
-  const isUsingTemplate: Stream<boolean> = stream(false);
+  const isUsingTemplate: Stream<boolean> = stream();
   let config: PipelineConfig;
 
   beforeEach(() => {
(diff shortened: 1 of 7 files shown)


server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/template_editor_spec.tsx
```diff
@@ -60,6 +60,20 @@ describe("AddPipeline: TemplateEditor", () => {
...
   });
+
+  it("should not display dropdown and should display flash when cache fails to populate", () => {
+    helper.unmount();
+    helper.mount(() => <TemplateEditor pipelineConfig={config} isUsingTemplate={isUsingTemplate} cache={new FailedTestCache()}/>);
+    helper.click(helper.byTestId("switch-paddle"));
+    expect(isUsingTemplate()).toBeTruthy();
+
+    const flash = helper.byTestId("flash-message-warning");
+    expect(flash).toBeTruthy();
```
marques-work: `toBeTruthy()` is ok, but `.toBeInDOM()` is probably better.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/template_editor_spec.tsx
```diff
@@ -60,6 +60,20 @@ describe("AddPipeline: TemplateEditor", () => {
     expect(isUsingTemplate()).toBe(false);
     expect(config.template() === undefined).toBeTruthy();
   });
+
+  it("should not display dropdown and should display flash when cache fails to populate", () => {
+    helper.unmount();
+    helper.mount(() => <TemplateEditor pipelineConfig={config} isUsingTemplate={isUsingTemplate} cache={new FailedTestCache()}/>);
+    helper.click(helper.byTestId("switch-paddle"));
```
marques-work: There's also `helper.clickByDataTestId()` which does just this. Not a problem, just letting you know.

PR #6601, opened 2019-07-04
Allow only known users to login at each auth config

GaneshSPatil:
This PR addresses:

- [x] A Config Migration to move `allowOnlyKnownUsersToLogin` from `Security` tag to `AuthConfig` tag.
- [x] Remove `allowOnlyKnownUsersToLogin` field from Java SecurityConfig Entity.
- [x] Add `allowOnlyKnownUsersToLogin` field on Java SecurityAuthConfig Entity.
- [x] Use `allowOnlyKnownUsersToLogin` from SecurityAuthConfig to explicitly add a user.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
@@ -24,6 +24,9 @@ import java.util.Collection;
 @ConfigTag("authConfig")
 @ConfigCollection(value = ConfigurationProperty.class)
 public class SecurityAuthConfig extends PluginProfile {
+    @ConfigAttribute(value = "allowOnlyKnownUsersToLogin", optional = true)
+    protected Boolean allowOnlyKnownUsersToLogin;
+
     public SecurityAuthConfig() {
         super();
     }
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
@@ -25,7 +25,7 @@ import java.util.Collection;
 @ConfigCollection(value = ConfigurationProperty.class)
 public class SecurityAuthConfig extends PluginProfile {
     @ConfigAttribute(value = "allowOnlyKnownUsersToLogin", optional = true)
-    protected Boolean allowOnlyKnownUsersToLogin;
+    protected Boolean allowOnlyKnownUsersToLogin = false;
 
     public SecurityAuthConfig() {
         super();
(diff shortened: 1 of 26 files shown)


server/src/test-integration/java/com/thoughtworks/go/config/GoConfigMigrationIntegrationTest.java
```diff
@@ -1454,6 +1454,54 @@ public void shouldMigrateEverythingAsItIs_Migration123To124() throws Exception {
...
+                "    </security>\n" +
+                "  </server>\n";
+
+        String configXml = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>" +
+                        "<cruise schemaVersion=\"124\">" + originalConfig + "</cruise>";
+
+        final String migratedXml = migrateXmlString(configXml, 124, 125);
+
+        XmlAssert.assertThat(migratedXml).nodesByXPath("//security").doNotHaveAttribute("allowOnlyKnownUsersToLogin");
+        XmlAssert.assertThat(migratedXml).nodesByXPath("//authConfig").haveAttribute("allowOnlyKnownUsersToLogin", "false");
```
ketan: Since `false` is default, I'd think the attribute should not be present.

config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
```diff
@@ -36,6 +39,15 @@ public SecurityAuthConfig(String id, String pluginId, Collection<ConfigurationPr
         this(id, pluginId, configProperties.toArray(new ConfigurationProperty[0]));
     }
 
+    public SecurityAuthConfig(String id, String pluginId, boolean allowOnlyKnownUsersToLogin, ConfigurationProperty... props) {
```
ketan: No constructors, please.

config/config-api/src/main/java/com/thoughtworks/go/config/SecurityAuthConfig.java
```diff
@@ -66,4 +78,12 @@ public boolean hasRole(PluginRoleConfig role) {
     private AuthorizationMetadataStore metadataStore() {
         return AuthorizationMetadataStore.instance();
     }
+
+    public Boolean getAllowOnlyKnownUsersToLogin() {
```
ketan: Use lombok:

```java
@Getter
@Setter
@EqualsAndHashCode(callSuper = true)
@Accessors(chain = true)
class Foo {
    // exclude some fields from equals/hashcode
    @EqualsAndHashCode.Exclude
    private final ConfigErrors configErrors = new ConfigErrors();
}
```

GaneshSPatil:
Closing in favor of https://github.com/gocd/gocd/pull/7178

PR #6610, opened 2019-07-08
Remove rails based unversioned stages history API

GaneshSPatil:


GaneshSPatil:
The API models such as `PaginationAPIModel`, `StageHistoryAPIModel` cannot be removed as those are consumed by other API entities.
Removing only the routes and the controller as of now.

PR #6614, opened 2019-07-10
Upgrade essential node packages

marques-work:
keeping up to date and trying to fix a number of security alerts detected by GH on our node packages. Any changes in the actual source are a result of satisfying `eslint` and the newer version of TypeScript. They don't have an impact on the generated production assets.


ketan:
`false as boolean` seems redundant, is this something new?

marques-work:
> `false as boolean` seems redundant, is this something new?

It's a really strange behavior with TS expansion of `boolean` to `true|false` I think. A know a number of folks have run into issues with this (or similar).

Oddly, `tsc` doesn't complain about this, which might be a cleaner way of working around it:

```typescript
class Foo {
  private prop: Stream<boolean> = stream(); // TS doesn't complain while using an empty stream for default init
  constructor() {
    this.prop(false); // and this seems to be ok
  }
}
```

marques-work:
@ketan I've found that omitting an explicit type fixes most cases. For the ones that omission doesn't fix, the best I could come up with that doesn't need to declare new wrapper functions is:

```typescript
const flag: (val?: boolean) => Stream<boolean> = stream;

class Foo {
  private falsey: Stream<boolean> = flag(false);
  private truthy: Stream<boolean> = flag(true);
}
```

PR #6618, opened 2019-07-12
Added backup config API and a modal to configure backups

ketan:


ketan:
The Backup Config API is available under `/api/config/backup` and supports `GET/PUT/POST/DELETE` HTTP verbs.

There are a few things I've done in order to simplify the implementation and end-user experience (for API callers):
- No conflict checking based on etags/if-none-match.
- The API is quite lenient, in that a `GET` will always return an empty "no-op" backup config, even if there's none on the server, no `<backup/>` is the same as `<backup emailOnSuccess=false emailOnFailure=failure />`, and the API will render the JSON corresponding to this config.
- A `PUT/POST` will do the same operation, and always replace the existing config (if any)
- A `DELETE` is added in the API (for completeness sake), but not exposed in the UI (I was not sure where to put a "Delete" button in the UI, happy to hear suggestions)


The JSON representation of the backup API looks like:

```json
{
  "_links" : {
    "doc" : {
      "href" : "https://api.gocd.org/19.6.0/#backup-config"
    },
    "self" : {
      "href" : "http://localhost:8153/go/api/config/backup"
    }
  },
  "email_on_failure" : false,
  "email_on_success" : false,
  "post_backup_script" : null,
  "schedule" : "0 15 10 * * ?"
}
```

rajiesh:
@ketan IMO the delete can be called on the configure modal say with a button like `RESET`

ketan:
I thought of a "reset" like button in the config, but we'd probably need
@naveenbhaskar's input to make sure that there's some confirmation before a
"reset" is actually performed.

- Ketan

PR #6626, opened 2019-07-16
SecretParamResolver errors out for invalid secret config id #000

maheshp:
* Previously secret params were validated(by applying rules) on config
  save and that would ensure a valid secret config id in a secret param.
  Since we have moved to applying rules at runtime, a secret_param can
  be configured with a secret config id which can be invalid.
* Added check to verify presence of secret config id before secret
  resolution.


maheshp:
This scenario is taken care by RulesService, hence closing this PR.

PR #6667, opened 2019-07-24
Remove feature toggle for add pipeline button styling

varshavaradarajan:
@marques-work - I don't know why this was a feature toggle. Perhaps to get some user feedback? But this seems more like a UI decision rather than a feature toggle to have. Besides, it is disabled for 2 months now. Feel free to close the PR if you want to keep this around.


marques-work:
@varshavaradarajan hi varsha, actually, I was working on this myself today but you beat me to it :). Yes, it was for user testing, but we decided earlier today that we should just remove the toggle and make it the canonical style. thanks for taking care of it.

marques-work:
@varshavaradarajan though actually, we decided to go the opposite way - we want the button to be a primary button rather than stay the way it is right now.

Issue #6666, opened 2019-07-23
Bug: upgrading to v19.6 causes 'autostart at boot' issue after server restart


mehmetcavus:
##### Issue Type
- Bug Report

##### Summary

- Upgrading to v19.6 causes autostart at boot issue after server restart, Ubuntu Server 18.04 (bionic).

###### Basic environment details

* Go Version before upgrade: `19.5.0` installed via deb package
* Go Version: `19.6.0`
* JAVA Version: `OpenJDK 64-Bit Server VM (build 11.0.3+7-Ubuntu-1ubuntu218.04.1, mixed mode, sharing) 2019-04-16`
* OS: `Ubuntu 18.04.2 LTS (bionic)`

###### Additional Environment Details

There is no specific setting for this environment. Upgrading to v19.6 migrates previous configuration file to `/usr/share/go-server/wrapper-config/wrapper-properties.conf`, all lines are commented with # .

##### Steps to Reproduce

1. ssh to ubuntu server
2. Upgrade existing GoCD v19.5 installation to v19.6 via `sudo dpkg -i <package-name>.deb` or `sudo apt-get install go-server`
3. Run `sudo systemctl status go-server` and `journalctl -xe`, see that go-server is running
4. Restart ubuntu server via 'sudo reboot'
5. Run `sudo systemctl status go-server`and `journalctl -xe`, see that go-server is not running

##### Expected Results
After restarting ubuntu server, go-server and go-agent should autostart successfully as it always does before upgrading to v19.6.0.

##### Actual Results
go-server and go-agent starts successfully after upgrading, but does not autostart after ubuntu server restart. Tested purging go-server and go-agent, reinstalling both, reloading backup files. Nothing changed.

##### Log snippets
`sudo systemctl status go-server` output before Ubuntu restart:
```
 go-server.service - go-server
   Loaded: loaded (/etc/systemd/system/go-server.service; enabled; vendor preset: enabled)
   Active: active (running) since Tue 2019-07-23 22:34:10 +03; 5s ago
  Process: 4776 ExecStop=/usr/share/go-server/bin/go-server stop sysd (code=exited, status=0/SUCCESS)
  Process: 4861 ExecStart=/usr/share/go-server/bin/go-server start sysd (code=exited, status=0/SUCCESS)
 Main PID: 4957 (wrapper-linux-x)
    Tasks: 49 (limit: 2312)
   CGroup: /system.slice/go-server.service
[...]

PR #6671, opened 2019-07-25
Fix Linux Installers (#6666)

GaneshSPatil:
* Change PIDDIR from /var/run/{go-agent,go-server} to /var/lib/{go-agent,go-server}/run


GaneshSPatil:
Installers can be found https://build.gocd.org/go/tab/build/detail/installers-PR/1671/dist/1/dist

Issue #6674, opened 2019-07-25
Issues with upgrading to 19.6 


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Upgrading to 19.6.0 version of GoCD server and agent has below issues in migrating user defined working directory and server port

- [x] On Linux installations if user had updated `/etc/default/go-server` to change the default values of `GO_SERVER_PORT` from `8153` and `GO_SERVER_SSL_PORT` from `8154` to some value of their choice, on an upgrade to 19.6 server starts up on default ports `8153` and `8154`

**Workaround**: On server machine edit the file `/usr/share/go-server/wrapper-config/wrapper-properties.conf`

- Look for an entry like `wrapper.java.additional.<index number>=-Dcruise.server.port<non_ssl_port_set_by_user>` and `wrapper.java.additional.<some index number>=-Dcruise.server.ssl.port<ssl_port_set_by_user>`. 

- Update it to `wrapper.java.additional.<index number>=-Dcruise.server.port=<non_ssl_port_set_by_user>` for both ssl port and non-ssl port. Note the `=` sign included

- Restart the server

- [x] On Linux installations if user had updated server/agent working directories in `/etc/default/go-{server, agent}` using `SERVER_WORKING_DIR` and `AGENT_WORKING_DIR` its not honored on an upgrade to 19.6. Working directories gets defaulted to `/var/lib/go-server` and `/var/lib/go-agent`

**Workaround**: This has to be done before starting the server for first time after upgrading to 19.6.0 to avoid issue of missing data. On server/agent machine edit the file `/usr/share/go-{server, agent}/wrapper-config/wrapper-properties.conf`

- Add an entry as `wrapper.working.dir=<user specified working directory>` and save the file
- Start the server

Note: Doing this workaround after the first start of 19.6.0 server/agent process may lead to loss of some data. If you upgraded the server/agent and noticed that the working directory is picked up wrongly and then applied this workaround and restarted the server/agent may cause that some data goes into `/var/lib/go-{server, agent}` folder and does not show up on the server

PR #6676, opened 2019-07-25
Use {server,agent} work dir specified from older installations (#6674)

GaneshSPatil:


GaneshSPatil:
@rajiesh @adityasood -- The installers for testing could be found https://build.gocd.org/go/pipelines/installers-PR/1673/dist/1

Issue #6649, opened 2019-07-22
Make `New Pipeline` a dropdown component to allow the user to go to the PaC builder or the conventional add pipeline flow.


marques-work:
This will need a toggle.

Intentionally separate from #6598 so we don't have to wait for final styling to get something that users can test.

See #6678.

Issue #6598, opened 2019-07-03
Finalize styles for `New Pipeline` dropdown


marques-work:
Related to #6649.

kierarad:
Currently implemented as a dropdown which behaves like a menu (shows options on hover). But interaction should be considered further:

As a dropdown should it behave more like a menu (the application header is an example) or like a selection dropdown (the group by pipelines groups/environments is an example)?

<img width="225" alt="Screen Shot 2019-08-01 at 2 13 27 PM" src="https://user-images.githubusercontent.com/3814794/62340670-38591600-b495-11e9-9dc0-d001c8401a3d.png"> <img width="334" alt="Screen Shot 2019-08-01 at 2 15 53 PM" src="https://user-images.githubusercontent.com/3814794/62340682-4018ba80-b495-11e9-82c3-14b989a90ecb.png">

For a selection dropdown the current selection generally populates the text of the dropdown when closed. But we would want to have “New Pipeline” (or something similar) as the text when the dropdown is closed. So this behavior better matches a menu. 

<img width="301" alt="Screen Shot 2019-08-01 at 2 17 38 PM" src="https://user-images.githubusercontent.com/3814794/62340708-57f03e80-b495-11e9-8426-ef2717741f1a.png">


Another consideration of the behavior is whether the options display when the user clicks or hovers. For a menu dropdown the typical behavior (at least in gocd) is for the options to display on hover. But for a button the typical behavior is for an action to occur (ie display options) when you click it. The two are at odds with each other. So which should we do? On click or on hover?
Currently it’s implemented on hover to be consistent with the other menus.

streisguth:
I have two viable options for the ‘add new pipeline’ button, and I plan to do some quick user testing next week before I make a call on which will be the best choice. Nielson Norman recommends using a ‘split button’ (prototype #2) for _actions_ or _tool selection_, while recommending a more common ‘menu’ for something that take a user to another page or screen (prototype #1). Our use case covers both scenarios, and will set a bit of precedence in the product, hence the additional user testing.

Prototype #1: Menu
https://thoughtworks-products.invisionapp.com/share/6XTAN6AVRMT

Prototype #2: Split Button
https://thoughtworks-products.invisionapp.com/share/86TAN6CN7UC

The ‘add pipeline’ pages in the mockup are grabbed from old prototypes for the sake of illustration. I also have designs for hover states (not included in the prototype).

NN Group article for reference:
https://www.nngroup.com/articles/split-buttons/ (edited)

PR #6678, opened 2019-07-25
New pipeline button dropdown

marques-work:
This PR makes the New Pipeline button into a dropdown so users can select between creating a pipeline via the pipelines as code page or the add pipeline wizard. 

This is meant for the test drive experience and therefore behind a toggle `new_pipeline_dropdown`.
Related issue #6649 
Design considerations & feedback as part of #6598


marques-work:
@kierarad LGTM. I'll merge this as it's behind a toggle. We'll style this in a separate PR once we've finalized those decisions.

Issue #6682, opened 2019-07-26
Environments Get API does not honor If-None-Match header


SnehaPL:
##### Issue Type

- Bug Report

##### Summary

Environments Get API should return 304 (Not Modified) when appropriate  If-None-Match header is specified.

###### Basic environment details

* Go Version: 19.7.0 (9552-aa833fbf095e66489c765751dfa23daa741c35b4)
* JAVA Version: 1.8.0_211
* OS: Mac OS X 10.14.6

##### Expected Results
Environments Get API should return 304 (Not Modified) when appropriate  If-None-Match header is specified.

##### Actual Results
Environments Get API return 200 (Ok) when appropriate  If-None-Match header is specified.

GaneshSPatil:
Fixed as part of #6683

rajiesh:
Verified on 19.8.0 (9746-6bbad2a99f2a38bfc31210777c22d9bfda941e68)

PR #6683, opened 2019-07-26
Environments Get API returns 304 (Not Modified) when appropriate If-None-Match header is specified. (#6682)

SnehaPL:
The API used to render an ETag, but never consume it to return a 304 instead.

### DIFF
--- a/api/api-environments-v2/src/test/groovy/com/thoughtworks/go/apiv2/environments/EnvironmentsControllerV2Test.groovy
+++ b/api/api-environments-v2/src/test/groovy/com/thoughtworks/go/apiv2/environments/EnvironmentsControllerV2Test.groovy
@@ -120,6 +120,44 @@ class EnvironmentsControllerV2Test implements SecurityServiceTrait, ControllerTr
           .isNotFound()
           .hasJsonMessage(controller.entityType.notFoundMessage("env1"))
       }
+
+      @Test
+      void 'should render not modified when ETag matches'() {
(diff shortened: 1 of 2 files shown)


vrushaliwaykole:
LGTM

api/api-environments-v2/src/test/groovy/com/thoughtworks/go/apiv2/environments/EnvironmentsControllerV2Test.groovy
```diff
@@ -120,6 +120,44 @@ class EnvironmentsControllerV2Test implements SecurityServiceTrait, ControllerTr
...
+        env1.addPipeline(new CaseInsensitiveString("Pipeline1"))
+        env1.addPipeline(new CaseInsensitiveString("Pipeline2"))
+
+        when(entityHashingService.md5ForEntity(env1)).thenReturn("md5-hash")
+        when(environmentConfigService.getEnvironmentConfig(eq("env1"))).thenReturn(env1)
+
+        getWithApiHeader(controller.controllerPath("env1"), ['if-none-match': 'md5-hash'])
+
+        assertThatResponse()
+          .isNotModified()
```
ketan: Probably also assert that the body is blank

ketan: Apparently, `isNotModified` does an assert for empty body.

Issue #6684, opened 2019-07-26
Variable GO_ENVIRONMENT_NAME lost on v19.6.0


aamattos:
##### Issue Type

- Bug Report

##### Summary

With the upgrade to v19.6.0, The variable GO_ENVIRONMENT_NAME is no longer injected.

##### Environment

Running GoCd in an Openshift platform with Kubernetes elastic agents

###### Basic environment details

* Go Version: `19.6.0 (9515-4b674c10941b6c27d7ec2a28dd946518d9211b7a)`
* JAVA Version: `12.0.1`
* OS: `Linux 3.10.0-862.14.4.el7.x86_64`

##### Steps to Reproduce

1. Update GoCd and elastic agentes to version 19.6.0
2. Run a pipeline in an environmnet

##### Expected Results

Variable GO_ENVIRONMENT_NAME should be injected.

##### Actual Results

Variable GO_ENVIRONMENT_NAME is not injected.

##### Code snippets/Screenshots

V19.3.0:

![v19 6 0](https://user-images.githubusercontent.com/1260299/61943481-6c16d600-af93-11e9-98cb-1f1ec16006f3.png)

V19.6.0:

![v19 3 0](https://user-images.githubusercontent.com/1260299/61943483-6d480300-af93-11e9-89cd-98bca9d91f47.png)

rajiesh:
Verified on 19.7.0 (9560-dfc88b483e64eb3cc1205af3b3ff319ec71e085a)

PR #6685, opened 2019-07-26
Include GO_ENVIRONMENT_NAME environment variable while creating environment variable context (#6684)

GaneshSPatil:


server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -314,15 +315,15 @@ private Work createWork(final AgentInstance agent, final JobPlan job) {
         }
     }
 
-    private EnvironmentVariableContext getEnvironmentVariableContextFromEnvironment(String pipelineName) {
+    EnvironmentVariableContext getEnvironmentVariableContextFromEnvironment(String pipelineName) {
         EnvironmentConfig environmentConfig = environmentConfigService.environmentForPipeline(pipelineName);
```
maheshp: Optional: 
Consider renaming `environmentConfig` to something like `pipelineEnvironment` or `environmentForPipeline`?

server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -314,15 +315,15 @@ private Work createWork(final AgentInstance agent, final JobPlan job) {
...
         if (environmentConfig == null) {
             return null;
         }
 
         secretParamResolver.resolve(environmentConfig);
 
-        EnvironmentVariableContext environmentVariableContext = new EnvironmentVariableContext();
+        EnvironmentVariableContext environmentVariableContext = new EnvironmentVariableContext(GO_ENVIRONMENT_NAME, CaseInsensitiveString.str(environmentConfig.name()));
```
maheshp: Optional:

Can create context using `environmentConfig.createEnvironmentContext()`

maheshp: @GaneshSPatil and I spoke about this, calling `environmentConfig.createEnvironmentContext()` would mean loosing all the resolved environment variables. So let's not change it.

PR #6689, opened 2019-07-29
Remove sysvinit-tools dependency from RPM distro

GaneshSPatil:


GaneshSPatil:
Installers could be found: https://build.gocd.org/go/pipelines/installers-PR/1674/dist/1

Issue #6611, opened 2019-07-09
Check Material internal API


marques-work:
Given a material config and plugin ID, we should be able to:

1. Do a basic connection test
2. Test if said material contains config repo definition files
3. Return the discovered set of definition files (according to effective plugin file patterns)

Might be desirable to add as another endpoint to the internal API defined in #6612

PR #6694, opened 2019-07-30
Config Repo Contract V3

ibnc:
* Adds the ability to fetch configuration files for a plugin and repo

Part of #6611 

Plugin PRs implementing endpoint:
JSON: https://github.com/tomzo/gocd-json-config-plugin/pull/49
YAML: https://github.com/tomzo/gocd-yaml-config-plugin/pull/126

### DIFF
--- a/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/configrepo/ConfigFileList.java
+++ b/plugin-infra/go-plugin-access/src/main/java/com/thoughtworks/go/plugin/access/configrepo/ConfigFileList.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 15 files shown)


ibnc:
@tomzo @marques-work 

Would one of y'all mind reviewing this?

tomzo:
@ibnc  once you add the test, I'll merge this, then I'll release plugins and make a PR to bundle the new versions.

arvindsv:
This should only be merged after 19.7 (which should be coming out in the next day or so). There is a problem with the EC2 nodes running out of IOPS, causing some regression pipelines to fail, which is making it harder than usual to get the right builds out. Thank you!

ibnc:
@tomzo Is there a reason why we can't just make `ConfigFilesResponseMessage#gitFiles` return an empty list if files is null? Receiving null is essentially the same as receiving the empty list. I do agree tho that the plugins should be returning an empty list rather than null tho

tomzo:
My point was that if a plugin returns an invalid message, e.g.
```json
{}
```
Or
```json
{
   "files-bogus-field-name" : [ "a.yaml"] 
}
```
Then in both cases it would deserialize to an object with `null` in (the correct) `files` field.

ibnc:
@tomzo would you mind taking another look?

tomzo:
@ibnc thanks, looks perfect now.

I see that 19.7 is out, so we just need a rebase and this can be merged.

PR #6703, opened 2019-08-01
PAC list config files internal api endpoint

ibnc:
Part of https://github.com/gocd/gocd/issues/6611

### DIFF
--- a/api/api-pipelines-as-code-internal-v1/src/main/java/com/thoughtworks/go/apiv1/pipelinesascodeinternal/PipelinesAsCodeInternalControllerV1.java
+++ b/api/api-pipelines-as-code-internal-v1/src/main/java/com/thoughtworks/go/apiv1/pipelinesascodeinternal/PipelinesAsCodeInternalControllerV1.java
@@ -24,18 +24,27 @@ import com.thoughtworks.go.api.spring.ApiAuthenticationHelper;
 import com.thoughtworks.go.api.util.GsonTransformer;
 import com.thoughtworks.go.api.util.HaltApiResponses;
 import com.thoughtworks.go.api.util.MessageJson;
+import com.thoughtworks.go.apiv1.pipelinesascodeinternal.representers.ConfigFileListsRepresenter;
 import com.thoughtworks.go.apiv8.admin.shared.representers.PipelineConfigRepresenter;
+import com.thoughtworks.go.apiv8.admin.shared.representers.materials.MaterialsRepresenter;
 import com.thoughtworks.go.apiv8.admin.shared.representers.stages.ConfigHelperOptions;
 import com.thoughtworks.go.config.ConfigRepoPlugin;
 import com.thoughtworks.go.config.CruiseConfig;
 import com.thoughtworks.go.config.GoConfigPluginService;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.materials.PasswordDeserializer;
+import com.thoughtworks.go.config.materials.SubprocessExecutionContext;
 import com.thoughtworks.go.config.update.CreatePipelineConfigCommand;
+import com.thoughtworks.go.domain.materials.Material;
+import com.thoughtworks.go.domain.materials.MaterialConfig;
+import com.thoughtworks.go.domain.materials.Modification;
+import com.thoughtworks.go.plugin.access.configrepo.ConfigFileList;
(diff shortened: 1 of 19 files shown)


ibnc:
@marques-work @tomzo Would both of you mind reviewing this PR?

@tomzo I'd like to know your thoughts on the second commit `fbc3966`

ibnc:
> @ibnc I'd like to echo @tomzo's comments - great work here. Everything is straightforward to understand.
> 
> My _only_ concern is around potential performance chokes if a given material represents a very large repo. And if I've missed something, please correct me.
> 
> #### Favor shallow clones for `GitMaterial`
> I see that `com.thoughtworks.go.server.service.materials.GitPoller#checkout()` calls `toggleShallowCloneFeature(material)` but I don't think this actually sets `shallowClone = true` unless `SystemEnvironment.GO_SERVER_SHALLOW_CLONE` is set. Its default value is `false`.
> 
> We should explicitly `gitMaterial = gitMaterial.withShallowClone(true)` somewhere.

@marques-work @tomzo How does my most recent commit look as far as addressing the shallow clone concern? 


> #### Wrap the `checkout()` in a timeout and return an appropriate message
> I didn't see this in the controller but also didn't dig deep enough to see if it's buried a few levels down in any of the `checkout()` code. I only dug a little bit but didn't see anything that seemed to handle this case. Ideally, this API would inform the calling page that the `config_files` check couldn't be completed because we failed to clone the repo within XX seconds. The page could decide on how to proceed in this case (outside the scope of this PR), but we shouldn't let this check be too long-running on the server and we don't want the user to wait too long either for feedback, even if it's an error.
> 
> Could we bake these into this PR as well? Otherwise, everything looks great. And if I've just missed this functionality, please let me know.

I was thinking of letting the client handle the timeout. Even if we handle it on the server side, there are a few issues:
* There's currently no way to stop the cloning process of a material
* We might finish the checkout, but timeout before we finish scanning for config files. Making determining the appropriate error message more difficult than simply "failed to clone within XX seconds"
* We might use this endpoint on a different page in the future where we're willing to wait longer for results for one reason or another

Thoughts? @marques-work

tomzo:
@ibnc The shallow clone part looks fine to me.

I had some thoughts about handling too long clones:
 * In this PR, each clone is done to a new, clean directory, so if we force a fixed timeout, then it might be impossible to run a checkout *at all* for a slow material. For instance, if it actually requires 30s to perform a clone from scratch and we have timeout of 15s, then it would never complete, each attempt to clone from scratch would eventually fail and we've just broken the feature.
 * I think that the correct way to handle all cases would be to have a similar system to the MDU, where requests are placed on a queue and there is a concept of already running material updates. That would handle the unresponsive or just long polling cases. A subsequent query could leverage an already running clone from previous (failed) call.
 * But for this PR, I would go with a solution that @marques-work suggested, perhaps a long timeout of 1-5 minutes for checkout to finish.

ibnc:
@tomzo How does the timeout that I pushed up look?

ibnc:
@tomzo @marques-work if the timeout commit looks good, and the build passes, then this PR is ready to be merged

api/api-pipelines-as-code-internal-v1/src/main/java/com/thoughtworks/go/apiv1/pipelinesascodeinternal/PipelinesAsCodeInternalControllerV1.java
```diff
@@ -85,11 +116,61 @@ public String controllerBasePath() {
...
+    protected void checkoutFromMaterialConfig(MaterialConfig materialConfig, File folder) throws ExecutionException, InterruptedException {
+        Material material = materialConfigConverter.toMaterial(materialConfig);
+        subprocessExecutionContext.setGitShallowClone(true);
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        Future future = executor.submit(() -> {
+            List<Modification> modifications = materialService.latestModification(material, folder, subprocessExecutionContext);
+            materialService.checkout(material, folder, Modification.latestRevision(modifications), subprocessExecutionContext);
+        });
+        try {
+            future.get(2l, TimeUnit.MINUTES);
```
tomzo: To be absolutely, 100% correct, I'd suggest that you add a configuration value to SystemEnvironment and define a default of 120 seconds. So that it could be changed by administrator on a specific deployment rather than hard coded here.

PR #6707, opened 2019-08-05
Updated stage animation to use gif

bdpiparva:
## Building
building-animation

## Failing
failing-animation


GaneshSPatil:
@bdpiparva -- The stage animation felt a little faster than what currently we have.

bdpiparva:
@GaneshSPatil - Yes, Now It is consistent in all gif files. earlier GIF and SVG had a different animation speed.

Issue #6715, opened 2019-08-06
Elastic Profiles and Cluster Profiles are rendered in smaller modal


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Elastic Profiles and Cluster Profiles are rendered in a smaller modal.

![Screen Shot 2019-08-06 at 11 19 30 AM](https://user-images.githubusercontent.com/15275847/62514391-abb7aa80-b83c-11e9-93f0-ad4833d6111e.png)

This has been broken after https://github.com/gocd/gocd/commit/af1bb43a9ed6ba2f119e15dac12106db179b93ed commit.

PR #6717, opened 2019-08-06
Fix Elastic Profiles and Cluster Profiles modal sizing (#6715)

GaneshSPatil:
Fixes: #6715

Revert "Use `classnames` instead of `classnames/bind` since we are ac…tually using it like plain old `classnames` anyway."

This reverts commit af1bb43a9ed6ba2f119e15dac12106db179b93ed.

----
Classes applied to the modal when styles are bound with classnames (using: `classnames.bind`):
```
index__overlay___IyXVO index__extra-large-hack-for-EA-profiles___157i3
```

Classes applied to the modal when NO styles are bound with classnames:
```
index__overlay___IyXVO extraLargeHackForEaProfiles
```

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/counts/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/counts/index.tsx
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
-import classnames from "classnames";
+import {bind} from "classnames/bind";
 import {MithrilViewComponent} from "jsx/mithril-component";
 import * as _ from "lodash";
 import * as m from "mithril";
 import * as s from "underscore.string";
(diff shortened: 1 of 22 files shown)


GaneshSPatil:
@marques-work -- I remember explicitly using the `classnames.bind` to use the bind method feature at a couple of places, so this commit possibly has caused other issues too.

This PR reverts the https://github.com/gocd/gocd/commit/af1bb43a9ed6ba2f119e15dac12106db179b93ed commit.

If you think https://github.com/gocd/gocd/commit/af1bb43a9ed6ba2f119e15dac12106db179b93ed commit is absolutely necessary, then I can close this PR and send a fix only for the modals changes (which will fix #6715) .

PR #6735, opened 2019-08-09
Rename menu and route for the elastic agent configuration

bdpiparva:
- Renamed menu to `Elastic Agent Configurations`
- Rename page header
- Changed spa route from `/go/admin/elastic_profiles` to `/go/admin/elastic_agent_configurations`
- Added redirect from `/go/admin/elastic_profiles` to `/go/admin/elastic_agent_configurations`(in case of bookmarks)


server/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -389,6 +389,12 @@
     <to last="true">/spark/admin/plugins</to>
   </rule>
 
+  <rule>
+    <name>Elastic Agent Configurations SPA</name>
+    <from>^/admin/elastic_profiles(/?)$</from>
+    <to last="true">/spark/admin/elastic_agent_configurations</to>
```
ketan: This looks like an internal rewrite, not a redirect.

ketan: See http://tuckey.org/urlrewrite/manual/4.0/#to

`<to last="true" type="temporary-redirect" />`

bdpiparva: Sure will update this

GaneshSPatil: @ketan -- just  a question regarding `permanent-redirect` (301) vs `temporary-redirect` (302).

As the old URL is moved permanently and will be removed (in near future), should we use `permanent-redirect`?

ketan:
I'd prefer a temporary redirect, allows us to reuse the old url for
something else. A permanent redirect is cached by the browser, not sure how
long for.

Issue #6711, opened 2019-08-06
Move the agents SPA to typescript


bdpiparva:
### Part 1
- [x] Render agents spa without elastic agent. This includes action like `enable`, `disable`, `delete`, `Modifying resources` and `Modifying environment association`

### Part 2 
- [x] Add Support for agent analytics

### Part 3
- Support for elastic agents. which will be done as part of https://github.com/gocd/gocd/issues/6730

GaneshSPatil:
Done as part of GoCD release v19.10.0

PR #6747, opened 2019-08-09
Initial skeleton of Agents SPA (#6711)

SnehaPL:
- Added route for new agents SPA on /go/new-agents
	- Show agents information in tabular format


GaneshSPatil:
A new route has been added for this page (`/go/new-agents`).

Once this page is fully-functionality wise ready, we will change the route back to `/go/agents` and will get rid of the old page.

Issue #6724, opened 2019-08-06
Rename 'Elastic Profile' name in admin menu to 'Elastic Agent Configuration'


bdpiparva:
##### Summary

After discussing it with the team we have decided to rename the menu item `Elastic Profile` under the `Admin` menu to `Elastic Agent Configuration`. 

We came to this decision as the SPA is not only for the elastic agent profile. It has the cluster profile as well. The plugin will require both the configuration to get a job running using the elastic agent.

adityasood:
This was done as part of 19.8.0 release.

PR #6748, opened 2019-08-09
Renaming models directory for the elastic agent configurations

bdpiparva:
Part of: #6724
This is just a rename of the directory. No change in code.


GaneshSPatil:
@bdpiparva -- why is this required?
As we're going to rewrite the entire page anyway, these changes will be removed?

bdpiparva:
>@bdpiparva -- why is this required?
As we're going to rewrite the entire page anyway, these changes will be removed?

I renamed the route and as a result, I had to rename a couple of view directories to follow the load the script files based on the controller name. This is the only one that was left behind. This is not a mandatory change but thought of keeping it consistent with the controller name. For now we can close the PR.

PR #6751, opened 2019-08-12
Upgrade babel to 7.5.5

ketan:
- Add some consistency in code to confirm with babel requirements:
  - don't intermix `require/import` and `export`, if using `export`,
    then you must use `import`.
  - changed all `module.export = foo` => `export [class|function|...] foo`
  - Change all `require` to `import`, except when requiring images, and
    non TS code from within TS. This is used for importing angular
    plugin, form_helper and friends.
  - No more `export default` in any module.


ketan:
For context: https://github.com/gocd/gocd/pull/6751/files

PR #6756, opened 2019-08-13
Moved DES to AES migration outside of GoCipher

maheshp:
* Removing support for DES to AES as part of post construct
* Moved DES to AES migration outside of GoCipher

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
@@ -268,6 +268,10 @@ public class ConfigurationProperty implements Serializable, Validatable {
         }
     }
 
+    public void setEncryptedValue(String encryptedValue) {
+        setEncryptedValue(new EncryptedConfigurationValue(encryptedValue));
+    }
+
     public void validateKeyUniqueness(HashMap<String, ConfigurationProperty> map, String entity) {
         String name = configurationKey.getName().toLowerCase();
         ConfigurationProperty propertyWithSameKey = map.get(name);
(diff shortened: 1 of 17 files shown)


config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
```diff
@@ -266,14 +266,6 @@ public void initialize() {
         if (configurationValue == null && encryptedValue == null) {
             configurationValue = new ConfigurationValue("");
         }
-
-        if (encryptedValue != null && isNotBlank(encryptedValue.getValue())) {
-            setEncryptedValue(cipher.maybeReEncryptForPostConstructWithoutExceptions(encryptedValue.getValue()));
-        }
-    }
-
-    private void setEncryptedValue(String encryptedValue) {
```
ketan: Leave this be. Might be used from within rails.

ketan: Make it a public method, perhaps.

config/config-api/src/test/java/com/thoughtworks/go/security/DesToAesMigratorTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.spy;
+
+public class DesToAesMigratorTest {
```
ketan: Use junit5?

maheshp: Done

config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
maheshp: Got it back.

Issue #6752, opened 2019-08-12
Upgrade mithril from 1.1.6 to 2.0.3


GaneshSPatil:
##### Issue Type

- Library Upgrade

GaneshSPatil:
Closed due to automation.
Reopening...

PR #6759, opened 2019-08-14
Upgrade Mithril to v2.0.3

GaneshSPatil:
Upgrade mithril. Fixes: #6752


server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
```diff
@@ -319,7 +319,7 @@ export class SparkRoutes {
   }
 
   static apiPluginInfoPath(query: PluginInfoQuery) {
-    const queryString = m.buildQueryString(query);
+    const queryString = m.buildQueryString(query as any);
```
ketan:
```
import Mithril from "mithril";
...
query as Mithril.Params
```

is more appropriate.

server/webapp/WEB-INF/rails/webpack/models/auth_configs/spec/auth_configs_crud_spec.ts
```diff
@@ -75,7 +75,7 @@ describe("AuthorizationConfigurationCRUD", () => {
     expect(request.method).toEqual("DELETE");
     expect(request.data()).toEqual(toJSON({} as AuthConfigJSON));
     expect(request.requestHeaders.Accept).toEqual("application/vnd.go.cd.v1+json");
-    expect(request.requestHeaders["Content-Type"]).toEqual("application/json; charset=utf-8");
+    expect(request.requestHeaders["Content-Type"]).toEqual(undefined!);
```
ketan: why this change?

GaneshSPatil: As per the [mithril's changelog](https://github.com/MithrilJS/mithril.js/releases) it is stated that the request header `https://github.com/MithrilJS/mithril.js/releases` is set for all XHR methods by default, **provided they have a body `that's != null`**.

In case of delete, we don't send any request body, hence we are not required to set this header.
This assertion is just to verify, we are not setting this header (outside of mithril) for delete requests.

PR #6764, opened 2019-08-16
Removing default title from tooltip icon 

kierarad:
The title can cover tooltip and isn't helpful.
image


akshaydewan:
Late to the party, but can we ensure that these tooltips are accessible to screen-readers too? Might just be a matter of adding an `aria-` property

kierarad:
@akshaydewan sure that makes sense. can you take a look at #6791

Issue #6726, opened 2019-08-06
Issues with new pipeline creation wizard


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Placeholder for all issues related to new pipeline creation wizard

- [x] Task settings should have option to specify working directory - On pipeline creation wizard material has option to specify `Alternative checkout directory`. However task settings do not have that option. Which will result in task running from the parent directory of the material checkout folder

- [x] Mercurial material do not have field for Branch. As part of 19.4.0 release we made some changes to URL Normalization for Git and Hg materials - https://github.com/gocd/gocd/issues/6132 which introduced `branch` field for Mercurial material

- [x] While creating new pipeline with the new pipeline creation wizard cannot use template which uses parameters. More details [here](https://github.com/gocd/gocd/issues/6726#issuecomment-518634101)

- [x] New Pipeline has the pause message as  (No pause cause was specified when pipeline was created via API). This is misleading as the pipeline was created via UI.

rajiesh:
Unable to associate template with parameters usage to the newly created pipeline

- Create pipeline `foo`
- Add a parameter `test` to the pipeline `foo`
- Create an environment variable at stage/job level by name say `test_env` and set the value of it to refer to the param like - `#{test}`
- Extract a template from pipeline `foo` and call it as `foo_temp`
- Now on pipeline creation wizard start a new pipeline creation with name `bar` and in advanced settings of `Part 2` choose `use template` option
- Select template `foo_temp`
- Click on `Save` (one of the 2 save options on the page)

**Result**:

Pipeline save fails with below error - User dont have option to create param on this wizard due to which pipeline cannot be associated with the template from this wizard

<img width="1351" alt="Screen Shot 2019-08-06 at 5 10 36 PM" src="https://user-images.githubusercontent.com/6533924/62537123-5647c180-b86d-11e9-87e5-96ad5bddaf7b.png">

**Workaround**

User create pipeline without template form the wizard and them edit pipeline to add parameter and associate with template

arvindsv:
@ibnc, @kierarad and @loudaTW: Since @marques-work is away, I've added you to this. Can you please take a look and work with @rajiesh on this?

ibnc:
> Unable to associate template with parameters usage to the newly created pipeline
> 
> * Create pipeline `foo`
> * Add a parameter `test` to the pipeline `foo`
> * Create an environment variable at stage/job level by name say `test_env` and set the value of it to refer to the param like - `#{test}`
[...]

PR #6765, opened 2019-08-16
Add support for defining pipeline parameter in pipeline creation flow

kierarad:
I'm not sure this is how we want the experience to be. But here is a potential fix for #6726

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/parameters.ts
+++ b/server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/parameters.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {JsonUtils} from "helpers/json_utils";
+import _ from "lodash";
+import Stream from "mithril/stream";
+import {ValidatableMixin} from "models/mixins/new_validatable_mixin";
+
+export class PipelineParameters extends ValidatableMixin {
+  name: Stream<string> = Stream();
+  value: Stream<string> = Stream();
+
+  constructor(name: string, value: string) {
+    super();
+    this.name(name);
+    this.value(value);
+
+    this.validatePresenceOf("name");
+    this.validateIdFormat("name");
+  }
+
+  isEmpty() {
+    return "" === this.name() && "" === this.value();
+  }
+
+  toApiPayload() {
+    return JsonUtils.toSnakeCasedObject(this);
+  }
+}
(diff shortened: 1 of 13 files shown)


marques-work:
rebasing on upstream after merging another PR

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/parameters.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {JsonUtils} from "helpers/json_utils";
+import _ from "lodash";
+import Stream from "mithril/stream";
```
marques-work: This is redundant, should probably just use 1.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/parameters_editor.tsx
```diff
@@ -0,0 +1,83 @@
...
+                ];
+              })
+            } />
+          </div>
+        </Form>
+      </FormBody>
+    );
+  }
+
+  update(params: Stream<PipelineParameters[]>, param: PipelineParameters, event: Event) {
```
marques-work: Might I suggest the following?

Create an `isEmpty()` method on `PipelineParameter`:

```typescript
isEmpty() {
  return "" === this.name && "" === this.value;
}
```

And then `PipelineParametersEditor#update()` becomes:

```typescript
update(params: Stream<PipelineParameter[]>, event?: Event) {
  if (event) { event.stopPropagation(); }
  params(this.paramsList().filter((p) => !p.isEmpty());
}
```

and `PipelineParametersEditor#remove()` becomes:

```typescript
remove(params: Stream<PipelineParameters[]>, paramToDelete: PipelineParameters, event?: Event) {
  if (event) { event.stopPropagation(); }

  this.paramsList(this.paramsList().filter((p) => p !== paramToDelete);

  if (!this.paramList().length) {
    this.paramList([new PipelineParameters("", "")]);
  }

  this.update(params);
}
```

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/parameters_editor.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as Tooltip from "views/components/tooltip";
+import {TooltipSize} from "views/components/tooltip";
+import css from "./environment_variables_editor.scss";
+
+interface Attrs {
+  parameters: Stream<PipelineParameters[]>;
+}
+
+export class PipelineParametersEditor extends MithrilViewComponent<Attrs> {
+  paramList: Stream<PipelineParameters[]> = Stream([new PipelineParameters("", "")]);
```
marques-work: Here, you don't have to specify an explicit type if you don't want to. TS has enough info to infer the type, so you can simplify this to:

```typescript
paramList = Stream([new PipelineParameters("", "")]);
```

server/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
```diff
@@ -29,6 +30,7 @@ export class PipelineConfig extends ValidatableMixin {
   template: Stream<string> = Stream();
   materials: Stream<NameableSet<Material>>;
   stages: Stream<NameableSet<Stage>>;
+  parameters = Stream([] as PipelineParameters[]);
```
marques-work: We should probably have a name-uniqueness validation:

Perhaps:

```typescript
this.validateChildAttrIsUnique("parameters", "name", {message: "Parameter names must be unique"});
```

marques-work:
@kierarad LGTM; merging

Issue #6613, opened 2019-07-09
Check Material button and result components


marques-work:
See flow 2 prototype: https://thoughtworks-products.invisionapp.com/share/25SOQYDCDFW#/screens

Likely depends on #6611.

As part of the PaC add pipeline page, we need a component similar to `<TestConnection/>` that checks PaC definitions and returns a list of matching files.

ankitsri11:
Verified on GoCD Version: 19.9.0 (9929-91f828d47612be57eeb28377e7c3ca6206b8d992).

![Screenshot 2019-09-11 at 11 11 44](https://user-images.githubusercontent.com/6838313/64723273-120b9a80-d485-11e9-9404-f1068581479f.png)

PR #6767, opened 2019-08-16
Init work on check material component

ibnc:
#6613

TODO:

- [ ] Testing
- [ ] Error handling (e.g. what if one plugin doesn't support v3?)


ibnc:
@marques-work would you mind giving this a look, and giving me some feedback?

marques-work:
> @marques-work would you mind giving this a look, and giving me some feedback?

Sure, will do @ibnc

server/webapp/WEB-INF/rails/webpack/models/materials/config_file_list.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Stream} from "mithril/stream";
+import stream from "mithril/stream";
```
marques-work: If you rebase, you probably won't need two imports for `Stream`. Generally, this will do from now on (note the titlecase):

```typescript
```

server/webapp/WEB-INF/rails/webpack/views/components/materials/config_file_list_editor.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {MithrilViewComponent} from "jsx/mithril-component";
+import _ from "lodash";
+import m from "mithril";
+import {ConfigFileList} from "models/materials/config_file_list";
+
+interface Attrs {
+  configFileList: ConfigFileList;
+}
+
+export class ConfigFileListEditor extends MithrilViewComponent<Attrs> {
```
marques-work: Not sure if this should be called "editor"? I don't think we are allowing the user any input here; mostly just informational display, but correct me if I'm wrong.

server/webapp/WEB-INF/rails/webpack/views/components/materials/config_file_list_editor.tsx
```diff
@@ -0,0 +1,48 @@
...
+  view(vnode: m.Vnode<Attrs>): m.Children | void | null {
+    const list = vnode.attrs.configFileList;
+    let filesList;
+    if (list.hasErrors()) {
+      filesList = (<li>{list.errors()}</li>);
+    } else if (list.isEmpty()) {
+      return;
+    } else {
+      filesList = _.map(list.files(), (file) => {
+        return <li>
```
marques-work: You might already intend to do this as I know this is still WIP -- but don't forget to add either a class or `data-test-id` to the `<li/>` so that tests can validate the text content.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/scm_material_fields.tsx
```diff
@@ -27,7 +27,8 @@ import {
   TfsMaterialAttributes
 } from "models/materials/types";
 import {CheckboxField, PasswordField, TextField} from "views/components/forms/input_fields";
-import {TestConnection} from "views/components/materials/test_connection";
+// import {TestConnection} from "views/components/materials/test_connection";
```
marques-work: I imagine the `Attrs` will allow us to choose between the buttons in future revisions?

marques-work:
@ibnc looks good so far 👍.

After everything is working, we can reevaluate as to whether check material and test connection are better as separate components, or as a single component. I know when we talked some time ago I had suggested making them separate so as to reduce complexity at the cost of some minor duplication, but it'd be interesting to see if that bet ends up being true in the final product.

server/webapp/WEB-INF/rails/webpack/views/components/materials/config_file_list_editor.tsx
ibnc: @marques-work Well, I called it that because we eventually want to have it be a checkbox list for creating config repos. Editor might be the wrong suffix tho

server/webapp/WEB-INF/rails/webpack/views/components/config_repos/material_config_files_editor.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export class MaterialConfigFilesEditor extends MithrilViewComponent<Attrs> {
+  view(vnode: m.Vnode<Attrs>): m.Children | void | null {
+    return (
+      <div class={styles.materialConfigFiles}>
+        <span class={styles.resultsTitle}>
+          We found the following Configuration files.
+        </span>
+        <div class={styles.pluginConfigFiles}>
+          {_.map(vnode.attrs.materialConfigFiles.pluginConfigFiles(), (cfgList) => {
```
marques-work: Not a critique at all, just letting you know you could shorten even more with es6 syntax:

```jsx
_.map(vnode.attrs.materialConfigFiles.pluginConfigFiles(), (cfgList) => <ConfigFileListEditor configFileList={cfgList}/>)
```

Probably won't have any effect on the compiled JS.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/scm_material_fields.tsx
```diff
@@ -44,9 +46,13 @@ abstract class ScmFields extends MithrilViewComponent<Attrs> {
 
   view(vnode: m.Vnode<Attrs>): m.Children {
     const mattrs = vnode.attrs.material.attributes() as ScmMaterialAttributes;
+    let materialTest = <TestConnection material={vnode.attrs.material}/>;
```
marques-work: Minor nitpick, but maybe either:

- Use ternary, or
- `if`/`else`
- also, extract as function?

The minor benefit of using ternary or if/else is (I think, anyway) they are special forms where only 1 condition is evaluated. That way you are not potentially building both `<TestConnection/>` and `<MaterialCheck/>` and instead just one of them. In the big picture, this is a micro-optimization, but not a bad one IMO.

PR #6768, opened 2019-08-17
Move to a simplified webpack configuration.

ketan:
- Remove awesome-typescript-loader, instead use ts-loader with
  fork-ts-checker-webpack-plugin.
- Use thread-loader and cache-loader to improve performance of all
  loaders


ketan:
This seems to add some "UnusedFilesWebpackPlugin found some unused files" warnings in the console, not sure why they are around, and how to rid them, will sort this out later.

PR #6769, opened 2019-08-17
Support working directory syntax for add pipeline task entry

marques-work:
So far, looks like this:

image

With errors:
image

Needs:

- [x] Tests!
- [x] Error handling - where/how to display?
- [x] Client-side validation?
- [x] `help` section describing `USAGE` (maybe later make this a special `help` command)


ketan:
Does this handle CWD with whitespaces?

marques-work:
> Does this handle CWD with whitespaces?

@ketan yes, so long as you either escape spaces via `\` or quote the path a la `CWD:"my path/name with/   a   lot of whitespace"`

It uses `shellwords-ts` (port of modern Ruby's `Shellwords` module) to parse, so just apply valid shell escaping and you'll be good.

PR #6779, opened 2019-08-20
Upgrade css-loader to 3.X.X

marques-work:
- Removed `typings-for-css-modules-loader` as this package is abandoned
  - Wrote a simple loader to replace typings-for-css-modules-loader that works with modern css-loader
    - The other typings packages suggested in https://github.com/Jimdo/typings-for-css-modules-loader/issues/81 didn't work well (either involved too much change, side-effects, or didn't work at all with the current css-loader@3)
    - Our simple loader maintains the same output as the original loader


server/webapp/WEB-INF/rails/webpack/config/loaders/type-declarations-for-css-modules.js
```diff
@@ -0,0 +1,110 @@
...
+    typeDecls += `${options.banner}\n`;
+  }
+
+  const cssClassnames = [];
+
+  let match;
+
+  // content includes a huge payload which introduces extra/unwanted keys in the *.scss..d.ts
+  // so we strip them out by focusing on the locals
+  const locals = content.split("// Exports\nexports.locals = ").pop() || "";
```
ketan: I'm a bit worried about this rather overly specific regex.

Seems like [this implementation](https://github.com/seek-oss/css-modules-typescript-loader/blob/master/index.js#L70) splits on `exports.locals`, is there a reason why this is not good enough, or is there a worry that perhaps `exports.locals` might catch too much gunk that will need further filtering?

marques-work: @ketan well, I think `exports.locals` could match a css selector, especially being that the rendered CSS payload is included above that actual exports.

I could change this to be somewhat safer:

`content.split(/\nexports\.locals[\s]*=[\s]*\{/)`

That should be less brittle as it won’t rely on a comment and instead syntax that would be expected to stay the same over upgrades.

ketan: Sounds good to me. Make the change and merge it?

marques-work: @ketan will do - thanks for looking!

PR #6780, opened 2019-08-20
Introduce dropdown component

GaneshSPatil:
Screen Shot 2019-08-20 at 2 41 02 PM
---
Screen Shot 2019-08-20 at 2 41 08 PM
---
Screen Shot 2019-08-20 at 2 41 48 PM


akshaydewan:
@GaneshSPatil Do we need this? We already have `<SelectField>` in [input_fields.tsx](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx#L571)

GaneshSPatil:
@akshaydewan -- The styles of both the components are very different.

attaching a screenshot of `SelectField` component.
Screen Shot 2019-08-20 at 3 21 21 PM

Ideally, we should **not** have 2 components for the same purpose.
Sometimes (in near future), we will think about merging these 2 components, so the dropdown looks consistent on all the GoCD pages.

Issue #4967, opened 2018-07-19
Delete SCM objects over the API


dbalakirev:
##### Issue Type
- Feature enhancement

##### Summary
According to the documentation, I can create and update scm objects.
https://api.gocd.org/current/#create-a-scm-object

Would it be possible to also delete them via the API (like for config repositories)?

kritika-singh3:
Fixed in https://github.com/gocd/gocd/pull/6838 as a part of GoCD version 19.8.0

PR #6784, opened 2019-08-20
Delete pluggable scm API (#4967)

varshavaradarajan:
Request:
```
curl -i 'https://ci.example.com/go/api/admin/scms/pluggable.scm.name' \
      -u 'admin:badger' \
      -H 'Accept:application/vnd.go.cd.v1+json' \
      -X DELETE
```

Response:
```
Scm deleted successfully if valid
Associated pipelines list
```

The API is accessible to admins and group admins. Just like the other SCM APIs.


ketan:
Update the API docs? There is also a changelog in there now.

varshavaradarajan:
Thanks @ketan. Will do

maheshp:
@ketan @varshavaradarajan shouldn't we be bumping up the version of the API?

varshavaradarajan:
No. There was no delete API before. The contract has not been changed to warrant a new API version. We have done this many times before. From what I remember, template config and template auth API. There were other examples too..

maheshp:
There was a similar conversation on https://github.com/gocd/gocd/pull/6395

varshavaradarajan:
https://github.com/gocd/gocd/pull/6395#issuecomment-500889863 - new feature added or removed from the existing API, right? Not for adding a new API.

PR #6791, opened 2019-08-21
Accessibility for tooltips

kierarad:
We removed the title from the icon for tooltips as part of #6764 
This is to make them accessible by screen-readers


akshaydewan:
lgtm 👍

Issue #6737, opened 2019-08-09
[PipelineConfig] Materials UI


akshaydewan:
Related to Epic: #6362

This issue is for the UI only. API calls will be stubbed out

Allow the user to add SCM Materials, Package Repositories, Pluggable SCMs

## UX issues
1. Some users are confused between Pipeline and Package in the current dropdown
<img width="202" alt="image" src="https://user-images.githubusercontent.com/6024038/62759588-621aca00-ba9f-11e9-9c35-0dbfbe1a49bb.png">
2. Autocomplete fields in the current UI are not intuitive

## Other changes
1. Separate out the Package Repositories flow and allow users to directly select a Package Material

PR #6816, opened 2019-08-26
Materials section in new pipeline config SPA

vrushaliwaykole:
Issue: #6737 

Description:

- Handled `Add material` action
- Handled `Delete material` action
- Removed unused functions from material types.
- Test cases for views and models


server/webapp/WEB-INF/rails/webpack/views/pages/create_pipeline_page.tsx
```diff
@@ -15,18 +15,37 @@
...
     vnode.state.pipelineConfig = new PipelineConfig("", [], []);
+
+    vnode.state.materialOperations = {
+      onAdd(material: Material) {
+        vnode.state.pipelineConfig.materials().add(material);
+        m.redraw();
+      },
+      onDelete(material: Material) {
+        vnode.state.pipelineConfig.materials().delete(material);
+        m.redraw();
```
akshaydewan: I don't think we need a `m.redraw()` here

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+interface Attrs {
+  materials: Materials;
+  materialOperations: MaterialOperations;
+}
+
+export class MaterialsWidget extends MithrilViewComponent<Attrs> {
+  view(vnode: m.Vnode<Attrs>) {
+    let materials = [];
+    let expanded  = true;
+    if (_.isEmpty(vnode.attrs) || _.isEmpty(vnode.attrs.materials) || vnode.attrs.materials.length === 0) {
```
akshaydewan: Don't think we need all these checks. `vnode.attrs` and `vnode.attrs.materials`  will always be defined due to the type safety of `Attrs`

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      materialContent.push(<td>{material.name()}</td>);
+      materialContent.push(<td>{this.getTypeForDisplay(material.type())}</td>);
+      materialContent.push(<td>{material.materialUrl()}</td>);
+      materialContent.push(<td class={styles.textAlignRight}>
+        <IconGroup>
+          <Delete onclick={() => vnode.attrs.materialOperations.onDelete(material)} title={"delete-material"}
+                  data-test-id={"delete-material-button"}/>
+        </IconGroup>
+      </td>);
+      tbodyContent.push(<tr>{materialContent}</tr>);
```
akshaydewan: You could avoid all the `array.push` by defining a single element such as:
```
const materialContent =
              <tr>
                <td>{material.name()}</td>
                <td>{this.getTypeForDisplay(material.type())}</td>
                <td>{material.materialUrl()}</td>
                <td class={styles.textAlignRight}>
                  <IconGroup>
                    <Delete onclick={() => vnode.attrs.materialOperations.onDelete(material)} title={"delete-material"}
                            data-test-id={"delete-material-button"}/>
                  </IconGroup>
                </td>
              </tr>;
```
If you used a `map` instead of `forEach`, you could avoid creating the arrays altogether.

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    });
+    return <tbody>{tbodyContent}</tbody>;
+  }
+
+  getTypeForDisplay(type: string | undefined) {
```
akshaydewan: Since this is repeated in [MaterialEditor](https://github.com/gocd/gocd/pull/6816/files#diff-358d82298e09369e31e9f98320b48e21R38), I think it can move to the model itself.

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/modals.tsx
```diff
@@ -0,0 +1,60 @@
...
+    this.onSuccessfulAdd = onSuccessfulAdd;
+  }
+
+  body(): m.Children {
+    return <MaterialEditor material={this.material}/>;
+  }
+
+  title(): string {
+    const materialAttributes = this.material.attributes();
+    if (_.isUndefined(materialAttributes) || _.isEmpty(materialAttributes.name())) {
```
akshaydewan: Might not be a good idea to rely to `materialAttributes.name()` since it may be undefined even for an existing material.

akshaydewan:
Couple of things that seem odd: 
- The material name section is hidden under advanced settings. If users don't enter it, the first column of the table is empty, and it looks a bit odd.
- The placeholder text for the material name doesn't fit the textbox
Not necessary to fix immediately, we can look into these in subsequent PRs as well.

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
vrushaliwaykole: It's a `set` so used `forEach`.

vrushaliwaykole:
> * The material name section is hidden under advanced settings. If users don't enter it, the first column of the table is empty, and it looks a bit odd.

Yes. In the current pipeline config index page, we display url in `Name` column if name is not provided

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
vrushaliwaykole: I have re-used a constant `SUPPORTED_MATERIAL` which was already there in MaterialEditor.

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
vrushaliwaykole: In subsequent PR, we will be getting rid of `NameableSet` for materials. Mainly because name can be blank. `Materials` class will extend an array which will have its customized method to add material. When this change is done, I will change this code to use `map` instead of `forEach`

vrushaliwaykole:
Merging the PR as jasmine tests are passing.

server/webapp/WEB-INF/rails/webpack/views/pages/pipeline_configs/materials/index.tsx
akshaydewan: Ok. I'm not a fan of *extending* collection classes, like we do in the Java code. In ES6 and typescript, the collection classes already have some good in-built functionality and there lodash for everything else. 

In other words, a `Collection<T>` is fine, but `T extends Collection` may be overkill in JS.

PR #6824, opened 2019-08-26
New Elastic agent Configuration Flow spa

dhanasp:
Elastic agent flow rewrite.

Some open questions for which this will be pending:

- [ ] Help text blurb on all wizard pages
- [ ] Images on all wizard pages (I saw some of the images in the wireframes/mockups are docker specific, but not all plugins may be docker specific)
- On the "associate jobs page" of the wizard
  - [ ] Do we show templates as well (note that by default, this page is visible only to group admins, who do not get to see templates)
  - [ ] Do we show jobs that are backed by templates? Changing the association on such jobs will actually affect the template, and affect other jobs (without the user knowing/understanding the consequences). Also, see above about permission issues.
- How will the review page look for situations as above?


GaneshSPatil:
@ketan @dhanasp -- ping!!

Is it possible for us to get this code merged (even if it is not feature-complete)?
Looking at [this](https://github.com/gocd/gocd/pull/6824/files#diff-87f857fae0abf00c40526113842bdcb7) piece of code, I believe this page is on a new route and shouldn't cause any issues with the existing cluster profiles SPA.

akshaydewan:
Current status:
**Add flow** - uses the wizard to add a cluster profile followed by an elastic profile. The user has the option to save cluster profile and exit.

**Add elastic profile to an existing cluster** - The wizard opens in the second step to add an elastic profile. The user can go to the first step of the wizard and view the Cluster Profile. All the fields of the Cluster Profile are disabled.

**Edit Cluster Profile** - A full page modal opens to edit the cluster profile, showing the same view as the Add flow. But there is no second step in the wizard for Elastic Profile.

**Edit Elastic Profile** - The wizard opens in the second step to edit the selected elastic profile. The user can go to the first step of the wizard and view the Cluster Profile. All the fields of the Cluster Profile are disabled.

**Clone Cluster Profile** - the old modal (which is not fullscreen/wizard) opens and allows cloning.

**Clone Elastic Profile** - the old modal (which is not fullscreen/wizard) opens and allows cloning.

**Note: The previous page allowed moving an elastic profile to another cluster. The new wizard doesn't allow this.**

spark/spark-spa/src/test/groovy/com/thoughtworks/go/spark/spa/ElasticAgentsControllerTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  ElasticAgentsController createControllerInstance() {
+    return new ElasticAgentsController(new SPAAuthenticationHelper(securityService, goConfigService), templateEngine)
+  }
+
+  @Nested
+  class Index {
+    @Nested
+    class Security implements SecurityTestTrait, GroupAdminUserSecurity {
```
GaneshSPatil: @akshaydewan -- I am a bit confused with this page's authorization.
So,
- The cluster profiles can only be accessed by super admins
- The elastic agent profiles can be accessed by pipeline group admins

On page load, we make an API call to fetch cluster profiles and elastic agent profiles. As a pipeline group admin user will not have access to cluster profiles API, the API would fail with a 403 error resulting in page load failure.

I haven't tested this behavior with these changesets, but I can confirm this behavior on the existing SPA.

akshaydewan: You're right, the page won't work unless a user is a super admin. I can't think of a good answer until we implement Policies for Cluster Profiles and Elastic Profiles. Should we keep it the way it is right now until the policies are implemented?

cc: @maheshp @arvindsv @adityasood

maheshp: Currently a Pipeline Group admin gets an error on visiting the Elastic Agent Configuration page. I would prefer showing this link only to System Admins for now. 

Once we support granular auth for these configurations we can decide on the appropriate behavior, that will most likely be showing information based on permissions.

vrushaliwaykole:
Improvements: 

- The left alignment for cluster and elastic profile form is not consistent.
- For the plugins with v4 extension, for creating cluster profile it should show the error message that `can not define Cluster profiles for <pluginId> plugin as it does not support cluster profiles.`
 Currently, it shows blank for both cluster and elastic profile.
- Should not allow updating plugin ID of cluster profile.
- Select incompatible plugin id. Click on `save + next` it will give an error for the name. now when you type name, multiple error messages are displayed for name field on the cluster profile form.

Screenshot 2019-12-11 at 2 03 20 PM

server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
```diff
@@ -459,4 +459,8 @@ export class SparkRoutes {
   static cancelStageInstance(pipelineName: string, pipelineCounter: string, stageName: string, stageCounter: number) {
     return `/go/api/stages/${pipelineName}/${pipelineCounter}/${stageName}/${stageCounter}/cancel`;
   }
+
+  static apiAdminInternalPipelinesPath() {
+    return "/go/api/internal/pipeline_structure";
+  }
```
vrushaliwaykole: Can we remove this method as we are not using it anywhere?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/elastic_agents/wizard.tsx
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    if (!result) {
+      result = vnode.attrs.pluginInfos()[0];
+    }
+
+    return result;
+  }
+
+  private findPluginWithId(vnode: m.Vnode<Attrs>, pluginId?: string) {
```
vrushaliwaykole: Suggestion:

We can use `findByPluginId` method on `PluginInfos` instead of this.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/elastic_agents/wizard.tsx
```diff
@@ -0,0 +1,618 @@
...
+            <p>An Elastic Agent Profile usually contains the configuration for your agent.<br/>
+              Depending on the plugin used, this may contain the machine image (ami, docker image), size of the
+              CPU/memory/disk, network settings among other things.</p>
+          </ConceptDiagram>
+        </div>
+      </div>
+    );
+  }
+
+  private findPluginWithId(vnode: m.Vnode<Attrs>, pluginId?: string) {
```
vrushaliwaykole: Suggestion:

We can use `findByPluginId` method on `PluginInfos` instead of this.

server/src/main/webapp/WEB-INF/rails/webpack/single_page_apps/elastic_agents.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+import {RoutedSinglePageApp} from "helpers/spa_base";
+import {ElasticAgentsPage} from "views/pages/elastic_agents";
+
+export class ElasticAgentsSPA extends RoutedSinglePageApp {
+  constructor() {
+    super({"/": ElasticAgentsPage});
+  }
```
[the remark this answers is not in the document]
akshaydewan: I didn't write this part, but maybe there was an intention of having multiple routes to go to a specific elastic profile or cluster profile? If so, I can implement it.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/elastic_agents.tsx
```diff
@@ -0,0 +1,361 @@
...
+        this.flashMessage.clear();
+
+        new NewClusterProfileModal(vnode.state.pluginInfos(), vnode.state.onSuccessfulSave).render();
+      },
+
+      onClone: (clusterProfile: ClusterProfile, event: MouseEvent) => {
+        event.stopPropagation();
+        this.flashMessage.clear();
+
+        new CloneClusterProfileModal(clusterProfile.id()!,
```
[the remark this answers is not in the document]
akshaydewan: > The previous page allowed moving an elastic profile to another cluster. The new wizard doesn't allow this. But an alternate way would be to clone to elastic profile into a different cluster and delete the original.

Based on that, the elastic profile cloning would use the old modal, but the cluster profile would use the wizard. IMO, it's just more consistent to use the modals for both clone operations.

server/src/main/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
```diff
@@ -237,7 +237,8 @@ export class HelpText extends RestyleViewComponent<Styles, HelpTextComponentAttr
 
   view(vnode: m.Vnode<HelpTextComponentAttrs>) {
     if (!_.isEmpty(vnode.attrs.helpText)) {
-      return (<span id={vnode.attrs.helpTextId} class={classnames(this.css.formHelp)}>
+      return (<span data-test-id={`form-field-help-`}
```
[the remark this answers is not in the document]
akshaydewan: Removed the data-test-id altogether since it wasn't used anywhere

server/src/main/webapp/WEB-INF/rails/webpack/views/components/no_plugins_installed/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  extensionType: AbstractExtensionType<T>;
+}
+
+export class NoPluginsOfTypeInstalled<T extends ExtensionJSON> extends MithrilViewComponent<Attrs<T>> {
+  view(vnode: m.Vnode<Attrs<T>, this>): m.Children | void | null {
+    const message = (
+      <div>
+        To use this page, you must ensure that there are one or
+        more {vnode.attrs.extensionType.humanReadableName()} plugins installed. Please see <a
+        href={vnode.attrs.extensionType.linkForDocs()}>this page</a> for a list of supported plugins.
```
vrushaliwaykole: Can we open the doc link in a new tab? We can also use `Link` with `externalLinkIcon` as the link opens in a new tab.

Issue #6362, opened 2019-05-31
[EPIC] PipelineConfig SPA


maheshp:
##### Issue Type

- Feature enhancement

##### Summary

Introduce a PipelineConfig SPA to replace the existing rails pages to configure a Pipeline.

The SPA should solve the issues listed in the earlier attempt to [re-write](https://github.com/gocd/gocd/issues/1404) and also with a better UX.

akshaydewan:
Adding some things that were previously discussed:

(This was previously shared in a google doc)
Rethinking quick edit … again:

### Considerations:
- Show only the bare minimum and most important parts. Most options are hidden by default.
- Exec tasks are most important and should be default.
- Should show the first stage, first job by default (maybe only if it’s a single-stage, single-job pipeline?)
- Errors should show somewhere (at the top and inline).
- Needs to support too many stages (scrolls to the right and left using those arrows)
- Needs to support too many jobs (need to figure it out - box grows?).
- Needs to support too many tasks (page grows).
- Needs to support too many materials (box grows).
- No accordions, if possible.

- Haven’t figured out interaction when those edit icons / add buttons are clicked. - Modal or a separate SPA “page” from which they can press Back. I see those pages as normal scrolling settings pages, as in Chrome or Firefox. Nothing fancy with tabs and all.
- If edit (task, material, etc) is possible inline, it should. If not, similar interaction to above.
Didn’t want to use too many icons, if possible.

Try to: Use this for template edit as well, by removing the “Materials” section and making some other very small changes.
Try to: Use this for config-repo-pipeline view as well. View-only mode is needed anyway, since we also have a template view.

Link to latest image: [pipeline_config_edit.png](https://gist.githubusercontent.com/arvindsv/ce86b93c580da28bd15fe05229ba5ec3/raw/f6995230095d3f6d2e5fb5488a430e3a425fe0da/pipeline_config_edit.png)


Thoughts from Ganesh: (2019-01-22)

- We are showing minimum required information on the screen. How easy it is for a user to locate the hidden information? Is it just one click away or multiple clicks?
- Does our design change for the inline-template view? As we need to explicitly say that the stage configurations come from the template.
- Template views are read-only. How to show the other hidden information? Still on settings icons? or some other icon?
- Config repo - everything is readonly.
- How about another tab to programmatically editing pipeline? May be JSON we pick by default (as our typescript models will be mapped to json). So i can edit the json can hit save. Will show read only json/yaml for config repo plugins.
[...]

PR #6825, opened 2019-08-26
Add optional documentation link to help text

akshaydewan:
Related to #6362 

Description: A lot of fields in the pipeline config page have a link to more documentation. This adds an optional property called `docLink` to the help text.

Example:
Screenshot 2019-08-26 at 17 51 16
`More...` opens the linked gocd documentation in a new tab


GaneshSPatil:
@akshaydewan -- The "more" text initially gave me an impression that more information will be rendered inline. (similar to "show more.." and "show less..")

Can we change the wording to "Learn more"?

So:
- "more.."/"less.." - shows inline information.
- "learn more" - links to the external documentation.

Or is there a better way to reduce the confusion between inline help text vs external links?

GaneshSPatil:
\cc: @streisguth  @marques-work -- WDYT?

akshaydewan:
"Learn more..." seems ok to me.

server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
```diff
@@ -202,9 +205,16 @@ class HelpText extends MithrilViewComponent<HelpTextComponentAttrs> {
     }
   }
 
+  static maybeDocLink(docLink?: string) {
+    if (!_.isEmpty(docLink)) {
+      return <span>&nbsp;<Link href={docsUrl(docLink)} target="_blank">More&#8230;</Link></span>;
```
ketan: See https://www.owasp.org/index.php/Reverse_Tabnabbing. Perhaps make the `Link` component smart enough, based on the `href` attribute to add additional attrs?

akshaydewan: Thanks for pointing that out. About the `Link` component, do you mean checking if the link specifies `target` and then adding `rel="noopener noreferrer"` if needed?

akshaydewan:
I've made some changes 
- Added the `rel="noopener noreferrer"` when a `target` is specified
- Changed the text to "Learn More"
- Used the external link icon instead of ellipsis
Screenshot 2019-08-27 at 11 53 07

server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
ketan: What I meant is to make the `Link` component smart enough to specifically detect external links (`http(s)://`) and, based on that, automatically add the `target="_blank"` and `rel="noopener noreferrer"` attributes.

PR #6826, opened 2019-08-26
Reduce GoCD Server and Agent Docker Image size

GaneshSPatil:
* Perform chmod and chown operations on go-{agent,server} dirs as part of
  the same docker layer (same as of unzip jar layer)

### DIFF
--- a/buildSrc/src/main/resources/gocd-docker-agent/Dockerfile.agent.ftl
+++ b/buildSrc/src/main/resources/gocd-docker-agent/Dockerfile.agent.ftl
@@ -18,6 +18,9 @@
 ###############################################################################################
 
 FROM alpine:latest as gocd-agent-unzip
+
+ARG UID=1000
+
 <#if useFromArtifact >
 COPY go-agent-${fullVersion}.zip /tmp/go-agent-${fullVersion}.zip
 <#else>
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
@ketan -- with these changes, the size of the docker image has reduced (_significantly_).

The additional **chown/chmod layer** used to add the exact same size as of the go-{server,agent} files (**30.6MB** for Agent and **204MB** for Server). With these changes, the size of that layer has been reduced **to a few KBs**.

Screen Shot 2019-08-27 at 9 01 29 AM
Screen Shot 2019-08-27 at 9 06 28 AM

buildSrc/src/main/resources/gocd-docker-agent/Dockerfile.agent.ftl
```diff
@@ -90,8 +93,8 @@ COPY --chown=go:root agent-bootstrapper-logback-include.xml agent-launcher-logba
 COPY --chown=root:root dockerd-sudo /etc/sudoers.d/dockerd-sudo
 </#if>
 
-RUN chown -R go:root /go-agent /docker-entrypoint.d /go /godata /docker-entrypoint.sh \
-    && chmod -R g=u /go-agent /docker-entrypoint.d /go /godata /docker-entrypoint.sh
+RUN chown -R go:root /docker-entrypoint.d /go /godata /docker-entrypoint.sh \
```
GaneshSPatil: `/docker-entrypoint.d`, `/go` and `/godata` are empty directories **chown/chmod** on empty directories will not add additional filesize.

GaneshSPatil: Whereas, **chmod/chown** on `/docker-entrypoint.sh` file will add additional filesize (same as of the size of the `docker-entrypoint.sh`).

There is no workaround/way for this (as of today), as `ADD` command in docker does not support `--chmod`  flag (it supports `--chown`).

Issue #6832, opened 2019-08-28
User info synchronisation with external sources


kritika-singh3:
Parent Issue: https://github.com/gocd-contrib/github-oauth-authorization-plugin/issues/69

##### Issue Type
 - Feature enhancement

##### Summary

Any GoCD user information is not updated with the actual source after the first login.
E.g. if the email associated with the github profile is updated, the same is never reflected in the GoCD until that user is deleted and added again.

##### Suggestion

After successful authentication, the user info received should be compared with the info stored in the DB. In case of any update, the DB should be updated to reflect the same.

kritika-singh3:
Closed via automation. Reopening for testing.

rajiesh:
Verified on 19.8.0 (9912-33898b3adaf72cedea1604882d3554fd586571a8)

PR #6833, opened 2019-08-28
Updated user information stored in DB

kritika-singh3:
Issue: #6832
Description: The user information stored in GoCD DB should get updated on successful authentication.
 - On Successful Authentication of any user in GoCD, we originally use to add the user if it does not exist.
   Updated this flow and added additional check for change in either display name or email. If either of them have changed, we will update the DB entry to reflect the same

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/UserService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/UserService.java
@@ -411,31 +411,19 @@ public class UserService {
         synchronized (enableUserMutex) {
             if (!user.isAnonymous()) {
                 LOGGER.debug("User [{}] is not anonymous", user.getName());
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/thoughtworks/go/server/service/UserService.java
```diff
@@ -573,4 +586,15 @@ private boolean validate(LocalizedOperationResult result, User user) {
         return false;
     }
 
+    private boolean hasUserChanged(User newUser, User originalUser) {
```
ketan: Is an equals check not good enough?

kritika-singh3: We only wanted to update the user in case `displayName` or `email` changed not `username`. Hence, the explicit checks.

Issue #6827, opened 2019-08-26
Upgrade from 19.7.0 to 19.8.0 overwrites wrapper-properties.conf file


rajiesh:
##### Issue Type

- Bug Report

##### Summary

Upgrade from 19.7.0 release to 19.8.0 experimental build overwrites `wrapper-properties.conf` file. Verified on Centos-7 docker container

##### Steps to Reproduce

1. Install 19.7.0-9567 stable release version
2. Update `wrapper-properties.conf` file to add an additional property like `wrapper.java.additional.100=-Dcruise.server.port=8155`
3. Start the server and access the application on 8155 port
4. Now upgrade the server to 19.8.0-9815 experimental build
5. `wrapper-properties.conf` file changes are missing

Same behavior for agent installation too

rajiesh:
Verified on 19.8.0 (9912-33898b3adaf72cedea1604882d3554fd586571a8)

PR #6837, opened 2019-08-28
Fixes for linux package installers

viraj2712:
Issue: #6827 

Description:

1. `fpm` requires option `--config-files` to avoid overriding config files while upgrading. This option was not getting applied for `wrapper.conf` and `wrapper-properties.conf`
The issue was maily because of the `permissions.confFile` check. So, added `confFile` flag on both `wrapper.conf` and `wrapper-properties.conf`

2. While migrating from version `< 19.6.0`, temporary migration file is created (eg. To migrate properties from `/etc/default/go-agent` to `/usr/share/go-agent/wrapper-config/wrapper-properties.conf`)
     These files were not getting deleted after the upgrade. Added code to delete these file
3. Take backup of configuration file for GoCD versions `< 19.6.0`(eg. `/etc/default/go-agent` to `/etc/default/go-agent.pre-migration`)
     PS: This is already in place for agent installers, but was missing for server installers

### DIFF
--- a/installers/linux/shared/partials/_go-server-config-migration.sh.erb
+++ b/installers/linux/shared/partials/_go-server-config-migration.sh.erb
@@ -22,6 +22,8 @@
 
             . "/etc/default/<%= name %>"
 
+            mv "/etc/default/<%= name %>" "/etc/default/<%= name %>.pre-migration"
+
             # these variables are not to be exported (even if users explicitly `export`ed them), so we filter them out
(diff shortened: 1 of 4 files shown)


installers/linux/shared/partials/_go-server-config-migration.sh.erb
```diff
@@ -22,6 +22,8 @@
 
             . "/etc/default/<%= name %>"
 
+            mv "/etc/default/<%= name %>" "/etc/default/<%= name %>.pre-migration"
```
ketan: For anyone wondering, we had a similar line in agent migration, but we missed it in the server migration.

This introduces another bug where a server upgrade from 19.6->19.7->19.8 might keep doing the same migration again and again.

Issue #6738, opened 2019-08-09
[PipelineConfig] UI for Pipeline Level Settings


akshaydewan:
Related to Epic: #6362

This issue is for the UI only. API calls will be stubbed out.

## UX issues
1. Validation for pipeline name happens too late
2. Provide better (help) text for fields:
    - Automatic Pipeline scheduling
    - Cron timer specification
    - Pipeline locking
3. Show a consolidated view for all environment variables defined everywhere in the Environment, Pipeline, Stage, Job

## Notes
After a pipeline is saved, the pipeline name won't be editable. Making the pipeline name editable is desirable, but will be out of the scope of this story.

After a pipeline is saved, one option to show the UI could be like:
<img width="697" alt="image" src="https://user-images.githubusercontent.com/6024038/62760981-ffc3c880-baa2-11e9-877e-674560daa0a3.png">

akshaydewan:
Current WIP screenshots:
<img width="1056" alt="Screenshot 2019-08-29 at 14 56 03" src="https://user-images.githubusercontent.com/6024038/63928497-43d83680-ca6d-11e9-8daa-a637fa7ca6f2.png">
<img width="1060" alt="Screenshot 2019-08-29 at 14 56 17" src="https://user-images.githubusercontent.com/6024038/63928508-48045400-ca6d-11e9-8e4d-d0a04b3d4639.png">

- The helpText is displayed below the input field instead of the (?) button that showed a tooltip in the old page
- The labels still need to be made user-friendly
- The styles are the default ones applied to input components, but some of them look inconsistent, such as the space between the checkbox label and its helpText. This needs to be fixed, but will likely affect all pages using the input fields.

akshaydewan:
The Env Variables and Parameters sections are similar and straightforward. Here's the Project Management section:

<img width="1056" alt="Screenshot 2019-09-03 at 15 23 45" src="https://user-images.githubusercontent.com/6024038/64163485-fdb11780-ce5e-11e9-8f57-1d854b241878.png">
<img width="1062" alt="Screenshot 2019-09-03 at 15 24 24" src="https://user-images.githubusercontent.com/6024038/64163500-030e6200-ce5f-11e9-813a-99a0aeab9a6e.png">

For reference, the current screen looks like this - 
[...]

PR #6841, opened 2019-08-28
Radiobutton component

akshaydewan:
Related to: #6738

Radio buttons component, copied from a previous branch.

Changes:
- added helpText to the model so that each radio button can show a help text or description
- Some CSS improvements
- Removed the `boolean` type from the radio button value. I don't see any use case for a true/false value. A checkbox or switch would be more appropriate in those cases.

Screenshot 2019-08-28 at 16 42 05

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
@@ -572,13 +572,19 @@ export class TriStateCheckboxField extends FormField<TriStateCheckbox> {
   }
 }
 
+export interface RadioData {
+  label: string;
+  value: string;
+  helpText?: string;
+}
+
 export interface RadioButtonAttrs {
   label: string;
   errorText?: string;
   disabled?: boolean;
   required?: boolean;
-  property: (newValue?: string | boolean) => string | boolean;
-  possibleValues: Map<string, string | boolean>;
+  property: (newValue?: string) => string;
+  possibleValues: RadioData[];
 }
 
 export class RadioField extends MithrilViewComponent<RadioButtonAttrs> {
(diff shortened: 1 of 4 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
```diff
@@ -572,6 +572,66 @@ export class TriStateCheckboxField extends FormField<TriStateCheckbox> {
   }
 }
 
+export interface RadioData {
```
GaneshSPatil: Do we need to export this?

server/webapp/WEB-INF/rails/webpack/views/pages/kitchen_sink.tsx
```diff
@@ -385,6 +387,15 @@ export class KitchenSink extends MithrilViewComponent<null> {
                      placeholder="password"
                      property={encryptedPasswordValue}
                      helpText={"Lorem ipsum dolor sit amet, consectetur adipiscing."}/>,
+      <RadioField label="Choose your favorite pizza topping"
+                  property={radioValue}
+                  required={true}
```
GaneshSPatil: Is it possible to have a radio button with no default selected (no value selected)?
Should we support such behavior?

server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
akshaydewan: Hmm, apparently we don't need to export it. I thought it was needed since we're passing this object from the page where the radio buttons will be used.

server/webapp/WEB-INF/rails/webpack/views/pages/kitchen_sink.tsx
akshaydewan: Technically, it's possible, since a form can be submitted without any selection. But from a usability point of view, it's unlikely that nothing would be selected. (since it's not trivial to un-select a radio-button). In some cases, we might want to start off with nothing selected, if we want the user to explicitly make a choice with no "default" given to them.

However, your comment made me realize that I haven't added a "checked" or "selected" property to select one of the options by default. I will add that to the `RadioData`

server/webapp/WEB-INF/rails/webpack/views/components/forms/input_fields.tsx
ketan: That interface is purely for type-checking, not exporting will still allow typechecking, auto-complete hints etc.

Only reason to export might be if you want to use that "named interface" somewhere, (`foo extends ...`, define a variable )

server/webapp/WEB-INF/rails/webpack/views/pages/kitchen_sink.tsx
akshaydewan: Bound the `checked` property to the data model instead of a hardcoded boolean

Issue #6762, opened 2019-08-15
Indicators on main navigation bar


streisguth:
Wayfinding indicators on the main navigation bar. ie when a user is on the Dashboard, there is an corresponding indication on the navigation.

![menu-nav-indicator-admin](https://user-images.githubusercontent.com/1781525/63119238-96d1ca00-bf54-11e9-8040-58eb805f93b2.png)

![menu-nav-indicator-dashboard](https://user-images.githubusercontent.com/1781525/63118785-705f5f00-bf53-11e9-80f9-14de767da32a.png)

ankitsri11:
Verified on GoCD Version: 19.9.0 (9929-91f828d47612be57eeb28377e7c3ca6206b8d992

PR #6845, opened 2019-08-28
Show indicators for site nav bar

ibnc:
Issue: #6762 

Screenshot 1: 
Screen Shot 2019-08-28 at 13 25 29

Screenshot 2:
Screen Shot 2019-08-28 at 13 25 41

### DIFF
--- a/server/webapp/WEB-INF/rails/webpack/views/components/site_menu/index.tsx
+++ b/server/webapp/WEB-INF/rails/webpack/views/components/site_menu/index.tsx
@@ -20,6 +20,10 @@ import styles from "./index.scss";
 
 const classnames = bind(styles);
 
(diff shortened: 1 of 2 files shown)


server/webapp/WEB-INF/rails/webpack/views/components/site_menu/index.tsx
```diff
@@ -20,6 +20,10 @@ import styles from "./index.scss";
 
 const classnames = bind(styles);
 
+const activeItem = (href?: string): boolean => {
+  return href === window.location.pathname;
```
marques-work: Perhaps more robust would be a `startsWith()`, or even better, comparing `new URL(href).pathname === window.location.pathname`

Issue #6771, opened 2019-08-19
Issues after upgrading to mithril 2.0.3


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Issues occurred after upgrading to mithril 2.0.3 (#6752)

- [x] A generic error message is shown on users SPA in case of operation failure. [Gauge Spec](https://build.gocd.org/go/files/regression-SPAs-PR/953/Firefox/1/regression-runInstance-6/GaugeReports/reports/html-report/specs/EnableDisableUsers.html)(Fixed as part of #6777)

- [x] Remove webpack patches dir inclusion. Remove [this](https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/webpack/config/variables.ts#L37) line. (Fixed as part of #6777)

- [x] Plugin Icon is not rendered for the Elastic agents on the Agents SPA. (Fixed as part of #6792)

- [x] AgentSPA: agent analytics was showing error on click of analytics icon(Fixed as part of  #6813)

- [x] A generic error message is shown on the pipeline config SPA in case of operation failure.  (Fixed as part of #6852)

rajiesh:
Verified on 19.8.0 (9912-33898b3adaf72cedea1604882d3554fd586571a8)

PR #6852, opened 2019-08-29
Show server error on Pipeline Config SPA (#6771)

GaneshSPatil:
Issue: #6771

Description:
Show appropriate error message received from server instead of showing a generic error


GaneshSPatil:
##### Before:
Screen Shot 2019-08-29 at 2 13 20 PM

##### After:
Screen Shot 2019-08-29 at 2 07 35 PM

PR #6865, opened 2019-08-30
Adding current config schema version to version.json

kritika-singh3:
Adding current config schema version to version.json 
- which is an artifact along with our installers
 - would help run the functional specs without explicit changes required.

### DIFF
--- a/installers/build.gradle
+++ b/installers/build.gradle
@@ -1,4 +1,5 @@
 import groovy.json.JsonOutput
+
 /*
  * Copyright 2016 ThoughtWorks, Inc.
  *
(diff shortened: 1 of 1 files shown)


ketan:
I'd prefer to put this value in the API support page instead. Maybe [this place](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/support/ServerBasicInfoProvider.java)?

Reason being that this change (to version.json) likely ties the functional tests to presence of version.json (or a specific upstream pipeline structure), without this information, it'll be hard(er) to run functional tests locally.

kritika-singh3:
Closing WRT @ketan's comment. Will raise another PR for the change mentioned.

PR #6883, opened 2019-09-05
Option to have a fixed height for modal

akshaydewan:
Issue:  #6738

Description:
This is so that the modal height doesn't change when switching tabs


GaneshSPatil:
@akshaydewan -- The current overlay height is set to `500px`.
If a modal requires a different fixed height for `500px` should it be allowed?

akshaydewan:
The code we have in place currently sets the max-height to 500px
```
  @media(min-width: $screen-md) {
    max-height: $overlay-height;
  }
```
So I've just set the height to the same value. Maybe we can change it in the future if we need to?

Issue #6740, opened 2019-08-09
[PipelineConfig] UI for Stage Level Settings


akshaydewan:
Related to Epic: #6362

This issue is for the UI only. API calls will be stubbed out.

Allow the user to configure Stage Level Settings

## UX issues
1. Improve the (help) text of the some fields:
    - Fetch Materials (people are very confused with this)
    - onSuccess & Manual and its relation to Automatic Pipeline Scheduling on the pipeline level

akshaydewan:
Some feedback on "Never cleanup artifacts" in #2506

Issue #2506, opened 2016-07-31
UI text tweak - change "Never Cleanup Artifacts" to something less confusing


sriramnrn:
##### Issue Type
- Bug Report
##### Summary

The text "Never Cleanup Artifacts" is a bit confusing when read alongside "Fetch Materials" and "Clean Working Directory". This text may give the impression that after artifacts are uploaded, this checkbox may permit them to remain in the workspace.

Some alternatives:
"Preserve Uploaded Artifacts during server-side auto-cleanup"
"Keep Uploaded Artifacts during server-side auto-cleanup"
"Preserve Uploaded Artifacts during server-side auto-cleanup"

I considered the following, and felt they'd be confusing as well:
"Never Cleanup Uploaded Artifacts"
"Preserve Uploaded Artifacts".

As I discovered during a recent GoCD training, there are people who take some time to understand the difference between artifacts kept locally vs the server side.

I considered sending a pull request, but I see that there are various locations in code where "neverCleanupArtifacts" is used, so I wanted to discuss this first.

arvindsv:
I agree. Also, "Clean working directory" - many people seem to think that GoCD won't clean up before every build starts unless this is checked. But, as you all know, this empties the working directory and re-clones every material before starting off a job run.

/cc @maheshp @naveenbhaskar

GaneshSPatil:
@akshaydewan @viraj2712 @vrushaliwaykole -- can we consider this as part of the pipeline config SPA?

akshaydewan:
Yes, thank you! I've referenced this issue in the relevant story for Stage level settings

PR #6885, opened 2019-09-05
Add help text for stage settings fields (#6740)

GaneshSPatil:
Issue: #6740

Description:
Screen Shot 2019-09-05 at 4 51 30 PM


akshaydewan:
Two points from #2506 that we can consider - 
- We can re-word "Never cleanup artifacts" to  "Preserve Uploaded Artifacts during server-side auto-cleanup". We can also add a [learn more](https://docs.gocd.org/current/configuration/delete_artifacts.html) link.

- For "Clean working directory", we need to convey that the whole working-directory will deleted be re-cloned, but it is always reset even if this un-checked. IMO "Clean" is an ambiguous word, but we also need to make it SCM-agnostic. I think we can change it to -

- [ ] Delete working copy on each run
When selected, it will force the agent to checkout a fresh copy from the SCM.

But I still don't know how to convey that the working copy is always "reset", even if this is unchecked.

arvindsv:
Yeah, I don't know a good way to convey that it is reset either. :( I like your other suggestions.

GaneshSPatil:
Closing this PR as the new pipeline config SPA has been deprioritized.

PR #6887, opened 2019-09-05
Feature toggle the Properties API (and tab)

ketan:
- Disables the "Properties" tab on the job detail page
- Disable all "GET" API calls for properties.

POST is still supported, for now, but will be removed in a couple of
releases. Additionally, the `<property/>` tag is rarely used. This is
used to extract metrics from XML reports (via XPATH) and publish them
using the API. Support for `<property/>` tag will be removed in a couple
of releases as well.

The properties API provides the ability for GoCD to store arbitrary
key-value-pair data associated with each build. The intent of this API
(at the time of developing this) was to be able to store arbitrary
metrics (like test count, coverage%, warning count, et.al.) and be
able to track progress of these metrics over time.

Fast forward to 2019, there are (much) better tools available that can
do just this.


ketan:
Rebasing and merging. The current [VSM is here](https://build.gocd.org/go/pipelines/value_stream_map/trigger/5055)

PR #6299, opened 2019-05-17
Introduce ConsoleLogRequestProcessor

arvindsv:
Mainly for use by plugins to write information to the job console log. This is helpful to understand what's going on when an elastic agent is being brought up by a plugin.

### Todo:

- [ ] Update functional tests? Will check to see if any of them test messages.
- [ ] Update [next release notes branch](https://github.com/gocd/www.go.cd/tree/release-notes-19.4).
- [ ] Update plugin-api documentation.

### Testing / trying this out:

1. You'll need code from this PR with something like this in your config:

<details>
<summary>Part of GoCD config</summary>

```
  <elastic>
    <agentProfiles>
      <agentProfile id="mydocker" clusterProfileId="localdocker">
        <property>
          <key>Hosts</key>
          <value />
        </property>
        <property>
          <key>Mounts</key>
          <value />
        </property>
        <property>
          <key>Command</key>
          <value />
        </property>
        <property>
          <key>Privileged</key>
          <value />
        </property>
        <property>
          <key>ReservedMemory</key>
          <value />
        </property>
        <property>
          <key>Environment</key>
          <value />
        </property>
        <property>
          <key>Cpus</key>
          <value />
        </property>
        <property>
          <key>Image</key>
          <value>gocd/gocd-agent-alpine-3.9:v19.3.0</value>
        </property>
        <property>
          <key>MaxMemory</key>
          <value />
        </property>
      </agentProfile>
    </agentProfiles>
    <clusterProfiles>
      <clusterProfile id="localdocker" pluginId="cd.go.contrib.elastic-agent.docker">
        <property>
          <key>go_server_url</key>
          <value>https://192.168.0.100:8154/go</value>
        </property>
        <property>
          <key>environment_variables</key>
          <value />
        </property>
        <property>
          <key>max_docker_containers</key>
          <value>1</value>
        </property>
        <property>
          <key>docker_uri</key>
          <value>unix:///var/run/docker.sock</value>
        </property>
        <property>
          <key>auto_register_timeout</key>
          <value>3</value>
        </property>
        <property>
          <key>docker_ca_cert</key>
          <value />
        </property>
        <property>
          <key>docker_client_cert</key>
          <value />
        </property>
        <property>
          <key>docker_client_key</key>
          <value />
        </property>
        <property>
          <key>private_registry_server</key>
          <value />
        </property>
        <property>
          <key>private_registry_username</key>
          <value />
        </property>
        <property>
          <key>private_registry_password</key>
          <encryptedValue />
[...]

PR #6899, opened 2019-09-06
#6299 - Introduce v2 of ConsoleLogRequestProcessor

arvindsv:
The only change from v1 is that it does not use camel case in its message.

Issue: #6299 + https://github.com/gocd/kubernetes-elastic-agents/pull/136#pullrequestreview-284638323

Description: v1 of ConsoleLogRequestProcessor used a message which looked like this:

```
{
  "pipelineName": "p1",
  "jobName": "j1",
  "pipelineCounter": "1",
  "stageName": "s1",
  "text": "message1",
  "stageCounter": "2"
}
```

v2 uses a message which looks like this:

```
{
  "pipeline-name": "p1",
  "stage-name": "s1",
  "pipeline-counter": "1",
  "job-name": "j1",
  "stage-counter": "2",
  "text": "message1"
}
```

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/console/ConsoleLogRequestProcessor.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/console/ConsoleLogRequestProcessor.java
@@ -24,6 +24,7 @@ import com.thoughtworks.go.plugin.infra.PluginRequestProcessorRegistry;
 import com.thoughtworks.go.plugin.infra.plugininfo.GoPluginDescriptor;
 import com.thoughtworks.go.server.service.ConsoleService;
 import com.thoughtworks.go.server.service.plugins.processor.console.v1.MessageHandlerForConsoleLogRequestProcessorImpl1_0;
+import com.thoughtworks.go.server.service.plugins.processor.console.v2.MessageHandlerForConsoleLogRequestProcessorImpl2_0;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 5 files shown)


GaneshSPatil:
@arvindsv @ketan -- should we use `hyphens (-)`? The existing requests uses `undersores (_)`.

See: [disable](https://plugin-api.gocd.org/current/elastic-agents/#disable-agents) or [delete](https://plugin-api.gocd.org/current/elastic-agents/#delete-agents) agents' [request body](https://plugin-api.gocd.org/current/elastic-agents/#the-elastic-agent-object).

ketan:
This is weird. But the request names use hyphens, json keys in request/response bodies uses snake case underscored keys.

arvindsv:
@GaneshSPatil @ketan Ok. I'll make them underscores.

arvindsv:
Done.

GaneshSPatil:
@arvindsv -- Anything else pending on the PR?
the changes LGTM.
should we get this merge for the current release?

\cc: @ketan

arvindsv:
@GaneshSPatil Nothing else is pending as far as I know. The change is quite harmless as you can see and I think it can be merged, but it's your call.

ketan:
How does a plugin know what version of the processor is available?

GaneshSPatil:
@ketan -- from `plugin-api.gocd.org`?

If we look at the [Requests to the GoCD server
](https://plugin-api.gocd.org/current/elastic-agents/#requests-to-the-gocd-server) section, each request mentions which request version should be used.

Screen Shot 2019-09-19 at 5 12 05 PM

GaneshSPatil:
@arvindsv -- merging this PR...

arvindsv:
@GaneshSPatil The server knows the version of the plugin. But, the plugin never knows what the server supports. But, that's the case with _all_ of our endpoints. We don't send that information to the plugins, when they're initialized. So, they'll have to make that call and find out. Also, capabilities is only for the plugin to tell the server what it supports.

So, a new plugin with an old server, will break. I think that's what @ketan was hinting at.

PR #6900, opened 2019-09-07
Polishing PaC form components

marques-work:
- [x] Config repo ID should be called "Name"
- [x] `configFiles` endpoint should be able to check just the specified plugin
- [x] Clean up styles for check material results (omit plugin id?); check fonts, layout, etc here.
- [x] Update copy for Download instructions
- [x] Update copy for Check Material instructions
- [x] Update copy for Check Material timeout error
- [x] Update copy for `New Pipeline` dropdown options
- [x] Preview content should try to remain visible as the user scrolls down


marques-work:
### Here's what it looks like now:

image

server/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
```diff
@@ -164,8 +164,8 @@ export class SparkRoutes {
     return `/go/api/internal/config_repos/${id}/trigger_update`;
   }
 
-  static pacListConfigFiles() {
-    return `/go/api/admin/internal/pac/config_files`;
+  static pacListConfigFiles(pluginId?: string) {
+    return pluginId ? `/go/api/admin/internal/pac/config_files?${m.buildQueryString({ pluginId })}` : `/go/api/admin/internal/pac/config_files`;
```
ibnc: This is a POST endpoint so passing the param like this seems off

ibnc: Beyond that, everything looks good to me

marques-work: @ibnc `POST` requests can also accept query params; AFAIK that _shouldn't_ break any conventions?

I could make this part of the path if you think that's better (this would make it required rather than optional, meaning you will only see a single plugin's results) or even the request body (though for the body it's probably more work).

marques-work: @ibnc and maybe it's better as part of the path because, at the moment, we don't need to query more than 1 plugin.

marques-work:
@ibnc thanks for reviewing!

marques-work:
Fixes #6910

PR #6929, opened 2019-09-13
Removing support for java version 9 and 10

bdpiparva:
- Removed support for java version 9 and 10
- Run builds with java version 11 and 12
- Keeping bootstrap source compatibility to Java 8. This is required as bootstrap should start with java version < 11 to print unsupported java version on console.
- Removed javasysmon dep

### Note

Adding support for java 13 requires us to move to gradle version 6.0(not released yet). The current version of gradle(`5.6.x`) uses groovy `2.5.4` and it has some issues with java 13


ketan:
🤘

agent-bootstrapper/build.gradle
```diff
@@ -18,6 +18,9 @@ import com.thoughtworks.go.build.VerifyJarTask
 
 description = 'GoCD Agent Bootstrapper'
 
+sourceCompatibility = 8
```
ketan: I'd suggested that we do this for `jar-class-loader` module, and not this module.

PR #6930, opened 2019-09-13
Move new agents spa behind a toggle

kritika-singh3:
Description: Moving the new Agents SPA behind a toggle `show_new_agents_spa`. Default is `true`.
To turn it off:
```
curl http://localhost:8153/go/api/admin/feature_toggles/show_new_agents_spa -d toggle_value=off -H 'Confirm:true' -u username:password
```


server/webapp/WEB-INF/rails/webpack/single_page_apps/agents.tsx
```diff
@@ -13,36 +13,57 @@
...
  */
-import $ from "jquery";
+
+import {SinglePageAppBase} from "helpers/spa_base";
+import {AgentsPage} from "views/pages/agents";
+
+import {AjaxPoller} from "helpers/ajax_poller";
 import m from "mithril";
 import Stream from "mithril/stream";
+// @ts-ignore
```
GaneshSPatil: why do we need to do a `ts-ignore`? What the lint is failing for?

GaneshSPatil:
@kritika-singh3 -- 

Do we need to do all of this for adding a toggle to route to either new or old SPA.

The current implementation merges the controllers at the frontend, causing tight coupling between new and the old agent spa's controller.

IMO, we should keep these two pages separate. (helps to easily get rid of the old page).

The toggle can be added at the url-rewrite level:
when toggle is off, `/go/agents` goes to `/go/agents` (exisitng route),
And, when the toggle is on, it will do `/go/agents` => `/go/new-agents`

Note, its an internal rewrite that spring does, so the right-hand side URL can be changed based on the toggle value.

See: https://github.com/gocd/gocd/pull/6335/commits/aa45676180c42a1e6fd13c61150c270cefe87f72 for reference.

server/webapp/WEB-INF/rails/webpack/single_page_apps/agents.tsx
kritika-singh3: The following errors pops up if `ts-ignore` is removed: 
`TS7016: Could not find a declaration file for module 'models/agents/agents'. '/Users/kritikasingh/gocd/server/webapp/WEB-INF/rails/webpack/models/agents/agents.js' implicitly has an 'any' type`

kritika-singh3: Some of the files needed does not have an `export`. Hence, `import` is throwing the above mentioned error. For that purpose `ts-ignore` is needed. For the rest, used `require` instead of `import`.

kritika-singh3:
@GaneshSPatil agreeing with you. Closing this PR and will add the toggle the way you suggested.

Issue #6940, opened 2019-09-16
Prepopulate parameter names if select template depending on parameters


marques-work:
See issue #6726

ankitsri11:
verified on GoCD Version: 19.9.0 (10194-34fca8f7682a65ef5fbd6726d51fd52b0c041bc6).

PR #6941, opened 2019-09-17
Param defaults

kierarad:
Issue: #6940 https://github.com/gocd/gocd/issues/6940

### Description

This PR will populate the param names when creating a pipeline from a template that references params. 

As part of this PR I updated the template config API (now, V5) to return the params with the template config object. Incidentally, I updated the template config API to depend on the latest `StageRepresenter` (so the approval object will include `allow_only_on_success` if `true`)


marques-work:
@kierarad the template selection doesn't fill in the PaC preview :/. We should look into that.

image

marques-work:
@ibnc I just fixed it in 64b7a4a

ibnc:
@marques-work I think you pushed that as soon as I hit submit for my review lol. You're changes look fine

ibnc:
@marques-work @kierarad Once the build goes green with Marques' changes, and if you're happy with the PR Kiera, then I'll go ahead and merge

marques-work:
> It'd be nice if we had a route that'd return the JSON for List<PipelineTemplateConfig> with the name and params. It'd mean we wouldn't need to fetch the data when they select a template.

@ibnc I had a similar thought, but that could be addressed down the line as you said. FWIW, we still would need v5 of the template config API because V4 used the old representers which didn't include recent updates to the stage config objects; pipelineconfig v9 and templateconfig v4 APIs were out-of-sync as of the last release. templateconfig v5 fixes this.

server/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/template_editor_spec.tsx
```diff
@@ -73,6 +89,18 @@ describe("AddPipeline: TemplateEditor", () => {
     const dropdown = helper.byTestId("form-field-input-template");
     expect(dropdown).not.toExist();
   });
+
+  it("should populate parameters when present", () => {
+    spyOn(TemplateConfig, "getTemplate").and.callFake((name, onSuccess) => {
+      onSuccess(new TemplateConfig(name, [new PipelineParameter("paramName", "")]));
+      expect(paramList().length).toBe(1);
```
marques-work: Putting expectations here isn't quite enough because if the `onSuccess()` handler never runs, the test won't fail. Even though this function is no longer asynchronous because of the stubbing, it could be valuable to make the test asynchronous and call `done()` within the `onSuccess()` to validate that it runs at all.

I did check out the code and try this, and I found the callback does not actually run :)

marques-work: @kierarad it's probably never getting there because the component is mounted with `new EmptyCache()`

marques-work: These issues should now be fixed in d5fa251

Issue #4931, opened 2018-07-09
Optional upstream dependencies


Stono:
Hola folks!
We have a dependency that we *do not* want to trigger all of the downstream pipelines, but we do want it to pull that dependency when the downstream pipelines trigger of their own accord.

If you look at this screenshot, we want:

  - platform-helm-at-service *not* to trigger docker.mercury-admin-tool (and the other apps)
  - docker.mercury-admin-tool to only trigger of their git repo, and when doing so, will consume the latest upstream of platform-helm-at-service

<img width="1067" alt="screen shot 2018-07-09 at 19 13 42" src="https://user-images.githubusercontent.com/1486729/42468262-463559ca-83ac-11e8-8174-c02c6d999d12.png">

So basically, it's like "configurable upstream material triggers", we want to control which materials cause a pipeline to build.

Is there any way to do this, or is this a feature request we could get in?

Stono:
ping.... I can't tell you how valuable this feature would be.  Deterministic yet not auto-triggering upstream dependencies

remy-tiitre:
Yeah, same here. With docker image build flow its really problematic, when your base images (os, java etc) trigger new builds. For example, when you are a Java house and you update your Java base images it could start building all your applications at the same time. (Sometimes this might be something you want - if java has some serious security vulnerability for example). But in most cases it would be fine when application plan picks up the new version when they release new version of the app. Yet, I would still want to have VSM support to see what plans have used the latest version or any previous versions.

Stono:
Exactly!

akshaydewan:
The most straightforward way I can think of is to add an option at the material level to toggle whether the pipeline should be triggered when a new revision is available. 
For e.g.
```
      <materials>
        <git url="https://github.com/gocd/ruby-functional-tests" shallowClone="true"/>
        <pipeline pipelineName="installers" stageName="docker" materialName="docker-images" triggerOnNewRevision="false" />
      </materials>
```
The `triggerOnNewRevision` (or whatever we choose to call it) attribute can be set to false.


From a UX point of view, we have a lot of options that determine whether a pipeline will be triggered:
- On the pipeline level, we have "Automatic Pipeline Scheduling"
- Making the first stage manual is the same as above
- A timer can be specified, which can optionally trigger only when new material is available.
- On the material level, we have an option to "Poll for new changes". My understanding is that this relates to server-side polling (MDU), but it can have implications on whether the build is automatically triggered too.

[...]

PR #6950, opened 2019-09-17
Optional upstream dependencies

akshaydewan:
Issue #4931 - Optional upstream dependencies

Description:
Add a `skipScheduling` attribute to a dependency material. This can be used to prevent a Pipeline from triggering a downstream build.

### DIFF
--- a/domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java
+++ b/domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java
@@ -231,7 +231,8 @@ public abstract class AbstractMaterial extends PersistentObject implements Mater
     }
 
     @Override
-    public boolean skipSchedulingOnChange() {
+    public boolean ignoreForScheduling() {
+        //overridden in DependencyMaterial. Other materials don't support this flag
         return false;
     }
 }
--- a/domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java
+++ b/domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java
@@ -229,4 +229,9 @@ public abstract class AbstractMaterial extends PersistentObject implements Mater
     @Override
     public void populateAgentSideEnvironmentContext(EnvironmentVariableContext context, File workingDir) {
     }
+
+    @Override
+    public boolean skipSchedulingOnChange() {
+        return false;
+    }
 }
--- a/server/src/main/java/com/thoughtworks/go/server/service/AutoBuild.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/AutoBuild.java
@@ -18,6 +18,7 @@ package com.thoughtworks.go.server.service;
 import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.CruiseConfig;
 import com.thoughtworks.go.config.PipelineConfig;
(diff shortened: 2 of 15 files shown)


akshaydewan:
@maheshp This is still WIP as I'm adding some more test scenarios, but can you take a look and let me know if you have any feedback?

The main logic for skipping the scheduling is in `AutoBuild.java`. I had considered adding a `SchedulingChecker` implementation instead, but that wouldn't give me the material revisions that have changed.

For manual and timer triggers of the pipleine, the latest upstream dependency will be picked up.

server/src/main/java/com/thoughtworks/go/server/service/AutoBuild.java
```diff
@@ -88,6 +89,10 @@ public void canProduce(PipelineConfig pipelineConfig, SchedulingCheckerService s
     public boolean isValidBuildCause(PipelineConfig pipelineConfig, BuildCause buildCause) {
         for (MaterialRevision materialRevision : buildCause.getMaterialRevisions()) {
             if (materialRevision.isChanged()) {
+                if (materialRevision.getMaterial() instanceof DependencyMaterial) {
+                    boolean skipSchedulingOnChange = ((DependencyMaterial) (materialRevision.getMaterial())).skipSchedulingOnChange();
+                    return !skipSchedulingOnChange;
```
arvindsv: This is just a drive-by comment. Suppose there are two dependency materials which are considered changed (a simple diamond dependency will do), and if the first one is set to skip scheduling on change and the second one is set to not skip scheduling on change, will this work?

akshaydewan: Yeah, Mahesh pointed that out yesterday. I'll be changing the logic and adding tests to verify it.

akshaydewan:
In this PR, the only way to set `skipScheduling` is by editing the XML. We can make the API and UI changes for pipeline config separately

arvindsv:
One thing to watch out for, that I've seen happen in cases where we allow editing by XML only is: You go and edit the XML. Everything looks fine. Then, you go and edit something else in the UI. Your XML change gets overwritten.

akshaydewan:
> One thing to watch out for, that I've seen happen in cases where we allow editing by XML only is: You go and edit the XML. Everything looks fine. Then, you go and edit something else in the UI. Your XML change gets overwritten.

Thanks for pointing it out. Verified this, the xml isn't getting overwritten.

domain/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterial.java
```diff
@@ -229,4 +229,9 @@ public void updateFromConfig(MaterialConfig materialConfig) {
     @Override
     public void populateAgentSideEnvironmentContext(EnvironmentVariableContext context, File workingDir) {
     }
+
+    @Override
+    public boolean skipSchedulingOnChange() {
```
maheshp: Probably add a test for this with the reason why its always false?

maheshp:
> One thing to watch out for, that I've seen happen in cases where we allow editing by XML only is: You go and edit the XML. Everything looks fine. Then, you go and edit something else in the UI. Your XML change gets overwritten.

>>Thanks for pointing it out. Verified this, the xml isn't getting overwritten.

Editing through UI does not overwrite, but the changes will be lost if using the PipelineConfig API V9, as mentioned above PipelineConfig API needs to bumped up to support `skipScheduling`

PR #6979, opened 2019-09-20
Add missing working dir replacement during agent duplication

SebastianGoral-md:
Issue: # none

Description: While using the provided script to duplicate an agent, we noticed that the working dir in the wrapper-properties.conf was not updated. This caused the new agents to fail during startup with the message "Already running agent launcher in this folder.".


bdpiparva:
@SebastianGoral-md PR LGTM. 

**Context:** This is required if installed agent version < 19.6.0 and then upgraded to version >= 19.6.0. During the upgrade, the **wrapper-properties.conf** file will be updated with the **wrapper_working_dir=/var/lib/go-agent**. As **wrapper-properties.conf** is included at the end in **wrapper.conf** file, It will override the property(wrapper_working_dir) configured in `wrapper.conf`. 

Earlier script to run multiple agents is only updating the `wrapper_working_dir=/var/lib/go-agent-${AGENT_ID}` in `wrapper.conf` and it will work fine except above-mentioned use case.

Issue #6661, opened 2019-07-23
Migrate internal Pipeline list API from rails to spark


GaneshSPatil:
https://github.com/gocd/gocd/blob/master/server/webapp/WEB-INF/rails/app/controllers/api_v1/admin/internal/pipelines_controller.rb

rajiesh:
Verified on 19.10.0 (10328-5cb62ca14461d56576551e0bbec2978c909ea9b2)

PR #7013, opened 2019-09-26
Migrate Internal Pipeline list API from rails to spark

viraj2712:
Issue- #6661


ketan:
Does the pipeline structure api work instead of this?

viraj2712:
Looks like `Pipeline Structure API` can be used by non-admin users as well, whereas `Pipelines API` is only for admin and group admin users.

Also, there is change in response of both the APIs:
- Response of `Pipeline Structure API` returns list of pipelines grouped by pipeline groups and templates.
- Response of `Pipelines API` returns list of pipelines.

api/api-internal-pipelines-v1/src/main/java/com/thoughtworks/go/apiv1/internalpipelines/InternalPipelinesControllerV1.java
```diff
@@ -0,0 +1,92 @@
...
+    public void setupRoutes() {
+        path(controllerBasePath(), () -> {
+            before("", mimeType, this::setContentType);
+            before("/*", mimeType, this::setContentType);
+
+            before("", mimeType, this.apiAuthenticationHelper::checkAdminUserOrGroupAdminUserAnd403);
+
+            get("", mimeType, this::index);
+
+            exception(HttpException.class, (Exception exception, Request request, Response response) -> {
```
bdpiparva: No need to do this. See `HttpException.class` for more information.

api/api-internal-pipelines-v1/src/main/java/com/thoughtworks/go/apiv1/internalpipelines/representers/PipelineConfigsWithMinimalAttributesRepresenter.java
```diff
@@ -0,0 +1,43 @@
...
+import com.thoughtworks.go.config.PipelineConfig;
+import com.thoughtworks.go.config.PipelineConfigs;
+import com.thoughtworks.go.spark.Routes;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class PipelineConfigsWithMinimalAttributesRepresenter {
+    public static void toJSON(OutputWriter outputWriter, List<PipelineConfigs> listOfPipelineConfigs) {
+        List<PipelineConfig> pipelines = listOfPipelineConfigs.stream().flatMap(pipelineConfigs -> pipelineConfigs.getPipelines().stream()).collect(Collectors.toList());
```
bdpiparva: Do we really need stream in `pipelineConfigs -> pipelineConfigs.getPipelines().stream()`?

viraj2712: done

viraj2712: `flatMap` requires `mapper element` as a `Stream` which we pass as second argument
Refactored

api/api-internal-pipelines-v1/src/main/java/com/thoughtworks/go/apiv1/internalpipelines/InternalPipelinesControllerV1.java
viraj2712: Removed the exception handling code

Issue #4756, opened 2018-05-18
Provide an option to sort by duration on the job run history page


varshavaradarajan:
##### Issue Type

- Feature enhancement

##### Summary

For all static agents, the job run history page provides some tabular data about the job and duration. Provide an option to sort the jobs based on the duration of the jobs.

https://build.gocd.org/go/agents/b9a506d0-1443-40ab-81a0-225c13c453a3/job_run_history

neoprez:
This url gives me a 403. Cant really understand the ask given that I can't view the url. Any instructions?

varshavaradarajan:
@neoprez - you need to have admin permissions to access the above URL. Sorry. 
If you are trying to contribute for hacktoberfest, I suggest you bring up your own GoCD setup - a GoCD server and a GoCD agent to reproduce this and then work on this issue.

https://developer.gocd.org/ gives you information on how to setup your dev environment.

neoprez:
I’ll install it on my local. This is for Hacktoberfest  

Best,
Rodny Perez

neoprez:
Please see:

https://github.com/gocd/gocd/pull/7069

GaneshSPatil:
Done as part of https://github.com/gocd/gocd/pull/7069 in GoCD release v19.10.0

PR #7069, opened 2019-10-09
Add sort by duration in job run history

neoprez:
Issue: #4756 

Description:

Add sort by duration in job run history. Both ASC and DESC

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/JobInstanceService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/JobInstanceService.java
@@ -259,21 +259,10 @@ public class JobInstanceService implements JobPlanLoader, ConfigChangedListener 
     }
 
     public JobInstancesModel completedJobsOnAgent(String uuid, JobHistoryColumns columnName, SortOrder order, Pagination pagination) {
-        List<JobInstance> jobInstances = jobInstanceDao.completedJobsOnAgent(uuid,
-                columnName == JobHistoryColumns.duration ? JobHistoryColumns.completed : columnName,
-                order, pagination.getOffset(), pagination.getPageSize());
+        List<JobInstance> jobInstances = jobInstanceDao.completedJobsOnAgent(uuid, columnName, order, pagination.getOffset(), pagination.getPageSize());
         CruiseConfig cruiseConfig = goConfigService.getCurrentConfig();
         for (JobInstance jobInstance : jobInstances) {
             jobInstance.setPipelineStillConfigured(cruiseConfig.hasPipelineNamed(new CaseInsensitiveString(jobInstance.getPipelineName())));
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/thoughtworks/go/server/service/JobInstanceService.java
```diff
@@ -261,11 +259,23 @@ public int totalCompletedJobsCountOn(String uuid) {
     }
 
     public JobInstancesModel completedJobsOnAgent(String uuid, JobHistoryColumns columnName, SortOrder order, Pagination pagination) {
-        List<JobInstance> jobInstances = jobInstanceDao.completedJobsOnAgent(uuid, columnName, order, pagination.getOffset(), pagination.getPageSize());
+        List<JobInstance> jobInstances = jobInstanceDao.completedJobsOnAgent(uuid,
+                columnName == JobHistoryColumns.duration ? JobHistoryColumns.completed : columnName,
+                order, pagination.getOffset(), pagination.getPageSize());
```
akshaydewan: We don't have a `duration` column in the sql resultset, is that why you're using the `completed` column? From what I see in the query in [JobInstance.xml](https://github.com/gocd/gocd/blob/master/server/src/main/resources/com/thoughtworks/go/server/dao/maps/JobInstance.xml#L290), the `completed` column is evaluated to `BuildStateTransition.stateChangeTime` for completed jobs.

So, in the sql, we end up sorting by when the job was completed, but that won't necessarily give us the correct order for the duration. I tested this out locally by running a few pipelines and then reducing the pagesize. The sort doesn't work properly when there are multiple pages.

This seems tricky to implement, but one way to do it would be to calculate the duration in the sql itself.

neoprez: Correct, I sorted by completed given that there is no duration column in table. I totally disregarded the fact that pagination happens in the sql instead of the code hence it doesn't work properly because of pagination. 

I'll do the calculation of duration in the sql query instead. I'll update the PR.

neoprez:
Please review updates

akshaydewan:
Thanks for making the change. It looks fine to me. I just triggered the PR pipeline once to check if it goes green.

@maheshp Since we're now doing some extra work in the query, the performance may be affected slightly. I think we can just check it on build.gocd after this change gets deployed, unless you have any concerns.

peterrosell:
Is this tested with PostgreSQL as database?
TIMESTAMPDIFF function seems to not exist in PostgreSQL.

akshaydewan:
> Is this tested with PostgreSQL as database?
> TIMESTAMPDIFF function seems to not exist in PostgreSQL.

Yes, I have added a different query for postgres (which is in a private repo)

Issue #7079, opened 2019-10-09
[Config Repo SPA] Use a more conspicuous loading animation; the spinner in the corner is not noticeable


ibnc:
- Maybe a striped line across the width, a lá personalization view loader?

PR #7080, opened 2019-10-09
pac spa loading style changes

ibnc:
Issue: #7079


ibnc:
Screenshot of loading (note, it's animated the same as building pipelines):

Screen Shot 2019-10-09 at 14 58 05

PR #7094, opened 2019-10-11
Use pipeline config api v10 to create pipelines

ibnc:
Older api versions were removed, and this caused the add pipeline flow to throw a 404 for create. Using v10 of the Pipeline Config api now


ketan:
It is now possible to use `latest` and not worry about bumping version up :)

PR #7106, opened 2019-10-15
Force the user to scan the repository before allowing config repo creation

### LINKED ISSUE #7101: Make `Scan Repository` required before allowing "Finish" (with caveats)
- Essentially, need to user to have tried to scan at least 1x
- Should deny completion if _and only if_ scan completed and definitively says "no config files found -- Did you push the file to this repo?"
    - This ensures that when other errors out of our control happen, the user can still proceed with creation if they know better

marques-work:
- Only enable "Finish" button when scan repo returns at least 1 file or when the repo clone times out
  - Change the "no config files found" message to be an alert style message and ask user to verify that they pushed the config files upstream

Fixes #7101


ibnc:
@marques-work There's a loop hole in the logic. 

Steps to reproduce:

1. Scan repository that does have configuration files
2. Change the repository URL or branch to one without config files

At this point the finish button is clickable, and will create the config repo. The finish button will only become disabled after scanning.

marques-work:
@ibnc I've added an `onchange` listener to reset the disabled state when the material URL changes

Issue #1875, opened 2016-02-02
Pipeline Config API Endpoint doesn't return the pipeline group


henriquegemignani:
The information returned by the `/go/api/admin/pipelines/:pipeline_name` doesn't return the pipeline group the pipeline belongs to.

Part of my workflow resolves on creating a new pipeline based on another. I was fetching the configuration of one pipeline, changing some fields, and calling `POST /go/api/admin/pipelines` to create the new one, however the request fails because of missing pipeline group.

ketan:
Conversation and context here — https://gitter.im/gocd/gocd?at=56b0d19e586242210ae10e0f

PR #7113, opened 2019-10-17
Added 'group' property to show call for Pipeline Config API

kritika-singh3:
Issue: #1875


kritika-singh3:
@marques-work Since, `PipelineAsCodeController` utilised the same representer, I have updated the same to return `group` property (which will always be `null`). Let me know if you think that will cause an issue.

Issue #6982, opened 2019-09-23
[EPIC] Environments Page re-write


maheshp:
##### Summary

EPIC issue to capture everything related to environments page re-write.

PR #7115, opened 2019-10-18
Updating New Environment SPA to support add new env

viraj2712:
Issue: #6982 

Description:
- Added ability to add new environment by specifying name.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
@@ -24,7 +24,7 @@ import {
   EnvironmentVariablesWithOrigin
 } from "models/new-environments/environment_environment_variables";
 import {EnvironmentsAPIs} from "models/new-environments/environments_apis";
-import {Origin, OriginJSON} from "models/origin";
+import {Origin, OriginJSON, OriginType} from "models/origin";
 
 export interface EnvironmentJSON {
   name: string;
@@ -67,8 +67,14 @@ export class EnvironmentWithOrigin extends ValidatableMixin {
   }
 
   static fromJSON(data: EnvironmentJSON) {
+    let origins = [];
+    if (data.origins) {
+      origins = data.origins.map((o) => Origin.fromJSON(o));
+    } else {
+      origins.push(new Origin(OriginType.GoCD));
+    }
     return new EnvironmentWithOrigin(data.name,
-                                     data.origins.map((o) => Origin.fromJSON(o)),
+                                     origins,
                                      Agents.fromJSON(data.agents),
                                      Pipelines.fromJSON(data.pipelines),
                                      EnvironmentVariablesWithOrigin.fromJSON(data.environment_variables));
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments_apis.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments_apis.ts
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-import {ApiRequestBuilder, ApiResult, ApiVersion} from "helpers/api_request_builder";
+import {ApiRequestBuilder, ApiResult, ApiVersion, ObjectWithEtag} from "helpers/api_request_builder";
 import {SparkRoutes} from "helpers/spark_routes";
 import {EnvironmentVariableJSON} from "models/environment_variables/types";
 import {PipelineStructure} from "models/internal_pipeline_structure/pipeline_structure";
-import {Environments} from "models/new-environments/environments";
+import {EnvironmentJSON, Environments, EnvironmentWithOrigin} from "models/new-environments/environments";
 
 export class EnvironmentsAPIs {
   private static LATEST_API_VERSION_HEADER = ApiVersion.latest;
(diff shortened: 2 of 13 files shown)


viraj2712:
@streisguth 
Does this look ok to you. Do you have any recommendations?

Preview for add-env-flow:
add_env_flow

\cc: @adityasood

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments.tsx
```diff
@@ -19,11 +19,15 @@ import m from "mithril";
...
 import {FlashMessage, MessageType} from "views/components/flash_message";
+import {HeaderPanel} from "views/components/header_panel";
+import {CreateEnvModal} from "views/pages/new-environments/create_env_modal";
 import {EnvironmentsWidget} from "views/pages/new-environments/environments_widget";
 import {Page, PageState} from "views/pages/page";
+import {AddOperation, SaveOperation} from "views/pages/page_operations";
 
-interface State {
+interface State extends AddOperation<EnvironmentWithOrigin>, SaveOperation {
   onSuccessfulSave: (msg: m.Children) => void;
```
vrushaliwaykole: `saveOperation` has this method. You can remove this line

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments/create_env_modal.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  private validateAndPerformSave() {
+    if (!this.environment.isValid()) {
+      return;
+    }
+    EnvironmentsAPIs.create(this.environment)
+                   .then((result) => {
+                     result.do(
+                       (successResponse: SuccessResponse<ObjectWithEtag<EnvironmentWithOrigin>>) => {
+                         this.environments().push(successResponse.body.object);
```
vrushaliwaykole: I don't think there is a need to add the environment here as we fetch data on successful save. Can you please check once?

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
```diff
@@ -67,8 +67,14 @@ export class EnvironmentWithOrigin extends ValidatableMixin {
   }
 
   static fromJSON(data: EnvironmentJSON) {
+    let origins = [];
+    if (data.origins) {
```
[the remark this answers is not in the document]
kritika-singh3: The origin will only be returned if you use the merged internal api. But while calling update API, no origin will be returned. Hence, the use of `EnvironmentJSON`.

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments_apis.ts
```diff
@@ -43,7 +43,26 @@ export class EnvironmentsAPIs {
...
+
+  static create(environment: EnvironmentWithOrigin) {
+    return ApiRequestBuilder.POST(SparkRoutes.apiEnvironmentPath(), this.LATEST_API_VERSION_HEADER, {payload: environment})
+                            .then(this.extractObjectWithEtag());
+  }
+
+  private static extractObjectWithEtag() {
+    return (result: ApiResult<string>) => {
+      return result.map((body) => {
+        const environmentJSON = JSON.parse(body) as EnvironmentJSON;
```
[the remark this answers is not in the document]
kritika-singh3: This method is utilized for the response of update call which does not have origin.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments/create_env_modal.tsx
kritika-singh3: No we don't fetch the data on `onSuccessfulSave`. We just update the flash message.

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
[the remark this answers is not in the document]
vrushaliwaykole: My bad. Alright.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments.tsx
```diff
@@ -19,23 +19,37 @@ import m from "mithril";
...
 
 export class NewEnvironmentsPage extends Page<null, State> {
   private readonly environments: Stream<Environments> = Stream(new Environments());
 
   oninit(vnode: m.Vnode<null, State>) {
     super.oninit(vnode);
+
+    vnode.state.onAdd = (e: MouseEvent) => {
+      e.stopPropagation();
+      this.flashMessage.clear();
```
GaneshSPatil: Avoid using `this` inside callback functions as a general practice.
Using `this` inside a callback function would require callee to bind `onAdd` method with an appropriate context object. 
This would mean doing `vnode.state.onAdd.bind(vnode.state)()` everytime we want to use this method.

In this example, `this` can be replaced with `vnode.state`.

vrushaliwaykole: Got it. will update the code

vrushaliwaykole: I forgot that this callback is an arrow function. And AFAIK, arrow functions bind to the context they are defined in. I am not sure when the issue that you mentioned will occur.

GaneshSPatil: > arrow functions bind to the context they are defined in

This is something new for me. Can you send me some documentation where I can read more about it?

GaneshSPatil: @vrushaliwaykole @dhanasp -- You are right.
With arrow functions, the `this` keyword always represents the object that defined the arrow function. In our case its the `vnode` object.

If we're using arrow function, can we remove the unnecessary `fn.bind(vnode)`?

vrushaliwaykole: Yes. sure. I will look into it.

PR #7124, opened 2019-10-21
Updated new environment SPA

kritika-singh3:
Issue: #6982 

Description: Updated the new Env SPA
 - Showing delete confirmation modal on delete env
 - Added API calls to edit pipeline and edit agents flow
 - Showing `hostname` instead of `uuid` in the env body

## TODO
 - [x] add environment api specs for patch (dependent on #7115)

### DIFF
--- a/api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
+++ b/api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
@@ -16,13 +16,18 @@
 
 package com.thoughtworks.go.apiv1.internalenvironments;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonElement;
 import com.thoughtworks.go.api.ApiController;
 import com.thoughtworks.go.api.ApiVersion;
 import com.thoughtworks.go.api.base.JsonOutputWriter;
+import com.thoughtworks.go.api.representers.JsonReader;
 import com.thoughtworks.go.api.spring.ApiAuthenticationHelper;
+import com.thoughtworks.go.api.util.GsonTransformer;
+import com.thoughtworks.go.api.util.MessageJson;
 import com.thoughtworks.go.apiv1.internalenvironments.representers.MergedEnvironmentsRepresenter;
-import com.thoughtworks.go.config.CaseInsensitiveString;
 import com.thoughtworks.go.config.EnvironmentConfig;
+import com.thoughtworks.go.server.service.AgentService;
 import com.thoughtworks.go.server.service.EnvironmentConfigService;
 import com.thoughtworks.go.spark.Routes;
 import com.thoughtworks.go.spark.spring.SparkSpringController;
@@ -33,8 +38,9 @@ import spark.Response;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Optional;
 
+import static java.lang.String.format;
 import static spark.Spark.*;
 
 @Component
@@ -42,12 +48,14 @@ public class InternalEnvironmentsControllerV1 extends ApiController implements S
 
     private final ApiAuthenticationHelper apiAuthenticationHelper;
     private final EnvironmentConfigService environmentConfigService;
+    private final AgentService agentService;
 
     @Autowired
-    public InternalEnvironmentsControllerV1(ApiAuthenticationHelper apiAuthenticationHelper, EnvironmentConfigService environmentConfigService) {
+    public InternalEnvironmentsControllerV1(ApiAuthenticationHelper apiAuthenticationHelper, EnvironmentConfigService environmentConfigService, AgentService agentService) {
         super(ApiVersion.v1);
         this.apiAuthenticationHelper = apiAuthenticationHelper;
         this.environmentConfigService = environmentConfigService;
+        this.agentService = agentService;
     }
 
     @Override
@@ -66,6 +74,7 @@ public class InternalEnvironmentsControllerV1 extends ApiController implements S
 
             get("", mimeType, this::index);
             get("/merged", mimeType, this::indexMergedEnvironments);
+            put(Routes.InternalEnvironments.ENV_NAME, mimeType, this::updateAgentAssociation);
         });
     }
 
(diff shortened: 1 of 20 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments.tsx
```diff
@@ -22,9 +22,11 @@ import {EnvironmentsAPIs} from "models/new-environments/environments_apis";
 import {FlashMessage, MessageType} from "views/components/flash_message";
 import {EnvironmentsWidget} from "views/pages/new-environments/environments_widget";
 import {Page, PageState} from "views/pages/page";
+import {DeleteConfirmModal} from "../components/modal/delete_confirm_modal";
+import {DeleteOperation, SaveOperation} from "./page_operations";
 
-interface State {
-  onSuccessfulSave: (msg: m.Children) => void;
+interface State extends SaveOperation, DeleteOperation<EnvironmentWithOrigin> {
+  // onSuccessfulSave: (msg: m.Children) => void;
```
vrushaliwaykole: Can we remove this commented line as it is not required?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments/edit_agents_modal.tsx
```diff
@@ -102,12 +105,15 @@ export class AgentFilterWidget extends MithrilViewComponent<AgentFilterWidgetAtt
...
 
-  constructor(env: EnvironmentWithOrigin, environments: Environments) {
+  constructor(env: EnvironmentWithOrigin, environments: Environments, onSuccessfulSave: (msg: m.Children) => void) {
     super(Size.medium);
-    this.modalState  = ModalState.LOADING;
+    this.onSuccessfulSave = onSuccessfulSave;
+    this.modalState = ModalState.LOADING;
     this.fixedHeight = true;
-    this.agentsVM    = new AgentsViewModel(env, environments);
+    this.agentsVM = new AgentsViewModel(env, environments);
```
vrushaliwaykole: Do we need to clone the environment model here as well?

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments_apis.ts
```diff
@@ -45,11 +45,23 @@ export class EnvironmentsAPIs {
   static patch(name: string, payload: EnvironmentPatchJson) {
     return ApiRequestBuilder.PATCH(SparkRoutes.apiAdminEnvironmentsPath(name), this.LATEST_API_VERSION_HEADER, {payload});
   }
+
+  static updateAgentAssociation(envName: string, agentUuids: string[]) {
+    return ApiRequestBuilder.PUT(SparkRoutes.apiAdminInternalEnvironmentsPath(envName), this.LATEST_API_VERSION_HEADER, {
```
vrushaliwaykole: Can we use public API instead of internal API? or this is a conscious decision we have made?

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environment_agents.ts
```diff
@@ -15,27 +15,45 @@
...
 import Stream from "mithril/stream";
-import {Agent, AgentJSON} from "models/environments/types";
 import {Origin, OriginJSON} from "models/new-environments/origin";
 
+export interface AgentJSON {
+  uuid: string;
+  hostname: string;
+}
+
+export class Agent {
```
vrushaliwaykole: Can we merge this information with `AgentWithOrigin`? Keeping two models for information of agent seems redundant. Also, we have reused agent class from 'environments/types'. So we can just add hostname in `AgentWithOrigin` class.

api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
```diff
@@ -66,6 +75,7 @@ public void setupRoutes() {
 
             get("", mimeType, this::index);
             get("/merged", mimeType, this::indexMergedEnvironments);
+            put(Routes.InternalEnvironments.ENV_NAME, mimeType, this::updateAgentAssociation);
```
vrushaliwaykole: I am not sure if we use internal APIs for modifications. We use them just to get data in the format required by view. Can you please confirm once?

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environment_agents.ts
kritika-singh3: This is a conscious move as `environments/types` is the older page which will be removed eventually. Hence, wanted to keep the models separate for both pages so that removal is easy. But yes a separate class doesn't makes sense. 
Should I just remove `Agent` class?

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments_apis.ts
kritika-singh3: There is no public API which can be used to modify single env-multiple agent association. Hence had to create an internal API for usage in SPAs.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments.tsx
kritika-singh3: sure, will do.

GaneshSPatil:
@kritika-singh3 @vrushaliwaykole -- We might not need to edit the merged environments index API to get the agent hostname.

As the environments API index is the shallow API, it should only return the unique identifier of the entities. Eg: Pipeline name, agent UUID.
Adding agent hostname to the API _would defeat the purpose of the shallow API_. 

---

> How to get the agent hostname on the UI?

On the environments SPA, we make merged environments index call to get all the environment information. We can make another API call to get all the agent's information. 
With `UUID` coming from environments API and `UUID` and `hostname` coming from agents API, we could do mapping of UUID to hostname at the frontend itself.

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environment_agents.ts
vrushaliwaykole: Yes you can remove Agent class and move `hostname` to `AgentWithOrigin` class

api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
vrushaliwaykole: As discussed with @maheshp, we have decided to keep this as it is as this is a very customized request for our front end.

PR #6921, opened 2019-09-12
API for server site urls config

viraj2712:
Issue: #4189 

Description:
- Add support for GET, PUT and POST call for `server-site-urls`

PR #7133, opened 2019-10-24
Return user friendly errors for server site urls

viraj2712:
Issue: #4189 #6921 

Description:
- Currently both UI and Api directly show the xsd validation errors for server site urls.
- Added errors on Site Urls in order to return user friendly errors.

Error Response:
```
{
  "message" : "Validations failed for siteUrls. Error(s): [Invalid format for site url. 'htp://foo.bar' must start with http/s, Invalid format for secure site url. 'http://foo.bar' must start with https]. Please correct and resubmit.",
  "data" : {
    "site_url" : "htp://foo.bar",
    "secure_site_url" : "http://foo.bar",
    "errors" : {
      "site_url" : [ "Invalid format for site url. 'htp://foo.bar' must start with http/s" ],
      "secure_site_url" : [ "Invalid format for secure site url. 'http://foo.bar' must start with https" ]
    }
  }
}

### DIFF
--- a/api/api-server-site-urls-config-v1/src/main/java/com/thoughtworks/go/apiv1/serversiteurlsconfig/ServerSiteUrlsConfigControllerV1.java
+++ b/api/api-server-site-urls-config-v1/src/main/java/com/thoughtworks/go/apiv1/serversiteurlsconfig/ServerSiteUrlsConfigControllerV1.java
@@ -24,9 +24,10 @@ import com.thoughtworks.go.api.spring.ApiAuthenticationHelper;
 import com.thoughtworks.go.api.util.GsonTransformer;
 import com.thoughtworks.go.api.util.MessageJson;
 import com.thoughtworks.go.apiv1.serversiteurlsconfig.representers.ServerSiteUrlsConfigRepresenter;
-import com.thoughtworks.go.config.ConfigTag;
+import com.thoughtworks.go.config.ErrorCollector;
 import com.thoughtworks.go.config.SiteUrls;
 import com.thoughtworks.go.config.exceptions.GoConfigInvalidException;
+import com.thoughtworks.go.server.controller.Message;
 import com.thoughtworks.go.server.service.EntityHashingService;
 import com.thoughtworks.go.server.service.ServerConfigService;
 import com.thoughtworks.go.spark.Routes;
@@ -40,7 +41,6 @@ import spark.Response;
 import java.io.IOException;
 import java.util.function.Consumer;
 
-import static com.thoughtworks.go.i18n.LocalizedMessage.entityConfigValidationFailed;
 import static spark.Spark.*;
 
 @Component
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: api/api-server-site-urls-config-v1/src/main/java/com/thoughtworks/go/apiv1/serversiteurlsconfig/ServerSiteUrlsConfigControllerV1.java [hunk 1]
```diff
@@ -77,8 +82,16 @@ public void setupRoutes() {
...
+
+        try {
+            serverConfigService.createOrUpdateServerSiteUrls(siteUrls);
+        } catch (GoConfigInvalidException e) {
+            response.status(HttpStatus.UNPROCESSABLE_ENTITY.value());
+            String errorMessage = entityConfigValidationFailed(siteUrls.getClass().getAnnotation(ConfigTag.class).value(), e.getAllErrorMessages());
+            return MessageJson.create(errorMessage, jsonWriter(siteUrls));
+        }
+
+        return index(request, response);
```
kritika-singh3: Any specific reason for using `index()` over `writerForTopLevelObject`?

viraj2712: Not really.
As we've same implementation in place under `index` method, thought of using the same.

kritika-singh3: Ok sound logical. But, IMO, we should keep the 2 calls different as in this way, the `create` response is dependent upon the implementation of `index`. If the latter changes, the former might need a change too. WDYT?

viraj2712: Agree.
Will update it.

api/api-server-site-urls-config-v1/src/main/java/com/thoughtworks/go/apiv1/serversiteurlsconfig/ServerSiteUrlsConfigControllerV1.java
```diff
@@ -77,19 +82,29 @@ public void setupRoutes() {
 
     String createOrUpdate(Request request, Response response) throws IOException {
         SiteUrls siteUrls = buildEntityFromRequestBody(request);
-        serverConfigService.createOrUpdateServerSiteUrls(siteUrls);
+
+        try {
+            serverConfigService.createOrUpdateServerSiteUrls(siteUrls);
+        } catch (GoConfigInvalidException e) {
```
GaneshSPatil: any particular reason for using `try-catch` instead of `exception` in the `setupRoutes`?

viraj2712: not specific, will update it

PR #7151, opened 2019-10-28
Style the initial state of the preview pane to make it look "not-editable"

### LINKED ISSUE #7111: Make preview pane look less "editable"
Maybe center (vertically - and maybe horizontally) the initial text or make it look like a watermark.

marques-work:
Fixes #7111

Looks like this now:

image


marques-work:
@streisguth what do you think? Ping me if you lack context.

marques-work:
Talked with @streisguth about the styles. Merging now.

PR #7155, opened 2019-10-30
Added support for group attribute in update pipeline config api

kritika-singh3:
- `group` attribute is mandatory for a `PUT` request
 - if the group name is different, a move operation will be performed

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
@@ -87,7 +87,7 @@ public class UpdatePipelineConfigCommand extends PipelineConfigCommand {
     }
 
     private boolean isRequestFresh(CruiseConfig cruiseConfig) {
-        boolean freshRequest = entityHashingService.md5ForEntity(cruiseConfig.getPipelineConfigByName(pipelineConfig.name())).equals(md5);
+        boolean freshRequest = entityHashingService.md5ForEntity(cruiseConfig.getPipelineConfigByName(pipelineConfig.name()), getPipelineGroup()).equals(md5);
 
         if (!freshRequest) {
             result.stale(EntityType.Pipeline.staleConfig(pipelineConfig.name()));
--- a/server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
@@ -18,26 +18,30 @@ package com.thoughtworks.go.config.update;
 import com.thoughtworks.go.config.CruiseConfig;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.PipelineConfigSaveValidationContext;
+import com.thoughtworks.go.config.PipelineConfigs;
 import com.thoughtworks.go.config.exceptions.EntityType;
 import com.thoughtworks.go.server.domain.Username;
 import com.thoughtworks.go.server.service.EntityHashingService;
 import com.thoughtworks.go.server.service.ExternalArtifactsService;
 import com.thoughtworks.go.server.service.GoConfigService;
 import com.thoughtworks.go.server.service.result.LocalizedOperationResult;
+import org.apache.commons.lang3.StringUtils;
 
 import static com.thoughtworks.go.config.update.PipelineConfigErrorCopier.copyErrors;
 
 public class UpdatePipelineConfigCommand extends PipelineConfigCommand {
     private final EntityHashingService entityHashingService;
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
```diff
@@ -53,6 +57,11 @@ private String getPipelineGroup() {
     @Override
     public void update(CruiseConfig cruiseConfig) {
         cruiseConfig.update(getPipelineGroup(), pipelineConfig.name().toString(), pipelineConfig);
+        if (!StringUtils.equalsIgnoreCase(group, updatedGroupName)) {
+            PipelineConfigs group = cruiseConfig.findGroup(this.group);
+            group.remove(pipelineConfig);
+            cruiseConfig.getGroups().addPipeline(updatedGroupName, pipelineConfig);
```
varshavaradarajan: So, this might be a little troublesome. The config save flow for entity enters a synchronized block in https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/config/GoConfigDao.java#L88, which means that a group might be deleted before the thread for this request acquires the write lock. 

Another thing is that the `addPipeline` method creates a group if the group name is not found. Is that intentional behaviour? If so, please add a test for the same.

kritika-singh3: > The config save flow for entity enters a synchronized block in https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/config/GoConfigDao.java#L88, which means that a group might be deleted before the thread for this request acquires the write lock.

That is a case I missed. Would adding a group existence check in `canContinue` handle it or would I need to add a check over here itself?

> Another thing is that the addPipeline method creates a group if the group name is not found. Is that intentional behaviour? If so, please add a test for the same.

Yes, it is intentional (the same happens in `CREATE`). I have added a test `UpdatePipelineConfigCommandTest.java#shouldUpdateThePipelineGroupIfDifferent`. Probably would rename it better.

varshavaradarajan: > That is a case I missed. Would adding a group existence check in canContinue handle it or would I need to add a check over here itself?
Update: there is a check of canEditPipeline in UpdatePipelineConfigCommand#canContinue which will blow up if the pipeline group does not exist.

I suggest moving this block to the update method in `UpdatePipelineConfigCommand.java`.  Anything that updates the cruise config is best added in the entity command's update method. Similar to how `CreatePipelineConfigCommand#update` does it.

kritika-singh3: @varshavaradarajan I took a deeper look into how the Update pipeline config and delete pipeline group works. As far as I understood, if a group is deleted that means it is an empty group and therefore the pipeline in it would have been either deleted or moved to some other group.
In the 1st case, a `RecordNotFoundException` would be thrown by the controller. For the 2nd case, the etag comparison should fail. So, I think we are in safe place over here to directly access the group.
WDYT?

cc:\ @maheshp

server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
```diff
@@ -74,11 +82,11 @@ public boolean canContinue(CruiseConfig cruiseConfig) {
     }
 
     private boolean canEditPipeline() {
-        return goConfigService.canEditPipeline(pipelineConfig.name().toString(), currentUser, result, getPipelineGroup());
+        return goConfigService.canEditPipeline(pipelineConfig.name().toString(), currentUser, result, getExistingPipelineGroupName());
```
maheshp: We need to check if user is admin of the new group to add the pipeline right?

kritika-singh3: Yes, we need to do. Will add that in the `update` method of the Command.

server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java
varshavaradarajan: Depends actually. You're considering just a delete pipeline group request - 
Supposing there's a create pipeline group request that is scheduled to be served right after this operation occurs. And suppose that the group name is the same as the new group name used in the update pipeline config API, then the create call (even though it enters the synchronized block before the update pipeline config API request), will error out because a group with that name already exists. 

There are always such unforeseen situations when updating the cruise config without acquiring the write lock.  So, it is better to perform anything related to updating the cruise config in the entity's update command.

Issue #6640, opened 2019-07-17
Pipeline pause message is not shown completely on dashboard


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Pipeline pause message is stripped on the dashboard.

##### Actual Result
![Screen Shot 2019-07-17 at 3 49 21 PM](https://user-images.githubusercontent.com/15275847/61368459-8671f500-a8ab-11e9-8718-b94c655ed4a3.png)

##### Expected Result
![Screen Shot 2019-07-17 at 3 56 36 PM](https://user-images.githubusercontent.com/15275847/61368468-896ce580-a8ab-11e9-8e78-43be9f29589e.png)

##### Possible Fix
Remove [`truncate-to-lines($line-count: 3);`](3a1817eaaa167a4da4dcf6fb6286498cc6/server/webapp/WEB-INF/rails/app/assets/new_stylesheets/single_page_apps/new_dashboard.scss#L338) line and add `font-size: 12px;` instead.

So the final `.pipeline_pause-message` class CSS looks like:
```
.pipeline_pause-message {
  font-size: 12px;
  margin:    5px 0 0 0;
  float:     left;
  clear:     both;
}
```

ketan:
Or maybe show a tooltip? Do we have a limit on how long the DB column is?

GaneshSPatil:
[`The pause_cause` has a DB limit of `255` characters](https://github.com/gocd/gocd/blob/master/server/db/migrate/h2deltas/300002_add_pause_info_columnsql#L17). 

_I feel 255 characters can be shown easily on the pipeline box._
Here is a screenshot with `255` character long pause message.

![Screen Shot 2019-07-17 at 4 10 20 PM](https://user-images.githubusercontent.com/15275847/61369276-68a58f80-a8ad-11e9-9899-cfadc763abec.png)



> Or maybe show a tooltip?

As a user, I would like to know the entire pause message upfront. A tooltip will cause another (unnecessary) user interaction?

GaneshSPatil:
This was closed by automation.
Reopening for testing..

rajiesh:
@GaneshSPatil @SnehaPL While dashboard is shows 255 char long pause message clearly, pipeline history page needs to be fixed for such large messages

[...]

PR #7161, opened 2019-10-31
Fixes for pipeline pause message on pipeline history page

viraj2712:
Issue: #6640


dhanasp:
Truncate the long pause message and after hovering on the pause message it will show full message as following - 

Screenshot 2019-11-01 at 1 32 57 PM
 
This change has been made on both pipeline-history and pipeline-edit page.

GaneshSPatil:
We might need to check whether any of the functional tests requires a tweak with this change.
\cc: @rajiesh @viraj2712

Issue #2990, opened 2016-12-12
Make 'material name' accessible through 'pipeline instance' api. 


tifigs:
##### Issue Type

- Feature enhancement

##### Summary

Make "material name" (dependency pipeline name) accessible through "pipeline instance" api. 

##### Details

Gocd has the nice habit to provide environment variables for all your dependencies named after the "material name". 
```
GO_DEPENDENCY_LABEL_BACKEND
GO_DEPENDENCY_LOCATOR_BACKEND
```
Thanks to this, you can create source-code-branches of whole pipeline hierarchies and you don't need to rename anything. So i can configure my pipelines `backend` and `backend.3.0.2` to both set the environment variable `GO_DEPENDENCY_LABEL_BACKEND` and not `GO_DEPENDENCY_LABEL_BACKEND_3_0_2`. 

Now the GoCD API does only provide the 'description', but not the 'name' of any matierals; 
https://api.go.cd/current/#get-pipeline-instance

So I propose to create a new version of the API which goes 

`$ curl 'http://ci.example.com/go/api/pipelines/PipelineName/instance/1'` 
```
[...]
        "material": {
          "description": "URL: https://github.com/gocd/gocd, Branch: master",
          "fingerprint": "61e2da369d0207a7ef61f326eed837f964471b35072340a03f8f55d993afe01d",
          "type": "Git",
          "name": "backend",
          "id": 4
        },
```
instead of 
```
[...]
        "material": {
          "description": "URL: https://github.com/gocd/gocd, Branch: master",
          "fingerprint": "61e2da369d0207a7ef61f326eed837f964471b35072340a03f8f55d993afe01d",
          "type": "Git",
          "id": 4
        },
```

Kind of related to #1907

cheers

maheshp:
Changes done as part of - https://github.com/gocd/gocd/pull/7174
Needs additional changes to the API response which will be done in 19.12.0 as part of - https://github.com/gocd/gocd/issues/7360

rajiesh:
Verified on 20.1.0 (11109-7a59f0f111d0d51960d0e3105d85d6e9585cb940)

PR #7174, opened 2019-11-04
Create versioned API for fetching pipeline instance

kritika-singh3:
Issue: #2990

Description:
 - timestamp in the form of `yyyy-mm-dd'T'hh:mm:ss'Z'` instead of a long value
 - added `name` property to material

### DIFF
--- a/api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1.java
+++ b/api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1.java
@@ -77,12 +77,12 @@ public class PipelineOperationsControllerV1 extends ApiController implements Spa
             before("", mimeType, this::verifyContentType);
             before("/*", mimeType, this::verifyContentType);
 
-            before(Routes.Pipeline.PAUSE_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
-            before(Routes.Pipeline.UNPAUSE_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
-            before(Routes.Pipeline.UNLOCK_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
-            before(Routes.Pipeline.TRIGGER_OPTIONS_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
-            before(Routes.Pipeline.SCHEDULE_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
-            before(Routes.Pipeline.INSTANCE_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateUserAnd403);
+            before(Routes.Pipeline.PAUSE_PATH, mimeType, apiAuthenticationHelper::checkPipelineGroupOperateOfPipelineOrGroupInURLUserAnd403);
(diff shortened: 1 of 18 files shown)


api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1.java
```diff
@@ -138,4 +153,16 @@ private PipelineScheduleOptions getScheduleOptions(Request req) {
         JsonReader jsonReader = gsonTransformer.jsonReaderFrom(req.body());
         return PipelineScheduleOptionsRepresenter.fromJSON(jsonReader);
     }
+
+    private Integer getCounterValue(Request request) {
+        try {
+            int counter = Integer.parseInt(request.params("pipeline_counter"), 10);
```
GaneshSPatil: Do we need to pass the radix? can't we simply use `Integer.parseInt(request.params("pipeline_counter"))`? where, by default, it converts the input to a signed decimal integer.

server/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -60,6 +60,20 @@
...
+    <to last="true">/spark/api/pipelines/${escape:$1}/instance/${escape:$2}</to>
+  </rule>
+
+  <rule>
+    <name>Pipeline Instance API</name>
+    <condition name="Accept">application\/vnd\.go\.cd\.v1\+json</condition>
+    <from>^/api/pipelines/([^/]+)/instance/([^/]+)(/?)$</from>
+    <to last="true">/spark/api/pipelines/${escape:$1}/instance/${escape:$2}</to>
+  </rule>
+
```
GaneshSPatil: Instead of using two different rules can we use only one rule and use regex in the condition?
`<condition name="Accept">application\/vnd\.go\.cd(.v*)*\+json</condition>` condition can be used instead of using 2 different rules with different conditions.

api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/representers/StageInstanceModelRepresenter.java
```diff
@@ -0,0 +1,43 @@
...
+import com.thoughtworks.go.api.base.OutputWriter;
+import com.thoughtworks.go.presentation.pipelinehistory.StageInstanceModel;
+
+public class StageInstanceModelRepresenter {
+    public static void toJSON(OutputWriter outputWriter, StageInstanceModel stageInstanceModel) {
+        if (stageInstanceModel.getResult() != null) {
+            outputWriter.add("result", stageInstanceModel.getResult().toString());
+        }
+        if (stageInstanceModel.getRerunOfCounter() == null) {
+            outputWriter.add("rerun_of_counter", "null");
```
GaneshSPatil: adding a string "null" will render it as a string in JSON (not at null value).

GaneshSPatil: #### Actual output:
`"rerun_of_counter" : "null"`

#### Expected output:
`"rerun_of_counter" : null`

api/api-pipeline-operations-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineoperations/PipelineOperationsControllerV1.java
kritika-singh3: No need to pass in the radix. Was just being cautious. Will remove the same.

PR #7175, opened 2019-11-04
Remove unversioned cancel stage api

kritika-singh3:
Description: We have introduced a versioned Stage cancel api in 19.8.0. Hence, removing the unversioned one.


server/src/main/webapp/WEB-INF/rails/app/controllers/api/stages_controller.rb
```diff
@@ -35,14 +35,6 @@ def index
     end
   end
 
-  def cancel_stage_using_pipeline_stage_name
-    pipeline_name = params[:pipeline_name]
-    stage_name = params[:stage_name]
-    result = HttpLocalizedOperationResult.new
-    schedule_service.cancelAndTriggerRelevantStages(pipeline_name,stage_name, current_user, result)
```
ketan: This method becomes unused, I think.

kritika-singh3: Yes, this method is no longer used. Removing the same.

Issue #6571, opened 2019-06-24
Option to allow plugin users to login without explicitly being added at each plugin level


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Summary

Currently, User Management section on the Server Configuration Page specifies an option to allow users to login via plugin into GoCD, even if they haven't been explicitly added to GoCD.

This is a global setting for all the plugins.
If turned on, all the users from all the auth-configs are allowed to log-in into the system without being explicitly added.
If turned off, no user is allowed to login into the system, without being explicitly added.

Imagine the following scenario,
I have 2 following auth configs:
1. `file-based-auth`: where I have added all the users from my team.
2. `github-auth`: where I have configured public GitHub for all organizations.

Now, with this setup, I want to allow all the users from `file-based-auth` and selected users from `github-auth`.
Turning on the global setting isn't a solution as it will allow all the users from GitHub to login into the system.
Turning off the global setting isn't too a solution as it will require administrators to manually go through all users list (from passwd file) and add them into the system explicitly.

##### Possible Solution
Allowing this flag (`allow users to login via plugin into GoCD, even if they haven't been explicitly added to GoCD`) to be configurable per auth config will provide more control. And administrators can choose to toggle it on or off based on the requirement.

#### Task Checklist
 - [x] Add config migration to migrate allowOnlyKnownUsersToLogin attribute from server to each auth config. (https://github.com/gocd/gocd/pull/7178)
 - [x] Remove old rails based server configuration page. (https://github.com/gocd/gocd/pull/7188)
 - [x] Introduce auth config API v2 to support allowOnlyKnownUsersToLogin. (https://github.com/gocd/gocd/pull/6605)
 - [x] Fix auth config SPA to support allowOnlyKnownUsersToLogin. (https://github.com/gocd/gocd/pull/7205)
 - [x] Document allowOnlyKnownUsersToLogin at auth config on docs.gocd.org
 - [x] Document Auth Config Api v2

GaneshSPatil:
@gocd/committers -- Thoughts??

bdpiparva:
@GaneshSPatil - I agree with the proposed solution. The flag should be provided for each plugin. For organization, an admin may want to add the LDAP user without any manual intervention but for the plugins like GitHub, they may want to disable it.

PS: if we decide to do it, we have to retain the existing selection for the current flag and apply it for all the installed plugin auth config.

arvindsv:
[...]

Issue #6400, opened 2019-06-10
Server Configuration Page Rewrite


GaneshSPatil:
##### Issue Type
- Feature enhancement

##### Summary

Migration existing server configuration page written in rails to a SPA.
- [ ] Server Management 
	 - [ ] API (#4189)
	 - [ ] UI
- [ ] User Management
	 - [ ] API
	 - [ ] UI
- [ ] Email Notification
	 - [ ] API (#6401)
	 - [ ] UI
- [ ] Pipeline Management
	 - [ ] API (#4189)
	 - [ ] UI
- [ ] Command Repository Management
	 - [ ] API (#4189)
	 - [ ] UI

adityasood:
![IMG_1592](https://user-images.githubusercontent.com/17826464/65592518-5a11dd80-dfac-11e9-8b43-add1000778c5.png)

GaneshSPatil:
Fixed as part of https://github.com/gocd/gocd/pull/7102

PR #7178, opened 2019-11-04
Allow only known users to login at each auth config level

GaneshSPatil:
Issue: #6400

Description:

Reattempt of https://github.com/gocd/gocd/pull/6601


GaneshSPatil:
Fixes: https://github.com/gocd/gocd/issues/6571

GaneshSPatil:
Unit and Integration Tests for the given changes passed: https://build.gocd.org/go/pipelines/value_stream_map/trigger/5291.

Functional Tests will not go green because of XML schema migration.

Issue #7152, opened 2019-10-28
Number (or re-number) the sections of the PaC flow.


marques-work:
### Context

The sections below the pipeline configuration may appear "optional". Having them numbered as "Step N" (just like the pipeline config) indicates to users that these are mandatory.

### Question to Explore

Do we keep the same numbering or change the numbering to better represent & group the actual conceptual hierarchy?

PR #7199, opened 2019-11-06
pac builder form changes

ibnc:
* config file lang selection now under builder form
* Numbered the steps

Related to #7152


ibnc:
screenshot of what's currently in this PR:

Screen Shot 2019-11-07 at 10 48 32

ibnc:
Screenshot from latest commit:


Screen Shot 2019-11-07 at 11 02 46

ibnc:
Updated screenshot after feedback:


Screen Shot 2019-11-07 at 14 03 24

marques-work:
@ibnc what do you think of 1b46452?

I gave a shot at a different approach for the layout fix because I was somewhat concerned about using `47.65%` which felt like a magic number. I ~think~ hope that it captures the intent of the SCSS rules to make the columns equal width while allowing the browser to calculate any width adjustments via `flex-shrink` and `flex-grow`. It was my hope that it would also be more robust across browsers.

PR #7213, opened 2019-11-11
Added validations to policy inside role

kritika-singh3:
- type can be `environment` and `*`
 - action can be `view` and `*`


arvindsv:
These actions can change. cc: @adityasood

arvindsv:
You might want to start with either `view` only or `*` only.

PR #7218, opened 2019-11-12
Fixes for the new server config SPA -

dhanasp:
Issue: #6400

Description:
  - Use `docLink` attribute for siteUrl and secureSiteUrl
  - Enable cancel button always
  - Fixed cancel for mail server section.
  - Added delete button in the SPA.


dhanasp:
Added delete button in mail server configuration - 

Screenshot 2019-11-12 at 12 35 52 PM

Also, on click of Cancel button we fetch data again to reset the changes. But `Mail server configuration` returns 404 as SMTP config may not be defined. This avoids re-rendering of page as it is captured as error response. So for now, we have ignored and handled it like:

```
return Promise.resolve(MailServerCrud.get())
          .then((results) => {
            if (results.getStatusCode() === 404) {
              vnode.state.mailServer(new MailServer());
            } else {
              results.do((successResponse) => {
                vnode.state.mailServer(successResponse.body);
              }, () => this.setErrorState());
            }
          });
```
Let me know if there is better way to do this.

vrushaliwaykole:
Updated delete buttons position - 

Screenshot 2019-11-29 at 1 50 45 PM

vrushaliwaykole:
Confirmation on click of cancel button

cancel button confirmation

PR #7219, opened 2019-11-12
Introduce versioned pipeline status API

kritika-singh3:
Description: Introduce versioned pipeline status API


maheshp:
This endpoint replaces the existing un-versioned Pipeline Status API  - https://api.gocd.org/a9.11.0/#get-pipeline-status

PR #7220, opened 2019-11-12
Removed `bind` on arrow functions.

vrushaliwaykole:
Description:

Context cannot be passed to arrow functions.
`this` always represents the context in which the arrow function is defined.
So binding arrow function does not make any sense.

In reference with https://github.com/gocd/gocd/pull/7115#discussion_r344556165


ketan:
Closing. While the implementation is correct, it adds cognitive load on the reader.

Issue #3326, opened 2017-03-28
Go API define page size


marko-milenkovic-myob:
##### Issue Type

- Feature proposal

##### Summary

As an API user, I would like to have the ability to define "Page Size" in endpoints that provide paging, e.g.

GET /go/api/pipelines/:pipeline_name/history[/:offset]
GET /go/api/stages/:pipeline_name/:stage_name/history[/:offset]
GET /go/api/jobs/:pipeline_name/:stage_name/:job_name/history[/:offset]

At the moment, only 10 can be retrieved per request.

Thank you!

CheolminConanShin:
I have same question!

victor-sim:
Me too. It's critical for us.

Stono:
Ditto, this frankly sucks :D

Issue #1364, opened 2015-07-20
Jobs API improvement


juhijariwala:
Currently, we don't have any API endpoint to get job instance. I would like to propose new API endpoint to get job instance.

**Old Job history API which has a list of job instances in the response.** 

URL:  `curl 'http://localhost:8153/go/api/jobs/mypipeline/defaultStage/job1/history' -u 'username: password'`

``` json
{
  "jobs": [
    {
      "agent_uuid": "278fb0b6-d3b8-47e1-9443-67f26bfb5c15",
      "name": "job1",
      "job_state_transitions": [],
      "scheduled_date": 1437369102373,
      "original_job_id": null,
      "pipeline_counter": 2,
      "rerun": false,
      "pipeline_name": "mypipeline",
      "result": "Passed",
      "state": "Completed",
      "id": 8,
      "stage_counter": "2",
      "stage_name": "defaultStage"
    },
    {
      "agent_uuid": "278fb0b6-d3b8-47e1-9443-67f26bfb5c15",
      "name": "job1",
      "job_state_transitions": [],
      "scheduled_date": 1437368981728,
      "original_job_id": null,
      "pipeline_counter": 2,
      "rerun": false,
      "pipeline_name": "mypipeline",
      "result": "Passed",
      "state": "Completed",
      "id": 6,
      "stage_counter": "1",
      "stage_name": "defaultStage"
    },
    {
      "agent_uuid": "278fb0b6-d3b8-47e1-9443-67f26bfb5c15",
      "name": "job1",
      "job_state_transitions": [],
      "scheduled_date": 1437367504683,
      "original_job_id": null,
      "pipeline_counter": 1,
      "rerun": false,
      "pipeline_name": "mypipeline",
      "result": "Passed",
      "state": "Completed",
      "id": 3,
      "stage_counter": "1",
      "stage_name": "defaultStage"
    }
  ],
  "pagination": {
    "offset": 0,
    "total": 3,
    "page_size": 10
  }
}
```

**New job instance API**

URL :  `curl 'http://localhost:8153/go/api/jobs/mypipeline/19/defaultStage/2/job1' -u 'username:password'`

``` json
{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/jobs/mypipeline/19/defaultstage/2/job1"
    },
    "doc": {
      "href": "http://api.go.cd/#get-job-instance"
    }
  },
  "agent": {
    "_links": {
      "self": {
[...]

PR #7221, opened 2019-11-12
Introduce versioned job history and instance API

kritika-singh3:
Issue: #3326, #1364, #7360 

Description: Introduce versioned job history API with support for providing page size and offset.
Format: 
`/api/jobs/:pipeline_name/:stage_name/:job_name/history?page_size=10&offset=1`

Also introducing API to get a particular job instance via `pipeline_counter` and `stage_counter`.
Format: `api/jobs/:pipeline_name/:stage_name/:job_name/instance/:pipeline_counter/:stage_counter`

 - Default value for `page_size: 10` and `offset:0`
 - The end user can provide any value between 10 and 100 for page_size

Also removed all the `id`s from the json output.


kritika-singh3:
This PR replaces the existing unversioned job history api https://api.gocd.org/current/#get-job-history.
and introduces a get job instance api

api/api-job-instance-v1/src/main/java/com/thoughtworks/go/apiv1/jobinstance/representers/JobInstanceRepresenter.java
```diff
@@ -0,0 +1,63 @@
...
+        if (jobInstance.getState() != null) {
+            outputWriter.add("state", jobInstance.getState().toString());
+        }
+        if (jobInstance.getResult() != null) {
+            outputWriter.add("result", jobInstance.getResult().toString());
+        }
+        if (jobInstance.isNull()) {
+            renderNullJobInstance(outputWriter);
+            return;
+        }
```
GaneshSPatil: How about returning a 404 instead of 200 for a `NullJobInstance`.

As we're bumping up the API version, we could consider this as bug-fix.

kritika-singh3: Agreed. Throwing a `RecordNotFoundException` in case a `NullJobInstance` is received from the service.

Issue #7222, opened 2019-11-12
Granular Authorization for different entities such as Environments, Config-Repos, Pipeline Groups etc.  


adityasood:
### Task Checklist

- [x] Introduce XSLT migration to support defining policy in cruise-config.xml
- [x] Restrict policy definition only for environments.
- [x] Use `view` only policy permissions to grant a user access to environments.
- [x] Bump up roles config API to support policy definition.
- [x] Document Roles config API
- [x] Allow Roles Config SPA to define policy per role.
- [x] Show `Admin` tab with `Environment` Subtab on UI when a user has permission to view the environment.
- [x] Allow users to access environments SPA with read-only permissions to environments.
- [x] Policy evaluation when dealing with multiple roles. See [this](#multiple-roles), [this](https://github.com/gocd/gocd/issues/7222#issuecomment-553301672) and [this](https://github.com/gocd/gocd/issues/7222#issuecomment-553332705)
- [x] Allow policy definition for config repos.
- [x] Document policy tag under [GoCD Configuration Reference](https://docs.gocd.org/current/configuration/configuration_reference.html)
- [x] Document Granular auth feature on [docs.gocd.org](https://docs.gocd.org) similar to [pipeline as code](https://docs.gocd.org/current/advanced_usage/pipelines_as_code.html)
- [ ] (Optional) Write a blog post demonstrating granular auth feature with environments.

### Issue Description

Initial discussions and conversations are documented here: https://github.com/gocd/gocd/issues/4735

**Current Issues:**

Today most of the global entities pipeline_group, template, SCM, package-repo, config-repo, environment, elastic agent profile, artifact store can be configured only by System Admins. In an enterprise setup with multiple teams using the same instance of GoCD, System Admins can have an annoying experience if they have to step in for things like,

1. Creating pipeline_groups
2. Creating environments

**Suggested approach:** Allow auth policies to be defined for roles configured in GoCD.

### Policy to be inlined at a Role level

```xml
<role name="config-repo-admins">
    <policy>
        <allow action="administer" type="configRepo">*</allow>
    </policy>
    <users>
        <user>foo</user>
        <user>bar</user>
    </users>
</role>
```
With the above policy all users belonging to the role config-repo-admins will be able to create, view, edit and delete any ConfigRepo.

```xml
<role name="config-repo-editors">
    <policy>
        <allow action="edit" type="configRepo">*</allow>
    </policy>
    <users>
        <user>foo</user>
        <user>bar</user>
    </users>
[...]

PR #7230, opened 2019-11-14
Roles config api v3 (#7222)

GaneshSPatil:
Issue: #7222

Description:

* Introduce role configs API v3 to add support for defining policy for a role


GaneshSPatil:
@kritika-singh3 -- Only the second commit needs to be reviewed from this PR.
The first commit is simply a copy-paste from v2

ketan:
> Only the second commit needs to be reviewed from this PR.
The first commit is simply a copy-paste from v2

such-smart-very

api/api-roles-config-v3/src/main/java/com/thoughtworks/go/apiv3/rolesconfig/representers/DirectiveRepresenter.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public static Directive fromJSON(JsonReader jsonReader) {
+        String permission = jsonReader.getString("permission");
+        if (permission.equals("allow")) {
+            return new Allow(jsonReader.getString("action"), jsonReader.getString("type"), jsonReader.getString("resource"));
+        } else if (permission.equals("deny")) {
+            return new Deny(jsonReader.getString("action"), jsonReader.getString("type"), jsonReader.getStringOrDefault("resource", null));
+        } else {
+            throw new JsonParseException(String.format("Invalid permission '%s'. It has to be one of 'allow' or 'deny'.", permission));
+        }
```
kritika-singh3: Few suggestions:
 - can we create individuals vars for action, type, etc and then create `Allow` inside `if`. 
    Reason: better readability 
- instead of throwing `JsonParseException`, can we create a dummy `UnknownDirective` and add a config error to the same. This will allow us to return the original payload with errors on them.

Reference: https://github.com/gocd/gocd/blob/master/api/api-secret-configs-v1/src/main/java/com/thoughtworks/go/apiv1/secretconfigs/representers/RulesRepresenter.java#L67-L88

GaneshSPatil: @kritika-singh3 -- Fixed the `JsonParseException` as suggested.
The API gives the following response in case of invalid permission:

```json
{
  "message" : "Save failed. An error occurred while saving the role config. Please check the logs for more information.",
  "data" : {
    "policy" : [ {
      "errors" : {
        "permission" : [ "Invalid permission, must be either 'allow' or 'deny'." ]
      },
      "permission" : "awllow",
	  ...
    } ],
	...
  }
}
```

PR #7233, opened 2019-11-14
Add support for defining policy to roles spa

kritika-singh3:
Issue: #7222 

Description: Updated the Roles Config SPA
 - to use Role Config API v3
 - render `policy` info if defined
 - support for defining `policy` during create/edit/clone operations.

Preview:
policy_in_role


server/src/main/webapp/WEB-INF/rails/webpack/models/roles/roles_crud.ts
```diff
@@ -19,7 +19,7 @@ import {SparkRoutes} from "helpers/spark_routes";
 import {BulkUserRoleUpdateJSON, GoCDRole, PluginRole, Role, Roles} from "models/roles/roles";
 
 export class RolesCRUD {
-  private static API_VERSION_HEADER = ApiVersion.v2;
+  private static API_VERSION_HEADER = ApiVersion.v3;
```
GaneshSPatil: use latest?

kritika-singh3: My initial thought as well but in all the SPAs we use targeted versions. Should I use latest?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/roles/policy_widget.tsx
```diff
@@ -0,0 +1,252 @@
...
+  role: GoCDRole | PluginRole;
+}
+
+export class PolicyWidget extends MithrilViewComponent<PolicyAttrs> {
+  static headers() {
+    return [
+      "Permission"
+      , "Action"
+      , "Type"
+      , "Resource"
```
GaneshSPatil: indentation?

server/src/main/webapp/WEB-INF/rails/webpack/models/roles/roles_crud.ts
GaneshSPatil: Hmm, with SPAs, it's tricky.
As if we use `latest`, A change in the API can possibly cause the SPA to break (unintentionally).

Let's use the targeted version instead of latest!

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/roles/policy_widget.tsx
kritika-singh3: Can u pls explain a bit more?

kritika-singh3: Moved the `,` to the same line as value.

kritika-singh3:
Updated preview:
policy_in_role

PR #7267, opened 2019-11-18
Fix inconsistent ordering when ordering by `lastTransitionTime`

### LINKED ISSUE #7138: GOCD - Agent Job History is inconsistent
##### Issue Type
Bug Report

##### Summary
We are using the API to fetch the agent job history and have found that when there are jobs that has got rescheduled they are missing a timestamp (database column buildstatetransitions.stateChangeTime). This has the effect that when using the web GUI all the rescheduled jobs are shown at the bottom of the first page and when using the API to fetch the latest 10 we always get only Rescheduled ones.

I have looked at the SQL-query for this API and see it's due to a missing "NULLS LAST" on the inner query where the order 'lastTransitionTime' is done.
We are running on version 18.12 at the moment, but are planning to upgrade quite soon. I have checked SQL-query in the master branch and it looks the same so I guess it's not fixed there either.

As a quick fix I have altered the SQL-query to solve the problem temporarily. See the attached file, [JobInstance.txt](https://github.com/gocd/gocd/files/3771282/JobInstance.txt), patch is on line 334-339.
A better fix would be to also add a timestamp on the Rescheduled buildStateTransitions so they get sorted in time in the job history or maybe just remove them, but you might have a reason for including them. Just fixing the timestamp will not work due to existing NULL values in the stateChangeTime column.

##### Environment

We are running in GOCD Enterprise version 18.12

ketan:
The `buildStateTransition` table contains a `NULL` transition time for
`completed` state for jobs that are re-scheduled. This behavior by
itself is not a problem as such. However when sorting based on the
`lastTransitionTime`, `NULL` values need to be handled more gracefully
by adding a `NULLS FIRST` or `NULLS LAST` depending on the desired
ordering.

Fixes #7138

### DIFF
--- a/server/src/main/resources/com/thoughtworks/go/server/dao/maps/JobInstance.xml
+++ b/server/src/main/resources/com/thoughtworks/go/server/dao/maps/JobInstance.xml
@@ -307,7 +307,6 @@
         joined.pipelineCounter,
         joined.stageCounter,
         joined.rerunOfCounter,
-        joined.duration,
 
         buildstatetransitions.id AS stateId,
         buildstatetransitions.buildid AS buildid,
@@ -324,10 +323,7 @@
             stages.name AS stageName,
             pipelines.counter AS pipelineCounter,
             stages.counter AS stageCounter,
-            stages.rerunOfCounter AS rerunOfCounter,
-            (
-                SELECT TIMESTAMPDIFF(MILLISECOND, stateChangeTime, bst.stateChangeTime) FROM buildStateTransitions WHERE buildid = builds.id AND currentState = 'Scheduled'
-            ) as duration
+            stages.rerunOfCounter AS rerunOfCounter
             FROM builds
             INNER JOIN stages on builds.stageid = stages.id
             INNER JOIN pipelines on stages.pipelineid = pipelines.id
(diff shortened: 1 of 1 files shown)


server/src/main/resources/com/thoughtworks/go/server/dao/maps/JobInstance.xml
```diff
@@ -324,10 +323,7 @@
             stages.name AS stageName,
             pipelines.counter AS pipelineCounter,
             stages.counter AS stageCounter,
-            stages.rerunOfCounter AS rerunOfCounter,
-            (
-                SELECT TIMESTAMPDIFF(MILLISECOND, stateChangeTime, bst.stateChangeTime) FROM buildStateTransitions WHERE buildid = builds.id AND currentState = 'Scheduled'
-            ) as duration
```
ketan: I checked, apparently there is no `duration` on the POJO, so removing the `duration` here and above in line 310 seems reasonable.

akshaydewan: The duration is used for sorting on JobHistory columns: https://github.com/gocd/gocd/pull/7069/files#diff-eac6a96b60afa44709aca82f5fa7a416R299

Issue https://github.com/gocd/gocd/issues/4756

ketan: Thanks. Bringing back that line.

PR #7271, opened 2019-11-18
Update Policy to support config repo

viraj2712:
Issue: #7222

Description:
- Added support for `config_repo` as type in policy.
- Updated Roles Config SPA to support `config_repo` as supported types.
- Updated Config Repo API v3 to implement policy.
- Show user specific config repos on Config Repositories SPA as per user permissions.

### DIFF
--- a/api/api-config-repos-v2/src/main/java/com/thoughtworks/go/apiv2/configrepos/ConfigReposInternalControllerV2.java
+++ b/api/api-config-repos-v2/src/main/java/com/thoughtworks/go/apiv2/configrepos/ConfigReposInternalControllerV2.java
@@ -27,6 +27,7 @@ import com.thoughtworks.go.config.GoRepoConfigDataSource;
 import com.thoughtworks.go.config.PartialConfigParseResult;
 import com.thoughtworks.go.config.exceptions.EntityType;
 import com.thoughtworks.go.config.exceptions.RecordNotFoundException;
+import com.thoughtworks.go.config.policy.SupportedAction;
 import com.thoughtworks.go.config.remote.ConfigRepoConfig;
 import com.thoughtworks.go.config.remote.PartialConfig;
 import com.thoughtworks.go.domain.materials.MaterialConfig;
@@ -43,8 +44,10 @@ import spark.Response;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import static com.thoughtworks.go.config.policy.SupportedEntity.CONFIG_REPO;
 import static com.thoughtworks.go.util.CachedDigestUtils.sha256Hex;
 import static spark.Spark.*;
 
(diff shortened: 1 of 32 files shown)


api/api-config-repos-v2/src/main/java/com/thoughtworks/go/apiv2/configrepos/ConfigReposInternalControllerV2.java
```diff
@@ -76,33 +79,46 @@ public String controllerBasePath() {
...
             before("", mimeType, this::verifyContentType);
 
             before("/*", mimeType, this::setContentType);
-            before("/*", mimeType, authHelper::checkAdminUserAnd403);
             before("/*", mimeType, this::verifyContentType);
 
             get(ConfigRepos.INDEX_PATH, mimeType, this::listRepos);
+            before(ConfigRepos.REPO_PATH, mimeType, (request, response) -> {
+                authHelper.checkUserHasPermissions(currentUsername(), getAction(request), CONFIG_REPO, request.params(":id"));
+            });
```
GaneshSPatil: extract the lambda into a separate function and reuse?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/roles.tsx
```diff
@@ -127,7 +128,7 @@ export class RolesPage extends Page<null, State> {
   }
 
   fetchData(vnode: m.Vnode<null, State>): Promise<any> {
-    return Promise.all([PluginInfoCRUD.all({type: ExtensionTypeString.AUTHORIZATION}), AuthConfigsCRUD.all(), RolesCRUD.all(), EnvironmentCRUD.all()])
+    return Promise.all([PluginInfoCRUD.all({type: ExtensionTypeString.AUTHORIZATION}), AuthConfigsCRUD.all(), RolesCRUD.all(), EnvironmentCRUD.all(), ConfigReposCRUD.all()])
```
GaneshSPatil: 😕 Going forward, this section of code would make a bunch of API calls based on what entity types are supported.
We should consider this area for a possible piece of improvement going forward to avoid making tons of API calls.

GaneshSPatil: How about trying out:

- Sending entity information as part of the embedded `meta`  field on the page (as this information is fetched only on the page load). This approach would avoid all API calls.
- Replacing all the API calls with only one internal API which returns all the entity information.

We can see what works the best for us and then take a decision about the approach!!

kritika-singh3: Yes agreed. We won't address this as a part of this PR. Can you create an issue for the same and link it here so that we can keep track and not lose the context. 
WDYT @GaneshSPatil

GaneshSPatil: Make Sense. Created an issue: https://github.com/gocd/gocd/issues/7381

GaneshSPatil:
These changes have been applied to config repo API **v2** and **not** API **v3**.

There is a discrepancy in the commit message.

PR #7296, opened 2019-11-24
Do not fail if directories already exist

amasucci:
Issue: #

Description: When mounting directories as docker volume the entrypoint fails with the following error:
`/docker-entrypoint.sh: cannot mkdir -v /godata/`


GaneshSPatil:
@amasucci @ketan -- Don't we need to add the similar flag while dir creation for Agent Image. [here](https://github.com/gocd/gocd/blob/master/buildSrc/src/main/resources/gocd-docker-agent/docker-entrypoint.sh#L77-L79) ?

PR #7300, opened 2019-11-25
Allow ajax button operation state to be monitored

akshaydewan:
There was a bit of duplication between https://github.com/gocd/gocd/commit/a0146465bd12447221eef6aa8c398cc989079905 and https://github.com/gocd/gocd/commit/c271c57004d3d5a907629a3b43a001e16374c07c

Since the use-case for Delete Confirmation Modal seems to be common, I moved the functionality in the button class itself and added an optional `ajaxOperationMonitor` that can be passed to disable other buttons in the group when the ajax operation is in progress.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
@@ -17,6 +17,7 @@ import classnames from "classnames";
 import {RestyleAttrs, RestyleComponent, RestyleViewComponent} from "jsx/mithril-component";
 import m from "mithril";
 import Stream from "mithril/stream";
+import {OperationState} from "views/pages/page_operations";
 import * as defaultStyles from "./index.scss";
 
 type Styles = typeof defaultStyles;
@@ -45,6 +46,7 @@ export interface Attrs extends OnClickHandler, RestyleAttrs<Styles> {
   align?: Alignment;
   dataTestId?: string;
   ajaxOperation?: (e: MouseEvent) => Promise<any>;
+  ajaxOperationMonitor?: Stream<OperationState>;
 }
 
 function dataTestIdAttrs(attrs: Attrs) {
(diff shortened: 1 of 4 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
```diff
@@ -88,18 +92,27 @@ abstract class Button extends RestyleViewComponent<Styles, Attrs> {
...
+
   view(vnode: m.Vnode<Attrs>) {
     const isSmall    = vnode.attrs.small;
     const isDropdown = vnode.attrs.dropdown;
     let clickHandler = vnode.attrs.onclick;
+
     if (vnode.attrs.ajaxOperation) {
       clickHandler = (e: MouseEvent) => {
-        if (this.ajaxInProgress()) {
+        if (this.ajaxOperationMonitor!() === OperationState.IN_PROGRESS) {
```
vrushaliwaykole: as `this.ajaxOperationMonitor` is already initialized I don't think `!` is required here and other places

server/src/main/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
```diff
@@ -108,10 +121,11 @@ abstract class Button extends RestyleViewComponent<Styles, Attrs> {
       <button {...Button.onlyHtmlAttrs(vnode.attrs)}
               {...dataTestIdAttrs(vnode.attrs)}
               onclick={clickHandler}
+              disabled={vnode.attrs.disabled || this.ajaxOperationMonitor() === OperationState.IN_PROGRESS}
               class={classnames(
                 this.css.button,
                 {[this.css.btnDropdown]: isDropdown},
-                Button.iconClass(this.ajaxInProgress, vnode.attrs.icon, this.css),
+                Button.iconClass(this.forceSpinner, vnode.attrs.icon, this.css),
```
vrushaliwaykole: was there any specific reason to add another state `forceSpinner`? This can be achieved by `this.ajaxOperationMonitor() === OperationState.IN_PROGRESS` right? We can move this to a method and call the same at both places(line 124 and 128)?

akshaydewan: I did this to handle cases where there are two or more ajax operations possible. E.g. On the new Pipeline creation page, there are two possible ajax operations - `Save + Edit full config` or `Save + Run this pipeline`. If we bind to the operationMonitor, both the buttons will show the spinner. Since the `forceSpinner` is local to the button, it will show the spinner only on the button which was clicked.

server/src/main/webapp/WEB-INF/rails/webpack/views/components/buttons/index.tsx
akshaydewan: Fixed

Issue #7301, opened 2019-11-25
Add wrapper-config files to the GoCD backup process


ankitsri11:
##### Issue Type

- Feature enhancement

##### Summary

Currently, the files under `/usr/share/go-server/wrapper-config` are not backed up and needs to be recreated when the GoCD server is restored with the backup data.

malkk:
Without these files being backed up, a restore of the server will fail on all but the most basic implementations.  Prior to the changes in 19.6, you could rely on the backup to have all the necessary information.  For this reason, I consider this a break/fix and not a feature enhancement.  Lack of this is preventing us from upgrading to current versions.

rajiesh:
Verified on 19.11.0 (10687-c532a61bb240ffbe59fe356445bea6433005563f)

PR #7312, opened 2019-11-26
Backup tanuki wrapper config

ibnc:
Issue: #7301


server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
```diff
@@ -85,9 +85,11 @@
     private final Database databaseStrategy;
     private volatile ServerBackup runningBackup;
     private static final String CONFIG_BACKUP_ZIP = "config-dir.zip";
+    private static final String WRAPPER_CONFIG_BACKUP_ZIP = "wrapper-config-dir.zip";
 
     private static final String CONFIG_REPOSITORY_BACKUP_ZIP = "config-repo.zip";
     private static final String VERSION_BACKUP_FILE = "version.txt";
+    public static final String WRAPPER_CONFIG_DIR = "wrapper-config";
```
ketan: For the correct location of the wrapper config dir see: https://wrapper.tanukisoftware.com/doc/english/props-envvars.html#default-definitions

akshaydewan:
I've added a commit to read the `WRAPPER_CONF_DIR` env variable.

One thing to note is that on the dev environment, we don't have the wrapper config. So the backup will always fail on that step when run from the IDE.

Issue #7176, opened 2019-11-04
Placeholder for Issues with new Environments SPA


GaneshSPatil:
#### Issues with new Environments SPA

- [x] No confirmation on the delete of an environment.
- [x] Edit environment variables popup should have a fixed height as of pipelines, agents edit popup.
- [x] When there are no physical agents, the edit agents modal shows `No agents matching search text 'undefined' found!` error.
- [x] No Cancel button for edit pipelines, agents and environment variables modal.
- [x]  Misaligned `No pipelines matching search text 'adsfasdf' found!`.
- [x] Misaligned `No agents matching search text 'adsfasdf' found!`.
- [x] Inconsistent popup headers
	- For pipelines - `Edit Pipelines Association`
	- For agents - `Edit Agents Association`
	- For Environment Variables - `Environment Variables`
- [x] Inconsistent save button for edit pipelines, agents and environment variables popup.
- [x] Show agent hostname instead of agent UUID.
- [x] a yellow border is shown at the top of the environment denoting the environment has no pipelines or agents configured. But there is no clear message denoting the same.
- [x] Show message when no environments are configured.
- [x] Add a button group to the delete environment button for consistent styling (across all pages).
- [x] Perform client-side validation of empty-environment-variable-name when an empty environment variable is added to the environment.
- [x] Clicking delete on a collapsed environment causes just <pre>`Are you sure you want to delete environment                      ?` </pre> message to show up, the environment name is rendered a couple of seconds later.
- [x] While the delete operation is in progress, disable `Yes Delete` and `Cancel` buttons and show spinner (to avoid user click `Yes Delete` option multiple times).
- [x] While the create operation is in progress, disable `Save` and `Cancel` buttons and show spinner (to avoid user click `Save` option multiple times).
- [x] Do not allow deletes for config repo environments
- [x] Disable remove environment variable button when it is defined in the config repository.
- [x] Invalid agent count is shown when the environment contains remote agents. (https://github.com/gocd/gocd/issues/7176#issuecomment-559349516)
- [x] Non-existing agent in the remote environment causes the addition of other agents to fail (https://github.com/gocd/gocd/issues/7176#issuecomment-559353500)

#### Possible Improvements

- [ ] Perform the first level of duplicate environment validation at the client-side.
- [ ] Show information about which entity has been defined in which config repository. (Similar to what we've today for the `Unavailable pipelines`)

GaneshSPatil:
Add an ability to link to the individual environment?

[...]

PR #7320, opened 2019-11-27
Environment config SPA fixes

vrushaliwaykole:
Issue: #7176 

Description:
1. Disabled delete button on new environments SPA if the environment is partially defined in config repo.
2. Rename method `editEnvironmentsPath` to `getEnvironmentPathOnSPA`
3. Disable remove environment variable button when it is defined in the config repository.
4. Fix for: Invalid agent count is shown when the environment contains remote agents. https://github.com/gocd/gocd/issues/7176#issuecomment-559349516
5. Fix for: Non-existing agent in the remote environment causes the addition of other agents to fail https://github.com/gocd/gocd/issues/7176#issuecomment-559353500


server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments/environments_widget.tsx
```diff
@@ -79,12 +79,18 @@ export class EnvironmentWidget extends MithrilViewComponent<EnvAttrs> {
         </div>
       </div>;
     }
+    let deleteTitle;
+    if (!environment.canAdminister()) {
+      deleteTitle = `You are not authorized to delete the '${environment.name()}' environment.`;
+    } else if (!environment.isLocal()) {
+      deleteTitle = `Cannot delete '${environment.name()}' environment as it is partially defined in config repo.`;
```
GaneshSPatil: can we change `Cannot delete '${environment.name()}' environment as it is partially defined in config repository.`?

Changing `repo` => `repository`?

vrushaliwaykole: will change it.

server/src/main/webapp/WEB-INF/rails/webpack/views/components/environment_variables/spec/index_spec.tsx
```diff
@@ -42,45 +53,46 @@ describe("Environment Variables Widget", () => {
 
     it("should have plain text environment variable fields", () => {
       const plainTextValue = plainTextEnvVar.value();
-      expect(helper.byTestId("env-var-name")).toHaveValue(plainTextEnvVar.name());
-      expect(helper.byTestId("env-var-value")).toHaveValue(plainTextValue ? plainTextValue : "");
-      expect(helper.byTestId("remove-env-var-btn")).toBeInDOM();
+      const wrapper        = helper.allByTestId("environment-variable-wrapper")[0];
+      expect(helper.byTestId("env-var-name", wrapper)).toHaveValue(plainTextEnvVar.name());
+      expect(helper.byTestId("env-var-value", wrapper)).toHaveValue(plainTextValue ? plainTextValue : "");
```
viraj2712: Should this be explicit assertion?
`if/else` in assertion doesn't make sense

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/new-environments/spec/environments_widget_spec.tsx
```diff
@@ -49,67 +43,99 @@ describe("Environments Widget", () => {
...
+    expect(deleteIcons).toHaveLength(2);
     expect(deleteIcons[0].title).toBe(`You are not authorized to delete the '${xmlEnv.name}' environment.`);
     expect(deleteIcons[0]).toBeDisabled();
     expect(deleteIcons[1].title).toBeFalsy();
     expect(deleteIcons[1]).not.toBeDisabled();
-    expect(deleteIcons[2].title).toBeFalsy();
-    expect(deleteIcons[2]).not.toBeDisabled();
+  });
+
+  it("should render disabled delete icon for the environments that user can not administer", () => {
```
viraj2712: How about `should render disabled delete icon for the environments defined in config repo and user can not administer`
as it verifies the same
https://github.com/gocd/gocd/pull/7320/files#diff-16caf4c51b6acb7dd2c1cb0bcadf106aR109

api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
```diff
@@ -108,17 +109,15 @@ public String indexMergedEnvironments(Request request, Response response) throws
...
-        Optional<List<String>> uuids = jsonReader.readStringArrayIfPresent("uuids");
-        if (uuids.isPresent()) {
-            return uuids.get();
+        JsonReader agents = jsonReader.readJsonObject("agents");
+
+        List<String> uuidsToAssociate = agents.readStringArrayIfPresent("add").orElse(Collections.emptyList());
+        List<String> uuidsToRemove = agents.readStringArrayIfPresent("remove").orElse(Collections.emptyList());
+
+        if (!uuidsToAssociate.isEmpty() || !uuidsToRemove.isEmpty()) {
+            agentService.updateAgentsAssociationOfEnvironment(envConfig, uuidsToAssociate, uuidsToRemove);
```
GaneshSPatil: @vrushaliwaykole @dhanasp -- There is a null check inside `updateAgentsAssociationOfEnvironment` as there is only one usage of this method and we're already making sure that the environment exists as part of `environmentConfigService.getEnvironmentConfig` should we remove the null check verification from `updateAgentsAssociationOfEnvironment` ?

api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
```diff
@@ -108,17 +109,15 @@ public String indexMergedEnvironments(Request request, Response response) throws
...
         }
-        throw halt(422, MessageJson.create(format("Json `%s` does not contain property '%s'.", new Gson().fromJson(request.body(), JsonElement.class), "uuids")));
+        return renderMessage(response, 200, "Environment '" + envName + "' updated successfully!");
```
GaneshSPatil: How about `return renderMessage(response, 200, EntityType.Environment("bob").updateSuccessful());` ?

We will need to define `updateSuccessful` utility method similar to `deleteSuccessful`

GaneshSPatil:
#### Issues with internal Environments API update:

- [ ] Use PATCH as request method instead of PUT.
- [ ] Improve error message. Change `Agents with uuids 'agent-1,agent-2' were not found!` to `Agents with uuids ["agent-1", "agent-2"] were not found!`
- [ ] Removal of an agent associated in config repository should fail with Unprocessible Entity 422 error. Whereas the current API returns 200 with no change in the environment.
- [ ] Should the API fail when a non-associated agent is removed from the environment?
- [ ] Should the API fail when the already associated agent is added to the environment again?
- [ ] Should the API fail when the same agent is is in both `add` and `remove` list?


#### Open question:
- [ ] Should this be a public API?
- [ ] Should this API support bulk update for pipelines and environment variables too (in case it is a public API)?

kritika-singh3:
> * [ ]  Improve error message. Change `Agents with uuids 'agent-1,agent-2' were not found!` to `Agents with uuids ["agent-1", "agent-2"] were not found!`

The error message is a generic error message generated via `EntityType#notFoundMessage` when a `RecordNotFoundException` is thrown. Do you still recommended the change?

> * [ ]  Removal of an agent associated in config repository should fail with Unprocessible Entity 422 error. Whereas the current API returns 200 with no change in the environment.
> * [ ]  Should the API fail when a non-associated agent is removed from the environment?
> * [ ]  Should the API fail when the already associated agent is added to the environment again?
> * [ ]  Should the API fail when the same agent is is in both `add` and `remove` list?

All these scenarios are not achievable via SPA. This internal API in question was created for SPA use only. Do we still need to think of fixing these scenarios?

api/api-internal-environments-v1/src/main/java/com/thoughtworks/go/apiv1/internalenvironments/InternalEnvironmentsControllerV1.java
kritika-singh3: @GaneshSPatil agreed that the method `updateAgentsAssociationOfEnvironment` is currently only used in the Internal controller. But is that enough to remove the check.
The service can be called from anywhere (in future) and shouldn't it be the responsibility of the method itself to handle all the cases? WDYT?
cc:\ @viraj2712

GaneshSPatil: There should be no reason a `null` is sent for environment config. I couldn't think of any scenario where an environment is null, hence the thought of removing that check.

PR #7321, opened 2019-11-27
Pipeline history spa

bdpiparva:
Description: Moving pipeline history spa to typescript

- Changed url  from `http://localhost:8153/go/tab/pipeline/history/up42` => `http://localhost:8153/go/pipeline/activity/up42`

- [x] Page Header
	- [x] Shows pipeline name
	- [x] Pause/Unpause button button next to pipeline name :
		* Pipeline is defined in config.xml
		* User is admin or a group admin
		* If pipeline is paused, that button becomes “Unpause” - invisible for users without permissions.

- [x] Approval gates:
	- [x] Has stage approval information in-between stage bars
	- [x] If the manual and waiting, it will be enabled(dark grey). On click, it should ask for confirmation.
	- [x] If approved already, and the next stage is in progress or finished, it will be disabled(light grey)
	- [x] If the next stage is not done, but the user does not have permission to approve (operate), it will be light grey
    - [x] Mouse-over on these two gates shows who approved the stage, `Approved by blah`, `Automatically approved`, or says  `Automatically approved` if done automatically.

- [x] Above the corresponding column of stage bars, the name of the stage is present - suitably wrapped to keep every column’s width the same
- [x] Stage bar shows the status of the stage run. Four types:
	- [x] Success - Green with check icon in the center
	- [x] Failure - Red with into circle icon in the center
	- [x] Cancelled - Yellow with ban icon in the center
	- [x] Not run - Grey

- [x] When config changes in between pipeline run, a new row of stage names appears, showing the old stage names
	* This happens for added/deleted/renamed stages
	* This also happens if you change a stage from “Automatic” to “Manual”
            
- [x] Mouse-over on stage to uncover actions
	- [x] All users get an “i” (information) icon, which takes them to the stage details page
	- [x] Users with permission to operate get a “Rerun stage” action if the stage has finished
	- [x] Users with permission to operate get a “Cancel stage” action if the stage is in progress

- [x] Ugly purple run button(Now it is play button)
    * If the first stage is manual and the user has operated permissions, the play button shows up, which, upon clicking, it triggers the pipeline and goes into a disabled state and then (upon the next refresh) will bring back information about the build you just triggered.
    * If the user has no permission to trigger a build, then it will be rendered disabled

- [x] Information about every pipeline instance:
	- [x] Pipeline revision - This is the revision of the latest modification (Shows like: revision: db962aed102)
	- [x] Time of run (actually that of latest modification) - In a human readable format (Shows like: “about 5 hours ago”)
	- [x] Link to modifications popup for that pipeline instance
	- [x] Handles content too big for column (by showing ellipsis - “...”)
            
- [x] Modifications popup. Clicking on  the `Trigger by ..` link brings up the modifications popup
	- [x] For SCM materials, it shows the committer, the commit message, and the revision. Within the commit message, escaping is sometimes broken (especially &gt; and &lt;).

- [x] Pagination, Shows 10 to a page.
- [x] Page updates every 10 seconds with new information.

#### TODO
- [ ] Beta feature of pipeline comment: Shows pipeline comment if it exists and the feature is toggled on 

@arvindsv - Should we keep or remove the pipeline comment feature?


server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity.tsx
```diff
@@ -0,0 +1,191 @@
...
+import {PipelineActivityWidget} from "views/pages/pipeline_activity/pipeline_activity_widget";
+import {AjaxPoller} from "../../helpers/ajax_poller";
+import {ApiResult} from "../../helpers/api_request_builder";
+import {SearchField, TextField} from "../components/forms/input_fields";
+import {HeaderPanel} from "../components/header_panel";
+import {PaginationWidget} from "../components/pagination";
+import {Pagination} from "../components/pagination/models/pagination";
+import {ConfirmationDialog} from "./pipeline_activity/confirmation_modal";
+import styles from "./pipeline_activity/index.scss";
+import {PipelineActivityHeader} from "./pipeline_activity/page_header";
```
kritika-singh3: Can we change all relative imports to absolute ones (where ever possible).

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity.tsx
```diff
@@ -0,0 +1,191 @@
...
+                                          unpausePipeline={this.unpausePipeline.bind(this)}
+                                          pausePipeline={this.pausePipeline.bind(this)}
+                                          isAdmin={Page.isUserAnAdmin()}
+                                          isGroupAdmin={Page.isUserAGroupAdmin()}
+                                          isEditableFromUI={this.meta.isEditableFromUI}/>;
+    return <HeaderPanel title={title} sectionName={this.pageName()} buttons={
+      <SearchField property={this.filterText} label={"Search"}
+                   dataTestId={"search-field"}
+                   placeholder={"Filter history..."}
+                   onchange={this.fetchData.bind(this)}/>
```
kritika-singh3: Should we debounce the API call on `onchange`?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/page_header.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {PipelineActivityHeader.pauseStatusText(vnode.attrs.pipelineActivity)}
+      </div>
+      {PipelineActivityHeader.pipelineSettingsLink(vnode)}
+    </div>;
+  }
+
+  private static pipelineSettingsLink(vnode: m.Vnode<Attrs>) {
+    if ((vnode.attrs.isAdmin || vnode.attrs.isGroupAdmin) && vnode.attrs.isEditableFromUI) {
+      return <div class={styles.iconContainer} data-test-id="page-header-pipeline-settings">
+        <Link target={"_blank"} href={`/go/admin/pipelines/${vnode.attrs.pipelineActivity.pipelineName()}/general`}>
```
kritika-singh3: Suggestion: can we move the link to `SparkRoutes` and use the reference here.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/build_cause_widget.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  protected doRenderButton(vnode: m.Vnode<DropdownAttrs & Attrs>): m.Children {
+    if (!vnode.attrs.pipelineRunInfo.revision()) {
+      return;
+    }
+
+    return <a class={styles.buildCauseButton}
+              data-test-id="trigger-with-changes-button"
```
kritika-singh3: It is not actually a button rather a link. So can we use `trigger-with-changes-link` as a data-test-id?

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/PipelineActivityController.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public String controllerBasePath() {
+        return Routes.PipelineActivity.SPA_BASE;
+    }
+
+    @Override
+    public void setupRoutes() {
+        path(controllerBasePath(), () -> {
+            before("", authenticationHelper::checkUserAnd403);
```
kritika-singh3: Should we check if the user has access to view the pipeline?

bdpiparva:
@kritika-singh3 - I will implement your suggestions in separate PR.

Issue #7332, opened 2019-11-28
Inconsistent delete behavior of Pipeline Group SPA and Pipeline Group API


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Pipeline Group API allows the deleting of the pipeline group containing only remote pipeline(s), whereas, on the Pipeline Config SPA, the delete pipeline group button is disabled.

##### Steps to Reproduce

1.  Define `pipeline-1` pointing to the group `group-1` in a config repository.
2. Verify `pipeline-1` and `group-1` shows up on the UI
3. Visit the Pipeline Group SPA page; Verify delete icon for the `group-1` is disabled.
4. Use Pipeline Group API to delete the `group-1`. The deletion of the `group-1` is successful with `Pipeline group with name 'group-1' was deleted successfully!`

##### Expected Results
Pipeline Group SPA and Pipeline Group API should have consistent behavior. Pipeline Group API should not allow deletion of pipeline group when it contains a pipeline (nevertheless of the pipeline origin)

##### Actual Results
Pipeline Group API allows deletion of pipeline group when it contains only remote pipelines

### Weird Behavior

Pipeline groups are not defined config repository, hence, when the config repository pipeline `pipeline-1` is merged with the GoCD config, we define an empty pipeline group `group-1`.

**After the deletion of the `group-1`, the GoCD configuration on the disk and the config object in memory goes out of sync.**

Because after the deletion of the group, the `pipeline-1` points to a non-existing group, the config object in the memory believes there is a group with name `group-1`, whereas there is no pipeline group!

As `pipeline-1` points to the non-existing `group-1`, the group shows up on the UI. Visiting the `edit`  page of the `group-1` fails with a **404** `Pipeline group with name 'group-1' was not found!` error.

rajiesh:
Verified on 19.11.0 (10687-c532a61bb240ffbe59fe356445bea6433005563f)

PR #7334, opened 2019-11-28
Fix inconsistent behaviour for delete group

kritika-singh3:
Issue: #7332

Description: If the group contains only config-repo pipelines, the DeletePipelineCommand would receive any empty group as we send configForEdit to the same which does not contains the PartialConfigs. Hence, delete is successful which leads to errors later on.

Moved the empty check to `canContinue` of the command wherein we get the group from go config service which takes into account the PartialConfigs.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/domain/PipelineGroups.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/domain/PipelineGroups.java
@@ -19,7 +19,6 @@ package com.thoughtworks.go.domain;
 import com.thoughtworks.go.config.*;
 import com.thoughtworks.go.config.exceptions.EntityType;
 import com.thoughtworks.go.config.exceptions.RecordNotFoundException;
-import com.thoughtworks.go.config.exceptions.UnprocessableEntityException;
 import com.thoughtworks.go.config.materials.PackageMaterialConfig;
 import com.thoughtworks.go.config.materials.PluggableSCMMaterialConfig;
 import com.thoughtworks.go.config.materials.ScmMaterialConfig;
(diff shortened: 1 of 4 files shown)


api/api-pipeline-groups-v1/src/main/java/com/thoughtworks/go/apiv1/admin/pipelinegroups/PipelineGroupsControllerV1.java
```diff
@@ -148,10 +148,15 @@ public String update(Request req, Response res) {
...
+        String groupName = req.params("group_name");
+        PipelineConfigs pipelineConfigs = fetchEntityFromConfig(groupName);
 
         HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
-        pipelineConfigsService.deleteGroup(SessionUtils.currentUsername(), pipelineConfigs, result);
+        if (!pipelineConfigs.isEmpty()) {
+            result.unprocessableEntity(String.format("Failed to delete group %s because it was not empty.", groupName));
+        } else {
+            pipelineConfigsService.deleteGroup(SessionUtils.currentUsername(), pipelineConfigs, result);
+        }
```
GaneshSPatil: @kritika-singh3 @viraj2712 -- How about performing this config entity related validation under [`DeletePipelineConfigsCommand.isValid`](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/config/update/DeletePipelineConfigsCommand.java#L45-L47)?

Reasons:
1. Controllers are not responsible for validations, they should just call the service methods.
2. This validation will not be performed for anyone directly calling `pipelineConfigsService.deleteGroup` method.

kritika-singh3: Agree that the current implementation leaves some margin of error. Moving the empty check to `DeletePipelineConfigsCommand.canContinue`.

Issue #7297, opened 2019-11-25
Place holder for issues on Roles SPA


maheshp:
As part of supporting granular auth https://github.com/gocd/gocd/issues/7222, the Roles  SPA was modified to support `policy`. Below is the list of issues on the Roles SPA,

- [x] Error while adding policy to roles if none of the attributes are selected. Add a client side error handling before making a API call. 
- [x] When a new role is added, a default Deny policy is added. Is this required or do we leave it upto to the user to add a policy?
- [x] The current message `The default policy is to deny access to all GoCD entities. Configure permissions below to override that behavior.` seems confusing. We need to change the message and provide link to Policy documentation.

kritika-singh3:
Created PR #7292 for first 2 points:
 1. Added client side validations for policy 
 2. Removed the default policy added when a new role is added. The user needs to add a new permission as per his/her need.

adityasood:
- [ ] `All` under the action list is misleading. `Administer` is what `All` would mean. We should remove that

- [ ] `All` under the type should be removed. We can add it once we have more than one entity to support

- [ ] Order of the actions can be `Allow` and `Deny`. This is more semantic. can be ignored

kritika-singh3:
@adityasood In reference to point 1 and 2, we have support for `*` for `type` and `action`.
Semantically, yes `*` and `administer` are one and the same. But if someone specifies the following: `<allow type="*" action="*">*</allow>`, without `All` the dropdown for respective attribute will be rendered blank. Hence, our decision of adding `All`.

![policy](https://user-images.githubusercontent.com/41165891/69776425-38a6bb00-11c2-11ea-8a8b-f0eecd42e39d.gif)

Point 3 sounds right. Will implement the same.

cc:/ @GaneshSPatil

adityasood:
@kritika-singh3 @GaneshSPatil 

Point 1: Since both All and administer are essentially the same I am of the opinion we should drop the support for * from the policy for actions. 

For point 2: We can keep the (*). We will document it properly

rajiesh:
Verified on 19.11.0 (10687-c532a61bb240ffbe59fe356445bea6433005563f)

PR #7339, opened 2019-11-29
Update policy to not accept wildcard as a valid action

kritika-singh3:
Issue: #7297 

Description:
 - Removed support of wildcard(`*`) as a valid action in `policy`
 - Added validation to `role` to validate policy during creation/updation
 - updated the roles spa regarding the same

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/RoleConfigCommand.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/RoleConfigCommand.java
@@ -17,13 +17,17 @@ package com.thoughtworks.go.config.update;
 
 import com.thoughtworks.go.config.BasicCruiseConfig;
 import com.thoughtworks.go.config.CruiseConfig;
+import com.thoughtworks.go.config.ErrorCollector;
 import com.thoughtworks.go.config.Role;
 import com.thoughtworks.go.config.commands.EntityConfigUpdateCommand;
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/thoughtworks/go/config/update/RoleConfigCommand.java
```diff
@@ -60,12 +64,12 @@ public boolean canContinue(CruiseConfig cruiseConfig) {
...
-            BasicCruiseConfig.copyErrors(preprocessedRole, role);
-            return false;
-        }
+        preprocessedRole.validateTree(RolesConfigUpdateValidator.validationContextWithSecurityConfig(preprocessedConfig));
+
+        List<ConfigErrors> allErrors = ErrorCollector.getAllErrors(preprocessedRole);
+        BasicCruiseConfig.copyErrors(preprocessedRole, role);
 
-        return true;
+        return allErrors.isEmpty();
```
GaneshSPatil: `BasicCruiseConfig.copyErrors` is an iterator over config object. Can we avoid traversing the entire config object (including children) if there are no errors by changing this to:
```java
        List<ConfigErrors> allErrors = ErrorCollector.getAllErrors(preprocessedRole);
        boolean isEmpty = allErrors.isEmpty();
        if(!isEmpty) {
                BasicCruiseConfig.copyErrors(preprocessedRole, role);
        }
        return isEmpty;
```

PR #7368, opened 2019-12-04
Remove Server State API

GaneshSPatil:
#### Description:
* Server State API is used to turn on/off server state (active/passive)
  while using business continuity addon.

* Changing GO_SERVER_STATE to active requires a restart as timer-threads
  are not scheduled when the server is in passive mode. Changing
  GO_SERVER_STATE to active without server restart will cause some sub-
  systems of GoCD to not function appropriately.

* Changing GO_SERVER_STATE to passive requires a restart as timer-threads
  are scheduled when the server is in active mode. Changing
  GO_SERVER_STATE to passive without server restart will cause some sub-
  sytems of GoCD to continue to function even in passive mode.


GaneshSPatil:
Do not merge until all functional specs are green!!

GaneshSPatil:
Build passed: https://build.gocd.org/go/pipelines/value_stream_map/regression-SPAs-PR/1046

Issue #7336, opened 2019-11-28
Missing "Send Test Email" option on the new Email SPA


adityasood:
The "Send Test Email" option seems to be missed when we moved the Smtp Config from Rails to Spark.

adityasood:
spoke to @kritika-singh3 regarding the position of `Send Test Email` button. Closing this out. We can make that change later.

Verified on 19.12.0 (10874)

PR #7405, opened 2019-12-06
Added Send test mail on the email server configuration

kritika-singh3:
Issue: #7336

Description:
- added a send test mail method in mail server crud
 - updated the relevant widgets with the method


dhanasp:
Suggestion: Currently, there is no confirmation message after sending the test email. It will be good if there is some information regarding the same so that user will know if test Email was successfully sent or not.

Issue #7391, opened 2019-12-06
Tooltip for Resource gets cut from view on Roles page


akshaydewan:
<img width="1680" alt="Screenshot 2019-12-06 at 11 16 28 AM" src="https://user-images.githubusercontent.com/6024038/70299091-e6951380-1819-11ea-9816-021b3e224041.png">

rajiesh:
Verified on 19.12.0 (10842-29a6623b57f6c7224c75549a451965a2580dfada)

PR #7413, opened 2019-12-09
Reduced the width of tooltip wrapper as the same was cut off

kritika-singh3:
Issue: #7391 

Description:
 also removed the default permission added due to rebase


kritika-singh3:
Screen Shot 2019-12-09 at 2 28 56 PM

akshaydewan:
Other ways to (long term) fix this could be to tweak the z-index of the tooltip. It could be set higher than the modal overlay. Or we can change the alignment of the tooltip when it is near the edge of the view, but that might be tricky on a modal.

kritika-singh3:
@akshaydewan  Tweaking the z-index of the tooltip might not work as we have a fixed width on our [modals](https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/rails/webpack/views/components/modal/index.scss#L54-L64). The extra width on the tooltip will get cut off.

akshaydewan:
Ah, I see. Got it 👍

Issue #7390, opened 2019-12-06
Add support for edit as action in policy


kritika-singh3:
##### Issue Type

- Feature enhancement

##### Summary

Add support for `edit` as `action` in policy.

###### Implementation for entities:

1. Environment: The users with this permission will have access to view and update the specified resource.
2. Config-Repo: The users with this permission can view, update the config-repo as well as trigger an MDU.

GaneshSPatil:
### Verificiation: Support Edit Action for Policies
  - [ ] Support defining edit action on policies via xml.
  - [ ] Support defining edit action on policies via roles SPA.
  - [ ] Environments Edit Policy
  	- [ ] Environments Public API
      - [ ] A user with edit policy permission should be allowed to get all visible environments.
      - [ ] A user with edit policy permission should be allowed to view permissible environment.
      - [ ] A user with edit policy permission should be allowed to edit the environment.
      - [ ] A user with edit policy permission should be allowed to patch the environment.
      - [ ] A user with edit policy permission should not be allowed to delete the environment.
      - [ ] A user with edit policy permission should not be allowed to create an new environment.
  	- [ ] Environments Internal API
      - [ ] A user with edit policy permission should be allowed to get all visible environments.
      - [ ] Environments Internal API should return permissions (`can_edit`, `can_administer`) as part of response
      - [ ] A user with edit policy permission should be allowed to update environment agent association.
  	- [ ] Environments SPA
      - [ ] A user with edit policy permission should be allowed to get all visible environments.
      - [ ] A user with edit policy permission should be allowed to modify environment pipeline association.
      - [ ] A user with edit policy permission should be allowed to modify environment agent association.
      - [ ] A user with edit policy permission should be allowed to modify environment variables.
      - [ ] A user with edit policy permission should not be allowed to delete the environment. (Disabled delete icon)
      - [ ] A user with edit policy permission should not be allowed to create an new environment. (Failure on create)
  - [ ] Config Repository Edit Policy
  	- [ ] Config Repos Public API
      - [ ] A user with edit policy permission should be allowed to get all visible config repos.
      - [ ] A user with edit policy permission should be allowed to view permissible config repo.
      - [ ] A user with edit policy permission should be allowed to edit the config repo.
      - [ ] A user with edit policy permission should be allowed to perform preflight check on a config repo.
      - [ ] A user with edit policy permission should not be allowed to delete the config repo.
      - [ ] A user with edit policy permission should not be allowed to create a new config repo.
[...]

PR #7424, opened 2019-12-11
Support for edit action on Policy

GaneshSPatil:
##### Description:

* Add an ability to define edit action as part of the policy.
* Allow users to define edit action policy on roles SPA.
* Modify Internal Environments API v1 to return user permissions.
* Modify Internal Config Repos API to return user permissions.
* Fix environments SPA to enable/disable action buttons based on permissions.
* Fix config repo SPA to enable/disable action buttons based on permissions.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
@@ -113,7 +113,7 @@ export class EnvironmentWithOrigin extends ValidatableMixin {
       origins.push(new Origin(OriginType.GoCD));
     }
     return new EnvironmentWithOrigin(data.name,
-                                     Permissions.fromJSON(data.permissions),
+                                     data.permissions ? Permissions.fromJSON(data.permissions) : new Permissions(false, false),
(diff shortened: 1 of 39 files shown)


GaneshSPatil:
Things to verify are documented at: https://github.com/gocd/gocd/issues/7390#issuecomment-564388931

server/src/main/webapp/WEB-INF/rails/webpack/models/new-environments/environments.ts
```diff
@@ -142,7 +161,7 @@ export class EnvironmentWithOrigin extends ValidatableMixin {
 
   clone(): EnvironmentWithOrigin {
     return new EnvironmentWithOrigin(this.name(),
-                                     true,
+                                     new Permissions(true, true),
```
kritika-singh3: Why not retain the original value or set it to default `false`?

kritika-singh3:
Create environment modal - always shows the spinner : 

Screen Shot 2019-12-11 at 1 55 18 PM

GaneshSPatil:
These changes have been reverted as part of https://github.com/gocd/gocd/pull/7435 PR.

PR #7457, opened 2019-12-18
Templates config api v6

GaneshSPatil:
Description:
* Introduce template config api v6
* Use template config api v6 on templates spa

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/admin_templates/templates.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/admin_templates/templates.ts
@@ -27,13 +27,13 @@ export namespace TemplateSummary {
   export interface TemplateSummaryTemplate {
     name: string;
     can_edit: boolean;
-    is_admin: boolean;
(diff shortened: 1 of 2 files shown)


GaneshSPatil:
Following issue has been resolved as part of this changeset:

##### Issue: 
pipelines using templates and defined in config repository have edit action disabled on the templates page.

##### Expected Behavior: 
Such pipelines should have edit action enabled.

##### Implementation Details:
* On the Templates page, A pipeline is editable or not is decided based on `can_edit` attribute returned by templates API.
* The value of `can_edit` is a function of whether `pipeline defined in config repository` and `the current user has permissions to administer the pipeline`.
To fix this issue, we can simply modify `can_edit` to be a function of only whether `the current user has permissions to administer the pipeline`.

##### Puzzles:
1. `can_edit` field is not documented as part Templates API v5
2. Is it OK to change the behavior of the API?

##### solution:
Bump up Templates API to` v6`, change `can_edit` to `can_administer` which returns user permissions and document `can_administer` field on api.gocd.org.

PR #7463, opened 2019-12-18
Updated Job History API v1 to support cursor based pagination

kritika-singh3:
Issue:  #1353

Description:
Query params supported:

 - page_size: the end user can specify the number of records they wish to see in a single page. Should be between 10 and 100
 - after: the cursor value for fetching the next set of records
 - before: the cursor value for fetching the previous set of records


api/api-job-instance-v1/src/main/java/com/thoughtworks/go/apiv1/jobinstance/JobInstanceControllerV1.java
```diff
@@ -36,12 +35,14 @@
 
 import java.io.IOException;
 
+import static com.thoughtworks.go.server.service.PipelineHistoryService.BAD_CURSOR_MSG;
```
GaneshSPatil: should we find a better home for this constant?

kritika-singh3: What about in `ApiController` ?

kritika-singh3: Not moving it in `ApiController` as the same is also accessed in Controller classes. Creating a Constant class to store the relevant constants

kritika-singh3:
> `ServiceConstants` would be a too generic class. Should we use subclass nesting as what we have in `Routes.java`?

Sounds ok to me.

PR #7483, opened 2019-12-20
Move API to list pipeline instance based on search query to spark

kritika-singh3:
Description: For Comparison SPA, we have an endpoint (`/go/compare/:pipeline_name/list/compare_with/:other_pipeline_counter`) which returns a list of pipeline instances based on the search query. The said API return a HTML based json response. 

This PR moves the same to Spark and returns the list of matched pipeline instances as json.
Endpoint:
`go/api/internal/compare/:pipeline_name/list`

Command: 
```curl
curl 'http://go-server-api/go/api/internal/compare/up42/list?pattern=10&page_size=15' \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd.v1+json'
```
Query Params supported:

 - `pattern` : the pattern which needs to be matched
 - `page_size` : the no. of records needed. Can be between 10 and 100

### DIFF
--- a/api/api-compare-v1/src/main/java/com/thoughtworks/go/apiv1/compare/representers/ModificationRepresenter.java
+++ b/api/api-compare-v1/src/main/java/com/thoughtworks/go/apiv1/compare/representers/ModificationRepresenter.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.apiv1.compare.representers;
+
+import com.thoughtworks.go.api.base.OutputWriter;
+import com.thoughtworks.go.domain.materials.Modification;
+
+public class ModificationRepresenter {
+    public static void toJSON(OutputWriter outputWriter, Modification modification) {
+        outputWriter
+                .add("revision", modification.getRevision())
+                .addIfNotNull("modified_time", modification.getModifiedTime())
+                .add("user_name", modification.getUserName())
+                .add("comment", modification.getComment())
+                .add("email_address", modification.getEmailAddress());
+    }
+}
--- a/api/api-compare-v1/src/test/groovy/com/thoughtworks/go/apiv1/compare/representers/JobHistoryItemRepresenterTest.groovy
+++ b/api/api-compare-v1/src/test/groovy/com/thoughtworks/go/apiv1/compare/representers/JobHistoryItemRepresenterTest.groovy
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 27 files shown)


kritika-singh3:
JSON response:

```json
{
  "pipeline": [ {
    "name" : "up42",
    "counter" : 10,
    "label" : "10",
    "natural_order" : 10.0,
    "comment" : null,
    "build_cause" : {
      "trigger_message" : "modified by GoCD Test User <devnull@example.com>",
      "approver" : "",
      "trigger_forced" : false,
      "material_revisions" : [ {
        "material" : {
          "name" : "test-repo",
          "fingerprint" : "4879d548de8a9d71224a532bc1b02",
          "type" : "Git",
          "description" : "URL: test-repo, Branch: master"
        },
        "changed" : false,
        "modifications" : [ {
          "revision" : "0b21229ac2c7fa72dd890a03616ccd0353eaefdd",
          "modified_time" : "2019-12-19T05:05:34Z",
          "user_name" : "GoCD Test User <devnull@example.com>",
          "comment" : "Initial commit",
          "email_address" : null
        } ]
      } ]
    }
    ,...
  ]
}
```

api/api-compare-v1/src/main/java/com/thoughtworks/go/apiv1/compare/representers/ModificationRepresenter.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.thoughtworks.go.apiv1.compare.representers;
+
+import com.thoughtworks.go.api.base.OutputWriter;
+import com.thoughtworks.go.domain.materials.Modification;
+
+public class ModificationRepresenter {
+    public static void toJSON(OutputWriter outputWriter, Modification modification) {
+        outputWriter
```
bdpiparva: Same as above 

https://github.com/gocd/gocd/blob/4d47d8686ef7557db2358efe9762d786dffdcb9b/api/api-build_cause-v1/src/main/java/com/thoughtworks/go/apiv1/buildcause/representers/ModificationRepresenter.java#L26-L34

https://github.com/gocd/gocd/blob/4d47d8686ef7557db2358efe9762d786dffdcb9b/api/api-build_cause-v1/src/main/java/com/thoughtworks/go/apiv1/buildcause/representers/ModificationRepresenter.java#L26-L34

https://github.com/gocd/gocd/blob/8fc14776005961368530f4097c97962242df1421/api/api-materials-v2/src/main/java/com/thoughtworks/go/apiv2/materials/representers/ModificationRepresenter.java#L27-L33

https://github.com/gocd/gocd/blob/d9d5aaef384cfc1604e76a5d8cdb894b4ad888ee/api/api-pipeline-instance-v1/src/main/java/com/thoughtworks/go/apiv1/pipelineinstance/representers/ModificationRepresenter.java#L24-L29

api/api-compare-v1/src/test/groovy/com/thoughtworks/go/apiv1/compare/representers/JobHistoryItemRepresenterTest.groovy
```diff
@@ -0,0 +1,97 @@
...
+  void 'should deserialize into json'() {
+    Date date = new Date();
+    def jobHistoryItem = new JobHistoryItem("jobName", JobState.Completed, JobResult.Passed, date)
+    jobHistoryItem.setId(5)
+
+    def actualJson = toObjectString({ JobHistoryItemRepresenter.toJSON(it, jobHistoryItem) })
+
+    def expectedJson = [
+      "name"          : "jobName",
+      "scheduled_date": date.getTime().toString(),
```
bdpiparva: Any reason for serializing long? I think it should be a string with TZ information.

kritika-singh3: @maheshp suggested that we should prefer date in milliseconds as it is easier to be converted into appropriate time. With string with TZ info, users might have to parse the string.

kritika-singh3:
Create issue #7495 to make sure that the JSON representations are same across APIs

PR #7494, opened 2019-12-24
Improve the position of send test mail button

kritika-singh3:
Description: On mail server configuration, the send test mail button was not positioned correctly. Moved it a little to the left.

Preview:
Screen Shot 2019-12-24 at 11 17 59 AM

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/server-configuration/index.scss
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/server-configuration/index.scss
@@ -90,7 +90,7 @@
   }
 
   .admin-mail-input {
-    width: 60%;
+    width: 40%;
   }
 }
 


GaneshSPatil:
@kritika-singh3 -- the button and the text field looked a little far from each other. 
can we use `text-field-with-button` component instead?

kritika-singh3:
IMO, the said component makes sense when the two entity (the input and the action via button) are closely related. In this case, the send test mail is not only dependent on the administrator mail but also on other values in the form. Hence, decided against using it.
But if think otherwise, will make that change.

cc:\ @adityasood

adityasood:
@kritika-singh3 @GaneshSPatil What do you folks think of adding a `Send Test Email` option on the left-side in the same row as currently, we have save and cancel?

Screen_Shot_2019-12-24_at_1_52_02_PM

kritika-singh3:
For the sake of consistency, should the Send Test button to placed next to save (as on Create New Auth Config Modal) and Cancel button in the highlighted place?

adityasood:
@streisguth can you help with this? The context is shared above, let me know if you need any more information

PR #7503, opened 2019-12-26
Migrate Comparison SPA to typescript

kritika-singh3:
Migrated the current rails page for the comparison SPA to mithril based SPA.
Made a few changes:

 - The page header is changed to render "Compare" first followed by pipeline name (link to history SPA)
 - Broke `Modified By` column into `Modified By` and `Modified On` for non-dependency materials
 - The timestamps will be shown in local time (the title on hover will show server time) instead of showing `Z` based time string
 - There was a popup on click on the input which is now visible as the label on the same.
 - The browse timeline link is now present as a help text below the input instead on the popup
 - The timeline modal will make use of the new cursor-based history api and will contain the next and previous link(if present) instead of rendering page numbers.
 - The results on the typing in the input box will not contain the highlighted pattern string present as was the case with older page.

Edit: 
- In case of comparison involving a bisect instance - the results will be rendered directly with an info message stating the same.

The old SPA is behind a toggle `show_old_comparison_spa`. The curl command to view the old SPA is:
```curl
 curl -k 'https://go-server-url/go/api/admin/feature_toggles/show_old_comparison_spa' 
       -H 'Accept: application/vnd.go.cd+json' 
       -uusername:password 
       -X PUT 
       -d '{ "toggle_value": "on"}'
```


kritika-singh3:
Preview:
new-compare-spa

kritika-singh3:
In case of a comparison involving bisect instance:
Screen Shot 2020-01-03 at 10 04 55 AM

kritika-singh3:
Updated styles for timeline modal:
timeline

kritika-singh3:
Showcasing modifications from dependency materials in the search results:

Screen Shot 2020-01-31 at 9 45 33 AM

maheshp:
@kritika-singh3 comparison of pipeline runs which involves bisect seems a little confusing with regard to messaging. While there is a message which says - `This pipeline instance cannot be used to perform a comparison because it was triggered with a non-sequential material revision.` we actually compare the runs and show the result. The rails implementation seems to be better in this case,

New SPA
Screen Shot 2020-03-16 at 12 17 25 PM


Old Compare Page
Screen Shot 2020-03-16 at 12 17 16 PM

PR #7530, opened 2020-01-02
Add a toggle to fall back to rails based test connection API

kritika-singh3:
Description:
 - Added `admin` in the spark based API to make it consistent with the rails one.
 - Added a toggle at `urlrewrite` level which based on toggle `use_rails_based_material_test_connection_api` will redirect to spark or rails based API. Default is spark based.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/SparkOrRailsToggle.java
+++ b/server/src/main/java/com/thoughtworks/go/server/SparkOrRailsToggle.java
@@ -30,4 +30,8 @@ public class SparkOrRailsToggle {
             request.setAttribute("sparkOrRails", "spark");
         }
     }
+
+    public void testConnectionAPI(HttpServletRequest request, HttpServletResponse response) {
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/thoughtworks/go/server/MaterialTestConnectionToggle.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package com.thoughtworks.go.server;
+
+import com.thoughtworks.go.server.service.support.toggle.Toggles;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class MaterialTestConnectionToggle extends SparkOrRailsToggle {
```
GaneshSPatil: we don't need a separate class for this, we can add a method inside `SparkOrRailsToggle` class

PR #6897, opened 2019-09-06
Guess webhook repo url when it starts with ssh://

atbentley:
Issue: #3973

Description: When using ssh keys for authentication to private git repositories the git url can (has to?) start with ssh:// [0]. The current webhook doesn't check for these urls when guessing what repository a webhook corresponds to.

The linked issue identifies many issues such as customer ports etc. This pull request only addresses the starting with ssh:// part.

[0]: For example, the repo url may look like:
ssh://git@github.com/gocd/gocd.git

PR #7539, opened 2020-01-06
Guess webhook url without username for gitlab

GaneshSPatil:
Relates issue: #6897

Description:
Guess more ssh based URLs (without user)


bdpiparva:
Closing in favor of https://github.com/gocd/gocd/pull/7619

GaneshSPatil:
Reopening...

This PR was closed in favor of https://github.com/gocd/gocd/pull/7619. Spark-based implementation of the Webook Notification API.

Wherein, a user might choose to [toggle back to the rails based API](https://github.com/gocd/gocd/pull/7619/files#diff-ab70e01c50f545279431eed55e6a9743R45) (if in case there is a bug in the spark based API), which will cause these URLs not to work.

Adding these URLs to keep both (Rails and Spark) based API consistent in terms of behavior.

\cc: @bdpiparva

PR #7544, opened 2020-01-06
Update Pipeline Group Edit Modal

kritika-singh3:
Issue: #7551 

Description:
 - the errors from API on roles and users are rendered in case of a failed Update call
 - the failure on Users and Roles API is now ignored as they are used to provide completion suggestions. This will fix the page error on save of edit pipeline grp authorization if the user is a pipeline-grp admin


kritika-singh3:
Preview:
Screen Shot 2020-01-08 at 4 12 37 PM

kritika-singh3:
For fixing the auto-completion suggestion, there are 2 possible fixes:
 1.  Make the Users/Roles Index API open to pipeline grp admin
 2. Add a internal API to fetch all pipeline grps along with any auto-complete suggestions we need for the SPA.

bdpiparva:
@kritika-singh3 - Ineassted of making a change to existing APIs, I would prefer to have one internal API to provide autosuggestion information for the SPA as the page will not require all the information from the Roles and Users  API, WDYT?

kritika-singh3:
@bdpiparva Correct, we only require the loginName for the users and name for the roles.

maheshp:
@kritika-singh3 do we need 2 endpoints now, the `/api/internal/pipeline_structure` and `/api/internal/pipeline_structure/with_suggestions`?
If not, can we just have `/api/internal/pipeline_structure` which returns the users and roles. 

Also wondering if it would have been ok to return the list of users and roles as a data field on the SPA over adding a new endpoint.

kritika-singh3:
@maheshp The original endpoint was being called from 3 SPAs (The environment, templates and pipelines). Only the pipelines SPA was the one where we need the suggestions. So thought of not modifying the original but adding another.

> Also wondering if it would have been ok to return the list of users and roles as a data field on the SPA over adding a new endpoint.

Sending roles via data field might be ok but not so sure about users. Generally the team size can vary from 10 to 100s and sending such info via data field I'm not so sure.

PR #7571, opened 2020-01-10
Add internal pipeline structure api with suggestions

kritika-singh3:
Issue: Comment on #7544 

Description:
In order to always show auto-complete suggestions, introduced an internal api which will fetch the relevant groups with the needed information


kritika-singh3:
Wrapping both entities makes sense. Will do that.
WRT to API path, I couldn't think of anything better. Hence, just added `with_suggestions` to the already present Internal API. Let me know if someone has some suggestions for the same.

Issue #7572, opened 2020-01-10
Dashboard UI not updating on pipeline grp authorization updates


kritika-singh3:
##### Issue Type

- Bug Report

##### Summary

Dashboard view not getting updated on editing pipeline group authorization either via the new modal UI or Pipeline Group Config API.

###### Basic environment details

* Go Version: `20.1.0 (10977-495eebb16395605d283592b2c45c96ad9dd7da4b)`
* JAVA Version: `11.0.1`
* OS: `Mac OS X 10.13.6`

##### Steps to Reproduce

1. Login as admin user `admin`. Create a pipeline group `group` with a pipeline `test` in it.
2. Login via user `operate` and try to edit the pipeline group authorization via dashboard UI. You should not be able to edit as the user does not have operate permissions on it.
3. Logout and login as `admin`. Edit the authorization for `group` and provide user permissions `admin` to user `operate`. 
4.  Logout and login as `operate` and try to edit the pipeline group authorization via dashboard UI. 

##### Expected Results

You i.e. the user `operate` should now be able to edit the pipeline group authorization as we have granted the permissions

##### Actual Results

Still, the user `operate` can't edit the authorization via the dashboard UI.
Surprisingly, if we navigate to the Pipelines SPA, we can edit the permissions just not from the dashboard.

##### Any other info

Redo the step 3 via [Pipeline Group Update API](https://api.gocd.org/current/#update-a-pipeline-group). The results would remain the same.

However, if the `admin` or any user edits and saves the config xml (via UI), the expected results are achieved.

GaneshSPatil:
#### Possible Issue:

This issue possibly has caused due to dashboard cache not being updated upon modifying the pipeline group authorization.

#### Workaround:
* Perform full config save operation
  1. Login into GoCD as admin
  2. Visit **Config XML** tab (`/go/admin/config_xml`)
  3. Edit and Save config without doing any config changes.

* Restart GoCD.

kritika-singh3:
@GaneshSPatil I don't think a GoCD restart is required for the workaround.

GaneshSPatil:
@kritika-singh3 -- These are two different workarounds:
1. Either we perform a full config save operation.
2. Or we restart the GoCD server.

In both of the cases, the entire dashboard cache will be reloaded and relevant pipelines for the users will be loaded on the dashboard.

rajiesh:
Verified on 20.1.0 (11090-752c45e6489aa30ef3a1fd911c82f97eb8104eaf)

PR #7573, opened 2020-01-11
Recompute Dashboard Cache when Pipeline Group Config Changes (#7572)

GaneshSPatil:
Issue: #7572

Description:
* Recompute the entire dashboard cache when pipeline group
  config changes.


GaneshSPatil:
Hi @arvindsv,

These changes will cause the entire dashboard cache to be recomputed every time when the pipeline groups are updated.

#### Reason:
The current `GoDashboardConfigChangeHandler` exposes only two interfaces to update the dashboard cache. For **single pipeline** and for the **entire cruise-config.xml**. 
Implementing a new interface to update the dashboard cache for a **pipeline group config** would require us to write a lot of boilerplate code.

Also, the old pipeline group config authorization page used to do a full config save on every update, resulting in the entire dashboard cache to be recomputed. Hence, there is no change from the performance perspective in this approach.

arvindsv:
I wouldn't consider that boilerplate. :) But I can understand why. I'm fine with having it re-cache everything. We can see if we need to restrict it to the group, if there is a performance issue (unlikely).

Issue #7575, opened 2020-01-11
Expose current agent "resources" as GO_* environment variable


Antauri:
##### Feature enhancement

It would be ideal for example for enhanced Ansible-based continuous deployment scenarios, to know from the environment what "resources" does an agent have, by reading/parsing an environment variable of the same GO_* prefix as the others. 

The invoked Ansible playbooks can then react/adjust to the agent they run on (eg. based on agent resource/tag, filter some subset of the inventory, allowing basically unlimited scale for continuous deployment with Go.CD and Ansible).

##### Summary

In continuous deployment/delivery situations, one use-case for Go.CD agents is to run Ansible playbooks against an infrastructure. To avoid concurrency issues, using the inherent "locking" mechanism from Go.CD providing logical and physical safety of the infrastructure configuration, the team decides to use one (1) single agent for the whole infrastructure running "continuously" as each Git commit/on-demand scheduling triggers the next configuration (management) update.

To some degree, that scales but from a certain number upwards, the team no matter how they tune Ansible, they can't get faster runs. So the team decides to use multiple Go.CD agents and code their way towards a segregated/isolated execution of the same pipelines by adding some "agent-based" logic in the playbook code.

The team wants to get "smart" and based on agent tags (resources) eg. "core-infrastructure", "edge-infrastructure", "etc-infrastructure" have all agents assigned to the environment, run the same pipeline, but on a different subset of the infrastructure.

When looking up an environment variable that says on which agent does the playbook run from, there's no environment variable to expose such information (other than GO_JOB_NAME which is of the form "PipelineName-runOnAll-#" which doesn't say much). The alternative would be to maintain a mapping of agent hostnames to resources, but that defeats the purpose of being agnostic to the CD infrastructure (go.cd master + agent).

We think that adding a **GO_AGENT_RESOURCES** environment variable with a comma-separated or JSON/YAML list of assigned agent resources would help in continuous delivery situations where the DevOps people want to get smart/fancy about how pipelines execute. In our case it unblocks using more agents for CM with Ansible.

##### Environment

Debian-based master/agents.

###### Basic environment details

* Go Version: `19.12.0 (10888-29b0f854605987c8edab9dced4814b62dc751a11)`
* JAVA Version: `13.0.1`
* OS: `Linux 5.0.18-1-pve`
* Browser vendor and version (if relevant): `Google Chrome (latest)`

###### Additional Environment Details
We're in an Ansible based Go.CD continuous deployment environment.

[...]

PR #7589, opened 2020-01-13
Adding GO_AGENT_RESOURCES as environment variable

Antauri:
The issue fixes upstream GoCD #7575 issue by serializing the agent resources to an environment variable. Note I have signed the CLA.

Issue: #7575 

Description:
We think that adding a GO_AGENT_RESOURCES environment variable with a comma-separated or JSON/YAML list of assigned agent resources would help in continuous delivery situations where the DevOps people want to get smart/fancy about how pipelines execute. In our case it unblocks using more agents for CM with Ansible.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
@@ -292,9 +292,11 @@ public class BuildAssignmentService implements ConfigChangedListener {
 
                     final EnvironmentVariableContext environmentVariableContext = buildEnvVarContext(job.getIdentifier().getPipelineName());
 
-                    // Agent may have a NULL "resources" so we check and set an empty default here so that tests pass
-                    String agentResourcesAsCsv = agent.getResourceConfigs() == null ? "" : agent.getResourceConfigs().getCommaSeparatedResourceNames();
-                    environmentVariableContext.setProperty(GO_AGENT_RESOURCES, agentResourcesAsCsv, false);
+                    // Agent may have a NULL "resources"
+                    if (agent.getResourceConfigs() != null) {
+                        // Users relying on this env. var. can test for its existence rather than checking for an empty string
+                        environmentVariableContext.setProperty(GO_AGENT_RESOURCES, agent.getResourceConfigs().getCommaSeparatedResourceNames(), false);
+                    }
 
                     final ArtifactStores requiredArtifactStores = goConfigService.artifactStores().getArtifactStores(getArtifactStoreIdsRequiredByArtifactPlans(job.getArtifactPlans()));
                     BuildAssignment buildAssignment = BuildAssignment.create(job, pipeline.getBuildCause(), builders, pipeline.defaultWorkingFolder(), environmentVariableContext, requiredArtifactStores);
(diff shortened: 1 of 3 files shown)


kritika-singh3:
@Antauri Thanks for the PR. The change that you are proposing uses `jobPlan.getResources()` that will return the list of resources set on the `Job` via JobSettings Page. 

Please note that the resources set on Job and on Agent are not the same but a subset. Meaning the former (i.e. the job resources) needs to be a subset of the latter (i.e. agent resources).
E.g. the job can have resources: `java,api` and the agent can have resources: `java,jdk,api`.
So, technically, we are not adding `GO_AGENT_RESOURCES` but `GO_JOB_RESOURCES`.

Please make that change and also add a test verifying that the said env var was added to the context.

I hope I'm clear. Let me know if you have any query.

cc:\ @GaneshSPatil

Antauri:
Looking a bit at the code using the BuildAssignment class, I guess a better modification place is in the BuildAssignmentService.java around line 269 in the [createWork](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java#L269) method due to the fact I already have the "agent" variable there and I don't need to modify the BuildAssignment signature to pass that inside. 

What do you think? I guess the service class is better suited (as it also has the buildEnvVarContext method for the initial one). Suggestions welcomed.

kritika-singh3:
Yes @Antauri, I also think that the `createWork` method in `BuildAssignmentService.java` would be a good place to set the environment context after the line [292](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java#L292).

Antauri:
So:
- made the changes, reverting the BuildAssignment.java class to its "master" state;
- added the agent resources as comma-separated to the GO_AGENT_RESOURCES environment variable. Kept the same 'constants' usage defining this as a constant on the BuildAssignmentService class;
- enhanced two tests that already had sufficient scaffolding (eg. shouldResolveSecretParamsFromEnvironmentConfig) which had sufficient preparation of agent instances and pipeline and environment configuration. The two tests call "assignWorkToAgent" which calls "createWork" (from BuildAssignmentService) which takes it through the code-path that was just modified. I figured adding all that scaffolding for a new simple "environment variable" test would be useless and repetitive since this path is already exercised by these 2 tests already.

Please comment/feedback or why not ... merge :)
It built and tests ran (for this test class alone) on my "dojo" container.

server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -291,6 +292,10 @@ private Work createWork(final AgentInstance agent, final JobPlan job) {
 
                     final EnvironmentVariableContext environmentVariableContext = buildEnvVarContext(job.getIdentifier().getPipelineName());
 
+                    // Agent may have a NULL "resources" so we check and set an empty default here so that tests pass
+                    String agentResourcesAsCsv = agent.getResourceConfigs() == null ? "" : agent.getResourceConfigs().getCommaSeparatedResourceNames();
+                    environmentVariableContext.setProperty(GO_AGENT_RESOURCES, agentResourcesAsCsv, false);
+
```
kritika-singh3: How about we don't set the property if the agent doesn't have any resources set on it.

kritika-singh3:
Thanks. I took a look at the changes. I just have one query which I have posted above. I have also triggered a PR build. If both of them are resolved I'll merge it :)

server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
Antauri: It actually makes more sense to do that, true. I've guarded it with an if. Downstream to our Ansible use-case we can for check for its existence and default to "None". It leads towards much stronger code. Good call. Thanks!

Issue #7549, opened 2020-01-07
[Granular Auth] Introduce Granular Auth for Elastic Agent Cluster Profile and Elastic Profile


adityasood:
1.  Permission set for Creating a Cluster Profile

```xml	
<role name=”cp_admins”>
	<policy>
		<allow action=”administer” type=”cluster_profile”>cp_*</allow>
	</policy>
	<users>
		<user>cp_admin1<user>
		<user>cp_admin2<user>
	</users>
</role>
```
#### Implicit permissions

If a user is able to administer a Cluster Profile, they will be able to administer all elastic profiles within it.

2. Permission set for viewing a Cluster Profile

```xml	
<role name=”cp_viewers”>
	<policy>
		<allow action=”view” type=”cluster_profile”>cp_*</allow>
	</policy>
	<users>
		<user>cp_view1<user>
		<user>cp_view2<user>
	</users>
</role>
```
#### Implicit permissions

If a user is able to view a Cluster Profile, they will be able to view all elastic profiles within it. I suggest only view as these can be elevated by making someone administer an EP or all. 

3. Permission set for viewing EP in a cluster profile

**A user with the below policy can view all Elastic Profiles within cluster profile CP1.**

```xml
<allow action="view" type="elastic_profile">cp_1:*</allow>
```
##### Implicit permissions
View permission on cp_1

**A user with the below policy can view all Elastic Profiles within cluster profile CP1.**
```xml
<allow action="view" type="elastic_profile">cp_1:ep_*</allow>
```
##### Implicit permissions
View permission on cp_1

**A user with the above policy can view all Elastic Profiles** 
```xml
<allow action="view" type="elastic_profile">*:*</allow>
```
##### Implicit permissions
View permission on all cluster profiles

A user with the below policy can view all Elastic Profiles matching the wildcard ep_* within cluster profile CP1

```xml
<allow action="view" type="elastic_profile">cp_*:ep_*</allow>
```

##### Implicit permissions
View permission on cluster profiles matching the pattern cp_*

4. Permission set for creating an EP in a cluster profile
**A user with the below policy can administer all Elastic Profiles within cluster profile CP1.**

```xml
<allow action="administer" type="elastic_profile">cp_1:*</allow>
```
##### Implicit permissions
View permission on cp_1

**A user with the below policy can administer all Elastic Profiles within cluster profile CP1.**
```xml
<allow action="administer" type="elastic_profile">cp_1:ep_*</allow>
```
##### Implicit permissions
[...]

PR #7611, opened 2020-01-15
Allow normal users to access elastic agent configurations page (#7549)

GaneshSPatil:
Issue: #7549

Description:
	- Make elastic agent configurations page accessible to normal users
	- Use elastic agent profiles API v3 on SPA
 	- Use cluster profiles API v2 on SPA
 	- Disable cluster profile and elastic agent profiles action buttons
 	  when user does not have administer permissions
 	- Show error message on the cluster profiles wizard
 	- Remove dead code

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java
+++ b/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java
@@ -73,10 +73,13 @@ public class AuthorizeFilterChain extends FilterChainProxy {
                 .addAuthorityFilterChain("/admin/materials/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
                 .addAuthorityFilterChain("/admin/package_repositories/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
                 .addAuthorityFilterChain("/admin/package_definitions/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
-                .addAuthorityFilterChain("/admin/elastic_agent_configurations/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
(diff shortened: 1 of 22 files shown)


server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java
```diff
@@ -73,10 +73,13 @@ public AuthorizeFilterChain(AllowAllAccessFilter allowAllAccessFilter,
                 .addAuthorityFilterChain("/admin/materials/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
                 .addAuthorityFilterChain("/admin/package_repositories/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
                 .addAuthorityFilterChain("/admin/package_definitions/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
-                .addAuthorityFilterChain("/admin/elastic_agent_configurations/**", genericAccessDeniedHandler, ROLE_SUPERVISOR, ROLE_GROUP_SUPERVISOR)
-                //allow /go/admin/environments page to be accessible by normal users
+
                 .addAuthorityFilterChain("/admin/environments/**", genericAccessDeniedHandler, ROLE_USER)
                 .addAuthorityFilterChain("/admin/config_repos/**", genericAccessDeniedHandler, ROLE_USER)
+                .addAuthorityFilterChain("/admin/elastic_agents/**", genericAccessDeniedHandler, ROLE_USER)
+                .addAuthorityFilterChain("/admin/admin/elastic_profiles/**", genericAccessDeniedHandler, ROLE_USER)
```
GaneshSPatil: these are redirects, refer [this](https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/urlrewrite.xml#L133-L137) and [this](https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/urlrewrite.xml#L617-L621) section of code.

Not whitelisting these URLs will cause the redirects to not work the page to throw a 403 for non-admin users.

Issue #7482, opened 2019-12-20
Move APIs to spark


bdpiparva:
## APIs in rails

- [x] Process list(#7484)
- [x] Notification filter(#7491)
- [x] GoCD Version info(stale, latest_version, go_server)
- [x] Web hooks(#7619)
- [x] fanin(Not sure where this is used, Disscussed with @arvindsv, I will send PR to remove it) - #7510
	* Trace
	* Debug
	* Index
- [x] XML Feed (#7555)
	* Id.xml
	* Stages.xml
	* Pipelines.xml
	* Scheduled.xml

## API Docs

- [x] Process list - not needed
- [x] Notification filter
- [x] GoCD Version info(stale, latest_version, go_server)
- [x] Web hooks
- [x] XML Feed
	* Id.xml
	* Stages.xml
	* Pipelines.xml
	* Scheduled.xml

## Already moved and deprecated(or scheduled for deprecation)but still around

- [x] Admin template authorization
- [x] pipeline_group_config_list_api(config/pipeline_groups)
- [x] pipeline_status_api(pipelines/:pipeline_name/status)
- [x] History api
  * Pipeline history
  * Job history
  * Agent job run history

PR #7619, opened 2020-01-16
Move webhook api to spark

bdpiparva:
Issue: #7482


api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/GuessUrlWebHook.java
```diff
@@ -0,0 +1,53 @@
...
+            format("git@%s:%s.git", hostname, repoFullName),
+            format("git@%s:%s.git/", hostname, repoFullName),
+            format("ssh://git@%s/%s", hostname, repoFullName),
+            format("ssh://git@%s/%s/", hostname, repoFullName),
+            format("ssh://git@%s/%s.git", hostname, repoFullName),
+            format("ssh://git@%s/%s.git/", hostname, repoFullName),
+            format("ssh://%s/%s", hostname, repoFullName),
+            format("ssh://%s/%s/", hostname, repoFullName),
+            format("ssh://%s/%s.git", hostname, repoFullName),
+            format("ssh://%s/%s.git/", hostname, repoFullName)
```
GaneshSPatil: (Additionally) Read an environment variable? so that a user can pass more possible URLs?

bdpiparva: That can happen separately as It may involve templating the URLs

PR #7635, opened 2020-01-20
Enable Elastic Agent Status Report button for users with view permission for elastic profile (#7549)

GaneshSPatil:
Issue: #7549

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/SecurityService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/SecurityService.java
@@ -16,6 +16,8 @@
 package com.thoughtworks.go.server.service;
 
 import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.config.policy.SupportedAction;
+import com.thoughtworks.go.config.policy.SupportedEntity;
 import com.thoughtworks.go.server.domain.Username;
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/thoughtworks/go/server/service/SecurityService.java
```diff
@@ -207,6 +209,30 @@ public boolean isAuthorizedToViewTemplates(Username username) {
         return goConfigService.cruiseConfig().isAuthorizedToViewTemplates(username.getUsername());
     }
 
+    //todo: needs refactoring use AbstractAuthenticationHelper.doesUserHasPermissions
+    //a specific method to check whether a user has permission to access agent status report has been added
+    public boolean doesUserHasPermissionsToViewAgentStatusReport(Username username, String elasticAgentProfileId) {
```
GaneshSPatil: This method needs to be removed.
Instead, use `AbstractAuthenticationHelper.doesUserHasPermissions`

As of today, `JobController` can not autowire `ApiAuthenticationHelper` component, as it is a circular dependency (Requires dependency on module 'com.thoughtworks.go.api-base.main').

To refactor this,
We will need to move the Authorization related to methods from `AbstractAuthenticationHelper` to `SecurityService`.

PR #7646, opened 2020-01-20
Add support for multiple databases 

ketan:
Quick spike to demonstrate support for multiple DBs. Currently only briefly tested with H2, PG, and MySQL (mariadb may work, but haven't tested it)

**tl;dr;**

- Modify the file `development-utility/development-server/build.gradle` to change the dependency `:db-support:db-support-{mysql,postgresql,h2}` to point to a DB of your choice
- If using PG or mysql, create the file `config/db.properties`
    ```properties
    # For postgresql
    #db.driver=org.postgresql.Driver
    #db.url=jdbc:postgresql://localhost:5432/go
    #db.user=ketan
    
    # for MySQL
    db.driver=com.mysql.cj.jdbc.Driver
    db.url=jdbc:mysql://localhost:3306/go
    db.user=root

    # for H2, it should be possible to point to an external DB, but haven't tested it yet.
    ```
- If using MySQL, turn off identifier case sensitivity. See [this doc](https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html). This is required because some of the HQL and mybatis queries use table and column names in a case sensitive manner :-(

---

The long version

The original [`Database`](https://github.com/gocd/gocd/blob/19.12.0/addon-api/database/src/main/java/com/thoughtworks/go/database/Database.java) seemed to be doing quite some DB specific operations (startup/shutdown/backup), and some operations that should not have been DB specific (upgrade, for e.g.)

The new implementation:

- moves the DB migrations into a separate `DatabaseMigrator` class that performs migrations on `@PostConstruct`. This is similar to the earlier implementation that called `Database#upgrade` except that the migration is now managed using liquibase.
- Provides for DB specific operations (like backup and queryextensions), is still delegated to the DB specific implementation. See the `Database` and `ConnectionManager` class. The `ConnectionManager` class loads up a DB specific `BackupProcessor` and `QueryExtensions` class. It is still to be determined how to package/distribute these DB specific implementations, a quick way would be to package a fatjar/uberjar containing the `BackupProcessor`, `QueryExtensions` class along with the DB driver classes.


ketan:
For anyone asking why MySQL/H2/PostgreSQL and not other databases. The reason why it's easy to support these DBs is out-of-the-box support for case-insensitive column searches. This is needed because pipeline/stage/job names in GoCD are case-insensitive, and as such the DB needs to support case-insensitive lookups on columns that store these names.

While it's possible to change the queries to use `WHERE lower(name) = lower(?)` (or apply case insensitive indices, in some cases), this requires significant effort on changing a lot of SQL and HQL code in GoCD.

ketan:
This PR adds support for multiple databases, while at the same time, moving from an unmaintained dbdeploy to a well supported liquibase. Since this requires some manual intervention, I'm also taking this as an opportunity to move from H2 `1.3.x` to the latest `1.4.x`.

I'll put this comment in a better location later, but here is a quick summary of what users are expected to do (and some background on why). As an example, assume that this PR is merged for release `X` of GoCD:

- Version `X` will sniff for the presence of the `changelog` table. The `changelog` table is used by dbdeply to keep track of change scripts. The presence of this table will cause a startup failure, and point users to instructions on how to upgrade.
- The instructions for upgrade are:
  - Upgrade to version `X-1` of GoCD, this will make sure that the DB schema is latest wrt dbdeploy. More on this in (1) below.
  - Once the schema is on `X-1`, they should use a separate tool that "converts/syncs" data from the `X-1` dbdeploy managed DB to a liquibase managed DB. More on what this tool does in (2) below
  - Users should then point GoCD to the new database and upgrade from `X-1` to `X`, this upgrade should complete without errors. As far as db migrations are concerned, none should execute, because this is taken care of the sync tool used in the step before.


1. GoCD team needs to ensure that that there are no DB schema changes between `X-1` and `X`. We can't possibly be changing schema while at the same time changing the tool that manages the schema :)
2. The external tool runs all liquibase migrations on an empty DB. This effectively does a sql `create table/view/index` and creates an empty DB structure. It then iterates each table in the old DB (except for dbdeploy's `changelog` table), and copies the records (in batches) into the a table in the new database. You can also import from H2 and copy it into PostgreSQL/MySQL as an example. Users can also use this tool at any point in time if they wish to migrate data between different DB backends at a later point in time.

GaneshSPatil:
#### QA for Multiple Database Support Story

- [ ] Fresh Installation with:
    - [ ] H2
    - [ ] PostgreSQL
    - [ ] MySQL

- [ ] Upgrade:
	- [ ] Upgrade from v20.4.0 to v20.5.0 with:
		- [ ] H2
		- [ ] PostgreSQL
		- [ ] Migrating H2 to PostgreSQL
		- [ ] Migrating H2 to MySQL
		- [ ] Migrating PostgreSQL to H2 (highly unlikely someone would do this, but as we support moving from one database to another, we should test this.)

- [ ] Upgrade:
	- [ ] Upgrade from versions prior to v20.4.0 (at least last 12 releases) to v20.5.0 with:


- [ ] Upgrade from versions prior to v20.4.0 to v20.6.0 (or above) (assuming there is a new DB Migration in v20.6.0):
	Lets say user is upgrading from v20.1.0, this test will verify, the data migrates from `v20.1.0` to `v20.5.0` and then to `20.6.0`.
	Under the hood, it will do following things in order:
	1. Perform DB migrations from `v20.1.0` to `v20.4.0` on the existing database using `gocd-db-migrator` tool.
	2. Sync data from old database to new databse using `gocd-db-migrator` tool.
	3. Once the server is started with the new database, GoCD Server will perform new migrations using liquibase.

db-support/db-support-base/src/test/java/com/thoughtworks/go/db/ConnectionManagerTest.java
```diff
@@ -13,8 +13,14 @@
...
-    void migrate();
+package com.thoughtworks.go.db;
+
+import org.junit.jupiter.api.Test;
+
+class ConnectionManagerTest {
+    @Test
+    void foo() {
+
+    }
```
GaneshSPatil, line 25: Fix test!

db-support/db-support-h2/src/main/java/com/thoughtworks/go/server/database/h2/UpdateLastTransitionedTim.java
```diff
@@ -22,28 +23,24 @@
 import java.sql.SQLException;
 import java.sql.Timestamp;
 
-/**
- * @understands
- */
-public class Migration_230007 implements Trigger {
+public class UpdateLastTransitionedTim implements Trigger {
```
GaneshSPatil, line 26: rename `UpdateLastTransitionedTim` to `UpdateLastTransitionedTime`

kritika-singh3:
Testing with postgres (Rebased the PR branch with master)
 - [ ] Fresh installation with postgres db failed - the dashboard shows the msg `Dashboard is being processed, this may take a few seconds. Please check back later.`

 - [ ] Ran the pipelines with the default h2 db (which was named as `cruise.mv.db`). Migrated the said db using the db migrator tool to a postgres db. No data was migrated. Only schemas.
Also, the server when brought up with thee migrated postgres db - the dashboard showed `Dashboard is being processed, this may take a few seconds. Please check back later.`. Later performed a full config save and then the dashboard was visible.

Issue #7609, opened 2020-01-15
Imrprovements to Pipeline SPA


kritika-singh3:
##### Issue Type

- Bug Report and SPA enhancement

##### Summary

Parent Issue: #7551 
Possible improvements to Pipeline SPA 

I. Move Pipeline Operation

 - [ ] The move pipeline modal currently shows a dropdown listing all the available pipeline groups. Will it be better to show an input field with auto-complete suggestions. What is the entered value doesn't exist?
 - [ ] In case there exist only one pipeline group (or have permission to view only one), what should be the expected behaviour?
   - Disable move button with hint `No other pipeline group exists to move this pipeline`?
   - Let the user enter a name and we create the specified group? What should be the behaviour for a `pipeline group admin` user in such scenario?

II. Clone Pipeline Operation

 - [ ] The clone pipeline modal currently shows an input field for specifying the new group name. It can be upgraded to show auto-complete suggestions.
 - [x] Currently, if a `pipeline group admin` user attempts to move the pipeline to a group `grp`, following things can happen:
    1. If `grp` exists and the user has access to it, the clone is successful.
    2. If `grp` exists but the user does not have access to it, the clone fails with error: `You are not authorized to perform this action.`. The pipeline create API returns a `403`.
    3. If `grp` does not exists, the clone fails with error: `Pipeline group with name 'grp' was not found!`. The pipeline create API returns a `404`.

     Shouldn't case 2 and 3 return the same thing. In both scenario it should be `403` as the user does not have permission to either access the group or to create a new group?
     Edit: Taken care via #7656 in 20.1.0

PR #7656, opened 2020-01-24
Pipeline Grp Admin should receive a 403 on create if group doesnot exist

kritika-singh3:
Issue: #7609 (See point 2 in II)

Description:
 - the before filter on create pipeline config api - checks whether or not the current user is an admin of the given group (if user is not a system admin) - which throws RecordNotFoundException if the group does not exist.
  Updated the check to include the group existence test before admin check

### DIFF
--- a/spark/spark-base/src/main/java/com/thoughtworks/go/spark/spring/AbstractAuthenticationHelper.java
+++ b/spark/spark-base/src/main/java/com/thoughtworks/go/spark/spring/AbstractAuthenticationHelper.java
@@ -37,6 +37,7 @@ import spark.Response;
 import java.util.List;
 
 import static com.thoughtworks.go.config.exceptions.EntityType.Pipeline;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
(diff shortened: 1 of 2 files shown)


spark/spark-base/src/main/java/com/thoughtworks/go/spark/spring/AbstractAuthenticationHelper.java
```diff
@@ -181,7 +181,8 @@ public void checkPipelineCreationAuthorizationAnd403(Request request, Response r
             throw new UnprocessableEntityException("Pipeline group must be specified for creating a pipeline.");
         } else {
             String groupName = group.getAsString();
-            if (StringUtils.isNotBlank(groupName) && !securityService.isUserAdminOfGroup(currentUsername(), groupName)) {
+            if (StringUtils.isNotBlank(groupName) &&
+                    (!goConfigService.groups().hasGroup(groupName) || !securityService.isUserAdminOfGroup(currentUsername(), groupName))) {
```
GaneshSPatil: indent?
May be extract into a variable?

Issue #7654, opened 2020-01-23
Server upgrade to v19.12.0 fails when pipeline label template contains package material name


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Upgrading server to `v19.12.0` fails when the pipeline label template contains package material name specified in format `${PACKAGE:REPO}`

###### Basic environment details

* Go Version: `v19.12.0`

##### Steps to Reproduce

1. Install GoCD Server `v19.11.0`
2.  Install any Package Repository Plugin. (Say `go-npm-poller`)
3. Define a Package Repository by visiting `go/admin/package_repositories/new` (Say `npm`)
4. Define a package material using `npm` repository. (Say `my-package`)
5. Define the pipeline label using package material name as `${npm:my-package}`
6. Verify configurations.
7. Upgrade GoCD Server to `v19.12.0`.
8. The server start fails with error `Error: You have defined a label template in pipeline 'up42' that refers to a material called 'npm:my-package', but no material with this name is defined.`

##### Expected Results
GoCD Server `v19.12.0` should start with configurations which were previously valid.

##### Actual Results
The server start up fails with error: `Error: You have defined a label template in pipeline 'up42' that refers to a material called 'npm:my-package', but no material with this name is defined.`

##### Possible Fix
Add an xml migration to migrate the labeltemplate from `${npm:my-package}` to `${npm_my-package}`

##### Additional Information

<details><summary>Use the following cruise-config.xml to reproduce this issue:</summary>
<p>

```xml
<?xml version="1.0" encoding="utf-8"?>
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="135">
  <server agentAutoRegisterKey="323040d4-f2e4-4b8a-8394-7a2d122054d1" webhookSecret="3d5cd2f5-7fe7-43c0-ba34-7e01678ba8b6" commandRepositoryLocation="default" serverId="60f5f682-5248-4ba9-bb35-72c92841bd75" tokenGenerationKey="8c3c8dc9-08bf-4cd7-ac80-cecb3e7ae86c">
    <security>
      <authConfigs>
        <authConfig id="9cad79b0-4d9e-4a62-829c-eb4d9488062f" pluginId="cd.go.authentication.passwordfile">
          <property>
            <key>PasswordFilePath</key>
            <value>config/password.properties</value>
          </property>
        </authConfig>
[...]

PR #7659, opened 2020-01-24
Add schema migration to migrate pipeline label template consisting of package repo material name (#7654)

GaneshSPatil:
Issue: #7654

Description:

- As part of GoCD release v19.12.0 package repository material name format has
changed from 'repo:package' to 'repo_package'.

- Perform config xml schema migration to change such material name references
from pipeline label.

---

- Migrate `${repo:package}` to `${repo_package}`
- Do not migrate truncating colon `${svn[:7]}`
- Do not migrate env vars `${env:var}`
- Do not migrate env vars consisting colon in them `${env:my:fancy:env}`
- Do not migrate normal colons `release:v1-${github}`
- Do not migrate parameters `#{my:fancy:param}`

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/domain/PipelineLabelTest.java
+++ b/common/src/test/java/com/thoughtworks/go/domain/PipelineLabelTest.java
@@ -26,9 +26,7 @@ import com.thoughtworks.go.helper.ModificationsMother;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 
 import static com.thoughtworks.go.domain.InsecureEnvironmentVariables.EMPTY_ENV_VARS;
 import static com.thoughtworks.go.domain.label.PipelineLabel.defaultLabel;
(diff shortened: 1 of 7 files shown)


common/src/test/java/com/thoughtworks/go/domain/PipelineLabelTest.java
```diff
@@ -282,6 +280,34 @@ public void shouldReturnEmptyIfThereIsNoMatchingEnvironmentVariable() {
...
+                Arrays.asList("${repo:package}", "${repo_package}"),
+                Arrays.asList("start-${repo:package}-end", "start-${repo_package}-end"),
+
+                Arrays.asList("release:v1-${repo:package}", "release:v1-${repo_package}"),
+                Arrays.asList("release:v1-${repo:package}:${svn}", "release:v1-${repo_package}:${svn}"),
+                Arrays.asList("release:v1-${repo:package}:${github[:7]}:${svn}", "release:v1-${repo_package}:${github[:7]}:${svn}"),
+                Arrays.asList("release:v1-${repo:package[:5]}:${github[:7]}:${svn}", "release:v1-${repo_package[:5]}:${github[:7]}:${svn}"),
+                Arrays.asList("start-release:v1-${repo:package}:${github[:7]}:${svn}-end", "start-release:v1-${repo_package}:${github[:7]}:${svn}-end"),
+
+                Arrays.asList("${env:var}", "${env:var}"),
```
GaneshSPatil: @arvindsv @maheshp 

As part of GoCD release `v18.10.0`, we've added support to specify pipeline label as Environment Variable with pattern `${env:env_var_name}`. But this pattern conflicts with package repository material with name `env`.

So lets say,

We've defined a package repository with name `env` and a package with name `my_package` and specified the label template of pipeline  as `${env:my_package}`. Even though, the package-repository material name is specified, it won't be resolved. Because, as per our implementation, we always resolve `${env:my_package}` to environment variable.

Hence, in this scenario, pipeline label would be set to empty as there is no env-var with name `my_package` is specified.

This bug is only applicable for GoCD releases from `v18.10.0` to `v19.11.0` as the pipeline label pattern for package material has changed to `${repo_package}` in GoCD `v19.12.0`

**For GoCD releases from `v18.10.0` to `v19.11.0`, we should say, for patterns `${env:something}`, `env` is a reserved GoCD keyword and will always be resolved to environment variable.**

With the above-mentioned behavior, for pipeline-label migrations containing `${env:something}`, the pattern template will be treated as environment variable pattern and will not migrate `:` to `_`.

config/config-server/src/main/resources/upgrades/135.xsl
```diff
@@ -0,0 +1,42 @@
...
+    <!-- Copy everything -->
+    <xsl:template match="@*|node()">
+        <xsl:copy>
+            <xsl:apply-templates select="@*|node()"/>
+        </xsl:copy>
+    </xsl:template>
+
+    <xsl:template name="migratePipelineLabel">
+        <xsl:param name="templateValue"/>
+        <xsl:value-of select="migratorClass:migratePipelineLabelTemplate($templateValue)"/>
```
GaneshSPatil: @maheshp -- This `xsl` invokes java code for migrations. I couldn't manage to write an integration spec for this behavior. 
Hence, We would need to test this manually. :(

common/src/test/java/com/thoughtworks/go/domain/PipelineLabelTest.java
arvindsv: I understand. Makes sense to me.

Issue #7605, opened 2020-01-14
Add rules to the config repo


kritika-singh3:
##### Issue Type

- Feature enhancement

##### Summary

Epic Issue: #5175 (search for `Better authorization`), #3636 

Proposal: Add `Rules` to the config repo to allow/deny creation/access to environments, pipeline groups and pipelines.

The rules defined for that config repo will govern what all entities can be added/referred to the GoCD. Proposed actions are:

| Action | Entity Type               | Description        |
|--------|--------------------|-----------------|
| create | environment/pipeline_group/pipeline | allow/deny ability to create a new environment/pipeline group/pipeline with the specified pattern |
| refer  | environment        | allow/deny pipelines(and agents?) to be added to the specified environment                        |
| refer  | pipeline_group    | allow/deny pipelines to be added to the specified pipeline group                                  |
| refer  | pipeline                | allow/deny the specified pipeline to be used as an upstream dependency                      |

Note: <strong>`Deny`</strong> rule will take precedence.

Example of the config:

```xml
<config-repo pluginId="json.config.plugin" id="json">
    <git url="/tmp/config-repo" />
    <rules>
        <allow action="refer" type="environment">env_*</allow>
        <allow action="create" type="pipeline_group">grp_*</allow>
      </rules>
</config-repo>
```

##### Need to decide:
 - Should the agents addition to environment be restricted as well?
 - Will the default rule to be `deny` all? The existing GoCD installations using config-repos may break if so. (Maybe we can do a config migration to get around this?)
 - Should we consider the rules while checking the definitions (via repoId) on [preflight API](https://api.gocd.org/current/#preflight-check-of-config-repo-configurations)?

#### More questions based on comments
 - Will there be a need to modify RBAC for config-repo?

#### Possible actions

| Action | Entity Type               | Description        |
|--------|--------------------|-----------------|
| refer  | environment        | allow/deny pipelines and agents to be added to the specified environment                      |
| refer  | pipeline_group    | allow/deny pipelines to be added to the specified pipeline group                              |
| refer  | pipeline  | allow/deny the specified pipeline to be used as an upstream dependency                  |

#### Additional Info

 - <strong>`deny`</strong> rule will take precedence.
 - Should the agents addition to environment be restricted as well?: **Yes**
[...]

PR #7663, opened 2020-01-28
Added config migration to support rules in config repo

kritika-singh3:
Issue: #7605 

Description:
 - Added support for `rules` to the config repo to allow/deny creation/access to environments, pipeline groups and pipelines.
 - Since the general rule will be `deny all`, added a default `allow all` rule to all existing config repos in the migration.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
@@ -97,7 +97,7 @@ public class ConfigRepoConfig extends RuleAwarePluginProfile implements Cacheabl
     }
 
     private boolean isValidMaterial() {
-        MaterialConfig materialConfig = getMaterialConfig();
+        MaterialConfig materialConfig = getRepo();
         if (materialConfig == null) {
             return false;
         }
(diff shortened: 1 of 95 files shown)


config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
```diff
@@ -142,29 +88,8 @@ private void validateUniquenessOfId(ValidationContext validationContext) {
         }
     }
 
-    public boolean validateTree(ValidationContext validationContext) {
-        validate(validationContext);
-        boolean isValid = errors.isEmpty();
-
-        if (getMaterialConfig() != null) {
-            isValid = getMaterialConfig().errors().isEmpty() && isValid;
```
maheshp: @kritika-singh3 where has this check moved? This check ensured a config repo is invalid if the material config within is invalid.

kritika-singh3: @maheshp The check for `materialConfig` is present in `validate(ValidationContext )` [method](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java#L55-L65). The same will get called from the `validateTree` [method](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/rules/RuleAwarePluginProfile.java#L127-L136) from `RuleAwarePluginProfile` class.

PR #7669, opened 2020-01-28
Remove all x509 related code

ketan:
- The server will no longer generate any self signed SSL certificates.
  It will however initialize the jetty SSL connector if the keystore is
  available.
- Agents, for quite some time have not been using mutual TLS, this
  removes all code that generates the agent certs/private keys and
  transmits the keys/certs and stores it in various
  keystores/truststores


jetty9/src/main/java/com/thoughtworks/go/server/Jetty9Server.java
```diff
@@ -76,7 +76,10 @@ public Jetty9Server(SystemEnvironment systemEnvironment, String password) {
     public void configure() throws Exception {
         server.addEventListener(mbeans());
         server.addConnector(plainConnector());
-        server.addConnector(sslConnector());
+        Connector connector = sslConnector();
+        if (connector != null) {
+            server.addConnector(connector);
+        }
```
ketan: @arvindsv and @maheshp is this OK for an implementation? Basically don't configure ssl on fresh installs.

arvindsv: @ketan My opinion: I'm happy to replace these kinds of things with two pieces of documentation / help:

1. Upon an upgrade, what do I need to do to keep everything the way it was? [Probably nothing in this case?]

2. What do I need to do and in what order to get a new installation to use TLS termination between the agents and server, if I want.

3. Optional: Warning users that it is not secure to run in an untrusted/open network, without doing what point 2 says.

As long as that is clear, I think it's fine.

ketan:
The goal of this PR is to simply the SSL setup within GoCD, all the while allowing GoCD system admins to setup SSL in a manner that is simple and secure. To this end:

GoCD server's jetty instance will (in a few months from now) no longer listen on a SSL/TLS endpoint (port 8154). There will be a way to be backward compatible for system admins to slowly (re)configure their agents, if they desire to use SSL/TLS for agents along with mutual TLS.

To setup TLS for GoCD, system admins will be required to front the GoCD server with a reverse proxy that supports TLS (like apache, NGINX). Any existing agents that are using TLS, can connect to this reverse proxy. Reverse proxies have the advantage that they make it a lot easier and convenient to setup and configure various TLS connection parameters.

For system admins that want to additionally setup mutual TLS between the server and agents, GoCD agents will allow configuring a private key, certificate and an optional passphrase, which the agent will use to authenticate with the reverse proxy.

To keep backward compatibility and minimize disruption:

- New versions of GoCD (>= 20.2) will not bring up TLS port 8154. There will be no way for such users to configure built-in TLS. A reverse proxy will need to be the one terminating TLS if needed.
- Users upgrading to GoCD (20.2-20.4) will notice that the server does not bring up TLS on port 8154, but emits a warning indicating how they can turn it on, and go back to how it was earlier. The warning message will include a timeline when the toggle will go away, along with instructions on how to set it up correctly using a reverse proxy.
- Version 20.5 of GoCD server will no longer contain any functionality to configure TLS. At this point, users will be expected to terminate SSL elsewhere.

PR #7675, opened 2020-01-29
Fix read only templates view styles on the pipeline config edit page

GaneshSPatil:
#### Description:
* Old application styles are applied to the templates read only view,
  when it is rendered on the pipeline config edit page.

* Explicitly define required styles on the templates read only view.

##### Before
Screen Shot 2020-01-29 at 2 51 22 PM

##### After
Screen Shot 2020-01-29 at 2 49 41 PM


GaneshSPatil:
**Note:** The styles will not be 100% identical with the one rendered on the templates page, as on the old `application.scss` styles, the default styles of the html elements are modified. There will be a subtle difference in margin/padding. 
Example: Check `Show job: JOB_NAME` header on both the pages.

Issue #3103, opened 2017-01-18
Missing parameter in dependent pipeline leads to unknown error in template configuration


lenucksi:
##### Issue Type
- Bug Report

##### Summary
I have pipelines **A, B** and **X** which depend on template **T**.
X defines parameter _foo_, A and B do not.
T uses foo in its resource section, it defines, "#{foo}, git" as required resources. (As described in  https://docs.gocd.io/current/configuration/admin_use_parameters_in_configuration.html
However, upon saving only a red bar appears with "Save failed, see errors below” appears on top with no errors shown below.
An attempt to save an explicitly wrong resource  such as  ‘#{foo$}’ gets the expected "Resource name ‘#{foo$}' is not valid. Valid names can contain valid parameter syntax or valid alphanumeric with hyphens,dots or pipes” below and the red bar above.

As suggested by @varshavaradarajan and @jyotisingh this appears to be due to the parameter not being present on A and B, i.e. not on all dependent pipelines. 
The actual is error thus occurs on a dependent pipeline and is not propagated to the template view.

This has also been discussed on Gitter here: https://gitter.im/gocd/gocd?at=587f82a861e516c157916bda

###### Basic environment details
* Go Version: 16.11
* JAVA Version: OpenJDK 7
* OS: Linux

##### Steps to Reproduce

Described above in summary

##### Expected Results
Pipelines (in the example A and B) which do not specify the parameter (in the example T) would have it defaulted to “blank"
There is a ticket suggesting this here: https://github.com/gocd/gocd/issues/829

##### Actual Results
See above

##### Code snippets/Screenshots
http://pasteboard.co/nBhf876Kq.png

arvindsv:
Closing as fixed through https://github.com/gocd/gocd/pull/7688

PR #7688, opened 2020-02-02
Show parameter related resource errors while editing pipeline and template config 

varshavaradarajan:
Issue: #3103

Description:

The pipeline config edit view uses the entity config save flow which copies errors from a preprocessed config to the pipeline config object being updated. We had missed copying errors on resource configs before.

The template config edit uses the old config save flow and does not show a specific error when non existent params are used as resources. If non-existent params were used elsewhere, like environment variables, the exception `Parameter x is not defined is shown`. This is fixed by displaying the global errors that are created as a result of the param exception in the old config save flow.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/PipelineConfigErrorCopier.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/PipelineConfigErrorCopier.java
@@ -56,6 +56,7 @@ public class PipelineConfigErrorCopier {
                 copyCollectionErrors(fromJob.artifactTypeConfigs(), toJob.artifactTypeConfigs());
                 copyCollectionErrors(fromJob.getTabs(), toJob.getTabs());
                 copyCollectionErrors(fromJob.getVariables(), toJob.getVariables());
+                copyCollectionErrors(fromJob.resourceConfigs(), toJob.resourceConfigs());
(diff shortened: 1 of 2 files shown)


arvindsv:
Verified on 20.2.0 (11308-05dc4d14977d4bf10d5f5c9e9d6f88817df1ec1e). Looks alright to me.

### Config used:

```xml
<pipelines group="secondGroup">
  <pipeline name="p2" template="t1">
    <params>
      <param name="RESOURCE">r1</param>
    </params>
    <materials>
      <git url="/tmp/test.git" />
    </materials>
  </pipeline>
</pipelines>
<templates>
  <pipeline name="t1">
    <stage name="s2">
      <jobs>
        <job name="j2">
          <tasks>
            <exec command="ls" />
          </tasks>
          <resources>
            <resource>#{RESOURCE}</resource>
            <resource>my-other-resource</resource>
          </resources>
        </job>
      </jobs>
    </stage>
  </pipeline>
</templates>
```

### Scenario:

Change the resource in job config to a non-existent resource.

### Before

In 20.1.0 (11114-f2c1771e333ea92dae82cd78d2efa1310c1086f0):



### After

In 20.2.0 (11308-05dc4d14977d4bf10d5f5c9e9d6f88817df1ec1e):

PR #7670, opened 2020-01-29
Introduce API endpoint for retrieving parameters of a template

kritika-singh3:
Issue: #3194

Description:
 - added an endpoint `/go/api/admin/templates/:template_name/parameters` which returns a list of parameters that needs to be defined while using it to create a pipeline

curl:

```curl
  curl "http://go-server-url/go/api/admin/templates/template-name/parameters" \
     -u 'username:password' \
     -H 'Accept: application/vnd.go.cd.v7+json'
```
Response:

```json
{
  "_links" : {
    "self" : {
      "href" : "http://go-server-url/go/api/admin/templates/template-name"
    },
    "doc" : {
      "href" : "https://api.gocd.org/#template-config"
    },
    "find" : {
      "href" : "http://go-server-url/go/api/admin/templates/:template_name"
    }
  },
  "name" : "template-name",
  "parameters" : [ "resources", "command" ]
}
```

PR #7696, opened 2020-02-03
Set etag based on parameters and not on template config

kritika-singh3:
Setting etag on parameters api


arvindsv:
Marking this as verified on 20.2.0 (11308-05dc4d14977d4bf10d5f5c9e9d6f88817df1ec1e) based on this being tested here: https://github.com/gocd/gocd/pull/7670#issuecomment-597034987 [Scenarios 2 and 3].

cc: @maheshp

PR #7697, opened 2020-02-03
Applied rules after a successful config repo parse

kritika-singh3:
Issue: #7605 

Description:
 - The PartialConfig is validated against the rules defined in the config repo while updating the config in the PartialConfigUpdateCommand.
 - In case of pre-flight api where no config-repo id is specified, the check is skipped.

Preview:

Screen Shot 2020-02-03 at 5 20 54 PM

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java
@@ -16,8 +16,18 @@
 package com.thoughtworks.go.config.update;
 
 import com.rits.cloning.Cloner;
+import com.thoughtworks.go.config.ErrorCollector;
 import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.config.materials.dependency.DependencyMaterialConfig;
+import com.thoughtworks.go.config.remote.ConfigRepoConfig;
 import com.thoughtworks.go.config.remote.PartialConfig;
+import com.thoughtworks.go.plugin.access.configrepo.InvalidPartialConfigException;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.thoughtworks.go.config.rules.SupportedEntity.*;
+import static java.lang.String.format;
 
 public class PartialConfigUpdateCommand implements UpdateConfigCommand {
     private static final Cloner CLONER = new Cloner();
@@ -25,11 +35,13 @@ public class PartialConfigUpdateCommand implements UpdateConfigCommand {
     private final PartialConfig partial;
     private final String fingerprint;
     private final PartialConfigResolver resolver;
+    private final ConfigRepoConfig configRepoConfig;
 
-    public PartialConfigUpdateCommand(final PartialConfig partial, final String fingerprint, PartialConfigResolver resolver) {
+    public PartialConfigUpdateCommand(final PartialConfig partial, final String fingerprint, PartialConfigResolver resolver, ConfigRepoConfig configRepoConfig) {
         this.partial = partial;
         this.fingerprint = fingerprint;
         this.resolver = resolver;
+        this.configRepoConfig = configRepoConfig;
     }
 
     @Override
@@ -41,7 +53,12 @@ public class PartialConfigUpdateCommand implements UpdateConfigCommand {
                 cruiseConfig.getPartials().remove(config);
             }
 
-            cruiseConfig.getPartials().add(CLONER.deepClone(partial));
+            PartialConfig cloned = CLONER.deepClone(partial);
+            cruiseConfig.getPartials().add(cloned);
+
+            if (!validateEntityForRules(cloned)) {
+                return cruiseConfig;
+            }
 
             for (PartialConfig partial : cruiseConfig.getPartials()) {
                 for (EnvironmentConfig environmentConfig : partial.getEnvironments()) {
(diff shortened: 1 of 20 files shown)


kritika-singh3:
Preflight API Preview:

```curl
curl "https://go-server-url/go/api/admin/config_repo_ops/preflight?pluginId=json.config.plugin&repoId=json" \
      -u 'username:password' \
      -H 'Accept: application/vnd.go.cd+json' \
      -F 'files[]=@test.goenvironment.json' \
      -F 'files[]=@test.gopipeline.json'
```

```json
{
  "valid": false,
  "errors": [
    "Cannot refer environment: 'testing123' from the config repository.",
    "Cannot refer pipeline: 'up42' from the config repository.",
    "Cannot refer pipeline group: 'test' from the config repository."
  ]
}
```

arvindsv:
Would it be useful to explain _why_? Meaning: Due to permissions / rules for this repo?

kritika-singh3:
@arvindsv  What about this message:
`Rules validation failure: Cannot refer environment: 'testing123' from this config repository.`?

cc:\ @GaneshSPatil @maheshp

arvindsv:
I think it's better. Another approach would be to say: "Not allowed to refer ..."

kritika-singh3:
Something like this:
`Not allowed to refer environment: 'testing123' from this config repository.`?

arvindsv:
Yes. Then, we won't need to say "Rules validation failure".

GaneshSPatil:
> Something like this:
> `Not allowed to refer environment: 'testing123' from this config repository.`?

may be without `:` (colon)?
`Not allowed to refer environment 'testing123' from this config repository. `

arvindsv:
Yes.

Another aspect to look into: If I refer to another pipeline in a different config repo (which is also not allowed), it should show a different message.

adityasood:
@kritika-singh3 @GaneshSPatil would it make sense to let them know what they can do if they see this error?  like contact sys-admin ?

adityasood:
Also keep the sentence like " Not allowed to refer environments/pipelines/pipeline groups " assuming it is possible to have more than one of those referred?

GaneshSPatil:
> Yes.
> 
> Another aspect to look into: If I refer to another pipeline in a different config repo (which is also not allowed), it should show a different message.

When we say different message, what it should be?

Something like `Not allowed to refer pipeline  'upstream' (defined in `blah` repository) from `my-repo` config repository. ` ?

This would require us to find the origin of each referenced pipeline.

GaneshSPatil:
> @kritika-singh3 @GaneshSPatil would it make sense to let them know what they can do if they see this error? like contact sys-admin ?
> Also keep the sentence like " Not allowed to refer environments/pipelines/pipeline groups " assuming it is possible to have more than one of those referred?

Hmm. This is something we can consider doing.

Currently, we are not differentiating between config-errors and reference-errors. Both the errors are shown as validation failures.

We possibly could split these errors into two sections:
- Config Errors (Represent them as they're represented today)
- Reference Errors (and show the message??)

\cc: @arvindsv @kritika-singh3

server/src/main/java/com/thoughtworks/go/config/update/PartialConfigUpdateCommand.java
```diff
@@ -58,4 +75,44 @@ public CruiseConfig update(CruiseConfig cruiseConfig) {
...
+                .forEach(pipelineGrpThatCannotBeReferred -> {
+                    pipelineGrpThatCannotBeReferred.addError(PIPELINE_GROUP.getType(), format("Cannot refer pipeline group: '%s' from the config repository.", pipelineGrpThatCannotBeReferred.getGroup()));
+                });
+
+        List<DependencyMaterialConfig> dependencyMaterialConfigs = new ArrayList<>();
+        partialConfig.getGroups().forEach((pipelineGrp) -> {
+            pipelineGrp.forEach((pipelineConfig) -> dependencyMaterialConfigs.addAll(pipelineConfig.dependencyMaterialConfigs()));
+        });
+        dependencyMaterialConfigs.stream()
+                .filter(dependencyMaterialConfig -> !doesPipelineExistInPartialConfig(partialConfig, dependencyMaterialConfig.getPipelineName()))
```
GaneshSPatil: @kritika-singh3 @arvindsv @adityasood 

**Pipeline reference rule defined on a config repository is not applicable for the pipelines coming from the same repository.**

Description:

For the following rule:
`<allow action="refer" type="pipeline">up42</pipeline>`

Following partials are coming from config repository:
* `pipeline1` (which has a pipeline-dependency `up42`)
* `pipeline2` (which has a pipeline-dependency `pipeline1`)

Even though, there is no explicit allow permission mentioned for `pipeline1` pipeline dependency, it is allowed because both the pipelines are defined in the same repository.

**Note:** We need to mention this behavior on our docs explicitly.

arvindsv:
> Something like Not allowed to refer pipeline 'upstream' (defined in blahrepository) frommy-repoconfig repository. ?

No. I just mean that they should be able to differentiate between a permissions issue and a reference issue like the case I mentioned. If that means adding "Rules validation error" more prominently, so be it.

kritika-singh3:
> Also keep the sentence like " Not allowed to refer environments/pipelines/pipeline groups " assuming it is possible to have more than one of those referred?

@adityasood We add an error for every pipeline/pipeline group/environment that cannot be referred.

Screen Shot 2020-02-04 at 2 39 22 PM

kritika-singh3:
@arvindsv @GaneshSPatil Currently this is the error message that comes up if the pipeline does not exist and it is not allowed to be referred:

Screen Shot 2020-02-04 at 3 00 19 PM

arvindsv:
@kritika-singh3 Cool. Seems different enough to me. Though we should fix the `&quot;`.

GaneshSPatil:
@kritika-singh3 - The messages looks okay to me.

Even though for the same upstream dependency, there are two different errors, both of these are valid errors, one for validation and another for reference.

kritika-singh3:
Great. Will fix that `&quot;`.

maheshp:
> We possibly could split these errors into two sections:
> - Config Errors (Represent them as they're represented today)
> - Reference Errors (and show the message??)

@GaneshSPatil @kritika-singh3 I think it makes sense to show in 2 separate sections.

kritika-singh3:
@maheshp @GaneshSPatil @arvindsv Something like this:

I. Error on config repo:
   Screen Shot 2020-02-05 at 11 22 41 AM

II. Error in server health messages:

   Screen Shot 2020-02-05 at 11 22 56 AM

kritika-singh3:
@maheshp @arvindsv @GaneshSPatil Something like this:

I. Config Repo SPA

Screen Shot 2020-02-05 at 11 27 08 AM

II. Server health messages

Screen Shot 2020-02-05 at 11 27 14 AM

Issue #6031, opened 2019-03-27
better pipeline material errors


mbotmcc:
##### Issue Type

- Feature proposal

##### Summary

We have a few projects sharing the same repo. Recently I've tried adding a new pipeline material and get the following error.... 

>  _is specified more than once in the configuration with different values for the autoUpdate attribute. All copies of this material must have the same value for this attribute

._But there's no mention what pipeline has a different value for autoupdate. This means I now need to go through all the repos and check where it might be. It'd be good if the error included the pipeline which has the different autoupdate setting

arvindsv:
Makes sense. No one is working on this area, @mbotmcc, as far as I know. We should pick this up when re-doing pipeline config.

For the record, this will need some changes around [MaterialConfigs.java#L267](https://github.com/gocd/gocd/blob/2be3951f5dc2b2ce07f5ddd34a314305834550ca/config/config-api/src/main/java/com/thoughtworks/go/config/materials/MaterialConfigs.java#L267) and then [ConfigSaveValidationContext.java#L254](https://github.com/gocd/gocd/blob/2be3951f5dc2b2ce07f5ddd34a314305834550ca/config/config-api/src/main/java/com/thoughtworks/go/config/ConfigSaveValidationContext.java#L254) and possibly [PipelineConfigSaveValidationContext.java#L230](https://github.com/gocd/gocd/blob/2be3951f5dc2b2ce07f5ddd34a314305834550ca/config/config-api/src/main/java/com/thoughtworks/go/config/PipelineConfigSaveValidationContext.java#L230).

PR #7708, opened 2020-02-05
Improving error message on autoUpdate mismatch

kritika-singh3:
Issue: #6031

Description:
 - showing the list of pipelines which uses the material in question (irrespective of autoUpdate value)
 - also showing the id of the config repository - if used

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/ConfigSaveValidationContext.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/ConfigSaveValidationContext.java
@@ -27,6 +27,7 @@ import com.thoughtworks.go.domain.packagerepository.PackageRepository;
 import com.thoughtworks.go.domain.scm.SCM;
 
 import java.util.HashMap;
+import java.util.Map;
 
 /**
  * @understands providing right state required to validate a given config element
@@ -281,6 +282,19 @@ public class ConfigSaveValidationContext implements ValidationContext {
         }
         MaterialConfigs matchingMaterials = fingerprintToMaterials.get(fingerprint);
         return matchingMaterials == null ? new MaterialConfigs() : matchingMaterials;
+    }
+
+    @Override
+    public Map<CaseInsensitiveString, Boolean> getPipelineToMaterialAutoUpdateMapByFingerprint(String fingerprint) {
+        Map<CaseInsensitiveString, Boolean> map = new HashMap<>();
+        getCruiseConfig().getAllPipelineConfigs().forEach(pipeline -> {
+            pipeline.materialConfigs().stream()
+                    .filter(materialConfig -> materialConfig.getFingerprint().equals(fingerprint))
+                    .findFirst()
+                    .ifPresent(expectedMaterialConfig -> map.put(pipeline.name(), expectedMaterialConfig.isAutoUpdate()));
+
+        });
+        return map;
     }
 
     private void primeForMaterialValidations() {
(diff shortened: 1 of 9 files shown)


kritika-singh3:
Preview:

Editing material:

Screen Shot 2020-02-05 at 1 03 14 PM

Screen Shot 2020-02-05 at 1 03 58 PM

Adding config-repo for a material which has `autoUpdate` set to `false`:

Screen Shot 2020-02-05 at 1 04 50 PM

arvindsv:
I know the message is already big, but it is unclear exactly what is happening. For instance, it talks about `autoUpdate` attribute. Maybe we can clarify that it is "Poll for new changes" as well.

How about: 

```
This material is used elsewhere with a different value for autoUpdate ("Poll for changes"). Those values should be the same. Other places it is used in are:

This pipeline: Auto update disabled
Config repository "sdsd": Auto update enabled
Pipeline "up42": Auto update enabled
```

... or something similar?

arvindsv:
This material is used elsewhere with a different value for autoUpdate ("Poll for changes"). Those values should be the same. This materail: auto update enabled. Pipelines: up42 (auto update disabled), up43 (auto update disabled). Config repository: sdsd (auto update disabled).

kritika-singh3:
Update message:

Screen Shot 2020-02-05 at 5 15 20 PM

Screen Shot 2020-02-05 at 5 11 20 PM

config/config-api/src/main/java/com/thoughtworks/go/config/ConfigSaveValidationContext.java
```diff
@@ -283,6 +284,19 @@ public MaterialConfigs getAllMaterialsByFingerPrint(String fingerprint) {
         return matchingMaterials == null ? new MaterialConfigs() : matchingMaterials;
     }
 
+    @Override
+    public Map<CaseInsensitiveString, Boolean> getPipelinesWithMaterial(String fingerprint) {
```
bdpiparva: Don't you think method name is mis-leading?

config/config-api/src/main/java/com/thoughtworks/go/config/ValidationContext.java
```diff
@@ -84,5 +88,9 @@ default SystemEnvironment systemEnvironment() {
     default PolicyValidationContext getPolicyValidationContext() {
         return null;
     }
+
+    default Map<CaseInsensitiveString, Boolean> getPipelinesWithMaterial(String fingerprint) {
```
bdpiparva: Same as above

kritika-singh3: Couldn't think of a better name. Any suggestions?

config/config-api/src/main/java/com/thoughtworks/go/config/ConfigSaveValidationContext.java
kritika-singh3: This is the implementation of the method in `ValidationContext`

config/config-api/src/main/java/com/thoughtworks/go/config/ValidationContext.java
kritika-singh3: Updated to `getPipelineToMaterialAutoUpdateMapByFingerprint`

maheshp:
@kritika-singh3 there are couple of things which needs to be looked at,
1. This error does not show up when you try to add a new pipeline, there is just a generic validation error.
2. The usages of the material comes as part of the error string and I think it's hard to read for UX perspective. I understand that the entire error message is returned as a single string, so you think we can do something to show the usages on a separate line.

maheshp:
> This error does not show up when you try to add a new pipeline, there is just a generic validation error.

There is a PR https://github.com/gocd/gocd/pull/7859 to fix the issue on the add pipeline wizard.

> The usages of the material comes as part of the error string and I think it's hard to read for UX perspective. I understand that the entire error message is returned as a single string, so you think we can do something to show the usages on a separate line.

Fixed by - https://github.com/gocd/gocd/pull/7849

Issue #6123, opened 2019-04-10
feature propsal - scan multiple branches


thatsk:
##### Issue Type
- Feature proposal

##### Summary
gocd config repositories should support regex for scanning multiple branches.

![image](https://user-images.githubusercontent.com/15324013/55874301-92b12c80-5baf-11e9-8a93-dd40dc27853a.png)

##### Environment
gocd  19.1.0

arvindsv:
I've been playing around with an idea to support branches with pipelines as code, in my head. I'm putting it out there as a nascent idea to be shot down / improved.


Assumption 1: Some kind of plugin-level templating mechanism (as in https://github.com/gocd/gocd/issues/5675) exists in the config-repo plugins.

#### Setup:
1. Imagine a pipeline defined in YAML / JSON with a single material (for now).
2. Assume that the material is marked, in some way, to need support for branches.

#### Example:

```yaml
track_branches_for: https://www.github.com/gocd/gocd
  patterns: pull/.*/head

  template:
    {{ foreach branch in CONTEXT['branches'] }}
      pipeline-${branch}:
        tag: mytag1
        group: mygroup
        materials:
          mygit:
            git: https://www.github.com/gocd/gocd
            branch: ${branch}
        stages:
          - stage1:
            ...
          
      pipeline-downstream-${branch}:
        tag: mytag1
        group: mygroup
        materials:
          mygit:
            git: https://www.github.com/gocd/gocd
            branch: ${branch}
          myupstream:
            pipeline: pipeline-${branch}
            stage: stage1
        stages:
          mystage:
            ...
    {{ end }}
```

#### Work:
  - The plugin would need to be aware of this and be able to understand branches + filter them.
  - The plugin would need to provide that as context to the template (see assumption that there is a templating mechanism).
  - Depending on how it is implemented, once a PR is merged or a branch is deleted, the context won't include that branch and the pipeline automatically goes away from GoCD.

#### Benefits of this approach:

1. Flexibility: You can use this to build whole VSMs for branches, if you want.
2. Proper branch support: Unlike with branches and PRs using the GoCD PR plugin, there won't be only one pipeline across all branches.
3. With some appropriate use of a concept such as tags, etc. all pipelines tagged with the same "tag" can be shown together in a dashboard, for visualization reasons, to reduce the clutter of having some many pipelines per branch.
[...]

PR #7710, opened 2020-02-05
Major refactoring of webhook request and payload classes in preparation for supporting different payloads

marques-work:
Extracted behaviors and provider-specific details into a collection of mixins so that new payloads, such as pull request and branch CRUD payloads can flexibly reuse common parts via composition instead of generations of inheritance.

See issue #6123 (PR and branch support for config repos)


marques-work:
### Context

The future intent is to add `PullRequestWebhookController` as well as `GitHubPrRequest`, `GitLabPrRequest`, `BitBucketCloudPrRequest`, and `BitBucketServerPrRequest` (plus payloads) in order to support reparsing config repos when PRs are opened or closed (which may result in new pipeline configs emitted). These will live alongside the existing "push event notification" webhook controller.

I might also do the same thing for branch creation/deletion events as well.

PR #7712, opened 2020-02-06
Introduce Permissions API (#7659)

GaneshSPatil:
Issue: #7659

Description:
Add a permissions API endpoint which tells whether a user has following permissions:
  * Environments - VIEW, ADMINISTER
  * Config Repos - VIEW, ADMINISTER
  * Cluster Profiles - VIEW, ADMINISTER
  * Elastic Agent Profiles - VIEW, ADMINISTER

Request:
```
curl http://localhost:8153/go/api/auth/permissions -u view:badger -H 'Accept: application/vnd.go.cd.v1+json'
```

Response:
```json
{
  "_links": {
    "self": {
      "href": "http://localhost:8153/go/api/auth/permissions"
    },
    "doc": {
      "href": "https://api.gocd.org/20.2.0/#permissions"
    }
  },
  "permissions": {
    "environment": {
      "view": [ "DEV", "UAT", "QA", "PROD" ],
      "administer": [ "DEV" ]
    },
    "config_repo": {
      "view": [ "dev-pipelines-repo", "prod-pipelines-repo" ],
      "administer": [ "dev-pipelines-repo" ]
    },
    "cluster_profile": {
      "view": [ "dev-cluster", "prod-cluster" ],
      "administer": [ "dev-cluster" ]
    },
    "elastic_agent_profile": {
      "view": [ "build-agent", "deploy-agent"  ],
      "administer": [ "build-agent" ]
    }
  }
}
```


kritika-singh3:
`administer` implies `view` permission. Any reason why to add in both?

GaneshSPatil:
@kritika-singh3 -- This API information will be used by clients to take actions based on permissions. 
Implicit permissions would cause confusion. Hence added `view` as well.

kritika-singh3:
We are passing all information available. So both of them will be present to the client and
checks can happen sequentially.
E.g. if not `administer` permission, does the user have `view` permission.

GaneshSPatil:
> We are passing all information available. So both of them will be present to the client and
checks can happen sequentially.
E.g. if not administer permission, does the user have view permission.

@kritika-singh3 - I am sorry did not understand this comment. Can you help me understand this again?

kritika-singh3:
Sure. 
Suppose there is a requirement that say an action `A` needs to be performed if the user `u` has `view` permissions on environment `E`.
We give a call to this API and get the result. Now the result will contain a list of entities that the user has `view` permissions on (say viewList) and another list with `administer` permissions (say administerList).

So, we can check our requirement via `administerList.contains('E') || viewList.contains('E')` removing the need to send the entity twice.

GaneshSPatil:
> Sure.
Suppose there is a requirement that say an action A needs to be performed if the user u has view permissions on environment E.
We give a call to this API and get the result. Now the result will contain a list of entities that the user has view permissions on (say viewList) and another list with administer permissions (say administerList).
So, we can check our requirement via administerList.contains('E') || viewList.contains('E') removing the need to send the entity twice.

In this example, we could avoid performing `administerList.contains('E')` check completely. As whether the user has view only permission matters.

Lets take an example of implicit `view` permissions. User has view access to `prod` environment and administer access to `dev` environment. So, our response will look like:

```json
{
    "environment": {
         "view": ["prod"],
    }
}
```

kritika-singh3:
What is the url which will give me only the `view` permissions?

GaneshSPatil:
> What is the url which will give me only the view permissions?

Users will use the same API to perform CRUD operations on the entity. Regardless of this API endpoint, the end users will use the same endpoint. 

This API is just telling what permissions a user has. Also, it is not tied to the granular-auth implementation at all. This API can also include other entities which does not support granular auth and other permissions too, example: `Pipeline Group`, and the permissions for it could be `view`, `operate` and `administer`.

May be someone else has more opinion towards this?
\cc: @arvindsv @ketan

arvindsv:
I agree it makes sense to send it explicitly. That way clients of this API don't need to have knowledge about these implicit assumptions, which can also change over time.

maheshp:
@GaneshSPatil should we consider providing an ability to filter permissions for a specific entity type. 
Since this API will be used on our SPA's, based on the SPA we can fetch permissions accordingly.

GaneshSPatil:
@maheshp - Added an ability to filter permissions based on `type` request param. 

Examples:
1. `/go/api/auth/permissions` => Returns **all** permissions
2. `/go/api/auth/permissions?type=environment` => Returns **environment** permissions
3. `/go/api/auth/permissions?type=environment,config_repo` => Returns **environment** and **config_repo** permissions

PR #7689, opened 2020-02-03
Add support for rules in config repo api and spa

kritika-singh3:
Issue: #7605 

Description: Added support for `rules` in config-repo in API v3 and SPA

PR #7731, opened 2020-02-12
Code refactoring

kritika-singh3:
In PR #7689, we introduced components to showcase rules configured and to define the same.
This PR is to utilize these components on Secret Configs SPA.


kritika-singh3:
Previous:

Screen Shot 2020-02-12 at 3 46 59 PM

Now:

Screen Shot 2020-02-12 at 3 45 45 PM

Issue #7698, opened 2020-02-03
default -Xmx8G in wrapper-properties.conf has no effect on agent.jar


suside:
##### Summary
After encountering for a few times this error:
```
2020-01-24 14:48:38,619 ERROR [scheduler-2] TaskUtils$LoggingErrorHandler:95 - Unexpected error occurred in scheduled task.
java.lang.OutOfMemoryError: Java heap space
```
I decided to increase memory for the go-agent.
According to [this comment](https://github.com/gocd/gocd/blob/master/installers/include/wrapper-properties.conf.example#L36) and [performance tuning](https://docs.gocd.org/current/installation/performance_tuning.html) docs `wrapper-properties.conf` is the right place. So I removed `#` from `wrapper.java.additional.105=-Xmx8G` in `/usr/share/go-agent/wrapper-config/wrapper-properties.conf`

It turned out it has no effect.

After some digging I added `-Xms1g -Xmx2g` to `set.default.AGENT_STARTUP_ARGS` parameter in `wrapper.conf` (instead `wrapper-properties.conf`) and that fixed the `java.lang.OutOfMemoryError` problem.
I've noticed that:

1. `wrapper.java.additional.*` in `wrapper-properties.conf` works only for `agent-bootstrapper.jar`
1. `AGENT_STARTUP_ARGS` in `wrapper.conf` works for `agent.jar`, besides `Xmx` it was also tested with `-Dtoggle.agent.git.clean.keep.ignored.files`

###### Basic environment details

* Go Version: `go-agent 19.7.0-9567`
* JAVA Version: `OpenJDK 64-Bit Server VM AdoptOpenJDK (build 12.0.1+12, mixed mode, sharing)`
* OS: `Ubuntu 16.04 xenial`

###### Additional Environment Details

Agent installation was upgraded from `/etc/default/go-agent`

##### Possible Fix
I'm not familiar with java service wrapper but this line is commented out https://github.com/gocd/gocd/blob/master/installers/include/additional-properties.conf#L3

ketan:
Any JVM args you specify using `wrapper.java.additional` are passed to what we call an "agent bootstrapper". The bootstrapper is a process that for the actual agent process.

So `AGENT_STARTUP_ARGS` is the right place to setup JVM arguments for the actual agent. You'd rarely have to setup JVM args for the bootstrapper, since all it does is execute the agent if it dies.

[...]

PR #7736, opened 2020-02-13
Improvements to wrapper properties examples

arvindsv:
Issue: #7698 

Description: This PR tries to make the wrapper.conf and wrapper-properties.conf situation a little clearer.
 
1. Currently, most installers include `wrapper-config/wrapper.conf`, `wrapper-config/wrapper-properties.conf` and `wrapper-config/wrapper-properties.conf.example`. However, in most of these cases (except Windows), the `wrapper-properties.conf` file includes everything inside `wrapper-properties.conf.example`. See: https://github.com/gocd/gocd/blob/fbe640c871dfe4bff5b3e3ddead658e6ca9694a0/installers/generic.gradle#L218

    This PR removes the redundant `wrapper-properties.conf.example` file, except on Windows [See point 3]. The content is in `wrapper-properties.conf`.

2. The `wrapper-properties.conf.example` was often confusing. That's because it was a common file for both agent and server. So, it wouldn't be able to show how to set agent properties properly (which is different from how they are set on the server).

    This PR uses two separate example files (internally) and concatenates the right content into `wrapper-properties.conf` (point 1). 

3. On Windows, the exe is not packaged with a wrapper.properties file. It contains the example file.

    This PR keeps the example file around, but makes it specific to the agent/server as mentioned in point 2.

4. In the agent-specific example content, it is now more clear how to set agent-specific properties and bootstrapper-specific properties. Similarly, the server-specific example content shows the right way to change Xmx.

--------------------------------------

#### The agent-specific wrapper-properties file looks like this:

```config
# Use this file to configure your GoCD agent

# setup the GoCD server URL
wrapper.app.parameter.100=-serverUrl
wrapper.app.parameter.101=https://localhost:8154/go

# For a detailed description about setting various environment variables, see the page at
# https://wrapper.tanukisoftware.com/doc/english/props-envvars.html

# For a detailed description about setting JVM options and system properties, see the page at
# https://wrapper.tanukisoftware.com/doc/english/props-jvm.html

# To understand how properties can be overridden in this file, see the page at:
# https://wrapper.tanukisoftware.com/doc/english/props-cascading.html

# To use a custom JVM:
# wrapper.java.command=/path/to/java

# To set any environment variables of your choice use the following syntax
#set.SOME_ENVIRONMENT_VARIABLE=some-value
#set.GRADLE_HOME=/path/to/gradle
#set.M2_HOME=C:\apache-maven-3.2.1
#set.MAVEN_OPTS=-Xmx1048m -Xms256m -XX:MaxPermSize=312M

# Appending to existing environment variables:
# This config supports environment variable expansion at run time within the values of any environment variable.
# To maintain the platform independent nature of this configuration file, the Windows syntax is used for all platforms.

# On windows:
#set.PATH=..\lib;%PATH%

# On Unix/Linux/macOS. Note that you must use %PATH%, and not $PATH, as you'd normally do on this platform
#set.PATH=../lib:%PATH%

###### System properties for GoCD agent bootstrapper ######
# Set any system properties. We recommend that you begin with the index `100` and increment the index for each system property
#wrapper.java.additional.100=-Dcom.example.enabled=true
#wrapper.java.additional.101=-Dcom.example.name=bob

# If a property contains quotes, these can be stripped out, but this works only on Unix/Linux/macOS
#wrapper.java.additional.103=-Dmyapp.data="../MyApp Home"
#wrapper.java.additional.104.stripquotes=TRUE

###### System properties for GoCD agent ######
# Set a memory limit of 1GB for the agent process. We recommend that you do not use more than half your system memory.
#set.AGENT_STARTUP_ARGS=-Xms128m -Xmx1024m -Dsome.java.property=somevalue
```

#### The server-specific wrapper-properties file looks like this:

```config
# Use this file to configure your GoCD server
# For a detailed description about setting various environment variables, see the page at
# https://wrapper.tanukisoftware.com/doc/english/props-envvars.html

# For a detailed description about setting JVM options and system properties, see the page at
# https://wrapper.tanukisoftware.com/doc/english/props-jvm.html

# To understand how properties can be overridden in this file, see the page at:
# https://wrapper.tanukisoftware.com/doc/english/props-cascading.html

# To use a custom JVM:
# wrapper.java.command=/path/to/java

# To set any environment variables of your choice use the following syntax
#set.SOME_ENVIRONMENT_VARIABLE=some-value
#set.GRADLE_HOME=/path/to/gradle
#set.M2_HOME=C:\apache-maven-3.2.1
#set.MAVEN_OPTS=-Xmx1048m -Xms256m -XX:MaxPermSize=312M

# Appending to existing environment variables:
# This config supports environment variable expansion at run time within the values of any environment variable.
# To maintain the platform independent nature of this configuration file, the Windows syntax is used for all platforms.

# On windows:
#set.PATH=..\lib;%PATH%

# On Unix/Linux/macOS. Note that you must use %PATH%, and not $PATH, as you'd normally do on this platform
#set.PATH=../lib:%PATH%

# Set any system properties. We recommend that you begin with the index `100` and increment the index for each system property
#wrapper.java.additional.100=-Dcom.example.enabled=true
#wrapper.java.additional.101=-Dcom.example.name=bob

# If a property contains quotes, these can be stripped out, but this works only on Unix/Linux/macOS
#wrapper.java.additional.103=-Dmyapp.data="../MyApp Home"
#wrapper.java.additional.104.stripquotes=TRUE

# Set a memory limit of 8GB (usually needed on your GoCD server for large setups). We recommend that you do not use more than half your system memory.
#wrapper.java.additional.105=-Xmx8G
```

### DIFF
--- a/installers/include/wrapper-properties.go-server.conf.example
+++ b/installers/include/wrapper-properties.go-server.conf.example
@@ -1,10 +1,11 @@
-# This is an example configuration file to override any GoCD configuration.
-
 # For a detailed description about setting various environment variables, see the page at
 # https://wrapper.tanukisoftware.com/doc/english/props-envvars.html
 
 # For a detailed description about setting JVM options and system properties, see the page at
 # https://wrapper.tanukisoftware.com/doc/english/props-jvm.html
+
+# To understand how properties can be overridden in this file, see the page at:
+# https://wrapper.tanukisoftware.com/doc/english/props-cascading.html
 
 # To use a custom JVM:
 # wrapper.java.command=/path/to/java
--- a/installers/generic.gradle
+++ b/installers/generic.gradle
@@ -205,17 +205,11 @@ def configureGenericZip(Zip zipTask, InstallerType installerType) {
       rename 'wrapper.conf.in', 'wrapper.conf'
     }
 
-
-    // the example properties file for reference
-    from("include/wrapper-properties.conf.example") {
-      into 'wrapper-config'
-    }
-
     // the actual default config file
     from("include/wrapper-properties-${installerType.baseName}.conf") {
       into 'wrapper-config'
       rename ".*", "wrapper-properties.conf"
-      filter(ConcatFilter, append: file("include/wrapper-properties.conf.example"))
+      filter(ConcatFilter, append: file("include/wrapper-properties.${installerType.baseName}.conf.example"))
     }
 
     from("${extractDeltaPack.outputs.files.singleFile}/wrapper-delta-pack-${project.versions.tanuki}-st/bin") {
--- a/installers/include/additional-properties.conf
+++ b/installers/include/additional-properties.conf
@@ -1,4 +1,5 @@
 
 # Create this file and specify any overrides for environment/jvm args
 #include ../wrapper-config/wrapper-properties.conf
+# (The line above is NOT a comment. It's a "#include" - see: https://wrapper.tanukisoftware.com/doc/english/props-cascading.html)
 
--- a/installers/include/wrapper-properties-go-agent.conf
+++ b/installers/include/wrapper-properties-go-agent.conf
@@ -1,5 +1,4 @@
 # Use this file to configure your GoCD agent
-# See the `wrapper-properties.conf.example` file for the configuration syntax
 
 # setup the GoCD server URL
 wrapper.app.parameter.100=-serverUrl
--- a/installers/include/wrapper-properties-go-server.conf
+++ b/installers/include/wrapper-properties-go-server.conf
@@ -1,2 +1,1 @@
 # Use this file to configure your GoCD server
-# See the `wrapper-properties.conf.example` file for the configuration syntax
--- a/installers/include/wrapper-properties.go-agent.conf.example
+++ b/installers/include/wrapper-properties.go-agent.conf.example
@@ -1,0 +1,40 @@
+# For a detailed description about setting various environment variables, see the page at
+# https://wrapper.tanukisoftware.com/doc/english/props-envvars.html
+
+# For a detailed description about setting JVM options and system properties, see the page at
+# https://wrapper.tanukisoftware.com/doc/english/props-jvm.html
+
+# To understand how properties can be overridden in this file, see the page at:
+# https://wrapper.tanukisoftware.com/doc/english/props-cascading.html
+
+# To use a custom JVM:
+# wrapper.java.command=/path/to/java
+
+# To set any environment variables of your choice use the following syntax
+#set.SOME_ENVIRONMENT_VARIABLE=some-value
+#set.GRADLE_HOME=/path/to/gradle
+#set.M2_HOME=C:\apache-maven-3.2.1
+#set.MAVEN_OPTS=-Xmx1048m -Xms256m -XX:MaxPermSize=312M
+
+# Appending to existing environment variables:
+# This config supports environment variable expansion at run time within the values of any environment variable.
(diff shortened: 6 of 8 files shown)


arvindsv:
The files for each OS are here: [files.zip](https://github.com/gocd/gocd/files/4199209/files.zip)

```
$ tree files/
files/
├── deb
│   ├── agent
│   │   ├── wrapper-properties.conf
│   │   └── wrapper.conf
│   └── server
│       ├── wrapper-properties.conf
│       └── wrapper.conf
├── osx
│   ├── agent
│   │   ├── wrapper-properties.conf
│   │   └── wrapper.conf
│   └── server
│       ├── wrapper-properties.conf
│       └── wrapper.conf
├── rpm
│   ├── agent
│   │   ├── wrapper-properties.conf
│   │   └── wrapper.conf
│   └── server
│       ├── wrapper-properties.conf
│       └── wrapper.conf
├── win
│   ├── agent
│   │   ├── wrapper-properties.conf.example
│   │   └── wrapper.conf
│   └── server
│       ├── wrapper-properties.conf.example
│       └── wrapper.conf
└── zip
    ├── agent
    │   ├── wrapper-properties.conf
    │   └── wrapper.conf
    └── server
        ├── wrapper-properties.conf
        └── wrapper.conf

15 directories, 20 files
```

installers/include/wrapper-properties.go-server.conf.example
```diff
@@ -34,4 +35,5 @@
 #wrapper.java.additional.104.stripquotes=TRUE
 
 # Set a memory limit of 8GB (usually needed on your GoCD server for large setups). We recommend that you do not use more than half your system memory.
-#wrapper.java.additional.105=-Xmx8G
+# This needs to be index 2, since it needs to override the Xmx set in wrapper.conf
+#wrapper.java.additional.2=-Xmx8G
```
ketan: It is my understanding is that `java -Xmx1g -Xmx2g` will force the JVM to use the heap specified last. The intention of using indices > 100 was that it allows GoCD to reserve indices upto 100 for its internal use.

arvindsv: Ah, sorry. I didn't know that a second one overrides the previous one. Then, it makes sense to leave it at 105. I force-pushed over it to remove that change.

arvindsv:
Note to me: Verify that there are no problems during upgrades in all platforms.

GaneshSPatil:
Verified on `20.2.0-11342` (across distros) installer downloaded from: https://build.gocd.org/go/tab/build/detail/installers/2128/dist/1/dist#tab-artifacts


#### Normal Installtion

```bash
1. yum -y localinstall --nogpgcheck go-agent-20.2.0-11342.noarch.rpm

2. ls /usr/share/go-agent/wrapper-config/
wrapper-properties.conf wrapper.conf

```

#### Upgrade:

```bash
1. curl https://download.gocd.org/gocd.repo -o /etc/yum.repos.d/gocd.repo

2. yum install -y go-agent

3. rpm -q go-agent
go-agent-20.1.0-11114.noarch

4. ls /usr/share/go-agent/wrapper-config/
wrapper-properties.conf  wrapper-properties.conf.example  wrapper.conf

5. echo wrapper.java.additional.106=-Xmn1G >> /usr/share/go-agent/wrapper-config/wrapper-properties.conf

6. yum -y localinstall --nogpgcheck go-agent-20.2.0-11342.noarch.rpm

7. rpm -q go-agent
go-agent-20.2.0-11342.noarch

8. ls /usr/share/go-agent/wrapper-config/
wrapper-properties.conf wrapper.conf
```

PR #7739, opened 2020-02-14
Bump up APIs Info API to v2

GaneshSPatil:
* Add deprecated API information

Description:

```
{
  "method" : "get",
  "path" : "/api/admin/config_repos/:id",
  "version" : "application/vnd.go.cd.v2+json",
  "path_params" : [ ":id" ],
  "deprecation_info" : {
    "is_deprecated" : true,
    "deprecated_api_version" : "v2",
    "successor_api_version" : "v3",
    "deprecated_in" : "20.2.0",
    "removal_in" : "20.5.0"
  }
}
```


maheshp:
@GaneshSPatil strangely a request to v1 of API Info also returns the deprecation info. Can you please check. 
I dint find this API documented, this should be public right because its not in internal namespace.
Finally, the controller and representer tests need to be added.

chadlwilson:
Hi @GaneshSPatil and @maheshp - sorry to dredge up this old PR, but do you know if this was ever completed?

I was looking at various APIs that have more than one version available as well as looking at https://github.com/gocd/ruby-functional-tests/pull/622 and it seems that perhaps this one perhaps wasn't completed?

I infer the intent was to make it easier to discover deprecated APIs on a running server at `/api/internal/apis` and to support some test automation to discover non-tested APIs at https://github.com/gocd/ruby-functional-tests/pull/594

However the v2 logic looks a bit brittle in terms of detecting the controllers with `@DeprecatedApi`. I was trying to migrate the v1 representer and controller tests, but couldn't really get it to work reliably due to the code below which tends to return `null` class canonical names for the lambda-driven routes (which is most of them). It probably needs to be `getClass().getName()` and then some other logic to filter out ones that don't belong to an `ApiController` which then seems to work OK.

https://github.com/gocd/gocd/blob/834d0761c7c5af499eacdf35d5a3a5ee303bd6b0/api/api-api-info-v2/src/main/java/com/thoughtworks/go/apiv2/apiinfo/representers/RouteEntryRepresenter.java#L39-L50

Do you think it is worth me fixing to then get rid of v1, **or** am I best to just drop this v2 version? If you could share with me the intent of the `v2` version of the API it might help.

GaneshSPatil:
Hello @chadlwilson 

> sorry to dredge up this old PR, but do you know if this was ever completed?

I think the API was functional and was returning all the required information about Java APIs. 
Did you find any functional issues with this API?

> I was looking at various APIs that have more than one version available as well as looking at gocd/ruby-functional-tests#622 and it seems that perhaps this one perhaps wasn't completed?

Hmm, I see. I vaguely remember that, the ruby functional test was pending, because:
    * we wanted to add some logic in our functional tests to fail the test if there are more than 2 deprecated API for the some entity and if the duration is more than 3 months (to help us track and remove older deprecated APIs)
    *  to include apis from ruby.
   

> Do you think it is worth me fixing to then get rid of v1, or am I best to just drop this v2 version?

AFAIK, was developed only for internal use by ruby functional test. 
My assumption is, its an internal use API hence, it should be okay just to keep one version rather than fixing all the APIs.

chadlwilson:
Hi @GaneshSPatil 👋

> > sorry to dredge up this old PR, but do you know if this was ever completed?
> 
> I think the API was functional and was returning all the required information about Java APIs. Did you find any functional issues with this API?

When I tried to resurrect the functional test from your branch to be able to merge it, my GoCD server wasn't returning the deprecation info for deprecated APIs. Unfortunately it's a bit difficult to easily test right now, as there are no deprecated APIs left on current build.gocd.org. `v21.3.0` does have stage-instance v2 as deprecated though, [removed in v21.4.0](https://github.com/gocd/gocd/pull/9987) :-)

I am not sure if the way the JVM returned `getClass().getCanonicalName()` (for inner classes from lambdas) has changed across Java versions, but after some digging it seemed to always be null for the lambda classes in the relevant APIs, causing it to assume they were non deprecated.

Since there are no unit tests, I can't easily validate whether there is something I am missing about the intent.

If you're feeling enthusiastic you could either replicate with v21.3.0 or alternatively replicate with https://github.com/chadlwilson/gocd/commit/ff2519fdb35cef43f798cd8bb94d40bab3aae289 where I initially deprecated `/api/internal/apis` `v1` (i now understand that is not required for internal APIs, but keeping around to support some testing for a bit) - you should be able to see that the v2 version of the API isn't returning the v1 version's deprecation information in the response.

In any case, I think I do have a fix [here](https://github.com/gocd/gocd/pull/10034/commits/647d9be2884dc60f164e251707057745a726393d) on the same branch/PR which seems to work fine and perhaps a little less hacky/coupled to Spark internals, the way the JVM generates class names for lambdas or the way we implement our controllers. However I wanted to check whether it was valuable since there's not much point having an internal API with extra functionality that is unused :-)

> Hmm, I see. I vaguely remember that, the ruby functional test was pending, because: 
> * we wanted to add some logic in our functional tests to fail the test if there are more than 2 deprecated API for the some entity and if the duration is more than 3 months (to help us track and remove older deprecated APIs) 
> * to include apis from ruby.

Well my understanding is that now there are not any non-internal Ruby APIs now (since you folks subsequently migrated them all) so that part is moot? Do you think the rest is still valuable given the current state of GoCD? :-)

chadlwilson:
I've concluded that it is still valuable, as the test has a nice link to header verification on the actual APIs for any of them that are declared as deprecated, seems to work fine with a minor tweak.

Issue #7333, opened 2019-11-28
Move Package Repositories SPA to typescript


akshaydewan:
The current Package Repo page needs to be moved from rails to typescript

Here's what the current UI looks like:
<img width="1440" alt="Screenshot 2019-11-28 at 14 36 42" src="https://user-images.githubusercontent.com/6024038/69792527-d44d2100-11ec-11ea-9159-5bba122f6483.png">
<img width="1440" alt="Screenshot 2019-11-28 at 14 37 50" src="https://user-images.githubusercontent.com/6024038/69792531-d57e4e00-11ec-11ea-9df7-31fcbd59dcb7.png">

Points to discuss:
- The package repo containing packages looks similar to Cluster Profiles and Elastic Profiles. Should we use a similar layout for the new SPA? The nesting of expandable panels can be confusing for users, as we've seen on that page. Alternately, we can retain the tree structure.
- Currently the page doesn't allow packages to be added inside the repo. This needs to be done via the Pipeline config page. I think we should allow this in the new SPA for a smoother flow, and for simplifying the Pipeline Config SPA in the future.

arvindsv:
I agree about point 2. It makes sense to allow creating packages here.

kritika-singh3:
For the point 1: I prefer a layout similar to what we have on Elastic Agent Configurations SPA.
Reason: It is a much cleaner and concise representation of what the package repositories and packages represent.
What I would suggest is that the outer panel is always expanded as on Pipelines SPA. It would help to reduce the confusion regarding nested expandable panels.

WDYT? @gocd/committers

maheshp:
Verified on GoCD Version: 20.3.0 (11521-f0170986bb47797b6ccf130e1ccb87736b45b372)

PR #7745, opened 2020-02-18
Migrate Package Repositories SPA to typescript

kritika-singh3:
Issue: #7333

Description:
 - migrated the package repository spa to typescript based
 - will include the feature of creating a new package as well as a package repository

#### TODO

 - [x] we don't ask for view from the template, instead we build one on the basis of the configuration. Need to build such a component using typescript and utilize the same.


kritika-singh3:
Preview till now:

package-repo-spa

api/api-internal-package-repositories-v1/src/main/java/com/thoughtworks/go/apiv1/internalpackagerepositories/InternalPackageRepositoriesControllerV1.java
```diff
@@ -0,0 +1,76 @@
...
+        return Routes.PackageRepository.INTERNAL_BASE;
+    }
+
+    @Override
+    public void setupRoutes() {
+        path(controllerBasePath(), () -> {
+             before("", mimeType, this::setContentType);
+             before("/*", mimeType, this::setContentType);
+
+            before("", mimeType, this.apiAuthenticationHelper::checkAdminUserOrGroupAdminUserAnd403);
```
GaneshSPatil: @kritika-singh3 - can we use the publicly documented [Package Repositories](https://api.gocd.org/current/#package-repositories) and [Packages](https://api.gocd.org/current/#packages) API to fetch this information.

IMO, we should always prefer public APIs for rendering entity information on the SPAs, unless its absolutely required to build a newer internal API.

kritika-singh3: Reason for going with an internal API:
  - The public API for Package Repositories does not contain all the info related to the packages (i.e. `auto_update` and `configurations`)
 - Yes, I make two API calls to get all package repositories and packages. But, the package name is not unique across the GoCD instance - it is unique only inside a particular package repository. Hence, merging the two list (IMO) got a bit of a overhead (checking with both repo id and package name).

Let me know if you thing otherwise.

GaneshSPatil: As per the API documentation,

> The public API for Package Repositories does not contain all the info related to the packages (i.e. auto_update and configurations)

The package repositories API will not contain detailed packages information (as it is a referenced entity). But, the public packages API does contain the `auto_update` and `configurations` information.

> Yes, I make two API calls to get all package repositories and packages. But, the package name is not unique across the GoCD instance - it is unique only inside a particular package repository. Hence, merging the two list (IMO) got a bit of a overhead (checking with both repo id and package name).

What is the overhead? I mean, it is the part of the validation that both repo-id and package-name can not be duplicated, we're not doing anything extra.

---

Nevertheless, If we ourselves can not use the entity API on the entity SPA, then the usability of such API is questionable :)

kritika-singh3: Agree with you. Will make the change.

GaneshSPatil:
Just a thought: should we consider wizard view for this SPA? similar to we have on the `elastic agent configurations`.

The repository define configuration can be overridden at the packages (such as `always poll latest version`). And it will be useful to view the repository settings while defining/editing package.

\cc: @kritika-singh3 @arvindsv @maheshp @ketan

kritika-singh3:
> Just a thought: should we consider wizard view for this SPA? similar to what we have on the `elastic agent configurations`.

I thought of the same and then decided against it as a full screen modal for package repositories did not make sense to me. But yes a wizard of a medium modal size does makes sense.

GaneshSPatil:
@kritika-singh3 -- Add an ability to specify `id` while creating a package repository?

arvindsv:
> Just a thought: should we consider wizard view for this SPA? similar to what we have on the elastic agent configurations.

@GaneshSPatil Could be nice to have, but I don't feel it's worth the effort, since this one is reasonably intuitive / exists. This is unlikely to be a very frequently used page, to make it worth the effort.

api/api-packages-v1/src/main/java/com/thoughtworks/go/apiv1/packages/InternalPackagesControllerV1.java
```diff
@@ -0,0 +1,76 @@
...
+import spark.Response;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.emptyList;
+import static spark.Spark.*;
+
+public class InternalPackagesControllerV1 extends ApiController implements SparkSpringController {
```
GaneshSPatil: remove this controller?

kritika-singh3: Should I make the Usages API public?

GaneshSPatil: yes

GaneshSPatil:
### Create Package Repository View
Screen Shot 2020-02-19 at 2 39 44 PM

### Edit Package Repository View
Screen Shot 2020-02-19 at 2 39 55 PM

### Edit Package Repository View with validation failures
Screen Shot 2020-02-19 at 2 40 17 PM

### Create Package View
Screen Shot 2020-02-19 at 4 29 24 PM

api/api-packages-v1/src/main/java/com/thoughtworks/go/apiv1/packages/InternalPackagesControllerV1.java
kritika-singh3: https://github.com/gocd/gocd/pull/7841

maheshp:

There are few small issues, will create a new issue for the same.

Issue #7676, opened 2020-01-29
GoCD sends `run_instance_count` as 0 during pipeline export


kritika-singh3:
##### Issue Type

- Bug Report

##### Summary

During export of any pipeline, GoCD sends a copy of current pipeline config to the plugin. The value of `run_instance_count` for a job is set to `'0'`.

###### Basic environment details

* Go Version: `20.1.0 (11115-af4bdadd3f6545617327c56f4cb3b92abe4f1181)r`
* JAVA Version: `11.0.1`
* OS: `Mac OS X 10.13.6`

##### Steps to Reproduce

1. Install GoCD Groovy [plugin](https://github.com/gocd-contrib/gocd-groovy-dsl-config-plugin/releases)
2. Define a pipeline `test` with a job to run with one instance.
<img width="308" alt="Screen Shot 2020-01-29 at 4 29 37 PM" src="https://user-images.githubusercontent.com/41165891/73351252-93d55680-42b4-11ea-9a14-889ff13c4fe6.png">

3. Go to Pipelines SPA.
4. Find `test` and click export. Select `GoCD Groovy Configuration plugin`.
5. Look at the file generated.

##### Expected Results

There should no `run_instance_count` present in the file. Even if present, it should be set to `null`.

##### Actual Results

The value of `run_instance_count` for the job is set to `0`

##### Any other info

If I send the same file as input to the GoCD server, the plugin throws an error stating `0 is not a valid value`. 
IMO, what the value of `run-instance_count` should mean: 
 - `0`: no job instances to run
 - `1` or `null`: run a single job instance
 - `x`: partition the task into x parts 
 - `all`: to run on all agents

Currently, if my groovy files contain `run_instance_count: '0'`, the plugin throws an error stating `0 is not a valid value` - which is correct.
On setting it to `run_instance_count: '1'`, the GoCD accepts it as a valid pipeline and schedules it. But the resulting job instance is suffixed by `-1` which is not the desired results.

maheshp:
Verified on GoCD Version: 20.2.0 (11290-b369eff87cba8fc33c38670177290da6d62c6391)

PR #7751, opened 2020-02-19
Do not set 0 as default for runInstanceCount

kritika-singh3:
Issue: #7676 

Description:
 - If no value is set for runInstanceCount - it should be defaulted to `null` rather than `0`.
    This should be the case when the job is set as run only a single instance.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/JobConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/JobConfig.java
@@ -226,11 +226,12 @@ public class JobConfig implements Validatable, ParamsAttributeAware, Environment
     }
 
 	public boolean isRunMultipleInstanceType() {
-		return getRunInstanceCountValue() > 0;
+        Integer runInstanceCountValue = getRunInstanceCountValue();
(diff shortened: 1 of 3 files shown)


config/config-api/src/main/java/com/thoughtworks/go/config/JobConfig.java
```diff
@@ -226,11 +226,12 @@ public void setRunOnAllAgents(boolean runOnAllAgents) {
...
 	public boolean isRunMultipleInstanceType() {
-		return getRunInstanceCountValue() > 0;
+        Integer runInstanceCountValue = getRunInstanceCountValue();
+        return runInstanceCountValue != null && runInstanceCountValue > 0;
 	}
 
 	public Integer getRunInstanceCountValue() {
-		return runInstanceCount == null ? 0 : Integer.valueOf(runInstanceCount);
+		return isBlank(runInstanceCount) ? null : Integer.valueOf(runInstanceCount);
 	}
```
kritika-singh3: Blow up if `runInstanceCount` is null and change the return type to `int`.

Issue #7754, opened 2020-02-19
Comparison SPA is broken for plugin and package materials


kritika-singh3:
##### Issue Type

- Bug Report

##### Summary

The new Comparison SPA is broken for pipelines which contains `package` or `plugin` materials

###### Basic environment details

* Go Version: `GoCD Version: 20.2.0 (11216-6361d67fe0c091f39cdccd84df2077e6e820e45e).`
* JAVA Version: `13.0.1`
* OS: `Linux 4.14.165-133.209.amzn2.x86_64`

##### Steps to Reproduce

Visit https://build.gocd.org/go/compare/trigger/5591/with/5592.

1. Create a pipeline which uses a `package` or `plugin` material. Use https://www.gocd.org/plugins/#package-repo or https://www.gocd.org/plugins/#scm.
2. Trigger few builds.
3. Click on `Compare` link on the dashboard for that pipeline.

##### Expected Results

The SPA should render the differences between the two instances

##### Actual Results

Page error

<img width="1015" alt="Screen Shot 2020-02-19 at 5 23 59 PM" src="https://user-images.githubusercontent.com/41165891/74832154-a6cfc980-533c-11ea-9fc8-4621264fcdb4.png">

maheshp:
Verified on - GoCD Version: 20.2.0 (11325-dcb8fa908ba08d4f82de172bc942de6f2ca1e28d).

PR #7758, opened 2020-02-20
Add support for plugin and package materials on Comparison SPA

kritika-singh3:
Issue: #7754

Description:
 - The plugin and package materials were not supported on the SPA. Added support for that
 - To uniquely identify such materials - added a `display_type` and `description` on the material json in the comparison result ( bumped up the API to v2)

API Doc: https://github.com/gocd/api.go.cd/pull/625


maheshp:
@kritika-singh3 probably this should have been a review comment, but I think `display_type` and `description` should be under `attributes` for each material.

kritika-singh3:
@maheshp `type` and `display_type` has a minor difference. Hence, decided to keep them at the same level.
Let me know if you think otherwise, will raise a PR to make the changes.

maheshp:
@kritika-singh3 I would probably move it to `attributes`.

maheshp:
Verified on - GoCD Version: 20.2.0 (11325-dcb8fa908ba08d4f82de172bc942de6f2ca1e28d).

PR #7804, opened 2020-03-03
Fix autocomplete text field styles to render the suggestions after help text

GaneshSPatil:
#### Before
Screen Shot 2020-03-03 at 2 36 04 PM

#### After
Screen Shot 2020-03-03 at 2 35 05 PM

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/components/forms/autocomplete.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/components/forms/autocomplete.tsx
@@ -67,10 +67,7 @@ export abstract class SuggestionProvider {
   }
 
   update(): void {
-    this.getData().
-      then(this.receiver).
-      catch(this.errorMsg).
-      finally(this.done);
+    this.getData().then(this.receiver).catch(this.errorMsg).finally(this.done);
   }
 
   abstract getData(): Promise<Awesomplete.Suggestion[]>;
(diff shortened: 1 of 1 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/views/components/forms/autocomplete.tsx
```diff
@@ -135,6 +133,14 @@ export class AutocompleteField extends RestyleViewComponent<Styles, AutoCompAttr
...
     }
 
-    return <TextField {...attrs} />;
+    return <TextFieldForAutoComplete {...attrs} />;
+  }
+}
+
+export class TextFieldForAutoComplete extends TextField {
+  renderInputField(vnode: m.Vnode<TextFieldAttrs>) {
+    return <div id="autocomplete-input-container">
```
marques-work: Since `id` is intended to be unique per page (the browser won't fail, but `getElementById()` will always return the first), should we use a different attr?

GaneshSPatil: @marques-work - Hmm, not against it.

Let's assume, someone consumes this component on an SPA, and reuses this ID, which will cause auto-suggestions not to work. But this failure will be constant, meaning, it will always fail (kind of a static error which does not depend on the data). Shouldn't this be a good enough hint for the consumer to fix the conflicting IDs?

Also, if we're using a different attribute, we can also change the value and append a random-uuid-string at the end of it.

marques-work:
Verified this is fixed on 5112d6824e0855d7e34f78513d98ccd6cb0f4106

server/src/main/webapp/WEB-INF/rails/webpack/views/components/forms/autocomplete.tsx
arvindsv: > Shouldn't this be a good enough hint for the consumer to fix the conflicting IDs?

This is very much not my place to say, but I would be _very_ confused if I added two autocomplete components in a page and this happened (or it started showing autocompletes in a component off screen). I think it is not recommended to use an id in a component, isn't it?

PS: Feel free to ignore me. This is just a drive-by comment -- and I'm pretty sure I haven't understood the context of what this is).

GaneshSPatil: @arvindsv  @marques-work - You're right. This possibly might cause problems if the component is reused multiple times on the SPA.

I will send a PR to fix this, to use `autocomplete-id` attribute which has a randomly generated uuid value.

PR #7807, opened 2020-03-04
[Pipeline Config SPA] Bind validation errors with the model

GaneshSPatil:
* Show server side validation errors

Screen Shot 2020-03-04 at 6 01 53 PM


arvindsv:
@GaneshSPatil If there are unexpected errors (keys don't match), do they get "lost"? Is there a way to show unbound errors at the top? I've seen the old page sometimes say: "Save failed. See errors below" and there are no errors below and it's quite irritating.

kritika-singh3:
@arvindsv The unbound errors are shown at the top of the page along with a message `Validations failed for pipeline 'test1'. Error(s): [Validation failed.]. Please correct and resubmit.`.

Scenario for verifying - try to delete a dependency material which has been used in a fetch artifact task.

PR #7817, opened 2020-03-05
Make internal resources controller accessible by all normal users

GaneshSPatil:
Resources list needs to be accessible on the new pipeline config SPA to provide resource autocompletions, thus following users needs to have access to this API:
- Super Admins
- Pipeline Group Admins
- Template Admins

Did not find any harm in opening up this API for all users as users can find the resources information through Agents SPA/API.


GaneshSPatil:
@arvindsv -- merging this. 
If you think otherwise, will submit another PR to revert these changes.

arvindsv:
@GaneshSPatil This is the list of resources across agents and jobs in the config, right? If so, this change is fine.

GaneshSPatil:
@arvindsv - Yes.

Thank you

maheshp:
Verified on GoCD Version: 20.2.0 (11308-05dc4d14977d4bf10d5f5c9e9d6f88817df1ec1e)

PR #7830, opened 2020-03-09
Don't showcase pending agents in edit agents modal

kritika-singh3:
Any pending agent cannot be added to an env. Hence, not showcasing them on the edit agents modal on the Environments SPA.


maheshp:
Verified on GoCD Version: 20.2.0 (11325-dcb8fa908ba08d4f82de172bc942de6f2ca1e28d)

PR #7832, opened 2020-03-10
Api Infos API fixes

GaneshSPatil:
Description:

* Honour API versioning.
* Move API under internal namespace.

Expect a Functional Test failure due to this change. Will fix it once it fails.


maheshp:
@GaneshSPatil I checked on GoCD Version: 20.2.0 (11325-dcb8fa908ba08d4f82de172bc942de6f2ca1e28d). It looks like a request with the `Accept` header `application/vnd.go.cd.v2+json` and `application/vnd.go.cd+json` gets routed to the v1 of the API. Can you please check.

Issue #7808, opened 2020-03-04
Compare SPA browse timeline UI issues


maheshp:
##### Issue Type

- Bug Report

##### Summary

The Comparison SPA, Browse Timeline view needs to be fixed.

<img width="1079" alt="Screen Shot 2020-03-04 at 3 01 55 PM" src="https://user-images.githubusercontent.com/1061051/75931247-5ae74f00-5e29-11ea-867a-f86f5bde3ac7.png">

marques-work:
@maheshp I looked at the compare modal yesterday on build.gocd.org and it looked like this is all fixed.

marques-work:
@maheshp actually I take that back. On MS Edge, it is still broken:

![image](https://user-images.githubusercontent.com/723642/76561506-aff61700-6460-11ea-8b98-b093a3669a51.png)

cc: @kritika-singh3

maheshp:
This issue has been verified - https://github.com/gocd/gocd/pull/7863#issuecomment-610328799

PR #7850, opened 2020-03-12
Fix for broken ui on timeline modal on compare spa

kritika-singh3:
Issue: #7808

Description:
 - changed from flex implementation  to table

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/modal.scss
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/modal.scss
@@ -54,10 +54,8 @@ $hover-background: #c5d9e0;
 }
 
 .pipeline-run {
-  display:        flex;
-  flex-direction: row;
-  cursor:         pointer;
-  padding:        7px 10px;
+  cursor:  pointer;
+  padding: 4px 2px;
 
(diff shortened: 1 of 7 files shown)


kritika-singh3:
Chrome:

Screen Shot 2020-03-12 at 2 10 51 PM

Firefox:

Screen Shot 2020-03-12 at 2 11 02 PM

marques-work:
MS Edge :(
image

GaneshSPatil:
@kritika-singh3 -  

1. As per the provided screenshots, the pipeline counter are split into 2 lines, which looked a bit misaligned to me.

2. Also, the stage tile(s) will occupy the remaining space, meaning, in case of 1 stage - 100% width, in case of n stages - (100/n)% width.
This was the similar case with old GoCD dashboard, in case of multiple stages, the stages would look cluttered, hence we moved to fixed width stage tile.

Can we do something similar? Show fixed width stage tile, and if there are more stages, overflow them to the next line?

\cc: @arvindsv @maheshp @marques-work - Thoughts?

arvindsv:
If it's easier to be consistent with the dashboard, I'm fine with that. I don't know how that will look when there are tens of stages, though.

maheshp:
Related PR with fixed width - https://github.com/gocd/gocd/pull/7863

PR #7852, opened 2020-03-12
Fix "Create first pipeline" prompt

marques-work:
For the test drive, the create first pipeline prompt should only appear on the first pipeline group; before this change, it appears for every group.


marques-work:
Note for me: merge after 20.2.0

PR #7863, opened 2020-03-16
On the timeline view, used a fixed width tile for stage

kritika-singh3:
Issue: #7808 

Description:
 - width for each stage tile is fixed at 25px

Pipeline with 9 stages:

Screen Shot 2020-03-16 at 11 27 53 AM

Pipeline with 2 stages:

Screen Shot 2020-03-16 at 11 29 16 AM


streisguth:
@kritika-singh3 My only feedback is that it would be great if the second line of stages aligned with the first.
76727836-20f84000-677b-11ea-9dd5-199534a97fe4

streisguth:
I have a second piece of feedback, actually. You might want to consider reducing each line of stages to 4 instead of 5 (if the sidebar is a fixed width), and then right-aligning the pipeline ID numbers so that the layout does not change when the pipeline ID is 2, 3, or 4 digits long.
76727836-20f84000-677b-11ea-9dd5-199534a97fe4

kritika-singh3:
Preview after the changes:

timeline-view

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/stages/stages_widget.tsx
```diff
@@ -24,11 +24,32 @@ const classnames = bind(styles);
...
+      case "failing":
+        return styles.failing;
+      case "cancelled":
+        return styles.cancelled;
+      case "passed":
+        return styles.passed;
+      default:
+        return styles.unknown;
+    }
+  }
```
GaneshSPatil: how about this?

```js
static stageStatusClass(status?: string = "unknown") {
    return styles[status.trim().toLowerCase()];
}
```

Secondly, if we want to make that status can yield only specific values, we can change the type of the `status` argument from string to the possible values as follows:

```js
static stageStatusClass(status?: "passed" | "failed" |  "unknown" = "unknown") {
    return styles[status.trim().toLowerCase()];
}
```

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/timeline_modal.tsx
```diff
@@ -123,6 +126,34 @@ export class TimelineModal extends Modal {
...
+      case 5:
+        return styles.medium;
+      case 6:
+      case 7:
+        return styles.large;
+      default:
+        return styles.xLarge;
+    }
+  }
+
```
GaneshSPatil: Hmm. I am not sure if there is a better way of writing this.

@marques-work -- Thoughts??

arvindsv:
Verified on 20.3.0 (11467-1fb7c3cc3a71abb71e827f5bbdc2cb139edcd660).

Issue #7865, opened 2020-03-16
Group with name already exists on material modification


Evesy:
##### Issue Type

- Bug Report

##### Summary

Material modifications on pipelines error with `Group with name 'delivery_engineering' already exists`

###### Basic environment details

* Go Version: `20.1.0`
* JAVA Version: `13.0.1`
* OS: `Linux 2.6.32-696.18.7.el6.x86_64`

##### Steps to Reproduce

1. Existing pipeline within a pipeline group (e.g. group name `example`)
2. Select to clone pipeline. In popup box provide a new pipeline name, and specify the _same_ group name, but different casing, e.g. `Example`)
3. The pipeline is cloned, and a new group is also created
4. Try editing the new pipeline material (Or any other material on any pipeline) and save. Save will fail with the message `Group with name 'example' already exists`

Duplicated pipeline group can be removed to resolve the issue.

maheshp:
This looks like an issue with the PipelineConfig API - https://api.gocd.org/20.2.0/#create-a-pipeline. The PipelineGroup name should be validated for case insensitive uniqueness. 

@Evesy we are close to a GoCD 20.2.0 release, if possible we will fix it else will be done as part of 20.3.0

PR #7867, opened 2020-03-17
Make pipeline group case-insensitive (#7865)

GaneshSPatil:
### Description:

* Perform case-insensitive pipeline group name equals check.

* When a user specifies a pipeline group name, while creating
  a new pipeline and the same group name with different case
  already exists, do not create a new group. Create the new
  pipeline in the existing pipeline group.

### DIFF
--- a/server/src/test-integration/java/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
+++ b/server/src/test-integration/java/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
@@ -195,6 +195,22 @@ public class PipelineConfigServiceIntegrationTest {
     }
 
     @Test
+    public void shouldCreatePipelineConfigWhenCaseInsensitivePipelineGroupIsSpecified() {
+        GoConfigHolder goConfigHolderBeforeUpdate = goConfigDao.loadConfigHolder();
+        pipelineConfig = GoConfigMother.createPipelineConfigWithMaterialConfig(UUID.randomUUID().toString(), git("FOO"));
+        pipelineConfigService.createPipelineConfig(user, pipelineConfig, result, groupName.toUpperCase());
+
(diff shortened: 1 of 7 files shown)


GaneshSPatil:
The bug mentioned as part of #7865, causes **"duplicate pipeline group name"** validation error (which is not caught by the pipeline config save validations).
Afterwards, all the full config save operations will fail because of invalid config.

It's highly unlikely that users will have servers running with mentioned bug (and invalid config). As the subsequent full config save calls and server restarts will fail.

Assuming the existing config is valid,
This PR makes sure that the existing pipeline group name is used for creating the new pipeline when specified group name matches the existing group name case-insensitively.

If in case users have duplicate pipeline groups in the config, such users are urged to fix their config manually (once) and going forward this fix will make sure that duplicate pipeline group names are avoided.

\cc: @maheshp @kritika-singh3

server/src/test-integration/java/com/thoughtworks/go/server/service/PipelineConfigServiceIntegrationTest.java
```diff
@@ -194,6 +194,22 @@ public void shouldCreatePipelineConfigWhenPipelineGroupDoesNotExist() throws Git
         assertThat(configRepository.getCurrentRevision().getUsername(), is(user.getDisplayName()));
     }
 
+    @Test
+    public void shouldCreatePipelineConfigWhenCaseInsensitivePipelineGroupIsSpecified() {
```
maheshp: This test is useful but we need to add a test where the pipeline group is created and ensure it does a case insensitive match.

GaneshSPatil:
@maheshp - fixed the pipeline group create API issue with duplicate pipeline group name when different case pipeline group name is specified.
The API now throws **422**, `another pipeline group with the same name already exists` error.

PR #7876, opened 2020-03-18
Introduce config repos v4 API

marques-work:
Changes:

  - Alternate serialization for configuration properties allows clients to specify plaintext values to be encrypted by GoCD:

```javascript
{
  "configuration": [
    { "key": "foo", "value": "bar", "is_secure": true },                                 // encrypts value -> encrypted_value
    { "key": "baz", "value": "quu" },                                                    // backwards compatible with v3; is_secure is optional
    { "key": "secret", "encrypted_value": "AES:abc123..." },                             // still works
    { "key": "still_a_secret": "encrypted_value": "AES:abc123...", "is_secure": false }, // ignores value of is_secret when encrypted_value explicitly provided
    { "key": "plain", "value": "wide open", "is_secure": false }                         // is_secure == false is basically a no-op
  ]
}
```


marques-work:
Builds passed; will rebase, reword, then merge.

ketan:
On Thu, Mar 19, 2020 at 3:01 AM marques-work <notifications@github.com>
wrote:

>     { "key": "still_a_secret": "encrypted_value": "AES:abc123...", "is_secure": false }, // ignores value of is_secret when encrypted_value explicitly provided
>
>
In a few places we consider this as ambiguous and raise an error.

- Ketan

PR #7879, opened 2020-03-19
Update the bisect warning msg

kritika-singh3:
Issue: https://github.com/gocd/gocd/pull/7503#issuecomment-599717827

Description:
  - showcasing stages along with the message that the selected instance is a bisect

Screen Shot 2020-03-19 at 9 23 40 AM

### TODO

 - [ ] Update the [docs site](https://docs.gocd.org/current/advanced_usage/compare_pipelines.html) to add explanation related to `bisect` comparison

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/comparison_result_widget.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/compare/comparison_result_widget.tsx
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
+import {docsUrl} from "gen/gocd_version";
 import {MithrilViewComponent} from "jsx/mithril-component";
 import _ from "lodash";
 import m from "mithril";
 import {Comparison, DependencyRevisions, MaterialRevisions} from "models/compare/compare";
 import {DependencyMaterialAttributes} from "models/compare/material";
 import {InfoCircle} from "views/components/icons";
+import {Link} from "views/components/link";
 import {DependencyRevisionsWidget} from "./dependency_revisions_widget";
 import styles from "./index.scss";
 import {MaterialRevisionsWidget} from "./material_revisions_widget";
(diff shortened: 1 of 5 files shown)


arvindsv:
Maybe we can warn about the consequence of this? I suppose this could mean that the changes view might not be dependable?

kritika-singh3:
@arvindsv I was thinking we can document the same about on the [docs site](https://docs.gocd.org/current/advanced_usage/compare_pipelines.html) and link the same. WDYT?

arvindsv:
@kritika-singh3 Agree. Makes sense.

kritika-singh3:
Added a `Learn More` link:

Screen Shot 2020-03-20 at 9 54 03 AM

GaneshSPatil:
@kritika-singh3 I have the least knowledge about bisect. May be we should ask someone else who understands this well.

kritika-singh3:
@GaneshSPatil Sure.
@arvindsv Do you some time to verify the changes?

arvindsv:
Of course. I'll take a look.

arvindsv:
I compared `20.3.0 (11450-b09b089652fae2543cda69a9a788038547a19054)` with `19.8.0 (9915-9ea99a72c338a132ae1ca83f363e16b2c95d920b)`. The config was like this:

```
<pipeline name="P1">
  <materials>
    <git url="/tmp/test.git" />
  </materials>
  <stage name="S1">
    <jobs>
      <job name="J1">
        <tasks>
          <exec command="ls" />
        </tasks>
      </job>
    </jobs>
  </stage>
</pipeline>
```

The pipeline runs were:

```
1. P1-1: With commit C1. [Manual]
2. P1-2: With commit C2. [Manual]
3. P1-3: With commit C3. [Automatic]
4. P1-4: Forced with commit C1 [bisect].
5. P1-5: With commit C4. [Automatic]
```

I tried a lot of combinations of compare pipeline with both releases of GoCD. The results were comparable.

One small difference is that in 19.8, there was a button to click and once you clicked it (to confirm that you do want to compare, it'll remove the message from the top of the compare pipeline screen. This is fine.

#### Old:



<hr>



#### New:

PR #7841, opened 2020-03-11
Bump up packages api

kritika-singh3:
Description:
Bumping up the packages API to v2.
Added a Usages API for a given package id.

Curl request:
```curl
curl "http://go-server-url/go/api/admin/packages/4f4d7748-d490-4d66-92f9-4870f9a68116/usages" \
     -u 'username:password' \
     -H 'Accept: application/vnd.go.cd+json'
```

Response:

```json
{
  "usages" : [ {
    "group" : "grp1",
    "pipeline" : "pipeline1"
  }, {
    "group" : "grp2",
    "pipeline" : "pipeline2"
  } ]
}
```
##### TODO
 - [ ] API documentation

PR #7881, opened 2020-03-19
Update Package Usages API

kritika-singh3:
- added `_links` with self, doc and find


arvindsv:
Verified on 20.3.0 (11467-1fb7c3cc3a71abb71e827f5bbdc2cb139edcd660) based on https://github.com/gocd/gocd/pull/7841#issuecomment-610354957.

Issue #7860, opened 2020-03-13
Pipeline Deletion Issue


OperationsEngineers:
##### Issue Type

- Bug Report
- Feature proposal

##### Summary

Before v20.1.0 when deleting a pipeline the delete icon was greyed out if the pipeline had a dependancy. There was also a note stating what the dependency was when hovering over the delete icon.
Now every pipeline allows you to select the delete icon, and once selected it appears to delete the pipeline. If the pipeline however has a dependancy on another pipeline, it doesnt actually delete, and we are not informed of what other pipeline is dependant on it.

##### Environment

All environments

###### Basic environment details

* Go Version: 20.1.0
* JAVA Version: 13.0.1
* OS: Linux 2.6.32-358.el6.x86_64
* Browser vendor and version (if relevant): Google Chrome

##### Steps to Reproduce

This is only for pipelines that have a dependancy for another pipeline
1. Select ADMIN - Pipelines
2. Select the delete icon on a relevant pipeline

##### Expected Results
Delete icon should not be clickable on a pipeline with dependancies and a message should be displayed stating the dependancy

##### Actual Results
Delete icon is available, process appears to be deleting the pipeline, but the pipeline remains

kritika-singh3:
@OperationsEngineers What does your screen looks like when you try to delete your upstream pipeline?

I tried a 2 pipeline setup - an `upstream` and a `downstream` pipeline. The `downstream` pipeline has the `upstream` as a material.
When I tried deleting `upstream`,  I got the following:
<img width="1673" alt="Screen Shot 2020-03-16 at 3 03 33 PM" src="https://user-images.githubusercontent.com/41165891/76742756-9e31ae00-6797-11ea-8e4d-ff91f4a7abd0.png">

My GoCD server version: `20.1.0 (11114-f2c1771e333ea92dae82cd78d2efa1310c1086f0)`

OperationsEngineers:
Hi,
We are not seeing the message regarding the dependancy. When selecting the delete icon we are presented with:
![delete1](https://user-images.githubusercontent.com/29859465/76749563-944a8380-6774-11ea-83dd-c8d7e2ccef0a.jpeg)
Then when confirming the delete, we are taken back to the pipeline page where the selected pipeline hasnt been deleted:
![delete2](https://user-images.githubusercontent.com/29859465/76749882-2a7ea980-6775-11ea-9937-f2ca776ebedc.jpeg)

[...]

PR #7886, opened 2020-03-19
Improve Pipelines SPA UX

kritika-singh3:
Issue: #7860 

Description:
 - Updated internal pipeline structures api v1 to return environment and dependant pipelines
 - Updated the Pipeline SPA to disable delete pipeline icon based on
      - origin:  if defined remotely
      - if a part of any environment
      - if has some pipelines dependant on it

### DIFF
--- a/api/api-internal-pipeline-structure-v1/build.gradle
+++ b/api/api-internal-pipeline-structure-v1/build.gradle
@@ -20,6 +20,8 @@ apply plugin: 'groovy'
 dependencies {
   compile project(':api:api-base')
 
+  compileOnly group: 'org.projectlombok', name: 'lombok', version: project.versions.lombok
+  annotationProcessor group: 'org.projectlombok', name: 'lombok', version: project.versions.lombok
   testCompile project(path: ':api:api-base', configuration: 'testOutput')
 
   testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: project.versions.junit5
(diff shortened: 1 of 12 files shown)


kritika-singh3:
Previews:

1. Delete icon when enabled 

delete-pipeline-2

2. Delete icon if pipeline is defined in a config-repo

delete-pipeline-1

3. Delete icon if pipeline is a part of environment

delete-pipeline-3

4. Delete icon when pipeline has downstream pipelines

delete-pipeline-4

api/api-internal-pipeline-structure-v2/build.gradle
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2019 ThoughtWorks, Inc.
```
GaneshSPatil: update copyright?

.gitignore
```diff
@@ -521,3 +521,7 @@ api/api-packages-v2/out/
 api/api-packages-v2/target/
 api/api-packages-v2/logs/
 api/api-packages-v2/config/
+api/api-internal-pipeline-structure-v2/out/
```
GaneshSPatil: as we're introducing v2 of internal pipeline-structure API, should we remove `v1`?

kritika-singh3: Decided against bumping up the api. Updated the v1 itself to return the desired values.

maheshp:
Verified on GoCD Version: 20.3.0 (11450-b09b089652fae2543cda69a9a788038547a19054).

PR #7887, opened 2020-03-20
Upgrading rails to 5.2.4.2 #000

maheshp:
* Upgrading to latest 5.x version to ensure latest vulnerablities are
  fixed.

Issue: #

Description:


marques-work:
The two big vulnerabilities this fixes are:

1. [CVE-2020-5267: XSS via `escape_javascript()`](https://groups.google.com/forum/#!topic/rubyonrails-security/55reWMM_Pg8P)
2. [CVE-2019-16782: Timing attacks for session ID](https://www.openwall.com/lists/oss-security/2019/12/18/2)

I'm not skilled enough to prove that session IDs are now protected from timing attacks by mounting such a PoC attack, so I'll take faith that the upgraded rack gem is sufficient.

I was able to validate that the XSS vulnerability is no longer an issue for us:

Pre-5.2.4.2, adding this piece of code to a view:

```erb
  <script type="text/javascript">
    let a = `<%= escape_javascript (params[:foo] || "") %>`;
  </script>
```

Paired with a URL constructed like this:

http://localhost:8153/go/preferences/notifications?foo=`%3B%20console.log(window.location)%3B`

Actually executes code.

After the upgrade, this no longer happens.

marques-work:
Otherwise, I think this PR looks good. I'll move this to Done.

PR #7904, opened 2020-03-24
Update add package material

kritika-singh3:
Issue: #6737

Description:
 - showing whether the selected package repo has a valid plugin
 - showing error if no packages are configured fr a given package repo

Screen Shot 2020-03-24 at 8 58 54 PM

Screen Shot 2020-03-24 at 8 59 01 PM


arvindsv:
I would remove "SPA" from there. It won't make sense to users.

GaneshSPatil:
is there a reason to not bind the error message `no packages defined for the selected package repository...` with the package select field?

kritika-singh3:
> is there a reason to not bind the error message `no packages defined for the selected package repository...` with the package select field?

@GaneshSPatil  No reason. I just thought that it is not an error but just a configuration that is missing. Let me know if think otherwise.

GaneshSPatil:
Hmm, how about doing one of the following:

1. Bind the error message to the `package repository`. As the selected package repository does not contain any packages.

2. Instead of the message being a normal error text, can we render it as a warning flash message?

\cc: @maheshp @arvindsv

kritika-singh3:
Created a PR https://github.com/gocd/gocd/pull/7920 with the suggestions.

PR #7914, opened 2020-03-25
Fixing pipeline creation error

kritika-singh3:
Setting material name as "" causes the create call to fail - setting them to `undefined`


arvindsv:
Verified on 20.3.0 (11467-1fb7c3cc3a71abb71e827f5bbdc2cb139edcd660).

@kritika-singh3 Not sure what the problem was. I'm able to create pipelines with and without materials names at this time.

kritika-singh3:
Checkout the commit `e333731fe4` on master - try to create a new pipeline

The request payload:

Screen Shot 2020-04-07 at 7 35 00 PM

Error:

Screen Shot 2020-04-07 at 7 35 07 PM

arvindsv:
@kritika-singh3 Got it. Thanks! Alright, I think I can move this to QA Done.

PR #7919, opened 2020-03-26
Add Pluggable SCM SPA

kritika-singh3:
Description:
Add a SPA to add/edit/delete a Pluggable SCM material which can be used as a material for any pipeline

Preview:

1. No plugins available

Screen Shot 2020-03-26 at 4 14 58 PM

2. Scm Usages (Currently points to old pipeline config materials edit)

Screen Shot 2020-03-26 at 4 41 13 PM

3. General

scm-spa

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
@@ -569,4 +569,8 @@ export class SparkRoutes {
   static pipelineStatusApiPath(pipelineName: string) {
     return `/go/api/pipelines/${pipelineName}/status`;
   }
+
+  static scmUsagePath(scm_name: string): string {
+    return `/go/api/admin/scms/${scm_name}/usages`;
+  }
 }
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/pluggable_scm.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/pluggable_scm.ts
@@ -16,6 +16,8 @@
 
 import {Configurations, PropertyJSON} from "models/shared/configuration";
 import Stream from "mithril/stream";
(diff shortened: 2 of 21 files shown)


kritika-singh3:
@GaneshSPatil @maheshp  Should we put this behind a toggle?

GaneshSPatil:
@kritika-singh3 - This is a completely new page. I wouldn't mind announcing it directly. 
if in case there are problems with this page, users still can modify the SCMs from Pipeline Config Page.

arvindsv:
I verified this on 20.3.0 (11467-1fb7c3cc3a71abb71e827f5bbdc2cb139edcd660).

One small aspect that caught me out a bit is that, if you click on the "Clone" (or "Duplicate") button, you can't just give that material a name and save it, since it will be considered a duplicate. You'll need to change some details to make it a new material.

--- --- ---

One small nitpick: If you set the name to contain spaces, it says: "Invalid Id. This must be alphanumeric and can contain underscores and periods (however, it cannot start with a period)". Two small problems with it:

1. It's invalid name, not invalid ID (which is a different field).
2. The name can actually contain hyphens too, and not just underscores and periods.

--- --- ---

While creating a new pluggable SCM: If you provide a name and then remove the (generated) ID and click "Save", nothing happens. No error messages are shown. It looks like no network call is made either. The ID field also does not have an asterisk next to it to show that it is mandatory.

--- --- ---

I've opened https://github.com/gocd/gocd/issues/7976 for a potential issue related to duplication / clone.

--- --- ---

When I try to edit an SCM: I am able to change the "Plugin" field and click "Save". It sends the new plugin type to the server and it comes back. But, nothing is changed in the view. Maybe that field should also be read-only?

arvindsv:
cc: @kritika-singh3

maheshp:
Created a new issue https://github.com/gocd/gocd/issues/7993 for all the outstanding issues to be fixed.

Issue #7022, opened 2019-09-30
Pipeline creation wizard missing package material type


donita85:
##### Issue Type

- Bug Report

##### Summary

The pipeline creation wizard introduced in version 19.8 of gocd is lacking the possibility to create a new pipeline by using the package material type. 

##### Environment

```
"Go Server Information": {
    "Version": "19.9.0 (10194-34fca8f7682a65ef5fbd6726d51fd52b0c041bc6)"
  }
```
-->

###### Basic environment details

* Go Version: `19.9.0`
* JAVA Version: `12.0.1`
* OS: `Linux`
* Browser vendor and version (if relevant): `Chrome and Firefox`

##### Steps to Reproduce

1. A gocd server is running on version 19.9.0
2. Trying to create a new pipeline via the web GUI
3. Trying to set as material type for the new pipeline the Package option
4. The package option is missing from the drop-down menu

##### Expected Results
Like in versions < 19.8, package should be one of the options for material type

##### Actual Results
Cannot select package as material type.

##### Any other info
When creating a new pipeline via the API with the package option for material type everything looks ok. Even in the GUI, afterwards, the pipeline created like this can be selected, a new material can be added, and the type of this new material can be set to package. It's just that the wizard is lacking this option.

adityasood:
@donita85 the issue is captured in a brief [here](https://github.com/gocd/gocd/issues/6445) with some thoughts on the user flow.

donita85:
Thank you for the answer @adityasood! Do you have an estimation when this feature would be implemented? We are just now looking to upgrade our gocd instances from 19.5 and this feature would be really important for us. Our workflow includes a lot of pipelines created with package as material type and sometimes they are manually created.
Can maybe a fallback be implemented when one would not want to use the new wizard for pipeline creation? In this case, the old menu could appear?

marques-work:
@donita85 Thanks for bringing this to our attention. I don't have a timeframe yet, but we haven't forgotten about this and will be working on it soon (i.e., adding an option to choose a package repo in the material selector).

jlusiardi:
Hello,

this is also missing in the docker image `gocd/gocd-server:v20.2.0`. 

You can't select `package` as material type.

arvindsv:
Maybe some of the work @kritika-singh3 and @GaneshSPatil are doing around pipeline config SPA could be re-used here? @marques-work might have some thoughts as well.

kritika-singh3:
[...]

PR #7920, opened 2020-03-26
Update material editor for supporting packages and scm materials

kritika-singh3:
Issue: #6737, https://github.com/gocd/gocd/pull/7904#issuecomment-603317125

Description: added support to add reference of a pkg/scm material to the pipeline
Show case error message
1. Packages
 - no package repo plugins are installed
 - no package repositories defined
 - if the selected pkg repo has no valid plugin
 - if there are no pkgs defined for the selected pkg repo
2. Pluggable SCMs
 - no scm plugins are installed
 - no scms defined
 - if there are no scms defined for the selected plugin

Preview:

Screen Shot 2020-03-27 at 12 34 08 AM

Screen Shot 2020-03-27 at 1 30 29 AM

Screen Shot 2020-03-27 at 1 30 35 AM

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
@@ -572,4 +572,8 @@ export class SparkRoutes {
   static scmUsagePath(scm_name: string): string {
     return `/go/api/admin/scms/${scm_name}/usages`;
   }
+
+  static pluggableScmSPA() {
+    return '/go/admin/scms';
+  }
 }
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/spec/pluggable_scm_crud_spec.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/spec/pluggable_scm_crud_spec.ts
@@ -35,7 +35,7 @@ describe('PluggableScmCRUDSpec', () => {
       expect(scms).toHaveLength(1);
       expect(scms[0].id()).toBe('scm-id');
       expect(scms[0].name()).toBe('pluggable.scm.material.name');
-      expect(scms[0].pluginMetadata().id()).toBe('github.pr');
+      expect(scms[0].pluginMetadata().id()).toBe('scm-plugin-id');
       expect(scms[0].pluginMetadata().version()).toBe('1');
 
       expect(scms[0].configuration().count()).toBe(1);
@@ -60,7 +60,7 @@ describe('PluggableScmCRUDSpec', () => {
 
       expect(packageRepository.id()).toBe('scm-id');
       expect(packageRepository.name()).toBe('pluggable.scm.material.name');
-      expect(packageRepository.pluginMetadata().id()).toBe('github.pr');
+      expect(packageRepository.pluginMetadata().id()).toBe('scm-plugin-id');
       expect(packageRepository.pluginMetadata().version()).toBe('1');
 
       expect(packageRepository.configuration().count()).toBe(1);
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/types.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/types.ts
@@ -19,7 +19,18 @@ import {JsonUtils} from "helpers/json_utils";
 import {SparkRoutes} from "helpers/spark_routes";
 import _ from "lodash";
 import Stream from "mithril/stream";
-import {DependencyMaterialAttributesJSON, GitMaterialAttributesJSON, HgMaterialAttributesJSON, MaterialAttributesJSON, MaterialJSON, P4MaterialAttributesJSON, PackageMaterialAttributesJSON, PluggableScmMaterialAttributesJSON, SvnMaterialAttributesJSON, TfsMaterialAttributesJSON,} from "models/materials/serialization";
+import {
+  DependencyMaterialAttributesJSON,
+  GitMaterialAttributesJSON,
+  HgMaterialAttributesJSON,
+  MaterialAttributesJSON,
+  MaterialJSON,
+  P4MaterialAttributesJSON,
+  PackageMaterialAttributesJSON,
+  PluggableScmMaterialAttributesJSON,
+  SvnMaterialAttributesJSON,
+  TfsMaterialAttributesJSON,
+} from "models/materials/serialization";
 import {Errors} from "models/mixins/errors";
 import {ErrorsConsumer} from "models/mixins/errors_consumer";
 import {ErrorMessages} from "models/mixins/error_messages";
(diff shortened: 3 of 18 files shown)


kritika-singh3:
Update Preview:

pkg-updated-preview

GaneshSPatil:
@kritika-singh3 - I might be late to the party :(

With the Package Repo and Package Material fields shown side by side, it makes it difficult to read these configurations as the sometimes the config values are split across two lines. Example: `Id`, `Description`  and `URL` in this case.

I would prefer rendering one key=value in 100% width.

We can render the `Package Repository` configurations under its dropdown,  **_Collapsed by default_**.
And, similarly we can render the `Package` configurations under the `Package` dropdown, **_Collapsed by default_**.

---

Another Thought,

With plugins providing the view for the configuration, rendering the config in a key-value view is not always the best idea, as the plugin might contain **_optional_**, **_mutually-exclusive_** and sometimes **internal**  config fields. And GoCD Server does not have this knowledge, to render them as per their importance.
An example of such plugin view is: `Kubernetes Elastic Agent Plugin Elastic Configuration View`, where there are `optional`, `mutually-exclusive` and `internal` fields too. Not sure if there is a package-repo plugin with these complexities :)

So, can we instead make use of the plugin view itself and render it in read-only format? And the view can be exapnd-collapsible, collapsed by default.

This is just a thought, can be ignored, if we've had these conversations. :)

kritika-singh3:
@GaneshSPatil You are not! 😄 
>We can render the Package Repository configurations under its dropdown, Collapsed by default.
And, similarly we can render the Package configurations under the Package dropdown, Collapsed by default.

Any reason for it being collapsable? IMO the widget is required to add the selected package/scm as a material to the pipeline. The configs are provided as readonly. Making them collapsable does not seem right to me.

> So, can we instead make use of the plugin view itself and render it in read-only format? And the view can be exapnd-collapsible, collapsed by default.

This is a question which differs from apps to apps. Some UX guide suggest (which I agree with) that showing readonly info in a disabled input raises the question as to why they are disabled - is there some switch which, when turned on, makes them editable or did the user make some mistake.

Hence, I'm not in favour of disabled/readonly input fields. But if my point does not seem valid for our app, will make that change.

cc:\ @streisguth @arvindsv @maheshp

arvindsv:
Yes, in an ideal world, disabled would give the wrong impression and we should replace it with a proper set of keys / values. Your point about UX is well made.

But, the point Ganesh made about the k8s plugin and ending up showing conflicting values is fair too. That might confuse users even more.

Maybe we can call out that it is readonly in some other way or make it clear somehow? I can't think of any way at this time. Maybe the others have ideas.

maheshp:
Does it make sense to use some kind of display value, similar to [getConfigForDisplay](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/domain/packagerepository/PackageRepository.java#L230) . Also provide a link to view/edit the package repository.

kritika-singh3:
@maheshp The said method `getConfigForDispaly` outputs a single string with non-secure configuration.
E.g. `Repository: [repo_url=https://npm.root.com]` (You can view the same by adding a package to a pipeline and viewing the rails based pipeline edit page).

Screen Shot 2020-03-31 at 10 12 17 AM

I would prefer a key-value combination over a single string.

>Also provide a link to view/edit the package repository.

Will a link to edit the package repository suffice or link to view is needed as well?

streisguth:
I agree with @kritika-singh3 about not having read-only inputs display the same way as disabled inputs. Can we append 'Read Only' to the title of a disabled selection?

Example:
Screen Shot 2020-03-31 at 12 54 59 PM

One more piece of feedback: there should be a break in the lines that span across both dropdown sections (highlighted below).

Screen Shot 2020-03-31 at 12 57 21 PM

kritika-singh3:
@streisguth The sections are disabled due to there being no options to select from.
E.g. there are no package repositories configured, or for a selected one, there are no packages to select.
IMO, in such cases, the `Read only` tag will not make sense because there may be no configurations to show. WDYT?

arvindsv:
I'm slowly leaning towards a key-value pair as shown above. One other option I'm considering is:



In this way, it is clear that the view is supposed to be read-only and it allows the rich plugin view that might come from the plugin. Otherwise, key-value pair might be better.

kritika-singh3:
Update Preview:

update-material-editor

kritika-singh3:
Preview on Add new pipeline wizard:

update-pipeline-wizard

arvindsv:
@kritika-singh3 I think it will be nice to add a link to create a new SCM or package to this screen / component.

Create_New

It would link to the corresponding page. It's possible a person doesn't have permission to go there and I think that's ok.

To think about: If that link opens in a new page and they create one and come back, is it possible to show the newly created package / SCM or do they need to refresh. Either is ok, but the first is preferable, of course.

kritika-singh3:
@arvindsv After creation of new package/scm, the users would need to refresh the page.
IMO creation of package repositories, packages (not sure about pluggable scms) are usually part of initial setup configuration. If we want to update the wizard as they are created - it would require us to poll the get all APIs for them. Not sure if that is a good idea network wise.

arvindsv:
@kritika-singh3 Ok. Understood. Let's then link to the page and expect needing a page refresh? It's better than not having that link there, imo.

kritika-singh3:
@arvindsv Sounds good to me!

mrmanc:
> It would link to the corresponding page. It's possible a person doesn't have permission to go there and I think that's ok.

This looks better than where we are now. What permissions would be required for this? I’m hoping that the average user will be able to create new materials, otherwise they won’t be able to self-create their pipelines.

arvindsv:
@mrmanc All pipeline group administrators can create them. It's not (yet?) incorporated into the [granular policy](https://docs.gocd.org/current/configuration/policy_in_gocd.html), and so it can't be controlled _which materials_ they will be able to control, but you just need to be a pipeline group admin.

arvindsv:
Given that: My point about "It's possible a person doesn't have permission to go there" is moot. View users won't be seeing the pipeline creation page anyway.

@kritika-singh3 asked me to say: Group admins or system admins can manage materials. If security is disabled, everyone is a system admin anyway.

kritika-singh3:
Preview:

1. No plugins installed

Screen Shot 2020-04-10 at 9 43 05 AM

Screen Shot 2020-04-10 at 10 23 15 AM

2. SCM selected

update-pipeline-wizard-scm

kritika-singh3:
3. No package repos defined

Screen Shot 2020-04-10 at 10 26 03 AM

4. Package selected

update-pipeline-wizard-package

maheshp:
There are further changes to this PR in - https://github.com/gocd/gocd/pull/8021

arvindsv:
@maheshp I've marked https://github.com/gocd/gocd/issues/7022 and https://github.com/gocd/gocd/issues/2947 as QA Done, based on this PR. Assuming you checked that behaviour as well.

maheshp:
There are changes to the new PipelineConfig SPA which needs to be verified, which can be done later as this PR is marked as part of the PipelineConfig Rewrite project as well.

PR #7933, opened 2020-03-29
Update the material editor for supporting SCM materials

kritika-singh3:
Issue:  #6737
GoCD PRs to wait for: #7919, #7920

Description: for supporting selection of pluggable scms as material

Preview when no scm plugins are installed: 

Screen Shot 2020-03-27 at 9 55 22 PM

General Preview:

scm-materials

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts
@@ -573,4 +573,8 @@ export class SparkRoutes {
   static scmUsagePath(scm_name: string): string {
     return `/go/api/admin/scms/${scm_name}/usages`;
   }
+
+  static pluggableScmSPA() {
+    return '/go/admin/scms';
+  }
 }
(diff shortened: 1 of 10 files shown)


kritika-singh3:
Updated preview:

Screen Shot 2020-03-30 at 1 42 57 PM

GaneshSPatil:
@kritika-singh3 - I might be late to the party :(

With the Package Repo and Package Material fields shown side by side, it makes it difficult to read these configurations as the sometimes the config values are split across two lines. Example: `Id`, `Description`  and `URL` in this case.

I would prefer rendering one key=value in 100% width.

We can render the `Package Repository` configurations under its dropdown,  **_Collapsed by default_**.
And, similarly we can render the `Package` configurations under the `Package` dropdown, **_Collapsed by default_**.

---

Another Thought,

With plugins providing the view for the configuration, rendering the config in a key-value view is not always the best idea, as the plugin might contain **_optional_**, **_mutually-exclusive_** and sometimes **internal**  config fields. And GoCD Server does not have this knowledge, to render them as per their importance.
An example of such plugin view is: `Kubernetes Elastic Agent Plugin Elastic Configuration View`, where there are `optional`, `mutually-exclusive` and `internal` fields too. Not sure if there is a package-repo plugin with these complexities :)

So, can we instead make use of the plugin view itself and render it in read-only format? And the view can be exapnd-collapsible, collapsed by default.

This is just a thought, can be ignored, if we've had these conversations. :)

kritika-singh3:
The commits from this PR has been merged into #7920.

PR #7939, opened 2020-03-30
[Pipeline Config SPA] Define new job

GaneshSPatil:
* Reuse models from job settings tab and tasks tab for defining
  a new job.


kritika-singh3:
@GaneshSPatil 
#### Steps to reproduce

1. Go to jobs tab on stage settings for any pipeline on the new pipeline config spa
2. Click on add new job
3. Give a name.
4. Click save. The save would fail.
5. Give a custom command, say, `ls`.
6. Click save. The save would go through successfully.
7. Click on the newly created job.

#### Expected Results

There will be only one task - `ls`

#### Actual Results

There are two task - both `ls`.

##### Additional info

The number of tasks is doubled every-time there is an error on the new job modal.

kritika-singh3:
In continuation of the steps in the [comment](https://github.com/gocd/gocd/pull/7939#issuecomment-629031613)

8. Go to the jobs tab of the stage for the same pipeline (without refreshing the page).
9. Delete the job created in the above steps.
10. Click on add new job.
11. Add a valid name.
12. Click save.

#### Expected

The save will fail with an error on command.

#### Actual

The save goes through with no error but no new job is added anywhere.

kritika-singh3:
The issue mentioned above was fixed in #8148

PR #7940, opened 2020-03-31
Update Package Repositories SPA

kritika-singh3:
Issue: #7333 

Description:

 - Added field for updating `id` during add/clone for package repo and package
 - showing `id` as part of configuration for both
 - updated title when actions buttons are disabled

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/package_repositories/package_modal_body.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/package_repositories/package_modal_body.tsx
@@ -69,6 +69,15 @@ export class PackageModalBody extends MithrilViewComponent<Attrs> {
         </Form>
       </FormHeader>
 
+      <div>
+        <TextField label="Id"
+                   readonly={vnode.attrs.disableId}
+                   property={vnode.attrs.package.id}
+                   placeholder={"Enter a unique id for this package"}
+                   helpText={"Each package is associated with an id by which it can be referenced in the pipelines. Once defined, cannot be updated"}
+                   errorText={vnode.attrs.package.errors().errorsForDisplay("id")}/>
+      </div>
+
       <PluginView pluginSettings={(pluginInfo.extensions[0] as PackageRepoExtension).packageSettings}
                   configurations={vnode.attrs.package.configuration()}/>
     </div>;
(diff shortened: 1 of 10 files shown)


maheshp:
## Repository

[Tested with yum and deb plugins]

#### Cloning packages
1. Do we need to give an option to change the plugin when cloning, since the configuration between plugins is never same.

#### View
1. Should all the packages be expanded by default. Does it make sense to keep it consistent with the Elastic Configuration Page. 
2. Is the `Name` key required on the listing view? 
3. There is too much spacing between the Key and the colon:

#### Create Package repository
1. Is there is specific reason for exposing the id and providing an option to edit it, can this be internal.  
2. The reset button only resets the password it seems a bit confusing. I assumed it would reset the entire form.

## Packages
#### Create Package
1. Why do we need an option to select a repository when creating a package, the package is created in context of the repository?
2. A package does not associate to the selected repository from the dropdown.
   - Create repositories rep1 and rep2
   - Click on create Package in rep1
   - Select rep2 in the dropdown
   - Package is created under rep1
3. Is there is specific reason for exposing the id and providing an option to edit it, can this be internal.  
4. Selecting a package from a different plugin does not show the appropriate config
   - Create repositories rep1(yum) and rep2(deb)
   - Click on create Package in rep1
   - Select rep2 in the dropdown
   - The form shown below is still for yum plugin

#### Delete Packages
1. Deleting a package if used by the pipeline does not show errors in the modal. Is it possible to show this error on the modal?

#### Cloning packages
1. Cloning packages should allow selecting repository for the same plugin type.

//cc @kritika-singh3

kritika-singh3:
@maheshp 
[Tested with gem and npm plugins]

> 1. Do we need to give an option to change the plugin when cloning, since the configuration between plugins is never same.

Not sure about this. This approach is consistent will all the SPAs created (elastic agent, auth config, etc).

> 1. Should all the packages be expanded by default. Does it make sense to keep it consistent with the Elastic Configuration Page. 

All the package repos are expanded. The packages are collapsed by default. It is consistent with Elastic Config SPA as packages are a part of package repository. https://github.com/gocd/gocd/issues/7333#issuecomment-585563196 

> 3. There is too much spacing between the Key and the colon:

We use a component `KeyValuePair`. The spacing is present in all the SPAs. Not sure if it was intentional. 

> 1. Why do we need an option to select a repository when creating a package, the package is created in context of the repository?

Yes, a package needs a reference to the package repository.

> 1. Cloning packages should allow selecting repository for the same plugin type.

Is there a specific reason?
 
### Fixed

> 1. Is there is specific reason for exposing the repo id (package repo) and id(package) and providing an option to edit it, can this be internal.

Yes, we provide the option of setting an id via API. Seems logical to provide the same here as well. Removed the default settings of id as it is an optional attribute. If not set, the server automatically sets one up.(#7969)

> 2. The reset button only resets the password it seems a bit confusing. I assumed it would reset the entire form.

The reset button is for password only. Fixed the styles along with other bugs.(#7926)

kritika-singh3:
Had a discussion with @maheshp - While the cloning bit is consistent across all SPAs, allowing the plugin/config to change does not make sense. Will raise a PR
 - to disable the plugin field while cloning a package repo
 - to make the repo field readonly while creating a package
 - to restrict the package repo selection to same plugin type while cloning a package

maheshp:

Issues listed in https://github.com/gocd/gocd/pull/7940#issuecomment-610092587 are fixed.

PR #7947, opened 2020-04-01
Add probot/stale app to close stale issues and PR

GaneshSPatil:
Add [probot/stale](https://github.com/probot/stale) Github app.


GaneshSPatil:
Just a note. The `probot/stale` runs run maximum of `30` actions an hour. so it might take a while to process all the existing open issues/pull requests.

PR #7949, opened 2020-04-01
ConfigRepo branching and PR support

marques-work:
This PR encompasses the server-side changes required for branching support.

### Context

The majority of work to support PR/branch support in config repos will be implemented on the plugin-side. Conceptually, a plugin will know how to fetch branches or any other ref (e.g., GitHub PRs) and filter them against a pattern. Combined with template interpolation capabilities, a config repo can render individual pipelines per matching ref, and return the final config to the GoCD server to merge into the active configuration. Initially, we've decided to only apply this this `git` and hosted `git` providers (GitHub, BitBucket, etc). That is not to say it couldn't or shouldn't work for other SCMs.

On a similar note, the only plugin that will support this is the groovy plugin as it can render templates by its intrinsic nature. Support in YAML and JSON plugins may appear later once we prove this works well.

### What this PR introduces

- [x] We now `parse-directory` every time we poll the config repo material, rather than only when we detect a new commit; new refs may have been introduced on the repo, even if no new commits happened on the branch that is watched by the material.
- [x] Support **user-defined properties** on a config repo by adding a UI frontend to allow arbitrary and easy creation of plaintext and encrypted properties; a developer will be able to look up/dereference these values from within the config definition file. A common use case would be to inject API keys/secrets into the config to support polling PR information from a provider like GitHub.
- [x] Error handling for freeform user-defined properties
- [x] Separate section/component to display user-defined properties in the collapsible details
- [x] Add a capabilities key to indicate support for user-defined properties lookup and interpolation; with this, we should only display the UI for defining properties if the plugin supports it

### Related (will probably implement as separate PRs)

- [ ] Modify `git` clone and checkout to fetch arbitrary refs, including those that do not fit `refs/{heads,tags}/*`, such as GitHub pull requests (which are`refs/pull/*/head`)
- [ ] Webhook support to control PR polling
- [ ] Ref/Branch caching mechanism on the plugin side, in combination with the aforementioned webhook to prevent GoCD from polling APIs or reaching out to remote repos too often

### Miscellany

One of the weird implementation details introduced here is the output of a configuration property view model. The view model in and of itself is not what is weird; what is weird is that the input to the view model is a `Configuration` instance, but the output is a plain JS object that is interface-compatible and is a direct representation of the resulting JSON. The reason for this is a bit hard to explain but ultimately has to do with the fact that the state tracked by the view model isn't completely mappable back to a `Configuration` instance.

I plan to fix this in a future PR that refactors `Configuration` into a true `Validatable` instance. I chose not to touch it right now because it's a widely used object and this PR already has a lot of changes. Theoretically, this work will simplify validation, error handling, and this view model once I get to it.

If anyone thinks otherwise that I should introduce these changes alongside this PR, leave a comment and we can discuss it.

**TODO (separate PR)**:

- [ ] Make `Configuration` and `Configurations` actual `Validatable` types

### Running this code yourself

This code makes use of `config_repos` API v4, which (at the time of writing) is behind a route toggle. Add this toggle to your `go.feature.toggles` (sorry, you must restart GoCD and you cannot use the toggle API for this):

```json
{
  "version": "1",
  "toggles": [
    {
      "key": "config_repo_v4",
      "value": true
    }
  ]
}
```

#### Modal entry

image

#### View in collapsible panel

image

### Usage in config file

The idea is to define these, and then be able to `lookup()` in the config file:

```groovy
GoCD.script { script ->

  branches {
    matching {
      pattern = ~/^refs\/pull\/.+/

      githubProvider {
        repo = "marques-work/sample-node"
        authToken = lookup("secret") // <-- user-defined encrypted property
      }

      each { ctx ->
        script.pipelines {
          pipeline("pr-${ctx.branchSanitized}") {
            materials { add(ctx.repo) }
            stages {
              stage('tests') {
                jobs {
                  job('units') {
                    tasks { bash { commandString = 'whoami' } }
                  }
                }
              }
            }
          }
        }
      }
    }
  } 
}
```

### Note for Reviewers

Sorry about the size of this PR (again). I try to make a cohesive narrative by keeping everything related together.

It may be helpful to go through the commits individually. I've tried to provide context and detail in each commit to maintain a cohesive narrative through code, comments, and commit messages.


server/src/main/java/com/thoughtworks/go/server/materials/ConfigMaterialUpdateListener.java
```diff
@@ -73,18 +69,19 @@ public void onMessage(MaterialUpdateCompletedMessage message) {
...
+                // Now, we need to reparse every time to support ref/branch scanning; even if there are no changes to
+                // the current material, new branches/refs could have been created that might change the output of
+                // templated config repo definitions that use this feature.
+                //
+                // In theory, this should be inexpensive, as the working copies of materials are cached on disk. This
+                // will cause more frequent `parse-directory` messages, however. Generally (crosses fingers), evaluating
+                // this is fast, but we may need to consider only merging into the main config if different?
+                //
+                // Open to any better ideas :).
+                updateConfigurationFromCheckout(folder, modification, material);
```
marques-work, line 84: This *should* be inexpensive for repos that do not take advantage of branch support.

For repos that do use branch support, it is up to the plugin to provide some caching of references so as not to pommel the remote repo or API.

server/src/main/webapp/WEB-INF/rails/webpack/models/base/accessor.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/** Higher order function to convert an {@link Accessor} into a {@link SerializableAccessor} by ensuring a `toJSON()` */
+export function serializing<T>(fn: Accessor<T>, thisArg?: any): SerializableAccessor<T> {
+  const f = ((arguments.length > 1) ? fn.bind(thisArg) : fn) as any;
+  if ("function" !== typeof f.toJSON) {
+    f.toJSON = () => f();
+  }
+  return f;
+}
```
marques-work, line 32: You might be thinking at this point, "Why not use a `Stream`?"

There is an edge case where one may find this useful (look at the configuration properties mixin): The `Accessor<T>` may do some magical abstraction when setting a value (in a way that is not supported via `Stream.map()` or `Stream.combine()`, but you still want the getter result to serialize.

This should be a rare case, admittedly, but it can be useful. This function is idempotent.

server/src/main/webapp/WEB-INF/rails/webpack/models/base/accessor.ts
```diff
@@ -0,0 +1,60 @@
...
+    value = initial!;
+  }
+
+  return function accessor(v?: T): T {
+    if (arguments.length) {
+      value = v!;
+    }
+    return value;
+  };
+}
```
marques-work, line 51: Great for when you want a storage/retrieval function that is `Stream-like` without having to litter your code with `__varname` members to make sure it is excluded from serialization.

server/src/main/webapp/WEB-INF/rails/webpack/views/components/rules/configure_rules_widget.tsx
```diff
@@ -45,7 +46,7 @@ export let RuleActions = {
 export type RuleInfos = RuleInfo[];
 
 interface Attrs {
-  rules: Stream<Rules>;
+  rules: Accessor<Rules>;
```
marques-work, line 49: Changed these types here to satisfy the types in the widget that imports this.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/config_repos/modals.tsx
```diff
@@ -39,18 +43,21 @@ import materialStyles from "./materials.scss";
...
+  & { userProps: EntriesVM }
   & { isNew: boolean }
   & RequiresPluginInfos
   & { error?: m.Children }
   & { resourceAutocompleteHelper: Map<string, string[]> };
 
 class MaterialEditWidget extends MithrilViewComponent<EditableMaterial> {
   view(vnode: m.Vnode<EditableMaterial>) {
-    const pluginList = _.map(vnode.attrs.pluginInfos(), (pluginInfo: PluginInfo) => {
+    const { repo, userProps, isNew, error, pluginInfos, resourceAutocompleteHelper } = vnode.attrs;
```
marques-work, line 54: No functional changes here, just tidying the subsequent code to make it a little more readable with less `vnode.attrs` noise.

marques-work:
### Note for Reviewers

Sorry about the size of this PR (again). I try to make a cohesive narrative by keeping everything related together.

It may be helpful to go through the commits individually. I've tried to provide context and detail in each commit to maintain a cohesive narrative through code, comments, and commit messages.

api/api-plugin-infos-v6/src/main/java/com/thoughtworks/go/apiv6/plugininfos/representers/extensions/ConfigRepoExtensionRepresenter.java
```diff
@@ -28,7 +28,10 @@ public void toJSON(OutputWriter extensionWriter, PluginInfo extension) {
 
         extensionWriter.addChild("capabilities", capabilitiesWriter ->
                 capabilitiesWriter.add("supports_pipeline_export", configRepoPluginInfo.getCapabilities().isSupportsPipelineExport())
-                        .add("supports_parse_content", configRepoPluginInfo.getCapabilities().isSupportsParseContent()));
+                        .add("supports_parse_content", configRepoPluginInfo.getCapabilities().isSupportsParseContent())
+                        .add("supports_list_config_files", configRepoPluginInfo.getCapabilities().isSupportsListConfigFiles())
+                        .add("supports_user_defined_properties", configRepoPluginInfo.getCapabilities().isSupportsUserDefinedProperties())
```
GaneshSPatil, line 33: @marques-work - here we're modifying the existing `Plugin Infos API v6`. 
Should we consider adding `Plugin Infos API v7`?

plugin-infra/go-plugin-domain/src/main/java/com/thoughtworks/go/plugin/domain/configrepo/Capabilities.java
```diff
@@ -46,17 +50,35 @@ public void setSupportsParseContent(boolean supportsParseContent) {
...
+    }
+
+    public boolean isSupportsUserDefinedProperties() {
+        return supportsUserDefinedProperties;
+    }
+
+    public void setSupportsUserDefinedProperties(boolean supportsUserDefinedProperties) {
+        this.supportsUserDefinedProperties = supportsUserDefinedProperties;
+    }
+
```
GaneshSPatil, line 68: should we consider using Lombok?
```java
@Getter
@Setter
@EqualsAndHashCode
@Accessors(chain = true)
@NoArgsConstructor(access = AccessLevel.PUBLIC)
@AllArgsConstructor(access = AccessLevel.NONE)
public class Capabilities2 {
    private boolean supportsPipelineExport;
    private boolean supportsParseContent;
    private boolean supportsListConfigFiles;
    private boolean supportsUserDefinedProperties;
}
```

marques-work: @GaneshSPatil I thought about that too. Hesitated to mess with the gradle deps at the time. But, yes, we probably should. All these objects are just `@Data` essentially.

api/api-plugin-infos-v6/src/main/java/com/thoughtworks/go/apiv6/plugininfos/representers/extensions/ConfigRepoExtensionRepresenter.java
marques-work: Yes, you're right. I probably should have moved this to a new `v7`. Curious though, does anyone actually use this API? It feels very, very specialized/niche. I suppose that's a larger discussion though.

marques-work:
@GaneshSPatil I'll create a PR for the items you mentioned above ☝️.

PR #7952, opened 2020-04-01
Use ephemeral auto register keys for elastic agents #000

maheshp:
* The ephemeral keys are short lived and expire based on the system
  property 'gocd.ephemeral.auto.register.key.expiry.millis' defaulted to
  30 mins.
* The ephemeral keys can be used only once, the keys will be revoked upon
  validation.
* Elastic agents can only register using the ephemeral keys, non-elastic
  agents continue to use the autoRegisterKey.

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/server/controller/AgentRegistrationControllerTest.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/server/controller/AgentRegistrationControllerTest.java
@@ -23,10 +23,7 @@ import com.thoughtworks.go.domain.JarDetector;
 import com.thoughtworks.go.helper.AgentInstanceMother;
 import com.thoughtworks.go.plugin.infra.commons.PluginsZip;
 import com.thoughtworks.go.server.domain.Username;
-import com.thoughtworks.go.server.service.AgentRuntimeInfo;
-import com.thoughtworks.go.server.service.AgentService;
-import com.thoughtworks.go.server.service.ElasticAgentRuntimeInfo;
-import com.thoughtworks.go.server.service.GoConfigService;
+import com.thoughtworks.go.server.service.*;
 import com.thoughtworks.go.util.SystemEnvironment;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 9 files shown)


ketan:
If we're indeed going down this route (which I think is a step in the right direction), I'd think we might want to consider doing an auto-register key for each agent. I think this should be simple to do, with the implementation that you currently have. (https://stackoverflow.com/a/3802420). We've used this in a few other places in code to handle some X509 based caches.

arvindsv:
Do you mean a unique auto-register key per agent? If so, we discussed that and the issue is that we don't have any information about the agent UUID, since it is generated on the agent side. So, we don't know which agent to _expect_. Right?

arvindsv:
Of course, we could change the elastic agent contract to pass in an agent GUID and expected auto-registration key and expect the agent to come up with that. However, two issues with that:

1. It's not necessarily much more secure.

2. All existing elastic agents will fail and all plugins will need to be updated.

maheshp:
Few scenarios verified using the docker elastic agent plugin:
1. **Server restart after providing an ephemeral key to the plugin.**
On a server restart the in-memory cache of ephemeral keys are lost. Elastic Agent registration would fail with server responding with a 422 for registration request. The agent should be deleted by the plugin based on configured `Agent auto-register Timeout`. The server logs an error with a message similar to  

```
ERROR [qtp1138410383-27] AgentRegistrationController:242 - Rejecting request for registration. Error: HttpCode=[422] Message=[Elastic agent registration requires an auto-register agent key to be setup on the server. The agentAutoRegisterKey: [dab13d3f-9169-4f82-af7f-122bd28595e9] is either not provided or expired. Agent-id: [ca5d28b4-c9e8-40ff-ac78-a4646d1446ad], Plugin-id: [cd.go.contrib.elastic-agent.docker]] UUID=[adf5cf1a-e08b-4161-b1d2-e4ff3d8e2033] Hostname=[1ab9d80581e5]ElasticAgentID=[ca5d28b4-c9e8-40ff-ac78-a4646d1446ad] PluginID=[cd.go.contrib.elastic-agent.docker]
```

2. **Ephemeral key expires(plugin takes time to bring up an agent)**
Elastic Agent registration would fail with server responding with a 422 for registration request. The agent should be deleted by the plugin based on configured `Agent auto-register Timeout`.
3. **All create agent threads are blocked for some reason.**
The CreateAgent messages added to Queue have a ttl less than the elasticJobStarvationThreshold and hence would expire if it reaches the threshold and a new message with a new ephemeral key would be added to the queue.

server/src/test-fast/java/com/thoughtworks/go/server/controller/AgentRegistrationControllerTest.java
```diff
@@ -383,6 +383,28 @@ public void shouldAutoRegisterElasticAgent() {
...
+        final String token = token(uuid, serverConfig.getTokenGenerationKey());
+
+        when(agentService.isRegistered(uuid)).thenReturn(false);
+        when(goConfigService.serverConfig()).thenReturn(serverConfig);
+        when(agentService.createAgentUsername(uuid, request.getRemoteAddr(), "host")).thenReturn(new Username("some-agent-login-name"));
+        when(ephemeralAutoRegisterKeyService.validateAndRevoke(any())).thenReturn(false);
+
+        String elasticAgentId = "elastic-agent-id";
+        String elasticPluginId = "elastic-plugin-id";
+        AgentRuntimeInfo.fromServer(new Agent(uuid, "host", request.getRemoteAddr()), false, "location", 233232L, "osx");
```
marques-work, line 401: @maheshp Is this line necessary? Looking at its impl, it doesn't seem to have any special side effects that were obvious to me. It doesn't _appear_ to hit any critical paths under test. Its return value is not used.

Also, when commenting out, the test still passes.

maheshp: Yes, it's redundant. I have changed it.

GaneshSPatil:
QA: Following scenarios are verified:

* Elastic Agents registers with ephemeral auto register key.
* An already registered elastic agent reconnects to the server on server restart.
* An elastic agent fails to auto-register if the server restarts after providing the auto-register key.
* Changing the `gocd.ephemeral.auto.register.key.expiry.millis` works as expected.


Question:

1. Should we validate `gocd.ephemeral.auto.register.key.expiry.millis` user input?
Providing a low number can possible cause elastic agents to not work.

* Should we log the value of `gocd.ephemeral.auto.register.key.expiry.millis` while reading it at [EphemeralAutoRegisterKeyService](https://github.com/gocd/gocd/pull/7952/files#diff-4004e49093a28db94774bc0235502bbfR33)?
will it be useful for debugging problems with elastic agents?

arvindsv:
I think we shouldn't expect too many people to mess with that value. If they do, they _should_ know what they're doing. Also: It will log that the key has expired, right? If they've changed it, they will know to expect it, I hope. My opinion: I don't think we need to do more.

PR #7969, opened 2020-04-07
Update Package Repo SPA

kritika-singh3:
Issue: https://github.com/gocd/gocd/pull/7940#issuecomment-610092587

Description:
 - removed name from config listing - added repo id
 - removed the default setting of repo_id/id - it is a optional attribute and should only be set by the user if required - else the server creates one
 - updated the package repo and package modals to not use foundation hax style - this fixed the reset button placement - it is only for the password field not for the whole form
 - updated the plugin view on change of package repo during package add/edit/clone
 - made sure that the pkg is created wrt to the repo selected
 - all the errors on the package repo/package delete will be shown on the modal itself


maheshp:

PR #7970, opened 2020-04-07
Add an annotation that allows API routes to be conditionally mapped to `latest` version via feature-toggle

marques-work:
Useful for developing new API versions that may not be yet finalized for release. Now you don't really have to revert your code only to undo the revert after release.

Also hides config repo v4 from being aliased as `latest`.

#### Note

Toggling a route will require a restart. It won't hot-update due to how we map `latest` aliases.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/support/toggle/FeatureToggleService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/support/toggle/FeatureToggleService.java
@@ -21,7 +21,6 @@ import com.thoughtworks.go.config.exceptions.RecordNotFoundException;
 import com.thoughtworks.go.server.cache.GoCache;
 import com.thoughtworks.go.server.domain.support.toggle.FeatureToggle;
 import com.thoughtworks.go.server.domain.support.toggle.FeatureToggles;
-import javassist.NotFoundException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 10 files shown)


server/src/main/java/com/thoughtworks/go/server/service/support/toggle/FeatureToggleService.java
```diff
@@ -42,25 +42,56 @@ public FeatureToggleService(FeatureToggleRepository repository, GoCache goCache)
...
 
     public boolean isToggleOn(String key) {
-        FeatureToggle toggle = allToggles().find(key);
+        // handle any key, really.
+        FeatureToggle toggle = userToggles().find(key);
+        if (toggle != null) {
+            return toggle.isOn();
+        }
+
+        toggle = allToggles().find(key);
```
marques-work, line 51: It may not be obvious, but `allToggles()` is _not really_ all toggles. It filters out toggles that are not known to `available.toggles`. Because we _may_ have dynamically generated toggle keys for route toggles, we need to account for these in case users set them in `<configDir>/go.feature.toggles`.

Thus, we first search `userToggles()` alone before defaulting to `allToggles()`.

maheshp: @marques-work I realized we cannot change the value for a dynamic toggle key. If you try to update the dynamic toggle key `config_repo_v4` using the feature_toggles API it fails.
The reason for the failure is this check - https://github.com/gocd/gocd/blob/3423ef97167847e1f57af32ee157e405a205ebcf/server/src/main/java/com/thoughtworks/go/server/service/support/toggle/FeatureToggleService.java#L102

maheshp:

**Note:** 
1. The dynamic toggles cannot be set using the the Feature Toggle API because the API expects the toggle key to be part of the known keys in available.toggles.
2. To change the dynamic toggle users will have to update the `config/go.feature.toggles` with the appropriate value and the GoCD server needs to be restarted for changes to take affect. e.g
```json
{
  "version": "1",
  "toggles": [
    {
      "key": "config_repo_v4",
      "value": true
    }
  ]
}
```

cc: @marques-work

Issue #7975, opened 2020-04-07
Duplicating a package repo does not work (sometimes?)


arvindsv:
##### Issue Type

- Bug Report

###### Basic environment details

* Go Version: 20.3.0 (11467-1fb7c3cc3a71abb71e827f5bbdc2cb139edcd660).
* Browser vendor and version (if relevant): Both Firefox 74.0.1 and Chrome Version 80.0.3987.149.

##### Steps to Reproduce

1. Get the Nuget poller plugin (or any plugin really): https://github.com/gocd-contrib/go-nuget-poller-plugin-2.0/releases/tag/1.0-62-exp

2. Use this config:

    ```
    <repository id="92bbb3b6-3919-45f4-8e02-6db6a0d0e996" name="repo1">
      <pluginConfiguration id="nuget" version="" />
      <configuration>
        <property>
          <key>REPO_URL</key>
          <value>https://www.nuget.org/api/v2/</value>
        </property>
        <property>
          <key>USERNAME</key>
        </property>
        <property>
          <key>PASSWORD</key>
        </property>
        <property>
          <key>PROXY</key>
        </property>
        <property>
          <key>TIME_ZONE</key>
        </property>
        <property>
          <key>LATEST_VERSION_TAG</key>
        </property>
      </configuration>
      <packages>
        <package id="d5a60d9a-e445-4260-b6a2-1d1d6fa34c85" name="pack1" autoUpdate="false">
          <configuration>
            <property>
              <key>PACKAGE_ID</key>
              <value>Newtonsoft.Json</value>
            </property>
            <property>
              <key>POLL_VERSION_FROM</key>
            </property>
            <property>
              <key>POLL_VERSION_TO</key>
            </property>
            <property>
              <key>INCLUDE_PRE_RELEASE</key>
            </property>
          </configuration>
        </package>
      </packages>
    </repository>
    ```

3. Try to clone **that package repository** (not package) through the UI.

##### Expected Results

It should clone the repo.

##### Actual Results

Nothing happens. Seems like there is no network call either. Some silent validation failure?

arvindsv:
This config also does not work:

```xml
    <repository id="4a6d9a76-0bcd-4858-8c6b-f3f956b938bd" name="repo1">
      <pluginConfiguration id="maven-repo" version="" />
      <configuration>
        <property>
          <key>REPO_URL</key>
          <value>https://repo1.maven.org/maven2</value>
        </property>
        <property>
          <key>USERNAME</key>
[...]

PR #7980, opened 2020-04-08
Fix package clone issue

kritika-singh3:
Issue: #7975, https://github.com/gocd/gocd/pull/7940#issuecomment-610739973

Description:
 - Fixed silent failure on client side while cloning a package repo - issue was that individual packages were also validated which failed. Since, there was no client side binding, the error were not represented - removed the package validation from package repo model.
 - disabled the plugin field during edit/clone of package repository
 - disabled the package repo field during creation as it should be within the context of package repo
 - disabled the package repo field during edit as you cannot update the package repo of a package
 - during clone of package repo - only show the repos which belong to the same plugin


arvindsv:
@kritika-singh3 Why was the package validation failing? And, is it ok that the validation no longer happens?

kritika-singh3:
The package validation was failing because it validated that every package needs to have a package repo id which wasn't present. It was a bit of overdrive.

Yes, it is perfectly ok as during create/update of package repository, packages are not in play.

maheshp:
GoCD Version: 20.3.0 (11490-3423ef97167847e1f57af32ee157e405a205ebcf)

PR #7987, opened 2020-04-09
Update package repo and scm spa

kritika-singh3:
Issue: https://github.com/gocd/gocd/pull/7919#issuecomment-610483965, 

Description:
 - updated the regex error message to say name instead of id
 - removed the repo id field from package repo modals
 - removed the id field from package modals
 - made the package repo widget collapsible (collapsed by default)
 - added a warning icon on package repo/scm panel when the plugin is missing

Screen Shot 2020-04-09 at 1 03 22 PM
Screen Shot 2020-04-09 at 1 57 30 PM


maheshp:
GoCD Version: 20.3.0 (11490-3423ef97167847e1f57af32ee157e405a205ebcf)

PR #7989, opened 2020-04-09
Add scm spa endpoint to authorize filter chain

kritika-singh3:
The Pluggable SCM can be viewed by super admin or group admin.


maheshp:
Verified On GoCD Version: 20.3.0 (11498-820055ee4db23c131ff954b054cfb36a564da9fc).

PR #7990, opened 2020-04-09
Fix the error on unauthorized access to scm spa

kritika-singh3:
Description:
The earlier PR rendered the error as if the request was an API - added necessary headers.
But the endpoint is an SPA - only the generic error is enough. This PR makes that change


maheshp:
Related to https://github.com/gocd/gocd/pull/7989

PR #7992, opened 2020-04-09
Invalidate User Toggles cache on change of value

maheshp:
* Related to changes introduced as part of #7970
* #7970 introuduced USER_DEF_TOGGLES cache which was not invalidated
  on change of toggles. Hence any changes to toggle was not reflected
  because of the stale cache.

Issue: #

Description:


marques-work:
LGTM

Issue #7993, opened 2020-04-09
Pluggable SCM SPA issues


maheshp:
##### Issue Type

Placeholder for all issues related to the new PluggableSCM SPA

**Tested using github.pr plugin**

While https://github.com/gocd/gocd/pull/7920 provides an option to select a Pluggable SCM material while adding a material on the pipeline wizard, the old PipelineConfig edit page still does not allow it. This means a material added using the new SPA cannot be used while editing the PipelineConfig. I know this option will be provided on the new PipelineConfig edit page. **[No action required for now]**

View
- [x] Password does not see to be encrypted and is showing in clear text in the key:value view
- [x] Hover on icons broken

Create
- [x] Check connection missing

Usages
- [x] Wrong message - "No usages for package '<material_name >' found."

maheshp:
Verified on GoCD Version: 20.3.0 (11521-f0170986bb47797b6ccf130e1ccb87736b45b372).

PR #7999, opened 2020-04-10
Fixing Pluggable SCM SPA issues

kritika-singh3:
Issue: #7993

Description:
 - Fix password shown as plain text
     - the scm api would not encrypt as secure variable if the value is passed instead of encrypted_value. They would get encrypted when a full config save happens. Fixed that for all versions
 - Fixed broken hover message and no usage found message


maheshp:
@kritika-singh3 I have added few tests for the changes - https://github.com/gocd/gocd/pull/7999

maheshp:

Issue #7994, opened 2020-04-10
Packages Repo SPA Issues


maheshp:
##### Issue Type

Placeholder for all issues related to the new Package Repo SPA

Create
- [ ] Check connection missing

maheshp:
Verified on GoCD Version: 20.3.0 (11498-820055ee4db23c131ff954b054cfb36a564da9fc).

maheshp:
Check connection is required for packages as well, reopening the issue.

maheshp:
Verified on GoCD Version: 20.3.0 (11521-f0170986bb47797b6ccf130e1ccb87736b45b372)

PR #8000, opened 2020-04-10
Package repo spa check connection (#7994)

GaneshSPatil:
Issue: #7994

Description:

Screen Shot 2020-04-10 at 5 12 23 PM
Screen Shot 2020-04-10 at 5 12 59 PM

### DIFF
--- a/api/api-package-repository-v1/src/main/java/com/thoughtworks/go/apiv1/packagerepository/PackageRepositoryInternalControllerV1.java
+++ b/api/api-package-repository-v1/src/main/java/com/thoughtworks/go/apiv1/packagerepository/PackageRepositoryInternalControllerV1.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2020 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.thoughtworks.go.apiv1.packagerepository;
+
+import com.thoughtworks.go.api.ApiController;
+import com.thoughtworks.go.api.ApiVersion;
+import com.thoughtworks.go.api.CrudController;
+import com.thoughtworks.go.api.base.OutputWriter;
+import com.thoughtworks.go.api.representers.JsonReader;
+import com.thoughtworks.go.api.spring.ApiAuthenticationHelper;
+import com.thoughtworks.go.api.util.GsonTransformer;
+import com.thoughtworks.go.apiv1.packagerepository.representers.PackageRepositoryRepresenter;
+import com.thoughtworks.go.apiv1.packagerepository.representers.VerifyConnectionRepresenter;
+import com.thoughtworks.go.config.exceptions.EntityType;
+import com.thoughtworks.go.domain.packagerepository.PackageRepository;
+import com.thoughtworks.go.server.service.EntityHashingService;
+import com.thoughtworks.go.server.service.materials.PackageRepositoryService;
+import com.thoughtworks.go.server.service.result.HttpLocalizedOperationResult;
+import com.thoughtworks.go.spark.Routes;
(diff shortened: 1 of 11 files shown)


maheshp:

kritika-singh3:
@GaneshSPatil Are you planning to add a toggle to this API (to revert back to rails one).

We need to have a check connection for packages and pluggable scms as well, so was thinking of moving all the three checks to `InternalMaterialTestControllerV1` with the following endpoint:
 - `/api/admin/internal/material_test`: for scm materials
 - `/api/admin/internal/repository/material_test`: for package repository
 - `/api/admin/internal/package/material_test`: for package
 - `/api/admin/internal/scm/material_test`: for pluggable scms

Let me know your thoughts.

GaneshSPatil:
> @GaneshSPatil Are you planning to add a toggle to this API (to revert back to rails one).


@kritika-singh3 - The rails and the spark are test connection APIs are on different endpoints.
* Rails Test Connection API:
    - URL: `admin/package_repositories/check_connection`
    - does not require accept version header.
* Spark Test Connection API:
    - URL: `/go/api/admin/internal/repositories/verify_connection`
    - require `v1` (or latest) accept version header.

These APIs are used from different pages, the rails based package repository page and the new package repository SPA, which keeps them separate from usage. Also, the new spark test connection API is under internal namespace. Hence, i feel it should be okay not to have a toggle for this API.

> We need to have a check connection for packages and pluggable scms as well, so was thinking of moving all the three checks to InternalMaterialTestControllerV1 with the following endpoint:

Within GoCD, we have built controllers for entities and not for API methods. Here, we are planning to merge all the test connections (API methods) together in a single controller. 
Note: This is just from a convention perspective, not a rule. :)

Also, not sure if this might make it difficult to look for a class, meaning, today, if ~we~ I want to look for any entity APIs, ~we~ I look for `<ENTITY>Controller<VERSION>` (eg: `PipelineConfigControllerV10`). 

Before that, what are the problems with keeping these APIs on separate controllers?

kritika-singh3:
There is no problem. Just my view of keeping all the check connection APIs in a single place. No other reason.

kritika-singh3:
Also, going by your preferred way of looking for a class - if I wish to see the API which manages connection check my first guess would be `Material` and not package repository or packages or pluggable scms.

GaneshSPatil:
> Also, going by your preferred way of looking for a class - if I wish to see the API which manages connection check my first guess would be `Material` and not package repository or packages or pluggable scms.

Hmm. Hence in [my previous comment](https://github.com/gocd/gocd/pull/8000#issuecomment-612754216), i had to strike through `we` and write `I` 😄

kritika-singh3:
So what is your final vote?

GaneshSPatil:
> So what is your final vote?

I would prefer them separate from a consistency perspective. As mentioned in [this](https://github.com/gocd/gocd/pull/8000#issuecomment-612754216) comment.

kritika-singh3:
Ok. Thanks.
@maheshp @marques-work @arvindsv Would like to know your views as well.

maheshp:
Is there a reason why these endpoints are internal. I would prefer documenting them and moving these endpoints to appropriate controllers. We might have to bump up the version though.

kritika-singh3:
No reason. The original ones (rails) were namespaced under internal. Hence, were keeping the same. We can make public and document them. Sounds ok to me.

GaneshSPatil:
@maheshp - The existing check connection APIs are under internal namespace

Example:
1. [Material Check Connection](https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/rails/config/routes.rb#L172-L174)
2. [Auth Config Test Connection](https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spark_routes.ts#L296-L298)

Hence, the new ones too are added under internal namespace. 

I don't know if there was a reason for these APIs to be internal. If we choose to make these APIs public, from the consistency perspective, we should make all these check connection APIs public.

arvindsv:
I don't know what use a check connection API will be for a user looking at GoCD's APIs. It seems like a niche operation for the UI.

maheshp:
> I don't know what use a check connection API will be for a user looking at GoCD's APIs. It seems like a niche operation for the UI.

Yeah, I understand its currently used only through the UI. I was thinking it might be useful for users who prefer to configure Pipelines using API's over UI.

maheshp:
> I don't know if there was a reason for these APIs to be internal. If we choose to make these APIs public, from the consistency perspective, we should make all these check connection APIs public.

Ok, in that case for the original question of having one controller for all the material check connections, I think we should have appropriate controllers for each material type. Something that we have done in this PR and https://github.com/gocd/gocd/pull/8005. If there is no strong reason to make this public, lets keep it internal for now.

PR #8008, opened 2020-04-15
Updated the material popup to disable scm based material type options

kritika-singh3:
Description: Updated the material popup to disable scm based material type options
  - if other materials in the pipeline have a blank destination
  - updated the material editor to disable options

material-blank-destination

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
@@ -21,7 +21,7 @@ import _ from "lodash";
 import Stream from "mithril/stream";
 import {EnvironmentVariableJSON, EnvironmentVariables} from "models/environment_variables/types";
 import {MaterialJSON} from "models/materials/serialization";
-import {Material, MaterialAttributes} from "models/materials/types";
+import {Material, MaterialAttributes, PluggableScmMaterialAttributes, ScmMaterialAttributes} from "models/materials/types";
 import {ValidatableMixin} from "models/mixins/new_validatable_mixin";
 import {Origin, OriginJSON} from "models/origin";
 import {NameableSet} from "./nameable_set";
@@ -99,7 +99,7 @@ export class Timer extends ValidatableMixin {
     }
 
     return {
-      spec: this.spec(),
+      spec:            this.spec(),
       only_on_changes: this.onlyOnChanges()
     };
   }
@@ -143,10 +143,10 @@ export class TrackingTool extends ValidatableMixin {
     }
 
     return {
-      type: "generic",
+      type:       "generic",
       attributes: {
         url_pattern: this.urlPattern(),
-        regex: this.regex()
+        regex:       this.regex()
       }
     };
   }
@@ -280,5 +280,21 @@ export class Materials extends Array<Material> {
     if (index > -1) {
       this.splice(index, 1);
     }
+  }
+
+  scmMaterialsHaveDestination(): boolean {
+    const hasScmMaterialWithEmptyDestination          = this
+      .filter((material) => material.attributes() instanceof ScmMaterialAttributes)
+      .some((material) => {
+        const attrs = material.attributes() as ScmMaterialAttributes;
+        return _.isEmpty(attrs.destination());
+      });
+    const hasPluggableScmMaterialWithEmptyDestination = this
+      .filter((material) => material.attributes() instanceof PluggableScmMaterialAttributes)
+      .some((material) => {
+        const attrs = material.attributes() as PluggableScmMaterialAttributes;
+        return _.isEmpty(attrs.destination());
+      });
+    return !hasScmMaterialWithEmptyDestination && !hasPluggableScmMaterialWithEmptyDestination;
   }
 }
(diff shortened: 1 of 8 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
```diff
@@ -281,4 +281,20 @@ export class Materials extends Array<Material> {
...
     }
   }
+
+  scmMaterialsHaveDestination(): boolean {
+    const hasScmMaterialWithEmptyDestination          = this
+      .filter((material) => material.attributes() instanceof ScmMaterialAttributes)
+      .some((material) => {
+        const attrs = material.attributes() as ScmMaterialAttributes;
+        return _.isEmpty(attrs.destination());
+      });
```
GaneshSPatil, line 291: this would iterate the collection twice.
How about moving the `material.attributes() instanceof ScmMaterialAttributes` check inside `some`?

server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
```diff
@@ -281,4 +281,20 @@ export class Materials extends Array<Material> {
...
+    const hasPluggableScmMaterialWithEmptyDestination = this
+      .filter((material) => material.attributes() instanceof PluggableScmMaterialAttributes)
+      .some((material) => {
+        const attrs = material.attributes() as PluggableScmMaterialAttributes;
+        return _.isEmpty(attrs.destination());
+      });
```
GaneshSPatil, line 297: this would iterate the collection twice.
How about moving the `material.attributes() instanceof ScmMaterialAttributes` check inside `some`?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/clicky_pipeline_config/modal/material_modal.tsx
```diff
@@ -71,9 +72,18 @@ export class MaterialModal extends Modal {
   }
 
   body(): m.Children {
+    const allScmMaterialsHaveDestination = this.materials().scmMaterialsHaveDestination();
+    const maybeMsg                       = this.isNew && !allScmMaterialsHaveDestination
+      ? <FlashMessage type={MessageType.warning} dataTestId={"materials-destination-warning-message"}>
+        In order to configure multiple SCM materials for this pipeline, each of its material needs have to a 'Destination Directory' specified. Please
+        edit the existing material and specify a 'Destination Directory' in order to proceed with this operation.
+      </FlashMessage>
+      : undefined;
```
GaneshSPatil, line 81: should we use `if` for better readability?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/material_editor_spec.tsx
```diff
@@ -148,4 +148,25 @@ describe("AddPipeline: Material Editor", () => {
...
+          .forEach((option) => {
+            if (option.hasAttribute('disabled')) {
+              disabled.push(option.textContent!);
+            } else {
+              enabled.push(option.textContent!);
+            }
+          });
+
+    expect(enabled).toEqual(["Another Pipeline", "Package"]);
+    expect(disabled).toEqual(["Git", "Mercurial", "Subversion", "Perforce", "Team Foundation Server", "SCM"]);
```
GaneshSPatil, line 170: Generally, we should avoid writing loops and logic in the tests. Or else we will need to write a test for the test too :)

Here, can we write multiple assertions, one for each material type?

kritika-singh3: Why would you write a test for looping on a list of `option`? There is no logic here.

server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
kritika-singh3: IMO, the iteration over the collection twice is the worst case scenario. Looping over collection twice is not necessary a performance issue. I factored readability before it.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/clicky_pipeline_config/modal/material_modal.tsx
kritika-singh3: Is the multi-line message an issue or the involvement of two boolean variables?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/spec/material_editor_spec.tsx
GaneshSPatil: The test includes a loop and an a condition. Isn't that logic?

I mean, its just a good practice to keep the test simple, which reduces the possibility of bugs in the tests itself.

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/clicky_pipeline_config/modal/material_modal.tsx
GaneshSPatil: Umm, in this particular scenario, 
the condition has two parts, `this.isNew && !allScmMaterialsHaveDestination` which itself was complex to understand.
Also, the ternary contains `undefined` as one of the value evaluation, Where in javascript, if a variable isn't initialized, by default it has `undefined` value. Which makes the `undefined` part of the ternary redundant.

server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/pipeline_config.ts
GaneshSPatil: Sure, everyone will have their own preferences of writing code. :)

In this particular scenario, we are iterating over the collection 4 times in worst case. Best case, still has to do 2 iterations. Where, everything can be achieved in just 1 iteration.

arvindsv:
People might still get confused about _why_ it is disabled. I don't have a good suggestion for it, though.

PR #8014, opened 2020-04-15
Fix test failures

kritika-singh3:
Issue: #8013

### DIFF
--- a/config/config-api/src/test/java/com/thoughtworks/go/config/materials/AbstractMaterialConfigTest.java
+++ b/config/config-api/src/test/java/com/thoughtworks/go/config/materials/AbstractMaterialConfigTest.java
@@ -15,87 +15,85 @@
  */
 package com.thoughtworks.go.config.materials;
 
-import com.thoughtworks.go.config.*;
+import com.thoughtworks.go.config.CaseInsensitiveString;
+import com.thoughtworks.go.config.PipelineConfig;
+import com.thoughtworks.go.config.PipelineConfigSaveValidationContext;
+import com.thoughtworks.go.config.ValidationContext;
 import com.thoughtworks.go.config.materials.dependency.DependencyMaterialConfig;
 import com.thoughtworks.go.config.materials.mercurial.HgMaterialConfig;
 import com.thoughtworks.go.domain.BaseCollection;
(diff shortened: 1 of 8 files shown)


config/config-api/src/test/java/com/thoughtworks/go/config/materials/AbstractMaterialConfigTest.java
```diff
@@ -15,87 +15,85 @@
...
     @Test
-    public void shouldRecomputeSqlCriteriaAndXmlAttributeMap_whenAttributesChanged() {
+    void shouldRecomputeSqlCriteriaAndXmlAttributeMap_whenAttributesChanged() {
         AbstractMaterialConfig testMaterialConfig = new TestMaterialConfig("foo");
 
         Map<String, Object> sqlCriteria = testMaterialConfig.getSqlCriteria();
         testMaterialConfig.setConfigAttributes(m("bar", "baz"));
-        assertThat(testMaterialConfig.getSqlCriteria(), not(sameInstance(sqlCriteria)));
-        assertThat(testMaterialConfig.getSqlCriteria().get("foo"), is("baz"));
+        assertThat(testMaterialConfig.getSqlCriteria()).isNotEqualTo(sqlCriteria);
```
kritika-singh3, line 61: @GaneshSPatil  I hope the conversion was correct.

GaneshSPatil: @kritika-singh3 - https://github.com/gocd/gocd/pull/8013/files#diff-1ff479f423bf1a1145900a784673600aR52-R53 is the only change which can possibly cause the spec failures. I am not sure why this would fail.

Also, we must bring back `assertThat(testMaterialConfig.getSqlCriteria().get("foo"), is("baz"));` assertion.

kritika-singh3: This is just the conversion from Junit4 assertion to assertj assertions.
That line is still present.

GaneshSPatil: Ahh Okay. Got it. Thanks.

PR #8015, opened 2020-04-15
Introduced SCM verify_connection internal api

maheshp:
### Verify Connection JSON using github.pr plugin
```
curl 'http://localhost:8153/go/api/admin/internal/scms/verify_connection' \
  -H 'Accept:application/vnd.go.cd.v1+json' \
  -H 'Content-Type:application/json' \
  -X POST -d '{
  "id": "scm-id",
  "name": "foobar",
  "auto_update": true,
  "plugin_metadata": {
    "id": "github.pr",
    "version": "1"
  },
  "configuration": [
    {
      "key": "url",
      "value": "https://github.com/gocd/gocd.git"
    }
  ]
 }'
```
Response 
```json
{
  "status" : "success",
  "messages" : [ "Could connect to URL successfully" ],
  "scm" : {
    "id" : "scm-id",
    "name" : "foobar",
    "auto_update" : true,
    "plugin_metadata" : {
      "id" : "github.pr",
      "version" : "1"
    },
    "configuration" : [ {
      "key" : "url",
      "value" : "https://github.com/gocd/gocd.git"
    } ]
  }


```


api/api-internal-scms-v1/src/main/java/com/thoughtworks/go/apiv1/internalscms/InternalSCMsControllerV1.java
```diff
@@ -0,0 +1,108 @@
...
+import spark.Request;
+import spark.Response;
+
+import java.io.IOException;
+import java.util.function.Consumer;
+
+import static spark.Spark.*;
+
+@Component
+public class InternalSCMsControllerV1 extends ApiController implements SparkSpringController, CrudController<SCM> {
```
kritika-singh3, line 46: We are not using this controller for CRUD operations. Does it still make sense to use `CrudController` interface?

maheshp: Yeah, you are right. It's not required, changed it.

PR #8017, opened 2020-04-16
Fix issue with empty destination

kritika-singh3:
Description
 - destination if empty should not be send in the api payload

Issue: If destination is set and then removed - the create pipeline calls fails for invalid file path.

Steps to reproduce on master (before this commit):

1. Open up the dashboard 
2. Click on add new pipeline
3. Set a material url
4. Set a `Alternate Checkout Path`.
5. **Now remove the same**
6. Fill out other details and click save + edit full config.

Expected results: the pipeline is created and a page opens up to edit full config
Actual results: the pipeline does not get created and the errors shows invalid file path

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/spec/types_spec.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/spec/types_spec.ts
@@ -210,4 +210,22 @@ describe("Material Types", () => {
       expect(material.pluginMetadata().errors().errorsForDisplay('id')).toBe('Id must be present.');
     });
   });
+
+  describe('Serialization', () => {
+    it('should serialize dependency materials', () => {
+      const dependencyAttrs = new DependencyMaterialAttributes("name", false, "pipeline", "stage", false);
+
+      expect(Object.keys(dependencyAttrs.toJSON())).not.toContain('destination');
+    });
+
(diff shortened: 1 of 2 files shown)


maheshp:
Verified On GoCD Version: 20.3.0 (11529-74b1c423c6d296d821d10153ec51a8b80f1d04ac).

This should be fixed on the server side as well right, currently the API returns a 500 for the below request,
```
curl 'http://localhost:8153/go/api/admin/pipelines' \
      -u 'admin:badger' \
      -H 'Accept: application/vnd.go.cd.v10+json' \
      -X POST -d '{
  "group": "internal",
  "pipeline": {
    "name": "up43",
    "parameters": [],
    "materials": [
      {
        "attributes": {
          "auto_update": true,
          "destination": "",
          "url": "http://example.com",
          "password": ""
        },
        "type": "git"
      }
    ],
    "stages": [
      {
        "name": "s1",
        "approval": {
          "type": "success",
          "authorization": {
            "users": [],
            "roles": []
          }
        },
        "jobs": [
          {
            "name": "j1",
            "environment_variables": [],
            "tasks": [
              {
                "type": "exec",
                "attributes": {
                  "run_if": [],
                  "command": "ls",
                  "arguments": []
                }
              }
            ],
            "resources": []
          }
        ]
      }
    ]
  }
}'
```

PR #8019, opened 2020-04-16
Pluggable Scm check connection small tweaks

maheshp:
* PluggableSCM verify connection returns 422 for connection failure


kritika-singh3:
LGTM

PR #8021, opened 2020-04-17
Update material editor

kritika-singh3:
Issue: Better UI/UX

Description:
updated-material-editor


GaneshSPatil:
@kritika-singh3 - can we make the scm configuration expand-collapsible, similar to packages?

kritika-singh3:
@GaneshSPatil What about `Id` and `Name` - should we remove it or keep it?

GaneshSPatil:
> @GaneshSPatil What about `Id` and `Name` - should we remove it or keep it?

I would prefer removing them (from this page at least), as a user would want to see only the configurations while configuring material. If in case a user wants `Id` for some reason, the user will visit the `SCMs` page to get that information.

kritika-singh3:
Updated preview:

updated-material-editor

maheshp:
Verified On GoCD Version: 20.3.0 (11529-74b1c423c6d296d821d10153ec51a8b80f1d04ac).

Issue #7816, opened 2020-03-05
[Pipeline Config SPA] [Placeholder] Issues with new pipeline config SPA.


GaneshSPatil:
This is a placeholder issue to document issues with in-development pipeline config SPA.

#### Tabs ready for QA
All Tabs are ready for QA

#### Issues:
Check comments on this PR for issues

**Note:** If an issue needs more description, it is recommended to add a separate Github comment on this issue and create an item under issues section.

arvindsv:
@GaneshSPatil I took a look at the pipeline level of tabs, before I saw this issue, where you'd said which tabs are ready to be looked at. So, I'll leave my notes here for now, so that I can use them when the rest is ready.

This was on: GoCD Version: 20.4.0 (11553-42b340370c6c5fcdfda260c103ba00baaaaa86c1).

- General tab / all tabs: When you're at the bottom of the screen and you click "Save" and there's no screen refresh, it's hard to know whether it saved and was successful. How about a green checkmark / red "x" next to the button too?

- General tab / all tabs: The "Saved successfully!" message going away after a few seconds is a little jarring, since the whole page moves up. :/ Maybe we can't do much about it.

- Materials tab: When you delete a material without an explicit name, the message shown is: "Material '' deleted successfully". The material name shown in the table is the same as the URL (for git, for instance) and should be shown in the message too.

- Materials popup: There is a lot of space for the repository URL, but it is restricted in width and for even a small material URL, you need to scroll. Can it be bigger?

- Materials popup: When you click "Change" to change an encrypted value, it doesn't focus on the text box. It should.

- Materials popup: When you make some change and cancel, it doesn't reset the object behind. So, if you come back to the modal, the change is not really cancelled. And if you try to switch to a different part of the pipeline edit, then it tells you that there is some unsaved change.

- Materials popup: I know we decided to make other materials disabled if a material doesn't have a destination directory. But, it's confusing. That message in yellow which says: "In order to configure multiple SCM materials for this pipeline, each of its material needs have to a 'Destination Directory' specified. ..." is not easily recognizable as the reason for disabling the materials in the "Material Type" dropdown.

- Materials popup: Also the message in yellow mentions 'Destination Directory', but in the materials popup, it is now called "Alternate Checkout Path".

- Materials popup: 

    1. Create a pipeline P1, with stage S1.

    2. Add a material to the pipeline. Choose "Upstream pipeline".

    3. Select "P1" in the auto complete popup thing for "Upstream Pipeline".

[...]

PR #8063, opened 2020-04-28
[Pipeline Config SPA] Show pipeline config save feedback on save button

GaneshSPatil:
Issue: https://github.com/gocd/gocd/issues/7816#issuecomment-618515907

Description:
* Mark green check when pipeline config save is successful.
* Mark red exclamation mark when pipeline config save is failed.

#### Normal Buttons
Screen Shot 2020-04-28 at 4 06 29 PM

#### Buttons when save is in progress
Screen Shot 2020-04-28 at 4 06 36 PM

#### Buttons when save is successful
Screen Shot 2020-04-28 at 4 06 42 PM

#### Buttons when save is failed 
Screen Shot 2020-04-28 at 4 06 51 PM


GaneshSPatil:
@arvindsv - I tried adding these icons next the save button, but could not provide a link between the save button and the icon. Also, when the failure icon was outside of the button, it gave an impression that its clickable/hoverable which will provide the failure information.

Hence, moved both of them on the button itself.

arvindsv:
Looks good to me.

Will users think it's ok to click the button again, if it has a failure icon next to "Save"? Overall, it's better than not having it, in my opinion. It's even better from an accessibility perspective.

GaneshSPatil:
> Will users think it's ok to click the button again, if it has a failure icon next to "Save"? 

Hmm, when the save is in progress, the buttons are disabled, but on save completion (success/failure), they're enabled again. Also, the button on-hover styles gives the effect that its clickable.

PR #8077, opened 2020-04-30
Update materials tab 

kritika-singh3:
Issue: #7816

Description:
Showcase the destination dir blank message on selecting scm materials rather than showing that all the time


arvindsv:
Merging for @kritika-singh3 since some checks aren't running.

PR #8078, opened 2020-04-30
Cleanup rails code

maheshp:
* Removed rails material test API
* Removed rails webhook API
* Removing some dead rails code


kritika-singh3:
Verified:
 - updating the toggles `use_rails_based_material_test_connection_api` and `rails_webhook_api` gives 404
 - the rails based test connections calls the spark based materials test connection and not the rails based

PR #8090, opened 2020-05-04
Additional fixes on materials

kritika-singh3:
Issue: #7816  

Description:
 - updated the mandatory destination dir msg to 'Destination directory is required when a pipeline has multiple SCM materials.'
 - fixed the JS errors on browser when an update of a material throws 422 and then cancel is clicked.
 - made the validation of password on TFS attributes based on a variable
 - mapped the name on error in dependency material
 - added destination to clone method on material attributes
 - Added support for `shallow_clone` for git material as a part of advanced options


kritika-singh3:
Covered by PR #8091

PR #8091, opened 2020-05-04
Pipeline config spa fixes

GaneshSPatil:
Issue: #7816
Description:

- [x] Show job resources autocomplete when tabs are switched.
- [x] Copy over inherited permissions when stage permissions are toggled to specify locally.
- [x] Improve invalid tab name error message.
- [x] Change no artifact store configured message from error to warning.
- [x] Disable add stage when stage reordering is in progress.
- [x] Disable add task when task reordering is in progress.
- [x] Fix template navigation specs.
- [x] Remove password field validation from model.
- [x] Disable Automatic pipeline scheduling when pipeline is using template.
- [x] Use TEMPLATE as header to denote pipeline is extracted from template.
- [x] Improve agent working directory error message.
- [x] Autofocus input when password field is changed.
- [x] Break from switch instead of early return.
- [x] Add help text for run if condition.
- [x] Switch "Run if condition" and "Properties" on Tasks tab.
- [x] Change destination directory message.
- [x] Add Shallow clone option for Git materials.
- [x] Make tfs material password field non mandatory.
- [x] Do not fail when invalid material save is cancelled.
- [x] Persist material destination directory when editing.


server/src/main/webapp/WEB-INF/rails/webpack/views/pages/clicky_pipeline_config/tabs/pipeline/materials_widget.tsx
```diff
@@ -159,7 +164,10 @@ export class MaterialsWidget extends MithrilViewComponent<MaterialsAttrs> {
         break;
       case "plugin":
         const pluginAttrs = material.attributes() as PluggableScmMaterialAttributes;
-        const scmMaterial = vnode.attrs.scmMaterials().find((pkg) => pkg.id() === pluginAttrs.ref())!;
+        const scmMaterial = vnode.attrs.scmMaterials().find((pkg) => pkg.id() === pluginAttrs.ref());
+        if(!scmMaterial) {
+          return {};
```
kritika-singh3, line 169: Instead of early return, do a break.So you don't have to do a check for undefined at line 109

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/scm_material_fields.tsx
```diff
@@ -229,7 +229,7 @@ export class TfsFields extends ScmFields {
       <TextField label="Repository URL" property={mat.url} errorText={this.errs(attrs, "url")} required={true}/>,
       <TextField label="Project Path" property={mat.projectPath} errorText={this.errs(attrs, "projectPath")} required={true}/>,
       <TextField label="Username" property={mat.username} errorText={this.errs(attrs, "username")} required={true}/>,
-      <PasswordField label="Password" property={mat.password} errorText={this.errs(attrs, "password")} required={true}/>,
+      <PasswordField label="Password" property={mat.password} errorText={this.errs(attrs, "password")} required={false}/>,
```
kritika-singh3, line 234: You might have to update the model as well.

kritika-singh3:
Verified on `GoCD Version: 20.4.0 (11700-0f06e975d3cbcb0699beda94c65a2019ab17d8ba)`

Issue:
 - When the automatic scheduling is disabled due to pipeline using a template - the help text or hover text should be `Since this pipeline is based on a template, automatic/manual behaviour is determined by the template (first stage)`

kritika-singh3:
The issue was fixed in https://github.com/gocd/gocd/commit/0bae7d4c283289a318974466946de174c6c89e2a

PR #8100, opened 2020-05-05
use latest notification filter for the preference page

kritika-singh3:
Issue: Fix the error `The resource you requested was not found!` on the preference page

Description:
The page used the notification filter API v1 which was removed.
Updated the same to use `latest`.


maheshp:
Verified on GoCD Version: 20.4.0 (11688-953d1f7453a918a96abc89568829fdeffe101dcf)

PR #8111, opened 2020-05-06
set auto_update to true by default for package

kritika-singh3:
Description: the rails based package creation used to set the value of `auto_update` as true.

On the mithrill based page, it was defaulted to `false`. Changing it to reflect the original behaviour.


maheshp:
Related to https://github.com/gocd/gocd/pull/8118 . The SPA allows `auto_update` to be set now.
Verified on GoCD Version: 20.4.0 (11688-953d1f7453a918a96abc89568829fdeffe101dcf)

PR #8112, opened 2020-05-06
Open view template on click of view template icon

GaneshSPatil:
Issue: #7816

Description:
* As the click event was bubbled up to the parent div. The edit
  template handler was also handling the same click event,
  causing the page to go to template edit instead of view.

* Avoid propagating click event when view template is clicked.


kritika-singh3:
After the changes - clicking on the `Preview` icon takes the user to the templates page. The original behaviour was that a popup opened up on the same page with details of the template.

Was this change intentional? @GaneshSPatil

GaneshSPatil:
@kritika-singh3 -- I did the simplest possible thing. 😬  The readonly view was already implemented at a url. Hence, from pipeline config SPA, we just open the new url. 
Opening it in the same page would require us to call the template modal from the spa.

Issue #4067, opened 2017-12-07
State exhibited by Agent and pipelines in various scenarios


anishvenkat:
##### Issue Type

- Bug Report
##### Summary

Logging an issue for the various state exhibited by the Agent and the pipeline when the Agent is stuck in Building(Cancelled) state.

###### Basic environment details

* Go Version: 17.7
* JAVA Version: 1.8.0_151
* OS: Ubuntu 16.04

**Case 1**: Have a task that does not respond well to being killed. For instance, a config such as this:

> <pipeline name="pipeline2">
      <materials>
        <git url="https://github.com/anishvenkat/python.git" />
      </materials>
      <stage name="defaultStage">
        <jobs>
          <job name="defaultJob">
            <tasks>
              <exec command="bash">
                <arg>-c</arg>
                <arg>trap "" INT TERM HUP; while sleep 2; do date; done</arg>
              </exec>
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>

##### Expected Results
On canceling the task, we expect the Agent to come back into the idle state.

##### Actual Results
On canceling the run, the `On Cancel Task: Kills child process` kicked in and it ignored the SIGTERM and went on a forever loop and the Agent was hung in Building(Cancelled) state. I had to kill the process manually by running - `pkill -9 -f 'INT TERM'`
After which the Agent returned back to idle state.

##### Code snippets/Screenshots
 Console output screenshot:
<img width="1440" alt="screen shot 2017-12-06 at 11 57 12 2" src="https://user-images.githubusercontent.com/11550439/33700639-70d52d0a-db40-11e7-8b27-48b6c6473f55.png">

Agent Dashboard page:
<img width="1440" alt="screen shot 2017-12-07 at 11 21 07" src="https://user-images.githubusercontent.com/11550439/33700698-cb25fb54-db40-11e7-893a-0d8c50995c6c.png">

**Case 2**: Restarting Go server when the Agent is stuck on Building(Cancelled) state

For this case I have used the same pipeline config as in case 1.

##### Expected Results
On restarting Go server, the Agent should have killed the job.

##### Actual Results
I could see that the job was still running after the restart(even though it shows that the pipeline run was canceled). The console output was constantly printing the output. I had to manually kill the job from the command line. 

##### Log snippets
The logs show the below two errors.

1) Error occurred when agent tried to ping server.
2) Could not send console output to server.

##### Code snippets/Screenshots
Pipeline state screenshot:
[...]

PR #8114, opened 2020-05-06
Allow tasks on Agent to be forced killed if stuck in Building(Cancelled) state

maheshp:
Issue: https://github.com/gocd/gocd/issues/4067

When a Job is cancelled, an instruction is sent to the agent to destroy all running tasks. On getting the instruction the agent tries to clean destroy all the running tasks but there might be instances where these tasks are not terminated. While the job is cancelled the agent remains in `Building(Cancelled)` state thereby not available for picking any new jobs. If an agent is stuck in this state there is no notification hence users will not know about it unless they visit the Agents page.
 
This PR  does the follwoing,
1. If an agent is stuck in `Building(Cancelled)` state for more than 10 mins a server health message is added indicating the same.
2. Add a new agents endpoint to send an instruction to`killRunningTasks` on an agent. The agent would **force** kill all running tasks on the agent upon getting this instruction.
3. TODO: Add an ability to `killRunningTasks` from the agents page.


kritika-singh3:
Verified on `GoCD Version: 20.5.0 (11818-b3932994090f442d17d0829804f44554ad04d309)` using the API agents v7

One observation: after the API is posted, there is no indication in the logs that a kill signal is send. It is only after a few seconds that the agent logs says `Stopping transmission`.

It would be a _**nice to have**_ feature: the agent/server logs saying that a kill running task signal is received.

maheshp:
> It would be a nice to have feature: the agent/server logs saying that a kill running task signal is received.

Thanks @kritika-singh3 I will take a look at it.

PR #8132, opened 2020-05-12
 Template API update fails if paramter resolution has errors

maheshp:
* During config save if a template is used by a pipeline, the pipeline is
      preprocessed and parameters are resolved. If there are any errors
      during parameter resolution, the errors are added on the pipeline and
      not the template, hence earlier the update used to go through.
* This PR ensures the template update fails if there are errors in
      preprocessed config.
* Currently just picking up the first error in the preprocessed config
      and adding to the template. This is not optimal but if template is
      associated with multiple pipelines the same error will be populated
      for each pipeline, hence picking up the first error.


kritika-singh3:
Verified on `20.4.0 (11700-0f06e975d3cbcb0699beda94c65a2019ab17d8ba)`

Issue #8119, opened 2020-05-07
Adding a package material via Package Repo page causes the pipeline to fail


kritika-singh3:
##### Issue Type

- Bug Report

##### Summary

Adding a package material a pipeline via the new pipeline config spa fails the pipeline. If added via the old rails based page,  the pipeline is green. 

###### Basic environment details

* Go Version: `GoCD Version: 20.4.0 (11669-dcb1fc4c5a26445c1091b81acc326d7219a53eed)`

##### Steps to Reproduce

1. Add a package repo and package via the Package Repositories SPA. 
2. Use the create pipeline wizard to create a pipeline with the package just created. Keep the task something simple - say `ls`.
3. Trigger the pipeline 

##### Expected Results

The pipeline status should be green

##### Actual Results

The `ls` task fails.

> Note: if you do a full config save and then trigger the same pipeline with no changes - the pipeline will turn green.

##### Any other info

You can use the following config to duplicate the error:

```xml
<repositories>
    <repository id="5ef6a39c-18fd-4240-8843-40e538b1229d" name="repo-1">
      <pluginConfiguration id="yum" version="" />
      <configuration>
        <property>
          <key>REPO_URL</key>
          <value>file:///tmp/packagerepo/file_repo</value>
        </property>
        <property>
          <key>USERNAME</key>
        </property>
        <property>
          <key>PASSWORD</key>
        </property>
      </configuration>
      <packages>
        <package id="58a804c6-f243-4d5f-9ec0-dd4ed3e9b21b" name="go-agent">
          <configuration>
            <property>
              <key>PACKAGE_SPEC</key>
              <value>go-agent</value>
            </property>
          </configuration>
        </package>
      </packages>
    </repository>
  </repositories>
  <pipelines group="defaultGroup">
    <pipeline name="pipeline_with_yum_repo_package_as_materialb327a3e3">
      <materials>
        <git url="/tmp/materials/gitmaterial-1588841060-fdd9f723/sample.git" dest="git" autoUpdate="false" materialName="git" />
        <package ref="58a804c6-f243-4d5f-9ec0-dd4ed3e9b21b" />
      </materials>
      <stage name="defaultStage">
        <jobs>
          <job name="defaultJob">
            <tasks>
              <exec command="echo">
                <arg>GO_REPO_TW_REPO_GO_AGENT_PASSWORD</arg>
              </exec>
              <exec command="/bin/sh">
                <arg>-c</arg>
[...]

PR #7982, opened 2020-04-08
Fix fingerprint generation for SCM and Package materials

maheshp:
Fixes #7974 and #7976

PR #8138, opened 2020-05-13
Send blank string when config value is null

kritika-singh3:
Issue: #8119

Description:
When a config value is not specified - it is stored as null. When it is converted to json to be send to the plugin, the nulls are ignored.
This PR proposes sending blank string instead when a config value is null.

Also, the config values are set as environmental variables. When the config value is not set, the env var value is set as null. This value when send to agent leads to a `NullPointerException` [here](https://github.com/gocd/gocd/blob/master/commandline/src/main/java/com/thoughtworks/go/util/ProcessManager.java#L55). Hence, setting this value as empty string as well.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
@@ -181,7 +181,7 @@ public class ConfigurationProperty implements Serializable, Validatable {
                 throw new RuntimeException(format("Could not decrypt secure configuration property value for key %s", configurationKey.getName()), e);
             }
         }
-        return configurationValue == null ? null : configurationValue.getValue();
+        return configurationValue == null ? EMPTY : configurationValue.getValue();
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


maheshp:
@kritika-singh3 I am just wondering if we are fixing this issue in patches. While this PR fixes the Package materials wouldn't this fail even for Pluggable SCM materials?

On similar lines I had fixed an issue https://github.com/gocd/gocd/pull/7982 with regard to fingerprint.

The ConfigurationProperty as part of [PostContruct](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java#L272) defaults ConfigurationValue to an empty string. But, this PostConstruct is not called if de-serializing from JSON, hence the ConfigurationValue is set as null.

I am leaning towards changing [getValue()](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java#L184) to return an EMPTY string over fixing this is multiple places. WDYT?

kritika-singh3:
@maheshp Yes, you are right. Updated `getValue` method makes more sense. Will make that change.

PR #8140, opened 2020-05-13
Improvement to config repo error message

arvindsv:
The error message is sometimes a little confusing:

```
INVALID MERGED CONFIGURATION
Number of errors: 1+
I. Rule Validation Errors:
       1. Not allowed to refer pipeline group 'PaymentGateway' from the config repository.;;
II. Config Validation Errors:
- For Config Repo: ...
```

It's not always clear that when it says: `Not allowed to refer`, it means that it is related to the `Rules` configured for that config repository. Adding that to the end of the message.

It should now read:

```
Not allowed to refer to pipeline group 'PaymentGateway'. Check the "Rules" of this config repository.
```

I don't see any functional tests with this message.

--------------------------------

Looks like this now:

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/exceptions/EntityType.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/exceptions/EntityType.java
@@ -51,7 +51,7 @@ public enum EntityType {
 
     private final String entityType;
     private final NameOrId nameOrId;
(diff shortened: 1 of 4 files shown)


kritika-singh3:
Verified on `GoCD Version: 20.4.0 (11689-566fce8d309fc7650ffd5152de636c72cd4505e1)`

PR #8156, opened 2020-05-18
Merge PartialConfigs only when they are different from the last attempt

marques-work:
We should only merge partial configs that are actually different in content from the last attempt

This will help prevent excessive config updates/merges now that we parse-directory all the time.


kritika-singh3:
Verified on `20.4.0 (11742-d19089b46ee5d376564103d6d4f77484e111426d)` with the following observations:
 - the configs are parsed on every MDU of the config material
 - the difference is noted - if none found - the merge does not happen
 - the merge only happens if a new revision is found or the partials are different or the config repo was changed

PR #8158, opened 2020-05-19
Use SHA-512/256 for all entity hashing

marques-work:
Trying to move away from `md5`, which is broken WRT cryptographic standards, to SHA-512/256. This isn't all of the usages (e.g., various parts of `CruiseConfig` and artifact checksums still use `md5`, but this is the majority of cases.

### DIFF
--- a/api/api-admins-config-v2/src/main/java/com/thoughtworks/go/apiv2/adminsconfig/AdminControllerV2.java
+++ b/api/api-admins-config-v2/src/main/java/com/thoughtworks/go/apiv2/adminsconfig/AdminControllerV2.java
@@ -118,7 +118,7 @@ public class AdminControllerV2 extends ApiController implements SparkSpringContr
 
     @Override
     public String etagFor(AdminsConfig entityFromServer) {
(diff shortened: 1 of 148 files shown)


arvindsv:
I assume this does not affect the name of the flyweight directory, etc. which can cause problems. It looks like it's for MD5 in API calls and stuff around diffs of config.

arvindsv:
Might be good to consider perf as well. Should be ok, but still.

marques-work:
@arvindsv yes, I don't think it would affect flyweight dirs. Generally, it should only affect APIs (mostly etag generation I believe).

yes, performance is a question. SHA-256 is more computationally expensive, but whether that becomes a choke point (I'm guessing not) is worth observing.

marques-work:
@arvindsv I have read that SHA-512 might be more performant than SHA-256 (esp on 64-bit CPUs). Those hashes are super long though, but I believe there are truncated variants (or we can truncate them ourselves. It would not be difficult to swap this out.

marques-work:
Now, SHA-512/256. Updated title and description to match.

PR #8167, opened 2020-05-21
Do not prematurely evict config repo validation errors from `ServerHealthState`

marques-work:
Stop optimistically removing config repo merge/validation errors from ServerHealthState

The errors in ServerHealthState should only be removed when a) the
partial is marked as valid and b) when destroying a configrepo config.

This commit has a lot of noise from renaming classes and methods, but is
logically very short -- effictively a one line change.

The heart of this fix is removing the line `serverHealthService.removeByScope(...)`
from `GoConfigRepoConfigDataSource#onCheckoutComplete()`.
`CachedGoPartials#markAsValid()` already removes the error from `ServerHealthState`.

This should fix at least one of the regression failures around ConfigRepo validations.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/GoConfigRepoConfigDataSource.java
+++ b/server/src/main/java/com/thoughtworks/go/config/GoConfigRepoConfigDataSource.java
@@ -46,21 +46,21 @@ import static java.util.Collections.synchronizedSet;
  * Parses partial configurations and exposes latest configurations as soon as possible.
  */
 @Component
-public class GoRepoConfigDataSource implements ChangedRepoConfigWatchListListener {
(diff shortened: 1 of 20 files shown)


server/src/main/java/com/thoughtworks/go/config/GoConfigRepoConfigDataSource.java
```diff
@@ -164,7 +165,8 @@ post event about completed (successful or not) parsing
 
                 newPart.setOrigins(new RepoConfigOrigin(repoConfig, modification.getRevision()));
                 this.configReposMaterialParseResultManager.parseSuccess(fingerprint, modification, newPart);
-                serverHealthService.removeByScope(scope);
+
+                // it is the responsibility of the success listeners to clear the ServerHealthState
```
marques-work, line 169: **This is the main change here**, among all the method/class rename noise.

PR #8168, opened 2020-05-22
How to add a new liquibase migration in GoCD

GaneshSPatil:
Description:

This is a test PR which demonstrates how to introduce a new database migration using liquibase.

As this PR is based on #7646 the older commits can be ignored. See [a9d49f384ea56a2ffeeb60b0adba4ef6c19d611d](https://github.com/gocd/gocd/pull/8168/commits/a9d49f384ea56a2ffeeb60b0adba4ef6c19d611d) commit for liquibase migration.

##### 2. Add a new migration under `migrations/` directory, lets say [`migrations/2020_06_1_test_migration.xml`](https://github.com/gocd/gocd/pull/8168/commits/a9d49f384ea56a2ffeeb60b0adba4ef6c19d611d#diff-66eb52e41df9fe48c8c225220339a7cf)
```xml
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">

  <changeSet id="2020_06_1_test_migration" author="gocd">
    <createTable tableName="TempTableForTestPurpose">
      <column name="id" type="int">
        <constraints primaryKey="true" nullable="false"/>
      </column>
      <column name="name" type="varchar(50)">
        <constraints nullable="false"/>
      </column>
    </createTable>
  </changeSet>

</databaseChangeLog>
```
* Specify the migration as part of the chaset.
* Specify `gocd` as the author of the migration. If required, we can include developers name.
* Provide a unique `id` to the changeset. 
**Note:** A changeset is uniquely tagged by both an author and an id attribute (author:id), as well as the changelog file path. As we have a common author `gocd`, the tag is based on `id`.  Also, liquibase uses the changeSet id to calculate the checksum, so its recommended to use a unique value. (Do not auto-generate id, as it will cause changeset recreation every time).

##### 3. Verify the migration by adding a test under [`LiquibaseMigtrationTest.java`](https://github.com/gocd/gocd/pull/8168/commits/a9d49f384ea56a2ffeeb60b0adba4ef6c19d611d#diff-d14e6f579763e692567af5becf50ac98).
**Note:** As of now, the migration tests are running only against `H2` database. As its easier to configure `H2` for unit/integration tests. To verify the functionality against other databases, we should rely upon our functional specs :)


GaneshSPatil:
\cc: @maheshp @arvindsv @kritika-singh3

PR #8177, opened 2020-05-27
Removing empty-task list validation

kritika-singh3:
Description:  as there is no corresponding check at the API side


marques-work:
There may not be a validation on the server side, but I think it makes sense to keep the validation for UX. Originally, it was put in for the UX improvements to "add pipeline". New users would be able to erroneously create jobs that do nothing.

Is there any other reason this was removed? It feels like a bug on the server side to allow taskless jobs. I would rather have chosen to make the server side match this validation.

/cc @arvindsv @maheshp @GaneshSPatil, WDYT?

PR #8203, opened 2020-05-31
Copy specific wrapper files needed to run the gocd-server in a container

varshavaradarajan:
Issue: #000

Description:
* Remove deprecated MAINTAINER instruction
* Reduce number of intermediate layers

### DIFF
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
@@ -54,7 +54,7 @@ enum Distro implements DistroBehavior {
       return [
         '# install glibc and zlib for adoptopenjdk',
         '# See https://github.com/AdoptOpenJDK/openjdk-docker/blob/ce8b120411b131e283106ab89ea5921ebb1d1759/8/jdk/alpine/Dockerfile.hotspot.releases.slim#L24-L54',
-        '  apk add --no-cache --virtual .build-deps curl binutils',
+        '  apk add --no-cache --virtual .build-deps binutils',
         '  GLIBC_VER="2.29-r0"',
         '  ALPINE_GLIBC_REPO="https://github.com/sgerrand/alpine-pkg-glibc/releases/download"',
         '  GCC_LIBS_URL="https://archive.archlinux.org/packages/g/gcc-libs/gcc-libs-9.1.0-2-x86_64.pkg.tar.xz"',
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/resources/gocd-docker-server/Dockerfile.server.ftl
```diff
@@ -28,10 +28,18 @@ RUN \
   curl --fail --location --silent --show-error "https://download.gocd.org/binaries/${fullVersion}/generic/go-server-${fullVersion}.zip" > /tmp/go-server-${fullVersion}.zip
 </#if>
 RUN unzip /tmp/go-server-${fullVersion}.zip -d /
-RUN mv /go-server-${goVersion} /go-server && chown -R ${r"${UID}"}:0 /go-server && chmod -R g=u /go-server
```
arvindsv, line 31: My worry about this is that will silently break if the structure of the zip changes. I tried this locally and the files copied over are 197MB and the files left behind due to this are 17MB. Not sure if the risk is worth that.

Also, there is a LICENSE file there, that would be nice to copy over, if going with this.

buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
```diff
@@ -54,7 +54,7 @@ enum Distro implements DistroBehavior {
       return [
         '# install glibc and zlib for adoptopenjdk',
         '# See https://github.com/AdoptOpenJDK/openjdk-docker/blob/ce8b120411b131e283106ab89ea5921ebb1d1759/8/jdk/alpine/Dockerfile.hotspot.releases.slim#L24-L54',
-        '  apk add --no-cache --virtual .build-deps curl binutils',
```
arvindsv, line 57: I see we use curl later here. ls it not needed?

varshavaradarajan: We have already added curl

buildSrc/src/main/resources/gocd-docker-server/Dockerfile.server.ftl
varshavaradarajan: There are windows, mac, freebsd etc files that are not needed to run on alpine. But, yeah, I guess I'll remove this change if you think that the structure is likely to change.

arvindsv: Not sure if it's likely to change, but I'd like something to fail in an obvious way if it changes ...

arvindsv:
On second thoughts, I think it's fine. I'll merge this soon, once the pipelines are a little greener.

varshavaradarajan:
I'll copy the LICENSE. Thanks, Duck!

buildSrc/src/main/resources/gocd-docker-server/Dockerfile.server.ftl
varshavaradarajan: I guess if something important does change, `docker run` will fail to start the server. There used to be some docker tests we were running to make sure basic things(bringing up a server and all agents) worked. I don't see it on build.go.cd anymore though.

PR #8196, opened 2020-05-28
Improve environments pipeline edit

maheshp:
Fixes Issue: #8127

PR #8209, opened 2020-06-02
Fix issue with Available Pipelines for environment pipelines edit

maheshp:
* This issue was introduced as part of changes in PR
  https://github.com/gocd/gocd/pull/8196

* Available pipelines should not include pipelines for
  environment associated through config repo.


kritika-singh3:
Verified on `GoCD Version: 20.5.0 (11785-0018daff441c43f817cbf6fbe5bea849b5cc43a1)` via build.gocd.

Issue #8223, opened 2020-06-11
[UI Lacking Description] Artifact cleanup options do not specify units


badbort:
##### Issue Type
- Documentation Report

##### Summary

No indication of units for artifact cleanup setting in Artifacts Management

![image](https://user-images.githubusercontent.com/4551792/84355412-2687e980-abf5-11ea-8dc3-e707ad840141.png)

MB? GB

###### Basic environment details

* Go Version: `GoCD Version: 20.4.0 (11749-5553066d29d315951efa5ead0c087374df6338b9).`
* JAVA Version: `13.0.2`
* OS: `Windows 10 10.0`
* Browser vendor and version (if relevant): `Chrome - Version 83.0.4103.97 (Official Build) (64-bit)`

arvindsv:
Thanks, @badbort. It will be fixed in 20.5.0. I think it is GB.

arvindsv:
Sorry, re-opening to verify this later.

arvindsv:
Verified on: 20.5.0 (11818-b3932994090f442d17d0829804f44554ad04d309).

PR #8239, opened 2020-06-15
Update label to mention units in artifacts config UI

kritika-singh3:
Issue: #8223

Description:
 - Updated UI to contain units in the label

Screen Shot 2020-06-15 at 4 28 21 PM


arvindsv:
Looks good to me. Thanks, @kritika-singh3.

Issue #8141, opened 2020-05-14
Pipeline Config SPA - Possible Improvements


GaneshSPatil:
##### Issue Type

- Feature enhancement

##### Description

Following enhancements could be done to the Pipeline Config SPA:

- [x]  An error adding stage shows up on the main page and doesn't go away (This behaviour is same while adding jobs)
  Steps to reproduce
   - Click on add new stage
   - Do not enter anything and click save
   - Errors show up on the Add stage modal
   - Close the modal
   - A global error shows up with the message `Validations failed for template with name                'deploy_template'. Error(s): [Validation failed.]. Please correct and resubmit.` which does not go away.

- [x] Secure environment variable should have an option to view the value while creating/editing it.

- [x] Errors adding fetch aritifact tasks does not show up on the modal
  Steps to reproduce
  - Add a pipeline, stage etc which is not an upstream of any of the pipelines using template
  - The validation errors show up on the main templates page and not the modal
  Add task buttons are disabled on fetch artifact task errors - Fixed  
  - Add fetch artifact task with all attributes as parameters and save it
  - Edit the fetch artifact task and provide an invalid parameter and save it
  - There are validation errors
  - Close the modal
    - The invalid values still show up on the task preview
    - The add task dropdown and button are disabled

**Template/Stage/Job/Job Settings**
- [x] Client side validation for required fields
- [ ]  Elastic profile errors are not mapped to the text field, shows as global errors
   Steps to reproduce
   - Provide a non-existent elastic profile id and save
   - Global error shown, ideally it should be mapped to the input field.
   - Looks like the template API needs to be fixed.

- [x] Run on instances allows a negative number to be selected, provide a min="0"  for the input element.
- [x]  Unable to select resources from autocomplete list
   - Navigate the Job Settings page
   - Move cursor on the resources textbox and enter a value
   - Select a resource from the dropdown(this works) and save.
   - Now move cursor again on the resources textbox and enter a value
   - Try selecting a resource from dropdown, this does not work.

**Template/Stage/Job/Environment Variables**
- [x] Client side validation for mandatory fields

**Template/Stage/Job/Custom Tabs**
- [x] Client side validation for mandatory field, in this case Path cannot be empty if Tab Name is specified.

- [x] Pipeline Stages or Templates (See: https://github.com/gocd/gocd/issues/8141#issuecomment-630626360)

GaneshSPatil:
[...]

PR #8256, opened 2020-06-23
Fixes on new pipeline config spa:

kritika-singh3:
Issue: #8141 

Description:
 - The global error on client validation failure will now go away after 5 secs
 - Added min="0" for run instance count and job timeout
 - Added validation for custom tab path when name is specified

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/spec/tab_spec.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/pipeline_configs/spec/tab_spec.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2020 ThoughtWorks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Tab} from "../tab";
+
+describe('TabSpec', () => {
+  it('should throw error when name is specified but not path', () => {
+    const tab = new Tab("name", "");
+
+    expect(tab.isValid()).toBeFalse();
+    expect(tab.errors().errorsForDisplay('path')).toBe('Path must be present.');
+  });
+
+  it('should not error out if name is not specified', () => {
+    const tab = new Tab("", "");
+
+    expect(tab.isValid()).toBeTrue();
+    expect(tab.errors().count()).toBe(0);
+  });
+});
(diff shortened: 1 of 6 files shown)


GaneshSPatil:
@kritika-singh3 - which issue point in the #8141 will be fixed by `The global error on client validation failure will now go away after 5 secs` ?

kritika-singh3:
@GaneshSPatil The first one: `A global error shows up with the message "Validations failed for template with name 'deploy_template'. Error(s): [Validation failed.]. Please correct and resubmit." which does not go away.`

GaneshSPatil:
@kritika-singh3 -

1. Clearing the global errors after 5 secs may not fix the above-mentioned issue completely. If the user hits save and closes the popup (by clicking `esc`), the error would be visible for 5 secs. Also, such error suddenly disappearing without an interaction might be confusing.

2. Not clearing global errors (on a specific timeout) was an intentional change. Reasons:
  - There is a possibility of multiple global errors and can may require more than 5 secs to read. 
  - Both, global errors and field specific errors are validation errors and should not go away unless user fixes them.

To allow users clearing the global errors, we changed the global errors from normal to dismissible errors, providing users an option to remove them (if the user think it may've been fixed by updating some other field from the form).

Coming back to the original issue,
To fix the above-mentioned error, we should not populate the global errors when user is operating inside a modal. IIRC, the global error holder `flashMessage` has been passed down to the individual widgets and modals. We should add a check to not populate global error when inside the modal context.

kritika-singh3:
@GaneshSPatil Only when the client side validation fails, the global error will be removed after 5 secs.
 
For point 2 - the global errors populated from API calls will remain till the user click close. Also, only the global error goes away. The field validation errors stays. Also, various other info messages also disappear on our UI. I don't support the idea that it is confusing.

Your fix will not work as there are various cases (e.g. adding a fetch artifact task) in which the error are not fully mapped to the fields. In such cases, only the global errors represent the errors from the API call.

GaneshSPatil:
> Also, various other info messages also disappear on our UI. I don't support the idea that it is confusing.

The `save/update/deleted successfully` info messages or errors occur on user interaction. User is aware of the fact that these messages are shown because (s)he performed some action on the page.

When i say "it would be confusing", i was thinking of the following scenario:
1. The user is in the popup modal context, performs save, some error happens and now closes the popup.
2. Upon closing the popup, the user is out of the edit modal context, and (s)he sees the error `Validation Failed! Please fix the below errors before submitting on the page and after a couple of seconds, it goes away. 
Which i thought would be confusing as without an interaction on the top level page, an error was shown, and went away.

> Your fix will not work as there are various cases (e.g. adding a fetch artifact task) in which the error are not fully mapped to the fields. In such cases, only the global errors represent the errors from the API call.

IIRC, we have a top level flashMessage, and then we've a flashMessage for each modal. Currently, all the global errors are mapped to the page level flashMessage. The logic of unbounded error being mapped to global flashMessage still remains the same. The only difference there would be is, instead of the global flashMessage being displayed outside, we would show that global flashMessage inside the modal.
Do you think that would work for server side errors too? or is there anything that I am missing?

kritika-singh3:
The confusing scenario you mentioned, I also thought of the same. But my views were,
 - if the user remains on the modal popup - the field validation will be able to convey the info we want to convey
 - if the user closes the modal popup, we anyways discard the changes made. So showcasing the error on the discarded data is of no use anymore.

There is a difference between the page level global error and modal level global error - which is the entity on which it is mapped.
The 'consumeError' method works on an entity. The global level entity is the entire pipeline config, hence, it will catch all the errors. The modal level entity is a subset (say stage, job, task) and it will only use the entity specific part of the API response. Hence, there is a chance that edit in one part raises an error in another part which might get missed (I don't remember an exact example but one such e.g. is dependency material and fetch artifact).

PR #8257, opened 2020-06-23
More fixes for pipeline config spa

kritika-singh3:
Issue: #8141 

Description:
 - do not send elastic profile id in post if empty
 - should not validate name or path in custom tab if both of them are empty. otherwise, do validate the presense
 - added validation for name and stages for template config


GaneshSPatil:

PR #8259, opened 2020-06-23
Removed rails template auth API #000

maheshp:
* There is replacement spark API for this and has been in use for few
  releases.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11907-9e5297fcda5504baae150d0696d797e759fba53b).`

PR #8260, opened 2020-06-23
Remove SSL support #000

maheshp:
* GoCD 20.1.0 removed support for SSL by default. In-order for users to
  migrate, as a temporary measure SSL could be enabled using a system
  property. This PR removes complete support for SSL.

Scenarios verified:
1. The GoCD server does not open a SSL port on 8154 even when the System Property `-Dgo.server.enable.tls=true` is specified in wrapper-properties.conf.
2. Upgrade Scenarios 
  - **GoCD Server from 19.6.0 => 20.6.0**
    - GoCD agent goes missing.
    - To get the agent connect to the server, the server url needs to be updated in wrapper-properties.conf to
    ```
      wrapper.app.parameter.100=-serverUrl
      wrapper.app.parameter.101=http://localhost:8153/go
    ``` 
  - **GoCD Server from 20.5.0 => 20.6.0** (20.5.0 tls is enabled)
    - GoCD agent goes missing.
    - To get the agent connect to the server, the server url needs to be updated in wrapper-properties.conf to
    ```
      wrapper.app.parameter.100=-serverUrl
      wrapper.app.parameter.101=http://localhost:8153/go
    ``` 
  - **GoCD Server from 20.5.0 => 20.6.0** (20.5.0 tls is disabled)
    - GoCD agent continues to connect to server.

3. New installations which works without any changes required.
    - GoCD Server 20.6.0 and GoCD Agent 20.6.0
    - GoCD Server 20.6.0 and GoCD Agent (20.2.0 .. 20.5.0)


GaneshSPatil:
Verified on `GoCD Version: 20.6.0 (11995-0500dc2e9647d2ccffb976c3d9228bad7426fdbd).`

PR #8285, opened 2020-06-30
Pipeline config spa fixes (#8141)

GaneshSPatil:
Issue: #8141

Description:
* On Job Settings Tab, allow selecting resources from autocomplete list.
* On Stages Tab, reset the page state to the original (define stages),
  when user toggles to use template and clicks reset.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11907-9e5297fcda5504baae150d0696d797e759fba53b)`

PR #8287, opened 2020-07-01
Fix bug with material password field

kritika-singh3:
Issue: #8261

Description: 
When a password field is set as show value, the style is applied which increased its size.
Added a selector which will negate this


GaneshSPatil:

PR #8298, opened 2020-07-01
Fix duplicate error message from the Stage Trigger API

GaneshSPatil:
#### Description:

* Do not add description from the server health state when the
  server health state description is same as the result message.

#### Before

```json
{
  "message" : "User does not have operate permissions for stage [up42_stage2] of pipeline [up42] { User does not have operate permissions for stage [up42_stage2] of pipeline [up42] }"
}
```

Screen Shot 2020-07-01 at 2 24 48 PM

#### After

```json
{
  "message" : "User does not have operate permissions for stage [up42_stage2] of pipeline [up42]"
}
```

Screen Shot 2020-07-01 at 3 06 07 PM


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11907-9e5297fcda5504baae150d0696d797e759fba53b)` with user `operate`
 - granted `operate` permissions on pipeline group G1
 - but not stage trigger permissions for stage S in pipeline P.

PR #8302, opened 2020-07-02
Add user can_operate field on individual stages of the dashboard pipeline

GaneshSPatil:
Description:

* Add an ability to compute individual stage operate permissions as part of dashboard pipeline permissions
* Add can_operate field to the dashboard API.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11907-9e5297fcda5504baae150d0696d797e759fba53b)`

Issue #8308, opened 2020-07-03
Validate empty tasks in a job


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Currently, GoCD allows defining an empty job which contains no tasks. GoCD should not allow defining an empty job or removing all the tasks from a job.

Related Issues: #8286 #8177 #8194

kritika-singh3:
Completed and Verified as per [comment](https://github.com/gocd/gocd/pull/8309#issuecomment-659316492)

PR #8309, opened 2020-07-03
Validate job contains at least one task

GaneshSPatil:
Issue: #8308

Description:

- [x] Define XML schema migration to add an echo task for the jobs with no tasks.
- [x] Validate job contains at least one task.
- [x] Fix create template API call from templates SPA to include an echo task.


GaneshSPatil:
https://build.gocd.org/go/pipelines/value_stream_map/trigger/5772

Unit and Integration tests are green. Functional specs will not pass due to schema migration.

kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11971-2fe29190f64d8fb2f6bc1089ab0e743b943ecf30)`.
 - An echo task was added during XML migration from 135 to 137
 - The Pipeline Config Create/Update API threw `422` if no tasks were provided
 - the Template Create API created a template with default echo task

Issue #6722, opened 2019-08-06
Help component for concepts


bdpiparva:
##### Issue Type
- Feature proposal

##### Summary

Create a component to show help on the page about the concepts or any other contextual help. 

- [ ] Create a prototype of the component
- [ ] Force all pages to have a dependency on the component
- [ ] The should be a way to open the component when there is text on the SPA

PR #8310, opened 2020-07-03
Showcase a help icon on all SPA

kritika-singh3:
Issue: #6722

Description:
All SPAs should have the ability to showcase some help text.
This PR introduces a info icon on each which on click will open up a small modal containing help w.r.t. to the current page

Preview:

Screen Shot 2020-07-03 at 3 48 24 PM

Screen Shot 2020-07-03 at 3 48 31 PM


GaneshSPatil:
@kritika-singh3 - I am not sure about the UX of the help text. The current position of the help text icon looked more like a help text for the `Add` button.

Lets see if others have any other ideas about this. \cc: @marques-work @arvindsv @maheshp 

I find AWS help text observable, as they've a one liner about the topic and a link to go to the documentation.
Example:
Screen Shot 2020-07-06 at 8 58 49 AM

maheshp:
We already show the help text on most of the pages when a configuration is not yet added. Do we still need this PR?

If the help text is missing on any page we should add with links to the docs page.

arvindsv:
I think it would be nice to show the help text even if there is a configuration. It might be new people or someone who just needs a bit of help. I like the option of showing the text all the time personally.

maheshp:
> I think it would be nice to show the help text even if there is a configuration. It might be new people or someone who just needs a bit of help. I like the option of showing the text all the time personally.

Agree, this would be useful. Showing the text always should be fine since our current help text is just a one liner with a link to the docs. But if we have show something like what we have for elastic agent configurations(with images) then having the help text always might not be feasible.

maheshp:
> We already show the help text on most of the pages when a configuration is not yet added. Do we still need this PR?

@kritika-singh3 I remember we had this conversation and then agreed to have the help icon, so my question is irrelevant. 
It was more of question for myself.

kritika-singh3:
>But if we have show something like what we have for elastic agent configurations(with images) then having the help text always might not be feasible.

Yes you are right @maheshp . This help icon is not present at all SPAs rather the one that needs it. E.g. elastic agent config, agent job run history - these SPAs don't need the help icon. Whereas auth configs, role configs or config repo - a help icon is beneficial there.

In my implementation, a dependency is there from `Page` - but if send `undefined` the icon is not shown

kritika-singh3:
Updates:
 - made the messages consistent
 - moved the help icon to near the header title.

Preview:

Screen Shot 2020-07-08 at 2 18 58 PM
Screen Shot 2020-07-08 at 2 19 05 PM

maheshp:
@kritika-singh3 I am ok with the position of the help icon. I have couple of observations,

1. The size of the help icon seems large, probably we should reduce it.
2. Clicking the help icon opens up a modal. What do you think about using a callout over a modal?

kritika-singh3:
@maheshp Reduced the help icon size to 15px. By callout do you mean, the help text will show up as an on-hover text?

maheshp:
> By callout do you mean, the help text will show up as an on-hover text?

Yes. e.g the callout for help icon on the roles page next to disabled plugin role radio button.

kritika-singh3:
Updated to show callout:

Screen Shot 2020-07-13 at 9 59 24 AM

GaneshSPatil:

GaneshSPatil:
@kritika-singh3 - 

#### 1. the following list of page does not have help text implemented:

- Cruise Config XML (can be omitted? due to rails page?)
- Server Configuration
- Server Maintenance Mode
- Backup
- Analytics
- Users
- User Preferences


####  2. Also, for the following pages the placement of the help icon was not immediately next to the page title:

- Pipeline History
Screen Shot 2020-07-20 at 2 52 56 PM

- Compare Pipeline
Screen Shot 2020-07-20 at 2 52 47 PM


Are we planning to fix this for the current release?

kritika-singh3:
@GaneshSPatil  point 1: The pages you mentioned either have the help text as a part of body itself (Backup, server maintenance mode) or doesn't require one (Cruise Config XML, Users). So, the omission was intentional.

For point 2: No, the header is specifically customized for these pages. The fix is not straightforward for them. 

I have no such plan. However, if needed I can spend some time on them.

GaneshSPatil:
> @GaneshSPatil point 1: The pages you mentioned either have the help text as a part of body itself (Backup, server maintenance mode).

Hmm I see your point. The only worry is, it might cause inconsistencies in showing the help text on different pages. 
> or doesn't require one (Cruise Config XML, Users).

May I know the reasoning why help text is not required for the above-mentioned pages?

> For point 2: No, the header is specifically customized for these pages. The fix is not straightforward for them.

It looked a little odd (only to me??) to not have help text next to the header, hence asked.

@marques-work @maheshp - Thoughts??

kritika-singh3:
>May I know the reasoning why help text is not required for the above-mentioned pages?

My opinion was that such pages are self-explanatory. E.g. Users SPA. It is clear that the SPA deals with users who have access to GoCD.

maheshp:
I think we should be consistent with the position of the help icon, the icons on the history and compare page looks off.

@kritika-singh3 sorry I should have spoken to earlier when I looked at the PR. I don't think it's a blocker for the release but we should fix it.

maheshp:
@kritika-singh3 I think all links in the help text should open in a new tab. Also we need to finalize on the discussion brought up by Ganesh - https://github.com/gocd/gocd/pull/8310#issuecomment-660912605
I will move it QA done as it's been verified.

Issue #2555, opened 2016-08-12
How to delete artifacts based on days


Sanket0510:
##### Issue Type

- Feature proposal
- Feature enhancement

Currently for GO CD we have feature of deleting artifacts once certain amount of memory is exceeded in server.
Consider a scenario in which we have more than 1 services pipelines. Now if we are going to use this feature how can we make sure it is going to delete old artifacts only. Do we have any surety for this??

Is it possible that we have functionality which will take number of days in places of memory so it is going to confirm delete older artifacts only??
##### Summary

Currently for GO CD we have feature of deleting artifacts once certain amount of memory is exceeded in server.
Consider a scenario in which we have more than 1 services pipelines. Now if we are going to use this feature how can we make sure it is going to delete old artifacts only. Do we have any surety for this??

Is it possible that we have functionality which will take number of days in places of memory so it is going to confirm delete older artifacts only??
##### Environment

production
###### Basic environment details

- Go Version: 16.1.0
- JAVA Version: 1.8
- OS: UNIX
- Browser vendor and version (if relevant): Google Chrome

arvindsv:
Closing as stale.

PR #8311, opened 2020-07-06
Artifact purge on demand API

varshavaradarajan:
Issue: #000 (this means @arvindsv knows what the issue # is) 

Description:

Provide an API to purge old artifacts on demand. This introduces a config option in `ArtifactConfig` called `onDemandPurgeThreshold`, which allows for setting the threshold (in hours) beyond which artifacts are considered old and deleted. This config is only for the on demand run and does not interfere with the ArtifactDiskCleaner which deletes artifacts if the specified amount of space is not available.


maheshp:
Is this PR WIP, I don't see the purge endpoint.

Also, I am sorry but I want to understand the use case for this PR.  Today, there is an option to purge artifacts based on the available disk space and GoCD ensures artifacts are cleaned up in background. So, in what cases would an admin want to do a onDemand purge? 

What would be the behavior if the `onDemandPurgeThreshold` is not set and purge artifacts request is made?

arvindsv:
@maheshp I saw it as an optional way to force an artifact purge, an alternate strategy for it. For situations such as https://github.com/gocd/gocd/issues/2555.

Other approaches _could have been_ (or could be), defining at a pipeline-level the number of instances of artifacts to keep. Like: "Preserve artifacts of the last 5 runs".

maheshp:
@arvindsv one thing I found a bit odd with approach was, a GoCD system admin can purge artifacts without even knowing the `onDemandPurgeThreshold` value. This can happen when you have multiple system admins.

arvindsv:
@maheshp Isn't that true of almost anything, where if someone doesn't read the API documentation and makes a call to the API that it can have unintended consequences?

In this case, we would document what happens and say that it depends on the value of the config property, wouldn't we?

maheshp:
> @maheshp Isn't that true of almost anything, where if someone doesn't read the API documentation and makes a call to the API that it can have unintended consequences?

I agree to an extent what you are saying.

@varshavaradarajan @arvindsv instead of setting `onDemandPurgeThreshold` in config, does it makes sense pass that value when making an API call?

arvindsv:
In this case, since the value is not used by anything other than the API call, I can see that passing it in makes sense. If there is a chance of us introducing support for auto-purge from within GoCD, then the config value makes sense.

varshavaradarajan:
@maheshp - I thought that we will use this in the server configuration page. So, asking users to repeatedly set this (or `onDemandPurgeRevisions` , which I plan to introduce next) might be annoying like Duck mentioned. So, okay to leave this as a config option or should I change the API to parse the request body?

@arvindsv - I didn't try any other FS. Will do it and let you know. Thanks

maheshp:
@varshavaradarajan my preference would be to have `onDemandPurgeRevisions` or `onDemandPurgeThreshold` sent as part of the API request over configuring it in the xml, unless we plan to use these values in a background task. Would the configuration`onDemandPurgeRevisions` set per pipeline?

If we plan to pass these values as part of the API request, I think the purge API can do the following,
- Purge artifacts which are older than x minutes.
- Purge artifacts for pipeline runs between x..y

Also, the time for purging the artifacts might vary based on the size of the artifacts. So we might have to think of the behavior of the API, should it be a blocking call or should it trigger an event in the background.

varshavaradarajan:
@maheshp - I think we should leave `onDemandPurgeThreshold` as a config option even if it is used only on demand and not in the background because I don't think it likely that the threshold will change for every API request. @arvindsv - thoughts?

arvindsv:
I don't have much of an opinion, but whether it changes or not shouldn't really decide that. We need to think about where it makes sense. I know we have had similar values in the config before, but when I think about having that value there, I think it will be hard for someone to make sense of why / how it is used.

arvindsv:
There's no reason we cannot find a way forward with it. Reopening it.

varshavaradarajan:
Sorry, @arvindsv , @maheshp - had been a little busy. I'll address this this weekend.

varshavaradarajan:
@maheshp - closing this. I don't think it's a big feature request anyway.

maheshp:
@varshavaradarajan sorry for missing this, I was of the assumption that you are still working on it.

If it's done, can you please rebase and reopen the PR, I will take a look at it.

varshavaradarajan:
@maheshp - thanks! :) But, sorry, no longer have the time to pick this up. :( If needed, I will add you as a collaborator on my fork and you or someone else on the team can finish this. I don't think it's worth the effort though. It's not a popular feature request anyway.

PR #8314, opened 2020-07-06
Return appropriate return code for update current user

kritika-singh3:
Description:
 - current implementation always returns 200 even if the user wasn't updated
 - this PR corrects this issue and returns the code and message


maheshp:

PR #8315, opened 2020-07-06
Added title to disabled plugin role selector

kritika-singh3:
There were a few concerns raised around disabled plugin role selector on add new role modal
 - to make sure that the users understand why it is disabled  - added a title

Screen Shot 2020-07-06 at 12 38 29 PM


arvindsv:
Could consider a "?" help text icon too, right? Using the tool-tip for this might be less discoverable.

kritika-singh3:
After adding a info icon:

Screen Shot 2020-07-06 at 7 17 57 PM

arvindsv:
My preference is this. I think it's ok to keep the old tooltip too, if needed.

maheshp:
@kritika-singh3 I added a commit to tweak the message a bit - https://github.com/gocd/gocd/pull/8315/commits/3991e4f3a9e69421d0f1f7e5cc0144ddde13f2f8

maheshp:

PR #8326, opened 2020-07-08
Removed materials version v1

kritika-singh3:
Description:
A materials API v2 was introduced in [GoCD version 19.10.0](https://api.gocd.org/current/#changes-in-19-10-0).
But v1 was never removed. Hence, doing so for this release.


maheshp:

PR #8334, opened 2020-07-10
Migrate version infos api

kritika-singh3:
Issue: #7482

Description: Migrating version_infos API from rails to spark keeping the version as v1

 - one difference - earlier if the user was not valid it use to return 404. Now it returns 403.

PR #8336, opened 2020-07-10
Fix Version Infos Updater

GaneshSPatil:
PROBLEM:

* Version infos update check has been failing from GoCD v19.10.0
  after changes done as part of PR #6932

* Changing the rails root from 'server' to 'server/src/main/WEB-INF/rails'
  causes the rails version infos controller to fail reading the resource
  from path 'config/go_update_server.pub'

FIX:

* Read the 'config/go_update_server.pub' certificate at java side.


GaneshSPatil:
Due to differences in the root (pwd) for rails vs spring, reading the resource would fail for rails but would work for spring.

Closing this issue, as the rails based version infos API has been migrated to spark https://github.com/gocd/gocd/pull/8334

Issue #8335, opened 2020-07-10
Using access token causes java.lang.NullPointerException


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

In GoCD, personal access tokens belongs to a user. And a user is a part of an authorization configuration, which in turn causes personal access token to belong to an auth config.

Every personal access token has a reference to the belonging auth config using the auth config id. 
Every time a personal access token is used, the validity and the authenticity of the personal access token is verified against the belonging auth config id.

When the auth config id(s) are changed, all the belonging access token has a reference to the old auth config ids, which in turn causes a causes java.lang.NullPointerException when such access tokens are used.

###### Basic environment details

* Go Version: `20.5.0`

##### Steps to Reproduce

The error scenario is explained as follows:

1. Let's say a user `Bob` belongs to an Authorization Configuration `Admins`.
2. User `Bob` creates a personal access token and uses the newly created access token to access GoCD APIs (and everything works).
3. Someone modifies the GoCD configurations by editing the `cruise-config.xml` (_either on disk or from UI_) and renames the authorization configuration from `Admins` to `SuperAdmins`.
4. Now, user `Bob` again tries to access the GoCD APIs and here, the request fails with 500 error.

**Note:** This scenario is also applicable for deleted auth configs.

##### Log snippets

```
2020-07-10 12:03:46,587 DEBUG [qtp1894369629-25] AccessToken:113 - Generating secret using algorithm: PBKDF2WithHmacSHA256 with spec: DEFAULT_ITERATIONS: 4096, DESIRED_KEY_LENGTH: 256
2020-07-10 12:03:46,593 DEBUG [qtp1894369629-25] AccessToken:131 - [Bearer Token Authentication] Authenticating bearer token for: GoCD User: 'admin'. GoCD API endpoint: '/go/api/current_user', API Client: 'curl/7.64.1', Is Admin Scoped Token: 'true', Current Time: '2020-07-10 12:03:46.593'.
2020-07-10 12:03:46,596 WARN  [qtp1894369629-25] HttpChannel:591 - /go/api/current_user
java.lang.NullPointerException: null
	at com.thoughtworks.go.server.newsecurity.providers.AbstractPluginAuthenticationProvider.authenticateUser(AbstractPluginAuthenticationProvider.java:109)
	at com.thoughtworks.go.server.newsecurity.filters.AccessTokenAuthenticationFilter.filterWhenSecurityEnabled(AccessTokenAuthenticationFilter.java:145)
[...]

PR #8339, opened 2020-07-13
Access token fix npe

GaneshSPatil:
Issue: #8335

Description:

#### 1. Revoke access token when belonging auth config id is removed (#8335)

* Revoke personal access token when the auth config id changed by modifying
  the cruise-config.xml from UI or directly editing the config on disk.

* Revoke the personal access tokens when the auth config is deleted from the
  UI or using API.

* Revoke the personal access tokens which belongs to the auth config which
  is not anymore in the config after a restart.
  Steps:
        - Create an Auth Config.
        - Create a personal access token.
        - Stop the server.
        - Rename the auth config id to a new value by directly modifying the
          cruise-config.xml
        - Start the server.
  On server startup, the created personal access token must be revoked.

#### 2. Do not fail with NPE when auth config is null

* Render 401 error along with appropriate error when the auth config
      belonging to the access token is null.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11971-2fe29190f64d8fb2f6bc1089ab0e743b943ecf30)` all the scenarios listed in Point 1.

Issue #8331, opened 2020-07-09
Add page to display all materials used in GoCD


kritika-singh3:
#### Issue Type

- Feature proposal

Parent Issue: #1105, #1145, #1545

### Summary

Propose a materials SPA accessible to all users containing a brief about the materials used in GoCD

### Details

The SPA will be visible to all users. The SPA will

 - display the list of unique materials visible to the current user
 - show the latest commit for each material as available in GoCD
 - has an option to list the modifications i.e. commits available for each material
 - for each of the modifications - display a link to its VSM.

Good to have feature for this SPA:

 - if `auto_update` is set to `false` for non-dependency materials - show an option to trigger an MDU
 - also display the current material's update - whether is in progress or not (similar to config-repo)
 - has an option to list all the pipelines which uses this material

### Challenges or TODO

 - No MDU trigger API support for TFS, Perforce and Package materials
 - Add an API which reports whether an MDU is in progress for a given fingerprint

### Questions

 - Should the materials be grouped by their type, i.e., git, mercurial, etc?
 - Material fingerprint - in the header or body?
 - There will no delete icon but should we provide a link to materials tab instead?
 - While displaying modifications - should we also display modified files?

Thoughts and suggestions are welcomed!

#### Suggested UI

![IMG_20200709_133003](https://user-images.githubusercontent.com/41165891/87013930-07c34580-c1e9-11ea-8b72-34118991707d.jpg)

![IMG_20200709_133028](https://user-images.githubusercontent.com/41165891/87014115-448f3c80-c1e9-11ea-9459-aba3bcf6d0d1.jpg)

maheshp:
@kritika-singh3 listing out the things we discussed,

Materials Page

1. List all unique materials(This list view would be very similar to the Config Repo SPA) 
   What should the list show?
     - Material name
     - Material attributes e.g url/branch
     - latest commit revision
     - latest commit message
     - latest commit author
     - Indicate if auto_update is enabled or not
     - Icon for material type

2. Search for a material by name/url within the page
   - Serach materials by type/branch?

3. Actions(Buttons) on the list
   - Trigger MDU
   - View VSM for the latest commit
   - A edit button which redirects to materials page on pipelines, scm or package material (What happens when a material is associated with multiple pipelines)

4. Details
   - List pipelines which are dependent on the material
   - List last 5 modifications
   - Option to saerch for a modification based on revision

5. Permissions
   - Should not show all materials, show materials with view permission

6. Errors & Warnings
[...]

PR #8341, opened 2020-07-14
Add Materials SPA

kritika-singh3:
Issue: #8331

Description:
Added the initial bit of Materials SPA:
 - showcase all the unique materials viewable to the current user
 - sorted by type
 - showcase basic attributes
 - show icon
 - search among materials based on type, name and display url

Edit: Added a toggle `show_materials_spa` - when enabled the menu would contain a link to the page. (Default is disabled).

materials_spa


maheshp:
@kritika-singh3 are you planning to list even dependency materials? Probably we should not.

GaneshSPatil:
@kritika-singh3 - Moving this to project `20.7.0`. Will be verified for the next release.

PR #8342, opened 2020-07-14
Add default echo task for config repo pipelines (#8308)

GaneshSPatil:
Issue: #8308

Description:

* Add a config repo migrations to add a default echo tasks for the
  jobs which does not have any tasks defined.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11971-2fe29190f64d8fb2f6bc1089ab0e743b943ecf30)`. 
When a config repo contains a pipeline with no tasks - a default echo task was added.

PR #8343, opened 2020-07-14
Adding business-continuity as part of core #000

maheshp:
- Moved business continuity code from enterprise to core as is.
- Support for virtual IP assignment is not added as this requires a CLI, will try to cover this in documentation.
- GoCD does not load/include any addons going forward.

- [x] Verified BC on 20.6.0
- [x] Verified GoCD 20.5.0 upgrade to 20.6.0 with BC addon
- [ ] Document BC in docs.gocd.org https://github.com/gocd/docs.go.cd/pull/475


GaneshSPatil:
Verified on GoCD `20.6.0-11980 (87819ade10267f6dc055d77848819f640a4c9511)`

Scenarios covered:
1. Verified upgrade scenario from 20.5.0 to 20.6.0
2. Fresh Installation on 20.6.0

Could not verify the agent connectivity through Virtual IP (verified through a record entry DNS)

Issue #8345, opened 2020-07-16
Stale response while using Stage History API


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

The stage history api does not return the newly triggered stage instance information as part of the stage history.

###### Basic environment details

* Go Version: `20.5.0`

##### Steps to Reproduce

1. Create a pipeline with one stage.
2. Trigger the pipeline and wait for it to pass.
3. Use [Get Stage History API](https://api.gocd.org/current/#get-stage-history) to fetch the stage history. (It should return 1 instance)
4. Trigger pipeline again (or re-ran the stage).
5. Use [Get Stage History API](https://api.gocd.org/current/#get-stage-history) to fetch the stage history. It should return 2 instances, but it returns only first.

##### Expected Results
The Get Stage History API should return all the instances.

##### Actual Results
The Get Stage History API returns only the firstly cached instance information.

##### Possible Fix

[findDetailedStageHistoryViaCursor](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L455-L478) method is called when the stage history of a stage is requested.

To avoid fetching the history information again from the database, the fetched stage instances are stored under GoCache.
The key used to store the fetched instances is generated using cursor, pageSize and feedModifier, see [this](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L460).
This key would be duplicate for all the stage instances until count reaches more than 10.

The fix would be to flush the cached stage histories on a stage update. Similar to [removal of other stage histories](https://github.com/gocd/gocd/blob/master/server/src/main/java/com/thoughtworks/go/server/dao/StageSqlMapDao.java#L126-L127)

GaneshSPatil:
Closed by automation. Reopening for testing...

kritika-singh3:
Fixed and verified in #8346

PR #8346, opened 2020-07-16
Fix Stage History API stale response (#8345)

GaneshSPatil:
Issue: #8345

Description:

* Clear cached stage history instances when stage status is updated.


kritika-singh3:
Verified on `GoCD Version: 20.6.0 (11974-cf63fcc59358bec8dfd4c5b017743853818fcc79)`. The Stage History API gave expected results.

PR #8350, opened 2020-07-17
Resolve secret params during check connection

kritika-singh3:
Issue: The bug was identified by a user in [go-cd group](https://groups.google.com/g/go-cd/c/uM1ir8fIQqo)

Description:
 - resolve secret params before sending the request for check connection
 - send pipeline group name in payload for check connection API


kritika-singh3:
@GaneshSPatil I have added this to 20.6.0 release. If you think otherwise, please update the same.

GaneshSPatil:
@kritika-singh3 - the changes LGTM.

@kritika-singh3 @maheshp  - Should we've a functional spec for this behavior?

maheshp:
This endpoint was moved from rails to spark. I checked the rails controller, the secret params were resolved before making a check connection call. Guess it was missed out when ported to spark.
I am ok with a unit test to ensure the params are resolved, looks like the PR is missing tests for these changes. @kritika-singh3 can be please add the relevant tests.

GaneshSPatil:
Verified on GoCD Version: 20.6.0 (11973-db44bc3ddbb4418fde09114086603d72dd3d5823).

Issue #8352, opened 2020-07-17
Config repo with no pipelines fails to parse


kritika-singh3:
##### Issue Type

- Bug Report

##### Summary

A config repo with no pipelines fails to parse with NPE

###### Basic environment details

* Go Version: `GoCD Version: 20.6.0 (11974-cf63fcc59358bec8dfd4c5b017743853818fcc79)`
* OS: `Mac OS X 10.13.6`

##### Steps to Reproduce

1. Add a config repo pointing to a repo with no pipelines configured with a single environment.
2. Wait for changes to be synced.

##### Expected Results

Repo should be parsed and show the environment

##### Actual Results

<img width="970" alt="Screen Shot 2020-07-17 at 2 40 06 PM" src="https://user-images.githubusercontent.com/41165891/87770071-230f0000-c83c-11ea-9ec9-fbc488968644.png">

##### Possible Fix

This was introduced due to changes in #8342 where in we try to migrate the pipeline to add a default task if no tasks are provided.

##### Log snippets

```
2020-07-17 14:45:00,689 ERROR [115@MessageListener for ConfigMaterialUpdateListener] JsonMessageHandler3_0:146 - Unexpected error when handling plugin response
java.lang.NullPointerException
java.lang.NullPointerException: null
	at com.thoughtworks.go.plugin.access.configrepo.AddEmptyTaskJoltTransformer.transform(AddEmptyTaskJoltTransformer.java:31)
	at com.bazaarvoice.jolt.Chainr.doTransform(Chainr.java:240)
	at com.bazaarvoice.jolt.Chainr.transform(Chainr.java:175)
	at com.thoughtworks.go.plugin.access.configrepo.ConfigRepoMigrator.migrate(ConfigRepoMigrator.java:33)
	at com.thoughtworks.go.plugin.access.configrepo.v3.JsonMessageHandler3_0.migrate(JsonMessageHandler3_0.java:160)
	at com.thoughtworks.go.plugin.access.configrepo.v3.JsonMessageHandler3_0.responseMessageForParseDirectory(JsonMessageHandler3_0.java:130)
	at com.thoughtworks.go.plugin.access.configrepo.ConfigRepoExtension$4.onSuccess(ConfigRepoExtension.java:145)
	at com.thoughtworks.go.plugin.access.configrepo.ConfigRepoExtension$4.onSuccess(ConfigRepoExtension.java:132)
[...]

PR #8356, opened 2020-07-20
Fix NPE during config repo migration 11 (#8352)

GaneshSPatil:
Issue: #8352

Description:

* Do not fail to parse when config repo pipelines are defined using templates


kritika-singh3:
Verified on `20.6.0 (11982-17fa4dff6338a9da1230258e84ae089916a207ec)`.

PR #8360, opened 2020-07-20
Fix hash generation for pipeline config export #000

maheshp:
* ConfigRepoPlugin#etagForExport relied on Objects.hash to generate
  hash for a pipelineconfig and plugin. It was observered, a change
  in material for a  pipeline did not generate a new hash and hence
  PipelineExport and PipelineAsCode preview endpoints relied on a invalid etag 
  returned stale responses.
* EntityHashingService generates hash for a PipelineConfig for a given
  plugin.


maheshp:
Verified On GoCD Version: 20.6.0 (11995-0500dc2e9647d2ccffb976c3d9228bad7426fdbd). using GoCD YAML plugin export.

PR #8376, opened 2020-07-24
Fixing build - functional spec fails after 8372

kritika-singh3:
Description:
The stage and job were being validated before adding the job and tasks respectively.
This caused the validation to occur but without the error getting mapped to the UI

Fixed this by removing the entity validation at modal level and clearing the global errors irrespective of presence of error response.


maheshp:
Tests are passing - https://build.gocd.org/go/pipelines/regression-SPAs/2101/Firefox/2

PR #8411, opened 2020-08-03
Implement Permissions for Config Repository Material Test Connection

GaneshSPatil:
Implement Permissions for Config Repository Material Test Connection

* Introduce config repo material test connection API
    * Introduce AbstractMaterialTestController which contains a controller method to verify check connection.
    * Use the AbstractMaterialTestController in InternalMaterialTestControllerV1 and ConfigReposControllerV4.

* Allow only admins and pipeline group admins to perform check connection on pipeline materials.
    * This commit brings back the authorization check to allow only super admin and group admin users to perform material check connection.

* Allow users to perform config repo check connection when user can administer config repo.

* Context:
    * Prior to GoCD release v20.6.0, GoCD had a common material check connection API to perform material test for pipeline as well as config repo materials.
  
    * As part of GoCD release v20.2.0, GoCD added support for granular auth on config repositories. Which in turn required users with granular auth on config repositories to perform material test connection.
  
    * Due to implementation complexities to allow users with access to config repos to perform material test on such config repo materials, was not possible, hence, the material test connection API was made accessible for all authenticated GoCD users (in GoCD release v20.2.0).
  
    * This PR allows pipeline material check connection to be performed by super admin and group admin users; And config repo material check connection to be performed by users who can administer config repo.


api/api-config-repos-v4/src/main/java/com/thoughtworks/go/apiv4/configrepos/ConfigReposControllerV4.java
```diff
@@ -57,16 +57,18 @@
 
 @ToggleRegisterLatest(controllerPath = Routes.ConfigRepos.BASE, apiVersion = ApiVersion.v4, as = "branch_support")
 @Component
-public class ConfigReposControllerV4 extends ApiController implements SparkSpringController, CrudController<ConfigRepoConfig> {
+public class ConfigReposControllerV4 extends AbstractMaterialTestController implements SparkSpringController, CrudController<ConfigRepoConfig> {
```
maheshp, line 60: Are you planning to document the test connection endpoint? I remember there was a conversation earlier if we should document the test connection API or keep it internal. I think it was decided to keep it internal as the assumption was it will be used only on the SPA. If we are planning to keep it internal we should not add the endpoint in this controller.

GaneshSPatil: @maheshp - Yes, this needs to be an internal API. I will move it to internal config repos api

maheshp:
Verified on GoCD Version: 20.7.0 (12066-b7bedcef7f27f0fd956dc0dbfd428de5a5d785a0)

Issue #8416, opened 2020-08-04
Agent SSLVerification with reverse proxy fails


maheshp:
##### Issue Type

- Bug Report

##### Summary
See - https://gitter.im/gocd/gocd?at=5f22fd25fe6ecd2888843f6e for context.

If a GoCD Server is fronted by a reverse proxy, SSL verification for GoCD Agent does not work if SLLVerificationMode is set to `FULL`. 
This is because, even though the required SSL parameters are provided in the agent's `wrapper-properties.conf` the same are not passed on to the agent process when launched by the Agent-Launcher. The properties which are not passed are, `sslCertificateFile`, `sslPrivateKeyFile` and `sslPrivateKeyPassphraseFile`

maheshp:
Fixed in - https://github.com/gocd/gocd/pull/8417

PR #8417, opened 2020-08-04
Launch agent with all SSL properties if provided #8416

maheshp:
Issue: #8416

* The agent launcher while starting an agent did not pass all the configured
  SSL properties in the wrapper config. The agents were not able to
  connect to a reverse proxy which used to terminate SSL connections.
* This PR ensures all the configured SSL properties are passed to
  the agent.


GaneshSPatil:
Verified on `GoCD Version: 20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed).`

PR #8437, opened 2020-08-07
[Materials SPA] Fix bug with rendering pipeline usage for a searched material.

GaneshSPatil:
#### Issue

The MaterialVMs model passed to the view in a normal render and in the searched render are different.
Every time a search string is specified, a new MaterialVMs is created with the filtered MaterialVMs.
Mithril views are cached and works with references. Even when the individual materialVM are same, the refering
MaterialVMs are different. And the newly fetched usage data was stored in the old referenced MaterialVMs,
causing pipeline usage to be always rendered as loading.

#### Fix

Add Mithril key to the MaterialsWidget which forces mithril to redraw the entire view with the updated MaterialVMs.


kritika-singh3:
Verified on `GoCD Version: 20.7.0 (12069-7ccd8de2ab296ca2c64f33086ae2e8a7ae26b696)`

PR #8443, opened 2020-08-10
Debug agent authentication issue

maheshp:
* This is temporary commit to debug the issue with agent loosing contact
  after a while. Looks like the agent authentication fails after a
  while. Logging the token to verify if a right token is passed.


maheshp:
This was temporary commit and has been reverted as part of https://github.com/gocd/gocd/pull/8473

PR #8460, opened 2020-08-14
Mask user info on materials tab on pipeline config spa

kritika-singh3:
Use url-parse library to mask user-info (if present) on the materials tab

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/types.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/materials/types.ts
@@ -51,6 +51,26 @@ export const mapTypeToDisplayType: { [key: string]: string; } = {
   plugin:     "Plugin"
 };
 
+function urlForDisplay(url?: string) {
+  if (!url) { return undefined;}
+
+  const parsed = urlParse(url, {});
+  // do not mask passwords
+  if (_.isEmpty(parsed.auth) || parsed.protocol.includes('ssh') || parsed.protocol.includes('svn+ssh')) {
+    return parsed.href;
+  }
+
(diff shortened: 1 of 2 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/models/materials/types.ts
```diff
@@ -51,6 +51,24 @@ export const mapTypeToDisplayType: { [key: string]: string; } = {
...
+      return parsedUrl.href;
+    }
+    if (parsedUrl.protocol.includes('ssh') || parsedUrl.protocol.includes('svn+ssh')) {
+      return parsedUrl.href;
+    }
+    parsedUrl.set('username', '******');
+  }
+  return parsedUrl.href;
+}
+
```
GaneshSPatil, line 71: * how about renaming the method name `urlForDisplay`?

* How about simplifying the above code as:
```javascript
function urlForDisplay(url?: string) {
  if (!url) { return undefined;}

  const parsed = urlParse(url, {});
  // do not mask passwords
  if (_.isEmpty(parsed.auth) || parsed.protocol.includes('ssh') || parsed.protocol.includes('svn+ssh')) {
    return parsed.href;
  }

  if (parsed.auth.includes(":")) {
    // mask password when username and password is provided as username:password@url
    parsed.set('password', '******');
  } else {
    // mask token when credentials are provided as token@url
    parsed.set('username', '******');
  }

  return parsed.href;
}
```

maheshp:
Verified on GoCD Version: 20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed).

Issue #8438, opened 2020-08-07
Dashboard Stage Overview Issues


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

- [x] Misaligned stage success / failure / cancelled icon. (aligned more to bottom).
- [x] Increase width of the stage name on the stage overview.
- [x] Job tooltip position when scrolling
- [x] Job tooltip position for the jobs (where index of the job is more than 5)
- [x] Job state indication (building/passed/failed) bar shrinks in width when job name is longer.
- [x] Capitalize job state.
- [x] Rerun tooltip should say "Rerun Stage".
- [x] Use external link icon for stage details page link.
- [x] Show server time on hover for:
  - [x] Stage triggered on.
  - [x] Stage cancelled on.
- [x] [Bug] Rerun failed doesn't seem to work (when the first stage is failed and second stage hasnt yet run).
- [x] Change job table header field from 'job' to 'job name'.
- [x] Sortable table header.
- [x] Increase the height of the scrollable window to denote there are more items.
- [x] Tooltip opacity issue
- [x] Increase polling interval to 10sec
- [x] Always show ellipsis on agent column.
- [x] Position job rerun buttons to the left
- [x] Relative job progress bar

##### Good to have:
- [ ] Should close stage overview when pressed ESC?

kritika-singh3:
<img width="873" alt="Screen Shot 2020-08-10 at 9 54 19 AM" src="https://user-images.githubusercontent.com/41165891/89751870-8add0180-daef-11ea-86e6-a298ce806e25.png">

 - [ ] Inconsistent icon size when one job is completed and rest are in progress (or vice-versa)
 - [ ] `building` and `in progress` should be capitalized as in `Passed`
 - [ ] The agent column is not consistent - maybe be can set the max chars as 8 and show an ellipse
 - [ ] If the current stage is in `Building` or `Failing` state - the checkboxes and the action buttons should not show up

GaneshSPatil:
@kritika-singh3 -

> * [ ]  Inconsistent icon size when one job is completed and rest are in progress (or vice-versa)
> * [ ]  `building` and `in progress` should be capitalized as in `Passed`
> * [ ]  The agent column is not consistent - maybe we can set the max chars as 8 and show an ellipse

Point 1/2/3 has been fixed and moved to the issue description checklist.

> * [ ]  If the current stage is in `Building` or `Failing` state - the checkboxes and the action buttons should not show up

[...]

PR #8461, opened 2020-08-14
Dashboard stage overview sortable jobs (#8438)

GaneshSPatil:
* Allow jobs to be sorted based on columns:
	- job name
	- job state
	- job duration
	- job assigned agent

* Job state sort priority is building -> passed -> cancelled -> failed.

* Show unassigned agents at the bottom when table is sorted
  in ascending order on agents column.

* Implement sort by duration
	- show in progress jobs at first.
	- sort by the total durtion taken by the job.

* Change sorting order when clicked twice on the column.

* Sort by job state in descending order by default to always
  show failed jobs first.

Issue: #

Description:


maheshp:
Verified on GoCD Version: 20.7.0 (12084-250a04bf83ebd3fc7faf00345518946d7dc78d59)

maheshp:
@GaneshSPatil we might need to re-look at the sort order based on Job state. I was looking at a stage with multiple jobs, couple of jobs in the stage had failed and it rightly showed on the top of the list. When I re-ran those failed jobs, the jobs went to the bottom of the list and was not visible in the viewport of the stage overview. I would have preferred to see the the re-run jobs on the top of the list. WDYT?

GaneshSPatil:
Hello @maheshp,
Yes, the scenario that you mentioned makes sense to me. 

It is due to the priority of the jobs based on job state (in desc order) is as follows:
```javascript
[
    Result.Failed,
    Result.Cancelled,
    Result.Passed,
    Result.Building
]
```

How about changing the priority to:
```javascript
[
    Result.Failed,
    Result.Cancelled,
    Result.Building,
    Result.Passed
]
```

So, when the failed jobs are re-run, the building jobs will be on top.

Issue #8463, opened 2020-08-14
Fail to perform any GoCD updates due to invalid config repository configuration


GaneshSPatil:
[Please refer to the YAML configuration provided in the Steps to Reproduce point 1]

#### 1. Confusion between GoCD Parameters and YAML Comment.
  - A GoCD parameter should be used within quotes in YAML definition. Example:
    ```yaml
        arguments:
            - "#{COMMAND_TO_EXECUTE}"
    ```
  - A Yaml comment should be used only as #my_comment. Example:
    ```yaml
        arguments:
            - "#{COMMAND_TO_EXECUTE}"  #this is an argument specified as a pipeline parameter
    ```
    In the given configuration, the argument is specified as `- #{COMMAND_TO_EXECUTE}` instead of `- "#{COMMAND_TO_EXECUTE}"` making it read as comment and evaluated to `-` (causing null argument).

#### 3. Missing GoCD full config save safety check to allow config updates regardless of validity of config repository.

---

##### Issue Type

- Bug Report

##### Summary

An invalid config repository blocks all the GoCD config saves.

###### Basic environment details

* Go Version: `20.6.0` (This error only occurs on GoCD Release v20.6.0, works fine with other releases)

##### Steps to Reproduce

1. Create a config repository with following Yaml Configuration.
<details>
  <summary>`config-repo-pipeline.gocd.yaml`</summary>
  
```yaml
COMMAND_TO_EXECUTE: ls
format_version: 3
pipelines:
  config-repo-pipeline:
    group: first
    label_template: ${COUNT}
    lock_behavior: none
    display_order: -1
    materials:
      git:
        git: https://github.com/gocd/gocd
        shallow_clone: false
        auto_update: true
        branch: master
        encrypted_password: AES:OF+lIfTm57i3ZJm8uALG1g==:2VDyBzAfH8NGqoJF2QXXmA==
    stages:
    - first-stage:
        fetch_materials: true
        keep_artifacts: false
        clean_workspace: false
        approval:
          type: success
          allow_only_on_success: false
        jobs:
          first-job:
            timeout: 0
            tasks:
            - exec:
                arguments:
                - #{COMMAND_TO_EXECUTE}
                command: docker
                run_if: passed
    parameters:
      COMMAND_TO_EXECUTE: ls
```
</details>

2. Upon successful parse, no message is shown on the config repository page.
[...]

PR #8464, opened 2020-08-14
Do not fail while performing config save when last known config repo partial has structural errors (#8463)

GaneshSPatil:
Issue: #8463

Description:

* Add a safety check around config partial hash generation.
  The hash generation function serializes config entity into xml and then
  computes hex. There is a possibility of plugin providing a structurally
  invalid config which can fail during serializing into xml. Hence,
  add a try/catch safety net to avoid failures while serializing.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/Argument.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/Argument.java
@@ -17,9 +17,11 @@ package com.thoughtworks.go.config;
 
 import com.thoughtworks.go.domain.ConfigErrors;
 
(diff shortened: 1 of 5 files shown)


maheshp:
@GaneshSPatil while these changes fix the issue during comparing partials, the partial would be marked as valid and the pipelines would show up on the dashboard. I think we should validate the Arguments in Tasks for null values.

Also, once the pipeline shows up on the dashboard, trying to view the read only config errors out.

maheshp:
Verified on GoCD Version: 20.7.0 (12096-ccdde50c002926ff21e74930e5f4ee7f2d298b1b)

PR #8465, opened 2020-08-15
Stage overview fixes (#8438)

GaneshSPatil:
Issue: #8438

Description:

* Do not show stale stage overview.
  When clicking on a new stage overview, for a short duration,
  the stage overview information from the previous stage was shown.
  This was due to not clearing older stage overview model on close.

* Show stage status bar even when stage overview is loading.


kritika-singh3:

Issue #8467, opened 2020-08-17
[Stage Overview] Invalid job transition progress bar state for job tooltip


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Invalid job state transition progress bar state when the job takes **0 seconds** for preparing.
For a building job, when job takes just 0 seconds for preparing, the scheduled state is shown as in progress.

<img width="901" alt="Screen Shot 2020-08-17 at 10 34 15 AM" src="https://user-images.githubusercontent.com/15275847/90361177-2ab00780-e07b-11ea-8b05-2f79e2d9249a.png">

##### Steps to Reproduce

1.  Create a pipeline with any material. (say `upstream`)
2. Create another pipeline with `upstream` pipeline as (the only) material.
3. Trigger pipeline `upstream` and wait for it to complete.
4. When `downstream` pipeline is in progress, hover over the building job. When in building state, both `waiting` and `building` states are shown in progress.

##### Expected Results
Only the `building` state should be in progress.

##### Actual Results
Both, `scheduled` and `building` states are shown as in progress.

##### Possible Fix
This issue is caused due to bug with unknown state check, which says a state with "00s" is unknown. [[Code Ref](https://github.com/gocd/gocd/commit/1cb478ca5f4bf672e7a269d7a5a08fc843e0b25c#diff-ec609f5c1102a015884233f21819d80eR124)]

Instead of checking if the value is zero, rely upon job state transition response from the server, if a particular state does not exists in the response, consider it as unknown.
Eg: if the job state transition does not contain `completing` state, then `uploading artifacts` is an unknown field.

kritika-singh3:
Fixed in #8468 and verified on `GoCD Version: 20.7.0 (12075-9f82b238e56c8d63fb8f13094f57f64ecb8f042e)`.

PR #8468, opened 2020-08-17
[Stage Overview] Fix job tooltip invalid state transition state (#8467)

GaneshSPatil:
Issue: #8467

Description:

* Remove '00s' seconds check for considering a job state as unknown.
* Check if job state transition json contains a particular state
  to consider the state as unknown.


kritika-singh3:

Issue #8470, opened 2020-08-17
Misaligned stage overview when pipeline has manual stages


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

Stage overview is misaligned when pipeline has manual stages (in particular order).
See Performance pipeline on build.gocd.org (GoCD Version: 20.7.0 (12085-2e1fc89a21a1633aebeed37c6dbd66a23cedca1d). release).

<img width="684" alt="Screen Shot 2020-08-17 at 3 42 18 PM" src="https://user-images.githubusercontent.com/15275847/90385458-d029a200-e0a0-11ea-9d33-4d4212985bc6.png">

##### Steps to Reproduce

1. Create a pipeline with at least 5 stages.
2. Make any one of first 4 pipeline manual (let say 3rd).
3. Make the 5th pipeline manual.
4. Trigger all the stages one by one.
5. When the last stage is building, hover on it.

kritika-singh3:
Fixed and verified on `GoCD Version: 20.7.0 (12089-16e70753c861f6f772eda6dfcac31cd0f1659eea)`.

PR #8471, opened 2020-08-17
Fix stage overview position when fifth stage is manual (#8470)

GaneshSPatil:
Issue: #8470

Description:
* Consider width of the stage including the manual gate when the stage is manual.


kritika-singh3:

PR #8473, opened 2020-08-17
Synchronise token verification

maheshp:
* HMAC generation is not thread safe, if multiple agents try to authenticate at the same time the hmac
   generated using the Agent UUID would not match the actual token.


GaneshSPatil:
The changes LGTM,

Have not seen this issue happening on build.gocd.org on version `GoCD Version: 20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed).`

Marking as QA Done.

PR #8477, opened 2020-08-20
Showcase building job before passed jobs on the stage overview

kritika-singh3:
Issue: https://github.com/gocd/gocd/pull/8461#issuecomment-676978270

Description:
In case of re-run of failed jobs, the current running jobs should be shown upfront before the passed ones.


GaneshSPatil:
Verified on `GoCD Version: 20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed).`

PR #8486, opened 2020-08-26
Update materials spa modifications modal

kritika-singh3:
Issue: #8331 

Description:
Updated modifications modal on materials spa
 - render modifications in a tabular format
 - render comment as is. In case of longer messages - make it scrollable


kritika-singh3:
Opinions needed: Should we always show the scroll bar?

1. Scroll bar rendered always
    comment_scroll_1

2. Scroll bar rendered when in use
    comment_scroll_2

GaneshSPatil:
@kritika-singh3 - the scroll on the expanded body commit message looks good to me.

For the modifications list, instead of scroll, can we have the old expand-collapse style? Meaning, on the modifications modal, each modification should be rendered only in one line with (more/less or ...) and clicking on more it expands with the scrollable height (the styles we've currently will be used when expanded). 

@maheshp - Thoughts??

maheshp:
> Meaning, on the modifications modal, each modification should be rendered only in one line with (more/less or ...) and clicking on more it expands with the scrollable height (the styles we've currently will be used when expanded).

I agree and would prefer the same. I don't think the intention of users would be to look at the entire commit message on the modifications modal. On the modal the user should be able to easily identify a commit to look at the corresponding VSM, which I think should be possible using the revision, commit message header and the committer details.

kritika-singh3:
> (the styles we've currently will be used when expanded)

@GaneshSPatil @maheshp With the scroll bar always rendered (option 1) or the other (option 2)?

server/src/main/webapp/WEB-INF/rails/webpack/views/pages/materials/modal.tsx
```diff
@@ -241,3 +241,54 @@ export class ShowUsagesModal extends Modal {
...
+  expanded: Stream<boolean>;
+  setExpandedTo: (state: boolean, e: MouseEvent) => void;
+}
+
+class EllipseText extends MithrilComponent<EllipseAttrs, EllipseState> {
+  private static MIN_CHAR_COUNT = 80;
+
+  oninit(vnode: m.Vnode<EllipseAttrs, EllipseState>): any {
+    vnode.state.expanded = Stream();
+    vnode.state.expanded(false);
```
GaneshSPatil, line 259: IIRC, this is done because of typescript compilation error.
try this instead - `vnode.state.expanded = Stream<boolean>(false)`?

kritika-singh3: I'll add this change in the next PR change. Thanks!

PR #8489, opened 2020-08-27
Add mdu related actions for materials spa

kritika-singh3:
Issue: #8331 

Description: Show MDU status for the materials.
Also added a poller on the page which will fetch the materials info every 10 secs.


kritika-singh3:
Preview:

Screen Shot 2020-08-27 at 4 41 46 PM

PR #8492, opened 2020-08-28
ConfigRepoPlugin should decrypt all of the configuration properties when including them in the request.

marques-work:
Also, some refactoring to make testing easier when working with encrypted values.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
@@ -55,10 +55,14 @@ public class ConfigurationProperty implements Serializable, Validatable {
     private EncryptedConfigurationValue encryptedValue;
 
     private final GoCipher cipher;
-    private ConfigErrors configErrors = new ConfigErrors();
+    private final ConfigErrors configErrors = new ConfigErrors();
 
     public ConfigurationProperty() {
-        this.cipher = new GoCipher();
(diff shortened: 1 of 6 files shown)


GaneshSPatil (approved):
LGTM.

`ConfigurationProperty.java` class looks like a pojo with some behavior, should we consider using Lombok (with `@Accessors(chain = true)`)?

marques-work:
@GaneshSPatil it's nearly a POJO. I'm not certain `@Accessors(chain = true)` will give you a sizeable benefit here; note that the builder-style setters need to wrap the string values to match the field type. So, you really only save 3 setters, while the wrapping setters will remain.

That said, I'm certainly not opposed to it.

GaneshSPatil:
@marques-work Ah fine. Understood. then I think its okay to keep it as is.

marques-work:
@GaneshSPatil actually, I just tried `@Accessors(chain = true)` with `@Getter @Setter`. It's not as straightforward as I thought. Trying to have overloaded setters that take `String`s and wrapping with the correct field type seems to not work. Lombok might be seeing that it has the same name and chooses not to generate a naive setter, which in turn makes the overloaded setter fail. A bit catch-22; I'm sure I can get it to work, but it'll be more refactoring than it's worth, I think.

maheshp:
Verified in `GoCD Version: 20.8.0 (12204-0e9d2bd06e51e694040a89f9b6b8dcb6978d41c5).` using gocd-groovy-dsl-config-plugin-1.1.0-233

Issue #8234, opened 2020-06-14
Support for secrets management interpolation on Pluggable SCM materials


chadlwilson:
##### Issue Type
- Feature enhancement

##### Summary

It appears that currently GoCD does not interpolate variables defined using the [Secrets Management](https://docs.gocd.org/current/configuration/secrets_management.html) Plugins for use by pluggable SCM materials.

That is, if you define a `password` (or other `secure` property indicated by an SCM plugin) with a value like `{{SECRET:[file-secrets][test-password]}}` it is not interpolated/populated with the real value before sending to the SCM plugin requests.

I note that currently the [`PluggableSCMMaterial`](https://github.com/gocd/gocd/blob/master/domain/src/main/java/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java) is not [`SecretParamAware`](https://github.com/gocd/gocd/blob/master/config/config-api/src/main/java/com/thoughtworks/go/config/SecretParamAware.java) and only the built-in `ScmMaterial`s have this support; with some hard-coded logic for the `password` field. 

My thought is that if a plugin indicates a value as `secure: true` in the `scm-configuration` request for a property, that it may be possible for GoCD Server to use this as an indication that a value should be eligible for `SecretParam` interpolation/retrieval prior to building the requests to the plugin.

##### Environment

GoCD 20.3.0

arvindsv:
@chadlwilson The reason for not supporting secrets in plugin SCMs was that plugin load order is undefined. So, it's possible that the secret plugins have not loaded while other plugins have. So, this can create problems. Right, @maheshp?

In cases such as this, where the material will be checked once more. maybe it is possible? However, looking at the original issue you linked to, it looks like it is more about config repo, than pluggable SCM.

Maybe @marques-work has some thoughts.

maheshp:
The [Secret Management](https://docs.gocd.org/current/configuration/secrets_management.html) feature was introduced with support for secret params only in Environment Variables and Password field for built-in SCM Materials.

Secrets support for all Plugins, PluggableSCMMaterials, etc... was something we considered but it  never got prioritized. With plugins, the idea was that whenever GoCD makes a request to plugin it should pass the configuration with all secrets resolved. 

> @chadlwilson The reason for not supporting secrets in plugin SCMs was that plugin load order is undefined. So, it's possible that the secret plugins have not loaded while other plugins have. So, this can create problems. Right, @maheshp?

[...]

PR #8503, opened 2020-08-31
Adding support for secrets in pluggable scm materials

kritika-singh3:
Issue: #8234 

Description:
 - added `pluggable_scm` as an entity for secrets rules
 - added support for secrets in plugin configurations
 - resolved values will be send to the plugin while MDU
 - MDU: will pass if the pluggable scm config is allowed access to secret config
 - pipeline will get triggered on successful MDU 

### Plugin UI

Since the view comes from the plugin, the plugin has to be updated in order to allow users to view unmasked values for secure configurations while editing


domain/src/main/java/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -348,4 +350,38 @@ public boolean getInvertFilter() {
...
+
+    @Override
+    public boolean hasSecretParams() {
+        if (secretParams == null) {
+            parseForSecretParams();
+        }
+        return scmConfig != null && (secretParams != null && !secretParams.isEmpty());
+    }
+
+    private void parseForSecretParams() {
```
maheshp, line 362: Rename to `parseSecretParams`?

config/config-api/src/main/java/com/thoughtworks/go/domain/config/ConfigurationProperty.java
```diff
@@ -349,4 +348,26 @@ public ConfigurationProperty deserialize(String name, String value, String encry
...
+        return this.secretParamsForValue;
+    }
+
+    public String getResolvedValue() {
+        if (hasSecretParams()) {
+            return getSecretParams().substitute(getValue());
+        }
+
+        return getValue();
+    }
```
kritika-singh3, line 372: Reason behind introducing a new method: parsing of a secret param and substituting the same needs to check the following cases:
 - not secure values: null nd not blank
 - secure values: null nd successful decryption.

These checks are already performed in `getValue()` method. Hence, to avoid code duplication, decided to introduce a new method. Let me know if this is not ok.

domain/src/main/java/com/thoughtworks/go/config/materials/PluggableSCMMaterial.java
```diff
@@ -270,28 +269,14 @@ public void populateEnvironmentContext(EnvironmentVariableContext context, Mater
         context.setProperty(getEnvironmentVariableKey("GO_SCM_%s_%s", "LABEL"), materialRevision.getRevision().getRevision(), false);
         for (ConfigurationProperty configurationProperty : scmConfig.getConfiguration()) {
             context.setProperty(getEnvironmentVariableKey("GO_SCM_%s_%s", configurationProperty.getConfigurationKey().getName()),
-                    configurationProperty.getValue(), configurationProperty.isSecure());
+                    configurationProperty.getResolvedValue(), configurationProperty.isSecure() || configurationProperty.hasSecretParams());
```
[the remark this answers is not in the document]
kritika-singh3: The `isSecure` method is used to decide whether to send the encrypted value or the normal value. Hence, the same was not altered.

maheshp:
Verified on `GoCD Version: 20.8.0 (12188-50837287b9d032638df921edd2a5fe12fd6090d9).` using the Git Path Material plugin.

Issue #8502, opened 2020-08-31
Dashboard UI issue: Repeated trigger buttons


varshavaradarajan:
##### Issue Type

- Bug Report

##### Summary

<img width="275" alt="Screen Shot 2020-08-30 at 6 41 12 PM" src="https://user-images.githubusercontent.com/4715748/91675580-bb61fb00-eaf1-11ea-8102-208bcbc1b244.png">

###### Basic environment details

* Go Version: `20.8.0 (12113-927c25d15ecaeab2a160f1b23f40f44acc96a20b).`
* Browser vendor and version (if relevant): `Chrome Version 84.0.4147.135 (Official Build) (64-bit)`

##### Steps to Reproduce

Not sure really. This is when I observed it
1. Click on a stage on the dashboard and wait for a state change?
2. Stage state changes from building to failed
3. And after that every dashboard refresh seems to add a new trigger buttons

GaneshSPatil:
### Issue Description

This issue seems to be happening due to invalid-state response from Stage Instance API. When a job task is failed, and the job is in uploading artifact state, the API responds as **_job -> completing_** (uploading artifacts) and **_result -> failed_**. 

##### Screenshots
<img width="1785" alt="Screenshot 2020-08-26 at 12 22 12 PM" src="https://user-images.githubusercontent.com/15275847/91701268-a9856400-eb94-11ea-9732-a7a019a282f8.png">
<img width="654" alt="Screen Shot 2020-08-31 at 11 29 39 AM" src="https://user-images.githubusercontent.com/15275847/91701284-ae4a1800-eb94-11ea-9cf7-c88046c90a6c.png">


##### Stage Instance API Response
```json
{
  "name" : "asd",
  "counter" : 3,
  "approval_type" : "success",
  "approved_by" : "admin",
  "result" : "Unknown",
  "rerun_of_counter" : null,
  "fetch_materials" : true,
  "clean_working_directory" : false,
  "artifacts_deleted" : false,
  "pipeline_name" : "new-pipeline",
  "pipeline_counter" : 14,
  "jobs" : [ {
    "name" : "asd",
    "state" : "Completing",
    "result" : "Failed",
    "scheduled_date" : 1598853338897,
    "rerun" : false,
    "original_job_id" : null,
    "agent_uuid" : "5c2ef605-58ce-42d3-ac15-b8bbcc60c81c",
    "pipeline_name" : null,
    "pipeline_counter" : null,
    "stage_name" : null,
[...]

PR #8505, opened 2020-08-31
Fix stage overview rendering issue do to invalid state in stage instance API (#8502)

GaneshSPatil:
Issue: #8502


kritika-singh3:
Stage overview is rendered without any issue in case of a job state is `completing` but the result is `failed`.
Verified on `GoCD Version: 20.8.0 (12151-3f21177a39ee10a9c4e0b7826716564e1e8ee23d)`.

PR #8521, opened 2020-09-03
Upgrade to Gradle 6

marques-work:
- [x] Update Gradle wrapper
- [x] Fix incompatibilties

The real hard work is in #8557


maheshp:
Verified locally and our builds are green after the upgrade.

PR #8522, opened 2020-09-04
Exponential backoff material update in case of failures

maheshp:
* In cases where material update fails continuously, exponentially backoff subsequent updates to reduce unnecessary retries and errors in logs.
* Only material updates through the timer are checked for previous failures and backed-off exponentially.
* The exponential backoff logic ensures a delay between MDU for failing updates. The calculation for the delay is **(time_ between_2_subsequent_failures * MULTIPLIER)**. The MULTIPLIER is defaulted to 1.5, which can be changed by the GoCD System property `gocd.mdu.exponential.backoff.multiplier`
* The backoff logic does not consider any max retires, just that the delay interval increases exponentially.
* A successful MDU would ensure subsequent MDU for the material are not backed-off.
* To force an MDU, material update can be triggered using the manual pipeline trigger or the trigger through the new Materials page.
* Material Update through webhooks do not backoff.

### DIFF
--- a/base/src/main/java/com/thoughtworks/go/util/Clock.java
+++ b/base/src/main/java/com/thoughtworks/go/util/Clock.java
@@ -19,6 +19,7 @@ import com.thoughtworks.go.utils.Timeout;
 import org.joda.time.DateTime;
 
 import java.sql.Timestamp;
+import java.time.LocalDateTime;
 import java.util.Date;
 
 public interface Clock {
@@ -27,6 +28,8 @@ public interface Clock {
     DateTime currentDateTime();
 
     Timestamp currentTimestamp();
+
+    LocalDateTime currentLocalDateTime();
 
     long currentTimeMillis();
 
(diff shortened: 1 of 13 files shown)


arvindsv:
What is the limit of the backoff? Usually, it goes something like 1, 2, 4, 8, 16, 32, 64, 64, 64, ... Otherwise, it can lead to overflows / never recovering etc.

Unrelated to this: Now that there is a materials page, it might be good to not log material failures at WARN level. Could maybe move to DEBUG and show in the materials page.

maheshp:
> What is the limit of the backoff? 

Currently there are no limits set, I was thinking if we should have a `MAX_RETRY_LIMIT` (maybe 10) and stop the MDU retires when the limit is reached. Or does it makes sense to have a `MAX_RETRY_INTERVAL`.

arvindsv:
I think a max interval makes sense. Otherwise, if the retries end, then you'd never pick up that material automatically if it was a temporary network issue or something.

maheshp:
Ok, let's go with a `MAX_RETRY_INTERVAL` will default it to an hour.

kritika-singh3 (approved):

kritika-singh3:
Verified on `GoCD Version: 20.8.0 (12178-568f4d8355b95a48759e83181c313092d267e90a)` (via zip installer)

Issue #8524, opened 2020-09-04
Test Connection with ## in Repository URL doesn't work in Pipeline Wizard


JeroenOortwijn:
##### Issue Type

- Bug Report

##### Summary

Testing the connection to an SVN repository in the Pipeline Wizard doesn't work when it contains two hash signs (##).

###### Basic environment details

* Go Version: `20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed)`
* JAVA Version: `13.0.2`
* OS: `Linux 5.4.0-45-generic`
* Browser vendor and version (if relevant): `not relevant`

###### Additional Environment Details

None

##### Steps to Reproduce

1. From the dashboard, click on `New Pipeline` -> `Use Pipeline Wizard`
2. Select `Subversion` as material type.
3. Fill in `svn://10.0.0.250/Software/MoveTo/trunk/C##` as repository URL.
  The repository URL is `svn://10.0.0.250/Software/MoveTo/trunk/C#`, but the hash sign needs to be escaped by another hash sign.
4. Click on Test Connection.

##### Expected Results
The message
```
Connection OK
```
appears.

##### Actual Results
The message
```
svn: Malformed URL svn://10.0.0.250/Software/MoveTo/trunk/C## : 
Error performing command: --- Command ---
svn log --non-interactive --xml -v --limit 1 svn://10.0.0.250/Software/MoveTo/trunk/C##
--- Environment ---
{}
--- INPUT ----


--- EXIT CODE (1) ---
--- STANDARD OUT ---
<?xml version="1.0" encoding="UTF-8"?>
<log>
--- STANDARD ERR ---
STDERR: svn: E160013: File not found: revision 3737, path '/MoveTo/trunk/C##'
---
```
appears.

##### Possible Fix
Use the same code for the `Test Connection` button as the one on the `Edit material - Svn` page. The `Test Connection` button on that page does work with the `svn://10.0.0.250/Software/MoveTo/trunk/C##` repository URL.

##### Log snippets
None

##### Code snippets/Screenshots
Screenshot showing error in `Material Wizard`:
> ![Annotation 2020-09-04 140206](https://user-images.githubusercontent.com/51476918/92238451-971f6900-eeb9-11ea-818a-c61163918c6b.png)

Screenshot showing correct result on `Edit material - Svn` page:
[...]

PR #8526, opened 2020-09-07
Update material check connection in case pipeline name is not provided

kritika-singh3:
Issue: #8524

Description:
 - In case of check connection from new pipeline wizard, the pipeline name is not provided. Hence param expansion was not done.
   But it also caused the `#` to not being escaped. Hence, creating a dummy pipeline config object for such cases

### DIFF
--- a/api/api-abstract-material-test/src/main/java/com/thoughtworks/go/api/abstractmaterialtest/AbstractMaterialTestController.java
+++ b/api/api-abstract-material-test/src/main/java/com/thoughtworks/go/api/abstractmaterialtest/AbstractMaterialTestController.java
@@ -40,7 +40,6 @@ import com.thoughtworks.go.server.service.CheckConnectionSubprocessExecutionCont
 import com.thoughtworks.go.server.service.GoConfigService;
 import com.thoughtworks.go.server.service.MaterialConfigConverter;
 import com.thoughtworks.go.server.service.SecretParamResolver;
-import com.thoughtworks.go.spark.spring.SparkSpringController;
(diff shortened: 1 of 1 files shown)


api/api-abstract-material-test/src/main/java/com/thoughtworks/go/api/abstractmaterialtest/AbstractMaterialTestController.java
```diff
@@ -126,12 +123,18 @@ private String handleValidationBeanResponse(ValidationBean validationBean, Respo
...
+        PipelineConfig pipelineConfig;
+        // If the pipeline name is provided, find the pipeline and add the params to the new pipeline config object
+        if (isNotBlank(pipelineName)) {
+            PipelineConfig existingPipeline = goConfigService.pipelineConfigNamed(new CaseInsensitiveString(pipelineName));
+            pipelineConfig = new PipelineConfig(existingPipeline.name(), new MaterialConfigs());
 
+            GoConfigCloner goConfigCloner = new GoConfigCloner();
+            pipelineConfig.setParams(goConfigCloner.deepClone(existingPipeline.getParams()));
+        } else {
+            // If the pipeline name is not provided, this means that the pipeline is still in creation nd hence no params exist
```
maheshp, line 135: @kritika-singh3 do you think we can add a test for these scenarios over adding a comment?

maheshp:
Verified on `GoCD Version: 20.8.0 (12188-50837287b9d032638df921edd2a5fe12fd6090d9).`

Issue #8525, opened 2020-09-07
Backup breaks when WRAPPER_CONF_DIR is not set


dAnjou:
##### Issue Type

- Bug Report

##### Summary

Without having the environment variable WRAPPER_CONF_DIR set, the backup via admin interface breaks.

We've noticed this while upgrading our 19.6.0 deployment to 20.1.0. The relevant commit is https://github.com/gocd/gocd/pull/7312/commits/7768b20842fe8c3adc3f0ce7d8cd98475a5926ce and the issue triggering the change is https://github.com/gocd/gocd/issues/7301.

###### Basic environment details

* Go Version: 20.1.0, but looking at the code it seems the behavior has not changed in version 20.7.0
* JAVA Version: 11.0.8
* OS: Amazon Linux 2

##### Steps to Reproduce

1. Deploy GoCD server 
2. Perform backup via UI `/go/admin/backup`

##### Expected Results

Backup succeeds

##### Actual Results

Backup fails at step `Backing up Wrapper Configuration` with error message `Could not find wrapper-config directory`.

##### Possible Fix

We worked around it by setting the environment variable to [the suggested directory for our installation](https://docs.gocd.org/current/installation/install/server/linux.html#location-of-gocd-server-files).

maheshp:
@dAnjou the environment variable `WRAPPER_CONF_DIR` is set by default by Tanuki the Java Service Wrapper used by GoCD. I verified backups on `v20.1.0` and `v20.7.0` using a zip installer and it works as expected.

Which installer are you using for you GoCD server?  Can you share the output of the API support logs, it can be fetched by opening - `http://<go-server-host>/go/api/support`. We don't need the entire output, I am interested in the `Environment Variables` set.

dAnjou:
Hi @maheshp, thanks for the response.

I don't feel comfortable posting all the environment variables here, there are over 1600 and some might contain sensitive information. We've checked `/go/api/support` while debugging the issue and `WRAPPER_CONF_DIR` wasn't set, after we've set it manually it showed up in that list.

We're installing GoCD on Amazon Linux 2, here's a snippet from our Dockerfile:
```
curl https://download.gocd.org/gocd.repo -o /etc/yum.repos.d/gocd.repo && \
  yum -y install go-server-${GOCD_PACKAGE_VERSION}-${GOCD_PACKAGE_BUILD_NUMBER} 
```
And then we're starting the server like this:
```
java -XX:+HeapDumpOnOutOfMemoryError
[...]

PR #8539, opened 2020-09-10
Backing up wrapper config dir is optional now #8525

maheshp:
* As part of backing up the wrapper config folder, the
  folder location is determined using the env variable
  WRAPPER_CONF_DIR which is set by Tanuki wrapper during startup.
  Starting the GoCD server not as a service would not set the
  env variable and would hence fail the backup. Hence making
  it optional.

Issue: #8525

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
@@ -173,13 +173,13 @@ public class BackupService implements BackupStatusProvider {
                 }
                 backupVersion(destDir, backupUpdateListeners);
                 backupConfig(destDir, backupUpdateListeners);
-                backupWrapperConfig(destDir, backupUpdateListeners);
+                boolean backedUpWrapperConfig = backupWrapperConfig(destDir, backupUpdateListeners);
                 backupConfigRepo(backupUpdateListeners, destDir);
                 backupDb(destDir, backupUpdateListeners);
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/thoughtworks/go/server/service/BackupService.java
```diff
@@ -224,8 +224,11 @@ private void notifyErrorToListeners(List<BackupUpdateListener> listeners, String
         listeners.forEach(backupUpdateListener -> backupUpdateListener.error(message));
     }
 
-    private void notifyCompletionToListeners(List<BackupUpdateListener> listeners) {
-        listeners.forEach(BackupUpdateListener::completed);
+    private void notifyCompletionToListeners(List<BackupUpdateListener> listeners, boolean backedupWrapperConfig) {
+        String message = backedupWrapperConfig
+                ? "Backup was generated successfully."
+                : "Backup was generated successfully. WrapperConfig backup was skipped, as the Wrapper Config Dir path is unknown.";
```
kritika-singh3, line 230: How about:
```
Backup was generated successfully. The backup of wrapper configuration was skipped as the 'wrapper_config_dir' path is unknown.
```

kritika-singh3 (approved):

kritika-singh3:
Verified on `GoCD Version: 20.8.0 (12178-568f4d8355b95a48759e83181c313092d267e90a)` (via generic installer)

PR #8540, opened 2020-09-10
Fix job rerun icon positioning on stage overview

GaneshSPatil:
BUG:
	Due to absolute positioning of the rerun icon, icons are rendered outside of the stage overview and are rendered at the fixed position even when scrolling.

Screen Shot 2020-09-10 at 3 43 56 PM
Screen Shot 2020-09-10 at 3 44 05 PM

FIX:
	Use relative positioning. Align rerun icon on the job state icon using transform and left positioning.


kritika-singh3:
Verified on `GoCD Version: 20.8.0 (12151-3f21177a39ee10a9c4e0b7826716564e1e8ee23d)`

PR #8557, opened 2020-09-15
Gradle 6 deprecations

marques-work:
Removing deprecated dependency configurations and other miscellany that will be removed in Gradle 7.0.

Many of these features have existed for several major versions of Gradle, so this is long overdue.

- [x] change dependency scopes away from deprecated `compile, testCompile, runtime, testRuntime` in favor of `implementation` and friends
- [x] upgrade plugins to remove other deprecations
    * NOTE: not all can be removed because certain plugins we are using (e.g., license reporting) have not resolved deprecation warnings

## Note to reviewers

I also removed an explicit dependency on `commons-digester3` in an attempt to clean up some old deps. It required quite a bit of refactoring in the `go-plugin-infra` module. I gave up doing other upgrades after that. If it makes the diff too complex, I can pull it out as a separate PR.


build.gradle
```diff
@@ -423,42 +483,11 @@ subprojects {
...
-          compileClasspath += configurations.compileOnly
-        }
-      }
-      test {
-        java {
-          compileClasspath += configurations.compileOnly
-          runtimeClasspath += configurations.compileOnly
-        }
-      }
-    }
```
marques-work, line 443: This bit is partly redundant since `compileClasspath` already `extendsFrom compileOnly`.

Adding the `compileClasspath` to `test{Compile,Runtime}Classpath` is achieved on line 296:

```groovy
    // it's rather convenient to allow compile and runtime classpaths for tests inherit from the main compile classpath
    testImplementation.extendsFrom compileClasspath
```

server/build.gradle
```diff
@@ -88,14 +88,7 @@ sourceSets {
 configurations {
   jrubyGems
 
-  sharedTestCompile.extendsFrom testCompile
-  sharedTestRuntime.extendsFrom testRuntime
-
-  fastUnitTestCompile.extendsFrom sharedTestCompile
-  fastUnitTestRuntime.extendsFrom sharedTestRuntime
```
marques-work, line 95: None of these were used (anymore?).

server/build.gradle
```diff
@@ -110,19 +103,15 @@ configurations {
...
 
-  compile.exclude(group: 'org.springframework', module: 'spring-support')
-  runtime.exclude(group: 'org.springframework', module: 'spring-support')
-
-  // this dependency is a "provided" dependency
-  compile.exclude(group: 'ch.qos.logback', module: 'logback-classic')
-  runtime.exclude(group: 'ch.qos.logback', module: 'logback-classic')
-  apiBase.exclude(group: 'ch.qos.logback', module: 'logback-classic')
-  api.exclude(group: 'ch.qos.logback', module: 'logback-classic')
-  spark.exclude(group: 'ch.qos.logback', module: 'logback-classic')
```
marques-work, line 125: All of this exclusion logic is handled during packaging in the `war` task. See the lines pertaining to configurations "arithmetic" when building the `classpath`.

server/build.gradle
```diff
@@ -725,23 +723,40 @@ task war(type: Jar) {
...
+  dependsOn project.configurations.allApis
   dependsOn project.configurations.spark
   finalizedBy 'verifyWar'
   archiveBaseName = 'cruise'
-  extension 'war'
+  archiveExtension = 'war'
+
+  // Using a Set<String> is more reliable than Set<File> because File.equals() may be affected by other internal fields
+  // aside from `path` that aren't important to us; thus, Set<File> may allow duplicate paths and Set<String> will not.
+  Set<String> classpath = (
```
marques-work, line 735: In other words, `Set<String>` is better at deduplicating paths to jars than `Set<File>`.

maheshp:
Manually verified  and we have our builds going for a while. build.gocd has been dog-fooded with the generated installers.

PR #8266, opened 2020-06-25
Update to allowlist/denylist

arvindsv:
Issue: #8263 

Description:
  - Change blacklist / whitelist in most places to either includes / excludes or allowlist / denylist.
  - In some places it's inside data (like the dashboard's personalization view sending it), which, if changed will break everyone's selected pipelines. It's internal and won't be seen anywhere. Haven't changed it. If anyone wants to, it's open-source and a PR will be accepted, as long as it doesn't affect existing users.
  - This PR also introduces `target_version: 10` for the internal config-repo JSON format (between CR plugins and GoCD). A similar PR will need to be opened for each plugin, so that they support v10. Just the YAML and Groovy plugins. The JSON one should work out of the box.

Todos:
- [x] Verify changes manually
- [ ] Submit PR to config repo plugins [Later, not necessarily immediately]
- [x] Update documentation (https://github.com/gocd/docs.go.cd/pull/468)
- [x] Check if any functional tests are affected (https://github.com/gocd/ruby-functional-tests/pull/670)

PR #8559, opened 2020-09-15
Updated yaml plugin to 0.13.0

tomzo:
Issue: #8266

Description:
Updates YAML plugin to 0.13.0 - which now supports format_version: 10.

What I tested:
Built GoCD from this branch.
Started a dev server.
Added a config repo with format_version:9 and some whitelist/blacklist uses in it
Renamed whitelist/blacklist to includes/ignore
Switched format version to 10

Commit history for the test: https://github.com/tomzo/gocd-yaml-config-example/commits/f10

Basically, a graceful way of upgrading and migrating the keywords will be possible now.

### DIFF
--- a/Dojofile
+++ b/Dojofile
@@ -1,2 +1,2 @@
 # For building GoCD in docker see https://github.com/kudulab/docker-gocd-dojo
-DOJO_DOCKER_IMAGE=kudulab/gocd-dojo:2.1.0
+DOJO_DOCKER_IMAGE=kudulab/gocd-dojo:2.2.0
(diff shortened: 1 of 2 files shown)


marques-work:
@tomzo I think (if I understand the changes correctly) that MaterialTransform.java might have the wrong keyword (`excludes` vs `ignores`)?

See https://github.com/tomzo/gocd-yaml-config-plugin/pull/148#discussion_r488908236

tomzo:
Thanks, let me investigate that

tomzo:
@marques-work it turns out that this piece of code is irrelevant. I've started to cleanup the inverse transform code and I'll release another version later. But for now, this plugin version is functional and tested, so I would go with it for the 20.8 release.

Let me know if you're OK with that.

marques-work:
@tomzo yes, I think that’s perfectly fine. I’ll merge this.

maheshp:
Verified the bundled plugin version is `0.13.0` on `GoCD Version: 20.8.0 (12211-5d6b4a7a7145f728857afb8108b31effbbd1cd0c).`

PR #8571, opened 2020-09-21
[Stage Overview] Fix in progress stage name on disabled rerun icon on stage activity page.

GaneshSPatil:
* Stage overview expects the stage name to be present on 'name' field,
  whereas, it was passed as 'stageName()' from the pipeline activity page.

* Add another field 'name' on the stage model before passing the model
  to the stage overview.

#### Before
Screen Shot 2020-09-21 at 8 51 17 AM

#### After
Screen Shot 2020-09-21 at 8 50 27 AM


kritika-singh3:
Verified on `GoCD Version: 20.8.0 (12204-0e9d2bd06e51e694040a89f9b6b8dcb6978d41c5)`

PR #8576, opened 2020-09-22
Show Stage Overview on the VSM pipelines.

GaneshSPatil:
* Add a DOMContentLoaded event listener to define handlers which
  renders stage overview when vsm pipeline is clicked.

* load typescript files on the vsm page.

* Mount the stage overview inside the temporary div of the vsm
  pipeline every time when the stage box is clicked.

* Render stage overview in the centre of the stage bar.

* Fix stage overview position when there are multiple stages in a
  pipeline. The total width of the stage is 237px, the width of each
  stage can be computed by dividing the total width by the
  total number stages (considering the initial margin and inbetween
  spaces).

* Fix stage overview positioning for unfocused pipelines. The focused
  pipeline is bigger in size than other pipelines. The focused pipeline
  has width 237px, whereas, the un-focused pipelines has width 192px.
  Determine the stage center to render the stage overview, based on the
  width of the pipeline.

* [Bug fix] Close the existingly opened stage overview.
  Clicking on multiple stage bars would keep the previous stage overview
  open.

* Fix the stage settings link for pipeline using template.
  Add the template information to the VSM API.

* Do not render rerun/cancel stage button.

* Do not render rerun-failed/rerun-selected button.

* Remove job checkboxes.

* Fix job count container mis-alignment.

* Do not refresh stage overview (Disable poller). The VSM page is not
  auto refreshed, rendering the fresh information on the vsm page may result
  into inconsistent stage state on the view.

Screen Shot 2020-09-22 at 8 52 57 AM


kritika-singh3:
Tested on `GoCD Version: 20.9.0 (12236-550797553265f0d8918d65356b5a59f846794cdc)` and issues reported in #8595

PR #8579, opened 2020-09-22
Fix stage overview caret icon positioning

GaneshSPatil:
##### Bug:

When the stage overview for the rightmost pipeline on the dashboard is opened, the overview caret icon is mis-positioned initially and on the subsequent redraw its positioned appropriately.

##### Why did it happen?:

The 'shouldAlignLeft' state was changed in the oncreate lifecycle method, whereas, any state changes done in the oncreate does not performs an immediate redraw. Such changes will be reflected on the next periodic redraw.
This behavior caused the caret icon to be positioned appropriately on the subsequent redraws.

##### Fix:
Do not store/update state as part of the oncreate method. As we've the handle to the created element's DOM, add appropriate classes to the DOM element within oncreate.


kritika-singh3:
Verified on `GoCD Version: 20.8.0 (12213-1e23a06e496205ced5f1a8e83d9b209fc0a290cb)`

Issue #1926, opened 2016-02-12
Job Timeout should apply even when the job could not be assigned to an agent


rajiesh:
We encountered this scenario in build.go.cd where the pipeline `installer-testing` was pending for assignment for over 18 hours since the agent was missing. Even though the job timeout on build.go.cd is setup to 180 mins, it do not apply in this scenario where the job assignment is not completed.

I think its good to apply the timeout on non-assigned jobs too since a job waiting for 18 hrs do not have any value. Better we timeout and fail the job so that it can be notified to take action.

Another observation was when the agent came up, the job `installer-testing` failed after having waited for 18 hours, and below exception seen in the logs,

```
2016-02-11 22:22:01,683  INFO [qtp399573350-130769] BuildRepositoryRemoteImpl:133 - [Agent Cookie] Agent [Agent [blrstdgoprf02.thoughtworks.com, 10.4.7.1, 1bdd029e-3c33-4928-a6a8-f3bd46d30c7e]] at location [/var/lib/go-agent] asked for a new cookie, assigned [4258312b-3fdf-4b72-874b-049af41253ce]
2016-02-11 22:22:02,008  INFO [67@MessageListener for WorkFinder] Stage:218 - Stage is being completed by transition id: 887554
2016-02-11 22:22:02,061 ERROR [67@MessageListener for WorkFinder] StageService:235 - error notifying listener for stage <Stage id='15283' name='fresh-install' stageCounter='1' createdTime='2016-02-11 04:23:57.133' result='Failed'>
java.lang.RuntimeException: Multiple independent transactions are not permitted inside single transaction surrounding.
        at com.thoughtworks.go.server.transaction.TransactionContext.transactionPushed(TransactionContext.java:47)
        at com.thoughtworks.go.server.transaction.TransactionTemplate$2.doInTransaction(TransactionTemplate.java:37)
        at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:130)
        at com.thoughtworks.go.server.transaction.TransactionTemplate.execute(TransactionTemplate.java:35)
        at com.thoughtworks.go.server.dao.EnvironmentVariableSqlMapDao.load(EnvironmentVariableSqlMapDao.java:60)
        at com.thoughtworks.go.server.dao.JobInstanceSqlMapDao.loadJobPlanAssociatedEntities(JobInstanceSqlMapDao.java:208)
[...]

PR #8585, opened 2020-09-23
Use jobTimeout to cancel unassigned jobs

kritika-singh3:
Issue: #1926 

Description:
- Earlier only the building jobs could be cancelled post a timeout set via `jobTimeout` property.
   Now, using the same value, unassigned jobs will be cancelled.

P.S. the original feature is still in place. This is additional feature over the existing

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/service/ConsoleActivityMonitor.java
+++ b/server/src/main/java/com/thoughtworks/go/server/service/ConsoleActivityMonitor.java
@@ -15,24 +15,27 @@
  */
 package com.thoughtworks.go.server.service;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-
 import com.thoughtworks.go.domain.JobIdentifier;
 import com.thoughtworks.go.domain.JobInstance;
 import com.thoughtworks.go.server.domain.JobStatusListener;
-import com.thoughtworks.go.serverhealth.HealthStateScope;
 import com.thoughtworks.go.serverhealth.HealthStateType;
 import com.thoughtworks.go.serverhealth.ServerHealthService;
 import com.thoughtworks.go.serverhealth.ServerHealthState;
 import com.thoughtworks.go.util.SystemEnvironment;
 import com.thoughtworks.go.util.TimeProvider;
-import org.slf4j.Logger;
 import org.joda.time.DateTime;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/thoughtworks/go/server/service/ConsoleActivityMonitor.java
```diff
@@ -124,6 +139,59 @@ private long jobTerminationThreshold(JobIdentifier jobIdentifier) {
...
+        }
+
+        @Override
+        public void jobStatusChanged(JobInstance job) {
+            JobIdentifier identifier = job.getIdentifier();
+            if (job.getState().isScheduled()) {
+                consoleActivityMonitor.jobScheduledMap.putIfAbsent(identifier, consoleActivityMonitor.timeProvider.currentTimeMillis());
+            } else if (job.getState().isActiveOnAgent() || job.isCompleted()) {
+                consoleActivityMonitor.jobScheduledMap.remove(identifier);
+                consoleActivityMonitor.removeHungJobWarning(identifier);
```
maheshp, line 190: Should we remove warning only if the `jobScheduledMap` has the identifier. Just wanted to make sure we do not clear out warnings added through `jobLastActivityMap`

maheshp (approved):
LGTM, can you please see if we need to change the logic around clearing the job hung warnings.

server/src/main/java/com/thoughtworks/go/server/service/ConsoleActivityMonitor.java
kritika-singh3: A job will be first added to `jobScheduledMap` when it gets scheduled. As soon as it gets assigned it will be removed from the map. As it starts building it will be added to `jobLastActivityMap`.
Hence, the same job identifier cannot be present in both the maps at the same time. 
Therefore, I think this is a safe operation.

maheshp: I get your point about how the jobs are added to the map.

Since the `ScheduledJobListener#jobStatusChanged` would be called for jobs which are part of `jobLastActivityMap`, was wondering if there will be cases where we end up clearing warnings for jobs which are in building state in this listener.

maheshp:
Verified in `GoCD Version: 20.9.0 (12292-25ed9fb08c2f5de81a8c4ddacc9c77bba732109f).`

PR #8589, opened 2020-09-24
Removed Scheduled Job API deprecated in 20.6.0 #000

maheshp:
Removed Scheduled Job API deprecated in 20.6.0


server/src/main/webapp/WEB-INF/rails/config/routes.rb
```diff
@@ -46,31 +46,11 @@
   put "admin/config_xml" => "admin/configuration#update", as: :config_update
   get "admin/config_xml/edit" => "admin/configuration#edit", as: :config_edit
 
-  resources :analytics, only: [:index], controller: "analytics"
-  get 'analytics/:plugin_id/:type/:id' => 'analytics#show', constraints: {plugin_id: PLUGIN_ID_FORMAT, id: PIPELINE_NAME_FORMAT}, as: :show_analytics
```
kritika-singh3, line 50: IIRC, this is still used on VSM page. Please check.

maheshp: I dint see the analytics controller, hence I removed it. Will check.

kritika-singh3:
Verified via API regression specs.

PR #8591, opened 2020-09-24
Update stage overview caret when stage status changes.

GaneshSPatil:
* Update the stage overview caret from older status to the updated
  status when the stage status changes.

#### Bug Screenshot
Screen Shot 2020-09-23 at 7 47 11 PM


kritika-singh3:
Verified on `GoCD Version: 20.9.0 (12236-550797553265f0d8918d65356b5a59f846794cdc)`

PR #8592, opened 2020-09-24
Add secrets support elastic agent configurations

kritika-singh3:
Description:
Add support for secret params in elastic configurations (cluster profile and elastic agent profile).
The secrets will be resolved before sending the following plugin requests:

 - validateClusterProfile
 - validateElasticProfile
 - createAgent
 - serverPing
 - shouldAssignWork
 - getPluginStatusReport
 - getClusterStatusReport
 - getAgentStatusReport
 - jobCompletion
 - clusterProfilesChanged

The secret resolution will not be done before `migrateConfig` as the plugin settings need to be migrated to cluster profile as in i.e. without secret resolution.


server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -134,14 +139,15 @@ public void heartbeat() {
 
         for (PluginDescriptor descriptor : elasticAgentPluginRegistry.getPlugins()) {
             List<ClusterProfile> clusterProfiles = clusterProfilesService.getPluginProfiles().findByPluginId(descriptor.id());
+            resolveSecrets(clusterProfiles);
```
maheshp, line 142: If secrets resolution for one of the cluster profile fails, the plugin would not get a ping for any of the clusters. 

How do we notify users in cases of secret resolution failures?

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -237,14 +244,15 @@ private void logToJobConsole(JobIdentifier identifier, String message) {
     }
 
     public boolean shouldAssignWork(ElasticAgentMetadata metadata, String environment, ElasticProfile elasticProfile, ClusterProfile clusterProfile, JobIdentifier identifier) {
-        Map<String, String> clusterProfileProperties = clusterProfile != null ? clusterProfile.getConfigurationAsMap(true) : Collections.EMPTY_MAP;
-        GoPluginDescriptor pluginDescriptor = pluginManager.getPluginDescriptorFor(metadata.elasticPluginId());
-        Map<String, String> configuration = elasticProfile.getConfigurationAsMap(true);
-
-        if (!StringUtils.equals(clusterProfile.getPluginId(), metadata.elasticPluginId())) {
+        if (clusterProfile == null || !StringUtils.equals(clusterProfile.getPluginId(), metadata.elasticPluginId())) {
```
maheshp, line 247: In what cases would the `clusterProfile` be `null`, I am guessing for the Azure plugin. Can you please check?
Returning `false` in these conditions might break things?

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -237,14 +244,15 @@ private void logToJobConsole(JobIdentifier identifier, String message) {
...
             return false;
         }
 
+        resolveSecrets(clusterProfile, elasticProfile);
```
maheshp, line 251: What happens when secret resolution fails and what should we do on failure?

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -311,9 +330,25 @@ public void jobCompleted(JobInstance job) {
...
         ElasticProfile elasticProfile = job.getPlan().getElasticProfile();
         ClusterProfile clusterProfile = job.getPlan().getClusterProfile();
-        Map<String, String> elasticProfileConfiguration = elasticProfile.getConfigurationAsMap(true);
-        Map<String, String> clusterProfileConfiguration = clusterProfile != null ? clusterProfile.getConfigurationAsMap(true) : Collections.EMPTY_MAP;
-
+        secretParamResolver.resolve(elasticProfile);
+        Map<String, String> elasticProfileConfiguration = elasticProfile.getConfigurationAsMap(true, true);
+        Map<String, String> clusterProfileConfiguration = emptyMap();
+        if (clusterProfile != null) {
+            secretParamResolver.resolve(clusterProfile);
```
maheshp, line 337: How do we handle secret resolution failures?

We are removing the job from `jobCreationTimeMap` before secrets are resolved. Will there be any repercussions of secret resolution failures after removing it from the map.

config/config-api/src/main/java/com/thoughtworks/go/domain/config/Configuration.java
```diff
@@ -158,10 +159,23 @@ public void addError(String fieldName, String message) {
...
+
+
+    /**
+     * Returns the configuration values a map of key-value pair
+     * @param addSecureFields should add secure fields
+     * @param useResolvedValue should try to resolve any secret params found.
+     * @throws UnresolvedSecretParamException if useResolvedValue is set to true and the secrets are not resolved
+     * @return map of configs
+     */
+    public Map<String, String> getConfigurationAsMap(boolean addSecureFields, boolean useResolvedValue) {
```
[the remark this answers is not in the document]
kritika-singh3: Yes. When we are calling the plugins which does not support secrets as of yet.
E.g. authorization, artifact.

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
kritika-singh3: Yes, in case the GoCD server was upgraded and their were few scheduled jobs which now won't have a cluster profile.
No it would not. The check was preset earlier as well. Also, we do a job reload based on timer, which checks for the above scenario and cancels the job with a message. Hence, sending false won't break things.

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
kritika-singh3: If secret resolution fails for any cluster profiles, the server ping call will do for the remaining of the clusters. Also, we will not be able to notify the users about the agents for which the plugin has been deleted.
While the error gets logged in server logs, do you think we should catch them nd display a server health message?

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
kritika-singh3: The `shouldAssignWork` is called when a elastic agent registers with GoCD. In case, secret resolution fails, the agent will not be assigned any job nd will remain idle.
We can catch them nd ignore it.

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
kritika-singh3: This is called when a job gets completed. Secret resolution failure would mean that the plugin is not notified nd hence will not be able to delete the job. Also, the job plan related information in DB is also not deleted.

kritika-singh3:
@maheshp In major cases of secret resolution failures for heartbeat, jobCompleted - we can catch the specific exceptions, maybe raise a server health message and continue with the normal flow.
For shouldAssign - we should just catch and ignore. WDYT?

kritika-singh3:
Implementation decided upon:
 - for createAgents and shouldAssign call: on secret failure - cancel the job. This is similar to wht we do for secret failure in pipelines
 - for job completed call - catch the rules related exception - add a server health message with a 5 minute timeout and log it
 - for heartbeat - catch the rules exception - add a server health message. This will get cleared up on a heartbeat call with the secrets fixed
 - for status reports - updated the controller to return 422 with the exception message

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -193,68 +201,70 @@ public void createAgentsFor(List<JobPlan> old, List<JobPlan> newPlan) {
...
     public boolean shouldAssignWork(ElasticAgentMetadata metadata, String environment, ElasticProfile elasticProfile, ClusterProfile clusterProfile, JobIdentifier identifier) {
-        Map<String, String> clusterProfileProperties = clusterProfile != null ? clusterProfile.getConfigurationAsMap(true) : Collections.EMPTY_MAP;
-        GoPluginDescriptor pluginDescriptor = pluginManager.getPluginDescriptorFor(metadata.elasticPluginId());
-        Map<String, String> configuration = elasticProfile.getConfigurationAsMap(true);
+        if (clusterProfile == null || !StringUtils.equals(clusterProfile.getPluginId(), metadata.elasticPluginId())) {
+            return false;
+        }
 
-        if (!StringUtils.equals(clusterProfile.getPluginId(), metadata.elasticPluginId())) {
+        boolean secretsResolved = resolveSecrets(identifier, clusterProfile, elasticProfile);
```
maheshp, line 244: It seems wrong for the `shouldAssignWork` method cancelling the job here. I think the caller of this method should handle the secret resolution exceptions and take necessary actions.

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/StatusReportsController.java
```diff
@@ -91,6 +93,9 @@ public ModelAndView pluginStatusReport(Request request, Response response) {
         } catch (UnsupportedOperationException e) {
             String message = String.format("Status Report for plugin with id: '%s' is not found.", pluginId);
             return errorPage(response, 404, "Plugin Status Report", message);
+        } catch (RulesViolationException | SecretResolutionFailureException e) {
+            LOGGER.error(e.getMessage(), e);
+            return errorPage(response, 422, "Plugin Status Report", e.getMessage());
```
maheshp, line 98: This should not be `422`, it's not a user error.

kritika-singh3: The exception can be raised due to a secret being configured for a non-existent secret config id. Which is a user error.
If not 422, then what? 500?

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
kritika-singh3: I don't get it. If the service can cancel the job in the `createAgent` call, why can't it do so in `shouldAssignWork`?

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/StatusReportsController.java
maheshp: > The exception can be raised due to a secret being configured for a non-existent secret config id. Which is a user error.

These would be user errors during config update right.

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
```diff
@@ -308,12 +323,73 @@ public void jobCompleted(JobInstance job) {
...
-        elasticAgentPluginRegistry.reportJobCompletion(pluginId, elasticAgentId, job.getIdentifier(), elasticProfileConfiguration, clusterProfileConfiguration);
+    private boolean resolveSecrets(JobIdentifier jobIdentifier, ClusterProfile clusterProfile, ElasticProfile elasticProfile) {
+        try {
+            if (clusterProfile != null)
+                secretParamResolver.resolve(clusterProfile);
+            secretParamResolver.resolve(elasticProfile);
+        } catch (RulesViolationException | SecretResolutionFailureException e) {
+            String cancellationMessage = format("\nThis job was cancelled by GoCD. This job is configured to run on an elastic agent, but the associated elastic configurations failed for secrets resolution: %s", e.getMessage());
+            logToJobConsole(jobIdentifier, cancellationMessage);
+            scheduleService.cancelJob(jobIdentifier);
```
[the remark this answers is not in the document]
kritika-singh3: > You will also need to add a message on the JobStatusTopic so that the JobStatusListener is called.

This is taken care by the schedule service class.

spark/spark-spa/src/main/java/com/thoughtworks/go/spark/spa/StatusReportsController.java
kritika-singh3: Yes but these are not caught during full config save.

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentPluginService.java
maheshp: Hmmm, I think in both the cases I would have preferred the ElasticAgentPluginService not cancelling the job, the way create `createAgent` is implemented it makes it tough to make those changes.

server/src/main/java/com/thoughtworks/go/server/service/BuildAssignmentService.java
```diff
@@ -202,9 +202,18 @@ JobPlan findMatchingJob(AgentInstance agent) {
...
-                    break;
+                try {
+                    if (jobPlan.requiresElasticAgent() && elasticAgentPluginService.shouldAssignWork(agent.elasticAgentMetadata(), environmentConfigService.envForPipeline(jobPlan.getPipelineName()), jobPlan.getElasticProfile(), jobPlan.getClusterProfile(), jobPlan.getIdentifier())) {
+                        match = jobPlan;
+                        break;
+                    }
+                } catch (RulesViolationException | SecretResolutionFailureException e) {
+                    JobInstance instance = jobInstanceService.buildById(jobPlan.getJobId());
+                    JobIdentifier jobIdentifier = jobPlan.getIdentifier();
+                    String failureMessage = format("\nThis job was failed by GoCD. This job is configured to run on an elastic agent, but the associated elastic configurations failed for secrets resolution: %s", e.getMessage());
```
maheshp, line 213:
```
This job was failed by GoCD. This job is configured to run on an elastic agent, there were errors while resolving secrets for the the associated elastic configurations.
Reasons: blah blah blah
```

maheshp (commented):
The changes look ok. 
Secrets are resolved when creating/updating the profiles. If there are secret resolution errors it does not show up on the SPA.

maheshp:
Verified on `GoCD Version: 20.9.0 (12311-2f64c36e2234eae2d5d37440ec5ae9cd9d8ba225).`

Issue #5787, opened 2019-01-29
Ability to start server in maintenance mode


GaneshSPatil:
##### Issue Type

- Feature proposal

##### Summary
Currently, we can only put GoCD server into maintenance mode, but can not start a server in the maintenance mode.

Upon Server startup, it will start building pipelines, assigning jobs to agents, performing MDUs, etc.

Imagine a situation where one has migrated a server from one instance to another. Even before we can verify if the new configurations are fine. GoCD will start building.

Add an ability to start GoCD server in maintenance mode using system property.

GaneshSPatil:
#### Possible Solution:

To Start GoCD Server in the maintenance mode, we can introduce a boolean system a property (may be named `gocd.server.enable.maintenance.mode`). And based on the value, the GoCD Server will either start normally, or in the maintenance mode.

#### Tasks

- [ ] Introduce maintenance mode system property to start GoCD Server in the maintenance mode.
- [ ] Fix message on the maintenance mode to mention clearly that the server was started in the maintenance mode (instead of saying `user X updated the maintenance mode state`)
- [ ] Fix message on all page footer to mention clearly that the server was started in the maintenance mode (instead of saying `user X updated the maintenance mode state`)
- [ ] Update [docs](https://docs.gocd.org/20.8.0/advanced_usage/maintenance_mode.html) to mention about the ability to start the GoCD Server in the maintenance mode.
-  [ ] Add a functional spec to start the GoCD Server in the maintenance mode.

GaneshSPatil:
@maheshp @kritika-singh3 @marques-work @gocd/committers -- Thoughts??

kritika-singh3:
@GaneshSPatil Few thoughts:

1. How about `gocd.server.start.in.maintenance.mode` for the system property.
2. Suppose the server is started in maintenance mode. The admin/user verifies everything and disabled it via SPA/API. Now if the server is restarted - will it come up normally or in maintenance mode?
3. Should the config-repo fetch and parse happen if we start the server in maintenance mode?

GaneshSPatil:
@kritika-singh3 --

> 1. How about `gocd.server.start.in.maintenance.mode` for the system property.

Sure. we can use this system property.

> 2. Suppose the server is started in maintenance mode. The admin/user verifies everything and disabled it via SPA/API. Now if the server is restarted - will it come up normally or in maintenance mode?

[...]

PR #8593, opened 2020-09-24
Start GoCD Server in maintenance mode. (#5787)

GaneshSPatil:
* Introduce 'gocd.server.start.in.maintenance.mode' system property
  to specify starting GoCD Server in maintenance mode.

* Update maintenance mode message when server is started in maintenance
  mode on:
	- maintenance mode spa.
	- global page footer.
	- maintenance mode banner at server_in_maintenance_mode.html

PR #8599, opened 2020-09-26
Load GoCD configurations from config repositories on GoCD startup.

GaneshSPatil:
#### Description:
Loading the GoCD configurations from config repositories on GoCD startup  is required when the GoCD server has config repositories and the server is started in maintenance mode.

When the server is started in maintenance mode, the MDU service will be disabled, and the config repository pipelines will never show up. This PR ensures to parse the existing cloned config repositories (from flyweight folder) on startup, making sure config repository pipelines are shown up on startup even in maintenance mode.

Similar conversation at: #5787

#### Uncovered Scenarios:

Following are the scenarios where the config repository pipelines will not be parsed when server is in the maintenance mode:

1. When the config repository is not available under flyweight directory (when material has never parsed).
2. When a pipeline from one config repository depends on pipeline coming from another config repository; And the downstream pipeline config repository is parsed first.
    Example:
    Lets say, config repo repo1 defines upstream pipeline and config repo repo2 defines downstream pipeline, and repo2 config repository is parsed before repo1. In such cases, the upstream pipeline coming from repo1 will not be shown on the UI.

#### Implementation:

* Add a config repo plugin load listener to initialize the GoCD configuration from the existing config repo material repository from flyweight folder.

* Skip initializing config repositories which are not available under flyweight folder (which are not cloned by GoCD yet).

* Always return plugin post load hook result as success, even on config repo initialization failure due to errors (such as config validation / rules violations). As returning a failure response on plugin post load hook will mark plugin as invalid.


maheshp:
@GaneshSPatil what is the need for this PR? ConfigRepo materials are auto updated, hence they would be parsed on the first MDU of the material right.

GaneshSPatil:
@maheshp -- It is required when the GoCD server has config repositories and the server is started in maintenance mode.

When the server is started in maintenance mode, the MDU service will be disabled, and the config repository pipelines will never show up. This PR ensures to parse the existing cloned config repositories (from flyweight folder) on startup, making sure config repository pipelines are shown up on startup even in maintenance mode.

GaneshSPatil:
Though, having said that, following are the scenarios where the config repository pipelines will not be parsed when server is in the maintenance mode:

1. When the config repository is not available under flyweight directory (when material has never parsed).
2. When a pipeline from one config repository depends on pipeline coming from another config repository; And the downstream pipeline config repository is parsed first.
    Example:
    Lets say, config repo `repo1` defines `upstream` pipeline and config repo `repo2` defines `downstream` pipeline, and `repo2` config repository is parsed before `repo1`. In such cases, the `upstream` pipeline coming from `repo1` will not be shown on the UI.

maheshp:
@GaneshSPatil I went through the PR, I had a few thoughts,
1. The ConfigRepo parse happens post the plugin load which is in a separate thread. Hence there could be cases where in the Config Repo is parsed before the main cruise config is migrated and loaded.
2. This PR will parse all the Config Repos irrespective of whether the server is in Maintenance Mode or not. Hence there can be scenarios where in a ConfigRepo parse can be triggered simultaneously through both the plugin load as well as the MDU, wondering if there can be any issues because of this.

@tomzo do you have any thoughts on this PR?

tomzo:
Hi @GaneshSPatil and @maheshp 

I had a quick look at the changes. I'll do a deeper review later today. 

> When a pipeline from one config repository depends on pipeline coming from another config repository; And the downstream pipeline config repository is parsed first.

@GaneshSPatil One thing that I don't immediately understand is why would we skip config-repos that have dependencies? Their dependencies would eventually be parsed, and eventually that should lead to a valid configuration.

> Hence there can be scenarios where in a ConfigRepo parse can be triggered simultaneously through both the plugin load as well as the MDU, wondering if there can be any issues because of this.

@maheshp I wouldn't suspect to have issues because of it. In worst case same thing is parsed twice. The parse and merge operations are both idempotent. We also have the 'refresh' functionality on the UI which is another source of forcing parsing outside of the requests generated in the MDU.

GaneshSPatil:
> > When a pipeline from one config repository depends on pipeline coming from another config repository; And the downstream pipeline config repository is parsed first.
> 
> @GaneshSPatil One thing that I don't immediately understand is why would we skip config-repos that have dependencies? Their dependencies would eventually be parsed, and eventually that should lead to a valid configuration.

@tomzo - sorry for not being clear here. Let me try to explain it again.

[GoCD added an ability to start the GoCD server in maintenance mode](https://github.com/gocd/gocd/pull/8593). When the server is started in maintenance mode, the MDU service will be disabled, and thus, for all the config repo materials, the MDU will be skipped and materials will never be parsed.

To overcome this issue, this PR introduces the config repository initializer service, which on server startup, will go through each config repository material (which is cloned under flyweight folder), and would perform config repo parse.
_This approach would do a parse on each repository only once per startup._

Now, coming back to our example, in case of dependency config repositories (mentioned in [point two of this comment](https://github.com/gocd/gocd/pull/8599#issuecomment-699756421)), lets say, if `repo2` is parsed before `repo1`, the pipelines coming from `repo2` would be invalid.
Also, as the server is in the maintenance mode, the MDU will not be triggered, and the config repository `repo2` will stay in the validation error state (till the admin resumes the maintenance mode, and the MDU starts again for config repo materials).

\cc: @maheshp

arvindsv:
> Now, coming back to our example, in case of dependency config repositories (mentioned in point two of this comment), lets say, if repo2 is parsed before repo1, the pipelines coming from repo2 would be invalid.
Also, as the server is in the maintenance mode, the MDU will not be triggered, and the config repository repo2 will stay in the validation error state (till the admin resumes the maintenance mode, and the MDU starts again for config repo materials).

This is because these repos are parsed only once, right?

In the normal case, during the first run of MDU "repo1" will be valid and the combination will be invalid, and then during the second run, they will both be valid. Right?

It'll be hard to know how many times to run this initialization, if we were to try and make it work. It might _never_ succeed, due to other failures.

server/src/main/java/com/thoughtworks/go/server/service/ConfigRepositoryInitializer.java
```diff
@@ -0,0 +1,150 @@
...
+            return;
+        }
+
+        // do nothing if the cruise config is not loaded yet..
+        if (!this.isConfigLoaded) {
+            return;
+        }
+
+        List<String> inUseConfigRepoPlugins = this.configRepoService.getConfigRepos().stream().map(RuleAwarePluginProfile::getPluginId).collect(Collectors.toList());
+        if (this.loadedConfigRepoPlugins.containsAll(inUseConfigRepoPlugins)) {
```
maheshp, line 121: @GaneshSPatil with this check, in cases where multiple ConfigRepo plugins are used none of the repos will be parsed if a single plugin is missing. Guess this was not intended.

GaneshSPatil: @maheshp - yes, you're right. Will change this check to process the config repositories as and when plugins are laoded.

server/src/main/java/com/thoughtworks/go/server/service/ConfigRepositoryInitializer.java
```diff
@@ -0,0 +1,146 @@
...
+        this.materialRepository = materialRepository;
+        this.goConfigRepoConfigDataSource = goConfigRepoConfigDataSource;
+        this.goConfigService = goConfigService;
+
+        this.pluginManager.addPluginChangeListener(this);
+    }
+
+    @Override
+    public void initialize() {
+        this.goConfigService.register(this);
```
maheshp, line 75: @GaneshSPatil can we register for ConfigChange in the constructor over the `intitalize` method. If we register in the `intialize` method, the config repos will most likely be parsed as part of the main thread and this will lead to increased startup time for GoCD instances with larger configs and multiple config repos. WDYT?

kritika-singh3:
Verified on `GoCD Version: 20.9.0 (12294-720a2fe626694cade5fccf55ef39d79dca6156f7)`

Issue #8546, opened 2020-09-14
Issue with pluggable scm config from config repo


kritika-singh3:
##### Summary

Starting from GoCD 19.2.0, the pluggable SCM configuration can be defined as part of config repo along with pipelines and environments.

For Pipeline Config API and Environment API, we send a readonly property named `origin` which specifies whether the said entity has been defined in GoCD or is coming from a config-repo. But the pluggable SCM API doesn't contains so.
IMO, the pluggable SCM API should also contain so.

#### UI

 1. Currently the Pluggable SCM SPA has the edit button enabled for all. It should be disable for SCM with `origin`=config-repo
2. The same goes for Materials SPA as well.

kritika-singh3:
Related Bug: Clicking on edit link of pluggable SCM materials takes to the pluggable SCM page where the message shows up `'scm_name' SCM has not been set up.`.

This happens sometimes (not always, randomly) due to the fact that the pluggable SCMs does not contain the SCM defined in the config-repo.

maheshp:
Verified on `GoCD Version: 20.9.0 (12301-4e16e35a91cc4698dd2281f5856d00211d0d7dc8).` Edit/delete button for SCM's from config_repo is disabled with an appropriate message shown on hover.

PR #8601, opened 2020-09-28
Solve issue around missing SCMs coming from config repo

kritika-singh3:
Issue: https://github.com/gocd/gocd/issues/8546#issuecomment-691819884

Description:
During conversion of configs from config repo to one stored by GoCD, we check if the said scm is already present or not. If former, we set the same as material config and ignore the one defined.
The search for presence was on all the configs including the ones present in the partials.
 - On the first parse, the SCM would not be found - hence the same would be added to the partial. On save, this will get integrated with all the scms.
 - However on the next parse it would be found - hence the same will not be added to the partial. On save, this will get integrated with the ones defined in the scm, deleting the older partial - effectively removing the scm

Fix: if the scm is already present - check the origin. If the origin is the current config repo - add it to the partial


marques-work (approved):
LGTM

PR #8603, opened 2020-09-29
Update SCM API and related SPAs

kritika-singh3:
Issue: #8546 

Description: Since GoCD 19.2.0, the pluggable scm configuration can be defined in config repo. Hence,

 - added `origin` information in the SCM API (upgraded to v4)
 - disabled edit/clone/delete operation on Pluggable SCMs SPA when defined in config repo
 - Also disabled edit operation on materials spa for same.


api/api-base/src/main/java/com/thoughtworks/go/api/representers/configorigin/ConfigRepoOriginRepresenter.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.thoughtworks.go.apiv9.admin.shared.representers.configorigin;
+package com.thoughtworks.go.api.representers.configorigin;
```
maheshp, line 16: I don't think we should be sharing representers between API versions. It's ok for each version to have it's own set of shared representers.

kritika-singh3: IMO, there are certain entities for which we have a fixed representation for all APIs. E.g. environment variable or plugin configuration property. Describing origin is also one such thing.
Hence, moving the same to api-base is a better option.

maheshp: I understand they have a same representation hence they are in a common place. We keep bumping up versions often and also delete the older versions post deprecation. I think getting into a mode of using representers across multiple versions would lead to confusion and cleaning up the representers would be hard as well.

kritika-singh3: Done.

api/api-scms-v4/src/main/java/com/thoughtworks/go/apiv4/scms/SCMControllerV4.java
```diff
@@ -0,0 +1,231 @@
...
+        haltIfEntityWithSameNameExists(scmFromRequest);
+        haltIfEntityWithSameIDExists(scmFromRequest);
+
+        HttpLocalizedOperationResult result = new HttpLocalizedOperationResult();
+        pluggableScmService.createPluggableScmMaterial(currentUsername(), scmFromRequest, result);
+
+        return handleCreateOrUpdateResponse(request, response, scmFromRequest, result);
+    }
+
+    public String update(Request request, Response response) {
```
maheshp, line 137: How are we handling updates to SCM's from ConfigRepo?

api/api-scms-v4/src/main/java/com/thoughtworks/go/apiv4/scms/SCMControllerV4.java
```diff
@@ -0,0 +1,231 @@
...
+            throw haltBecauseEtagDoesNotMatch(getEntityType().getEntityNameLowerCase(), existingSCM.getId());
+        }
+
+        final HttpLocalizedOperationResult operationResult = new HttpLocalizedOperationResult();
+        pluggableScmService.updatePluggableScmMaterial(currentUsername(), scmFromRequest, operationResult, getIfMatch(request));
+
+        return handleCreateOrUpdateResponse(request, response, scmFromRequest, operationResult);
+    }
+
+    public String destroy(Request request, Response response) throws IOException {
```
maheshp, line 157: Same as above, how is delete to SCM's from ConfigRepo handled?

api/api-scms-v4/src/main/java/com/thoughtworks/go/apiv4/scms/SCMControllerV4.java
kritika-singh3: Previously it sends 404 saying that the said scm id was not found.
Updated it to return 422 with the message saying it is defined in the config repo

maheshp (approved):

PR #8609, opened 2020-10-04
Move java dependency declarations to separate file and format for Dependabot

### LINKED ISSUE #8232: Move gradle dependency version information to `dependencies.gradle` so that dependabot can actually inspect our deps
Like many Java projects, we typically store our dependency versions in a common location so that we can share dependency versions between submodules.

Dependabot isn't that smart though.

According to dependabot/feedback/issues/345 we'd need to store this info in a file that contains the word `dependencies` in order for dependabot to parse our dependencies. Otherwise, we aren't going to reap the benefits of dependabot (i.e., the onslaught of PRs) for our Java libraries.

marques-work:
Fixes #8232

Most of our dependencies have been declared in the following format:

```groovy
implementation group: 'mygroup', name: 'mylib', version: project.versions.mylib
```

Currently, Dependabot is not able to parse most of our java dependencies because it isn't able to dereference variables or interpolate strings. This is because Dependabot's `gradle` engine does not actually evaluate Gradle scripts; it uses a simple regex to extract information.

So, this PR creates a `dependencies.gradle` file to hold all the dependency declarations. Dependabot will examine any gradle script with the work `dependencies` in it, in addition to `build.gradle`.

The formatting in this file is particular. Instead of using a groovy map to declare dependencies, we use the maven-like string declarations, e.g. `groupId:artifiactId:versionString`. Dependabot can more reliably extract this info as compared to the map format. Using a map (i.e., `[group: 'foo', name: 'bar', version: '1.0.0']`) might fail because the regex does not match the necessary square brackets needed by groovy's map declaration when used out of context from a gradle dependency configuration statement.

In other words, dependabot can match this:

```groovy
implementation group: 'foo', name: 'bar', version: '1.0.0
```

And this:

```groovy
implementation(group: 'foo', name: 'bar', version: '1.0.0) {
  // some closure body
}
```
**But NOT this**:

```groovy
project.ext.libraries = [
  foo: [group: 'foo', name: 'bar', version: '1.0.0],
  ...
]
```

So it _must_ be in this form instead:

```groovy
project.ext.libraries = [
  foo: 'foo:bar:1.0.0',
  ...
]
```

Version strings must be plaintext string literals instead of interpolated variables because of the reasons mentioned before. Related libraries that share versions with other dependencies are defined in a separate section in the file. See inline comments for explanations.

## Note to reviewers

This PR is built atop of #8557, so it only makes sense to review this after the other PR has been merged.

### DIFF
--- a/agent-bootstrapper/build.gradle
+++ b/agent-bootstrapper/build.gradle
@@ -20,19 +20,19 @@ description = 'GoCD Agent Bootstrapper'
 
 dependencies {
   implementation project(':agent-common')
-  implementation group: 'commons-codec', name: 'commons-codec', version: project.versions.commonsCodec
+  implementation project.deps.commonsCodec
 
-  testRuntimeOnly group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: project.versions.bouncyCastle
-  testRuntimeOnly group: 'org.bouncycastle', name: 'bcpkix-jdk15on', version: project.versions.bouncyCastle
+  testRuntimeOnly project.deps.bouncyCastle
+  testRuntimeOnly project.deps.bouncyCastlePkix
   testImplementation project(':test:test-utils')
-  testImplementation group: 'org.mockito', name: 'mockito-core', version: project.versions.mockito
-  testImplementation group: 'org.hamcrest', name: 'hamcrest-library', version: project.versions.hamcrest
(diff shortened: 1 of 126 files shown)


marques-work:
@maheshp any chance of merging this PR in the coming release? :)

Issue #8542, opened 2020-09-11
Changes to global Plugin Material Config and Package Config does not refelect on pipeline materials


maheshp:
##### Issue Type

- Bug Report

##### Summary

Plugin Materials and Package Materials are defined globally and referred in a pipeline. The in-memory PipelineConfig object has a copy of these global materials. Any updates to the configuration of the global materials does not reflect in the PipelineConfig materials. Hence, incorrect values are passed to plugin as part of material update.

**Current Workaround:**   A full config update of the config xml ensures pipeline references are updated.

kritika-singh3:
Tested on `GoCD Version: 20.8.0 (12202-5224cab793f1fc77ed51523d667087b4c29ad7f4)`

Plugins Used: [Maven Poller](https://github.com/1and1/go-maven-poller/releases/download/go-maven-poller-1.5.1/go-maven-poller-plugin.jar)

#### Package material
Steps:

1. Define a package repository using the maven poller with url: `https://jcenter.bintray.com`.
2. Define a package for the repo defined above with values:
     - Group Id: `com.oneandone`
     - Artifact Id: `snmpman`
     - Packaging: `jar`
3. Create a pipeline with the package as the material
4. Wait for MDU to go through and pipeline to be triggered
5. Verify that the Changes dropdown on the dashboard and materials spa modification contains: `"TRACKBACK_URL":"https://oosnmp.net/dist/release"`
6. Update the package with the following values:
     - Group Id: `cd.go.plugin`
     - Artifact Id: `go-plugin-api`
7. Wait for MDU and pipeline to be triggered
8. Verify that the changes on dashboard and modifications on materials spa contains: `"TRACKBACK_URL":"http://gocd.org"`
9. Do some config change not related to pipeline config. E.g. create a new pluggable SCM or a secret config

Expected Results
The pipeline should not be triggered

Actual Results
The pipeline gets triggered with the value which corresponds to the original package defined i.e. `snmpman` one. The changes for the latest instance as well as the modifications on the materials spa will now contain `"TRACKBACK_URL":"https://oosnmp.net/dist/release"`.

The polling for the material will now contain the values from the originally configured package configuration
Moreover the manual trigger of the pipeline will also use the modifications w.r.t. the original package configuration.

kritika-singh3:
Tested on `GoCD Version: 20.8.0 (12202-5224cab793f1fc77ed51523d667087b4c29ad7f4)`

[...]

PR #8610, opened 2020-10-05
Fix scm and package update

maheshp:
Issue: #8542 

Description: Update pipeline config on changes to to SCM, Package repo or package.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/UpdatePackageConfigCommand.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/UpdatePackageConfigCommand.java
@@ -16,6 +16,7 @@
 package com.thoughtworks.go.config.update;
 
 import com.thoughtworks.go.config.CruiseConfig;
+import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.exceptions.EntityType;
 import com.thoughtworks.go.domain.packagerepository.PackageDefinition;
 import com.thoughtworks.go.domain.packagerepository.PackageRepositories;
@@ -26,6 +27,8 @@ import com.thoughtworks.go.server.service.GoConfigService;
 import com.thoughtworks.go.server.service.materials.PackageDefinitionService;
 import com.thoughtworks.go.server.service.result.HttpLocalizedOperationResult;
 
+import java.util.List;
+
 public class UpdatePackageConfigCommand extends PackageConfigCommand {
     private final GoConfigService goConfigService;
     private final String oldPackageId;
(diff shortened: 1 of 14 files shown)


server/src/main/java/com/thoughtworks/go/config/update/UpdatePackageConfigCommand.java
```diff
@@ -68,6 +75,17 @@ public boolean canContinue(CruiseConfig cruiseConfig) {
         return isAuthorized() && isRepositoryPresent(cruiseConfig) && isIdSame() && isRequestFresh();
     }
 
+    private void updatePackageConfigurationOnAssociatedPipelines(CruiseConfig modifiedConfig) {
+        List<PipelineConfig> pipelinesWithPackage = modifiedConfig.pipelinesAssociatedWithPackage(newPackage);
+        pipelinesWithPackage.forEach(pipelineConfig -> {
+            pipelineConfig.packageMaterialConfigs().forEach(packageMaterialConfig -> {
+                if (packageMaterialConfig.getPackageId().equals(newPackage.getId())) {
```
kritika-singh3, line 82: The comparison check has been performed in the `modifiedConfig.pipelinesAssociatedWithPackage` method. Do we require it again.

server/src/main/java/com/thoughtworks/go/config/update/UpdatePackageRepositoryCommand.java
```diff
@@ -50,16 +53,29 @@ public void update(CruiseConfig modifiedConfig) {
...
     @Override
     public boolean canContinue(CruiseConfig cruiseConfig) {
         return super.canContinue(cruiseConfig) && isIdSame() && isRequestFresh();
     }
 
+    private void updatePackageRepositoryConfigurationOnAssociatedPipelines(CruiseConfig modifiedConfig) {
+        List<PipelineConfig> pipelinesWithPackageRepo = modifiedConfig.pipelinesAssociatedWithPackageRepository(newRepo);
+        pipelinesWithPackageRepo.forEach(pipelineConfig -> {
+            pipelineConfig.packageMaterialConfigs().forEach(packageMaterialConfig -> {
+                if (packageMaterialConfig.getPackageDefinition().getRepository().getId().equals(newRepo.getId())) {
```
kritika-singh3, line 69: Same here: the check has performed in `modifiedConfig.pipelinesAssociatedWithPackageRepository`

server/src/main/java/com/thoughtworks/go/config/update/UpdatePackageConfigCommand.java
maheshp: I was thinking of a scenario where a PipleineConfig has more than one package material. We would need this check in those cases right?

kritika-singh3: Yes right - we would need this check then.

kritika-singh3 (approved):
LGTM. Neither of the 2 scenarios mentioned in #8542 came up while testing.

kritika-singh3:
Verified on `GoCD Version: 20.9.0 (12264-30bf28b07277798dfc36761dad864e170cee5336)`

PR #8612, opened 2020-10-05
Fix the order of the yes and no button as well as the style

kritika-singh3:
Description:
The confirm delete on the preferences page had the major button as `No`. Updated the styles

Before:
Screen Shot 2020-10-05 at 3 33 23 PM

Now:
Screen Shot 2020-10-05 at 4 12 21 PM


maheshp:
@kritika-singh3 the button colour `red` is not consistent with other pages, check Pipeline Config page it is purple. Should we change this to purple as well?

kritika-singh3:
@maheshp  Only on the Pipeline config page it is purple. Rest all the places (Config Repo, Pluggable SCM, Server Configuration) it is red.
Should we update the one on pipeline config to red to make it consistent?

maheshp:
Guess you are right, yeah I think we should change it on the PipelineConfig page.

PR #8623, opened 2020-10-07
Fix the check connection button text on pluggable scm modal

kritika-singh3:
Description:
Reset the check connection button text when the check connection calls completes to the original value


maheshp:
Verified on `GoCD Version: 20.9.0 (12292-25ed9fb08c2f5de81a8c4ddacc9c77bba732109f)`

PR #8628, opened 2020-10-09
Do not run database migration for server in standby

maheshp:
* After moving to liquibase, as part of database upgrade
  liquibase tries to update the 'databasechangeloglock' table to
  take a lock. While using Business Continuity, this update fails on
  the secondary server running in a standby mode which is connected
  to a database which is configured to be in standby mode as well.
* This commit ensures a db migration is not run on a server in standby.
  When using business continuity, the database replication should take
  care of syncing migrations from primary to secondary db.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/database/Database.java
+++ b/server/src/main/java/com/thoughtworks/go/server/database/Database.java
@@ -20,6 +20,7 @@ import com.thoughtworks.go.security.GoCipher;
 import com.thoughtworks.go.server.database.migration.DatabaseMigrator;
 import com.thoughtworks.go.server.database.migration.DbDeploySchemaVerifier;
 import com.thoughtworks.go.util.SystemEnvironment;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/thoughtworks/go/server/database/Database.java
```diff
@@ -31,15 +32,23 @@
...
 
     private final ConnectionManager connectionManager;
     private SystemEnvironment systemEnvironment;
+    private DatabaseMigrator databaseMigrator;
 
     @Autowired
     public Database(SystemEnvironment systemEnvironment) {
-        this.connectionManager = new ConnectionManager(System.getProperties(), systemEnvironment.configDir(), decryptionFunction());
+        this(systemEnvironment, new ConnectionManager(System.getProperties(), systemEnvironment.configDir(), decryptionFunction()),
+                new DatabaseMigrator());
```
GaneshSPatil, line 45: intentional indent?

GaneshSPatil (approved):
LGTM

server/src/main/java/com/thoughtworks/go/server/database/Database.java
maheshp: Yes

kritika-singh3:
Verified via local `DevelopmentServer` on [master](https://github.com/gocd/gocd/commit/1f529832ad3d2ab6b88dfc9688d3a8446161ed9e)

Issue #6641, opened 2019-07-18
Add Elastic Profile Id at Pipeline and Stage level


kmugrage:
##### Issue Type

- Feature enhancement

##### Summary

Much (most?) of the time jobs in the same pipeline and pipeline stage use the same elastic agent profile. It would be great to be able to specify this at a higher level if appropriate much like environment variables.

PR #8632, opened 2020-10-09
Add support for elastic profile id at pipeline and stage config

kritika-singh3:
Issue: #6641 

Description:
 - Added a config migration to support specifying elastic profile id in pipeline config and stage config
 - Updated the instance factory to check for the elastic profile id in:
     - job config. If null then,
     - stage config. If null then,
     - pipeline config

#### TODO

 - [x] Update pipeline config API to v12
 - [x] Update template config API to v8
 - [x] Update the pipeline config and template config edit page
 - [ ] Add deprecation header to pipeline config v11 and template v7 API
 - [ ] Update the api docs as well as the user docs


olivierodo:
@kritika-singh3 Thank you for the this pull request, Did you have some information about this feature ?

kritika-singh3:
Hello @olivierodo What specific information are you looking for?
P.S. this PR is not in WIP. It has been moved to backlog for now.

olivierodo:
Hello @kritika-singh3 ,
Related to #8632 I was wondering if we will be able to specify the elastic profile id at the stage or the pipeline level.
That would help a lot to get rid of some redundancy.
I'm using YAML aliases to reuse some part of different pipeline but having the elastic profile id at the job level makes the reusability very challenging.

kritika-singh3:
Yes it is exactly what this PR was targeting. But this feature hasn't been prioritised yet. :)

Issue #8643, opened 2020-10-12
Elastic Agent Plugin always migrates the old plugin settings


GaneshSPatil:
##### Issue Type

- Bug Report

##### Summary

GoCD added support for elastic agents extension v5 to allow GoCD server to connect to multiple clusters to bring up elastic agents. To do so, GoCD moved away from a singular config entity plugin settings to a new concept called cluster profiles, which allowed users defining multiple cluster configurations.

As part of this new extension, all the older plugin settings were migrated to the cluster profile. Though, even after migration, the plugin settings were not deleted from the GoCD's database.

So, if a user deletes the newly migrated cluster profile after its use, GoCD will migrate the plugin settings again to the cluster profile.

###### Basic environment details

* Go Version: `20.8.0`

##### Steps to Reproduce

1. Start GoCD Server v20.8.0
2. Install docker elastic agent plugin (**docker-elastic-agents-2.1.0-174.jar**)
3. Configure plugin settings. 
4. Now, upgrade the docker elastic agent plugin to version **docker-elastic-agents-3.1.0-248.jar**
5. Verify a new cluster profile is created with the older plugin settings.
6. Now delete the newly migrated cluster profile.
7. Restart GoCD Server

##### Expected Results
The plugin settings are again migrated as a cluster profile

##### Actual Results
The plugin settings should be migrated only once to the cluster profile. Upon a successful migration of the plugin settings to cluster profile, GoCD should delete the plugin settings

maheshp:
Verified on `GoCD Version: 20.9.0 (12292-25ed9fb08c2f5de81a8c4ddacc9c77bba732109f)`

PR #8644, opened 2020-10-12
Delete plugin settings on a successful migration of the plugin settings to cluster profile. (#8643)

GaneshSPatil:
Issue: #8643

Description:

* Do not delete the plugin settings when the migrate-config call fails.
* Do not delete the plugin settings when the elastic agent plugin implements
  elastic agent extension v4, which relies on plugin settings.
* Delete the plugin settings when the plugin successfully migrates the
  plugin settings to cluster profile and the plugin implements elastic
  agent extension v5 or above, which does not rely on plugin settings.

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/dao/PluginSqlMapDao.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dao/PluginSqlMapDao.java
@@ -113,4 +113,23 @@ public class PluginSqlMapDao extends HibernateDaoSupport implements PluginDao {
             }
         });
     }
+
+    @Override
+    public void deletePluginIfExists(String pluginId) {
+        String cacheKey = cacheKeyForPluginSettings(pluginId);
+        Plugin plugin = this.findPlugin(pluginId);
+
+        if (plugin instanceof NullPlugin) {
+            return;
+        }
+
+        synchronized (cacheKey) {
+            transactionTemplate.execute((TransactionCallback) transactionStatus -> sessionFactory.getCurrentSession()
+                    .createQuery(String.format("delete from %s where pluginId=:pluginId", Plugin.class.getSimpleName()))
+                    .setParameter("pluginId", pluginId)
+                    .executeUpdate());
+
+            goCache.remove(cacheKey);
+        }
+    }
 }
--- a/server/src/main/java/com/thoughtworks/go/server/dao/PluginSqlMapDao.java
+++ b/server/src/main/java/com/thoughtworks/go/server/dao/PluginSqlMapDao.java
@@ -118,16 +118,20 @@ public class PluginSqlMapDao extends HibernateDaoSupport implements PluginDao {
     public void deletePluginIfExists(String pluginId) {
         String cacheKey = cacheKeyForPluginSettings(pluginId);
         Plugin plugin = this.findPlugin(pluginId);
-
         if (plugin instanceof NullPlugin) {
             return;
         }
 
         synchronized (cacheKey) {
-            transactionTemplate.execute((TransactionCallback) transactionStatus -> sessionFactory.getCurrentSession()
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentInformationMigratorImpl.java
```diff
@@ -68,12 +73,29 @@ private boolean migrate(GoPluginDescriptor pluginDescriptor) {
...
         Plugin plugin = pluginSqlMapDao.findPlugin(pluginId);
         String pluginConfiguration = plugin.getConfiguration();
         HashMap<String, String> pluginSettings = (pluginConfiguration == null) ? new HashMap<>() : JsonHelper.fromJson(pluginConfiguration, HashMap.class);
         ReplaceElasticAgentInformationCommand command = new ReplaceElasticAgentInformationCommand(clusterProfilesService, elasticProfileService, elasticAgentExtension, pluginDescriptor, pluginSettings);
 
-        return update(command, pluginDescriptor);
+        boolean updated = update(command, pluginDescriptor);
+
+        if (updated) {
+            LOG.debug("Done Migrating elastic agent information for {} plugin", pluginId);
```
maheshp, line 85: `Done migrating elastic agent configurations for plugin with id: '{}'`

server/src/main/java/com/thoughtworks/go/server/service/ElasticAgentInformationMigratorImpl.java
```diff
@@ -68,12 +73,29 @@ private boolean migrate(GoPluginDescriptor pluginDescriptor) {
...
+
+            // all the plugins implementing elastic agent extension v5, does not use plugin settings and hence delete them after successful migration
+            if (!pluginManager.resolveExtensionVersion(pluginId, ELASTIC_AGENT_EXTENSION, ElasticAgentExtension.SUPPORTED_VERSIONS).equals(ElasticAgentExtensionV4.VERSION)) {
+                LOG.debug("Deleting stale plugin settings for {} plugin, after successful plugin config migration", pluginId);
+                pluginSqlMapDao.deletePluginIfExists(pluginId);
+            } else {
+                LOG.debug("Skip deleting plugin settings for {} plugin, as the plugin implements elastic agent v4 extension which uses plugin settings", pluginId);
+            }
+        } else {
+            LOG.debug("Failed migrating elastic agent information for {} plugin", pluginId);
```
maheshp, line 95: Probably this should be logged as error in the `update ` method catch block.

server/src/main/java/com/thoughtworks/go/server/dao/PluginSqlMapDao.java
```diff
@@ -113,4 +113,23 @@ protected void doInTransactionWithoutResult(TransactionStatus status) {
...
+        String cacheKey = cacheKeyForPluginSettings(pluginId);
+        Plugin plugin = this.findPlugin(pluginId);
+
+        if (plugin instanceof NullPlugin) {
+            return;
+        }
+
+        synchronized (cacheKey) {
+            transactionTemplate.execute((TransactionCallback) transactionStatus -> sessionFactory.getCurrentSession()
+                    .createQuery(String.format("delete from %s where pluginId=:pluginId", Plugin.class.getSimpleName()))
```
maheshp, line 128: Couldn't we just do `sessionFactory.getCurrentSession().delete(plugin);`

GaneshSPatil: `transactionTemplate.execute` method expects a return value and `sessionFactory.getCurrentSession().delete` returns a void. This requires adding a new execute method which supports void return type.

maheshp (approved):

Issue #8595, opened 2020-09-24
Issues with Stage Overview


GaneshSPatil:
##### Issue Type

- Bug Report

#### On the VSM page

- [x] Misaligned stage overview (positioned above than expcted) when a pipeline has multiple run instances for the same upstream.

- [ ] (Selecting older stage run) Selecting the older stage instance counter does not change the status of the stage instance from the pipeline-box to the current stage status from overview.

kritika-singh3:
On the VSM page (related to #8576)

 - [ ] misaligned name with `>>` icon
 - [ ] Stage duration is rendered as `Invalid date` for a non-building stage
 - [ ] Poller is turned off by default - but if you change the stage counter - it turned on. The stage overview starts updating the job statuses. But the stage overview status bar remains in the original status.

<img width="539" alt="Screen Shot 2020-09-25 at 10 42 03 AM" src="https://user-images.githubusercontent.com/41165891/94229016-f0f4db00-ff1b-11ea-8161-637d558cf3bd.png">

kritika-singh3:
Tested on `GoCD Version: 20.9.0 (12236-550797553265f0d8918d65356b5a59f846794cdc)` on the dashboard and pipeline activity page. (Related to #8570)

1. Selecting an older failed instance from the stage overview changes the status of stage in the dashboard as well. This behaviour is not present on pipeline activity
![1](https://user-images.githubusercontent.com/41165891/94227954-5eebd300-ff19-11ea-8eb6-1055ed9a03be.gif)

2. Rerunning a stage/job temporary removes the stage counter selection. Present on pipeline activity page as well. 
![2](https://user-images.githubusercontent.com/41165891/94227957-6317f080-ff19-11ea-8fa3-f428dfe5cc27.gif)

3. The stage duration changes after time
![3](https://user-images.githubusercontent.com/41165891/94227958-63b08700-ff19-11ea-80e6-fb52348af70b.gif)

GaneshSPatil:
@maheshp @kritika-singh3 -

1. When an older stage instance is selected on the stage overview, the status of the stage-overview will change, but the status of the dashboard stage will always show the latest stage instance. This behavior is now consistent with other pages as well - dashboard, pipeline activity and vsm.

2. On the VSM page, when one of the pipeline has multiple instance runs, the stage overview will be misaligned, when there are more than pipeline instances for the same pipeline. 
   For upto 3 pipeline instances, the stage overview will be positioned appropriately.
[...]

PR #8645, opened 2020-10-12
Fix stage overview issues (#8595)

GaneshSPatil:
Issue: #8595

Description:

* Refresh stage overview status when the status of the pipeline
  activity stage changes.

* Always render a new stage overview when the stage bar is clicked.


kritika-singh3:
Verified on `GoCD Version: 20.9.0 (12278-00bbeec8dff839559809c65de50e0c141fa0c03e)`

Issue #8170, opened 2020-05-22
Support for webhook integration with Pluggable SCM materials


chadlwilson:
##### Issue Type
- Feature enhancement

##### Summary

Currently, GoCD supports incoming webhooks from a number of sources (GitHub, GitLab, Bitbucket Server + Cloud) all relating to Git materials, however my understanding is that it uses some heuristics to match the material and basically hard-codes the material search to filter for the inbuilt Git material.

There are a few SCM plugins in the ecosystem that also support Git, however I don't think there is a way for the webhooks to select these Git materials (and I'm not convinced there would be a way without an enhancement [to the `scm` API](https://plugin-api.gocd.org/current/scm/) that would allow 

- the plugin to inform the server upfront that it handles git materials
- the server to inform the plugin of the webhook invocation metadata
- **and**, the plugin to inform the server whether it was a possible match

In particular, the [Git Path Material plugin](https://github.com/TWChennai/gocd-git-path-material-plugin/issues/27) could benefit from this; as polling dozens of materials is expensive, and as noted [here](https://github.com/TWChennai/gocd-git-path-material-plugin/issues/27), can end up breaching API request limits.

##### Any other info
Even if such support existed, I'm not familiar enough with how material triggers work to know whether this would work with the git-path-material-plugin, since the plugin basically filters the `git log` when answering the `latest-revision` request.

Opening this for discussion/ideas about whether there might be a way to make this work.

chadlwilson:
Seems to relate to #5328

arvindsv:
@chadlwilson [The webhook notification API endpoints](https://api.gocd.org/current/#webhook) are analogous to the [notification API endpoints](https://api.gocd.org/current/#notify-materials), with the only difference being the matching you mentioned.

[...]

PR #8646, opened 2020-10-12
Update webhook endpoints to support `SCM_NAME`

kritika-singh3:
Issue: #8170 

Description:
 - added support for query param `SCM_NAME`. The param can be specified multiple times.
 - the names if received will be mapped w.r.t. schedulable pluggable scm materials - if matched, a MDU will be triggered fr the same.
 - the none-matches will be silently ignored
 - fixed the broken bitbucket authentication

#### TODO
 - [x] Test for Github
 - [x] Test for Gitlab
 - [x] Test for Bitbucket cloud
 - [ ] Update docs.

### DIFF
--- a/api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/request/mixins/bitbucketcloud/BitBucketCloudAuth.java
+++ b/api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/request/mixins/bitbucketcloud/BitBucketCloudAuth.java
@@ -63,7 +63,7 @@ public interface BitBucketCloudAuth extends HasAuth {
             }
 
             String encodedCredentials = new String(Base64.getDecoder().decode(credentials), UTF_8);
-            return encodedCredentials.contains(":") ? split(encodedCredentials, ":", 2)[1] : encodedCredentials;
+            return encodedCredentials.contains(":") ? split(encodedCredentials, ":", 2)[0] : encodedCredentials;
         }
         return null;
     }
--- a/api/api-webhook-v1/src/test/groovy/com/thoughtworks/go/apiv1/webhook/request/push/BitBucketCloudPushRequestTest.java
+++ b/api/api-webhook-v1/src/test/groovy/com/thoughtworks/go/apiv1/webhook/request/push/BitBucketCloudPushRequestTest.java
@@ -161,11 +161,10 @@ class BitBucketCloudPushRequestTest {
         void shouldBeValidWhenTokenMatches() {
             String body = "{ \"repository\": { \"scm\": \"git\" } }";
 
-            Request request = newRequestWithAuthorizationHeader("repo:push", "021757dde54540ff083dcf680688c4c9676e5c44", body);
+            Request request = newRequestWithAuthorizationHeader("repo:push", "021757dde54540ff083dcf680688c4c9676e5c44:x", body);
 
             assertThatCode(() -> new BitBucketCloudPushRequest(request).validate("021757dde54540ff083dcf680688c4c9676e5c44"))
                     .doesNotThrowAnyException();
-
         }
     }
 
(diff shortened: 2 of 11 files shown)


api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/request/mixins/bitbucketcloud/BitBucketCloudAuth.java
```diff
@@ -63,7 +63,7 @@ private String getTokenFromHeader(Request request) {
             }
 
             String encodedCredentials = new String(Base64.getDecoder().decode(credentials), UTF_8);
-            return encodedCredentials.contains(":") ? split(encodedCredentials, ":", 2)[1] : encodedCredentials;
+            return encodedCredentials.contains(":") ? split(encodedCredentials, ":", 2)[0] : encodedCredentials;
```
marques-work, line 66: I guess nobody noticed this was broken :).

marques-work, line 66: Wouldn't this take the username and not the personal token?

api/api-webhook-v1/src/test/groovy/com/thoughtworks/go/apiv1/webhook/request/push/BitBucketCloudPushRequestTest.java
```diff
@@ -93,6 +96,26 @@ void shouldGuessTheWebhookUrls(String body) {
...
+        when(request.queryMap()).thenReturn(map);
+        when(map.hasKey(KEY_SCM_NAME)).thenReturn(false);
+        assertThat(bitBucketCloudPushRequest.getScmNames()).isEmpty();
+
+        when(request.queryMap()).thenReturn(map);
+        when(map.hasKey(KEY_SCM_NAME)).thenReturn(true);
+
+        QueryParamsMap value = mock(QueryParamsMap.class);
+        when(map.get(KEY_SCM_NAME)).thenReturn(value);
+        when(value.values()).thenReturn(singletonList("scm1").toArray(new String[]{}));
```
marques-work, line 115: This isn't wrong, but isn't this more simply `new String[]{"scm1"}`?

api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/request/mixins/bitbucketcloud/BitBucketCloudAuth.java
kritika-singh3: In our API docs, we have mentioned to put the webhook secret as username and some random value as password: https://api.gocd.org/current/#hosting-on-bitbucket
I think Bitbucket does not recognize personal token in url - it only takes `username:password` format.

api/api-webhook-v1/src/test/groovy/com/thoughtworks/go/apiv1/webhook/request/push/BitBucketCloudPushRequestTest.java
kritika-singh3: Yes u r right - will update it

marques-work (commented):
I think it looks good, though 2 things:

1. ~Does bitbucket.com really send the token as the first segment of the userinfo?~ you answered this above
2. I hate to nit-pick, but _I think, per our conventions_ adding `SCM_NAME` support warrants a new API version since it's new behavior and requires documentation updates. /cc @maheshp @GaneshSPatil correct me if I'm wrong

api/api-webhook-v1/src/main/java/com/thoughtworks/go/apiv1/webhook/request/mixins/bitbucketcloud/BitBucketCloudAuth.java
marques-work: @kritika-singh3 ah ok, disregard my first point in my review summary then

kritika-singh3:
> I hate to nit-pick, but I think, per our conventions adding SCM_NAME support warrants a new API version since it's new behavior and requires documentation updates

Yes, u r right. It is a new behaviour but I thought our webhooks endpoint are not versioned as we don't verify the version info in the request headers. (Same as feeds api)

marques-work:
> Yes, u r right. It is a new behaviour but I thought our webhooks endpoint are not versioned as we don't verify the version info in the request headers. (Same as feeds api)

Hmm. Fair enough -- we can't expect 3rd party services to allow custom headers for our benefit. I think it's fine then. Is this ready to merge?

kritika-singh3:
Yes. The master is failing, hence the tests are failing. Otherwise, this is ready for merging.

marques-work:
@kritika-singh3 oops, I think I broke it - I'll fix it and I'll merge this as well.

PR #8653, opened 2020-10-14
Fix for partials ConfigConvertor race condition #000

maheshp:
* ConfigConvertor while generating SCMs for a partial used
  a list of SCMs which was defined as an instance variable.
  Since multiple partials could be converted at the same time
  multiple threads could be updating the same SCMs object which
  led to incorrect SCMs for a partial. This in-turn would lead
  to the merge failure with validation failing for duplicate SCMs.
* This commit fixes the race condition by not relying on the SCMs
  object defined as an instance variable.


GaneshSPatil (approved):
LGTM

PR #8654, opened 2020-10-14
Update secret config SPA 

kritika-singh3:
Description:
 - Introduce an internal secret configs API which contains the secret configs configured along with any auto-complete suggestions required
 - Updated the secret config SPA to use the same.


maheshp:
Verified on `GoCD Version: 20.9.0 (12301-4e16e35a91cc4698dd2281f5856d00211d0d7dc8).`

PR #8655, opened 2020-10-14
Do not purge the config repo defined structures cache on polled fetch

marques-work:
This fixes a UI flicker when a config repo widget is expanded. Previously, the
defined pipeline structure graph would quickly blank out, then reappear each time
GoCD fetched config repo data on interval. This was caused by mithril redrawing
an empty pipeline structures data cache (resulting in an empty graph with a loading
message) followed only milliseconds later by a redraw when data was full again on
fetch response. On most connections, even those that are relatively slow, this looks
like a UI "glitch" because it happens so quickly.

Even though the caches were designed to not clear their contents on `invalidate()`
precisely to address this issue, it was happening anyway because a new cache object
was recreated on each fetch. This commit fixes this by persisting object caches over
intervals for reuse. Data is still updated when it changes, but no longer flickers
because the widget continues to display stale data while waiting for the API response
until fresh data overwrites it upon response completion.

Here's what I'm talking about (slowed down network so it can be seen on GIF, but imagine this happens much more quickly so it looks like a flash)

config-repo-flicker

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/base/cache.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/base/cache.ts
@@ -88,8 +88,12 @@ export abstract class AbstractObjCache<T> implements ObjectCache<T> {
   }
 
   invalidate() {
-    this.empty();
+    this.markStale();
     this.primed = false;
+  }
+
+  protected markStale() {
+    this.empty();
   }
 
   protected empty() {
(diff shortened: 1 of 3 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/models/base/cache.ts
```diff
@@ -88,10 +88,14 @@ export abstract class AbstractObjCache<T> implements ObjectCache<T> {
   }
 
   invalidate() {
-    this.empty();
+    this.markStale();
     this.primed = false;
   }
 
+  protected markStale() {
```
marques-work, line 95: Adding another abstraction only because I like the idea of still allowing an explicit `empty()` if ever needed. The `data` field is marked `private` and would be otherwise inaccessible in typescript.

GaneshSPatil (approved):
LGTM

PR #8656, opened 2020-10-15
Show appropriate stage duration on the stage overview across mithril redraws.

GaneshSPatil:
Issue: #8595

Description:

* Compute stage compute time based on the oldest scheduled job from
  the stage. In case of reruning a single job, jobs from the stage will
  have different scheduled times. In such cases, pick the job with
  the oldest scheduled time.


kritika-singh3:
Verified on `GoCD Version: 20.9.0 (12294-720a2fe626694cade5fccf55ef39d79dca6156f7)`

PR #8667, opened 2020-10-20
Allow `ConfigRepoConfig` to set `autoUpdate` to `false`

marques-work:
Part of #8023

Provides a UI to configure `autoUpdate` and removes validation enforcing it to be enabled.

As a byproduct, it also fixes the API etags to be built from the serialized representations rather than the weak `hashCode()` for proper caching.


config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
```diff
@@ -65,7 +67,6 @@ public void validate(ValidationContext validationContext) {
         this.validateRepoIsSet();
         this.validateMaterial(validationContext);
         if (isValidMaterial()) {
-            this.validateAutoUpdateEnabled();
             this.validateAutoUpdateState(validationContext);
```
maheshp, line 70: The implementation for `validateAutoUpdateState` has to change to check if `auto_update` value for ConfigRepo material matches with regular material.

maheshp (commented):
Will take a look again once the changes for the MDU are pushed.

server/src/main/java/com/thoughtworks/go/server/service/EntityHashingService.java
```diff
@@ -397,6 +405,26 @@ public String hashForEntity(ArtifactConfig artifactConfig) {
         return getFromCache(cacheKey, () -> String.valueOf(Objects.hash(artifactConfig)));
     }
 
+    public String hashForEntity(PartialConfig entity) {
```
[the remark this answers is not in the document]
marques-work: I'll remove it. I think I had intended to use it from a controller, but realized it is really the `PartialConfigParseResult` that I needed to hash. Thanks for catching it.

config/config-api/src/main/java/com/thoughtworks/go/config/remote/ConfigRepoConfig.java
marques-work: Is this because if there is an existing material that has the same config, `auto_update` must be the same?

marques-work: Changed this to test if autoUpdate is the same across all matching materials

maheshp: > Is this because if there is an existing material that has the same config, auto_update must be the same?
Yes

config/config-api/src/test/java/com/thoughtworks/go/config/CruiseConfigTestBase.java
```diff
@@ -811,24 +814,27 @@ public void getAllUniquePostCommitSchedulableMaterials_shouldReturnMaterialsWith
...
         cruiseConfig.setConfigRepos(reposConfig);
 
 
         Set<MaterialConfig> materials = cruiseConfig.getAllUniquePostCommitSchedulableMaterials();
         assertThat(materials.size(), is(2));
-        assertThat(materials, hasItem(gitNonAutoMaterial));
-        assertThat(materials, hasItem(configRepoMaterial));
+        assertThat(materials, hasItem(gitMaterialManual));
+        assertThat(materials, hasItem(configRepoMaterialManual));
+        assertThat(materials, not(hasItem(configRepoMaterialAutoUpdate)));
```
marques-work, line 837: Changing this test to reflect that only materials and configrepos with `auto_update = false` will be considered.

maheshp:
Verified on `GoCD Version: 20.9.0 (12300-de4958ffa90153ecfc28519bf5347423b1b28ca2).`

Note: ConfigRepo behavior when the material `auto_update` is turned off.
1. On creation of a ConfigRepo the corresponding material is updated and the the ConfigRepo is parsed and merged.
2. On update of the ConfigRepo the corresponding material is updated and the the ConfigRepo is parsed and merged. This material update is required since there could be changes to material as well.
3. Otherwise, ConfigRepo materials are not polled. The materials can be updated by,
   - Refreshing the config repo
   - Manual trigger of pipeline which shares the same material with ConfigRepo
   - Through [Webhook](https://api.gocd.org/current/#webhook) or [Notify Materials](https://api.gocd.org/current/#notify-materials) API's

PR #8668, opened 2020-10-20
ClonerFactory to fix JDK 15 `deepClone()` issues

marques-work:
We need a special `Cloner` instance to handle `ImmutableCollections.{List12,Set12}` instances because their peculiar implementation results in an incorrect output from `deepClone()`.


maheshp (approved):
LGTM.

Config saves uses an instance of `GoConfigCloner` which extends the Cloner, we need to take care of it as well.

config/config-server/src/main/java/com/thoughtworks/go/config/GoConfigCloner.java
```diff
@@ -36,5 +37,6 @@ public GoConfigCloner() {
                 PipelineNameToConfigMap.class,
                 CachedPluggableArtifactConfigs.class,
                 CachedFetchPluggableArtifactTasks.class);
+        ClonerFactory.applyFixes(this);
```
marques-work, line 40: @maheshp This should register the custom `FastCloner`s for `GoConfigCloner`. Tests below to validate.

maheshp:
Verified on `GoCD Version: 20.9.0 (12318-b67833a26e7273f83019facdf22dfcfd290cdf4a).`

PR #8712, opened 2020-11-02
Fix comparison spa failure for view users

kritika-singh3:
Description:
When a user with `view` access to a pipeline visits the comparison page - it fails. This is due to get pipeline config API call failing as the same is accessible to only admins/group admins.
Fixed it by not setting the error state when the said call fails.


maheshp:
Verified on `GoCD Version: 20.9.0 (12331-99a8b0050394855cabc1dab750e10832c3c4d405)`

PR #8688, opened 2020-10-26
Unify the `autoUpdate` and `ignoreForScheduling` UI widgets

marques-work:
- Extract the autoUpdate radio button toggle as a common component
- Make a variant for ignoreForScheduling under the same file
- Refactor typings for ConfigRepo, Material, Scm, and Package to be more
  convenient (and more sensible?)
- Minor refactoring on RadioField component to allow for better value
  type-checking and clearer implementation

PR #8713, opened 2020-11-02
Fix issues related to autoUpdate UI

kritika-singh3:
Issue: https://github.com/gocd/gocd/pull/8688#issuecomment-718422581

Description:
 - Showcase error text w.r.t. material auto update toggle on the widget

  Screen Shot 2020-11-02 at 10 33 42 AM

 - The foundation hax styles were applying which caused an additional 1em margin bottom to the radio input.
Rectified that by explicitly setting margin to zero.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/material_auto_update_toggle.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/material_auto_update_toggle.tsx
@@ -32,8 +32,8 @@ interface Attrs {
 
 export class MaterialAutoUpdateToggle extends MithrilViewComponent<Attrs> {
   view(vnode: m.Vnode<Attrs>) {
-    const { noun, toggle, errors } = withDefaults(vnode.attrs);
-    const autoUpdate = this.radioState(toggle);
+    const {noun, toggle, errors} = withDefaults(vnode.attrs);
+    const autoUpdate             = this.radioState(toggle);
 
     return <RadioField<AutoUpdateRadioOptions>
       label={`${_.capitalize(noun)} polling behavior:`}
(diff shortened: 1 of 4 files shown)


server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipelines/material_auto_update_toggle.tsx
```diff
@@ -32,19 +32,22 @@ interface Attrs {
...
+        label={`${_.capitalize(noun)} polling behavior:`}
+        property={autoUpdate}
+        readonly={!!vnode.attrs.disabled}
+        errorText={errors?.errorsForDisplay("autoUpdate")}
+        dataTestId="material-auto-update"
+        possibleValues={[
+          {label: `Regularly fetch updates to this ${noun}`, value: "auto"},
+          {label: `Fetch updates to this ${noun} only on webhook or manual trigger`, value: "manual"},
+        ]}/>,
+      <ErrorText errorText={errors?.errorsForDisplay("autoUpdate")} errorId={`material-auto-update-error`}/>
```
marques-work, line 49: Hmm, looking at `RadioField`, it doesn't seem to render `errorText` on its own. Should we just add `<ErrorText/>` to `RadioField` like all the other input components? Seems like it may have been an oversight in its initial impl?

marques-work: Unless there is a good case for not showing errors normally, but I'd find that to be odd.

kritika-singh3: Hmm sounds right to me. I will make the change.

marques-work (approved):
LGTM

maheshp:
Verified on `GoCD Version: 20.9.0 (12331-99a8b0050394855cabc1dab750e10832c3c4d405).`

PR #8714, opened 2020-11-02
Do not build alpine 3.9 as it has reached EOL

kritika-singh3:
Description:
Stop building Alpine 3.9 based images as the same has reached EOL
Ref: https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases


maheshp:
Alpine 3.9 image is no longer built - https://build.gocd.org/go/tab/build/detail/installers/2487/docker/2/docker-agent#

PR #8715, opened 2020-11-02
Fixed the incorrect console log link

kritika-singh3:
When the stage counter is changed via the counter selector, the console log should point to the one selected rather than the latest one


maheshp:
Verified on `GoCD Version: 20.9.0 (12328-c3593d377f9f90efe46916c01bf239a7270b5d1f).`

PR #8766, opened 2020-11-13
Remove command repo

maheshp:
Command lookups is no longer supported, hence cleaning up code around CommandRepository


kritika-singh3:
Looks good. Found the command repository reference in a few more places. Please remove them as well:
 - [System Environment](https://github.com/maheshp/gocd/blob/remove_command_repo/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java#L83)
 - [service.rb](https://github.com/maheshp/gocd/blob/remove_command_repo/server/src/main/webapp/WEB-INF/rails/lib/services.rb#L52)
 - gradle task to [bumpCommandRepoVersion](https://github.com/maheshp/gocd/blob/remove_command_repo/release/build.gradle#L44)

Issue #8767, opened 2020-11-15
Get /go/api/agents is slow when there are too many secure variables in gocd environments


shakirshakiel:
##### Issue Type

- Bug Report

##### Summary

When there are too many secure variables in gocd environments, the GET agents API (/go/api/agents) starts to slow down. 

##### Environment

Output of GoCD support api - https://gist.github.com/shakirshakiel/ce5ad9338a8decdc20515229d4ab986d

###### Basic environment details

* Go Version: `20.7.0 (12097-f4f86ca8d433edaf9235ed92790c11e19d7be4ed)`
* JAVA Version: `13.0.2 `
* OS: `Linux 3.10.0-957.1.3.el7.x86_64`

##### Steps to Reproduce

1. Deploy a gocd-server (20.7.0) with /etc/go/cipher.aes value (e36a7c58b0fe642fd8dc8aa2d32c9465).
2. Deploy 1 gocd-agent and enable it.
3. Add this repo (https://github.com/shakirshakiel/gocd-environments) as a config-repo. Wait till config-repo is updated.
4. Enable all environments to the gocd-agent.
5. Hit /go/api/agents.

##### Expected Results

/go/api/agents should respond within < 3 seconds

##### Actual Results

/go/api/agents takes approximately 10 - 15 seconds to respond

##### Possible Fix

1. MergeEnvironmentConfig class, hashCode method is the slowest after some debugging. This calls getVariables method twice. In getVariables() method, allVariables is getting constructed by comparing with all other variables in the list. Basically O(n2). This method can be called once by storing in a variable.
2. In EnvironmentVariableConfig class equals method, decrypting both values can be last. This will add some performance benefit because in an environment no 2 variables will have same name.

https://github.com/shakirshakiel/gocd/commit/7ea11727b9983027f6472378ad6d4ded5961bebd

PR #8768, opened 2020-11-15
Fix performance issues in GET agents API

shakirshakiel:
Issue: #8767

Description:

1. MergeEnvironmentConfig class, hashCode method is the slowest after some debugging. This calls getVariables method twice. In getVariables() method, allVariables is getting constructed by comparing with all other variables in the list. Basically O(n2). This method can be called once by storing in a variable.

2. In EnvironmentVariableConfig class equals method, decrypting both values can be last. This will add some performance benefit because in an environment no 2 variables will have same name.

If there are 100 agents with each agent associated to 10 environments and each environment has 100 secure variables, then for each GET /go/api/agents - AES decryption will happen 100 * 10 * 100 * 100 * 2 * 2 = 4,00,00,000 times

Breakdown:

100 - Number of agents
10 - Number of environments
100 * 100 * 2 - Worst case scenario of getVariables method calling decrypt
2 - Getvariables getting called twice in MergeEnvironment

After this commit, decrypt function will be called 100 * 10 = 1000 times because in an environment no 2 secure variables will have the same name.

However decrypt will be called when computing hashcode.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/EnvironmentVariableConfig.java
@@ -116,13 +116,13 @@ public class EnvironmentVariableConfig implements Serializable, Validatable, Par
         if (isSecure != that.isSecure) {
             return false;
         }
-        if (!goCipher.passwordEquals(encryptedValue, that.encryptedValue)) {
-            return false;
-        }
         if (name != null ? !name.equals(that.name) : that.name != null) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


maheshp:
@shakirshakiel thanks for this!

maheshp:
Verified on `GoCD Version: 20.10.0 (12351-ddd81da42cd59eff9f78f4f486ef3d148b3af432)`. Tried to verify the performance using the environments defined in https://github.com/shakirshakiel/gocd-environments. On GoCD 20.9.0 the agents API took around ~30secs to respond which is reduced to less the 200ms post the fix.

PR #8775, opened 2020-11-19
Make the ignore_for_scheduling field readonly on pipeline config edit spa

kritika-singh3:
Issue: 
when the pipeline is from a config repo - all the fields are readonly except the `ignore_for_scheduling` on dependency materials

Description : use `attrs.readonly` attribute instead of `attrs.disabled`.


maheshp:
Verified on `GoCD Version: 20.10.0 (12351-ddd81da42cd59eff9f78f4f486ef3d148b3af432).`

PR #8776, opened 2020-11-19
If the window is in the background, even if the tab is active, reduce the polling interval by a factor of 3.

marques-work:
Just a minor tweak to be even stingier with connections when GoCD isn't in focus.


kritika-singh3 (approved):

kritika-singh3:
Verified on `GoCD Version: 20.10.0 (12354-b9d98c27050ca3d67d568270530c256d4c5f3750)`

PR #8777, opened 2020-11-19
Set default connection pool settings for h2 datasource #000

maheshp:
* DefaultH2DataSource maxIdle and maxTotal defaults to 32 connections.
  This is same as the defaults we had prior to adding support for
  multiple databases.

* H2 db url no longer supports MVCC flag, hence removing it.
  Check H2 1.4.200 changelog - https://github.com/h2database/h2database/releases/tag/version-1.4.200


marques-work (approved):
👍🏼

Issue #8774, opened 2020-11-18
bug - agent resource tags are case-sensitive when referenced from a job


jasonduff:
##### Issue Type
Bug Report

##### Summary
When a job has an agent resource tag that does NOT exactly match (e.g. case is different), it fails to find that agent.  This is an issue in 20.7.0.  It was not an issue in 18.7, which is what we upgraded from.

##### Environment
Currently running 20.7.0 on a windows server 2016 VM

###### Basic environment details
GO Version: 20.7.0
JVM: 13.0.2
OS Info: Windows Server 2016 10.0
Chrome browser
3500+ pipelines

kritika-singh3:
Fixed via #8778 and verified on `GoCD Version: 20.10.0 (12354-b9d98c27050ca3d67d568270530c256d4c5f3750)`

PR #8778, opened 2020-11-20
Agent, job resource matching is case-insensitive #8774

maheshp:
Issue: #8774


kritika-singh3 (approved):
LGTM. One suggestion: Should we also update the internal resources API (used on Agents SPA and pipeline config edit SPA) to return a list of unique resources in lower case?
E.g. if the agent has `TEsting` and job has `Testing`, the API should return only `testing`

maheshp:
> Should we also update the internal resources API (used on Agents SPA and pipeline config edit SPA) to return a list of unique resources in lower case?

I think we should just go with the users input and not change it.

kritika-singh3:
Verified on `GoCD Version: 20.10.0 (12354-b9d98c27050ca3d67d568270530c256d4c5f3750)`

PR #8782, opened 2020-11-23
Update pipeline activity page

kritika-singh3:
Description: 
 - make the manual gate same width as auto gate - improves alignment of stages
 - remove unused cancel stage instance method
 - use a confirmation dialog before triggering the pipeline


maheshp:
Verified on `GoCD Version: 20.10.0 (12356-b16b6bc3f7ad3bb39e465ff2168d1cc66c95b4d5).`

PR #8880, opened 2020-12-12
Agent-Server comms using JSON

maheshp:
Description: **WIP Do not merge**


agent/src/main/java/com/thoughtworks/go/agent/NewRemote.java
```diff
@@ -0,0 +1,235 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void reportCompleted(AgentRuntimeInfo agentRuntimeInfo, JobIdentifier jobIdentifier, JobResult result) {
+        Map<String, Object> request = Map.of(
+                "agent_runtime_info", agentRuntimeInfo,
+                "job_identifier", jobIdentifier,
+                "job_result", result);
```
arvindsv, line 135: I guess you don't need these maps, since you have those small objects?

agent/src/main/java/com/thoughtworks/go/agent/NewRemote.java
```diff
@@ -0,0 +1,235 @@
...
+
+        try {
+            this.httpClientHttpInvokerRequestExecutor.doPost("report_completed",
+                    gson.toJson(new ReportCompleteStatusRequest(agentRuntimeInfo, jobIdentifier, result)));
+        } catch (IOException e) {
```
arvindsv, line 140: I think the exception needs to be serialised and sent back from the server. It might not always be an IOException. Right?

common/src/main/java/com/thoughtworks/go/remote/adapter/RuntimeTypeAdapterFactory.java
```diff
@@ -0,0 +1,277 @@
+/*
+ * Copyright 2020 ThoughtWorks, Inc.
```
arvindsv, line 2: I think you should get the full original file, _along_ with its license and then add this license on top of this, with a comment explaining where it came from.

agent/src/main/java/com/thoughtworks/go/agent/NewRemote.java
maheshp: Yeah, it needs to be cleaned up. Will do.

common/src/main/java/com/thoughtworks/go/remote/adapter/RuntimeTypeAdapterFactory.java
maheshp: Will do.

agent/src/main/java/com/thoughtworks/go/agent/NewRemote.java
maheshp: Since this was a quick spike, haven't really thought about error handling. You are right tho, there could be other exceptions, need to take care of it.

maheshp:
This was a spike, actual changes done as part of https://github.com/gocd/gocd/pull/8898

PR #8887, opened 2020-12-15
Agent Client Impl

marques-work:
### Static agents

Configure `wrapper-properties.conf` to set the following property: `-Dgocd.agent.remoting.legacy=true`

### Elastic agents:

Set the `AGENT_STARTUP_ARGS` to include `-Dgocd.agent.remoting.legacy=true`

~This is a temporary PR while #8880 is being cleaned up. Just want to be able to run builds against it, then will merge into #8880 when ready and close this one.~

This PR will be cherry-picked into and superseded by #8880, after which will be closed.

### DIFF
--- a/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
@@ -92,6 +92,8 @@ public class AgentHTTPClientController extends AgentController {
         this.artifactExtension = artifactExtension;
         this.pluginRequestProcessorRegistry = pluginRequestProcessorRegistry;
         this.pluginJarLocationMonitor = pluginJarLocationMonitor;
+
+        LOG.info("Configured remoting type: {}", remote().getClass().getSimpleName());
     }
 
     @Override
--- a/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
@@ -29,8 +29,8 @@ import com.thoughtworks.go.plugin.infra.PluginRequestProcessorRegistry;
 import com.thoughtworks.go.plugin.infra.monitor.PluginJarLocationMonitor;
 import com.thoughtworks.go.publishers.GoArtifactsManipulator;
 import com.thoughtworks.go.remote.AgentIdentifier;
-import com.thoughtworks.go.remote.BuildRepositoryRemote;
 import com.thoughtworks.go.remote.AgentInstruction;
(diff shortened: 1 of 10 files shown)


arvindsv:
@marques-work Might need to be ready to use `-Dgocd.agent.extra.properties=` as in https://github.com/gocd/gocd/pull/5693 and might want to test that it works.

I mean: The reason we would want to use `-Dgocd.agent.remoting.legacy=true` is if the GSON serialization goes very wrong. In that case, if it's a large installation, it might be a bit too much to expect this property to be set on every agent. It will be easier to set this on the server side.

Of course, the challenge with this _specific_ change is that you need agent-server comms to be working for this to happen. :) But, the `gocd.agent.extra.properties` property works at the bootstrapper or launcher time and should not be affected by this change.

cc: @maheshp

maheshp:
@arvindsv, Marques indeed was looking for an option to configure this from the server side, my bad, I couldn't recollect about the option to set `gocd.agent.extra.properties`.

We should definitely use this over configuring multiple agents, elastic profiles etc. Thanks for pointing it out.

arvindsv:
Cool. I don't think setting that property restarts the agent, which might be needed in this case. Anyway testing it will let you know.

maheshp:
Ok, I verified setting the property `gocd.agent.extra.properties` on the server. We should be able to control the client used on the agent side by setting this property on the server.

```
wrapper.java.additional.100=-Dgocd.agent.extra.properties=gocd.agent.remoting.legacy=false
```

Updating the property would not restart the agent. An agent restart is not required in this case since the the remote client is decided before making a server request.

Thanks again @arvindsv!

arvindsv:
> Updating the property would not restart the agent. An agent restart is not required in this case since the the remote client is decided before making a server request.

But, if an agent is running and using the new JSON way, and something goes wrong and we decide that it needs to switch to using the old way, setting this property won't make the agent (without restart) start using the old way, right?

maheshp:
A restart is not required to switch between the old and new clients on the agent side. Check [this](https://github.com/gocd/gocd/pull/8887/files#diff-2ef9d8e91e4cefc0d4aaf31808772178950506496a8f79a6d1eff93b97af905eR147) and [this](https://github.com/gocd/gocd/pull/8887/files#diff-2ef9d8e91e4cefc0d4aaf31808772178950506496a8f79a6d1eff93b97af905eR140), every time before the agent makes a request to the server, the client to be used is decided based on the system property.

arvindsv:
Ah, good. Cool, thank you.

maheshp:
These changes are cherry-picked and merged in https://github.com/gocd/gocd/pull/8898.

Closing this PR.

PR #8895, opened 2020-12-16
Prevent gson serialization from serializing cipher-related instances

marques-work:
Introduces a rejecting type adapter that throws exceptions when trying to un/marshal during agent JSON remoting:

- `GoCipher`
- `AESEncrypter`
- `AESCipherProvider`
- `DESEncrypter`
- `DESCipherProvider`

### DIFF
--- a/common/src/main/java/com/thoughtworks/go/remote/Serialization.java
+++ b/common/src/main/java/com/thoughtworks/go/remote/Serialization.java
@@ -42,10 +42,13 @@ import com.thoughtworks.go.domain.materials.scm.PluggableSCMMaterialInstance;
 import com.thoughtworks.go.domain.materials.svn.SvnMaterialInstance;
 import com.thoughtworks.go.remote.adapter.RuntimeTypeAdapterFactory;
 import com.thoughtworks.go.remote.work.*;
+import com.thoughtworks.go.security.*;
 import com.thoughtworks.go.server.service.AgentRuntimeInfo;
 import com.thoughtworks.go.server.service.ElasticAgentRuntimeInfo;
 
 import java.lang.reflect.Type;
+
+import static java.lang.String.format;
 
 public class Serialization {
     private static class SingletonHolder {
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/thoughtworks/go/remote/Serialization.java
```diff
@@ -169,10 +194,9 @@ private JsonArray configurations(ArtifactStore store) {
         @Override
         public ArtifactStore deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
             Configuration configuration = new Configuration();
-            json.getAsJsonObject().get("configuration").getAsJsonArray().forEach(el -> {
-                configuration.add(new ConfigurationProperty(new ConfigurationKey(el.getAsJsonObject().get("key").getAsString()),
-                        new ConfigurationValue(el.getAsJsonObject().get("value").getAsString())));
-            });
+            json.getAsJsonObject().get("configuration").getAsJsonArray().forEach(el ->
+                    configuration.add(new ConfigurationProperty(new ConfigurationKey(el.getAsJsonObject().get("key").getAsString()),
+                            new ConfigurationValue(el.getAsJsonObject().get("value").getAsString()))));
```
marques-work, line 199: Just IDEA suggesting terser syntax by converting to expression-lambda.

common/src/test/java/com/thoughtworks/go/remote/SerializationTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+
+@SuppressWarnings("deprecation")
+class SerializationTest {
+    @Test
+    void rejectsSerializationOfGoCipher() {
+        final IllegalArgumentException e = assertThrows(IllegalArgumentException.class, () ->
+                Serialization.instance().toJson(new GoCipher(mock(Encrypter.class))));
```
maheshp, line 36: What happens when we try to serialize `ArtifactStore`, the store needs to be serialized but deep down it does have a cipher.

marques-work: It depends on the `ArtifactStore` type adapter. If it attempts to do a `toJson()` on it rather than computing a set of decrypted secrets, it will blow up.

marques-work: It looks like it would work, unless I'm missing something.

For serialization, we never descend that far into configuration properties. We do this instead:

https://github.com/marques-work/gocd/blob/ff49b4ff9ce81f2379e2382ad8db362ab7e331bc/common/src/main/java/com/thoughtworks/go/remote/Serialization.java#L184-L190

And for deserialization, we do the reverse. The `Gson` instance never encounters a `GoCipher` instance:

https://github.com/marques-work/gocd/blob/ff49b4ff9ce81f2379e2382ad8db362ab7e331bc/common/src/main/java/com/thoughtworks/go/remote/Serialization.java#L197-L199

So we shouldn't blow up here. We can add a test though.

marques-work:
Merging this now

PR #8898, opened 2020-12-17
Agent server comms over json

maheshp:
Introduced new internal agents API


api/api-internal-agent-v1/src/main/java/com/thoughtworks/go/apiv1/internalagent/InternalAgentControllerV1.java
```diff
@@ -0,0 +1,146 @@
...
+        boolean isIgnored = buildRepositoryMessageProducer.isIgnored(isIgnoredRequest.getAgentRuntimeInfo(), isIgnoredRequest.getJobIdentifier());
+
+        return valueOf(isIgnored);
+    }
+
+    public String getCookie(Request request, Response response) {
+        GetCookieRequest getCookieRequest = GetCookieRequestRepresenter.fromJSON(request.body());
+        ensureAgentIsMakingARequestForItself(getCookieRequest, request);
+
+        return buildRepositoryMessageProducer.getCookie(getCookieRequest.getAgentRuntimeInfo());
```
marques-work, line 124: @maheshp If I’m not mistaken, this returns raw text? For consistency, do you think we should quote it to be valid JSON?

maheshp: You are right, I will change it.

Issue #8916, opened 2020-12-26
arguments of Custom Command with spaces will failed to run


dongyuwei:
##### Issue Type
- Feature enhancement

##### Summary
When we define a Custom Command, if the argument starts with empty spaces or ends with spaces, the command will failed to run. Such as:
```bash
ls
 -al ##keep an eye the space surrounds the -al
```

or
```bash
bash
-c  ##keep an eye the space here
echo ok;
```

##### Environment
- go-agent-20.10.0
- go-server-20.10.0    20.10.0 (12356-b16b6bc3f7ad3bb39e465ff2168d1cc66c95b4d5)

macOS Big Sur or Windows 10

###### Basic environment details
* Go Version: GoCD Version: 20.10.0 (12356-b16b6bc3f7ad3bb39e465ff2168d1cc66c95b4d5).
* JAVA Version: 15.0.1
* OS: Mac OS X 10.16

##### Steps to Reproduce
1. Add a 'Custom Command' task,
2. input `ls` in command input field, and input `  -al  ` in the arguments input field.
3. run the task.

##### Expected Results
The command should passed without problem.

##### Actual Results
```
[go] Task: bash "-c " "echo ok;"took: 0.117sexited: 2
bash: - : invalid option
Usage:	bash [GNU long option] [option] ...
	bash [GNU long option] [option] script-file ...
GNU long options:
	--debug
	--debugger
	--dump-po-strings
	--dump-strings
	--help
	--init-file
	--login
	--noediting
	--noprofile
	--norc
	--posix
	--protected
	--rcfile
	--restricted
	--verbose
	--version
	--wordexp
Shell options:
	-irsD or -c command or -O shopt_option		(invocation only)
	-abefhkmnptuvxBCHP or -o option
[go] Task status: failed, took: 0.117s, exited: 2
```

Or

```
[go] Task: ls "-al "took: 0.112sexited: 1
ls: illegal option --  
usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]
[go] Task status: failed, took: 0.112s, exited: 1
```

Or

```
ls:     -al: No such file or directory
[go] Task status: failed, took: 0.109s, exited: 1
```

##### Possible Fix
trim the spaces of the arguments.

##### Code snippets/Screenshots
![image](https://user-images.githubusercontent.com/112451/103149096-5a95b700-47a1-11eb-826f-fa803510e95d.png)

[...]

PR #8917, opened 2020-12-26
trim spaces of arguments in Custom Command

dongyuwei:
Issue: #8916

Description: trim spaces of arguments when define and use Custom Command


marques-work:
@dongyuwei I can understand why you want this, but it may not be the best choice. There are times when one wants a space as an argument, and this would make it harder to do that.

For example, when using `cut`, we often want to pass a space as the argument to the `-d` flag. It might be a better idea to just highlight spaces in the UI so the user can decide whether they are intentional or not.

marques-work:
For reference, this is the editor used in create pipeline:

image

PR #8929, opened 2021-01-01
Disable RMI endpoint by default but allow toggling

marques-work:
Even though agents now use the JSON remoting API, the RMI endpoint on the server should be disabled unless the agents are configured to use the legacy RMI comms.

This endpoint can be enabled by including `gocd.agent.remoting.legacy=true` in the `-Dgocd.agent.extra.properties` property for the server process, which is the same method used to enable RMI on the agents side.


chadlwilson:
Hey folks - do you think it would be safe now to remove the RMI remoting entirely? Are you aware of any issues or reports of people needing to turn RMI remoting back on?

The only thing I am really aware of that might relate to this is #9641 which might relate to the change in comms approaches.

I noticed that the regression tests seem to have been running with `gocd.agent.remoting.legacy=true` set which is, err, not ideal, especially as this area is a bit sensitive for the Java 16/17 upgrade (regarding serialization to objects or JSON). Just switched them - hopefully it's fine.

PR #8959, opened 2021-01-13
Update to jgit 5.10.0.202012080955-r

marques-work:
Issue: #

Description:


marques-work:
Seems to cause issues loading plugins during regression tests.

See:

image

We need to figure this out before considering this upgrade.

marques-work:
Superceded by #8963

PR #9465, opened 2021-06-28
Fix expectation for JavaEWAH version post JGit `5.11.1` bump

chadlwilson:
In #9380, JGit was bumped, which brought in a newer transitive dependency version of JavaEWAH (`1.16` -> `1.17`).

```
|    +--- org.eclipse.jgit:org.eclipse.jgit:5.11.1.202105131744-r
|    |    +--- com.googlecode.javaewah:JavaEWAH:1.1.7
```

Current build is failing due to this: https://build.gocd.org/go/tab/build/detail/installers/2564/dist/1/dist


arvindsv:
Thank you! Merging. The `docker` build is failing, but it is unlikely to be due to this and it looks like this PR is needed for https://github.com/gocd/gocd/pull/9467 as well.

PR #9467, opened 2021-06-29
Remove dependency on legacy library junit-ext

chadlwilson:
The functionality this provided has been replaced within JUnit 5 by `EnabledOnOs` and `DisabledOnOs`.

This avoids the need to vendor JUnit-Ext and makes it easier to further migrate things to JUnit 5 which which reduces the cognitive overhead of working with the codebase.


arvindsv:
This looks good. More readable too. Thank you again!

chadlwilson:
Thanks for taking a look @arvindsv !

Do you know how to get it so guests can view the pull request builds on the team GoCD? The permissions don't seem to allow it right now so couldn't review the failure.

Or alternatively whether there is a community-oriented process to get a read account to relevant pipeline groups linked to a user's Google or GitHub account?

arvindsv:
I'll need to think about this a bit more. Not sure if there is an easy way to do this, since the pipelines are created dynamically by https://github.com/gocd-contrib/gocd-groovy-dsl-config-plugin which can't control permissions. Maybe @marques-work has an idea?

Interestingly, there was a pipeline created for https://github.com/gocd/gocd/pull/9479 which was running and one of the jobs was stuck. I killed it and re-ran the job, but the pipeline itself is gone. The plugin mentioned above controls the creation and destruction of the pipeline. Marques might know about that too ...

chadlwilson:
I see.

Not sure how that plugin works, but would it be sufficient for the plugin to place the relevant pipelines in a pre-created pipeline group that has the necessary permissions already assigned?

Yeah for #9479 I noticed that from the perspective of the PR it had failed... eventually. After that I forced pushed and noticed that the build "check" had completely gone from the list which was weird.

arvindsv:
That's the challenge. The plugin creates a new _dynamic_ pipeline group, with a name based on the PR. For instance, the one for #9479 is called "gocd-pull_9479-Attempt_to_get_the_JUnit_5_migration_closer_to_completion". So, it's a little more challenging.

I've given you access manually to it. Can you see if you can view it on the dashboard? I do see it come and go sometimes. I wonder if it is because the config repo fails to refresh sometimes. As I write this comment, I see that the pipeline is available.

PR #9479, opened 2021-07-04
Attempt to get the JUnit 5 migration closer to completion

chadlwilson:
Systematically go through every test & try to migrate what I can to make these more workable.

- migrating types to JUnit 5
- changing Mockito to use the MockitoExtension and fix unnecessary stubbing where possible
- removed use of `@Rule` where an alternate solution is possible
- otherwise still using `@EnableRuleMigrationSupport` 
- migrate all `@Theory` usages to `@ParameterizedTest`
- generally still using Hamcrest matchers except where migration to AssertJ was necessary to avoid confusing `assertThat` clashes

Methodology
- no production code changed.
- no tests removed
- some unreachable assertions due to incorrect `ExpectedException` and `@Test(expect = blah.class)` usage were removed where they were failing. 

I appreciate that this PR is massive and can't be reviewed line-by-line however suggest this is probably the only way to get this done so the tests are more consistent to work with. Perhaps can ignore all the import changes and focus on the limited test code change.

I aim to follow up by finding a more consistent/easy replacement for `TemporaryFolder` and remove the remaining `Rule` usage.


marques-work:
> I aim to follow up by finding a more consistent/easy replacement for TemporaryFolder and remove the remaining Rule usage.

I always found these bits to be the most frustrating when chipping away at this task. Other than me being obsessive about removing `assertThat` and `expect`-style assertions :) as a personal peeve.

chadlwilson:
> > I aim to follow up by finding a more consistent/easy replacement for TemporaryFolder and remove the remaining Rule usage.
> 
> I always found these bits to be the most frustrating when chipping away at this task. Other than me being obsessive about removing `assertThat` and `expect`-style assertions :) as a personal peeve.

Do you mean to say you prefer the AssertJ-style fluent `assertThat` rather than the one from Hamcrest `MatcherAssert`? Or the other way around?

chadlwilson:
I'm pretty confident that this change doesn't make things worse, and I had a vague attempt to make a cleaner diff without the import and basic annotation cruft, but it is still pretty big.

`git diff --unified=1 fe03e9b79f28606dcd177518e442457b99cf7991..HEAD -I 'import|@RunWith|@ExtendWith|@Before|@After|initMocks|ExpectedException|@Rule|@Mock' --ignore-space-at-eol -b -w --ignore-blank-lines > junit5-patch.txt`

[junit5-patch.txt](https://github.com/gocd/gocd/files/6952447/junit5-patch.txt)

chadlwilson:
Hello @marques-work - do you have any concerns with me merging this? If you can clarify how you would prefer the assertion styles to go I can possibly follow up with some cleanup here, or correct anything I'm doing which is perhaps inconsistent.

arvindsv:
I think this should be merged. You've done a lot here and if anything else needs to be done, it can be done later. No point sitting on so much of a change, in my opinion.

Issue #9509, opened 2021-07-10
Unable to render old VSM page when dependent pipelines are deleted or renamed


shakirshakiel:
##### Issue Type

- Bug Report

##### Summary

When dependent pipeline nodes of a pipeline is renamed or deleted, we are unable to view the old VSM pages. 

###### Basic environment details

* Go Version: `21.2.0`
* JAVA Version: `15.0.2 `
* OS: `Linux 3.10.0-957.1.3.el7.x86_64

##### Steps to Reproduce

1.  Deploy gocd-server (21.2.0) with an agent
2.  Add config-repo with git material (https://github.com/shakirshakiel/gocd-environments.git) & branch stage1
3. Trigger build & deploy pipelines 
4. Verify whether VSM page of deploy pipeline works (https://gocd.example.com/go/pipelines/value_stream_map/deploy/1)
5. Modify config-repo to branch stage2. In this branch, build pipeline is renamed to test.
6. After the config-repo is refreshed, view the VSM of deploy pipeline (https://gocd.example.com/go/pipelines/value_stream_map/deploy/1)
7. You will get an error "Value Stream Map of pipeline 'deploy' cannot be rendered. Please check the server log for details"

##### Expected Results

The VSM page should be rendered

##### Actual Results

The VSM page shows an error - "Value Stream Map of pipeline 'deploy' cannot be rendered. Please check the server log for details". In the server log NullPointerException is thrown

##### Possible Fix

Handle NullPointerException in ValueStreamMapService#removeRevisionsBasedOnPermissionAndCurrentConfig method

##### Log snippets

```
java.lang.NullPointerException: Cannot invoke "com.thoughtworks.go.config.PipelineConfig.getTemplateName()" because the return value of "com.thoughtworks.go.server.service.GoConfigService.findPipelineByName(com.thoughtworks.go.config.CaseInsensitiveString)" is null
        at com.thoughtworks.go.server.service.ValueStreamMapService.removeRevisionsBasedOnPermissionAndCurrentConfig(ValueStreamMapService.java:207)
        at com.thoughtworks.go.server.service.ValueStreamMapService.buildValueStreamMap(ValueStreamMapService.java:114)
        at com.thoughtworks.go.server.service.ValueStreamMapService.getValueStreamMap(ValueStreamMapService.java:79)
```

chadlwilson:
Looks like this might have been broken since GoCD `20.9.0`. Resurrecting this and the associated PR, as had another report of it.

PR #9510, opened 2021-07-10
Handle NullPointerException in ValueStreamMapService

shakirshakiel:
Issue: #9509

Description:

With this, we should be able to view old VSM pages whose dependent pipelines are renamed or deleted. In VSM, deleted pipelines is shown as "Pipeline has been deleted"


chadlwilson:
Made minor tweaks to the PR; including ensuring the test code accurately tests this case & tested locally.

PR #9573, opened 2021-08-09
Remove (likely unintentional) use of Apache Poi locales in test

chadlwilson:
Poi is a transitive dependency of DB Unit; we shouldn't be using it directly; and the test fails under later DBUnit/POI versions

Fixes issue caused by merge of #9538 (see https://build.gocd.org/go/tab/build/detail/build-linux/5518/build-non-server/1/FastTests-runInstance-1)


chadlwilson:
Could you please help me merge this @arvindsv ? Don't think I have the permissions to do so. Made the mistake of merging some minor dependabot updates when I don't have ability to merge any required fixes myself due to the CLA check issues. Doh - my bad :-(

PR #9596, opened 2021-08-13
Improve compatibility for cloning of config repo via Docker entrypoint

### LINKED ISSUE #8878: gocd-server fail on start with CONFIG_GIT_REPO
I'm trying to start a gocd-server: v20.10.0 docker containder using a git settings repository. But when cloning the configuration repository, you are giving the following error:

```console
/usr/local/sbin/git-clone-config: line 10: [: too many arguments
```

chadlwilson:
Issue: likely fixes #8878

Description: Currently this feature can fail (maybe only on some bash versions, still working out a specific reprod), I believe because the output of `git status` is a string with multiple args which can break single-bracket tests (`/usr/local/sbin/git-clone-config: line 10: [: too many arguments`).

The test bracketing seems unnecessary here given we just want to check the exit code of a command.

### DIFF
--- a/buildSrc/src/main/resources/gocd-docker-server/git-clone-config
+++ b/buildSrc/src/main/resources/gocd-docker-server/git-clone-config
@@ -7,7 +7,7 @@ try() { echo "$ $@" 1>&2; "$@" || die "cannot $*"; }
 try cd /godata/config
 # somehow this script and install-gocd-plugins called 2 times
 # so check if git repo already exists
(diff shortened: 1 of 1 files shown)


arvindsv:
I agree. Your change seems like the _right_ way to do this. The output of git status isn't _that_ predictable that it can be used there, especially without extra quoting around the backticks.

chadlwilson:
Still not quite sure on the use case for this env var (never used it personally) and how exactly this can break things irretrievably based on my comment in https://github.com/gocd/gocd/issues/8878#issuecomment-899596504 - however I don't think this is going to make anything worse.

PR #9686, opened 2021-09-10
Fix unstable integration test ordering

chadlwilson:
Attempts to fix unstable integration tests by aligning the approach to setup and teardown the config files and related database configuration.

It seems when tests use/involve the fixtures for pipelines (`PipelineWithTwoStages` etc) which make changes to config in both database config file, they can leave the server configuration/context in a broken state which is not reset correctly for subsequent tests. In particular `PipelineScheduleServiceTest` fails repeatably if run after `JobAssignmentIntegrationTest` or after `AgentServiceIntegrationTest`, as do some other tests which expect jobs to be run (or not run) on all agents. These failures feel very random and cause a lot of broken pipeline runs.

In general
* prefer `configFileHelper.onSetup` to `initializeConfigFile` (the former already does the latter)
* avoid using `initializeConfigFile` in test teardowns. It's confusing when there is already an `onTearDown` which most tests use
* ordering of `dbHelper.onTearDown` and `configHelper.onTearDown` seems to be important in some cases (for some reason I don't fully understand). Align these (although it really would be better to have a way to enforce this).

### DIFF
--- a/server/src/test-fast/java/com/thoughtworks/go/server/service/StageServiceTest.java
+++ b/server/src/test-fast/java/com/thoughtworks/go/server/service/StageServiceTest.java
@@ -117,7 +117,7 @@ public class StageServiceTest {
 
     @AfterEach
     public void teardown() throws Exception {
-        configFileHelper.initializeConfigFile();
+        configFileHelper.onTearDown();
         SessionUtils.unsetCurrentUser();
     }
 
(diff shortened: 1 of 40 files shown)


chadlwilson:
Integration tests have run 5 separate times on PR pipeline without experiencing these issues. Once they failed due to an hsqldb lock/deadlock issue (a separate issue that I have not attempted to address here). In general this gives me reasonable confidence that these changes do not introduce any additional instability, so merging now and will observe over time.

PR #9735, opened 2021-10-03
Avoid unnecessary full re-installs of Yarn dependencies & unnecessary yarn task re-runs

chadlwilson:
Speeds up the local dev process by fixing the Gradle results caching of Yarn installs.

In #8557 the working directory was redeclared as type `@InputDirectory` to address some Gradle 7 compatibility issues with annotating properties correctly. The problem is that using this type causes Gradle to observe the contents of this directory and assess whether it has changed between runs in order to decide whether the task is stale and needs to be re-run. https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputDirectory.html

Since this folder has many other files other than `node_modules`, `package.json`, `yarn.lock` etc, it basically always considers the task `outOfDate` which currently causes the code as written to wipe `node_modules` and do an unnecessary full reinstall and re-link, which is **very** slow.

Changing these types back to regular input strings will cause it to only consider them out of date if the path itself changes, which reverts to the behaviour before #8557, while still being Gradle 7 compliant in terms of annotating inputs correctly.

### DIFF
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/YarnInstallTask.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/YarnInstallTask.groovy
@@ -17,7 +17,10 @@
 package com.thoughtworks.go.build
 
 import org.gradle.api.DefaultTask
-import org.gradle.api.tasks.*
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.incremental.IncrementalTaskInputs
 import org.gradle.internal.os.OperatingSystem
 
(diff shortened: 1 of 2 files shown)


arvindsv:
Wow! This is huge! Thank you.

chadlwilson:
@arvindsv You're welcome! Was irritating me, so did some source code archaeology and figured out how it was *supposed* to work originally when the tasks were written. Just an unintentional side-effect of some of the rather painful Gradle 7 compatibility prep work Marques did a lot of the heavy lifting on late last year.

The code is still pretty conservative in removing the entire `node_modules` entirely locally for any detected change in `package.json` or `yarn.lock` however, which I left in place. I generally haven't found a need to do that with Yarn or indeed modern npm in more recent years (on other projects), but the hardcore JS experts may differ in that opinion, so left it in place.

I have a list of minor irritations I have been working through gradually as a way of understanding things a bit better :-)

arvindsv:
Yes, I've noticed that but didn't know enough to go into it that deep. I've also had a lot of trouble with incremental builds, which would be fast but somehow the rails part (where it compiles assets) used to always re-do a lot of stuff, seemingly unnecessarily.

> I have a list of minor irritations I have been working through gradually as a way of understanding things a bit better :-)

❤️ :)

Issue #9469, opened 2021-07-02
Unable to start Go CD server with Java 16


sguptatw:
##### Issue Type
Bug Report
 
##### Summary

Following the instruction on [GOCD installation](https://docs.gocd.org/current/installation/install/server/osx.html). Unable to start Go CD server. Getting an UnsatisfiedDependencyExceptions exception

###### Basic environment details
```
Mac OS 10.14.6
GoCD Version: 21.2.0-12498
jvm 1             Java Version: 16
jvm 1    xMac OS X(10.14.6)
```

##### Steps to Reproduce

* On Mac OSX, install using the zip/osx installer.
* Start the server using the command `./bin/go-server console` 

##### Expected Results
The exception stack trace is huge, hence have attached to the log file. The first exception is `UnsatisfiedDependencyExceptions` 

##### Actual Results
Should be able to run the server without an exception

##### Log snippets
Please find the logs here [logs.zip](https://github.com/gocd/gocd/files/6751377/logs.zip)

arvindsv:
Can you try with a Java version between 13 and 15?

But, overall, yes, I agree that it should be able to run.

chadlwilson:
The root of the problem is the use of `setAccessible` in deep cloning logic, which is disallowed by default on Java 16.

```
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private transient java.util.NavigableMap java.util.TreeSet.m accessible: module java.base does not "opens java.util" to unnamed module @3f371cc0
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:357)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:177)
	at java.base/java.lang.reflect.Field.setAccessible(Field.java:171)
	at com.rits.cloning.Cloner.addAll(Cloner.java:563)
	at com.rits.cloning.Cloner.allFields(Cloner.java:577)
	at com.rits.cloning.Cloner.registerStaticFields(Cloner.java:209)
	at com.rits.cloning.Cloner.registerKnownConstants(Cloner.java:190)
	at com.rits.cloning.Cloner.init(Cloner.java:85)
	at com.rits.cloning.Cloner.<init>(Cloner.java:56)
[...]

PR #9818, opened 2021-10-30
Upgrade GoCD to run and build with Java 17

### LINKED ISSUE #9635: GOCD new version with build-in Java 15.0.3+
Issue Type
Summary
Let us know which GOCD version support the Java 15.0.3.

After install GOCD 21.2.0 , GOCD build-in Java version is 15.0.2.
To fix Java vulnerability with Java 15.0.2, we would like to know the GOCD new version.
Kindly advise us.

Basic environment details
Go Version: 21.2.0
JAVA Version: 15.0.2
OS: Linux 3.10.0-1160.36.2.el7.x86_64
Browser vendor and version : Google chrome 86.0

is it possible that we use a LTS version of java, like java 11?

chadlwilson:
Fixes #9635, #9469

* Upgrades to Java 17
   * Since Gradle 6 was not compatible with Java 16+ when initially working on this, in order to decouple, introduces use of Gradle Java Toolchains to compile and test using a different JVM to that Gradle runs with (this is now moot, since the Gradle 7 upgrade was completed, but preserving ability to run tests with target JVM and Gradle with a newer JVM is possibly desirable)
   * Adds bare minimum `--add-opens` to open up internal JVM modules for reflection, that seems to be required in a number of places, including use of the deep object `cloner` (but not limited to this) and Hibernate 3.6/javassist
   * Avoids use of non-ideal`source`/`target` compatibility, preferring the `release` flag which ensures code doesn't depend on APIs from earlier JVMs it shouldn't.
* Switches Java builds to Eclipse Adoptium Temurin JRE builds
  * Adds ability to use an Alpine-specific Adoptium JRE, but not currently used due to issues with the Tanuki Service Wrapper on musl libc

Task list
* [X] Get pre-requisite JVM on build agents
* [X] Get tests passing
* [x] A lot of cleanups/TODOs/FIXMEs
* [x] Validate the various distributions
  * [X] Windows
  * [X] MacOS
  * [X] Linux (.deb via Ubuntu)
* [X] Factor out the `--add-opens` args so the tests for `agent` run with different set of args than `server` since they seem to require different packages to be open
* [X] Much more deep assessment of correctness of server and agent with the existing `--add-opens`. The current list was derived through trial and error.
  * [X] Review all the dodgy `setAccessible` reflection stuff throughout the code
    * There's a lot, but most of it seems to be via walking configuration, and params resolver stuff. I hope that this is OK with the existing opens. 
  * [X] Check the `SvnCommand` and such exception handling which sets exception internals. Is this used on agents?
    * The `smudgeException` code is only used when checking modifications on the server, it appears. So should be fine. 
  * [X] brute force through some more features "of interest" on server locally, and some relevant plugins?
* [ ] Consider changing compatibility back to Java 11 (previous LTS) rather than Java 13 which is not an LTS release.
* [X] Sanity check that 22.1.0 server can talk to 21.4.0 agent bootstrapper after server upgrade
* [X] Update developer docs to reflect need to `--add-opens` when running certain tests inside IDE
* [x] Rebase/squash

Before merge
* [x] Drop the final commit (that hacks the Gradle JDK version).
  * [x] Take a backup of build.gocd.org :-)
  * [x] Update the pipeline definition on `build.gocd.org` to `BUILD_ON_JDK=17`

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 # Configuration for https://github.com/asdf-vm/asdf as an alternative to jabba, rvm, nvm etc
 java temurin-17.0.2+8
 ruby jruby-9.3.3.0
-nodejs 16.13.1
+nodejs 16.13.2
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 # Configuration for https://github.com/asdf-vm/asdf as an alternative to jabba, rvm, nvm etc
-java adoptopenjdk-15.0.2+7
+java temurin-17.0.1+12
 ruby jruby-9.3.3.0
 nodejs 16.13.1
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 # Configuration for https://github.com/asdf-vm/asdf as an alternative to jabba, rvm, nvm etc
-java temurin-17.0.1+12
+java temurin-17.0.2+8
 ruby jruby-9.3.3.0
 nodejs 16.13.1
--- a/build.gradle
+++ b/build.gradle
@@ -166,10 +166,10 @@ project.ext.compilerOptions = [
 
 project.ext.packaging = [
   adoptOpenjdk: [
-    featureVersion: 15,
+    featureVersion: 17,
     interimVersion: 0, // set to `null` for the first release
-    updateVersion : 2, // set to `null` for the first release
-    buildVersion  : 7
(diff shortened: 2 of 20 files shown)


chadlwilson:
All is clear now; failing build is due to remove the BUILD_ON_JDK=17 being removed for merge (forgot to pause the PR build)

chadlwilson:
Workaround guide at https://github.com/gocd/gocd/issues/10265

PR #9824, opened 2021-10-31
Add specific fast cloners for `Date`s and `Timestamps`

chadlwilson:
Currently there are both `Date` and `Timestamp` objects deep within the structures that are cloned on pipelines.

While currently it works fine, it is relying on a deep field-by-field clone and reflection into private JVM classes. For `Date` and `Timestamp` this requires access to `sun.util.calendar`'s `BaseCalendar` which has a lazily set field inside `java.util.Date`. 

Such reflection is brittle, and discouraged heavily on Java 16/17. This change reduces need to open access to private Java module APIs for cloning these particular object structures by using a simpler cloning mechanism.

Reduces dependencies for upgrade in #9818.

### DIFF
--- a/util/src/main/java/com/thoughtworks/go/util/ClonerFactory.java
+++ b/util/src/main/java/com/thoughtworks/go/util/ClonerFactory.java
@@ -17,8 +17,12 @@
 package com.thoughtworks.go.util;
 
 import com.rits.cloning.Cloner;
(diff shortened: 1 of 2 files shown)


chadlwilson:
Test all pass, so merging this. Seems safe enough 🙏

PR #9901, opened 2021-11-27
Correct test imports

chadlwilson:
Corrects issues with #9899 test imports.

IntelliJ IDEA doesn't know how to resolve things from server/sharedTest for some reason, so when it optimizes imports on save from `.*`s it can remove things it needs.

Complicating this is the use of the same package across multiple different dependent jars/projects...


chadlwilson:
Removed the milestone, since this is just noise on top of the original PR.

Issue #9873, opened 2021-11-15
Debian 11 (Bullseye) agent docker image


netsandbox:
##### Issue Type

- Feature proposal

##### Summary

Provide a Debian 11 (Bullseye) agent docker image.

By the way you should also think about switching to the `-slim` variant of the Debian Docker images, 
which would result in smaller Debian agent images:
```
REPOSITORY             TAG            IMAGE ID       CREATED              SIZE
gocd-agent-debian-10   v21.3.0-slim   2e5a27a183bf   About a minute ago   474MB
gocd-agent-debian-10   v21.3.0        63d9e1d9ed59   4 minutes ago        543MB
```

chadlwilson:
Noted, thanks for the reminder.

I suspect the (potential?) difficulty in switching to the slim images might be expectation of particular things being available by default in the non-slim images, of use in people's agents if they are using the GoCD images as a base image for something else they layer on top. Would need to look through what is "missing" in slim to see if that would be of concern.

It's not the same, given musl vs glibc, but If you want the smallest images with the lowest attack surface area, the Alpine images are probably the best bet.

netsandbox:
Have a look at `.slimify-excludes` and `.slimify-includes`at https://github.com/debuerreotype/debuerreotype/tree/master/scripts for what is excluded from the slim images.
It doesn't look like something someone would expect in an agent image, but I understand your concern.

PR #9922, opened 2021-12-07
Switch to Debian "slim" variants as agent base image

chadlwilson:
As noted in #9873, there doesn't seem to be any major reason to include all the documentation and locales from the non-slim variants.

See https://github.com/debuerreotype/debuerreotype/blob/master/scripts/.slimify-includes and https://github.com/debuerreotype/debuerreotype/blob/master/scripts/.slimify-excludes for the differences from the full versions.

Seeking input on whether there are additional concerns with this.

### DIFF
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
@@ -164,10 +164,10 @@ enum Distro implements DistroBehavior {
     @Override
     List<DistroVersion> getSupportedVersions() {
       return [
-        new DistroVersion(version: '9', releaseName: 'stretch', eolDate: parseDate('2022-06-30')),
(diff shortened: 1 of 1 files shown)


chadlwilson:
Hi @arvindsv @GaneshSPatil @marques-work (and anyone with an opinion!) - do you think this is a good idea?

It cuts about 40-45MB (35-40% from base) off the uncompressed base image sizes of Debian 10 and 11 and mainly seems to remove man pages and documentation, but also removes some locale stuff for which I don't fully understand possible implications within an agent image.

Don't feel strongly about this as there are no security benefits per se.

arvindsv:
@chadlwilson I think this is a good idea. I don't think there should be any expectation of that documentation being part of the agent. It seems relatively harmless to me.

arvindsv:
I can't see why the locale files should be needed. If the agent comes up fine, then I think it's ok.

chadlwilson:
Alright, thanks - let's give it a go. I'll revert if there any other comments/complaints :-)

PR #9987, opened 2021-12-24
Removed stage-instance version v2

chadlwilson:
This was `deprecatedIn = "20.9.0"` scheduled for `removalIn = "21.1.0"` so should be fine.

We seem to have missed specifically noting the deprecation in https://api.gocd.org/21.3.0/#changes-in-20-9-0 however the API itself will have been returning deprecation information for some time now, so possibly OK?


arvindsv:
Would have been nice if it was announced, but since it has been showing a message, and given that it looks like a couple of fields have been added (no deletions), it should be ok. Looked at:

https://api.gocd.org/current/#get-stage-instance
https://api.gocd.org/20.8.0/#get-stage-instance

chadlwilson:
OK, thanks! Prepared the docs at https://github.com/gocd/api.go.cd/pull/1033

PR #8282, opened 2020-06-29
Migrate preferences page

kritika-singh3:
Description: Migrate the preferences page to mithril behind a toggle `show_old_preference_spa`.
Defaulted to new SPA.

Changes:
 - Added `edit` feature for filter.
 - Disabled add and edit button when SMTP settings are not configured.
 - Updated the Notification filter API v2 to verify whether or not SMTP settings have been configured only for create(POST) and update(PATCH) call.

PR #9988, opened 2021-12-25
Cleanup dead Ruby/Rails APIs and SPAs

chadlwilson:
Trying to do what I can to clean up some work deprecated/replaced in 2019-2020.

- Removes some infra left behind from Rails Scheduled Jobs/Feeds APIs, since moved to Spark/Java-land
  - https://api.gocd.org/21.4.0/#changes-in-20-6-0 
  - #8589
  - Removes the remaining infra that supported Rails APIs in general (etag helpers, deprecation helper)
- Removes the old Ruby/Rails Preferences SPA and the toggle
  - #8282 
  - Removes the remaining infra for Rails SPAs (this seemed to be the last one)
    - Rails layout, navigation elements etc
- Removes old feature toggle for rails template authorization/permissions API
  - API removed some time ago in `20.6.0` in #8268, but UI toggle left lying around
- Removes an unused Ruby AuthenticationHelper
  - Only used by "job details/history config diffs" function, available only to admins. Switch to using [`AuthorizationHelper`](https://github.com/gocd/gocd/blob/3e948218bfb163c5c3d2bf5140cb4a12f110769e/server/src/main/webapp/WEB-INF/rails/app/helpers/admin/authorization_helper.rb#L18-L25) which has an equivalent function and is used by legacy rails admin functions already.  
- Removes some unused code from the Ruby `MaterialsHelper`
  - This is still used on stage/job details to render diff/comment information. 
- Removes an old unused `my-cruise` Rails layout, since replaced with newer SPAs for preferences etc  
- Removes some remaining cruft from the old "properties" API from the UI along with some URL rewrite rules
  - Disabled in `19.8.0` #6887
  - Controller removed in `19.10.0` in #6903


chadlwilson:
Manually tested visually locally, as well as verifying all all regression SPAs are working fine.

If there are any comments from @kritika-singh3 or @arvindsv on things I might have missed cleaning up, or problems with approach I, would appreciate it. Each commit should be self contained for an item in the list.

Side note - the old scss is so confusing, there seems to be a lot of copy+paste/duplication craziness everywhere in GoCD's histroy and I find it very hard to figure out what is used/unused and could be cleansed 😓 Any tips appreciated! e.g
- `rails/app/assets/stylesheets/new-theme.scss`
- `rails/app/assets/stylesheets/css/_application-old.scss`
- `rails/app/assets/stylesheets/layouts/vm/_layout_for_old_ui.scss`
- `rails/app/assets/stylesheets/layouts/_layout.scss`
- overlapping bits from the above broken up into smaller widget-specific files

arvindsv (approved):
Looks good to me. The CSS changes are the ones I'm a little unsure about. Hopefully the selectors are selective enough and aren't reused.

arvindsv:
I think we might need to write / find something automated to deal with the CSS / SCSS. :/

chadlwilson:
> The CSS changes are the ones I'm a little unsure about. Hopefully the selectors are selective enough and aren't reused.

I brute force text search everything I remove due to paranoia 😅 These ones all seem specific to the particular views and I can't see any visual weirdness afterwards so I think all good.

I basically do a human-driven depth first search through the code path to the things it depends on and try to figure out if removing something higher will leave any orphans. I get that this can break with certain levels of dynamicness and rails magic but it generally serves me well enough :)

arvindsv:
:) Yes, with a human-driven approach, it makes sense and seems like the best that can be done. If there _was_ some method to the madness, it could be automated, but it's not ordered, and is confusing as you said.

As long as nothing major is obviously broken, let's go ahead with it.

kritika-singh3:
IIRC:
 - `rails/app/assets/stylesheets/new-theme.scss` - this is the theme which is used in the rails based UI/pages. E.g. stage details
 - `rails/app/assets/stylesheets/css/_application-old.scss` - this is used in the older rail based UI/pages. E.g. job details pages
 - `rails/app/assets/stylesheets/layouts/vm/_layout_for_old_ui.scss` - this is used in old javascript/template based pages. E.g. older pipeline history page (it was replaced with SPA one but not sure if the rails based page was removed), dashboard page
 - `rails/app/assets/stylesheets/layouts/_layout.scss` - all the new typescript based SPAs.

kritika-singh3 (approved):

chadlwilson:
Thanks @kritika-singh3 - that's useful pointers. 👍

PR #10034, opened 2022-01-02
Replace API Info v1 with v2

chadlwilson:
Ironically, while introducing v2 of the API which returns deprecation information in #7739, we seem to forget to deprecate/remove v1 :-)

This
* removes api-info v1 (assuming can be done without announcement as internal?)
* adds tests for api-info v2
* (_edit_) corrects behaviour of api-info v2 as it did not seem to be working correctly


arvindsv:
I think the idea for `/internal` APIs is that they can be removed without any announcement, since their only client is the GoCD UI.

chadlwilson:
Right, OK - thanks! I am digging a bit deeper as the V2 API doesn't seem to be working as expected (I was trying to get https://github.com/gocd/ruby-functional-tests/pull/622 to pass) and the representer is missing some tests.

chadlwilson:
It looks like the v2 version wasn't fully completed as it's not totally working and appears unused right now. I do have it working locally, but not sure on intent, so let me see whether Ganesh/Mahesh can recall on https://github.com/gocd/gocd/pull/7739#issuecomment-1008263765 and whether it is worth me fixing v2 so we can proceed to drop v1.

PR #10152, opened 2022-02-05
Downgrade Tanuki wrapper to 3.5.46 (as per last release)

chadlwilson:
Wrapper versions 3.5.47+ seem to have some funny behaviour in some cases where they spin in a tight CPU loop. Only observed this on build.gocd.org where there is a custom GoCD Server docker image, and supervisord is wrapping launch of the docker entrypoint and the tanuki wrapper, however it is replicable in that specific scenario on both 3.5.47 and 3.5.48. The problem goes away with a downgrade.

There may be a config oriented fix, but not sure right now. https://wrapper.tanukisoftware.com/doc/english/release-notes.html#3.5.47

This undoes #10033 for now.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -184,8 +184,8 @@ final Map<String, String> v = [
   ztExec              : versionOf(libraries.ztExec),
 
   // misc
-  tanuki              : '3.5.48-st',
-  tanukiSha256sum     : '7619e1d4320efcb7c2f5bafc36347993c3ff6417e2b8289e90c047033139c514',
+  tanuki              : '3.5.46-st',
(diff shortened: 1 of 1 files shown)


chadlwilson:
Asked the Tanuki folks, and they replied that a workaround is to set `wrapper.java.new_process_group=FALSE` however a fixed version should be available soon. Seems to relate to TTYs somehow.

Can be replicated locally:

**100% CPU Use** (pseudo-TTY, wrapper 3.5.48): `docker run -t gocdexperimental/gocd-server:v22.1.0-13695`

**OK/Normal** (_no pseudo-TTY_, wrapper 3.5.48): `docker run gocdexperimental/gocd-server:v22.1.0-13695`
**OK/Normal** (pseudo-TTY, _wrapper 3.5.46_): `docker run -t gocdexperimental/gocd-server:v22.1.0-13697`

PR #10157, opened 2022-02-06
Fix TFS usage under Java 17

chadlwilson:
Refactor the TFS command's NestedJarClassLoader to not require internal access

Fixes an issue with #9818 

Use of setAccessible is blocked by default on Java 17. The old implementation was directly referring to the parent classLoader in a strange way using reflection rather than just using the super class to handle that for us. Not really sure why or have super detailed classLoader understanding, but will validate whether this works adequately. First indication is that it looks OK but I can't run all the regression tests locally or work out how to authenticate to an Azure DevOps TFVC instance.

I think this was missed earlier because some of the regression tests were actually running under Java 15 rather than Java 17 on my branch builds as expected.


chadlwilson:
Everything passing; let's give this is a go.

PR #10241, opened 2022-03-04
Remove stray ;

JeroenOortwijn:
Previously, the help text on the General tab of a Pipeline showed:
`If unchecked, this pipeline will only schedule in response to a Manual/API/Timer trigger. Unchecking this box is the same as making the first stage manual.;`

Now it will be:
`If unchecked, this pipeline will only schedule in response to a Manual/API/Timer trigger. Unchecking this box is the same as making the first stage manual.`


chadlwilson:
Thanks!

PR #10243, opened 2022-03-05
Migrate to use of Charset type everywhere we can

chadlwilson:
* Switches to `Charset` type in almost all places (It's more implicitly type-safe, and more "modern java" than passing about String encodings, also more efficient generally as less `Charset.forName`s required)
  * Didn't migrate the field in `BuildWork` for now as this seems to be serialized onto internal queues with Java-built-in binary serialization and the `Charset` type isn't serializable
* Tidies up some of the utility packages and minor tweaks to related classes


chadlwilson:
This currently isn't suitable to merge, as apparently the legacy RMI remoting code is still there and `Charset` isn't serializable....

chadlwilson:
This is dependent on at least #10300 which will remove the legacy RMI comms.

chadlwilson:
Legacy remoting is removed, but there is still some Java-built-in serialization. Will follow up with addressing the last remaining bit.

PR #10300, opened 2022-03-29
Remove RMI communications from agent to server

chadlwilson:
This was deprecated in GoCD 21.1.0 due to introduction of HTTP/JSON comms in #8898 and disabled by default in the same release (#8929). 

I believe it should be safe to remove now, since it has been replaced by `RemotingClient` which uses regular HTTP/JSON comms.


chadlwilson:
Hi @marques-work, @maheshp, @arvindsv - any commentary on this RMI removal? Anything else I need to be aware of/do?

maheshp:
@chadlwilson this LGTM, nothing else I can think of. I have not tested the changes, hoping our PR pipelines will give the right feedback.

chadlwilson:
Thanks @maheshp - in addition to this, I did notice that there are some tests around the place using https://github.com/gocd/gocd/blob/2daf703480f3d33cf5f06eeb0efd218bf2f031ae/test/test-utils/src/main/java/com/thoughtworks/go/utils/SerializationTester.java which tests serialize/deserialize to objects. I assume this was to ensure the objects transferred by the earlier RMI mechanism could serialize OK. Still need to work out if there are similar tests for the replacement before I remove, but I assume so.

PR #10455, opened 2022-05-23
Attempt to fix TS errors in preparation to TS 4.6.4 upgrade

mapkon:
Related PR: https://github.com/gocd/gocd/pull/10387

Attempt to fix the TS compiler errors in preparation for TS upgrade to 4.6.4. 

- Await for the result in a promise, as opposed to using the promise function in a truthy way, as that will always pass since the function exists
- Currently using `unknown` as an intermediate type to see if it fixes the issue. Might require modifying the structure of the expected types in order to match.
- Enabled downleveling since we are targeting the not-so-recent greatest version of ECMAScript.


chadlwilson:
Thanks for the help! Looks fine to me - rebased and merging now.

PR #10461, opened 2022-05-26
Revert "Bump org.apache.felix.framework from 7.0.3 to 7.0.4"

chadlwilson:
Reverts gocd/gocd#10460

Looks like change broke something on Windows, perhaps related to https://github.com/apache/felix-dev/pull/154


chadlwilson:
See https://build.gocd.org/go/pipelines/build-windows/5741/build-non-server/1/overview

PR #10493, opened 2022-06-05
Update material userName/password in sync with one another when constructing BuildWork

### LINKED ISSUE #8986: GitMaterial Authentication failed for the second user who using same repository
##### Issue Type

- Bug Report

##### Summary

If two git materials are configured in two pipelines with same git repo and different username + accesstoken, the second user will failed to checkout.

##### Environment

```plain
"Go Server Information": {
    "Version": "20.6.0 (12005-12860aac6351e2a353728c7d7913f34d741c63e0)"
  }，
"OS Information": {
    "OS Name": "Linux",
    "OS Version": "3.10.0-862.el7.x86_64",
    "System Architecture": "amd64",
    "Available Processors": 24,
    "Average System Load": 0.67
  },
```

* Go Version: `20.6.0 (12005-12860aac6351e2a353728c7d7913f34d741c63e0)`
* JAVA Version: `13.0.2`
* OS: `Linux 3.10.0-862.el7.x86_64`

##### Steps to Reproduce

1. create two pipeline with git credential `https://gitlab.somehost/someproject.git`, username and password(gitlab accesstoken) are provided in advanced settings. 
2. the username/accesstokens of those git material are belongs to two person. now call them userA and userB.
3. While click `test connection` in `Edit material - Git`, both repositories says: "Connection OK".
4. trigger the first pipeline mannually, later it will success.
5. trigger the second pipeline mannually, later it will output `Authentication failed`

##### Expected Results
The second pipeline should works as the first one.

##### Actual Results
The second pipeline build failed with message `Authentication failed` and `Failed to run git clone command`

##### Possible Fix

I have read some source code, the access token are stored correctly in cruise file.The material editing page uses it too.

The method `com.thoughtworks.go.config.materials.git.GitMaterial#appendCriteria` is like:
```
        parameters.put(ScmMaterialConfig.URL, url.originalArgument());
        parameters.put("branch", refSpecOrBranch);
```
the `generateFingerprintFromCriteria` is using `type`, `url`, `branch`, it will using the first pipeline's material.

It seems that is agent JVM, the `com.thoughtworks.go.config.materials.ScmMaterial`:
```
    protected String userName; //the userName is always UserA.
    protected String password;
```

I cannot figure out why `GitMaterial` failed to checkout userB's pipeline.

##### Log snippets
To see command line logs, provide:
logback-agent-include.xml
```
<logger name="com.thoughtworks.go.util.command.CommandLine" level="DEBUG" >
     <appender-ref ref="commandLineLog"/>
</logger>
```

Agentlog
```
2021-01-20 17:15:51,543 DEBUG [scheduler-3] CommandLine:350 - Running git config remote.origin.url
2021-01-20 17:15:51,681 DEBUG [scheduler-3] CommandLine:362 - Output: 
https://UserB:UserB'Accesstoken@gitlab.host/project.git
2021-01-20 17:15:51,778 DEBUG [scheduler-3] CommandLine:368 - Running git clone --branch deploy https://UserA:******@gitlab.host/project.git /var/lib/go-agent/pipelines/HandyCreated
2021-01-20 17:15:52,164 ERROR [scheduler-3] DefaultGoPublisher:137 - Failed to run git clone command
java.lang.RuntimeException: Failed to run git clone command
        at com.thoughtworks.go.util.ExceptionUtils.bomb(ExceptionUtils.java:37)
        at com.thoughtworks.go.config.materials.git.GitMaterial.updateTo(GitMaterial.java:157)
        at com.thoughtworks.go.domain.MaterialRevision.updateTo(MaterialRevision.java:147)
        at com.thoughtworks.go.domain.materials.AbstractMaterialAgent.prepare(AbstractMaterialAgent.java:43)
......
Caused by: java.lang.Exception: Failed to run git clone command
        at com.thoughtworks.go.util.ExceptionUtils.bombIfFailedToRunCommandLine(ExceptionUtils.java:51)
        at com.thoughtworks.go.config.materials.git.GitMaterial.git(GitMaterial.java:384)
        at com.thoughtworks.go.config.materials.git.GitMaterial.updateTo(GitMaterial.java:151)
        ... 20 common frames omitted

```

```
[go] Start to update materials.

[go] Start updating files at revision 65d9a85e92e8284857409fe62fcef2d80fb9211c from https://gitlab.<host>.com/<project>.git
Cloning into '/var/lib/go-agent/pipelines/HandyCreated'...
STDERR: fatal: Authentication failed for 'https://<userA>:******@<host>/<project>.git/'
Failed to run git clone command
```

image
image

### LINKED ISSUE #9153: Git Material Credentials can't be changed
##### Issue Type
Bug Report

##### Summary

If you try to change the user / password of an git material the checkout is still using the old user but with the password of the new user. The connection test passes while updating the material config.

##### Environment
Docker based setup tested with gocd/gocd-server:21.1.0 and 21.2.0

* Stock gocd docker image with h2 database
* Only docker and docker swarm elastic agents

##### Steps to Reproduce

1. Create new Pipeline and define a git material with basic authentication 
2. Run Pipeline -> Success (at least the checkout)
3. Edit Pipeline and change the User + Password to another user with access to the Repository
4. Click Check Connection (will succeed)
5. Save and run the Pipeline again
Now it will fail on checkout and you will see that its using the username that you configured in the beginning but the password you set in step 3

##### Expected Results
GoCD should be able to update the credentials of git materials

##### Actual Results
GoCD does only update the password used on the material but not the user which causes the checkout to fail.

##### Any other info
This might be related to https://github.com/gocd/gocd/issues/8986

chadlwilson:
- fixes #8986
- fixes #9153

When loading a triggered run from the DB to construct buildWork, the userName from the material is, for some reason persisted. However the password is not and is dynamically loaded from the GoConfig based on the pipeline name and fingerprint. So when determining the correct password to use for construction of `BuildWork` to dispatch to the agent, it will use the one from the `MaterialConfig` attached to the pipeline, not the de-duplicated material used for triggering.

If the pipeline is configured with a different username for the same url/branch in two different pipelines, the password will be updated but not the userName, leaving them out of sync. For some source control targets this doesn't seem to matter when using tokens (e.g GitHub) but others require the username to match that whom generated the token, and  others still support legacy username/password.

In any case, the entire credential needs to be updated together to ensure that it is consistent, rather than assuming that the persisted `userName` is the same one corresponding to the `password`.

The other alternative would be to review the persisted material revisions to see whether the material usernames should be persisted with the correct one relevant to the material attached to each scheduled pipeline, but in any case, updating username/pass together seems to be the correct thing to do to me.

Testing
- [x] Replicated with unit-ish test on `ScheduledPipelineLoader`, the source of the problem
- [x] Characterisation unit tests added for location of the direct change/fix
- [x] Replicated and tested fix locally with Git via Github (two accounts with different usernames and tokens, by checking the username/token combinations are accurate with all combinations)

### DIFF
--- a/common/src/test/java/com/thoughtworks/go/domain/materials/AbstractMaterialTest.java
+++ b/common/src/test/java/com/thoughtworks/go/domain/materials/AbstractMaterialTest.java
@@ -18,6 +18,8 @@ package com.thoughtworks.go.domain.materials;
 import com.thoughtworks.go.config.PipelineConfig;
 import com.thoughtworks.go.config.materials.AbstractMaterial;
 import com.thoughtworks.go.config.materials.SubprocessExecutionContext;
+import com.thoughtworks.go.config.materials.svn.SvnMaterial;
+import com.thoughtworks.go.config.materials.svn.SvnMaterialConfig;
 import com.thoughtworks.go.domain.MaterialInstance;
 import com.thoughtworks.go.domain.MaterialRevision;
 import com.thoughtworks.go.util.command.ConsoleOutputStreamConsumer;
@@ -30,7 +32,7 @@ import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
-import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.*;
 
 public class AbstractMaterialTest {
 
@@ -206,5 +208,28 @@ public class AbstractMaterialTest {
         assertThatCode(() -> material.checkConnection(executionContext))
                 .isInstanceOf(UnsupportedOperationException.class)
                 .hasMessage("'checkConnection' cannot be performed on material of type name");
+    }
+
+    @Test
+    public void updateConfigShouldIgnoreNonPasswordAwareConfig() {
+        MaterialConfig config = mock(MaterialConfig.class);
+        SvnMaterial passwordAwareMaterial = new SvnMaterial("url", "username", "password", false);
+        passwordAwareMaterial.updateFromConfig(config);
+        verifyNoInteractions(config);
+    }
+
+    @Test
+    public void updateConfigShouldTakeUserPassFromPasswordAwareConfig() {
+        SvnMaterialConfig config = mock(SvnMaterialConfig.class);
+        when(config.getUserName()).thenReturn("newUsername");
+        when(config.getPassword()).thenReturn("newPassword");
+        SvnMaterial passwordAwareMaterial = new SvnMaterial("url", "username", "password", false);
+        passwordAwareMaterial.updateFromConfig(config);
+
+        assertThat(passwordAwareMaterial.getUserName()).isEqualTo("newUsername");
+        assertThat(passwordAwareMaterial.getPassword()).isEqualTo("newPassword");
+
+        verify(config).getUserName();
+        verify(config).getPassword();
     }
 }
--- a/config/config-api/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterialConfig.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/config/materials/AbstractMaterialConfig.java
@@ -121,8 +121,8 @@ public abstract class AbstractMaterialConfig implements MaterialConfig, ParamsAt
     public String getTruncatedDisplayName() {
         String displayName = getDisplayName();
         if (displayName.length() > TRUNCATED_NAME_MAX_LENGTH) {
-            StringBuffer buffer = new StringBuffer();
-            buffer.append(displayName.substring(0, TRUNCATED_NAME_MAX_LENGTH / 2));
+            StringBuilder buffer = new StringBuilder();
+            buffer.append(displayName, 0, TRUNCATED_NAME_MAX_LENGTH / 2);
             buffer.append("...");
             buffer.append(displayName.substring(displayName.length() - TRUNCATED_NAME_MAX_LENGTH / 2));
             displayName = buffer.toString();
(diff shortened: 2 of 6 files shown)


chadlwilson:
Hi @arvindsv - would you be in a position to review this? Regression tests passing, FWIW.

Pretty confident I am at the root of the problem, and can't see any downside to the "fix", but not sure what other specific scenario combinations I might need to test. e.g 
* could secret params or other parameter substitutions affect this somehow?
* is there a way that a usernames in the URL + a username directly specified might change semantics with this change? 🤔
* are there specific concerns for other material types I might be missing? The reported issues are to do with Git, but suspect that's more due to its ubiquity. All `PasswordAwareMaterial`s are potentially effected, so all SCMs except `Pluggable`.

arvindsv:
Hey @chadlwilson. Yes, I had my eye on this, because I know this came up as a problem so many times. I'm glad you found the cause. I'll take a look soon, and give it some thought. Thank you!

arvindsv:
Sorry about the delay, @chadlwilson. I thought the change was much bigger than it was, and I kept pushing it till I had a bigger chunk of time to spend on this. :(

I was a little worried about possible changes in the fingerprint method, but it doesn't look like anything changed there. Phew. Changing that by mistake is one of my regular nightmares. :)

This looks good, and the fact that all the regression tests are passing adds to the confidence. I tried to think about secret params, but feel that it should be fine. Template / parameter substitution should happen quite early anyway, much before this time. So, I will merge this. Thank you!

chadlwilson:
Thanks @arvindsv! 

Yeah, im aware of the criticality of the fingerprint, partly due to my intro to GoCD dev being via the git path material plugin which adds a proper path prefix to the fingerprint (and filters modifications at git level) to workaround issues with fan-in caused by use of the exclude/include filters (which don't affect the fingerprint).

Issue #9405, opened 2021-06-08
Error while updating config or settings


axcl:
##### Issue Type

- Bug Report

##### Summary

Whenever i tries to change config for roles or any plugins it throws following error
I have also tried completing removing and adding plugin again but somehow config still stucked.

```
Someone has modified the entity. Please update your copy with the changes and try again.

Someone has modified the configuration for role 'Admin'. Please update your copy of the config with the changes and try again.
```

###### Basic environment details

* Go Version: 21.2.0
* JAVA Version: 15.0.2
* OS: Linux 4.15.0-144-generic
* Browser vendor and version (if relevant): Latest chrome

##### Steps to Reproduce

1. 
2. 
3. 
4. 

##### Any other info

https://ibb.co/7yswJwK
https://ibb.co/2S0Qksd

msharbaji:
I am facing the same issue, anyone faces the issue and manage to solve it?

msharbaji:
@axcl have you fixed the issue?

axcl:
No i have switched to Jenkins

chadlwilson:
A similar problem was reported at #9964. Were you folks using GoCD behind some sort of caching reverse proxy or CDN, e.g Cloudflare?

msharbaji:
@chadlwilson Yes, indeed it has a caching reverse proxy which Cloudflare, my colleague reported the problem you mentioned on, We add page rule to respect the strong ETag, but I'd suggest if the GoCD can catch it and accept either weak ETag or strong ETag.

Thanks for your support, really helped us with this hint and unblock us with operational stuff :)

chadlwilson:
No worries, and thank you for confirming!

I am glad to have gotten to the bottom of the problem. Now to figure out the right way to address it - will track in the linked ticket.

If you have any information on other APIs/screens you have seen this type of problem with (or ones where updates definitely worked without the page rule) that'd be helpful, as currently we dont quite understand why the issue has been reported against plugin settings and roles, since almost all the APIs that update a single "entity" or piece of config have the same protection to avoid lost updates between different users' API calls, so I'd have expected issues to be reported against lots of APIs.

If folks are mainly using things such as pipelines-as-YAML for non-admin functions, that might explain it, since I imagine role config and plugin settings are the two more common admin functions which are more likely to involve UI configuration.

msharbaji:
@chadlwilson indeed not just the updating role, we had an issue with all updates, example: update pipeline template we were facing the same issue :)

[...]

Issue #6278, opened 2019-05-09
GET plugin API returns weak ETag validator


ankitsri11:
##### Issue Type

- Feature enhancement

##### Summary

Currently, GET plugin API returns "W/" in the ETag identifier whereas GET specific environment API returns ETag identifier without "W/".

https://api.gocd.org/current/#get-plugin-settings

For example:

```
curl -kv 'https://localhost:8154/go/api/admin/plugin_settings/com.thoughtworks.gocd.analytics' \
        -H 'Authorization: bearer XXXXXXXXXXXXX' \
      -H 'Accept:application/vnd.go.cd.v1+json' \
      -i

*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8154 (#0)
.
.
.
< ETag: W/"fd6d6c7c005834d62981a0c5505144ea"
ETag: W/"fd6d6c7c005834d62981a0c5505144ea"
.
.
.

```

arvindsv:
@ankitsri11 Whether it is a weak etag or not _should_ not matter. It should be an implementation detail.  Technically, providing whatever comes back, when making an update should work. Can you try using those different ETags and seeing what happens?

ankitsri11:
@arvindsv: Here is an example which I tried updating plugin setting using PUT request with and without **W/**:

- [GET request](https://api.gocd.org/current/#get-plugin-settings) to get the plugin settings

```
» curl -kv 'http://localhost:8153/go/api/admin/plugin_settings/com.thoughtworks.gocd.analytics' \                                                                                            
      -H 'Accept:application/vnd.go.cd.v1+json'

*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8153 (#0)
> GET /go/api/admin/plugin_settings/com.thoughtworks.gocd.analytics HTTP/1.1
> Host: localhost:8153
> User-Agent: curl/7.54.0
> Accept:application/vnd.go.cd.v1+json
>
< HTTP/1.1 200 OK
< Date: Mon, 13 May 2019 20:56:11 GMT
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< X-UA-Compatible: chrome=1
< Set-Cookie: JSESSIONID=node01k9xr0l9npccrxpojeek4vwci3.node0;Path=/go;Expires=Mon, 27-May-2019 20:56:11 GMT;Max-Age=1209600;HttpOnly
[...]

PR #10515, opened 2022-06-12
Normalize client-sent headers to strong ETags

### LINKED ISSUE #9964: Can't add user into a GoCD role behind CloudFlare minification proxy
##### Issue Type

- Bug Report

* Go Version: 21.3.0 
* JAVA Version: 15.0.2
* OS: Linux 5.4.129+
* Browser vendor and version (if relevant): Chrome 96.0.4664.93

##### Steps to Reproduce

1. MainNavigation:ADMIN->Roles configuration
2. Add  a new GoCD Role.
3. Edit the role just created. Add a user and Save .

##### Expected Results

The user was added into the role.
##### Actual Results

An error showed up like:Someone has modified the configuration for role 'XXX'. Please update your copy of the config with the changes and try again.

##### Possible Fix

sorry

##### Log snippets

sorry

image
##### Any other info

sorry

chadlwilson:
Fixes #9964, #9405, https://github.com/gocd-contrib/email-notifier/issues/46

**Background**
Currently GoCD always sets strong ETags in response headers as noted in #6278. Putting aside whether this is wise or not (as opposed to setting weak etags), we use these etags both to determine whether a resource is modified or whether there has been a race condition update in `If-None-Match` and `If-Match` headers respectively. 

https://datatracker.ietf.org/doc/html/rfc7232#section-3.1 says that when comparing to the `If-Match` header (for updates, basically) we should use "strong comparison" which effectively means the a weak etag validator is not considered equal to the strong etag validator with the same content. While accepting these RFCs were written with the intention of more old-school webserver file uploads _rather than REST APIs_, the requirement to do this is a problem because some caching proxies (e.g CloudFlare) have features to minify and re-compress JSON responses for optimization and [convert our strong etag to a weak etag](https://support.cloudflare.com/hc/en-us/articles/218505467-Using-ETag-Headers-with-Cloudflare). This minification can be disabled and strong etags kept in tact, however it doesn't seem this should be a requirement for GoCD use and there have been a number of tickets raised over the years. All have been linked to Cloudflare, although I imagine even a reverse proxy like NGINX may do the same thing.

This conversion to weak etags before client return causes updates of these resources to fail, which is confusing and not desirable.

**Addressing the issue**
Technically we already compare etags in a slightly "non-strong" way because the server ignores `--gzip` and `--deflate` suffixes that Jetty can add when compressing responses already. e.g the below (there are also some [hacks in JS client side code](https://github.com/gocd/gocd/blob/2daf703480f3d33cf5f06eeb0efd218bf2f031ae/server/src/main/webapp/WEB-INF/rails/webpack/views/dashboard/models/personalization_vm.js#L164-L166) to remove the some suffixes, maybe due to legacy Rails APIs since replaced, but ignoring these for now) 

https://github.com/gocd/gocd/blob/2daf703480f3d33cf5f06eeb0efd218bf2f031ae/api/api-base/src/main/java/com/thoughtworks/go/api/ControllerMethods.java#L54-L62

It also doesn't really make sense to compare objects "byte-by-byte" when considering API responses which do not really have a canonical octal/binary representation.

In any case, it seems to express the intent more clearly to have our UI JS _normalize the ETags it sends back in headers to strong ETags_, which is what this change does.

**Alternative Fix**
Alternative would be to change the matching on the server side to allow a weak etag to match a strong etag in the normalization logic in `ControllerMethods` location above, which would be a much simpler change.

However this seems to go against the spirit of the spec, although it is possible I am misreading it. When you read the description [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#weak_validation) it reads as if the server should do a weak validation if the client puts a weak validator in the header, but [the spec reads differently](https://datatracker.ietf.org/doc/html/rfc7232#section-3.1) as if any weak validator ever sent in an `If-Match` header will never match when doing "Strong Comparison" as it requires. I'm not sure how people implement this in practice.

The downside of the approach taken in this PR is that use of the GoCD APIs via other client code will likely have similar issues, and have to normalize their own ETags in PUT calls made to the API.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/spec/webpack/views/dashboard/model/personalization_vm_spec.js
+++ b/server/src/main/webapp/WEB-INF/rails/spec/webpack/views/dashboard/model/personalization_vm_spec.js
@@ -162,7 +162,7 @@ describe("Personalization View Model", () => {
       jasmine.Ajax.stubRequest(SparkRoutes.pipelineSelectionPath(), undefined, 'GET').andReturn({
         responseText:    JSON.stringify({filters: [{ name:"New", type: "whitelist", pipelines: ["a"] }]}),
         responseHeaders: {
-          ETag:           `"1234567"`,
+          ETag:           `W/"1234567"`,
           'Content-Type': 'application/vnd.go.cd.v1+json'
         },
         status:          200
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/api_request_builder.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/api_request_builder.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 import {CaseInsensitiveMap} from "helpers/collections";
+import {mrequest} from "helpers/mrequest";
 import _ from "lodash";
 import m from "mithril";
 
@@ -78,7 +79,7 @@ export class ApiResult<T> {
   }
 
   getEtag(): string | null {
-    return this.header("etag") || null;
+    return mrequest.normalizeEtag(this.header("etag"));
   }
 
   getRedirectUrl(): string {
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/mrequest.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/mrequest.ts
@@ -16,6 +16,7 @@
 
 import * as CONSTANTS from "helpers/constants";
 import $ from "jquery";
+import _ from "lodash";
 
 const setHeaders = (xhr: JQuery.jqXHR, version: string) => {
   xhr.setRequestHeader("Content-Type", "application/json");
@@ -60,10 +61,17 @@ export const mrequest = {
     }
   },
 
+  normalizeEtag(value: string | undefined | null) {
+    return _.isNil(value) ? null : value
+      .replace(/^W\//, "")
+      .replace(/"/g, "")
+      .replace(/--(gzip|deflate)$/, "");
+  },
+
   unwrapMessageOrEntity: (type: any, originalEtag: string) => (data: any, xhr: JQuery.jqXHR) => {
     if (xhr.status === 200) {
       const entity = type.fromJSON(data);
-      entity.etag(xhr.getResponseHeader("ETag"));
+      entity.etag(mrequest.normalizeEtag(xhr.getResponseHeader("ETag")));
       return entity;
     }
     if (xhr.status === 422 && !!data.data) {
--- a/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spec/api_request_builder_spec.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/helpers/spec/api_request_builder_spec.ts
@@ -134,6 +134,41 @@ describe("Api Request Builder", () => {
     });
   });
 
+  describe("ETag handling", () => {
+    it("should normalize quoted etags to unquoted", (done) => {
+      mockSuccessfulRequest("GET", undefined, '"hello"');
+      ApiRequestBuilder.GET("/foo", ApiVersion.v1).then((result) => {
+        expect(result.getEtag()).toEqual("hello");
+        expect(result.getStatusCode()).toEqual(200);
+        done();
+      }, () => {
+        done.fail("should have passed");
+      });
+    });
+
+    it("should normalize weak Etags to strong", (done) => {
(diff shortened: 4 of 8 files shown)


chadlwilson:
Hi @GaneshSPatil and @arvindsv - if you have any comments on this, I'd appreciate it.

arvindsv:
Oh, sorry. I see that I was supposed to take a look at this. Will do later today or early tomorrow.

arvindsv:
Can I make sure my understanding of the current process is right?

1. Client UI makes a request to the GoCD server (say a `GET`)
2. GoCD server (always?) sends strong ETags in response
3. It's possible that Cloudflare or other reverse proxy _changes_ that to a weak ETag while minifying the body of that response
4. Client sees the weak ETag ("W/....") and then uses that in a subsequent `PUT` response, in its `If-Match` header.
5. GoCD server sees the weak entity-tag in the `If-Match` header and refuses to match it with the entity, since it's doing a strong comparison

Is that right?

The fix being proposed is to strip out the "weakness" of the ETag between steps 3 and 4 above. This way, any subsequent response in step 5 will be sending a strong entity-tag as a part of the `If-Match` header and everything works.

This assumes that Cloudflare will not minify the response from the client to the server, in step 4, right? Is that a valid assumption? The page you linked to just talks about content coming from the origin server.

Your reading of the RFC seems right to me. It does, unambiguously, say that the comparsion MUST be a strong one. So, if one side of the two compared values is weak (as in step 4), then the comparison will have to fail.

If my understanding of all of the above is correct, then I am fine with this change. It removes double-quotes, the "W/" prefix and the "gzip/deflate" suffix properly.

chadlwilson:
@arvindsv sorry for the slow reply.

> Can I make sure my understanding of the current process is right?
> 
>     1. Client UI makes a request to the GoCD server (say a `GET`)
> 
>     2. GoCD server (always?) sends strong ETags in response
> 
>     3. It's possible that Cloudflare or other reverse proxy _changes_ that to a weak ETag while minifying the body of that response
> 
>     4. Client sees the weak ETag ("W/....") and then uses that in a subsequent `PUT` response, in its `If-Match` header.
> 
>     5. GoCD server sees the weak entity-tag in the `If-Match` header and refuses to match it with the entity, since it's doing a strong comparison
> 
> 
> Is that right?

Effectively, yes, because the server does a "strong comparison" of `W/"blah"` to `blah` and they won't match.

> The fix being proposed is to strip out the "weakness" of the ETag between steps 3 and 4 above. This way, any subsequent response in step 5 will be sending a strong entity-tag as a part of the `If-Match` header and everything works.

Yup, that's right.

> This assumes that Cloudflare will not minify the response from the client to the server, in step 4, right? Is that a valid assumption? The page you linked to just talks about content coming from the origin server.

It's "OK" if Cloudflare minifies the response because Cloudflare doesn't actually change the ETag's raw value, just says that it is weak since it doesn't retain its byte-by-byte equality anymore after minification. Essentially what we are saying (with a client-side change here) is that for the purposes of avoiding update races, we are OK if the two responses are "logically identical" rather than "byte-for-byte" identical, i.e we actually don't care about the strong comparison in the specific usage of etags here.

> If my understanding of all of the above is correct, then I am fine with this change. It removes double-quotes, the "W/" prefix and the "gzip/deflate" suffix properly.

The existing code is a bit weird here, because client size strips the `--gzip` or `--deflate` but when comparing on the server-side in the quoted code above, it also strips these. I suspect the client-side is doing such munging as well because at various historical points there were multiple places these etag comparisons were being done server side with the old Rails APIs. I haven't attempted to simplify/clean this up right now, so have just extended the client-side "normalization" of the ETags.

arvindsv:
> It's "OK" if Cloudflare minifies the response because Cloudflare doesn't actually change the ETag's raw value, just says that it is weak since it doesn't retain its byte-by-byte equality anymore after minification.

Ok. I was thinking of the communication from the _client_ to the server.

1. Client requests a list of users. It gets a list, and CloudFlare changes the ETag to "W/12345".
2. Client now wants to update one of the users, and submits with an If-Match of "12345".
3. CloudFlare modifies that _back_ to "W/12345".
4. Server rejects it again.

However, I think my thinking around step 3 is wrong. CloudFlare will likely not mess with `If-Match` headers.

chadlwilson:
Ahh right, yeah - my bad in the misunderstanding. Yeah, it shouldn't mess with those unless you've told it to, to my knowledge (like telling it to strip headers, add additional headers etc in its role as a reverse proxy).

PR #10520, opened 2022-06-13
Change GoCD to no longer force links to open in new windows/tabs

### LINKED ISSUE #9526: Please provide some way to stop new windows from being opened
It is remarkable to me that in 2021, I have a piece of software with RESTful URLs that will not let me choose to open, for example, console output in the same window. This is extremely annoying. I do not want 12 tabs for GoCD opened, let me choose if I want to open something in a new window or the same window. I am just trying to drill down into a pipeline to see what happened and I have to open a new window for that? Why?! This is so frustrating. Give me a choice!

chadlwilson:
Fixes #9526, related to #9613

Internal links will open by default in the same window/tab, returning control to the user to be able to middle-click (CMD + click) for new tab/window, and shift/middle-click for new background tabs, reducing a relatively age-old user complaint about excessive tabs being opened, and getting lost in tabs.

External and docs links continue to generally open in new tabs for now, but link display has been made more consistent to generally give a good indicaation with the icon if something is about to force-open in a new tab, e.g 
image

### DIFF
--- a/server/src/main/resources/server_in_maintenance_mode.html
+++ b/server/src/main/resources/server_in_maintenance_mode.html
@@ -75,7 +75,7 @@
       <div class="message">
         <span id="message_text"/>
         &nbsp;
-        <a target="_blank" href="%docsBaseUrl%/advanced_usage/maintenance_mode.html">Learn more..</a>
+        <a target="_blank" href="%docsBaseUrl%/advanced_usage/maintenance_mode.html">Learn More</a>
       </div>
     </div>
   </div>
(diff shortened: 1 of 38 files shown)


GaneshSPatil:
Hey @chadlwilson - the changes LGTM.

Lets wait to get an agreement if we want to make this change based on [this comment](https://github.com/gocd/gocd/discussions/9613#discussioncomment-3089112)

Also, ~I~ We might need to take a relook at the application to see at how many places we're opening links in new tab. 
I remember at some places, we've done this using javascript 😛

PR #10576, opened 2022-07-02
Return job state transitions from /api/jobs "Get Job Instance" API

### LINKED ISSUE #10566: Job Instance API does not return `job_state_transitions`
##### Issue Type

- Bug Report

##### Summary

The job instance API with 22.2.0 (master) is not returning any `job_state_transitions` as is mentioned in the documentation. When I run the following command, the `job_state_transitions` is empty:

`curl  -H 'Accept: application/vnd.go.cd+json' -s http://admin:badger@localhost:8153/go/api/jobs/up42/1/up42_stage/1/up42_job`

```
{
  "name" : "up42_job",
  "state" : "Scheduled",
  "result" : "Unknown",
  "original_job_id" : null,
  "scheduled_date" : 1656610300191,
  "rerun" : false,
  "agent_uuid" : null,
  "pipeline_name" : "up42",
  "pipeline_counter" : 1,
  "stage_name" : "up42_stage",
  "stage_counter" : "1",
  "job_state_transitions" : []
}
```

Digging a bit deeper, it looks like the [api controller](https://github.com/gocd/gocd/blob/54e657fd508ebb1f8764988456f5e78e42f076e8/api/api-job-instance-v1/src/main/java/com/thoughtworks/go/apiv1/jobinstance/JobInstanceControllerV1.java#L92) is invoking a service method that does not pull in the relevant job state transition data from DB.

chadlwilson:
Fixes #10566

- [x] Unit test gaps to be reviewed
- [x] Regression on usages of `findJobInstance` to be verified (should be OK, it's just the feed XML?)
- [x] Regression on usages of DAO to be verified
- [x] To check whether history should also return (decided that no, want address this now, possible performance implication and not worth the effort)
- [x] To update docs
   - [x] Add example of transitions
   - [x] Remove misleading empty transitions array from history api return
- [x] Looking for feedback on whether there are API, narrow integration tests or functional/regression tests that would normally be

### DIFF
--- a/api/api-job-instance-v1/src/main/java/com/thoughtworks/go/apiv1/jobinstance/JobInstanceControllerV1.java
+++ b/api/api-job-instance-v1/src/main/java/com/thoughtworks/go/apiv1/jobinstance/JobInstanceControllerV1.java
@@ -89,7 +89,7 @@ public class JobInstanceControllerV1 extends ApiController implements SparkSprin
         String jobName = request.params("job_name");
         Integer pipelineCounter = getValue(request, "pipeline_counter");
         Integer stageCounter = getValue(request, "stage_counter");
-        JobInstance jobInstance = jobInstanceService.findJobInstance(pipelineName, stageName, jobName, pipelineCounter, stageCounter, currentUsername());
+        JobInstance jobInstance = jobInstanceService.findJobInstanceWithTransitions(pipelineName, stageName, jobName, pipelineCounter, stageCounter, currentUsername());
         if (jobInstance.isNull()) {
             throw new RecordNotFoundException(format("No job instance was found for '%s/%s/%s/%s/%s'.", pipelineName, pipelineCounter, stageName, stageCounter, jobName));
         }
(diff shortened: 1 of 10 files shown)


chadlwilson:
What do you think @ketan ? Decided to change the existing implementation on `JobInstanceService` as it is only used by this API controller and the feed service API which discards the transitions anyway. Think should be low impact there.

ketan:
Your assessment of impact seems reasonable to me. Are you able to paste an API response here?

chadlwilson:
> Your assessment of impact seems reasonable to me. Are you able to paste an API response here?

Sorry @ketan  - missed the comment:

```
$ curl -H 'Accept: application/vnd.go.cd+json' -s http://admin:badger@localhost:8153/go/api/jobs/up42/1/up42_stage/1/up42_job
```

```json
{
  "name" : "up42_job",
  "state" : "Completed",
  "result" : "Passed",
  "original_job_id" : null,
  "scheduled_date" : 1658243537184,
  "rerun" : false,
  "agent_uuid" : "ccf946ac-a8e6-4c9f-b688-0eca23654e05",
  "pipeline_name" : "up42",
  "pipeline_counter" : 1,
  "stage_name" : "up42_stage",
  "stage_counter" : "1",
  "job_state_transitions" : [ {
    "state" : "Scheduled",
    "state_change_time" : 1658243537184
  }, {
    "state" : "Assigned",
    "state_change_time" : 1658243552110
  }, {
    "state" : "Preparing",
    "state_change_time" : 1658243562162
  }, {
    "state" : "Building",
    "state_change_time" : 1658243576441
  }, {
    "state" : "Completing",
    "state_change_time" : 1658243576700
  }, {
    "state" : "Completed",
    "state_change_time" : 1658243576740
  } ]
}
```

chadlwilson:
The timestamp serialization here is consistent with [stage instances](https://api.gocd.org/current/#get-stage-instance) but I do note that the [agent job run history](https://api.gocd.org/current/#agent-job-run-history) serializes timestamps inconsistently for some reason.

PR #10577, opened 2022-07-02
Replace Velocity with Freemarker for rendering build/job details

chadlwilson:
Replaces the deprecated use of `VelocityView`s in Spring with a `FreeMarkerView`. Support for Velocity was dropped in Spring 5, so this helps unblock Spring upgrades. FreeMarker is already used elsewhere in GoCD (for better or worse) so this at least doesn't introduce a new tech by now.

Allows removing of Commons Lang 3, Commons Digester, Commons BeanUtils.

Original commit and work by Ketan at https://github.com/ketan/gocd/commit/8fe85ed0af9973614ab541c5afcee1ccf5ec139e

**Sanity testing of job/build details page**
* [x] Console tab
* [x] Tests tab
  * [x] Empty?
  * [x] With results? 
* [x] Artifacts tab 
  * [x] Empty
  * [x] With artifacts
* [x] Materials tab
  * [x] Single git material
  * [x] Package material
  * [x] Multiple materials?
* [x] Custom tabs
  * [x] Empty (no artifact)
  * [x] With artifact to render
* [x] Job History Sidebar
* [x] Updating header etc while running
* [x] While anonymous?
* [x] Websockets flag etc?


chadlwilson:
Hi @ketan - thanks for your help with this. Finally got around to brute forcing my way through. Fixed a few bugs and missed bits, and then a few most merge fixes I and the tests missed, but the hard work was mostly already done by you. Looks to be going fine on https://build.gocd.org now.

Even managed to find and fix an old bug in the process - #10838  :-)

ketan:
I'm glad this worked. Now you migrating to a newer spring/hibernate should be a cakewalk. /s

PR #10582, opened 2022-07-05
Remove unused dynamic_form rails helper

chadlwilson:
I believe this is now unused after the remaining last batch of Rails-based editing pieces were removed.


chadlwilson:
Regression tests are green - merging. Should be easy to roll back if there is some hidden usage I can't find....

PR #10625, opened 2022-07-24
Fix issues running Windows commands that have spaces in paths

### LINKED ISSUE #4173: Windows path with spaces for custom command fails
##### Issue Type

- Bug Report

##### Summary

For running msbuild bundled in the microsoft build tools I've been using the following command

```
C:\PROGRA~2\MICROS~1\2017\BuildTools\MSBuild\15.0\Bin\amd64\MSBuild.exe
```

with the following arguments

```
/m
/nr:false
targets.msbuild
/t:BuildAndRunUnitTestsAndRelease
/p:BuildVersion="%GO_PIPELINE_LABEL%"
```

This works fine on the windows server 2012 r2 I've installed the go.cd agent on.

However when switching to a microsoft windows docker container it turns out 8.3 naming is disabled, so the short names no longer work.

I've tried various ways of writing the msbuild path with the spaces escaped or quoted, but none successful.

```
C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\bin\amd64\MSBuild.exe
```

```
""C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\bin\amd64\MSBuild.exe""
```

```
C:\""Program Files (x86)""\""Microsoft Visual Studio""\2017\BuildTools\MSBuild\15.0\bin\amd64\MSBuild.exe
```

```
C:\Program^ Files^ (x86)\Microsoft^ Visual^ Studio\2017\BuildTools\MSBuild\15.0\bin\amd64\MSBuild.exe
```

##### Environment

Go.cd agent 17.12.0 running in `microsoft/dotnet-framework-build:4.7.1-2017.12-windowsservercore-1709` container with openjdk JRE 8.0.152.

It also fails on windows 2012 R2, but there we still have the possibility of using 8.3 notation for the path.

* Go Version: `17.12.0 (5626-cb7df2ffe421e43f2a682a7a323cb3a3e30734cc)`
* JAVA Version: `1.8.0_131`
* OS: `Linux 4.4.86+`

##### Steps to Reproduce

1. Run go.cd agent on Windows 12 R2 or Windows 2016 Server or in microsoft/dotnet-framework-build:4.7.1-2017.12-windowsservercore-1709 windows docker container
2. Install chocolatey package `cinst microsoft-build-tools` version `15.0.26320.2`
3. Run custom command `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\bin\amd64\MSBuild.exe` in any of the escaped forms

##### Expected Results

Command to run equally well as with the 8.3 notation

##### Actual Results

Fails with either

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

```
The input line is too long.
```

or

```
The filename, directory name, or volume label syntax is incorrect.
```

### LINKED ISSUE #10521: Null parameter causes invalid command on Windows
##### Issue Type
- Bug Report

##### Summary

When using a parameter in a template job that is not set in the pipeline, the command path is incorrect.
```'C:\Program' is not recognized as an internal or external command```

This used to work. Last updated about a year ago.

##### Environment

Go Server: Windows 7
Go Agent: Windows 7 and Windows 10

* Go Version: GoCD Version: 22.1.0 (13913-f4c9c1650e2e27fe0a9962faa39536f94f57e297).
* JAVA Version: 17.0.2
* OS: Windows 7 6.1

##### Steps to Reproduce

Create a template that uses a property as an argument.
Create a pipeline using the template and leave the property unset.

I have not tested this in GoCD but trimmed a config file that produces the error.

```
<cruise xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="cruise-config.xsd" schemaVersion="139">
  <pipelines group="TestGroup">
    <pipeline name="TestPipeline" template="TestTemplate">
      <params>
        <param name="Property1">/p:TestValue=Test</param>
        <param name="Property2" />
      </params>
    </pipeline>
  </pipelines>
  <templates>
    <pipeline name="TestTemplate">
      <stage name="TestStage">
        <jobs>
          <job name="TestJob">
            <tasks>
              <exec command="C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin\MSBuild.exe">
                <arg>#{Property1}</arg>
                <arg>#{Property2}</arg>
                <runif status="passed" />
              </exec>
            </tasks>
          </job>
        </jobs>
      </stage>
    </pipeline>
  </templates>
</cruise>
```

##### Expected Results

Command should execute the full path correctly.

##### Actual Results

```
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

##### Possible Fix

Set a dummy property value that does not affect the actual command

```<param name="Property2">/p:Dummy=Dummy</param>```

### PR DESCRIPTION
**Pull request by chadlwilson**
- fixes #4173, #2153
- fixes #10521 

This change resolves issues repeatedly reported over the years on Windows, especially _when running executables that have spaces in the path_.

It does this by correcting the way things are quoted for use of `cmd`. GoCD runs everything via `cmd /c`, presumably to allow "shell" built-ins such as `dir` to be run.

Unfortunately the current strategy doesn't quote things correctly, so if the executable path has a space in it things will not work correctly. This change corrects the quoting by adding an extra pair of quotes around the args per https://superuser.com/questions/238810/problem-with-quotes-around-file-names-in-windows-command-shell which should allow things to run as expected.

_Relevant section from cmd /?_:
```
If /C or /K is specified, then the remainder of the command line after
the switch is processed as a command line, where the following logic is
used to process quote (") characters:

    1.  If all of the following conditions are met, then quote characters
        on the command line are preserved:

        - no /S switch
        - exactly two quote characters
        - no special characters between the two quote characters,
          where special is one of: &<>()@^|
        - there are one or more whitespace characters between the
          two quote characters
        - the string between the two quote characters is the name
          of an executable file.

    2.  Otherwise, old behavior is to see if the first character is
        a quote character and if so, strip the leading character and
        remove the last quote character on the command line, preserving
        any text after the last quote character.
```

This change does two things
- Adds the `/S` switch to ensure that the logic in the second clause is applied, regardless of the number of quotes or special chars.
- Surrounds the args with `" {argsList} "` to allow first and last quote characters to be stripped. Other quotes within will be retained as-is.

New behaviour is enabled by default but allows fallback to the old behaviour by defining the system property on server of `-Dgocd.agent.extra.properties=toggle.agent.windows.command.quote.all.args=N`. The intent is to remove this toggle if no issues are reported.

**Example commands**

1. Executable with spaces in it (previously ❌)
2. Arg non-quoted with spaces (previously ✅)
3. Arg _pre_-quoted with spaces (previously ✅)
4. cmd in-built multiple args with spaces in them (previously ✅)

image

**BEFORE**:
image

**AFTER**:
_Note_: quotes in below are inserted by the "display" logic for commands.
image

### DIFF
--- a/commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
+++ b/commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
@@ -24,7 +24,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
-class ScriptRunnerTest {
+class CommandLineScriptRunnerTest {
     @Test
     @EnabledOnOs(OS.LINUX)
     void shouldReplaceSecretsOnTheOutputUnderLinux() throws CheckedCommandLineException {
@@ -79,7 +79,7 @@ class ScriptRunnerTest {
 
     @Test
     @DisabledOnOs(OS.WINDOWS)
-    void shouldMaskOutOccuranceOfSecureEnvironmentVariablesValuesInTheScriptOutputOnLinux() throws CheckedCommandLineException {
+    void shouldMaskOutOccurrenceOfSecureEnvironmentVariablesValuesInTheScriptOutputOnLinux() throws CheckedCommandLineException {
         EnvironmentVariableContext environmentVariableContext = new EnvironmentVariableContext();
         environmentVariableContext.setProperty("secret", "the_secret_password", true);
         CommandLine command = CommandLine.createCommandLine("echo").withArg("the_secret_password").withEncoding(UTF_8);
@@ -93,7 +93,7 @@ class ScriptRunnerTest {
 
     @Test
     @EnabledOnOs(OS.WINDOWS)
-    void shouldMaskOutOccuranceOfSecureEnvironmentVariablesValuesInTheScriptOutput() throws CheckedCommandLineException {
+    void shouldMaskOutOccurrenceOfSecureEnvironmentVariablesValuesInTheScriptOutput() throws CheckedCommandLineException {
         EnvironmentVariableContext environmentVariableContext = new EnvironmentVariableContext();
         environmentVariableContext.setProperty("secret", "the_secret_password", true);
         CommandLine command = CommandLine.createCommandLine("cmd")
--- a/common/src/main/java/com/thoughtworks/go/domain/builder/BaseCommandBuilder.java
+++ b/common/src/main/java/com/thoughtworks/go/domain/builder/BaseCommandBuilder.java
@@ -21,13 +21,18 @@ import com.thoughtworks.go.plugin.access.pluggabletask.TaskExtension;
 import com.thoughtworks.go.plugin.infra.PluginRequestProcessorRegistry;
 import com.thoughtworks.go.util.command.*;
 import com.thoughtworks.go.work.DefaultGoPublisher;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.SystemUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.List;
 
 public abstract class BaseCommandBuilder extends Builder {
+    static final String QUOTE_ALL_WINDOWS_ARGS = "toggle.agent.windows.command.quote.all.args";
     private static final Logger LOG = LoggerFactory.getLogger(BaseCommandBuilder.class);
 
     protected String command;
@@ -83,7 +88,22 @@ public abstract class BaseCommandBuilder extends Builder {
         }
     }
 
-    protected abstract CommandLine buildCommandLine();
+    protected CommandLine buildCommandLine() {
+        if (SystemUtils.IS_OS_WINDOWS) {
+            return CommandLine.createCommandLine("cmd")
+                    .withWorkingDir(workingDir)
+                    .withArgs(windowsCmdArgsStart())
+                    .withArg(translateToWindowsPath(this.command))
+                    .withArgs(argList())
+                    .withArgs(windowsCmdArgsEnd());
+        } else {
+            return CommandLine.createCommandLine(this.command)
+                    .withWorkingDir(workingDir)
+                    .withArgs(argList());
+        }
+    }
+
+    protected abstract String[] argList();
 
     private void setTaskError(String errorMessage) {
         LOG.warn(errorMessage);
@@ -100,5 +120,25 @@ public abstract class BaseCommandBuilder extends Builder {
                 ", workingDir=" + workingDir +
                 ", errorString='" + errorString + '\'' +
                 "} " + super.toString();
+    }
+
+    private List<String> windowsCmdArgsStart() {
+        return shouldQuoteAllWindowsArgs()
+                ? List.of("/s", "/c", "\"")
+                : List.of("/c");
+    }
+
+    private List<String> windowsCmdArgsEnd() {
+        return shouldQuoteAllWindowsArgs()
+                ? List.of("\"")
+                : Collections.emptyList();
+    }
+
+    private boolean shouldQuoteAllWindowsArgs() {
+        return "Y".equalsIgnoreCase(System.getProperty(QUOTE_ALL_WINDOWS_ARGS, "Y"));
+    }
+
+    private String translateToWindowsPath(String command) {
+        return StringUtils.replace(command, "/", "\\");
     }
 }
--- a/common/src/main/java/com/thoughtworks/go/domain/builder/CommandBuilder.java
+++ b/common/src/main/java/com/thoughtworks/go/domain/builder/CommandBuilder.java
@@ -15,12 +15,10 @@
  */
 package com.thoughtworks.go.domain.builder;
 
-import java.io.File;
-
 import com.thoughtworks.go.domain.RunIfConfigs;
 import com.thoughtworks.go.util.command.CommandLine;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.SystemUtils;
+
+import java.io.File;
 
 public class CommandBuilder extends BaseCommandBuilder {
     protected String args;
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by chadlwilson** [comment]
Appreciate any comments from @ketan and anyone who might have background here.

**Comment by ketan** [comment]
No background. Your presumption around `cmd /c` is correct. I cannot find the issue, and it might as well had been a conversation on gitter — I was tempted to wrap everything up in `bash -c` for sake of consistency, but that might have broken way too much stuff.

If your tests pass, with the examples given in the issues you linked to — I suppose the change should be good. Looking at the tests, nothing stands out as odd to me.

**Comment by chadlwilson** [comment]
Thanks @ketan ! It's surprising that this issue seems to have persisted over the years, that is all. And couldn't really find any/many narrow integration tests that validated any of the "can actually run tasks in these forked processes". I guess the code here is rather old, like Cruise old though?

I'll have additional confidence once it's merged and GoCD's own pipelines run through on Windows but seems like it will be fine to me.

**Comment by ketan** [comment]
Most of the core of GoCD is old "Cruise". Not much has changed in the core.

I think the feature toggle should help folks roll back, if needed. However I noticed that the feature toggle needs to be enabled on each agent, which is *really* difficult to do across several windows VM. I recall doing something to send feature toggles from server over to agent, let me see if I can look that up.

**Comment by ketan** [comment]
My understanding about feature toggle seems incorrect. The server *I think* seems to serialize the task as a `CommandLine` instance that represents the final command that should be run.

**Comment by chadlwilson** [comment]
Yeah, you can enable it in each agent directly but the `gocd.agent.extra.properties` on server side does propagate to all agents via the launcher/bootstrapper. I validated this empirically.

**Comment by chadlwilson** [comment]
It goes via a response header when the agent binary is downloaded and eventually gets used at the below when the agent is launched.

https://github.com/gocd/gocd/blob/5638a49a35e5a96836f8870deae69a8e9c8f1093/agent-process-launcher/src/main/java/com/thoughtworks/go/agent/AgentProcessParentImpl.java#L66-L68

**Comment by arvindsv** [comment]
The second clause seemed worrying to me:

> Otherwise, old behavior is to see if the first character is a quote character and if so, strip the leading character and remove the last quote character on the command line, preserving any text after the last quote character.

The "remove the last quote character on the command line" seemed wrong. But, I see you've checked all the cases I could think of, and it's working. There is also a feature toggle which can be used from the server-side (along with an agent restart). That's the most amount of care that can be taken. So, I'm happy with this change.

**Comment by arvindsv** [comment]
Maybe check with "C:\Program\ Files\..." (backslash-space)?

**Comment by arvindsv** [comment]
Apparently, caret (^) and grave (`) can be used as escape characters instead of backslash, on Windows! TIL.

**Comment by chadlwilson** [comment]
> The second clause seemed worrying to me:
> 
> > Otherwise, old behavior is to see if the first character is a quote character and if so, strip the leading character and remove the last quote character on the command line, preserving any text after the last quote character.
> 
> The "remove the last quote character on the command line" seemed wrong. But, I see you've checked all the cases I could think of, and it's working. There is also a feature toggle which can be used from the server-side (along with an agent restart). That's the most amount of care that can be taken. So, I'm happy with this change.

What they are saying, basically is that if you give it a command line with

```
cmd /c /s " "C:\temp dir\something.exe" "hello world" "
```
It will see look at the start args, see it is a `"` and then remove that and the last `"`, then the remaining args will go to the underlying command/executable.

By counter-example, what is currently happening is that GoCD executes something like:
```
cmd /c "C:\temp dir\something.exe" "hello world"
```
what happens is that the 1st clause fails due to not having `exactly two quote characters`. So then it actually executes `C:\temp dir\something.exe" "hello world` (stripping first and last quote). That will then break with `C:\temp is not recognised as a valid internal or external command`.

At root, the problem is that clause 1 has two many crazy requirements. The only predictable behaviour across varieties of user-defined args seems to be to ensure that clause 2 is what is always executed. But then we need to force this with `/s` and compensate for the "strip the leading character and remove the last quote character" by prefixing and suffixing.

Overall when interpreting the args to `cmd` it seems to actually concatenate all the args together into one string when interpreting things and then re-parse on top of that. It's bizarre the way it's implemented, but I think if you ensure the command starts with a (double-)quote, and ends with a (double-)quote it will generally do what you expect with regard to the stuff between.

**Comment by arvindsv** [comment]
Oh wow! Thank you for that explanation. I had totally misunderstood that. Coming from a bash/sh background, this just feels insane to me. :)

> cmd /c /s " "C:\temp dir\something.exe" "hello world" "

Un-escaped quotes!

**Comment by chadlwilson** [comment]
> Oh wow! Thank you for that explanation. I had totally misunderstood that. Coming from a bash/sh background, this just feels insane to me. :)
> 
> > cmd /c /s " "C:\temp dir\something.exe" "hello world" "
> 
> Un-escaped quotes!

Yeah, it's bizarre to me too. It's also very hard to understand the way it is documented. I'm still not fully sure I understand _all_ of the behaviour myself, especially when putting things through the Java `exec` process. I mainly was trying things out manually on command line myself to try and make sense of things after reading a few of these replies on StackOverflow and elsewhere which seemed insane.

You did remind me that in #4173 the reporter tried caret-escaping spaces (`^ `) and that didn't work either, so want to dig a bit there to make sure I understand why that also didn't work, and understand what would happen now. However it [looks bizarre](https://www.howtogeek.com/694949/how-to-escape-spaces-in-file-paths-on-the-windows-command-line/).
> In the Command Prompt, the caret character (^ ) will let you escape spaces—in theory. ... Here’s the problem: While this should work, and it does sometimes, it doesn’t work all the time. The Command Prompt’s handling of this character is strange.

🤦‍♂️ I suspect it works inside a `.bat` file more consistently that it does for commands on the command line. All this nonsense is why there was some need for Powershell, I suppose.

e.g
```
C:\temp>attrib "C:\Program Files"
     R               C:\Program Files

C:\temp>attrib C:\Program^ Files
Parameter format not correct -

c:\temp>"c:\temp\Directory With Spaces\attrib.exe" "C:\Program Files"
     R               C:\Program Files
     
c:\temp>c:\temp\Directory^ With^ Spaces\attrib.exe "C:\Program Files"
Parameter format not correct -

c:\temp>c:\temp\"Directory With Spaces"\attrib.exe "C:\Program Files"
     R               C:\Program Files

c:\temp>c:\temp\Directory^ With^ Spaces\attrib.exe C:\temp
Parameter format not correct -
```

This is where I give up with carets or assuming anyone is relying on them in their GoCD tasks :-) As far as I can tell double-quoting is the only even quasi-reliable way to deal with spaces. But have no idea what kind of crazy workarounds people might have had to employ, hence the feature flag for now.

I also wanted to check any unintended effect to other special characters or attempts to escape them as well, but it's such a mystery as things just don't work like you might expect, especially across built-ins and various other commands. There's just no consistency.

**Additional weirdness**
There are certainly still some GoCD WTFs caused by the way it does things. e.g it will double-quote arguments on your behalf if there are spaces in them. It feels like these two should behave the same (with the latter just having a space in the path this is `attrib`ed:

image

But they don't. The space being there causes _something_ to seemingly put an extra set of double-quotes around it and the `^` no longer act as an escape as it seems to be pre-escaped:

image

This makes it actually behave like indicated in the display with the extra double quotes - but I thought the double quotes being displayed there was just _for display_, as the code has a distinction between `commandForDisplay` and the raw command `argsList` and I cannot see any such magic quoting in the code. It's possible the Windows `ProcessBuilder` does that behind the scenes).

https://github.com/gocd/gocd/blob/ba4d98fdd7779e901c1adbdd7269545fcc5e5abe/commandline/src/main/java/com/thoughtworks/go/util/ProcessManager.java#L47-L63

Anyway, the change in this PR doesn't make anything worse, nor change the behaviour in these screenshotted scenarios - just makes it difficult to test and understand what _should_ happen.

chadlwilson:
In any case, I'll merge this and see how things go on build.gocd.org, it seems we are OK with this given my testing. Will re-address anything else (or rollback) if additional thoughts come up.

arvindsv:
> This is where I give up with carets or assuming anyone is relying on them in their GoCD tasks :-) 

That's a fair assumption. There's always the workaround of asking them to use a `.bat` file instead.

> It's possible the Windows ProcessBuilder does that behind the scenes).

Could be. I think we assumed Java and ProcessBuilder would do the "right" thing, and treated it like a normal Unix process. But, your post shows that it wasn't the right assumption.

PR #10679, opened 2022-08-08
Ensure errors for FetchPluggableArtifactTasks are available after validation

### LINKED ISSUE #10668: Fetch External Artifact task gives generic Unprocessable Entity error for failing validations
##### Issue Type
- Bug Report

##### Summary
Validation failures for ArtifactStoreConfig entries produce a nice GUI representation and indicate the location and error message for the failure(s).
Validation failures for FetchArtifactConfig entries produce unprocessable entity responses and a less pleasing GUI/feedback.

Although I encountered this whilst writing a custom plugin, the same can be seen using the [Docker artifact plugin](https://github.com/gocd/docker-registry-artifact-plugin).

Further discussion and some troubleshooting can be found at discussion https://github.com/gocd/gocd/discussions/10639

* Go Version: `22.1.0`
* JAVA Version: `17.0.2`
* OS: `Windows 10`
* Browser vendor and version (if relevant): `Chrome 103.0`

##### Steps to Reproduce
See https://github.com/gocd/gocd/discussions/10639 for details

1. Download and install the [Docker artifact plugin]((https://github.com/gocd/docker-registry-artifact-plugin). (releases page, `1.3.1-270` was the version tested)
2. Create a valid(atable) Docker artifact store (no need to have a working store, junk values for `RegistryURL`, `Username` and `Password` are sufficient when choosing `other` for `RegistryType`)
3. Create a new pipeline with two stages
4. In the first stage, create an External artifact for the store created in step 2, provide some value for `id` and `Image` (test-docker-image and 'gocd/image' chosen here)
5. In the second stage, create a FetchArtifact task, choose `External` for type and then the first stage for the source and cd.go.artifact.docker.registry for the Plugin Id (might get a warning about not being able to determine the plugin, I think this happens because not built previously?)
6. (Optionally) Click save - configuration is valid and saves.
7. Set the `Environment Variable Prefix` to "1" (validation requires it not start with a number)
8. Click save - observe the error shown in Actual Results section

##### Expected Results
Feedback similar to this (this is for an invalid artifact store config, which displays well):
image

##### Actual Results
image

##### Possible Fix
Unknown

##### Log snippets
Some logs are given in #10639

Some code snippets are given in #10639 but mainly indicate that the output of the validation request is (as far as I can tell) of the structure specified by https://plugin-api.gocd.org/current/artifacts/#validate-fetch-artifact-config

chadlwilson:
Fixes #10668 

Currently the errors are not returned in the response via the API, nor displayed against the relevant fields on the UI. This addresses that. I don't really understand why, but for everything except the `PluggableTask`s it seems necessary to copy over errors from `preProcessedConfig` to the actual config being updated in:

https://github.com/gocd/gocd/blob/5c7c6c226dc2caa1fe729f5142536634912e795a/server/src/main/java/com/thoughtworks/go/config/update/UpdatePipelineConfigCommand.java#L67-L78

Seems this requires some dodgy `instanceof` stuff, already done for `ExecTask`, otherwise the errors disappear into the ether.

A response on fetch task edit via `PUT http://localhost:8153/go/api/admin/pipelines/up42` with the **Docker Registry Artifact Plugin**.

```json
{
  "message" : "Validations failed for pipeline 'up42'. Error(s): [Validation failed.]. Please correct and resubmit.",
  "data" : {
// ... stuff
       {
          "type" : "fetch",
          "attributes" : {
            "artifact_origin" : "external",
            "pipeline" : "up42",
            "stage" : "up42_stage",
            "job" : "up42_job",
            "run_if" : [ "passed" ],
            "artifact_id" : "test-thingy",
            "configuration" : [ {
              "key" : "EnvironmentVariablePrefix",
              "value" : "1",
              "errors" : {
                "EnvironmentVariablePrefix" : [ "Invalid environment name prefix. Valid prefixes contain characters, numbers, and underscore; and can't start with a number." ]
              }
            }, {
              "key" : "SkipImagePulling"
            } ]
          }
        } 
// ... stuff
```

Afterwards:
image

image

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/config/update/PipelineConfigErrorCopier.java
+++ b/server/src/main/java/com/thoughtworks/go/config/update/PipelineConfigErrorCopier.java
@@ -69,6 +69,9 @@ public class PipelineConfigErrorCopier {
                     if (toTask instanceof ExecTask) {
                         copyCollectionErrors(((ExecTask) fromTask).getArgList(), ((ExecTask) toTask).getArgList());
                     }
+                    if (toTask instanceof FetchPluggableArtifactTask) {
+                        copyCollectionErrors(((FetchPluggableArtifactTask) fromTask).getConfiguration(), ((FetchPluggableArtifactTask) toTask).getConfiguration());
+                    }
                 }
                 List<PluggableArtifactConfig> toPluggableArtifactConfigs = toJob.artifactTypeConfigs().getPluggableArtifactConfigs();
                 List<PluggableArtifactConfig> fromPluggableArtifactConfigs = fromJob.artifactTypeConfigs().getPluggableArtifactConfigs();
--- a/server/src/main/webapp/WEB-INF/rails/webpack/models/mixins/errors.ts
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/models/mixins/errors.ts
@@ -72,11 +72,11 @@ export class Errors {
   }
 
   errorsForDisplay(attrName: string) {
-    return _.map(this._errors[attrName] || [], s.terminateWithPeriod).join(" ");
+    return _.uniq(_.map(this._errors[attrName] || [], s.terminateWithPeriod)).join(" ");
   }
 
   allErrorsForDisplay(): string[] {
-    return _.map(this.keys(), (key) => this.errorsForDisplay(key));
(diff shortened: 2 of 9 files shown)


chadlwilson:
Testing notes on #10668.

PR #10699, opened 2022-08-15
Upgrade from Rails 5.2 to Rails 6.1

chadlwilson:
**Methodology**
Followed
* https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0
* https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1
* Run `./gradlew railsExec -Pcmd='-S rails zeitwerk:check'` and review

**Testing**
* [x] Build
  * [x] Rspecs passing
  * [x] Regression tests all passing
* [x] Local dev experience validation
  * [x] Error during start. (Not quite sure what happened here, but this went away. Maybe a build cleanliness problem)
  ```
    2022-08-15 23:54:40,682 ERROR [qtp973576304-28] Rails:-2 -
  NoMethodError (undefined method `+' for nil:NilClass):
  
  lib/log4j_logger.rb:21:in `silence'
  ```
  * [x] Logging for rails Slf4j working fine
  * [x] Dynamic reloading working fine/as normal
  * [x] Stack traces coming out on local while debugging 
  * [x] 500 errors from Rack are not being logged anywhere by default
    * Fixed in Rack 2.2.4 which came out recently, funnily enough, not related to Rails upgrade, just made it difficult initially to debug the below 500 error caused by Rails/Jetty interaction
  * [x] 500 errors for all pages in local dev being caused by excessive `Links` response header being added by Rails 6.1. Disabled in dev, as per general recommendations.

* [x] The *Great, Paranoid Brute force manual Ruby/Rails testing* list
   * [x] Overall rendering incl headers/footers etc
   * [x] Rails Stage Details page and all tabs
     * [x] Rails Stage Details page > Config Changed Stage History modal (admin function)
     * [x] Rails Stage Details page > "Run 1 of 2" run switcher dropdown
   * [x] Rails Job/Build Details page and all tabs
     * [x] Rails Job/Build Details page > Job History dropdown
   * [x] Rails Value Stream Map page
     * [x]  Rails Value Stream Map page > general Javascript navigation etc
   * [x] Legacy Admin pages
     * [x] Admin > Config XML (as admin) 
     * [x] Admin > Config XML (as group admin only)
   * [x] Review browser console logs for any unexpected errors from JS

PR #10703, opened 2022-08-17
Make preparatory forward-compatible changes for Rails 6

chadlwilson:
- Resolves behaviour deprecated in Rails 6.1 regarding template names with `.` in them (makes things consistent across the codebase)
- Avoid naming specs starting with `/` which seems to confuse rspec-rails 6 (or rails itself)
- Fix some minor technical debt


chadlwilson:
Main upgrade will be undertaken via https://github.com/gocd/gocd/pull/10699

PR #10734, opened 2022-08-22
Migrate from sass-rails 5 to sassc-rails 2.1

chadlwilson:
sass-rails is only a wrapper around sassc-rails now.

See https://github.com/sass/sassc-rails


chadlwilson:
The differences in generated CSS seem minimal to me after comparing so this looks fine to me.

PR #10762, opened 2022-09-03
Exclude assets pipeline gems from production installers

chadlwilson:
**Motivation**
Currently the packaged zips and installers include Gems that are not needed in production. It would better to remove these for a few reasons
- security attack surface area. If you're able to break he first layer and run arbitrary stuff inside the container, the less code/libs available the better.
- reduce size of container images and installers
- we are currently reliant on the Rails asset pipeline to compile some SASS/SCSS. `sass-rails` is EOL, and we are supposed to move to `sassc-rails` based on a native impl in #10734. If we do nothing, including this native gem will include an extra 40MB of random binaries and such in production.

**Background**
I'm no Rails historian, but I infer that in earlier Rails world, it seems there was a dedicated `assets` group, but this was removed as it caused some problems. This re-introduces the concept here in order to be able to run without these Gems during production. Don't fully understand how this all works, and best practice but...

**Changes**
- GoCD doesn't do a secondary `bundle install` after compiling assets. It just uses some random code to exclude non-default group Gems from the production war file. So from bundler's perspective they are still there.
- Thus we need to override during `boot.rb` to tell it to only setup the `default` gems, not everything implied by the `Gemfile` and `.bundle/config`, which has no `BUNDLE_WITHOUT` set. I use `RAILS_ENV=production` and `!RAILS_GROUPS` to determine this
- During asset compilation, tests, local dev `RAILS_GROUPS` is set to tell Rails to "add in" the assets pipelines gems to be able to compile on the fly.
- During `init.rb` some extra magic is required to require the rails groups, that I dont fully understand.
- We need to disable the `jsroutes.rb` initializer when running in production. Not sure if this should be an initializer but this seems to work fine.

These ideas came from https://nts.strzibny.name/removing-nodejs-dependency-from-a-rails-application/ and https://stackoverflow.com/a/39758749 along with a lot of trial and error. Some of the complications come from the JRuby-Rack based bootstrapping process, which seemingly works differently to other regular Rails apps in terms of env var visibility as well as recognition/ordering of `boot.rb`, `init.rb`, `application.rb` etc.

**Testing**
- [x] All tests run, etc etc
- [x] Can still do local dev just fine after `./gradlew prepare` including dynamically compiling assets
- [x] Diffed the zips before and after to make sure only the expected differences are discovered 

**Differences**

image

image

### DIFF
--- a/server/rails.gradle
+++ b/server/rails.gradle
@@ -44,6 +44,7 @@ task generateJSRoutes(type: ExecuteUnderRailsTask) {
 
   environment(
     'RAILS_ENV': 'production',
+    'RAILS_GROUPS': 'assets',
     'OUTPUT_DIR': outputDir,
   )
 
@@ -128,7 +129,6 @@ task initializeRailsGems {
   }
 }
 
-
 task yarnInstall(type: YarnInstallTask) {
   // delete this in the compile phase, because otherwise the yarnInstall detects a new folder and reinstalls.
   project.delete(project.file("${project.railsRoot}/node_modules/.cache"))
@@ -209,6 +209,7 @@ task compileAssetsRailsTest(type: ExecuteUnderRailsTask) {
 
   environment += [
     'RAILS_ENV' : 'test',
+    'RAILS_GROUPS': 'assets',
   ]
 
   args = ['-S', 'rake', '--trace', 'assets:clobber', 'assets:precompile']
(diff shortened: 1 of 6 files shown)


chadlwilson:
Hi @ketan and @arvindsv  - does this seem reasonable and correct to you folks? Anything I am missing?

ketan:
Did a quick look. Seems reasonable, no need to package those gems for use in production, esp because we have already compiled all assets.

chadlwilson:
Thanks a lot @ketan - much appreciated. A quick look is fine - pretty confident in the regression tests and my empirical testing as it took a while to get to a working combination across the various Ruby environments and combinations.

Ruby/Rails asset stuff and interactions with Bundler, rack etc are still a bit of a mystery to me, so not sure how canonical my approach is. 😅

PR #10776, opened 2022-09-05
Migrate from sass-rails to sassc-rails native implementation

chadlwilson:
Recreating #10734 to ensure it can be run on Windows.

sass-rails is only a wrapper around sassc-rails now - see https://github.com/sass/sassc-rails


chadlwilson:
Given the difficulties here on Windows, and libsass/sassc being deprecated in favour of Dart Sass, may look at migrating straight to https://github.com/rails/dartsass-rails instead. Will need to move the Webpack world to Dart Sass at some point too anyway.

chadlwilson:
Will likely abandon libsass and replace this with https://github.com/gocd/gocd/pull/10794 and https://github.com/ntkme/sassc-embedded-shim-ruby and dartsass-rails is not well updated, nor conveniently integrated with the assets pipeline, sprockets etc for development mode.

PR #10794, opened 2022-09-09
Switch to sassc-rails with sass-embedded

chadlwilson:
Uses dart-sass underneath and integrates properly with the assets pipeline and sprockets, as opposed to dartsass-rails which seems to have problems.


chadlwilson:
Build is currently failing on Windows due to https://github.com/ntkme/sass-embedded-host-ruby/issues/62 which I am trying to workaround on our build agents with https://github.com/gocd-contrib/gocd-oss-cookbooks/commit/bce0a9d9f4899e4b0a9d30c154a0145f2e190947

PR #10804, opened 2022-09-11
Upgrade Sprockets from v3 to v4

chadlwilson:
Sprockets 4 introduces need for mandatory app/assets/config/manifest.js (which is not actually JS) that replaces the old assets.precompile block. By default everything in the root of the load paths (app/assets/stylesheets etc) seems to be treated as an entry point which is not what we want, so continue to list out the individual entry points we need.

https://github.com/rails/sprockets/blob/main/UPGRADING.md

This
- removes a whole lot of dead entrypoints for pieces migrated fully to webpack land
- changes image loading to not just include ANY image in the load path (it's not possible like it was before anyway, it seems) and only include those referenced. This doesn't seem to have any effect.
- only includes woff/woff2 versions of fonts based on their inclusion. The others aren't required anyway it seems?

_Useful details_
- https://github.com/rails/sprockets/blob/main/UPGRADING.md
- https://www.rubydoc.info/gems/sprockets/4.1.1#link 
- https://bibwild.wordpress.com/2019/10/15/sprockets-4-and-your-rails-app/
- https://eileencodes.com/posts/the-sprockets-4-manifest/


chadlwilson:
Everything appears to be working fine, and the zip/war diff is as expected. Much cleaner and is ~5MB smaller compressed, and about ~10MB smaller uncompressed due to removing unnecessarily included assets duplicated between compiled and raw/uncompiled versions, unnecessary image inclusions from `node_modules` and unnecessary font variant inclusions.

PR #10888, opened 2022-10-03
Publish MacOS aarch64 installer builds

### LINKED ISSUE #10887: Provide arm64 Mac installers/zips rather than relying on Rosetta for Apple Silicon
##### Issue Type

Enhancement

##### Summary

I downloaded the latest stable GoCD release (22.2.0), and all commands executed in pipelines are running via Rosetta emulation on an Apple Silicon Mac.

Since the `java` binary in the embedded JRE is only compiled for Intel, all child processes it spawns also run as Intel.

##### Environment

Server:
https://download.gocd.org/binaries/22.2.0-14697/osx/go-server-22.2.0-14697-osx.zip

Agent:
https://download.gocd.org/binaries/22.2.0-14697/osx/go-agent-22.2.0-14697-osx.zip

Problematic binary:
```plain
↪ file jre/Contents/Home/bin/java 
jre/Contents/Home/bin/java: Mach-O 64-bit executable x86_64
```

Some binaries in the package are compiled for arm64, though, which is good.
```plain
↪ file wrapper/wrapper-macosx-arm-64
wrapper/wrapper-macosx-arm-64: Mach-O 64-bit executable arm64
```

chadlwilson:
Issue: fixes #10887

**Changes**
- Produces aarch6 installer for macOS/OSX including the Adoptium Eclipse Temurin jre built for that platform.
- Includes Tanuki wrappers only appropriate for each architecture's build
- Generates metadata for every installer type including the architecture, as well as whether a JRE is included (idea being to drive downloads etc etc based on this info)

**Tasks**
- [x] Sanity check it works on an Apple Silicon device (see https://github.com/gocd/gocd/pull/10888#issuecomment-1272237538)
- [x] validate downstream impact for codesigning + publishing
  - Needs https://github.com/gocd/codesigning/pull/58 which should be backward compatible.
- [x] Figure out what to do with downloads on www.gocd.org
  - Will need some changes to allow downloading these versions (similar to how we used to allow 32bit and 64bir windows version downloads independently)

**To do after merge & build**
- [ ] Allow the downloads to work, once they are being published for experimental builds
- [ ] Check other docs that reference installing for MacOS
- [ ] test-drive
  - produce arm64 test-drive `run-gocd` binaries or suggest everyone use x64 for test-drive and then switch archs when launching?


ketan:
I have an apple silicon mac, so can help test if needed.

chadlwilson:
Oh, cool - thanks @ketan !

In lieu of this, if you have a minute to try whether the test Drive `curl -fsSL 'https://www.gocd.org/test-drive-gocd/try.sh' | bash -s 'https://download.gocd.org/test-drive/installers/22.2.0/14697/gocd-22.2.0-14697-1779-osx.zip'` works now from a regular aarch64 terminal process via Rosetta that'd be useful. Just a temporary thing fixed [here](https://github.com/gocd/www.go.cd/commit/fd7b04be915fad5c1ffaee3ddff2019d37c557fe) until can get `./run-gocd` built for Apple Silicon too.

There are (unsigned) builds of this PR available below if you'd like to try. I don't see any reason they _wouldn't_ work :-)

https://build.gocd.org/go/files/installers-pull_10888/2/dist/2/dist/dist/osx/go-server-22.3.0-15099-osx-aarch64.zip
https://build.gocd.org/go/files/installers-pull_10888/2/dist/2/dist/dist/osx/go-agent-22.3.0-15099-osx-aarch64.zip

ketan:
I have an apple silicon mac, so can help test if needed.

> Oh, cool - thanks @ketan !
> 
> There are (unsigned) builds of this PR available below if you'd like to try. I don't see any reason they _wouldn't_ work :-)
> 
> https://build.gocd.org/go/files/installers-pull_10888/2/dist/2/dist/dist/osx/go-server-22.3.0-15099-osx-aarch64.zip
> https://build.gocd.org/go/files/installers-pull_10888/2/dist/2/dist/dist/osx/go-agent-22.3.0-15099-osx-aarch64.zip

These binaries work fine on apple silicon. The GoCD UI shows up in the browser and the agent shows up as registered.

chadlwilson:
Great, thank you @ketan! I'll continue with figuring out some of the signing, packaging and downloads stuff to move this forward.

chadlwilson:
@ketan if you'd like to have a go at a sanity check of "real" builds, there are some up now on https://www.gocd.org/download/#osx (experimental tab)

Unfortunately, as was the case before for x64 I've noticed that one still has to accept security warnings at OS level, for the `java` processes in particular, I guess because the binaries are not signed/notarised by Apple. I believe the Tanuki wrapper bins are signed and notarised though.

PR #10892, opened 2022-10-04
Don't crop pipeline instance labels on pipeline activity page

### LINKED ISSUE #10891: Build ids are truncated in pipeline activity page
##### Issue Type

- Bug Report

##### Summary

Build ids are truncated when over 17 characters on the pipeline activity page.

Build id on VSM page: 
image

Build id on pipeline activity page:
image

* Go Version: `21.4.0 (13469-f3b865c5018dec2bddf98496e0e74e8e70a80cb6)`
* JAVA Version: `15.0.2`
* OS: `amzn2`

##### Steps to Reproduce

1.  Have a build id over 17 characters 
2. Navigate to the pipeline activity page

##### Expected Results

The full build id should be present

##### Actual Results

Only the first 17 characters of the build id are shown 

##### Possible Fix

Code here is hcanged 
https://github.com/gocd/gocd/blob/master/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/pipeline_run_info_widget.tsx#L83

chadlwilson:
- fixes #10891

Not sure why this was being done earlier at all, but let's try without it. Should only have an effect when people are using custom labels for pipeline instances rather than normal counters.

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/dashboard/stage_overview/index.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/dashboard/stage_overview/index.tsx
@@ -94,30 +94,16 @@ export class StageOverview extends MithrilComponent<Attrs, State> {
     classNames.split(" ").filter(c => !!c).forEach(c => {vnode.dom.classList.add(c);});
 
     if (vnode.attrs.isDisplayedOnPipelineActivityPage) {
-      let top = 36;
-      let left = -6;
-
-      // for a user with no operate permission, the add comment feature is not available, making the stage overview mis-positioned,
-      // hence, position stage overview a little above for read only users.
-      if (!vnode.attrs.stageInstanceFromDashboard.canOperate) {
-        top = 27;
-      }
-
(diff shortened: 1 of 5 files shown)


arvindsv:
I recall something about the (dynamic) layout going berserk if one of the pipeline boxes was bigger than expected. Maybe something to do with that? But yeah, the code is _very_ specific.

chadlwilson:
Most of the existing rails pages go bezerk with long labels right now anyway 😅 I have a fix locally that looks fine to wrap the labels, will screenshot later.

chadlwilson:
Here's what it looks like now with some ludicrous labels :-)

There's some screwiness with the position of the stage tooltip popup though which seems to assume no overflow, and I need to check with lots of stages, so it's not quite right here.

image

Example of other pages things are still broken (and why this change is unlikely to make things worse overall)

image
image
image
image

arvindsv:
Ok. :) Then it's alright. My main worry was the dashboard.

chadlwilson:
Seems to be working fine now with these tweaks :-)

image

Issue #10895, opened 2022-10-05
Various problems with cruise-config encoding on Windows


cinex-ru:
##### Issue Type

- Bug Report

##### Summary

Various problems with cruise-config encoding

###### Basic environment details

```plain
"java.runtime.version": "17.0.4+8"
"os.arch": "amd64",
"os.name": "Windows 10",
"os.version": "10.0",
"Go Server Version": "22.2.0 (14697-4bdda4e0d769e66da651926c7066979740bd7ae7)"
"browser": "Chrome 105.0.5195.127"
```

##### Steps to Reproduce

Updating pipeline:
1. Add Custom Command task
2. Edit command to '開始'
3. Save task
4. Check cruise-config.xml tasks section

Editing config xml Part One
1. Go Admin / Config XML
2. Press Edit button
3. Edit command in tasks section
``` xml
              <exec command="開始">
                <arg />
                <runif status="passed" />
              </exec>
```
4. Try to save
5. Check new command

Editing config xml Part Two
1. Go Admin / Config XML
2. Press Edit button
3. Edit command in tasks section
``` xml
              <exec command="Команда">
                <arg />
                <runif status="passed" />
              </exec>
```
4. Try to save

chadlwilson:
I haven't validated, but is it possible that a workaround for this may be to set `-Dfile.encoding=UTF-8` in the `config\wrapper-properties.conf` of your installation? In theory the config files should be read, written and declared as `UTF-8` everywhere but I suppose it's possible something might have been missed somewhere along the line or a regression introduced on Windows.

The last line here does look a bit dubious
https://github.com/gocd/gocd/blob/5c7c6c226dc2caa1fe729f5142536634912e795a/server/src/main/java/com/thoughtworks/go/config/GoFileConfigDataSource.java#L405-L411

chadlwilson:
FWIW, setting`-Dfile.encoding=UTF-8 isn't enough. You need to set the wrapper specific property of either `wrapper.jvm.encoding=UTF-8` or `wrapper.lang.windows.encoding=UTF-8`. There's some weird wrapper behaviour at https://wrapper.tanukisoftware.com/doc/english/props-lang.html#platform

Example of error you can get on Admin when the default encoding is something like cp2252 and you try to save something

![image](https://user-images.githubusercontent.com/29788154/197383129-b4d0713c-2ccb-40a4-9bfe-1ab49575b75b.png)

PR #10896, opened 2022-10-06
Fix problems with cruise-config encoding

cinex-ru:
Issue: #10895

Description:
Fix problems with cruise-config encoding when saving from admin panel and from task.

### DIFF
--- a/config/config-api/src/main/java/com/thoughtworks/go/util/XmlUtils.java
+++ b/config/config-api/src/main/java/com/thoughtworks/go/util/XmlUtils.java
@@ -24,6 +24,7 @@ import org.jdom2.output.XMLOutputter;
 
 import java.io.*;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
 
 public class XmlUtils {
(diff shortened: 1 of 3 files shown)


chadlwilson:
Hey there! Any reason you decided to close this PR?

I think you're right that there is likely a problem here on Windows where the (Java) default encoding is not UTF-8. We'd need to think carefully about where/how to address it as changing the behaviour may break backwards compatibility for some folks where we change the default interpretation from cp-1252, and they have certain mismatched characters.

cinex-ru:
> t there is likely a problem here on Windows where the (Java) default encoding is not UTF-8. We'd need to think carefully about where/how to address it as changing the behaviour may break backwards compatibility for some folks where we change the default interpretation from cp-1252, and they have certain mism

Hello.
I found one more related problem and need some time to rework and test this PR. Hope it will be ready today.

chadlwilson:
OK. Yeah, we'd ideally want some unit tests too, although acknowledge that it's a bit difficult in some of these areas.

cinex-ru:
> OK. Yeah, we'd ideally want some unit tests too, although acknowledge that it's a bit difficult in some of these areas.

Yeah, feel like can't do without testing.

cinex-ru:
> OK. Yeah, we'd ideally want some unit tests too, although acknowledge that it's a bit difficult in some of these areas.

I finished but without tests yet. Checked only with manual testing. If thats ok, i can create new PR.

PR #10982, opened 2022-11-02
Fix Postgres backups via pg_dump on Windows

chadlwilson:
It seems the order of arguments can be important on some versions of PostgreSQL / `pg_dump`. Avoid this by using the `--dbname` option to specify the database name.

Technically non-options should be the last arg of the dump as documented for all versions (`9.6` -> `15`): https://www.postgresql.org/docs/15/app-pgdump.html

See https://groups.google.com/g/go-cd/c/AMtgQQS3Gck


chadlwilson:
This should be low risk. Will merge to produce experimental installers and revert if things get messy.

chadlwilson:
Manually validations
* Still works fine with MacOS pg_dump against Linux PostgreSQL 15
* User checked the fix is fine on Windows pg_dump with PostgreSQL 12

Issue #3974, opened 2017-10-31
No way to override bundled plugin when using Docker


chriswhelix:
##### Issue Type

Bug Report

##### Summary

The latest GoCD server version (17.11.0) does not come bundled with the latest YAML config plugin version (0.6.0). When using the official GoCD Docker image to run the server, there is no way to override the bundled plugin.

##### Environment

Using a Dockerfile that looks like:

```Dockerfile
FROM gocd/gocd-server:v17.11.0

ENV GOCD_YAML_CONFIG_VERSION=0.6.0

RUN mkdir -p /godata/plugins/external && cd /godata/plugins/external \
	&& wget "https://github.com/tomzo/gocd-yaml-config-plugin/releases/download/${GOCD_YAML_CONFIG_VERSION}/yaml-config-plugin-${GOCD_YAML_CONFIG_VERSION}.jar"
```

##### Desired Results

GoCD uses plugin version 0.6.0.

##### Actual Results

GoCD uses bundled plugin version 0.5.0. Note, this is as per documentation; the behavior is expected, but deeply frustrating.

On a persistent server, I could remove the bundled plugin, or install the new version over it, then restart GoCD. But because we're using Docker to manage plugin installation, all plugin installation activities occur before the first run of the server. So whatever I might do in the Dockerfile, as soon as the server boots up, it writes out its bundled plugins, overriding my desired versions.

##### Possible Fix

My immediate problem would be solved by bundling the latest version of the plugin (CC/ @tomzo).

However, it also seems to me there should be some way, as a user of GoCD via Docker, to disable or replace bundled plugins. I would view it as a defect that, in the context of Docker, certain plugins have been made mandatory and can not be upgraded or disabled (unless I'm missing something).

bdpiparva:
@chriswhelix --- You are right about the bundled plugin behaviour. On server boot up it first loads the bundled plugins first and then the external plugin. Only way I can see is,

1. Start `gocd/gocd-server:v17.11.0` with volume mount.
```bash
docker run --name go-server -d -p8153:8153 -p8154:8154 -v <PATH_TO_DATA_DIR>:/godata gocd/gocd-server:v17.11.0
```
2. Replace bundled plugin under <PATH_TO_DATA_DIR>/plugins/bundled
```bash
wget -O gocd-yaml-config-plugin.jar  https://github.com/tomzo/gocd-yaml-config-plugin/releases/download/0.6.0/yaml-config-plugin-0.6.0.jar
```
[...]

PR #11125, opened 2022-12-20
Remove bundling of the yum-repo-poller plugin

chadlwilson:
Having [this plugin](https://github.com/gocd/gocd-yum-repository-poller-plugin) bundled imposes a cost on all agent initialisations since it is a package repo plugin and a relatively uncommon use case in the current setup. Better to allow people to add this if they need it IMHO.

It also has some challenges due to compatibility with various RHEL/CentOS versions over time and the transition from yum to dnf (and aliasing of dnf back to `yum`) as per #3974 and https://github.com/gocd/gocd/issues/3974#issuecomment-1003149583

This is a **breaking change** and will need to be communicated in the release notes with indication of how to add it back.

Other things to do
- [x] Validate that config is not lost if server started without the plugin, and then functionality restored when plugin added
- [ ] Add it back to build.gocd.org plugin install automation (not used, but can be useful)
- [x] Ensure it is available for ruby-functional-test regression suite if used there
- [ ] Change the ruby-functional-test expectations to expect to be bundled
- [x] Update the README for the plugin itself
  - Done [here](https://github.com/gocd/gocd-yum-repository-poller-plugin/commit/f6d705d9afbdcfce58890ab581b6dad91329cb6c)
- [x] Prepare instructions for upgrade in release notes and update plugins page
  - Done [here](https://github.com/gocd/www.go.cd/pull/1436)
- [x] Update docs on docs.gocd.org  
  - Done [here](https://github.com/gocd/docs.go.cd/commit/8ee877bdfaba073aa9b1eb25e7ad8a9a9707904c)

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -1013,7 +1013,6 @@ task licenseReportAggregate {
                                  'gocd-yaml-config-plugin.jar',
                                  'gocd-filebased-authentication-plugin.jar',
                                  'gocd-ldap-authentication-plugin.jar',
-                                 'gocd-yum-repository-poller-plugin.jar',
                                  'gocd-file-based-secrets-plugin.jar']
 
   def getAllJarDependencies = {
--- a/tw-go-plugins/build.gradle
+++ b/tw-go-plugins/build.gradle
@@ -39,13 +39,6 @@ def dependencies = [
     checksum: 'f8603c66007feb5bccda62a7a9169a7d1c6dabad2775277fcd9b8641f4687065'
   ),
   new GithubArtifact(
-    user: 'gocd',
-    repo: 'gocd-yum-repository-poller-plugin',
-    tagName: 'v2.0.6-135',
-    asset: 'gocd-yum-repo-plugin-2.0.6-135.jar',
-    checksum: '5988384979750c4f450896ab52af8729f4a8602ec2c37ed45da5482a13539654'
-  ),
-  new GithubArtifact(
     user: 'tomzo',
     repo: 'gocd-yaml-config-plugin',
     tagName: '0.14.1',


chadlwilson:
Hello @arvindsv - wonder what you think about this? I am thinking I will use the faux-semver `23.x` to make a couple of minorish breaking changes like this :)

**Consequences**
My current understanding of this is that even removing the plugin will leave the various config in tact in people's `config.xml`, so when folks add it back as an external plugin things will go back how they used to be. Will validate this empirically though.

Will need release notes/changelog documentation obviously.

**Open question**

Should we consider removing the `ldap-authentication-plugin` too, forcing folks to make a choice between the two LDAP variants? How common is LDAP auth really? (I have no idea - but suspect for most folks, the auth'z plugin is the better choice now it's free)

The current list of bundled plugins is

https://github.com/gocd/gocd/blob/d0ef7eb3421802b12236537c14b80ec1bff36310/tw-go-plugins/build.gradle#L27-L69

All except ldap-auth and yum-repo seem reasonable defaults to me. _Personally_ I would add [script-executor](https://github.com/gocd-contrib/script-executor-task) as a default due to how much easier it makes pipelines-as-code, but ya, no powershell support and a few downsides probably makes it not suitable.

I don't have the background on the philosophy to decide whether a plugin should be bundled or not, however I think the situation has changed a bit here since some of the original decisions due to
- changing tech
- open-sourcing of some of the previous commercial plugins

**Pros to bundling plugins**
- makes it easier to do the "right thing", e.g use of JSON or YAML pipelines-as-code plugins
- provides a "real" entry point for new administrators to see examples of, and then "discover" other plugins that might help them.
- doesn't rely on users to upgrade plugins when security issues or issues in dependencies arise

**Cons to bundling plugins**
- slower server startup, marginally
- for task and package repository plugins, slower agent bootstrap and startup
- "opinions" can nudge people in the wrong direction, arguably (e.g LDAP authentication plugin is bundled, but the LDAP authorization plugin - which used to be commercial - is not)
- cannot remove/disable bundled plugins (https://github.com/gocd/gocd/issues/3974)
- cannot override or force upgrade/downgrade bundled plugins (https://github.com/gocd/gocd/issues/3974)
- can be confusing for admins - since you cannot disable LDAP auth'n plugin, if you are trying to use LDAP auth'z plugin you just have to ignore the auth'n plugin's existence.
- greater attack surface area for vulnerabilities in dependencies or the core plugin model
- greater testing surface area (since you cannot disable them, ensuring they work or at least dont degrade is kind of required for GoCD server/agent changes)

arvindsv:
> Hello @arvindsv - wonder what you think about this? I am thinking I will use the faux-semver 23.x to make a couple of minorish breaking changes like this :)

Makes sense to me. With yum, I think it's fine -- as long as we make sure we've documented the steps to be taken clearly. As you mentioned, a link in the release notes would be fine.

> Should we consider removing the ldap-authentication-plugin too, forcing folks to make a choice between the two LDAP variants? How common is LDAP auth really? (I have no idea - but suspect for most folks, the auth'z plugin is the better choice now it's free)

For the LDAP authentication plugin: I agree that the gocd-ldap-authorization-plugin is a better option for most. However, I worry about the impact of someone not checking the release notes and upgrading and getting locked out (temporarily). Won't be a pleasant experience. If we do go down that path, we need to document everything well, and make sure people can follow it easily.

I was thinking about whether the config can be automatically migrated, since a bundled plugin is essentially GoCD code. But, the configuration of the two plugins might not be directly compatible.

> I don't have the background on the philosophy to decide whether a plugin should be bundled or not ...

I wonder if you've put in more though than anyone else ever did. :P But no, your list of pros and cons make sense, and we used similar criteria, but ours were less-well articulated.

chadlwilson:
> For the LDAP authentication plugin: I agree that the gocd-ldap-authorization-plugin is a better option for most. However, I worry about the impact of someone not checking the release notes and upgrading and getting locked out (temporarily). Won't be a pleasant experience. If we do go down that path, we need to document everything well, and make sure people can follow it easily.

Yeah, true - I was thinking something similar. Maybe a bit too disruptive without doing something automatic with the config like you suggest - which might be a bit of work to evaluate and test as im not 100% sure the authz plugin is superset of the authn plugin. Might go into the 'too hard' basket for now 😅

Might be better to put the effort into resolving #3974 and also allowing some disabling, perhaps using something akin to ChatGPTs [suggested](https://groups.google.com/g/go-cd/c/0yvJMsLyvZs) design 😅

arvindsv:
> Might be better to put the effort into resolving #3974 and also allowing some disabling, perhaps using something akin to ChatGPTs [suggested](https://groups.google.com/g/go-cd/c/0yvJMsLyvZs) design 😅

🤣

PR #11139, opened 2022-12-28
Rationalize Alpine image JRE requirements

chadlwilson:
- Upgrade zlib
- Building image without gcc-libs, which seems not to be required.


chadlwilson:
All the tests pass, and a sanity-start locally seems to work fine.

chadlwilson:
gcc-libs was supplying 
* `libgcc_s.so.1`
* `libstdc++.so.6.0.28`

Cannot find any libs/binaries on our built images linking to these at these stage, so I think safe to remove. There is always a risk that some native code is being loaded into the JVM that is linked to these, however at this stage cannot find any evidence of this. Should be fine :-)

PR #11205, opened 2023-01-14
Simplify OS completeName determination 

chadlwilson:
This change replaces the Python+direct file reading hack for determining a "complete" OS family names with a more modern cross-platform library ([oshi](https://github.com/oshi/oshi)) that does the same. The library has ability to use native code to get a LOT more information out from the underlying OS than is being used here.

The hack was introduced in #2421 despite some of the initial objections (as to shelling out with Python, [here](https://github.com/gocd/gocd/pull/2421#issuecomment-231932261)), reverted, and then reintroduced in #2608 eventually. The attempt to shell out to Python happens regardless of OS, and I'm not sure it does any better than reading the name from `/etc/os-release` anyway.

This change also
- moves the logic back to the `agent` lib, to avoid dependencies propagating everywhere. Only the agent needs this logic.

Some examples (before `-->` after)

_Basically the same_ (linuxes)
- Alpine Linux v3.17 `-->` Alpine Linux 3.17.1 
- Ubuntu 22.04.1 LTS `-->`  Ubuntu 22.04.1 LTS (Jammy Jellyfish)
- CentOS Stream 9 `-->` CentOS Stream 9 
- Debian GNU/Linux 11 (bullseye)`-->` Debian GNU/Linux 11 (bullseye)
- Windows 11 `-->` Windows 11

_Enhanced_ (non-linuxes)
- Mac OS X `-->`  macOS 13.1 (Ventura)
- Windows Server 2022 `-->` Windows Server 2022 (Datacenter)

It also adds support for many others, so is far is less opinionated than before.

Task list
- [x] Prefix with manufacturer? (`Linux`, `GNU/Linux`, `Apple`, `Microsoft` etc)
  - The existing ones don't have a "manufacturer name" so let's leave it for now
- [x] Verify fallback behaviour of OSHI if can't determine at least the family + version


chadlwilson:
Any input from Mr @ketan here would also be appreciated :-)

ketan:
Also, I've not seen the code — but I'd make sure to "cache" the value, so as to not evaluate the OS version on every method invocation, mostly to avoid any expensive shelling out or file reading this library might be doing under the hood.

chadlwilson:
Thanks - yeah it caches the value. I'll double check the behaviour if the library can't read the value for some reason for a given distro/OS.

arvindsv (approved):
Looks good to me. The `SystemInfoTest` is the one I was worried about -- as not being future proof, and whether we needed such a strict regex to verify it. But, it's unlikely to change too quickly. Should be ok.

Also, TIL about `!!` [used in build.gradle](https://github.com/gocd/gocd/pull/11205/files#diff-7152b7052968f4fe41e7f92f5d5280b79c020ffb4993b173bdc67d072f7b9b45L23) ([gradle doc](https://docs.gradle.org/current/userguide/single_versions.html#simple_version_declaration_semantics))

chadlwilson:
> Also, TIL about `!!` [used in build.gradle](https://github.com/gocd/gocd/pull/11205/files#diff-7152b7052968f4fe41e7f92f5d5280b79c020ffb4993b173bdc67d072f7b9b45L23) ([gradle doc](https://docs.gradle.org/current/userguide/single_versions.html#simple_version_declaration_semantics))

Probably would be better to use the more descriptive version than the magic. The OSGI library depends on slf4j APi 2.x and it causes an implicit upgrade where which has some other problems I haven't been able to workaround yet with respect to the way we use classloaders within the agent and launcher. Their usage should still work with SLF4J 1.7.x as far as I can tell since it is largely API compatible.

Issue #11100, opened 2022-12-09
Guard against OOM errors for extremely long log lines


sandro-h:
##### Issue Type

- Feature enhancement

##### Summary

GoCD can run into Out Of Memory errors when handling build output consisting of extremely long lines. I'm wondering if we can find an effective way to guard against this.

Recently, our GoCD instance crashed with an Out Of Memory error because a build produced console output where a single line was 350 MB in size. When GoCD tried to display the console log in the UI, it crashed handling this extremely long line. Of course we are going to correct this build, since such a long line is virtually useless for human consumption anyway.

Nonetheless, I was wondering if we can guard against such crashes in the future. We have many different teams working with GoCD, and accidents can always happen again, and such crashes cause quite a lot of disruption. We can (and have in the past) increased the heap size available to GoCD, but that only goes so far. It would be nicer if GoCD guards against reading extremely long log lines.

I investigated briefly, but it doesn't look entirely trivial:
* Line-wise loading doesn't only happen when displaying the log in the GUI, but in other places as well, for example while transmitting the log from agent to server.
* Bounded line reading is harder than it seems, based on searching for existing solutions. At least when considering the many types of input streams and input data in the generic case. Not to mention losing out on the efficient implementation provided by the JDK.
  * https://stackoverflow.com/questions/5960554/maximum-line-length-for-bufferedreader-readline-in-java
  * https://bugs.openjdk.org/browse/JDK-4107821
  * https://github.com/seantmalone/BoundedBufferedReader
* What would be a good bound to begin with? Should it be configurable by the operator?
* What should happen when the bound is reached? If the line is truncated, how does the user know? If the line is split into two, it doesn't represent the reality anymore and could cause confusion.

My minimalist, egotistical ("works for us") approach would be to fail fast at the source: the StreamPumper class responsible for reading command process output could read lines in a bounded fashion and throw an error if the bound is exceeded. I'd rather make our teams aware that their builds are doing something bad, by failing the build, than trying to deal with the log all over the place.

Any thoughts on this are much appreciated.

chadlwilson:
Putting aside "how" to do it right now, it seems to me a conservative approach would be to
* define a default limit, intended to be imposed at agent level
* allow it to be overridden with a server-side system property (there is an existing mechanism to propagate such properties to all agents)
* if the limit is reached, crop the line, and then prepend or append something like `... [ cropped by GoCD ]` to every such line.
[...]

PR #11231, opened 2023-01-23
Crop overly long log lines

sandro-h:
Issue: #11100

Description:

Log lines exceeding a configurable threshold are cropped on agent side and a "...[ cropped by GoCD ]" notice appended to the line. This prevents extremely long log output from crashing the server with an Out Of Memory error.

The threshold defaults to 1 million characters per line, and can be configured on the server side like so:

-Dgocd.agent.extra.properties="go.console.log.max.line.length=1000000"

Cropping can be disabled by setting the threshold to 0.

This solution may still crash an agent, since a line is first read fully into memory before cropping. Rationale:

1. Protecting the central server is more important than individual agents
2. Secret redaction requires that the full line is known, otherwise we risk leaking a partially cropped secret.
3. Reading output into memory in a bounded fashion requires a deep dive into Java standard library code (i.e. BufferedReader). This could cause additional effort to keep in sync with later improvements or fixes (though it probably changes rarely).

The last two points are not impossible to solve, but the cost/benefit is questionable.

The cropping is implemented as an additional OutputStreamConsumer decorator and applied only in CommandLine::runScript. This method is used for user-defined pipeline tasks, but not for other commands, such as SCM tasks, where cropping is undesirable.

### DIFF
--- a/commandline/src/main/java/com/thoughtworks/go/util/command/CommandLine.java
+++ b/commandline/src/main/java/com/thoughtworks/go/util/command/CommandLine.java
@@ -327,7 +327,13 @@ public class CommandLine {
         ProcessWrapper process;
         int exitCode;
 
-        SafeOutputStreamConsumer streamConsumer = new SafeOutputStreamConsumer(new ProcessOutputStreamConsumer(outputStreamConsumer, errorStreamConsumer));
+        SafeOutputStreamConsumer streamConsumer =
+            new SafeOutputStreamConsumer(
+                new BoundedOutputStreamConsumer(
+                    new ProcessOutputStreamConsumer(outputStreamConsumer, errorStreamConsumer),
+                    new SystemEnvironment().get(SystemEnvironment.CONSOLE_LOG_MAX_LINE_LENGTH)
+                )
+        );
         streamConsumer.addArguments(getArguments());
         streamConsumer.addSecrets(environmentVariableContext.secrets());
         try {
--- a/commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
+++ b/commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
@@ -15,16 +15,25 @@
  */
 package com.thoughtworks.go.util.command;
 
+import com.thoughtworks.go.util.SystemEnvironment;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.OS;
 
+import static com.thoughtworks.go.util.SystemEnvironment.CONSOLE_LOG_MAX_LINE_LENGTH;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
 class CommandLineScriptRunnerTest {
+
+    @AfterEach
+    public void tearDown() {
+        System.clearProperty(CONSOLE_LOG_MAX_LINE_LENGTH.propertyName());
+    }
+
     @Test
     @EnabledOnOs(OS.LINUX)
     void shouldReplaceSecretsOnTheOutputUnderLinux() {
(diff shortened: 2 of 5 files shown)


commandline/src/main/java/com/thoughtworks/go/util/command/CommandLine.java
```diff
@@ -327,7 +327,13 @@ public void runScript(Script script, StreamConsumer buildOutputConsumer,
         ProcessWrapper process;
         int exitCode;
 
-        SafeOutputStreamConsumer streamConsumer = new SafeOutputStreamConsumer(new ProcessOutputStreamConsumer(outputStreamConsumer, errorStreamConsumer));
+        SafeOutputStreamConsumer streamConsumer =
+            new SafeOutputStreamConsumer(
+                new BoundedOutputStreamConsumer(
+                    new ProcessOutputStreamConsumer(outputStreamConsumer, errorStreamConsumer),
+                    new SystemEnvironment().get(SystemEnvironment.CONSOLE_LOG_MAX_LINE_LENGTH)
```
sandro-h, line 334: I'm not happy about reading from the system environment here. It would be better to pass this along the layers, like the `consoleLogCharset`. Requires a lot of changes though, and I wonder if these console related properties should be wrapped into a single `ConsoleLogProperties` data class then, for easier future adjustments.

sandro-h: Actually, it might not be so bad, using the `CommandLine::withXYZ` pattern to only set it on demand. Then we can propagate it along the BuildWork layers for the agent. And in all other cases we deliberately avoid setting a maximum line length (e.g. SCM tasks).

In that case, it would make sense to use the `BoundedOutputStreamConsumer` for all of `CommandLine`'s run methods, so the behavior is consistent when using `withMaxLineLength`.

chadlwilson:
Thanks for this, I'll get to it I promise 🙏

commandline/src/main/java/com/thoughtworks/go/util/command/CommandLine.java
chadlwilson: I'm not _overly_ worried about this, especially since the `runScript` method is a specific usage style of `CommandLine`, yet is just one (complicated) style with which you can use it.

Given the logic only has one consumer, arguably that's a signal that the logic should be in the `BaseComandBuilder` or in some wrapper class like `LowTrustScriptRunningCommandLine` or something. Honestly, I haven't bothered to try and get my head around all the various consumers and weird abstractions here. Seems overly complicated as it currently is :-) My point is that this is already pretty messy, and although that's not a reason to make it worse, it's probably not THAT much worse adding this :-)

This code is really rather old (and lifted from maven shared utils, based on the copyright); I suspect it might have been around since the original CruiseControl days 😅

The `consoleLogCharset` is even weirder in some ways as it is actually serialized and sent over as part of the `BuildWork`, as if it could reasonably be varied for one job vs another, rather than something defaulted as part of the agent code, and able to be overridden and propagated. So it kind of needs to be propagated through by necessity, rather than serving a function to do so...

commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
```diff
@@ -15,16 +15,25 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
 class CommandLineScriptRunnerTest {
+
+    @AfterEach
+    public void tearDown() {
+        System.clearProperty(CONSOLE_LOG_MAX_LINE_LENGTH.propertyName());
+    }
+
```
chadlwilson, line 36: The more modern of the tests use `SystemStub`s to ensure properties are restored to what they were before without lots of manual cruft. Could consider using these.

e.g

https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/domain/src/test/java/com/thoughtworks/go/config/materials/git/GitMaterialTest.java#L70-L89

chadlwilson (approved):

commandline/src/main/java/com/thoughtworks/go/util/command/CommandLine.java
chadlwilson: Your call if you want to have a think about this and submit another small PR :-)

chadlwilson:
Experimental builds `23.1.0-15827+` have this change in them if you want to try one from https://www.gocd.org/download/#zip or https://www.gocd.org/test-drive-gocd.html?dl=latest

commandline/src/test/java/com/thoughtworks/go/util/command/CommandLineScriptRunnerTest.java
chadlwilson: I tweaked this in a subsequent commit to do some other cleanups. Should get around to moving everything...

sandro-h:
Thanks for reviewing and accepting this! I briefly tested the experimental build and it looks to be working.

PR #11286, opened 2023-02-05
Improve agent work pickup speed with an exponential rather than fixed 10s backoff

### LINKED ISSUE #11215: Reduce the amount of time spent "waiting for an agent"
##### Issue Type

- Feature enhancement

##### Summary

My team has slowly been trying to improve our CD system and pipelines lately. One of the best ways we've found to do that is to split our longer running processes into different stages and jobs, which helps a lot with debugging when something goes wrong and tracking progress in general. 

However, as nice as this is from a usability point of view, each new job increases the runtime for that task for about 25-30 seconds. Slightly more if there is some load on the server, slightly less if there isn't. I haven't found anything online to indicate that this behavior is uncommon, if anything it seems like this is relatively fast from the other threads I've seen. While this may not sound like a lot of time, what used to be a 5-minute stage that we split into 5 1-minute stages now takes 7.5 minutes (50% longer) to run. The usability and maintenance benefits *are* huge, but they don't seem worth such a huge hit to runtime, especially for pipelines that may run a few times a day.

Increasing the number of agents and parallelizing things (splitting into jobs) has helped mitigate the issue somewhat, but there's still plenty of things that have to run sequentially, and we'd prefer not to have to decide between maintainability and speed, especially since the bottleneck seems arbitrary and unnecessary.

I did some looking into the logs and found a few concrete things that seem like they could maybe be improved (see below), but I'm no expert so I may be way off the mark. I also noticed that after upgrading from 20 to 22.3.0 things got better by about 10 seconds, which is an indication that either *someone* is looking into it or the base image just got better, but I'd say it's still slow enough to discourage us from using what we consider to be the better workflow. Case in point, we're now discussing merging everything back into a single stage wherever reasonable to avoid these time penalties.

##### Environment
We're using elastic agents with the [gocd-agent-ubuntu-22.04](https://github.com/gocd/docker-gocd-agent-ubuntu-22.04) image. The rest of the environment is:

```
* Version: "22.3.0 (15301-9d23ed19a9ea46eaf7f18bd16671ae0569871f53)"
* JAVA Version: 17.0.5
* OS: Linux 4.18.0-305.19.1.el8_4.x86_64
* Available Processors: 4
* Average System Load: 0.02
```

##### Log snippets

The only "suspicious" behavior I found was this weird time between getting a cookie and getting a response from the server, it's the last thing that happens before actually running the job commands and seems to be almost exactly 10 seconds every time:
```
jvm 1    | 2023-01-18 10:56:04,797 [stdout] - 2023-01-18 10:56:04,796 INFO  [main] DefaultPluginJarChangeListener:74 - Plugin load finished: /go/plugins/external/docker-registry-artifact-plugin-1.2.0-127.jar
jvm 1    | 2023-01-18 10:56:04,797 [stdout] - 2023-01-18 10:56:04,797 INFO  [scheduler-1] AgentHTTPClientController:129 - About to get cookie from the server.
jvm 1    | 2023-01-18 10:56:04,830 [stdout] - 2023-01-18 10:56:04,830 INFO  [scheduler-1] AgentHTTPClientController:132 - Got cookie: -redacted-
jvm 1    | 2023-01-18 10:56:15,067 [stdout] - 2023-01-18 10:56:15,067 INFO  [scheduler-2] HttpService:125 - Got back 200 from server
```
I'm not exactly sure what's happening between those two log lines, but I suspect it's something that could be fixed by someone who actually knows what they're doing. 

A lot of time (~10s) also seems to be spent downloading jar and zip files for agent plugins.

```
jvm 1    | 2023-01-18 10:55:52,766 INFO  [TouchLoopThread-1] Lockfile:103 - Using lock file: /go/.agent-bootstrapper.running
jvm 1    | 2023-01-18 10:55:53,281 INFO  [WrapperJarAppMain] ServerBinaryDownloader:114 - Downloading agent.jar
jvm 1    | 2023-01-18 10:55:53,392 INFO  [WrapperJarAppMain] ServerBinaryDownloader:121 - Got server response
jvm 1    | 2023-01-18 10:55:56,622 INFO  [WrapperJarAppMain] ServerBinaryDownloader:129 - Piped the stream to admin/agent
...
jvm 1    | 2023-01-18 10:56:04,797 [stdout] - 2023-01-18 10:56:04,796 INFO  [main] DefaultPluginJarChangeListener:74 - Plugin load finished: /go/plugins/external/docker-registry-artifact-plugin-1.2.0-127.jar
```

I suspect this is something that could be done periodically on the image itself to avoid having to update it every time a job runs, but again I may be wrong and am happy to be corrected by someone who actually knows the reasoning behind it.

Hopefully something can be done, because we would really like to take full advantage of all of GoCDs features and it seems silly to go back to having single long-running jobs after seeing all the advantages that multiple jobs can provide.

chadlwilson:
- Fixes #11215

### Summary

Prior to this change, agents had an initial once-off delay (1s) and then only attempted to get work every 10s after that. During start-up (and transitioning between pieces of work) this caused a lot of unnecessary delay, as it takes two server round trips _after agent registration_ to get a new piece of work - the first where the server doesn't yet know the agent is idle, and the second where it has asynchronously allocated some work.

This change mitigates some of this by
- changing the existing `agent.get.work.delay=1000` to be interpreted as "delay after each piece of work is retrieved" rather than "once-off at agent start"
- does a configurable exponential backoff `agent.get.work.backOffMultiplier=1.5`
- up to maximum wait between get-work-attempts of `agent.get.work.interval=10000`

In other words, each time the agent completes a piece of real work (with success or failure), it resets the wait time to the `agent.get.work.delay`. This has the effect of
- making agents much more responsive to picking up work when they initially start by default (hugely beneficial for elastic agents) (estimated default saving of 7-15s)
- making it **much** faster to transition between pieces of work and the next when there is work waiting on server for an agent (for static agents) 

If the server responds with "no work to do", "agent denied" or some other failure (say a 503 during server restart, or an agent version mismatch), the exponential backoff for trying to get more work will begin, waiting for the "problem" to go away, or work to be available for the agent. With defaults `1s` --> `1.5s` --> `2.25s` --> `3.4s` --> `5.1s` --> `7.6s` --> `10s` --> `10s` etc.

### Downsides

- More calls from agent to server, but not really that many, and checking for work is easy/low cost, happening all in memory.
- Does not address the problem that two "please give me work" calls are required to the server for a new agent, 1 for the agent to tell the server its ready for work, and other to actually give it some work.
- Implementation uses a perma-running task on the scheduler, which is not really good practice, but easiest way to fit into the current convention of Spring usage. To be ultra conservative, if the thread dies for some reason it gets immediately restarted on the scheduler.

### Other options considered

1. Do something on the server so that as soon as the agent is registered it is known as idle and starts to assign work to it, likely before getWork is called. Might lead to issues when server is restarted/upgraded? Not sure.
    - Would be nice, but far more complicated and invasive. This change has value regardless of server side changes, and is agent side only.
2. Have the agent try again nearly immediately on first attempt to get work
    - Things happen on the server asynchronously so it's hard to tell when the work will be ready for the agent. 

### Results

With this change and default settings, testing with the docker-elastic-agents plugin we get something representative like the below

| Section  | Time | Section Elapsed |
| ------------- | ------------- | ------------- |
| Start | 14:52:37.000 | 0 |
| Approx "elastic agent schedule" time  | 14:52:38.860 | 0-5s (unchanged, `cruise.build.assignment.service.interval` poll)  |
| Container started  | 14:52:39.361 | ~500ms (unchanged) |
| Agent binaries downloaded | 14:52:40,808 | ~1.4s (unchanged) | 
| Agent started + registered with server + cookie | 14:52:43,398 | ~2.6s (unchanged) | 
| Got First Work  | 14:52:46,913 | ~3.6s (was ~10-20s depending on luck) |
| TOTAL | | Probably ~7-12s (depending on poll luck, with extra for slow network speeds or slow agent downloads) |

This last step is the bit that is far faster now compared with `22.3.0`. Prior to `22.3.0` the "agent started" time had a big `10s` delay in it after binary download and a `1s` fudge factor sleep thrown in, so this builds on the changes in #10759

### DIFF
--- a/agent/src/main/java/com/thoughtworks/go/agent/AgentController.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/AgentController.java
@@ -25,6 +25,7 @@ import com.thoughtworks.go.domain.AgentRuntimeStatus;
 import com.thoughtworks.go.domain.AgentStatus;
 import com.thoughtworks.go.plugin.infra.PluginManager;
 import com.thoughtworks.go.plugin.infra.PluginManagerReference;
+import com.thoughtworks.go.plugin.infra.monitor.PluginJarLocationMonitor;
 import com.thoughtworks.go.remote.AgentIdentifier;
 import com.thoughtworks.go.server.service.AgentRuntimeInfo;
 import com.thoughtworks.go.server.service.ElasticAgentRuntimeInfo;
@@ -51,6 +52,7 @@ public abstract class AgentController {
     private final SubprocessLogger subprocessLogger;
     private final AgentUpgradeService agentUpgradeService;
     private final AgentHealthHolder agentHealthHolder;
+    private final PluginJarLocationMonitor pluginJarLocationMonitor;
     private final String hostName;
     private final String ipAddress;
 
@@ -60,12 +62,14 @@ public abstract class AgentController {
                            PluginManager pluginManager,
                            SubprocessLogger subprocessLogger,
                            AgentUpgradeService agentUpgradeService,
-                           AgentHealthHolder agentHealthHolder) {
+                           AgentHealthHolder agentHealthHolder,
+                           PluginJarLocationMonitor pluginJarLocationMonitor) {
         this.sslInfrastructureService = sslInfrastructureService;
         this.agentRegistry = agentRegistry;
         this.subprocessLogger = subprocessLogger;
         this.agentUpgradeService = agentUpgradeService;
         this.agentHealthHolder = agentHealthHolder;
+        this.pluginJarLocationMonitor = pluginJarLocationMonitor;
         PluginManagerReference.reference().setPluginManager(pluginManager);
         hostName = SystemUtil.getLocalhostNameOrRandomNameIfNotFound();
         ipAddress = SystemUtil.getClientIp(systemEnvironment.getServiceUrl());
@@ -75,28 +79,34 @@ public abstract class AgentController {
 
     public abstract void execute();
 
-    public final void loop() {
+    final WorkAttempt performWork() {
         try {
             LOG.debug("[Agent Loop] Trying to retrieve work.");
             agentUpgradeService.checkForUpgradeAndExtraProperties();
             sslInfrastructureService.registerIfNecessary(getAgentAutoRegistrationProperties());
-            work();
-            LOG.debug("[Agent Loop] Successfully retrieved work.");
+
+            if (pluginJarLocationMonitor.hasRunAtLeastOnce()) {
+                return tryDoWork();
+            } else {
+                LOG.debug("[Agent Loop] PluginLocationMonitor has not yet run. Not retrieving work since plugins may not be initialized.");
+                return WorkAttempt.FAILED;
+            }
+
         } catch (Exception e) {
             if (isCausedBySecurity(e)) {
                 handleSecurityException(e);
             } else {
                 LOG.error("[Agent Loop] Error occurred during loop: ", e);
             }
+            return WorkAttempt.FAILED;
         }
-
     }
 
     private void handleSecurityException(Exception e) {
         LOG.error("There has been a problem with one of GoCD's TLS certificates. This can be caused by a man-in-the-middle attack, or a change to the HTTPS certificates of the GoCD Server. Review the agent TLS trust settings and any mutual TLS configuration of the agent.", e);
     }
 
-    protected abstract void work();
+    protected abstract WorkAttempt tryDoWork();
 
     protected AgentAutoRegistrationProperties getAgentAutoRegistrationProperties() {
         return agentAutoRegistrationProperties;
--- a/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/AgentHTTPClientController.java
@@ -51,7 +51,6 @@ public class AgentHTTPClientController extends AgentController {
     private final SslInfrastructureService sslInfrastructureService;
     private final ArtifactExtension artifactExtension;
     private final PluginRequestProcessorRegistry pluginRequestProcessorRegistry;
-    private final PluginJarLocationMonitor pluginJarLocationMonitor;
 
     private final PackageRepositoryExtension packageRepositoryExtension;
     private final SCMExtension scmExtension;
@@ -75,7 +74,7 @@ public class AgentHTTPClientController extends AgentController {
                                      PluginRequestProcessorRegistry pluginRequestProcessorRegistry,
                                      AgentHealthHolder agentHealthHolder,
                                      PluginJarLocationMonitor pluginJarLocationMonitor) {
-        super(sslInfrastructureService, systemEnvironment, agentRegistry, pluginManager, subprocessLogger, agentUpgradeService, agentHealthHolder);
+        super(sslInfrastructureService, systemEnvironment, agentRegistry, pluginManager, subprocessLogger, agentUpgradeService, agentHealthHolder, pluginJarLocationMonitor);
         this.client = client;
         this.packageRepositoryExtension = packageRepositoryExtension;
         this.scmExtension = scmExtension;
@@ -84,7 +83,6 @@ public class AgentHTTPClientController extends AgentController {
         this.sslInfrastructureService = sslInfrastructureService;
         this.artifactExtension = artifactExtension;
         this.pluginRequestProcessorRegistry = pluginRequestProcessorRegistry;
-        this.pluginJarLocationMonitor = pluginJarLocationMonitor;
     }
 
     @Override
@@ -113,15 +111,9 @@ public class AgentHTTPClientController extends AgentController {
     }
 
     @Override
-    public void work() {
-        LOG.debug("[Agent Loop] Trying to retrieve work.");
-        if (pluginJarLocationMonitor.hasRunAtLeastOnce()) {
-            retrieveCookieIfNecessary();
-            retrieveWork();
-            LOG.debug("[Agent Loop] Successfully retrieved work.");
-        } else {
-            LOG.debug("[Agent Loop] PluginLocationMonitor has not yet run. Not retrieving work since plugins may not be initialized.");
-        }
+    public WorkAttempt tryDoWork() {
+        retrieveCookieIfNecessary();
+        return doWork();
     }
 
     private void retrieveCookieIfNecessary() {
@@ -133,22 +125,24 @@ public class AgentHTTPClientController extends AgentController {
         }
     }
 
-    void retrieveWork() {
+    private WorkAttempt doWork() {
         AgentIdentifier agentIdentifier = agentIdentifier();
         LOG.debug("[Agent Loop] {} is checking for work from Go", agentIdentifier);
-        Work work;
         try {
             getAgentRuntimeInfo().idle();
-            work = client.getWork(getAgentRuntimeInfo());
+            Work work = client.getWork(getAgentRuntimeInfo());
             if (!(work instanceof NoWork)) {
                 LOG.debug("[Agent Loop] Got work from server: [{}]", work.description());
             }
             runner = new JobRunner();
             final AgentWorkContext agentWorkContext = new AgentWorkContext(agentIdentifier, client, manipulator, getAgentRuntimeInfo(), packageRepositoryExtension, scmExtension, taskExtension, artifactExtension, pluginRequestProcessorRegistry);
             runner.run(work, agentWorkContext);
+            LOG.debug("[Agent Loop] Successfully executed work.");
+            return WorkAttempt.fromWork(work);
         } catch (UnregisteredAgentException e) {
             LOG.warn("[Agent Loop] Agent is not registered. [{}] Registering with server on next iteration.", e.getMessage());
             sslInfrastructureService.createSslInfrastructure();
+            return WorkAttempt.FAILED;
         } finally {
             getAgentRuntimeInfo().idle();
         }
--- a/agent/src/main/java/com/thoughtworks/go/agent/AgentWorkRetrievalScheduler.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/AgentWorkRetrievalScheduler.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2023 Thoughtworks, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.thoughtworks.go.agent;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.TaskScheduler;
+import org.springframework.scheduling.support.PeriodicTrigger;
+import org.springframework.stereotype.Component;
+import org.springframework.util.backoff.BackOff;
+import org.springframework.util.backoff.BackOffExecution;
+
+import javax.annotation.PostConstruct;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class AgentWorkRetrievalScheduler implements Runnable {
+    private static final Logger LOG = LoggerFactory.getLogger(AgentWorkRetrievalScheduler.class);
+
+    private final AgentController controller;
+    private final BackOff backoffStrategy;
+    private final TaskScheduler scheduler;
+
+    @Autowired
+    public AgentWorkRetrievalScheduler(AgentController controller, BackOff backoffStrategy, TaskScheduler scheduler) {
+        this.controller = controller;
+        this.backoffStrategy = backoffStrategy;
+        this.scheduler = scheduler;
+    }
+
+    @PostConstruct
+    public void schedule() {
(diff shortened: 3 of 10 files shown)


chadlwilson:
Not sure if I can tag unrelated folks - but @chrisb2244 - you might be interested in this or the workarounds discussed in #11215. Watched your sharing on use of GoCD with LabVIEW [here](https://www.youtube.com/watch?v=1xk8GsoWISI) and noted your comments about the 30s agent startup time.

Based on my experience as well as comments in #11215, the ~30s is probably accurate as of `22.2.0`. `22.3.0` probably reduced that to ~20s for a local docker elastic agent. This change I hope to reduce to more like 7s-12s (network-willing for folks with remote agents).

If you're interested in helping to validate when/if we decide to merge this, I can let you know. Right now I haven't actively tested with Windows WSL2 containers like I believe you use (easy availability of a machine/platform/arch where that works more than desire).

chrisb2244:
Hi Chad - thanks for the heads-up.

I'm shortly going to be working on moving my GoCD server from the workstation that I use daily to a separate (slower, but otherwise essentially unused) machine with a bunch more RAM, as a less "testing" and more "this is my new full-time build system", so I'll be working with GoCD over the next (probably) couple of weeks.

I'll set it up with the latest stable and then try comparing it with this branch and let you know what differences I see, but it's great to know it might be faster - lots of my jobs are pretty quick (especially if they don't have dependencies on other jobs) and so a 10-15s reduction per elastic agent creation could vastly increase the speed from commit to all outputs built.

chadlwilson:
I can save you building this branch if we either merge it, or I can upload the unsigned branch/PR builds somewhere.

chadlwilson:
@arvindsv Based on your offline comment that this generally looked OK, I'll merge this to try and get a bit of "real world" smoke time on build.gocd.org - but still interested in any of your more specific feedback and will either revert or make necessary changes.

arvindsv:
@chadlwilson Thank you. This looks good to me.

One of the things I was looking at was to see if the inital call happened at time T = 0 seconds instead of the T = 1 second that it was at before. But, because you delay for the backoff number of seconds first ([here](https://github.com/gocd/gocd/pull/11286/files#diff-d61d6700be0c79e6fc2a6d551f1c31fbbc795b8571a115095cc4ec9f2383ecc2R59)) instead of attempting work means that the earlier initial delay of 1s is maintained.

There used to be some weird dependency between `loop()` and `ping()` in the controller that I'm not able to remember now. Maybe it was just about the cookie + SSL infrastructure bit? Not sure. But, that's unlikely to be affected by this, I think.

This change looks good to me. Thanks!

chadlwilson:
> @chadlwilson Thank you. This looks good to me.
> 
> One of the things I was looking at was to see if the inital call happened at time T = 0 seconds instead of the T = 1 second that it was at before. But, because you delay for the backoff number of seconds first ([here](https://github.com/gocd/gocd/pull/11286/files#diff-d61d6700be0c79e6fc2a6d551f1c31fbbc795b8571a115095cc4ec9f2383ecc2R59)) instead of attempting work means that the earlier initial delay of 1s is maintained.

Thank you @arvindsv ! 🙏 Yes as much as possible I have tried to inherit/reuse any overridding folks might have done to the system properties. They will behave slightly differently, but if, for example you had reduced the `agent.get.work.interval` to `1000`, you'll get the same behaviour as before.

> There used to be some weird dependency between `loop()` and `ping()` in the controller that I'm not able to remember now. Maybe it was just about the cookie + SSL infrastructure bit? Not sure. But, that's unlikely to be affected by this, I think.

Yeah, there are some shared pieces of state and variables, some of which I have corrected to be volatile to avoid weird semantics across threads, but generally speaking `ping` at least in its _current_ state doesn't appear to be able to "fix" any problem that might exist (e.g upgrade needed, a bad token, a missing cookie, an out-of-date SSL context) and it relies on the agent work loop to trigger those pieces which makes it a little easier to get ones head around. 

But goodness me, this area has had some changes associated over time :-)

@chrisb2244 [Experimental builds](https://www.gocd.org/download/#windows) of `23.1.0-15962`+ have this change included if you want to try it out.

chrisb2244:
Thanks Chad, I got a chance to test the experimental build you linked today and it seems like my agent wait time (or at least, the grey bar time when viewing a pipeline from the dashboard) decreased from ~30s in `22.2.0 (14697)` and `22.3.0 (15301)` to ~21s in the linked build (showing as `23.1.0 (15992)`).

Given my builds typically take about 30 seconds, this is a something like a 10-15% reduction in time per job. 🚀 

<details>
<summary>Possible extra benefit - untested yet</summary>
I haven't checked yet if this has any impact on the time between jobs (if the number of agents required is more than the maximum configured value, then they also take a little while to shut down) but I don't know if checking work is related.
I suspect probably not - maybe this is a configuration issue on my part, or part of the Docker elastic-agent plugin.
However, this is only a (small) problem for me if everything gets triggered, especially if I move to a machine with more RAM/CPU cores available to increase the cap on agents.
</details>

chadlwilson:
@chrisb2244 Thanks for trying it out!

> the grey bar time when viewing a pipeline from the dashboard) decreased from ~30s in `22.2.0 (14697)` and `22.3.0 (15301)` to ~21s in the linked build (showing as `23.1.0 (15992)`).

Hmm, I thought it might save more than that relate to `22.2.0`, but there are a few other variables, including how long it takes for your containers to start, downloads to happen etc, and I don't really have any gut instinct for how fast that stuff is on native Windows compared to say MacOS/Linux. In my experience, usually such things seem relatively slow 😅

For example, on https://build.gocd.org the bare minimum agent "grey bar time" I have seen for a Windows ECS elastic agent container (verify similar to docker elastic agent plugin) on a pre-warmed, massively over-provisioned Amazon ECS-optimized Windows 2022 Server Core (Datacenter) instance is `23-24s` (4 cores, 8GB container limits).

This compares pretty unfavourably to `8-9s` for Amazon Linux 2 elastic agents (3 cores, 6GB container limits). Same theoretical SSD performance for disks underneath etc etc, but not dug into it in detail. 🤷‍♂️

chrisb2244:
I'd guess the size of the image doesn't help - I don't know how your images look, but the images I'm using are 15-20GB, which definitely compares unfavourably with what I'd guess is the typical expectation of maybe a 500MB/1GB Linux image... 

As a result, some of my waiting time is probably unavoidable given the bulkiness of what I'm spinning up.

I think that probably some small improvement between 22.2.0 and later could be hidden by the way that my images are built - I predownloaded the GoCD agent jar into the image at (image) build time, but this only helps when the required version is the same as the baked-in version - if not, then a new jar is downloaded from the GoCD server, as I understand it?
I think this contributes a couple of seconds in some previous testing I did (so `22.3` might have been a couple of seconds faster than `22.2` if those seconds were moved into downloading a new jar).

chadlwilson:
> I'd guess the size of the image doesn't help - I don't know how your images look, but the images I'm using are 15-20GB, which definitely compares unfavourably with what I'd guess is the typical expectation of maybe a 500MB/1GB Linux image...

Yeah, it doesn't help - but in this case the "pre-warming" also means the image is cached locally. Still [a 3.9GB image though](https://hub.docker.com/r/gocddev/gocd-dev-build/tags?page=1&name=windows), which is about as small as I've been able to get it [basing on servercore:ltsc2022](https://github.com/gocd-contrib/gocd-oss-cookbooks/blob/master/windowsservercore.Dockerfile) and try to [install as little as possible](https://github.com/gocd-contrib/gocd-oss-cookbooks/blob/master/provision/install-packages.ps1) :-)

> As a result, some of my waiting time is probably unavoidable given the bulkiness of what I'm spinning up.

For a pre-cached image, I'm not sure if the image size/# layers affects launch speed on Windows. Doesn't seem to make a big difference on Linux/MacOS, but filesystems and sub-systems are rather different on Windows.

> I think that probably some small improvement between 22.2.0 and later could be hidden by the way that my images are built - I predownloaded the GoCD agent jar into the image at (image) build time, but this only helps when the required version is the same as the baked-in version - if not, then a new jar is downloaded from the GoCD server, as I understand it?

Yeah, that's roughly correct - there is some extra slowness if the agent version is different because the launcher that bootstraps the agent starts once and then needs to re-launch but you might find it is still downloading a few jars each time anyway due to the way it is written. At cost of some extra coupling of server/agent versions, you can probably make sure it doesn't need to download anything by either baking or bind-mounting all of the expected jars as individual files into the agent working folder (e.g `C:\go-agent`) [similar to this for Linux](https://github.com/gocd/gocd/issues/11215#issuecomment-1400585574).

I've never attempted the same thing on Windows since GoCD does not distribute official Windows agent images and not familiar with all the host bind mounting capabilities of Docker on Windows. FWIW, neither the Linux or Windows images used to build/test GoCD itself pre-bake (or have a container cache for) the agent-related jars (`agent.jar`/`tfs-impl.jar`/`agent-launcher.jar`/`agent-plugins.zip`) so they are being downloaded dynamically from the server in both of those times above.

chadlwilson:
Actually, I should caveat that the https://build.gocd.org agents use an [alternate agent bootstrapper written in golang](https://github.com/gocd-contrib/gocd-golang-bootstrapper), so it's not the "normal" experience if you are using the regular agent bootstrapper from https://www.gocd.org.

That may be contributing to some speed improvement ... or maybe just use less memory? Not sure. 😬

chrisb2244:
> I've never attempted the same thing on Windows since GoCD does not distribute official Windows agent images and not familiar with all the host bind mounting capabilities of Docker on Windows. FWIW, neither the Linux or Windows images used to build/test GoCD itself pre-bake (or have a container cache for) the agent-related jars (`agent.jar`/`tfs-impl.jar`/`agent-launcher.jar`/`agent-plugins.zip`) so they are being downloaded dynamically from the server in both of those times above.

I wasn't too sure about which images were best to use, so I put this in my Dockerfiles:
```
FROM labview2019_base AS labview2019_base_gocd
# This is placed at the bottom to allow more easily rebuilding without reinstalling all of LabVIEW.
COPY go-agent.ps1 .
ARG GO_SERVER_URL="http://10.2.3.4:8153/go"
RUN $ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest ($env:GO_SERVER_URL + '/admin/agent') -OutFile 'agent.jar'
RUN $ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest ($env:GO_SERVER_URL + '/admin/agent-plugins.zip') -OutFile 'agent-plugins.zip'
RUN $ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest ($env:GO_SERVER_URL + '/admin/tfs-impl.jar') -OutFile 'tfs-impl.jar'
```

where `go-agent.ps1` is if I recall correctly a modified version of [an example script](https://github.com/gocd-contrib/docker-elastic-agents-plugin/blob/master/contrib/scripts/bootstrap-without-installed-agent/go-agent) in the Docker elastic images repo:
<details>
<summary>go-agent.ps1</summary>

```
# -Force prevents errors if the directory exists. It does not delete contained files.
New-Item -ItemType "directory" -Path ".\config" -Force | Out-Null

# Write connection keys into the autoregister file
Add-Content -Path .\config\autoregister.properties -Value ("agent.auto.register.key="+$env:GO_EA_AUTO_REGISTER_KEY)
Add-Content -Path .\config\autoregister.properties -Value ("agent.auto.register.environments="+$env:GO_EA_AUTO_REGISTER_ENVIRONMENT)
Add-Content -Path .\config\autoregister.properties -Value ("agent.auto.register.elasticAgent.agentId="+$env:GO_EA_AUTO_REGISTER_ELASTIC_AGENT_ID)
Add-Content -Path .\config\autoregister.properties -Value ("agent.auto.register.elasticAgent.pluginId="+$env:GO_EA_AUTO_REGISTER_ELASTIC_PLUGIN_ID)
Get-Content -Path .\config\autoregister.properties

function get_checksums {
	$url = $env:GO_EA_SERVER_URL+"/admin/latest-agent.status"
	(Invoke-WebRequest $url -UseBasicParsing).Headers
}

function checkAndUpdate {
	param(
		[string]$localPath,
		[string]$remotePath,
		[string]$remoteChecksum
	)
	if ((Test-Path $localPath -PathType Leaf) -and ((Get-FileHash -Algorithm MD5 $localPath).Hash -eq $remoteChecksum.ToUpper())) {
		Write-Host $localPath " is up-to-date"
	} else {
		Write-Host $localPath " requires update..."
		$url = $env:GO_EA_SERVER_URL+$remotePath
		Invoke-WebRequest $url -OutFile $localPath
	}
}

# Get MD5 values from the server. These might be used for authentication/registration?
$checksums = get_checksums
$agent_md5 = $checksums.'Agent-Content-MD5'
$tfs_md5 = $checksums.'TFS-SDK-Content-MD5'
$plugins_md5 = $checksums.'Agent-Plugins-Content-MD5'
$agent_launcher_md5 = $checksums.'Agent-Launcher-Content-MD5'

# Download files from the server/elastic host if not up-to-date in the base image.
# This will typically mean slower starts after updating the GoCD server, until the images are rebuilt.
$ProgressPreference = 'SilentlyContinue'
checkAndUpdate -localPath 'agent.jar' -remotePath '/admin/agent' -remoteChecksum $agent_md5
checkAndUpdate -localPath 'agent-plugins.zip' -remotePath '/admin/agent-plugins.zip' -remoteChecksum $plugins_md5
checkAndUpdate -localPath 'tfs-impl.jar' -remotePath '/admin/tfs-impl.jar' -remoteChecksum $tfs_md5

Write-Host "Launching the GoCD Agent, and waiting for it to exit?"
$RUN_CMD = "java --% -Dcruise.console.publish.interval=10 -Xms128m -Xmx256m -Djava.security.egd=file:/dev/./urandom -Dagent.plugins.md5=$plugins_md5 -Dagent.binary.md5=$agent_md5 -Dagent.launcher.md5=$agent_launcher_md5 -Dagent.tfs.md5=$tfs_md5 -jar agent.jar -serverUrl $env:GO_EA_SERVER_URL"

Invoke-Expression $RUN_CMD
```

</details>

chadlwilson:
Ahh ok, if you are using the bootstrap-without-installed-agent approach and baking things in at build time, the stuff about re-launching agents is somewhat irrelevant :-)

PR #11317, opened 2023-02-19
Fix agent upgrade check md5Header NullPointerExceptions

### LINKED ISSUE #4469: Agent Error occurred during loop NullPointerException: Cannot invoke "org.apache.http.Header.getValue()" because "md5Header" is null

chadlwilson:
Fixes #4469 

### The Problem

Currently calls to the `/admin/latest-agent.status` API from static agents will trigger a `NullPointerException` like the below, **every 30 minutes**.

```
jvm 1    | 2023-02-18 16:37:13,918 ERROR [scheduler-1] AgentController:99 - [Agent Loop] Error occurred during loop:
jvm 1    | java.lang.NullPointerException: Cannot invoke "org.apache.http.Header.getValue()" because "md5Header" is null
jvm 1    | 	at com.thoughtworks.go.agent.service.AgentUpgradeService.validateMd5(AgentUpgradeService.java:111)
jvm 1    | 	at com.thoughtworks.go.agent.service.AgentUpgradeService.checkForUpgradeAndExtraProperties(AgentUpgradeService.java:84)
jvm 1    | 	at com.thoughtworks.go.agent.service.AgentUpgradeService.checkForUpgradeAndExtraProperties(AgentUpgradeService.java:72)
jvm 1    | 	at com.thoughtworks.go.agent.AgentController.performWork(AgentController.java:85)
```

Eventually, the JVM will start cutting off stacktraces and you will only see

```
jvm 1    | 2023-02-18 16:37:13,918 ERROR [scheduler-1] AgentController:99 - [Agent Loop] Error occurred during loop:
jvm 1    | java.lang.NullPointerException: null
```

This doesn't cause any major problem, because on next attempt things will work again, a new authenticated session will be created and the agent will continue to check for upgrades and get work. But it will cause an unnecessary 10s break in getting work.

### The Root Cause

This is happening every 30 minutes due to `go.security.reauthentication.interval` at https://github.com/gocd/gocd/blob/781bc78dc473cf14715408adf9f2b054b8a814a3/base/src/main/java/com/thoughtworks/go/util/SystemEnvironment.java#L178 and some logic causing calls to this endpoint having `re-authentication` forced after 30 minutes.

- Currently the agent shares the `HttpClient` between calls to check for agent upgrades (can be anonymous) and calls to the agent APIs (generally require authentication/registration with an `AgentToken`).
- The `JESSIONID` session cookie is shared between different calls made by the agent, so sometimes the agent sends a cookie referring to a session which was earlier authenticated with an `AgentToken`.
- However, during **agent upgrade checks** there is no agent UUID and token sent. Nor the params required to register an agent or acquire an authentication token. It is expected to work anonymously, and the cookie being sent is a side effect of use of a shared client.
- The current authentication filter being applied to `/admin/latest-agent.status` looks like the below, which means calls to this API fall back to `invalidateAuthenticationOnSecurityConfigChangeFilter, reAuthenticationWithRedirectToLoginPage, assumeAnonymousUserFilter, basicAuthenticationWithRedirectToLoginFilter`
https://github.com/gocd/gocd/blob/246aeb95302d3969595e689b2a5c2c2efb1882ac/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthenticationFilterChain.java#L38-L58
- Since `AgentToken` sessions cannot be "re-authenticated", re-authentication fails and the server issues a `302` redirect back to `/go/login`
https://github.com/gocd/gocd/blob/3ec98138e5f309666bca72873cad8e0a7a39783e/server/src/main/java/com/thoughtworks/go/server/newsecurity/filters/AbstractReAuthenticationFilter.java#L88-L97

### The "fix"

To resolve this, this change
- tells agent upgrade not to follow redirects. This is the direct cause of the `NullPointerException`s but not the root cause. This is more for future.
- Tells the authentication filter to treat this route as `assumeAnonymousUserFilter`, and not to apply reauthentication, security config change or other filters during authentication. **Authorization** rules applying to the routes [remain unchanged](https://github.com/gocd/gocd/blob/8811f094a79fe822e4cc0188194e4ecd17c4707d/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthorizeFilterChain.java#L37-L55).

### Alternatives considered

It also works to tell the `AgentUpgradeService` to ignore cookies. This way it never refers to an existing session and a new "anonymous" session is always created for every call in the agent loop. I started with this approach, but it seemed more of a workaround for incorrect authentication on the server on the route, and would cause many new single-shot sessions to be created and managed by the server. Didn't feel quite right, but maybe safer?

### Advice needed

I'm not 100% sure if this is the right approach here, and the consequences of [AssumeAnonymousUser](https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/server/src/main/java/com/thoughtworks/go/server/newsecurity/filters/AssumeAnonymousUserFilter.java#L50-L70) if a given client is accessing multiple routes/APIs that have different authentication requirements. It seems to work fine, and be correct though? :-)

### DIFF
--- a/agent/src/main/java/com/thoughtworks/go/agent/service/AgentUpgradeService.java
+++ b/agent/src/main/java/com/thoughtworks/go/agent/service/AgentUpgradeService.java
@@ -19,6 +19,7 @@ import com.thoughtworks.go.agent.common.ssl.GoAgentServerHttpClient;
 import com.thoughtworks.go.util.SystemEnvironment;
 import com.thoughtworks.go.util.URLService;
 import org.apache.http.Header;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.slf4j.Logger;
@@ -38,6 +39,8 @@ import static com.thoughtworks.go.util.SystemEnvironment.*;
 public class AgentUpgradeService {
     private static final Logger LOGGER = LoggerFactory.getLogger(AgentUpgradeService.class);
     private static final Marker FATAL = MarkerFactory.getMarker("FATAL");
+    private static final RequestConfig NO_FOLLOW_REDIRECT = RequestConfig.custom().setRedirectsEnabled(false).build();
+
     private final GoAgentServerHttpClient httpClient;
     private final SystemEnvironment systemEnvironment;
     private final URLService urlService;
@@ -68,10 +71,16 @@ public class AgentUpgradeService {
     }
 
     public void checkForUpgradeAndExtraProperties() throws IOException {
-        if (!"".equals(systemEnvironment.getAgentMd5())) {
+        if (upgradesEnabled()) {
             checkForUpgradeAndExtraProperties(systemEnvironment.getAgentMd5(), systemEnvironment.getGivenAgentLauncherMd5(),
                     systemEnvironment.getAgentPluginsMd5(), systemEnvironment.getTfsImplMd5());
+        } else {
+            LOGGER.debug("[Agent Upgrade] Skipping check as there is no wrapping launcher to relaunch the agent JVM...");
         }
+    }
+
+    private boolean upgradesEnabled() {
+        return !"".equals(systemEnvironment.getAgentMd5());
     }
 
     private void checkForUpgradeAndExtraProperties(String agentMd5, String launcherMd5, String agentPluginsMd5, String tfsImplMd5) throws IOException {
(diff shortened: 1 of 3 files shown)


chadlwilson:
Hello @ketan - given your experience in this area, would you have any comments on whether this is the right approach to resolve this issue?

ketan:
Approach looks reasonable. Given the error is rare, it's probably reasonably OK to ignore, given it will eventually pass on the next try. Being explicit on the authentication filter is also fine. I'm also a bit curious, if we'll possibly see something similar when download the agent/plugin jars — should a similar filter be added on those URLs as well?

chadlwilson:
> Approach looks reasonable. Given the error is rare, it's probably reasonably OK to ignore, given it will eventually pass on the next try. Being explicit on the authentication filter is also fine.

Thanks! 🙏 Yeah, the NPE keeps getting brought up in various places (usually a red herring while people are trying to figure out other agent things), and it's confusing when trying to validate other changes to agent bootstrapping/startup/connectivity/work-getting, since an NPE is, err, not normal :-)

> I'm also a bit curious, if we'll possibly see something similar when download the agent/plugin jars — should a similar filter be added on those URLs as well?

Yeah, good thought.

I don't think they are affected since that part of downloading binaries happens in a different part of the logic in the `AgentLauncherImpl` (launcher download) and then the launcher's `AgentParentProcessImpl (agent/plugins/tfs-impl) ` which both use [single shot http clients via the ServerBinaryDownloader](https://github.com/gocd/gocd/blob/4968b50b4c1b7197eb2fd44ce46a8436de51d350/agent-common/src/main/java/com/thoughtworks/go/agent/launcher/ServerBinaryDownloader.java#L117-L138). Thus I think they do not "share" cookie storage or sessions or anything like that with the `Agent` itself.

Since they don't share anything or try to maintain connections, and there are no specific rules, they fall back to the below, where the `invalidateAuthenticationOnSecurityConfigChangeFilter` and `reAuthenticationWithRedirectToLoginPage` don't really do any "harm".
https://github.com/gocd/gocd/blob/3416442968dd0dae29aa393b35fd8d3718cbc178/server/src/main/java/com/thoughtworks/go/server/newsecurity/filterchains/AuthenticationFilterChain.java#L57

One could argue they should be added similarly - only reason I didn't was to reduce surface area for testing and make it easier to review.

PR #11323, opened 2023-02-20
Allow including paused or cancelled pipelines/stages in user dashboard filters

chadlwilson:
It's useful to be able to include `Cancelled` and `Paused` pipelines in such filtered views as 
- cancellation can often be considered a failed state when cancelled by the server (e.g due to timeout acquiring an agent, or a stuck build)
- paused pipelines can be a temporary thing and folks forget to unpause them so folks like to have a filtered view to quickly identify these

This does not change the default behaviour and is both backward and forward compatible (i.e if server is rolled back, filters saved with the added state still work, and just ignore the "cancelled" addition to the filter).

image

**As-is (for reference)**
image

### DIFF
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/DashboardFilter.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/DashboardFilter.java
@@ -29,8 +29,9 @@ public interface DashboardFilter {
 
     String DEFAULT_NAME = "Default";
     String BUILDING_STATE = "building";
+    String CANCELLED_STATE = "cancelled";
     String FAILED_STATE = "failing";
-    Set<String> VALID_STATES = Set.of(BUILDING_STATE, FAILED_STATE);
+    Set<String> VALID_STATES = Set.of(BUILDING_STATE, CANCELLED_STATE, FAILED_STATE);
 
     String name();
 
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/DashboardFilter.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/DashboardFilter.java
@@ -28,10 +28,7 @@ public interface DashboardFilter {
     }
 
     String DEFAULT_NAME = "Default";
-    String BUILDING_STATE = "building";
-    String CANCELLED_STATE = "cancelled";
-    String FAILED_STATE = "failing";
-    Set<String> VALID_STATES = Set.of(BUILDING_STATE, CANCELLED_STATE, FAILED_STATE);
+    Set<String> VALID_STATES = Set.of("paused", "building", "cancelled", "failing");
 
     String name();
 
--- a/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
+++ b/server/src/main/java/com/thoughtworks/go/server/domain/user/FilterValidator.java
@@ -21,7 +21,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
-import static com.thoughtworks.go.server.domain.user.DashboardFilter.*;
+import static com.thoughtworks.go.server.domain.user.DashboardFilter.VALID_STATES;
 
 class FilterValidator {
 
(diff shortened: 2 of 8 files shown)


arvindsv:
This is good. I can see it being useful.

Something I was thinking about is this change in `server/src/main/webapp/WEB-INF/rails/webpack/models/dashboard/dashboard_filter.js`: https://github.com/gocd/gocd/pull/11323/files#diff-33ab53f3c614d6411edcc4722b3bef3584543cd61fcfb7335c9c44d24e732c41

I know that "Building (Cancelled)" is a pseudo-state that GoCD can be in. That is, when a pipeline has been cancelled by the user or by the server, but one or more jobs have not finished cancellation. What happens in those states? If I create a filter for "Cancelled" pipelines only, will the one that is in "Building (Cancelled)" not show up, because Line 27 will pre-empt it?

If that works right, then this looks good to me. "Right" in my view is that the pipeline should show when looking for "Building" _or_ "Cancelled", meaning both filters for just "Building" and for just "Cancelled" should show it.

chadlwilson:
> I know that "Building (Cancelled)" is a pseudo-state that GoCD can be in. That is, when a pipeline has been cancelled by the user or by the server, but one or more jobs have not finished cancellation. What happens in those states? If I create a filter for "Cancelled" pipelines only, will the one that is in "Building (Cancelled)" not show up, because Line 27 will pre-empt it?
> 
> If that works right, then this looks good to me. "Right" in my view is that the pipeline should show when looking for "Building" _or_ "Cancelled", meaning both filters for just "Building" and for just "Cancelled" should show it.

Hmm, this is a very good point and I probably hadn't given enough thought to it. I thought "Building (Cancelled)" might only be displayed in the agents view model which necessarily works at `Job` scope, but I'm not 100% sure what you see on the dashboard when a stage is cancelled but the jobs haven't yet stopped.

I suspect the logic will need adjustment if a stage can be both "building" and "cancelled", similar to how it's treated differently for "paused" jobs, but need to dig into how the statuses come back to the UI for the dashboard view in particular.

Based on the below, it _looks like_ a stage is only considered "Building" OR "Cancelled" but not both at the same time from the perspective of the dashboard.

https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/server/src/main/webapp/WEB-INF/rails/webpack/models/dashboard/pipeline_instance.js#L31-L36

I believe the stage status is coming from the below which is derived from `JobResult` to determine cancellation, so think they will only change to `Cancelled` ones they have _actually completed the cancellation_.

https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/domain/src/main/java/com/thoughtworks/go/domain/StageState.java#L52-L82

chadlwilson:
> Based on the below, it looks like a stage is only considered "Building" OR "Cancelled" but not both at the same time from the perspective of the dashboard.

**TL;DR** - from perspective of the dashboard it doesn't seem possible to be in this pseudo-state; the stage is either considered Building OR Cancelled but not both (actually you immediately see "Cancelled" on dashboard while an agent is still processing the cancellation). I think this means this PR is OK, so will merge for now and see how we go in the real world on build.gocd.org. 🙏

**Longer Story**

I've tested this empirically locally by pausing the agent so it does not _process_ the cancellation.
- Almost immediately the server considers the stage cancelled and dashboard updates
  image
- Meanwhile from perspective of the agent we are in "Building (Cancelled)" limbo state
  image
- Agent Job Run History says it is cancelled
  image
- The Job itself looks cancelled for all intents and purposes (except its log isn't complete)
  image
  
Essentially this happens because the agent has not reported back that it is not building anymore and what you see on the agent page is actually based on the `AgentRuntimeStatus`.

https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/common/src/main/java/com/thoughtworks/go/domain/AgentRuntimeStatus.java#L22

When you cancel a stage, all jobs which agents are working on are cancelled and this status is immediately updated by the server to `Cancelled` which appears as "Building (Cancelled)" on the agent display.

https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/server/src/main/java/com/thoughtworks/go/server/ui/AgentViewModel.java#L83-L85

Eventually the agent will complete the cancellation, then report back to the server that it is Idle, along with any logs etc.

arvindsv:
Nice. Thanks for the investigation! Based on what you found, the decisions made in PR make sense.

PR #11483, opened 2023-04-13
Bump Alpine glibc to 2.35

chadlwilson:
This may or may not be a good idea/stable, but let's see.

Se https://github.com/sgerrand/alpine-pkg-glibc/issues/175 and https://github.com/sgerrand/alpine-pkg-glibc/issues/176 and https://github.com/sgerrand/alpine-pkg-glibc/issues/181 for the workaround on linking into `/lib64`


chadlwilson:
Currently this still leaves libc6-compat installed due to a dependency added in glibc-bin which is probably a risk. Probably not worth the risk here unless this is resolved upstream.

Issue #11629, opened 2023-06-03
Ability to store/retrieve session state within GoCD plugins to facilitate secure OIDC


JamesMcNee:
##### Issue Type

- Security Enhancement
- Feature proposal

##### Summary

For an OIDC (Open ID Connect) authentication flow to be truly secure, one must verify that the user who initiated the authentication flow is the same one from whom the callback is coming. In order to do this, OIDC has the concept of a `state` param, and often a `nonce`, these should be used by the client (the one implementing the flow) to check that the user remains consistent. The mechanism for doing this is storing some state against either the end user (e.g. Cookies, LocalStorage) or the session (server state) and then later verifying it.

Ideally when an OIDC flow is initiated a `state` and `nonce` value are generated and stored using one of these technologies against the current user, then on the callback, the IdP (Identity Provider) will pass these values back in the response. The OIDC client can then verify that the values match those that have been previously stored. This protects against CSRF attacks, [this Medium article is good at explaining further](https://medium.com/keycloak/the-importance-of-the-state-parameter-in-oauth-5419c94bef4c).

Currently the GoCD plugin API does not expose any mechanism in order to store state against the end user or the active session, therefore Authorization plugins that are using OIDC/OAuth are not able to do this additional validation step. Take the Google Authentication plugin as an example, although [it sets the state param on the flow](https://github.com/gocd-contrib/google-oauth-authorization-plugin/blob/789a26de9a448138304d5af50cc8d5a8885f7ce7/src/main/java/cd/go/authorization/google/GoogleApiClient.java#L66), it never verifies it, therefore not benefiting from the additional protection it provides.

##### Possible Fix

~My proposed fix is to add a mechanism onto the `GoPluginApiRequest` that will allow for access to a session context. This map will be scoped to the plugin, and will be backed by the HTTP Session, there would be a `String get(String key)` method and a `void set(String key, value)` method in order to store and retrieve context.~

Moved away from the above as not all requests to a plugin are attached to a HTTP request, so it feels a little to broad to add to the `GoPluginApiRequest` interface. 

Maybe something more localised to the Authentication plugin would work... Potentially if a response body is provided on the `go.cd.authorization.authorization-server-url` request, it could be stored against the session and then passed back as an additional property in the request body of `go.cd.authorization.fetch-access-token`.

This would allow for the OIDC flow to complete the verification and would make this change be localised to the Authorization plugin infrasturcture.

[...]

PR #11633, opened 2023-06-05
Add support for additional OAuth2/OIDC verification between redirects in authorization extensions

JamesMcNee:
Issue: #11629

Description:
Implements the ability for an Web based Auth plugin to store some sessions tate against the user between the `Get Redirect URL` and `Fetch Auth Token` flow. This is primarily to facilitate verifying the client state/nonce in an OIDC/OAuth2 flow.

Tested with a dummy plugin and behaves as expected!


chadlwilson:
FWIW the automated build and regression tests all passed. Bit of a pain making these more visible unfortunately.

image

chadlwilson:
For your information, the functional regression tests, in the context of authorization plugins appear to only
- test [bundled plugins such as the password file and LDAP authentication plugins](https://github.com/gocd/ruby-functional-tests/blob/6275afba96358889f8159b5318b6db77d692259b/specs/BundledAuthPlugins.spec)
  - they configure them via the UI and then validate logins, the latter using a test LDAP server running locally
- specific test the [authorization config single page app](https://github.com/gocd/ruby-functional-tests/blob/d9715b3824439880243786ba92116c6d75d75f17/specs/AuthConfigSPA.spec) (doesn't test actual login)

Generally speaking, most plugin variants are only really validated that plugins can load and be understood, and display correctly on the plugins page, then allow configuration on any of their dedicated UIs (e.g elastic agents can be configured, task plugins can then be used on jobs, secrets plugins can be used in builds to retrieve secrets etc)

I cannot see any specific regression test for a web-based/redirect authorization plugin (neither a "real" one, or a dummy plugin based off [the skeleton](https://github.com/gocd-contrib/authorization-skeleton-plugin)), which is quite a gap and seems the weakest automated validation of plugin variants.
- the authorization skeleton plugin is a bit out-of-date in general (tools, gradle version and such)
- it'd be "nice" if there was an easy way to validate a web based redirect oriented plugin like this by using a simple/dumb webserver or something like that which automatically redirects back (it wouldn't have to do "real" oauth2 to help us validate the whole flow, including perhaps session variable propagation, but seems there is nothing existing there 😬

JamesMcNee:
Hey @chadlwilson , thanks for confirming that everything passed 🎉 

I admit I did not actually look at the function tests, but agree that it is a shame there isn't a more comprehensive one for web based authentication plugins.

I think your idea around having a web server that acts as the IDP for the tests is sensible, are you thinking that you would like to have some additional tests in place before looking at this PR further?

chadlwilson:
Do you have any thoughts on how the forward/backward compatibility could be best handled in plugins? Is there an elegant way to do this that doesn't avoid rolling the plugin forward and changing its minimum GoCD version?

- One could include the session vars in the initial auth-server-url response and then see if they are supplied later, but it might be a bit weak here to only validate them if they exist (in case of bugs, or a server session that has been cleared for some reason)
- if the current GoCD server version was included in plugin requests the plugins could vary behaviour based on this, but is not there to my knowledge

JamesMcNee:
>Do you have any thoughts on how the forward/backward compatibility could be best handled in plugins? Is there an elegant way to do this that doesn't avoid rolling the plugin forward and changing its minimum GoCD version?

I did think about this a little yes, but did not come up with anything better than just targeting the latest version in the plugin 😓 The code should be backwards compatible in the sense that *old* plugins should be fine on new servers, but you're right that new plugins using this mechanism will not be compatible with old servers. It's likely worth at least calling this out in the docs thinking about it?

> if the current GoCD server version was included in plugin requests the plugins could vary behaviour based on this, but is not there to my knowledge

No, I do not think it is either. A plugin [can ask](https://plugin-api.gocd.org/current/authorization/#get-server-info) for `server-info` however this unfortunately doesn't include the current server version 😅, which is a shame.

> One could include the session vars in the initial auth-server-url response and then see if they are supplied later, but it might be a bit weak here to only validate them if they exist (in case of bugs, or a server session that has been cleared for some reason)

I agree it does feel a little weak, as you say the session may be cleared for whatever reason, it also feels like it also downplays the checking of these params on the whole as being a 'nice to do' check. Obviously this would be a choice for plugin authors to make, but I assume you're talking about the 'official' ones mainly.

JamesMcNee:
Oh, actually, it appears that the version is included in the V2 response (that is not documented): 

https://github.com/gocd/gocd/blob/de940cb3d466aff747b5dc2708da25d8e8f5cf67/server/src/main/java/com/thoughtworks/go/server/service/plugins/processor/serverinfo/v2/MessageHandlerForServerInfoRequestProcessor2_0.java#L39C60-L42

JamesMcNee:
Hey @chadlwilson -- Apologies for the radio silence, got distracted with some other things, although think this is with you now to look over when you have the capacity?

Let me know if I can do anything that would be useful e.g. Would it be useful to update the auth plugin skeleton project to be more up-to-date? Happy to help with this!

chadlwilson (approved):

chadlwilson:
This all looks pretty good to me. I can't think of/see any issues here.

In general the GoCD-side session gets completely cleared when transiting from "pre-auth" (while interacting with external authz server) to "post-auth" (after external authz server token exchange, and when GoCD is now happy with who you are) so probably not such a big deal manually clearing the metadata from the session so will leave it there :-)

Validated it with the previous GitHub Oauth plugin, and then implemented it myself within the same plugin and validated it with that too (https://github.com/gocd-contrib/github-oauth-authorization-plugin/pull/280) - GitHub seems to only support `state` since it's OAuth2 and not OIDC.

Will merge when the functional/regression tests pass again :-)

JamesMcNee:
@chadlwilson That's awesome, thanks for your support with this issue, and maintaining GoCD in general!

Shall look for further opportunities to contribute in the future too!

chadlwilson:
No problem @JamesMcNee - thanks for all the heavy lifting here.

I refreshed the general base dependencies of the https://github.com/gocd-contrib/authorization-skeleton-plugin however it doesn't include executors for the "web"-based authorization plugin requests yet anyway, so not sure if it's worth the effort there? If you're keen though, it would be welcome.

chadlwilson:
Hmm, just stumbled across these when looking across other plugins

- https://github.com/klinux/gocd-keycloak-oauth-authorization-plugin/blob/main/src/main/java/cd/go/authorization/keycloak/Constants.java (83,197 downloads over 2.5 years)
- https://github.com/szamfirov/gocd-okta-oauth-authorization-plugin/blob/master/src/main/java/cd/go/authorization/okta/Constants.java (2,785 downloads over 3 years)
- https://github.com/mgambini/crowd-integration-plugin/blob/master/src/main/java/it/ditech/go/authorization/crowd/Constants.java (663 downloads over 6 years)
- https://github.com/cnorthwood/gocd-tls-auth/blob/master/src/main/java/uk/me/cjn/gocd_tls_auth/TlsAuthorizationPlugin.java#L26 (29 downloads over 2.5 years)

I'll contact the maintainers for the first two, I think, rather than re-add the v1 API or add some other simpler compat logic.

JamesMcNee:
Ah damn, I didn't look at any of those 🙈The keycloak plugin seems to be widely used too (given the number of downloads).

Contacting the maintainers seems like a good approach, although some of those have not been updated in a few years so hopefully we are able to get some changes made.

I think if needed a V1 to V2 adapter would prove simpler than keeping all of the bespoke V1 mappers etc, especially given the two versions are not worlds apart.

chadlwilson:
I believe literally all that needs to be done by then is updating the version to `2.0` and the defaults will work fine as before. But yeah, they haven't been released for a while so there's that.

PR #11874, opened 2023-08-17
Ensure use of Spring TypeConverters is thread safe, and re-used

### LINKED ISSUE #11868: `runOnAllAgents="true"` added to jobs at GoCD startup
##### Issue Type

- Bug Report

##### Summary

GoCD surprisingly adds `runOnAllAgents="true"` to jobs.

##### Environment

👋  Let me know if you need this and I can share this individually, as it might contain sensitive information.

* Go Version: `23.1.0 (16080-54a6971915ff8d402c9fea8cd2ceeb6e31c8cdc8)`
* JAVA Version: `17.0.6`
* OS: `Linux 5.10.176+`

##### Steps to Reproduce

1. Restarting GoCD

##### Expected Results

1. GoCD config behaving like before.

##### Actual Results

1. GoCD surprisingly adds `runOnAllAgents="true"` to jobs.

##### Possible Fix

🤷 

Here is an example of a commit GoCD makes:

```diff
commit a29ab7eddd8a8d864f33a656f96ad1291abe53b7
Author: Upgrade <go-cd-dev@googlegroups.com>
Date:   Fri Aug 11 05:11:37 2023 +0000

    user:Upgrade|timestamp:1691730697644|schema_version:139|go_edition:OpenSource|go_version:23.1.0 (16080-54a6971915ff8d402c9fea8cd2ceeb6e31c8cdc8)|md5:536dedf85183254cabe2192eb1f7c5e2

diff --git a/cruise-config.xml b/cruise-config.xml
index 5234ae934..65a8023b5 100644
--- a/cruise-config.xml
+++ b/cruise-config.xml
@@ -34838,7 +34838,7 @@ fi</arg>
               <artifact type="test" src="some-tests/target/logs/docker" dest="logs" />
             </artifacts>
           </job>
-          <job name="some-tests">
+          <job name="some-tests" runOnAllAgents="true">
             <tasks>
               <fetchartifact artifactOrigin="gocd" srcfile="tag" stage="build" job="defaultJob" />
               <exec command="/bin/bash">
```

chadlwilson:
Likely fixes #11868

At some point TypeConverters were decided they should be re-used when loading config from XML; perhaps there was too much garbage being generated. In any case, these are NOT threadsafe. Perhaps there was only a single thread trying to use them at the time, but it certainly seems possible for multiple to try and do it now, and bizarre behaviour happens when doing so, such as not being able to convert a string 'false' to a primitive boolean and even returning null, which won't work.

Additionally, TypeConverters used when writing XML were NOT being re-used which seemed inconsistent. I don't know if they really need to be re-used/cached but probably better to do so consistently. Only risks here are that we have too many cached across threads (wasteful) or alternatively that we have ephemeral threads being created for these purposes, thus making the caching ineffectual, but I don't think either are true. If this turns out to be the case, could use a caching object pool or some other mechanism.

Also removed some low value/pointless tests.


chadlwilson:
Noting here for my own reference. When the server starts, there can definitely be a race between two threads trying to load the config from disk.

1) `main`/`WrapperJarAppMain` thread. `Upgrade` user. As soon as Spring finishes, `CachedGoConfig.upgradeConfig` starts to migrate the config to latest version, if necessary. This is the first thing to start. It then loads the config again from disk.
  https://github.com/gocd/gocd/blob/0f58107c851cf2df6ce7c6902eebde796dc1f742/server/src/main/java/com/thoughtworks/go/server/initializers/ApplicationInitializer.java#L112-L113
2) `goThreadPoolTaskScheduler-x` thread. When Spring starts, it starts an "every 5 seconds" check of the modified time and md5 of the config on disk.
  https://github.com/gocd/gocd/blob/aca340bb65e4472ad9b0a3d1031e0d79656e12b5/server/src/main/resources/spring-cruise-remoting-servlet.xml#L62-L64

Sometimes (1) is not complete before (2) is loading config. 

These two mechanism use different locks. Not sure if intentional, but maybe to avoid deadlocks. The code is confusing.
1) Locks on `CachedGoConfig` object monitor which represents the config state the rest of the server uses, and locks when swapping cached config, when doing migrations, when writing through UI-driven config mutations to disk and config history etc.
2) Locks on `GoConfigWriteLock.class` first to load the config, and only locks on `CachedGoConfig` when it needs to swap the config. `GoConfigWriteLock` is the outer lock and must be acquired first.

Consequence is that
- sometimes migration is "slow" and loading happens concurrently between these two threads.
- on a server upgrade, it seems **theoretically** possible that the "every 5 sec reload" could swap the `CachedGoConfig` into memory with a "stale" config which has not been "upgraded".
  1 - main thread starts loading config from disk to upgrade it
  2 - scheduled thread starts loading config from disk to cache it
  3 - main thread re-saves any changes (if any) to disk and cached config
  4 - scheduled thread caches config loaded earlier <-- without upgrade changes, dirty read is now cached
  5 - _5 seconds later_ the scheduled thread **might** fix the cached config as the md5 will have changed on disk from main thread's write. Might need to check the way this works in `GoConfigFileDataSource`.

This could be avoided by ensuring `CachedGoConfig.upgradeConfig` also acquires the `GoConfigWriteLock.class` and holds it until after it has  finished, thus ensuring the periodic config checks don't start until the server is fully loaded. But it's hard to keep track of the possible deadlocks here, so will leave it for now.

PR #12051, opened 2023-10-01
Remove subversion & mercurial from container images

chadlwilson:
All of the current container images (server and agent) include both subversion and mercurial, which GoCD does have support for.

However, these SCMs are very infrequently used, and including them does bring in a lot of other dependencies, including perl and python, increasing the security/attack surface area of the images.

This change removes these by default, leaving only git included by default in the images. Users who need these SCMs are suggested to build their own images which layer on the required SCMs.

This is estimated to reduce the image size by ~60MB (uncompressed), but will update these with the end results after official builds.

### DIFF
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/docker/Distro.groovy
@@ -49,7 +49,7 @@ enum Distro implements DistroBehavior {
     List<String> getInstallPrerequisitesCommands(DistroVersion v) {
       return [
         // procps is needed for tanuki wrapper shell script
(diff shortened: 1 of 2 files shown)


chadlwilson:
Changes in compressed image size

* Server Alpine -25MB
* Server Centos 9 -26MB
* Agent Alpine DIND -25MB
* Agent Alpine 3.18 -25MB
* Agent Alpine 3.17 -24MB
* Agent Alpine 3.16 -24MB
* Agent Alpine 3.15 -23MB
* Agent Debian 12 -22MB
* Agent Debian 11 -19MB
* Agent Debian 10 -16MB
* Agent Centos 9 -27MB
* Agent Centos 8 -17MB
* Agent Ubuntu 22 -19MB
* Agent Ubuntu 20 -16MB

PR #12202, opened 2023-11-02
Replace nanohttpd server with JDK builtin httpserver

ravikp:
Context: There were no updates to [NanoHttpd](https://github.com/NanoHttpd/nanohttpd) server since couple of years. Moreover, JDK has built-in httpserver, that can be used for the similar purpose. Following are the changes introduced in this PR:

* Re-write agent status endpoint using NanoHttpd to in-built jdk provided http server from 'com.sun.net.httpserver' package.
* Re-write test cases to adapt to the change introduced by removing NanoHttpd references


agent/src/test/java/com/thoughtworks/go/agent/statusapi/AgentStatusHttpdTest.java
```diff
@@ -37,80 +50,169 @@ class AgentStatusHttpdTest {
     private AgentHealthHolder agentHealthHolder;
     @Mock
     private SystemEnvironment systemEnvironment;
-    @Mock
-    private NanoHTTPD.IHTTPSession session;
+
     private AgentStatusHttpd agentStatusHttpd;
 
+    private String hostName = "0.0.0.0";
+    private int port = 8093;
```
chadlwilson, line 57: I understand that when mocked correctly the server should not **actually** start, but can we change this to `127.0.0.1` or `localhost` so that it fails a bit more gracefully if the mocking needs changes in future (won't trigger firewall noise)?

ravikp: Can be done. Initially when writing the code, I have used loopback address, later I changed to the most generic. Let me change to loopback.

ravikp: Change committed.

chadlwilson (approved):
LGTM! Thanks!

Tests are running now over at https://build.gocd.org/. Sadly there's not a great way to provide easy access to the PR pipelines right now - need to fix that.

Are there any of these config defaults that you feel we should tweak for these purposes (either for compatibility with nanohttpd or just to reduce chance of misbehaving status checkers overloading the status check port/dispatcher thread etc)?

https://github.com/openjdk/jdk17u/blob/57bf412983c106b2cc35ff522a7582278c4ffecf/src/jdk.httpserver/share/classes/sun/net/httpserver/ServerConfig.java#L38-L129

chadlwilson:
Merged now, can think about the defaults separately :-)

chadlwilson:
This is now running on https://build.gocd.org for what it's worth. 😅

PR #12212, opened 2023-11-03
fix: templates in alphabetical order

jales-els:
When you have a long list of templates it's hard to find the one you need. Having the list sorted alphabetically helps with this 

image

image


chadlwilson:
Thanks for this. I'm sympathetic to this change.

Currently I believe the order is deterministic but unhelpful, just that it's probably 'configuration definition order' in the underlying config file. Which roughly speaking is usually order or creation (similar to pipeline groups), but I'm not even sure that's 100% true in all cases 😅

We've need to double check that this doesn't have unintended consequences when editing/deleting/adding/cloning templates (probably shouldn't) and add some kind of unit test or assertion.

chadlwilson:
I assume it was intentional at this stage that you have only attempted to sort the dropdown selector rather than the admin > templates list in the second screenshot?

Although I do note that this list's order seems kind of random right now, it's not even the creation order of the templates it seems - so would consider a PR for that too, as long as it seems broadly consistent with behaviour elsewhere.

PR #12356, opened 2023-12-10
Rollback selenium-webdriver to 4.15.0

chadlwilson:
Seems to be some issue with 4.16.0 on Windows, to investigate later.


chadlwilson:
See https://github.com/SeleniumHQ/selenium/issues/13261

PR #12422, opened 2024-01-02
Update README-running-multiple-agents.md

brewpark:
hello. I use multiple agents in single machine.
when I use version 20.4.0, script worked fine.
but in version 23.4.0, it now doesn't work.

### changes
- change mode of /usr/share/${AGENT_ID} same as /usr/share/go-agent.

### details

i got an error like below when run the go-agent-1/bin/go-agent

```
./go-agent console
-bash: line 1: /usr/share/go-agent-1/bin/go-agent: Permission denied
Failed to access the script using an absolute path. Insufficient permissions may prevent the user "go" from traversing one of the folders. Please check the following permissions:
drwxr-xr-x.  12 root root   144 Dec 15 10:49 /usr
drwxr-xr-x. 108 root root 16384 Jan  2 17:06 /usr/share
drwx------.   4 root root    76 Jan  2 19:05 /usr/share/go-agent-1
drwxr-xr-x.   2 root root    22 Jan  2 18:59 /usr/share/go-agent-1/bin
```

then i check the mode of /usr/share/go-agent and go-agent-1,
i found they have different mode.

```
drwxr-xr-x.   7 root root    76 Jan  2 15:24 go-agent
drwx------.   4 root root    76 Jan  2 19:05 go-agent-1
```

after I set directory's mode, it works fine.

### DIFF
--- a/installers/go-agent/release/README-running-multiple-agents.md
+++ b/installers/go-agent/release/README-running-multiple-agents.md
@@ -53,6 +53,7 @@ Each agent needs to be configured to run in its own separate directory so that e
         # can write to these directories
         chown -R go:go /var/{lib,log}/${AGENT_ID}
         chmod -R 0750 /var/{lib,log}/${AGENT_ID}
+        chmod 0755 /usr/share/${AGENT_ID}
 
         # tweak the scripts and configs to use the correct directories
         sed -i -e "s@go-agent@${AGENT_ID}@g" /usr/share/${AGENT_ID}/bin/go-agent


chadlwilson:
Was your original `/usr/share/go-agent` install an upgrade of a previous install, or a freshly created one? Which mechanism did you use to install (rpm? deb? something else?) Are you trying to run `./go-agent console` as the `go` user, or a different user?

I'm not quite sure why this would be any different to earlier versions, but it's not really possible to comment on the differences over 3+ years of installers without a bit more detail on how you installed.

In any case, I don't quite like the idea suggesting folks change the permissions to all users (`0755` vs `0750`) without understanding exactly how we got there :-)

brewpark:
@chadlwilson 

It's a freshly created one. (machine and go-agent)
i use amazon linux2023(fedora).

and I installed with yum. with command below
```
sudo curl https://download.gocd.org/gocd.repo -o /etc/yum.repos.d/gocd.repo
sudo yum install -y go-agent
```
ref : https://docs.gocd.org/23.4.0/installation/install/agent/linux.html#rpm-based-distributions-ie-redhatcentosfedora

after install (i tried on the other machine)
`drwxr-xr-x.   7 root root    76 Jan  3 10:19 go-agent` in /usr/share.
it has `0755` mode and own by root:root 

I ran `/usr/share/go-agent-1/go-agent console` with root user. 
because `go` user didn't have permission of `/usr/share/go-agent-1`

thanks for your help.

chadlwilson:
@brewpark OK, thanks for clarifying - that helps. And also thanks for raising a PR to try and improve things for everyone!

Let me double-check in an AL2 container and see if there is a more elegant way the script can express the intent to match the permissions from the official package, in case there are any other unexpected differences.

PR #12635, opened 2024-03-22
Suggested redesign for the error.html page.

RedYetiDev:
This PR includes a suggested redesign for the `error.html` page. In the new page, a **Back to Home** link has been added, and the circle layout has been replaced with a rounded rectangle

[Old Page](https://refined-github-html-preview.kidonng.workers.dev/gocd/gocd/raw/master/jetty/src/main/resources/error.html)
[New Page](https://refined-github-html-preview.kidonng.workers.dev/redyetidev/gocd/raw/patch-1/jetty/src/main/resources/error.html)

### DIFF
--- a/jetty/src/main/resources/error.html
+++ b/jetty/src/main/resources/error.html
@@ -1,46 +1,66 @@
-
-<html>
+<!DOCTYPE html>
+<html lang="en">
 <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


chadlwilson:
Hiya @RedYetiDev - thanks for the suggestion. The current error page is broadly conforming to existing GoCD design language and colours (for whatever that means), as odd as they might be and as imperfect as this page is.

Are there specific things you're trying to achieve with a different design?
- Including an "oops" message?
- Including a link to go back home?

RedYetiDev:
Mostly I thought the circle layout was not as 'pretty' as it could be, and I thought a rounded rectangle would be nicer. While I was designing it, I figured a 'go back' button would also help, and the Oops! message followed shortly after.

Issue #11354, opened 2023-03-05
Provide arm64 arch-specific deb and rpms with correctly bundled JRE


chadlwilson:
##### Issue Type

- Feature enhancement

##### Summary

After the implementation of #8544 in `23.1.0`
- the Go Server and Agent are properly validated against arm64 architecture via docker images for both Debian and CentOS
- wrapper binaries for arm64 are included within the `deb` and `rpm`s built.

This means users _can use the installs on most distros to run arm64_ **ONLY IF** the user supplies their own arm64 JRE or JVM and overrides `wrapper.java.command` in `/usr/share/go-(server|agent)/wrapper-config/wrapper-properties.conf` to tell the wrapper where to find an arm64 JRE they have installed elsewhere.

Due to an accident of history, an x64 JRE is bundled with both the `deb` and the `rpm` despite them marked marked as `all` arch and `noarch` respectively. 

##### Expected Results

Running debs or rpms on arm64 come with a correct-arch pre-packaged JRE or multi-arch package dependency (as with x64), rather than requiring a separate manual JRE install and any need to manually override the JRE in the config.

##### Actual Results

Currently an x64 JRE is bundled, which is wasteful and unusable on arm64 anyway with a separate JRE/JDK install provided by the user.

##### Possible Fix

Possibly we could either
1. bundle an arm64 JRE into the same deb/rpm at cost of massively increasing the size, and undermining the intention
2. produce aarch64/arm64 specific installers which bundle a valid JRE version (similar to the docker images)
3. keep the existing `noarch`/`all` top level variants, but extract the JRE (and optionally other) arch-specific bits into a new dependent package
    - The idea of this would be to smooth the upgrade path as I am not sure if `dnf` etc will transparently handle an upgrade for a version while switching arch, e.g from say `24.1.0-noarch` --> `24.2.0-aarch64` ?
4. One might argue it'd be better to stop bundling the JRE in these specific packages and add a dependency to some other standard JDK/JRE package and reconfigure the wrapper config to point to it rather than the bundled JRE. (e.g `java-17-openjdk-headless` on RHEL/CentOS, or `openjdk-17-jre-headless` on Debian/Ubuntu).
5. arguably stop publishing the `noarch`/`all` arch variants entirely since they are "incorrect".

chadlwilson:
At this stage I am leaning towards **option 4**, but would need to dig a bit into the history of why the JRE was bundled, and the size/contents of these off-the-shelf packages.

[...]

Issue #10722, opened 2022-08-21
GoCD RPMs cannot be verified/installed with default RHEL 9 / CentOS Stream 9 configuration


chadlwilson:
##### Issue Type

- Bug Report

##### Summary

Currently it is not possible to install the GoCD RPMs on RHEL 9 and Centos Stream 9 - and other derived repos such as Rocky, Alma, Fedora. 

- RHEL 9/10 ❌
- CentOS Stream 9/10 ❌
- Rocky Linux 9/10 ❌
- AlmaLinux 9/10 ❌
- Fedora 36 + 37 ✅

Fedora 39+ will break also according to their [plans](https://www.redhat.com/en/blog/rhel-security-sha-1-package-signatures-distrusted-rhel-9) - it should be OK on Fedora 36-38 at the moment as they haven't yet changed policies to match the upstream RHEL/CentOS.

This appears to be because the GoCD GPG key includes at least one signature on a `SHA-1` hash which are [deprecated on RHEL 9 and friends.](https://www.redhat.com/en/blog/rhel-security-sha-1-package-signatures-distrusted-rhel-9) and removed in the default crypto policies.

We'd need to ensure all signatures are based on SHA256 digests to resolve this.

##### Environment

Tested with CentOS Stream 9

##### Steps to Reproduce

1.  `docker run -it quay.io/centos/centos:stream9`
2.  Follow instructions at https://docs.gocd.org/current/installation/install/server/linux.html#rpm-based-distributions-ie-redhatcentosfedora
3. `curl https://download.gocd.org/gocd.repo -o /etc/yum.repos.d/gocd.repo`
4. `dnf install -y go-agent`

```
Importing GPG key 0x8816C449:
 Userid     : "ThoughtWorks GoCD (GoCD Code signing keys) <support@thoughtworks.com>"
 Fingerprint: 3222 59C8 2D30 82B3 E32A EC2E D884 3F28 8816 C449
 From       : https://download.gocd.org/GOCD-GPG-KEY.asc
warning: Signature not supported. Hash algorithm SHA1 not available.
Key import failed (code 2). Failing package is: go-agent-22.2.0-14697.noarch
 GPG Keys are configured as: https://download.gocd.org/GOCD-GPG-KEY.asc
```

This is also the same error when trying to build GoCD itself during attempt to sign on CentOS Stream 9 (example [here](https://build.gocd.org/go/tab/build/detail/code-sign/1029/sign-and-upload/1/rpm))

##### Expected Results

[...]

PR #12677, opened 2024-04-07
Fixes #4011

### LINKED ISSUE #4011: RPM packages do not install on RHEL 7 unless SELinux is disabled
##### Issue Type

- Enhancement

##### Summary

RPM packages do not install on RHEL 7 unless SELinux is disabled

##### Environment

```plain
# cat /etc/system-release
Red Hat Enterprise Linux Server release 7.4 (Maipo)
```
###### Related issues

* https://github.com/gocd/gocd/issues/1833 Official package fails to install on redhat
* https://github.com/gocd/gocd/issues/1579 Official package fails to install on CentOS
* https://github.com/gocd/gocd/issues/1527 CentOS 7 go user isn't created
* https://github.com/gocd/gocd/issues/1467 Centos7 useradd fails during install

##### Workaround

    vi /etc/sysconfig/selinux

change to

    SELINUX=permissive

reboot

##### Expected result

The gocd rpm package plays nice with SELinux enabled.

##### Log snippets

```
[root@jd-server7x ~]# sudo yum install -y go-server
Loaded plugins: product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
Resolving Dependencies
--> Running transaction check
---> Package go-server.noarch 0:17.11.0-5520 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

==================================================================================================================================================================================
 Package                                     Arch                                     Version                                        Repository                              Size
==================================================================================================================================================================================
Installing:
 go-server                                   noarch                                   17.11.0-5520                                   gocd                                   137 M

Transaction Summary
==================================================================================================================================================================================
Install  1 Package

Total download size: 137 M
Installed size: 137 M
Downloading packages:
warning: /var/cache/yum/x86_64/7Server/gocd/packages/go-server-17.11.0-5520.noarch.rpm: Header V4 RSA/SHA1 Signature, key ID 8816c449: NOKEY===-] 4.4 MB/s | 136 MB  00:00:00 ETA 
Public key for go-server-17.11.0-5520.noarch.rpm is not installed
go-server-17.11.0-5520.noarch.rpm                                                                                                                          | 137 MB  00:00:28     
Retrieving key from https://download.gocd.io/GOCD-GPG-KEY.asc
Importing GPG key 0x8816C449:
 Userid     : "ThoughtWorks GoCD (GoCD Code signing keys) <support@thoughtworks.com>"
 Fingerprint: 3222 59c8 2d30 82b3 e32a ec2e d884 3f28 8816 c449
 From       : https://download.gocd.io/GOCD-GPG-KEY.asc
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
useradd: cannot create directory /var/go
error: %pre(go-server-17.11.0-5520.noarch) scriptlet failed, exit status 12
Error in PREIN scriptlet in rpm package go-server-17.11.0-5520.noarch
  Verifying  : go-server-17.11.0-5520.noarch                                                                                                                                  1/1 

Failed:
  go-server.noarch 0:17.11.0-5520                                                                                                                                                 

Complete!
```

matthiaskraaz:
I took the same approach as https://github.com/hashicorp/consul/issues/8510 was solved:

- removed the -m option from useradd so it does not try to create the directory
- added /var/go to directories to be created by package

### DIFF
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/InstallerTypeAgent.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/InstallerTypeAgent.groovy
@@ -80,7 +80,8 @@ class InstallerTypeAgent implements InstallerType {
       '/var/lib/go-agent'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
       '/var/lib/go-agent/run'             : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
       '/var/log/go-agent'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
-      '/var/run/go-agent'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true]
+      '/var/run/go-agent'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
+      '/var/go'                           : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
     ]
   }
 
--- a/buildSrc/src/main/groovy/com/thoughtworks/go/build/InstallerTypeServer.groovy
+++ b/buildSrc/src/main/groovy/com/thoughtworks/go/build/InstallerTypeServer.groovy
@@ -85,6 +85,7 @@ class InstallerTypeServer implements InstallerType {
       '/var/log/go-server'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
       '/var/run/go-server'                 : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
       '/etc/go'                            : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
+      '/var/go'                            : [mode: 0750, owner: 'go', group: 'go', ownedByPackage: true],
     ]
   }
 
--- a/installers/linux/shared/before-install.sh.erb
+++ b/installers/linux/shared/before-install.sh.erb
@@ -17,4 +17,4 @@
 -%>
 
     getent group go > /dev/null || groupadd -r go
-    getent passwd go > /dev/null || useradd -r -c "GoCD User" -g go -d /var/go -m -s /bin/bash go
+    getent passwd go > /dev/null || useradd -r -c "GoCD User" -g go -d /var/go -s /bin/bash go


chadlwilson:
Thx!


Did you happen to check 
- whether this has any difference in effect for `.deb`s on Ubuntu or Debian? IIRC the same script is used for both - I naively assume that the SELinux problems might be similar on both platforms.
- whether similar change needs to be made in `InstallerTypeAgent.groovy` (believe it does, as I imagine the agent installation has the same problem and `before-install.sh.erb` is used for both agent and server packages)

matthiaskraaz:
Oops. Let me check...

matthiaskraaz:
I added InstallerTypeAgent.groovy to the PR and also did a lot of testing.

### Ubuntu 22.04

I didn't succeed in enabling SELinux. 
According to https://askubuntu.com/questions/481293/selinux-implementation-in-ubuntu SELinux on Ubuntu is dead.
Is it worth investing more time to get an Ubuntu with SELinux running?
I tested my changes without SELinux.

- install agent
- connect agent to different host
- create and run pipeline
- install server
- connect agent to local server
- create and run pipeline
- fresh install
- install server
- connect agent from different host
- create and run pipeline
- install agent locally
- connect local agent to server

Result: changed deb works fine under Ubuntu 22.04 without SELinux

### Debian 12 without SELinux

- fresh install
- installed changed server deb
- installed git
- connected agent from different host
- created and run pipeline
- installed changed agent deb locally
- connected agent to server
- run pipeline
- fresh install
- installed changed agent dep
- installed git
- connect agent to different host
- created and run pipeline
- installed changed server deb locally
- connected agent to server
- run pipeline

Result: changed deb works fine under Debian 12 without SELinux

### Debian 12 with SELinux

Fresh install and activated SELinux then:
# apt update
# apt install selinux-basics selinux-policy-default auditd
# selinux-activate
# reboot
# check-selinux-installation
# setenforce 1
(Source: https://wiki.debian.org/SELinux/Setup)

Server fails to start:

```
INFO   | jvm ver. | 2024/04/07 15:03:06 | Error occurred during initialization of VM
INFO   | jvm ver. | 2024/04/07 15:03:06 | Failed to mark memory page as executable - check if grsecurity/PaX is enabled
ERROR  | wrapper  | 2024/04/07 15:03:06 | Failed to parse the version of Java. Resolving to the lowest supported version (1.4).
```

Tried also with official go-server_23.5.0-18179_all.deb. Server fails to start again with above error.

What do you think?

chadlwilson:
Wow, thanks for the detail here - that's far above and beyond!

If there are other problems with Debian SELinux (looks like the Tanuki wrapper's Java/JAVA_HOME detection isn't working for some reason we don't have to solve them here. Once we sort this our for RHEL, we can open a separate issue and see if anyone complains/cares. Part of me has been thinking about dropping Debian containers and raw Debian installs for official support anyway, as I don't hear much noise about it.

Even RHEL 9 / CentOS Stream 9 has issues right now due to #10722 and needs workarounds.

Anyway, couple of other things I'd like to check 
- I think the default **directory perms** are different when created by `useradd` vs here. I believe normally the group (`go`) has no default read permissions to the created home dir, i.e `0700` (or `cat /etc/login.defs | grep HOME_MODE` rather than `0750`. Whether this matters I am not sure, I can't recall if anything in particular is stored here by default (I think the relevant stuff is in `/etc/go`), or whether it's just for scripts/shells/ssh credentials etc. Am I missing something here? Do you think we should use something more restrictive by default?
- Do you anticipate if it makes any difference to **removal** of the package? I believe the current approach will leave `/var/go` behind because the directory is not package-owned. I don't know if it's a pro/con for package/removal to have this folder automatically cleaned up. Currently most important stuff is in `/etc/go` which **is cleaned up**, so probably doesn't matter to also clean up this dir and is maybe a pro. But if people have put SSH keys there for, e.g ssh git access, we would remove them on uninstall I suppose.

matthiaskraaz:
Let me check, what the old and new behavior of rpm and deb is regarding permissions of the dir. I suggest to keep the old behavior.

I believe that files not owned by the package are not removed even if in a directory owned by the package. Uninstall may give a warning though. But let me check.

chadlwilson:
To safe you some empirical checking, `0700` is the default on CentOS Stream 9 Minimal coming from `cat /etc/login.defs | grep HOME_MODE`, but I didn't check RHEL or earlier versions. To some extent the previous behaviour is dependent on distro-default, `umask`s etc I understand so we may not be able to match all, but `0700` seems sensible for a home dir, even with go using a dedicated `group` generally (and thus low risk of group readability being a problem).

When I checked the uninstall on CentOS Stream 9 with `23.5.0`'s rpm it did remove `/etc/go` entirely however I hadn't actually created files outside the package/started the server etc, so that might not have been a valid test. (excuse my ignorance, the intriciacies of rpms/debs etc is not my strong suit!)

Separately to this, I am impressed you seem to have managed to get a build going locally. If you had any major friction or feedback on the process of doing so, or the stuff at https://developer.gocd.org/current/ please raise issues/complaints/feedback and I'll fix it. Generally getting to the point of getting the installer builds going with `fpm` etc locally is even more painful than the rest of the build, and there are no specific instructions for linux where that stuff is actually easier, so good for you!

matthiaskraaz:
Thank you. And thank you for keeping GoCD alive.

Regarding the permissions: I think explicit `0700` has a very, very slight chance of breaking something somewhere, explicit `0750` has a very, very slight chance of being a security problem. Because it only matters if someone has actually added a user to the go group. Please make the call. You have the experience in that matter.

### Ubuntu 22.04 and Debian 12

apt purge:
- outputs: `dpkg: warning: while removing go-agent, directory '/var/go' not empty so not removed`
- keeps directory /var/go and file within

apt remove: 
- no warning
- keeps directory /var/go and file within

### CentOS 9

`rpm -e`:
- keeps directory /var/go and file within

I don't think rpm has a "purge" option.

### Build

I used WSL with Ubuntu. I had to

```
sudo apt install openjdk-17-jdk-headless
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
source ~/.bashrc
nvm install --lts
corepack enable
sudo apt install ruby-rubygems
sudo gem install fpm
sudo apt install rpm
```

I also had to 

`./gradlew --stop`

because nvm fiddling with the environment doesn't propagate to a running daemon.

I tried Java 21 first, but that gave me endlessly `waiting for isolate _isolateMain to check in` when trying to build.

matthiaskraaz:
OTOH, even the chrony package under CentOS removes /var/lib/chrony, the home directory of the chrony user.

chadlwilson:
OK, let's stick with this, should be fine I think!

> ### Build
> 
> I used WSL with Ubuntu. I had to
> 
> ```
> sudo apt install openjdk-17-jdk-headless
> curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
> source ~/.bashrc
> nvm install --lts
> corepack enable
> sudo apt install ruby-rubygems
> sudo gem install fpm
> sudo apt install rpm

Yeah, that looks about right, that's essentially what the build agents for GoCD itself do to pre-install.

https://github.com/gocd-contrib/gocd-oss-cookbooks/blob/a4dd63c7673963aac1c6a4023b764a70ed00400b/provision/provision-centos.sh#L148-L162

> ```
> 
> I also had to
> 
> `./gradlew --stop`
> 
> because nvm fiddling with the environment doesn't propagate to a running daemon.

Yeah, that's Gradle-life. Urggh.

> I tried Java 21 first, but that gave me endlessly `waiting for isolate _isolateMain to check in` when trying to build.

Yeah, I haven't validated either the build or GoCD itself with Java 21 yet, but that's a weird one - I wonder where that error is coming from, as it looks like a Dart error, and I can't think of a Java 21 interaction with Dart like that.... (if I recall correctly, only the sass compiler forked within the builds is built with Dart)

chadlwilson:
For what it's worth, the main build pipeline does run some rpm/deb driven installer tests which dont run on pull requests (due them needing to be post-signing and such), so when it has flowed through I'll let you know. It's the `installer-tests` pipeline at https://build.gocd.org/go (login with the `guest` button if you're interested)

If you feel like it, you can play with the experimental builds too :-)

matthiaskraaz:
My PR is just building. How exciting! :-)

chadlwilson:
_Experimental_ builds from https://www.gocd.org/download/#redhat `24.1.0-18170`+ have your fix in them; the installer tests were fine (it only tests a fresh install and an upgrade on CentOS Stream 8 of the RHEL adjacent distros)

`dnf install -y go-server-24.1.0-18710`

https://build.gocd.org/go/tab/build/detail/installer-tests/2410/install-tests/1/fresh-install-runInstance-4
https://build.gocd.org/go/tab/build/detail/installer-tests/2410/install-tests/1/upgrade-runInstance-4

chadlwilson:
Frontend isn't my strong suit either.

I'm not sure, depends whether you're looking to fix a bug, implement a feature, or experiment with finding an upgrade path for some technical debt/outdated libraries/tools?

It's not the most critical thing, but along the lines of this one, if you know something about rpms and debs and have an idea how we might approach #11354 at reasonable cost with the current repository infrastructure we have, that might be interesting?

matthiaskraaz:
LOL. I had this in my notes, because I tried to reproduce #4011 on an ARM instance first: 
_go-server-23.5.0-18179.noarch.rpm. says noarch. but contains x86-64 jre._

I'm on it.

Issue #10262, opened 2022-03-13
Upgrade to Hibernate 5


chadlwilson:
Currently GoCD uses Hibernate 3.6. In order to unblock Spring Framework upgrade to Spring 5+, we need to move to to Hibernate 5. Our current Spring version,  `4.3` does support at least Hibernate `5.0` to `5.2` (see [here](https://docs.spring.io/spring-framework/docs/4.3.x/spring-framework-reference/html/new-in-4.3.html)) and maybe later point releases.

If we could get to `5.3` we'd be getting back into a "supported" release, which would be ideal, as long as it works with Spring 4.3. There's nothing in [the Spring docs](https://docs.spring.io/spring-framework/docs/4.3.30.RELEASE/spring-framework-reference/htmlsingle/#orm-hibernate) to indicate other 5.x versions wont work.

https://hibernate.org/orm/releases/

* [4.0](https://developer.jboss.org/docs/DOC-16473)
* [5.0](https://github.com/hibernate/hibernate-orm/blob/5.0/migration-guide.adoc)
* [5.1](https://hibernate.org/orm/documentation/5.1/migration/)
* [5.2](https://github.com/hibernate/hibernate-orm/blob/5.2/migration-guide.adoc)
* [5.3](https://github.com/hibernate/hibernate-orm/blob/5.3/migration-guide.adoc)
* [...others](https://github.com/hibernate/hibernate-orm/wiki/Migration-Guides)

PR #12712, opened 2024-04-19
Replace bundled deb/rpm x64 JRE with arch-independent package deps

### LINKED ISSUE #11354: Provide arm64 arch-specific deb and rpms with correctly bundled JRE

matthiaskraaz:
Removed x86 JRE from architecture-independent package. Instead, added dependency on openjdk-17-jre-headless to DEB and dependency on jre-17-openjdk-headless to RPM. Fixes #11354.

Tested installation of go-agent and go-server.

Tested under:
- CentOS Stream 9
- CentOS Stream 8
- CentOS 7
- Ubuntu 22.04
- Ubuntu 20.04
- Debian 12
- Debian 11
- Debian 10

Each OS both in the x86_64 variant and arm64 variant.

Result: all ok except for:
- CentOS Linux 7: jre-17-openjdk-headless not available
- Debian 10: openjdk-17-jre-headless not available

### DIFF
--- a/installers/linux.gradle
+++ b/installers/linux.gradle
@@ -33,6 +33,8 @@ enum PackageType {
       // for `ps`
       fpmArgs += ['--depends', 'procps']
 
+      fpmArgs += ['--depends', 'openjdk-17-jre-headless']
+
       // HACK: for debian packages :(, since manifests cannot contain fine grained ownership
       def tmpDir = project.file("${project.buildDir}/tmp")
       tmpDir.mkdirs()
@@ -60,6 +62,8 @@ enum PackageType {
 
       // for `ps`
       fpmArgs += ['--depends', 'procps']
+
+      fpmArgs += ['--depends', 'jre-17-openjdk-headless']
 
       fpmArgs += ['--rpm-defattrfile', '0444']
       fpmArgs += ['--rpm-defattrdir', '0755']
@@ -96,22 +100,9 @@ private File destFile(String url) {
   new File(gradle.gradleUserHomeDir, "download-cache/${DigestUtils.md5Hex(url)}/${new File(new URL(url).path).name}")
 }
 
-task downloadLinuxJreChecksum(type: DownloadFile) {
-  src project.packagedJavaVersion.toSha256SumURLFor(OperatingSystem.linux, Architecture.x64)
-  dest destFile(src.toString())
-}
-
-task downloadLinuxJre(type: DownloadFile) {
-  dependsOn downloadLinuxJreChecksum
-  src project.packagedJavaVersion.toDownloadURLFor(OperatingSystem.linux, Architecture.x64)
-  dest destFile(src.toString())
-  checksum = { downloadLinuxJreChecksum.outputs.files.singleFile.getText("utf-8").trim().split(" ").first() }
-}
-
 def configureLinuxPackage(DefaultTask packageTask, InstallerType installerType, PackageType packageType, File buildRoot, Configuration configuration, File destDir) {
   packageTask.with {
     dependsOn configuration
-    dependsOn downloadLinuxJre
     dependsOn extractDeltaPack
     outputs.dir(destDir)
 
@@ -225,7 +216,10 @@ def configureLinuxPackage(DefaultTask packageTask, InstallerType installerType, 
             }
 
             if (eachLine == 'wrapper.java.command=java') {
-              eachLine = "wrapper.java.command=/usr/share/${installerType.baseName}/jre/bin/java"
+              // Debian Package post install script sets command depending on the architecture
+              if (packageType == PackageType.rpm) {
+                eachLine = 'wrapper.java.command=/etc/alternatives/jre_17/bin/java'
+              }
             }
 
             if (eachLine == 'wrapper.jarfile=../lib/wrapper.jar') {
@@ -277,24 +271,6 @@ def configureLinuxPackage(DefaultTask packageTask, InstallerType installerType, 
           include 'libwrapper-linux-arm-64*'
         }
       }
-
-      project.copy {
-        into "${buildRoot}/usr/share/${installerType.baseName}/jre"
-        from(tarTree(downloadLinuxJre.outputs.files.singleFile)) {
-          eachFile { FileCopyDetails fcd ->
-            List<String> segments = fcd.relativePath.segments
-            segments.remove(0)
-            fcd.relativePath = new RelativePath(!fcd.isDirectory(), (segments as String[]))
-            // Remove write access of JRE files for user group `others`.
-            // Perform `&` operation on existing mode with `0775`.
-            // eg. if existing mode is `0777`, the mode will be changed to `0775` after following operation
-            // eg. if existing mode is `0477`, the mode will be changed to `0475` after following operation
-            fcd.setMode(fcd.getMode() & 0775)
-          }
-          includeEmptyDirs = false
-        }
-      }
-
     }
 
     doLast {
--- a/installers/linux/deb/after-install.sh.erb
+++ b/installers/linux/deb/after-install.sh.erb
@@ -20,4 +20,8 @@
     # initialize init script symlinks on first install
     /usr/share/<%= name %>/bin/<%= name %> install 2>&1
 
+    arch=$(dpkg --print-architecture)
+    java=/usr/lib/jvm/java-17-openjdk-$arch/bin/java
+    sed -ie "s:^\\(wrapper.java.command=\\).*:\\1$java:" /usr/share/<%= name %>/wrapper-config/wrapper.conf
+
     <%= ERB.new(File.read(File.join(install_scripts_dir, 'shared', 'partials', "_#{name}_post_install_message.sh.erb")), trim_mode: '-', eoutvar: "_#{SecureRandom.hex}").result(binding) %>


chadlwilson:
Thanks for this, I'll have a play around.

While I am not an expert in the area I'm thinking that to avoid conflicts and maximise compatibility

1.  it might be good to add alternative optional dependencies that also satisfy it, such as a JDK or a non-headless variant - or maybe even the abstract capabilities for each platform?. It might be moot if the JDK and non-headless variants actually augment (and dont conflict with) the headless runtime variants.
2. we might consider using `--deb-recommends` package dependencies on Deb/Ubuntu. Don't think there is an equivalent on rpm though (and weak deps is not what we want), so maybe just causes more confusion than necessary?
3. On Ubuntu/Debian `openjdk-17-jre-headless` conflicts `openjdk-11-jdk-headless << 11~18-2` but I don't really understand why or the implications, as they install in different locations but might be misunderstanding the meaning (just keeping a note here so I can double check it). What I am thinking about is sanity checking that adding this dependency isn't going to cause issues for other things/versions people might need on their systems for other reasons (especially on agent systems where people typically install other runtimes and such for use in jobs).

matthiaskraaz:
DEB: I'd love to have a dependency on a virtual package instead of a concrete package, but I can't find a virtual package specifically for Java 17. The virtual package java17-runtime-headless is also provided by openjdk-18-jre-headless, openjdk-19-jre-headless, ... So "Depends: java17-runtime-headless" or "Depends: openjdk-17-jre-headless | java17-runtime-headless" would allow to try to use the go-* package with Java 18 or later, which would break our dependency on /usr/lib/jvm/java-17-openjdk-$arch/bin/java. And AFAICS it is not possible to specify a (major) version for a virtual package dependency.

RPM: Same, with the exception that RPM automatically selects the latest package that fulfills a dependency.

chadlwilson:
This is merged and automated tests are fine. https://build.gocd.org/go/pipelines/installer-tests/2456/install-tests/1 

I made some changes before this to improve #s/types of OSes this is tested against:
https://github.com/gocd/installer-testing/blob/47b859f67f2b508b74ba527060ebc5995364a4d2/Rakefile#L169-L175

It ensures server and agent both start up OK and agent can connect to server, with both a fresh install and a package upgrade from `20.5.0` to latest. I added validation against Ubuntu `24.04` and Almalinux 8/9 as a RHEL equivalent. Alma validates using systemd as init system and service runner since `/etc/init.d` isn't available on RHEL 9-compatibles.

I'll do a bit more sanity testing for alternate Java dependencies - ideally I'd like the package to not complain if the user has a JDK installed that may otherwise clash/`conflicts` with the main JRE dependency we add. Adding a dependency on the concrete package as primary seems the canonical way with an alternative virtual or concrete package in an `OR`; as long as we can locate its installation folder reliably (and noting that we don't want to "allow" Java 18+ so some virtual packages are not suitable).

matthiaskraaz:
I pondered this a little more:

- it would be a good thing to ensure someday that GoCD runs with Java 21. But given that the Java 17 EOL is 2029 or so, this really can't be a top priority.
- it's somewhat funny, that you can either depend on a specific implementation of a specific Java version or on some implementation of some Java version. I tend to ask the Debian and CentOS Java Package maintainers whether I have overlooked something.

chadlwilson:
Yes, I intend to get to Java 21 some time soon. Just want to limit versions by changing baseline to Java 17 first (which is a requirement of `24.1.0`) so I don't have to keep thinking about Java 11 as well as 21.

chadlwilson:
I did some more real-world sanity testing of this with both debs and rpms to see what happens if there is already a "full" JDK installed before installing the gocd packages.

In both cases, with the way packages are modelled in the respective ecosystems the JDKs either layer on top of the headless JREs and regular JREs OR "replace" the JRE package - so having those pre-installed will mean the package requirement is already met when installing gocd packages.

So there's nothing more needed to be done here on adding an "`OR`" dependency. Apologies if you already knew that, I hadn't followed!

Accordingly, released 24.1.0: https://www.gocd.org/releases/#24-1-0

Thanks for the help, this is a much better situation!

matthiaskraaz:
There is never anything wrong with more tests! I have written to openjdk-17@packages.debian.org to ask if there is some better approach than the one taken, but have not received any answer. So what are the most pressing issues now?

chadlwilson:
"Pressing" is probably a matter of perspective, but other than the likely-nasty long-standing EOL dependency problems (e.g [Hibernate](https://github.com/gocd/gocd/issues/10262)) - it'd be nice to to some sanity testing with Java 21 and start flushing out the various issues.

The "easy" stuff is at https://github.com/gocd/gocd/pull/12762 but failures in the pipelines there will probably be hard to debug as they will be full functional regression test failures. If you're interested in exploring, it's probably better to first try and run all the normal tests with Java 21 by having a Java 21 JRE available and then changing this to `21`:

https://github.com/gocd/gocd/blob/756866c6974df118928d1951265104a537238c51/build.gradle#L139

matthiaskraaz:
Sounds like fun. Are all tests in the main repository?

chadlwilson:
> Sounds like fun. Are all tests in the main repository?

The main JUnit driven unit, API and component tests that five the fastest feedback are, yes. They currently only run on a single Java release. They take a long time to run all of them locally so usually better to target just some of them (`./gradlew server:fastUnitTest` is a good one to aim for) . They still rely on a few dependencies documented at https://developer.gocd.org/current/ but without specifics for installing on Linux if that's your local Dev env. 😬

I imagine it'll take quite a few iterations to make various Java 21 forward compatible fixes for various pieces, and we might get stuck on some legacy dependencies which don't function properly on Java 22 in unfixable ways. But don't know any of these right now, so have to flush 'em out 😞 

There are also fuller functional regression tests (that run against packaged server/agent) at https://github.com/gocd/ruby-functional-tests but that is a bit suite and generally better to let the pipeline run those.

And then there are the rpm/deb tests at https://github.com/gocd/installer-testing (the ones I augmented for this issue) but those are at the very end of the pipeline running against signed packages, and really just basic smoke test of the Linux packages.

chadlwilson:
I really need to get around to refactoring https://build.gocd.org pull request pipelines to make them visible to guests which would make it a lot easier to collaborate on experimental stuff like this 😓

chadlwilson:
As a starting point, all the smoke tests and functional regression tests pass on Java 21 right now, so if there are issues they are likely only found at a lower layer with more specific unit tests (or relate to deprecated behaviour) But it's a good starting point to be at.

chadlwilson:
Sorry to give you the runaround, but it seems that Java 21 is actually pretty straight-forward in most areas - at least enough to merge to master and trial it out on https://build.gocd.org soon. I ran all of the Gradle tests and they were all fine.

There is a problem that seems specific to running local dev servers and the compile/build infrastructure (JRuby/Rails nonsense) which I'll track at https://github.com/gocd/gocd/pull/12762#issuecomment-2109259718 but might not be much we can about around that for now. If you're still planning to have a look yourself or look further forward to address deprecations, we can move the conversation to that PR :-)

chadlwilson:
Sigh, no Java 21 variant packages available on Debian 11 or 12 yet. 🤦‍♂️

matthiaskraaz:
Surprises me too that Debian hasn't a Java 21 package yet.

```
BuildWorkTest.shouldCleanAgentWorkingDirectoryIfExistsWhenCleanWorkingDirIsTrue()
BuildWorkTest.shouldCleanAgentWorkingDirectoryWithSymlinksIfExistsWhenCleanWorkingDirIsTrue()
BuildWorkTest.shouldCreateAgentWorkingDirectoryIfNotExist()
BuildWorkTest.shouldCreateAgentWorkingDirectoryIfNotExistWhenFetchMaterialsIsFalse()
BuildWorkTest.shouldNotBombWhenCreatingWorkingDirectoryIfCleanWorkingDirectoryFlagIsTrue()
BuildWorkTest.shouldReportBuildIsFailedWhenAntBuildPassed()
BuildWorkTest.shouldReportConsoleOut()
BuildWorkTest.shouldUpdateBothStatusAndResultWhenBuildHasPassed()
AntTaskBuilderTest.shouldFailWhenTargetDoesNotExist()
```

fail. I'm digging into it...

chadlwilson:
> ```
> BuildWorkTest.shouldCleanAgentWorkingDirectoryIfExistsWhenCleanWorkingDirIsTrue()
> BuildWorkTest.shouldCleanAgentWorkingDirectoryWithSymlinksIfExistsWhenCleanWorkingDirIsTrue()
> BuildWorkTest.shouldCreateAgentWorkingDirectoryIfNotExist()
> BuildWorkTest.shouldCreateAgentWorkingDirectoryIfNotExistWhenFetchMaterialsIsFalse()
> BuildWorkTest.shouldNotBombWhenCreatingWorkingDirectoryIfCleanWorkingDirectoryFlagIsTrue()
> BuildWorkTest.shouldReportBuildIsFailedWhenAntBuildPassed()
> BuildWorkTest.shouldReportConsoleOut()
> BuildWorkTest.shouldUpdateBothStatusAndResultWhenBuildHasPassed()
> AntTaskBuilderTest.shouldFailWhenTargetDoesNotExist()
> ```
> 
> fail. I'm digging into it...

Assume these are the tests that are failing for you? Might be able to make a guess if I can see a couple of the stack traces and error messages. Might be something weird with your git working dir and/or some cruft left behind (at a random guess).

Could try a ./gradlew clean first to start freshish if it's that.

matthiaskraaz:
So they are not supposed to fail?! May be my setup: Ubuntu 22.04.4 LTS via WSL on a Windows machine. Let me try on a real Linux machine instead.

chadlwilson:
> So they are not supposed to fail?! May be my setup: Ubuntu 22.04.4 LTS via WSL on a Windows machine. Let me try on a real Linux machine instead. PS: I did a git clean -fdx beforehand, so it shouldn't be any cruft.

All tests should and do pass in CI and locally - but will never have been tested via Linux on WSL2. Windows on its own, yes. Linux yes. The vagaries of filesystems and locks via WSL2 I've no idea, because they are a pain on Windows normally, and on Linux the underlying FS wouldn't be expected to affect things.

Depends on what the errors are though..

chadlwilson:
Some tests are targetting towards Windows or non-Windows where testing filesystem specific behaviour, for example.

I did try with WSL2 locally once I believe, but I only have windows via virtualisation and at the moment aarch64-only (so Win 11 insider preview) so I'm not sure I want to go down that rabbit hole 😅

matthiaskraaz:
Going through the output it became pretty fast pretty obvious that `sudo apt-get install subversion rake ant -y`  was all that's needed. Now `server:fastUnitTest` succeeds locally.

PR #12762, opened 2024-05-13
Build, test and package GoCD with Java 21 LTS by default

chadlwilson:
There seem to be no issues with using Java 21 by default at runtime, so let's get this going for build.gocd.org so we can smoke test things.


chadlwilson:
Unit and regression tests are all fine when running under Java 21, however running sprockets asset precompile via sass-embedded and dart-sass is currently broken on Java 21 (which prevents running the rails/ruby unit tests under Java 21).

Not sure of root cause, but trying at https://github.com/sass-contrib/sass-embedded-host-ruby/issues/208 for now.

PR #12827, opened 2024-06-03
fix: artifacts cronjob performance and use busybox tar

joshuarli:
when i wrote the zip stuff i didn't verify it locally as i was trying to save time by just doing it on staging, easy to forget alpine doesn't have zip lol

also a huge performance gain, see comments


joshuarli:
ugh sorry

chadlwilson:
No worries.

Are there any of your forked changes which you think could be beneficial to the wider community or altered upstream to make your lives easier?

At a high level, I am quite curious about incorporating some configurable artifacts clean-up in the container images; although haven't given it much detailed thought (running cronjobs in a container I suppose is "non canonical" for packaged software, but I understand the need).

joshuarli:
> No worries.
> 
> Are there any of your forked changes which you think could be beneficial to the wider community or altered upstream to make your lives easier?
> 
> At a high level, I am quite curious about incorporating some configurable artifacts clean-up in the container images; although haven't given it much detailed thought (running cronjobs in a container I suppose is "non canonical" for packaged software, but I understand the need).

We've only needed a few minor tweaks and we're pretty behind now, at some point i'd like to upgrade our branch and see if there's anything that can be upstreamed - but nothing major.

Running cronjobs in a container is unnatural, but it's the only choice we have as our gocd-server is a k8s pod and we're kind of all in on a full k8s deployment for gocd. And it was a quick solution for us. I wasn't able to find anything on deleting (or archiving) artifacts based on a retention policy.

chadlwilson:
No, there isn’t anything and IMHO it’s a big problem. Personally I’d like to have started by building a dedicated API for this that can be “aware” of the use of various artifacts, last n runs (regardless of age), ability to filter for certain pipeline groups etc, is aware of artifacts needed for downstream fetches.

But it’d all be quite a lot of work, so perhaps a pipe dream :)

Issue #10852, opened 2022-09-22
Upgrade to Webpack 5


chadlwilson:
##### Summary

Currently we are using Webpack 4. Many plugins/loaders etc have changed and are no longer getting patched/updated, so we probably need to migrate to Webpack 5.

Keeping track of work/research here, for want of a better place.

**Things to do**

_General research_
- [x] https://webpack.js.org/migrate/5/
- [x] Research whether the below loaders/plugin combinations are still the best/recommended approaches here

_Plugins/Loaders_
- [x] Upgrade `css-loader` from 5.x to 6+
  - https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#600-2021-07-14
  - Unclear if 5.x works on webpack 5?  (probably)
- [x] Upgrade `style-loader` from 2.x to 3+
  - https://github.com/webpack-contrib/style-loader/blob/master/CHANGELOG.md#300-2021-06-24
  - Unclear if 5.x works on webpack 5?  (probably)
- Consider switching `file-loader` & `url-loader` to https://webpack.js.org/guides/asset-modules/  
- [x] Upgrade webpack-build-notifier from 2.x to 3+

_Current version already supports webpack 5 / can be deferred_
- [x] Upgrade `ts-loader` from 8.x to 9+
  - https://github.com/TypeStrong/ts-loader/blob/main/CHANGELOG.md#v900
- [x] Upgrade `sass-loader` from 10.x to 13+
  - https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#1100-2021-02-05
- [x] Upgrade `stylelint-webpack-plugin` from 2.x to 3+
  - https://github.com/webpack-contrib/stylelint-webpack-plugin/blob/master/CHANGELOG.md#300-2021-07-19 
  - [ ] Also upgrade stylelint to 16+ which is blocked on this
  - [ ] Also upgrade stylelint-config-standard-scss to 12+ which is blocked on this
- [x] Upgrade `fork-ts-checker-webpack-plugin` from 6.x to 7+
  - https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/releases/tag/v7.0.0 
- [x] Upgrade `html-webpack-plugin` from 4.x to 5.x
  - https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md#500-2021-02-03
- [x] Upgrade `mini-css-extract-plugin` (and types) from 1.x to 2+
[...]

PR #12992, opened 2024-08-03
Downgrades stylelint to last version compatible with stylelint-webpack-plugin

chadlwilson:
While that later version works fine when there are no linting errors, it falls over as soon as there is a single linting error, which is why I didn't notice the problem before. This downgrades to the last compatible versions.


server/src/main/webapp/WEB-INF/rails/package.json
```diff
@@ -111,8 +111,8 @@
     "speed-measure-webpack-plugin": "^1.5.0",
     "stats-webpack-plugin": "^0.7.0",
     "style-loader": "^2.0.0",
-    "stylelint": "^16.8.1",
```
Mouvedia, line 114: Or you could update `stylelint-webpack-plugin` instead.

ref https://github.com/webpack-contrib/stylelint-webpack-plugin/releases/tag/v5.0.0

chadlwilson: Thanks - but unfortunately cannot, as need to do a webpack 5 upgrade first (#10852)

PR #13321, opened 2024-12-10
Sleep at least 1 second before checking dockerd status to workaround cgroupsv2 challenges

chadlwilson:
The `dind` hack script on cgroupsv2 needs to move processes to different groups which does not work reliably when there are other scripts (such as this one) creating new processes in the entrypoint creating forked processes at the same time due to race conditions.

This adds an initial wait to give the `dind` hack script time to do its thing before passing off to dockerd.
- See https://github.com/moby/moby/blob/master/hack/dind
- See https://github.com/docker-library/docker/blob/master/dockerd-entrypoint.sh for what this script replaces.


chadlwilson:
There is an attempt to fix this upstream at https://github.com/moby/moby/pull/48850

PR #13420, opened 2025-01-16
Revert "ui-dev: bump concurrent-ruby from 1.3.4 to 1.3.5 in /server/src/main/webapp/WEB-INF/rails"

chadlwilson:
Reverts gocd/gocd#13415


chadlwilson:
Sadly breaks rails and where are on an EOL version which won't be fixed :( https://github.com/ruby-concurrency/concurrent-ruby/issues/1077

chadlwilson:
Can possibly workaround with a `require 'logger'` somewhere during init.

PR #13425, opened 2025-01-17
Fix unparseable commit date (Y2038/Y2106 problem) causing materials pull to fail

romansorin:


chadlwilson:
Would you like to consider describing and getting a fix upstreamed for this problem rather than maintaining a fork? :-)

romansorin:
> Would you like to consider describing and getting a fix upstreamed for this problem rather than maintaining a fork? :-)

Absolutely - happy to open a PR once we validate it on our end!

PR #13779, opened 2025-07-30
default to current UTC time rather than bombing on 2038/2106 commits

romansorin:
commits made when experiencing clock drift (e.g., 2038, 2106, 2120) result in the git integration failing due to unparseable commit timestamps.

i've deployed these changes to our cluster earlier this year and have not experienced any build or deploy issues since


romansorin:
latest upstream sync shows that these commits are no longer valid; will re-open pr if this is still an outstanding issue on master

chadlwilson:
I imagine they are possibly still valid? The conflicts will just be that I've moved away from jodatime, but I didn't 100% "fix" all of the parsing logic around the place.

PR #14307, opened 2026-04-14
feat: show commit author and hash on pipeline activity instance cards

Danny-r-22:
The pipeline activity page shows a top-level pipeline revision but doesn't surface the developer name or per-material commit hash from `buildCause.materialRevisions`. This change adds both to each instance card using data already returned by the existing API, making it easier to identify who triggered a pipeline run and from which commit at a glance.

Note: branch name is not yet available in the pipeline activity JSON response (omitted from `ScmMaterial.toJson()`) — a follow-up backend change could expose it. Fyi @chadlwilson

### DIFF
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/pipeline_run_info_widget.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/pipeline_run_info_widget.tsx
@@ -87,6 +87,15 @@ export class PipelineRunWidget extends MithrilViewComponent<PipelineRunAttrs> {
           </span>
         </div>
         <div class={styles.revision}>Revision: {pipelineRunInfo.revision()}</div>
+        {pipelineRunInfo.materialRevisions().length > 0 && (() => {
+          const firstRev = pipelineRunInfo.materialRevisions()[0];
+          const shortHash = firstRev.revision().substring(0, 8);
+          return <div class={styles.revision} data-test-id={"commit-info"}>
+            <span data-test-id={"commit-author"} title={firstRev.user()}>{firstRev.user()}</span>
+            {" — "}
+            <span data-test-id={"commit-hash"} title={firstRev.revision()}>{shortHash}</span>
+          </div>;
+        })()}
         <div class={styles.scheduleInfo}
              data-test-id={"time"}
              title={PipelineRunWidget.getTimeServer(pipelineRunInfo.scheduledTimestamp())}>
--- a/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/pipeline_run_info_widget.tsx
+++ b/server/src/main/webapp/WEB-INF/rails/webpack/views/pages/pipeline_activity/pipeline_run_info_widget.tsx
@@ -74,9 +74,12 @@ export class PipelineRunWidget extends MithrilViewComponent<PipelineRunAttrs> {
 
   view(vnode: m.Vnode<PipelineRunAttrs>): m.Children | void | null {
     const pipelineRunInfo = vnode.attrs.pipelineRunInfo;
+    const revisions = pipelineRunInfo.materialRevisions();
+    const firstRev = revisions.length > 0 ? revisions[0] : undefined;
+    const shortHash = firstRev ? firstRev.revision().substring(0, 8) : undefined;
 
     return <tr class={styles.groupContent}
-               data-test-id={this.dataTestId("pipeline-instance", pipelineRunInfo.label())}>
+           data-test-id={this.dataTestId("pipeline-instance", pipelineRunInfo.label())}>
       <td class={styles.left} data-test-id={"meta"}>
         <div class={classnames(styles.run, styles.header)}>
           <span class={styles.label} data-test-id={"counter"}>
@@ -87,15 +90,13 @@ export class PipelineRunWidget extends MithrilViewComponent<PipelineRunAttrs> {
           </span>
         </div>
         <div class={styles.revision}>Revision: {pipelineRunInfo.revision()}</div>
-        {pipelineRunInfo.materialRevisions().length > 0 && (() => {
-          const firstRev = pipelineRunInfo.materialRevisions()[0];
-          const shortHash = firstRev.revision().substring(0, 8);
-          return <div class={styles.revision} data-test-id={"commit-info"}>
+        {firstRev && (
+          <div class={styles.revision} data-test-id={"commit-info"}>
             <span data-test-id={"commit-author"} title={firstRev.user()}>{firstRev.user()}</span>
             {" — "}
             <span data-test-id={"commit-hash"} title={firstRev.revision()}>{shortHash}</span>
-          </div>;
-        })()}
+          </div>
+        )}
         <div class={styles.scheduleInfo}
              data-test-id={"time"}
              title={PipelineRunWidget.getTimeServer(pipelineRunInfo.scheduledTimestamp())}>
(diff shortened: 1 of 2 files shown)


chadlwilson:
Thanks - this page indeed has need for more context to be useful.

Would be good if you could augment the unit tests and include a screenshot so I know what I’m expecting.

Danny-r-22:
Hey @chadlwilson please find additions to the tests in the latest [commit](https://github.com/gocd/gocd/pull/14307/changes/042a9cf7c1e7b5fc28bf537b3e3091f60eaa6315) 

In terms of a screenshot, I wasn't able to get the test mock server up and running but I got claude to do a html mock up. Please let me know what you think

image

chadlwilson:
Why haven't you been able to run it locally? Kinds means it might not actually work. I might be able to help with that, but otherwise will take a look at some point.

Not all material revisions represent commits as not all materials are git materials, nor that picking the first one makes asnse (for pipelines with multiple materials) so I'm not sure the change as it stands makes sense right now, but I don't know that page well off the top of my head so will need some effort to look in more detail at some point.

Danny-r-22:
Just a bunch of yarn, gradle and webpack version issues I think, I don't think it's related to the PR.

I don't really want to mess up my environment, if you have it available are you able to check out the branch and test easily?

If not I'll give it another go.

In terms of the pain point for this, when you have a team of 30+ dev's and a lot of commit's it makes difficult to find the correct pr and commit, it's not a massive deal but it would streamline the process quite a lot in my opinion

chadlwilson:
I agree the history needs work -.just needs effort, and I am just one person doing things for free.with many people to complain to me if anything goes wrong - far from a 30 person dev team.

So it might take some time if there is some extra thought needed into the design and edge cases, 

As suggested at https://developer.gocd.org/26.1.0/ I recommend using mise to keep relarively isolated envs. The config is all there in GoCD to use the correct versions of required tools without affecting other things on your machine and to make it relatively easy for a multi-lang project.

Danny-r-22:
Ok cool will check it out tomorrow thanks @chadlwilson, in the meantime, it doesn't look high risk right?

chadlwilson:
Not high risk, no - just need to see what it looks like and make sure it is semantically correct with respect to different material types.

The automated unit & functional tests are passing at least, so that's a start (the feedback back to GitHub for PR automated builds seems to be broken, so need to fix that).

Danny-r-22:
Hey @chadlwilson sorry it's been a little while, I've updated based on the copilot suggestion hope it looks ok.

Is it possible for you to run locally please? My env on this work laptop and restrictions just aren't having the setup.

chadlwilson:
Yeah, I will at some point soonish, before releasing 26.1.0.

Danny-r-22:
Sounds great, thanks @chadlwilson

chadlwilson:
After finally getting some time to take a look, unfortunately this change isn't logically or stylistically correct right now.

image

1. the "chosen" revision is actually already displayed in `Revision: abcdef`, so appending the `shortHash` is unnecessary and duplicative
  - the added `shorthash` is generally not visible anyway, as it's cropped behind the user
  - Revisions are not necessarily git-style sha hashes - materials can be of many types in GoCD. This impl is git-specific
2. The biggest problem is that the user/author information isn't consistent with the "chosen" revision displayed. The material revision displayed is that of the most recently modified material revision for that given instance of the pipeline, e.g if the pipeline has 3 materials, and only 1 is changed since the last pipeline instance run, it will display the revision of the most recent revision that is actually changed between the last pipeline run; not the revision of the first material in the list. This may not necessarily be in chronological order due to fan-in semantics.
- So if we displayed the user from the first material (regardless of its status) this would be misleading/wrong as it might be from a different material.

https://github.com/gocd/gocd/blob/943a2ff01eb61a3db4d6dc6cbdaae82f814d6266/domain/src/main/java/com/thoughtworks/go/domain/MaterialRevisions.java#L101-L112

Right now, this `latestRevision` (`pipelineRunInfo.revision`) is calculated server-side. If we are going to "flatten" information and display from a chosen revision, the calculation needs to be in the same place. The API used (`/pipelineHistory.json`) is rather ancient, and we probably _could_ move the selection to be done on the front-end.

FWIW, right now this information is available via the "Triggered by ..." popover; but not available "at a glance", to be fair.

Note how in the below sometimes the revision reflects an upstream pipeline "Dependency Material" and sometimes a source control material directly on the pipeline.

image

It's also fair to say that the "Revision: " displayed is possibly not entirely clear what it represents, since it is only picking a single revision from the set of relevant revisions.

chadlwilson (changes requested):
As above.

chadlwilson:
Closing this now, as there’s not been a response. Happy to discuss what you’re looking for (feel free to open an issue that describes what you’d like to be able to do at a glance) and to consider an alternate approach.

Danny-r-22:
> Closing this now, as there’s not been a response. Happy to discuss what you’re looking for (feel free to open an issue that describes what you’d like to be able to do at a glance) and to consider an alternate approach.

Hey @chadlwilson 

At the moment we see just see triggered by changes

image

So to see the dev's name we have to search in the terminal output which is all a bit of a faff and when you have a large team and lots of PR's it can be a bit of a nightmare.

Please let me know if there's anything we can do setup wise to change this or if there is a code change possible on your side to populate the info that would be great.

Kind regards
Danny

chadlwilson:
As I mentioned above, you can click the "triggered by changes" pop-over (rather than terminal output) to see the immediate triggering materials, or use the "compare" view to see all changes (incl. Individual commits) across materials between two pipeline runs.

The problem is that not all pipeline runs are directly triggered by code changes and that there may be multiple code changes (commits) bundled together that triggered a run even for a single material. Sometimes the run is triggered by another pipeline, or by a change to an artifact, package or any other material.

Maybe I can add a link to that view to make comparing between two the latest run of two sequential pipelines a single click.

Since you mention PRs, if you are using a single logical pipeline to handle lots of unrelated pull requests (e.g via an unofficial PR polling plugin like https://github.com/ashwanthkumar/gocd-build-github-pull-requests ) then that is a bit of a different problem since that plugin "fights" against how GoCD wants to be used (which is not really optimized for PR flows)

If you can clarify your use case for this view I can think about whether it'd be possible to change "Triggered by changes" to include some sort a suffix which summarises the materials that changed, and perhaps includes the most recent commit author if the changed materials include a source code material with new commits)

Danny-r-22:
> As I mentioned above, you can click the "triggered by changes" pop-over (rather than terminal output) to see the immediate triggering materials, or use the "compare" view to see all changes (incl. Individual commits) across materials between two pipeline runs.
> 
> The problem is that not all pipeline runs are directly triggered by code changes and that there may be multiple code changes (commits) bundled together that triggered a run even for a single material. Sometimes the run is triggered by another pipeline, or by a change to an artifact, package or any other material.
> 
> Maybe I can add a link to that view to make comparing between two the latest run of two sequential pipelines a single click.
> 
> Since you mention PRs, if you are using a single logical pipeline to handle lots of unrelated pull requests (e.g via an unofficial PR polling plugin like https://github.com/ashwanthkumar/gocd-build-github-pull-requests ) then that is a bit of a different problem since that plugin "fights" against how GoCD wants to be used (which is not really optimized for PR flows)
> 
> If you can clarify your use case for this view I can think about whether it'd be possible to change "Triggered by changes" to include some sort a suffix which summarises the materials that changed, and perhaps includes the most recent commit author if the changed materials include a source code material with new commits)

Ohhhhh I see, that's great thanks Chad!

chadlwilson:
Are you able to clarify whether you are using a plugin like https://github.com/ashwanthkumar/gocd-build-github-pull-requests or reusing a single plugin to switch across PR branches for the same logical pipeline?

Will help me figure out what might improve things a bit.

PR #14308, opened 2026-04-14
Improve rendering of long pipeline names in pipeline activity/history view

aembleton:
Long pipeline names are getting truncated by this max-width


chadlwilson:
Thx for the PR.

I'll take a look and see if this causes any (terrible) issues with the rendering. Some limit may be necessary to get things to flow vaguely correctly, but I am aware the UI has some problems with extremes (long pipeline group names cause mess on dashboard as well).

chadlwilson:
Unfortunately this breaks the rendering and causes everything to run off screen with a long pipeline name, removing the filter box, the unpause button, the pause message, the configure button etc.

The flex box css here needs some work to make these flex correctly, to make more space for the pipeline name.

chadlwilson (changes requested):
A

PR #14503, opened 2026-07-19
style: remove redundant public modifiers in JobInstancesTest

santoshbachar:
Remove unnecessary `public` modifiers from JUnit 5 tests (Sonar S5786).


chadlwilson:
I should probably mass fix these across the codebase.

For future, you probably want to send your pull requests from a branch that you intend to throw away after merge (rather than `master`) and keep your master able to be cleanly synced with upstream.

Otherwise your master branch will diverge from upstream master in ways which will be messy to sync with upstream. Some you won't be able to fix locally (to get a clean history) except doing by doing a "hard" reset to upstream master and then force-pushing to your fork.

For example, this one I merged with a squash commit to clean up the history here, which will probably make a mess locally when/if you try and pull from upstream (gocd/gocd) master using a rebase strategy.

PR #14508, opened 2026-07-21
fix: prevent XXE in XML parsing (Sonar S2755)

santoshbachar:
XpathUtils only had FEATURE_SECURE_PROCESSING set, which isn't enough on its own - explicitly disallow DOCTYPE now. Same gap in GoPluginDescriptorParser's SchemaFactory, so blocked external DTD/schema access there too.

Added tests for both that check a DOCTYPE-declared external entity gets rejected instead of resolved.

PR #14509, opened 2026-07-21
test: add regression tests for XXE handling in XML parsing

santoshbachar:
This is a clean resubmission of https://github.com/gocd/gocd/pull/14508.

As discussed there, this PR only keeps the two regression tests. Since `FEATURE_SECURE_PROCESSING` is sufficient on Java 21+, no source changes are required.

### DIFF
--- a/config/config-api/src/test/java/com/thoughtworks/go/util/XpathUtilsTest.java
+++ b/config/config-api/src/test/java/com/thoughtworks/go/util/XpathUtilsTest.java
@@ -96,6 +96,24 @@ public class XpathUtilsTest {
     }
 
     @Test
+    void shouldRejectXmlContainingDocTypeDeclarationToPreventXXE() throws Exception {
+        String maliciousXml = """
+            <?xml version="1.0"?>
+            <!DOCTYPE root [
+              <!ENTITY xxe SYSTEM "file:///etc/passwd">
+            ]>
+            <root>
+            <son>
+            <grandson name="&xxe;"/>
+            </son>
+            </root>""";
+
+        File file = getTestFile(maliciousXml);
+
+        assertThrows(XPathExpressionException.class, () -> XpathUtils.evaluate(file, "/root/son/grandson/@name"));
+    }
+
+    @Test
     public void shouldReturnEmptyStringWhenMatchedNodeIsNotTextNode() throws Exception {
         String xpath = "/root/son";
         String value = XpathUtils.evaluate(getTestFile(), xpath);
--- a/plugin-infra/go-plugin-infra/src/test/java/com/thoughtworks/go/plugin/infra/plugininfo/GoPluginDescriptorParserTest.java
+++ b/plugin-infra/go-plugin-infra/src/test/java/com/thoughtworks/go/plugin/infra/plugininfo/GoPluginDescriptorParserTest.java
@@ -16,6 +16,7 @@
 package com.thoughtworks.go.plugin.infra.plugininfo;
 
 import jakarta.xml.bind.JAXBException;
+import jakarta.xml.bind.ValidationException;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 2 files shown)


chadlwilson:
Ahh, did you sanity check my conclusions from there? I don't use Sonar, so not sure exactly what they were getting at or whether there is some edge case FSP misses.

config/config-api/src/test/java/com/thoughtworks/go/util/XpathUtilsTest.java
```diff
@@ -95,6 +95,24 @@ public void shouldThrowExceptionForBadXML() {
...
+            ]>
+            <root>
+            <son>
+            <grandson name="&xxe;"/>
+            </son>
+            </root>""";
+
+        File file = getTestFile(maliciousXml);
+
+        assertThrows(XPathExpressionException.class, () -> XpathUtils.evaluate(file, "/root/son/grandson/@name"));
```
chadlwilson, line 113:
```suggestion
    void shouldRejectXmlContainingExternalDtdReferenceToPreventXXE() throws Exception {
        String maliciousXml = """
            <?xml version="1.0"?>
            <!DOCTYPE root [
              <!ENTITY xxe SYSTEM "file:///etc/passwd">
            ]>
            <root>
              <son>
                <grandson>&xxe;</grandson
              </son>
            </root>""";

        File file = getTestFile(maliciousXml);

        assertThatThrownBy(() -> XpathUtils.evaluate(file, "/root/son/grandson"))
            .isExactlyInstanceOf(XPathExpressionException.class)
            .hasCauseExactlyInstanceOf(SAXParseException.class)
            .hasMessageContaining("not allowed due to restriction set by the accessExternalDTD property");
```


This test is a bit too loose and is passing for the _wrong reason_ right now.

The name implies it expects to reject the doctype, but the exception isn't due to that, and it doesn't check the exception message or cause. `XPathExpressionException` is a wrapping exception and could be thrown for any type of error, so the assertion should really be stronger, to ensure its actually testing what it says it's testing. 

Right now with or without secure processing you get

```
"org.xml.sax.SAXParseException; lineNumber: 7; columnNumber: 22; The external entity reference "&xxe;" is not permitted in an attribute value."
```

If the correct defenses are in place with the right type of malicious XML, you should get

```
"org.xml.sax.SAXParseException; lineNumber: 7; columnNumber: 16; External Entity: Failed to read external document 'passwd', because 'file' access is not allowed due to restriction set by the accessExternalDTD property."
```

This is because the XML is failing earlier validation - to get past the basic XML validation and validate that "external DTDs" are now allowed the test would need to be like the below (not putting the XXE ref in an attribute)

```xml
<?xml version="1.0"?>
<!DOCTYPE root [
  <!ENTITY xxe SYSTEM "file:///etc/passwd">
]>
<root>
  <son>
    <grandson>&xxe;</grandson>
  </son>
</root>
```

My suggestion above would need some other changes for imports etc; just to give an idea.

plugin-infra/go-plugin-infra/src/test/java/com/thoughtworks/go/plugin/infra/plugininfo/GoPluginDescriptorParserTest.java
```diff
@@ -73,6 +74,34 @@ void shouldParseValidVersionOfPluginXML() throws Exception {
...
+                  <url>https://github.com/gocd/fake-gocd-analytics-plugin</url>
+                </vendor>
+                <target-os>
+                  <value>Linux</value>
+                </target-os>
+              </about>
+            </go-plugin>""";
+
+            try (InputStream pluginXml = new ByteArrayInputStream(maliciousXml.getBytes(UTF_8))) {
+                assertThrows(ValidationException.class, () -> parseXml(pluginXml, "/tmp/", new File("/tmp/"), true));
```
chadlwilson, line 101: As above, would be good to strengthen the assertions here to check the test is working for the "correct" reason rather than some other syntax error.

chadlwilson:
Hiya @santoshbachar - I see you added a commit to a different branch in your fork (but not here). Are you still working on this PR before updating here?

santoshbachar:
> Hiya @santoshbachar - I see you added a commit to a different branch in your fork (but not here). Are you still working on this PR before updating here?

Hi Chad, I need to figure out my alerts, and yes I am preparing a review document for you.

PR #14540, opened 2026-08-03
feat: add sparse checkout support for git materials

palencharj:
Adds sparse checkout to git materials, so a pipeline can check out only the paths it needs instead of the whole repository. This is the git equivalent of a Perforce client view, which `p4` materials have always had via `<view>`, and it is the missing piece for anyone using GoCD against a large monorepo.

## Configuration

Modelled directly on the existing `<view>` element of `p4` materials — a nested element, one pattern per line:

```xml
<git url="https://example.com/monorepo.git" branch="main">
  <sparseCheckout>
    services/billing
    libs/shared/*
    build.gradle
  </sparseCheckout>
</git>
```

Patterns use the syntax accepted by `git sparse-checkout set --no-cone`, i.e. gitignore-style. Non-cone mode is deliberate: cone mode can only select whole directories, whereas non-cone can select individual files, which is what makes this a real analogue of a p4 view. Omitting the element leaves behaviour completely unchanged.

Also exposed as `sparse_checkout` on the config/materials/compare APIs, and as a field on the material form in the UI.

## Design notes

A few decisions worth surfacing for review:

- **Not part of the material fingerprint.** `appendCriteria` is untouched, matching how `shallowClone` is treated. Sparse checkout is a checkout strategy, not material identity — including it would orphan a pipeline's history whenever someone edited their patterns.
- **Does not affect change detection.** Polling still uses `git log` on the remote ref, so a commit touching only excluded paths still triggers the pipeline. Deciding *which* changes matter remains the job of the material `<filter>`/`invertFilter`. These two features compose rather than overlap.
- **Applied on every update, not just at clone.** This makes it self-healing: narrowing, widening or removing patterns repairs an existing working copy in place. `isRepositoryChanged` is untouched, so no re-clone is ever forced.
- **No git version bump for existing users.** `applySparseCheckout` is a no-op on repositories that neither are nor should be sparse, so a material that does not use the feature never invokes `git sparse-checkout` (which needs git 2.25+). Only opting in requires the newer git.
- **No config schema migration.** The element is additive and optional, so `cruise-config.xsd` needed no version bump, no frozen schema copy and no XSL. Existing configs validate unchanged — verified explicitly, including that a `<filter>`-only material still validates. It is placed before `<filter>` in the sequence because the config writer emits child-class fields before superclass ones (`sparseCheckout` is declared on `GitMaterialConfig`, `filter` on `ScmMaterialConfig`) — the same ordering `p4Type` already uses for `view` then `filter`.
- **No JSON fixture churn.** The representers use `addIfNotBlank`, so materials without sparse checkout serialise byte-identically and no existing expected-JSON fixture needed touching.

`GitSparseCheckoutConfig` implements `Validatable` deliberately: without it, an unresolvable `#{param}` inside the element would surface as a raw `IllegalStateException` from `ParamSubstitutionHandler` rather than a config error.

## Testing

New `GitMaterialSparseCheckoutTest` (9 tests) exercises the real git binary against the existing fixture repo, covering: full checkout by default, restriction to matching paths, one-pattern-per-line parsing, narrowing an existing working copy, restoring a full checkout after the config is removed, and config round-tripping. Two cases added to `MagicalGoConfigXmlLoaderTest` for the XML binding.

All green locally:

| Suite | Tests |
|---|---|
| `domain` git suites (22 classes, incl. submodules / refspec / shallow clone) | 200 |
| `MagicalGoConfigXmlLoaderTest` | 204 |
| `api-shared-v11` / `api-materials-v2` / `api-compare-v2` | 158 / 63 / 59 |
| `config-api` `GitMaterialConfigTest` | 27 |

Frontend `tsc --noEmit` reports no errors in the changed files.

One thing the tests caught that is worth knowing: `git sparse-checkout disable` leaves `.git/info/sparse-checkout` on disk and only flips `core.sparseCheckout`, so detecting "is this working copy sparse" has to read the config flag rather than test for the file.

## Notes

Happy to adjust the config shape if you would rather this were an attribute than an element, or to split the UI out into a follow-up. Documentation for `gocd/docs` can follow once the config surface is settled.

### DIFF
--- a/api/api-compare-v2/src/main/java/com/thoughtworks/go/apiv2/compare/representers/material/GitMaterialRepresenter.java
+++ b/api/api-compare-v2/src/main/java/com/thoughtworks/go/apiv2/compare/representers/material/GitMaterialRepresenter.java
@@ -26,5 +26,6 @@ public class GitMaterialRepresenter {
         jsonWriter.addWithDefaultIfBlank("branch", gitMaterialConfig.getBranch(), "master");
         jsonWriter.add("submodule_folder", gitMaterialConfig.getSubmoduleFolder());
         jsonWriter.add("shallow_clone", gitMaterialConfig.isShallowClone());
+        jsonWriter.addIfNotBlank("sparse_checkout", gitMaterialConfig.getSparseCheckout());
     }
 }
--- a/api/api-materials-v2/src/main/java/com/thoughtworks/go/apiv2/materials/representers/materials/GitMaterialRepresenter.java
+++ b/api/api-materials-v2/src/main/java/com/thoughtworks/go/apiv2/materials/representers/materials/GitMaterialRepresenter.java
@@ -28,6 +28,7 @@ public class GitMaterialRepresenter extends ScmMaterialRepresenter<GitMaterialCo
                     jsonWriter.addWithDefaultIfBlank("branch", gitMaterialConfig.getBranch(), "master");
                     jsonWriter.add("submodule_folder", gitMaterialConfig.getSubmoduleFolder());
                     jsonWriter.add("shallow_clone", gitMaterialConfig.isShallowClone());
+                    jsonWriter.addIfNotBlank("sparse_checkout", gitMaterialConfig.getSparseCheckout());
                 });
     }
 }
--- a/api/api-shared-v11/src/main/java/com/thoughtworks/go/apiv11/admin/shared/representers/materials/GitMaterialRepresenter.java
+++ b/api/api-shared-v11/src/main/java/com/thoughtworks/go/apiv11/admin/shared/representers/materials/GitMaterialRepresenter.java
@@ -27,6 +27,7 @@ public class GitMaterialRepresenter extends ScmMaterialRepresenter<GitMaterialCo
         jsonWriter.addWithDefaultIfBlank("branch", gitMaterialConfig.getBranch(), "master");
         jsonWriter.add("submodule_folder", gitMaterialConfig.getSubmoduleFolder());
         jsonWriter.add("shallow_clone", gitMaterialConfig.isShallowClone());
+        jsonWriter.addIfNotBlank("sparse_checkout", gitMaterialConfig.getSparseCheckout());
     }
 
     @Override
@@ -43,6 +44,7 @@ public class GitMaterialRepresenter extends ScmMaterialRepresenter<GitMaterialCo
         });
         jsonReader.readStringIfPresent("submodule_folder", gitMaterialConfig::setSubmoduleFolder);
         jsonReader.optBoolean("shallow_clone").ifPresent(gitMaterialConfig::setShallowClone);
+        jsonReader.readStringIfPresent("sparse_checkout", gitMaterialConfig::setSparseCheckout);
 
         return gitMaterialConfig;
     }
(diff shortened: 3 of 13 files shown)


chadlwilson:
I'm interested in adding support for this but probably does need some more work in adjacent repos like `ruby-functional-tests`. I need to document more of this stuff since implementing full features tends to require more than changes to this single repo.

Without doing something with the material fingerprint it would also lead to fan-in problems with repos that have different specs configured (similar to how the filters create similar problems) because they would be considered as the same material.

I had imagined trying to implement a consolidated feature that replaces the git-path-material plugin and support for partial checkouts (blob:none or tree:none etc) and avoid the fan-in problems the current approach tends to create.

palencharj:
Thanks — the fan-in point is a fair hit, and I'd got that trade the wrong way round. Pushed a fix.

**Fingerprint.** `sparseCheckout` now contributes to `appendCriteria` in both `GitMaterialConfig` and `GitMaterial`, so two materials on the same URL and branch with different specs are distinct. You're right that on a monorepo — the whole point of the feature — several pipelines deliberately share one repository, so that's the common case rather than an edge case, and resolving against a working copy that never contained a pipeline's paths is a much worse failure than losing history continuity.

One subtlety worth flagging, since it's the part that could have gone badly: the spec is contributed **only when configured**. `fingerprintFrom` joins every criteria entry as `key=value` with nulls included, so contributing it unconditionally would have re-fingerprinted every existing git material on upgrade and orphaned its history. Absent rather than null is what keeps it migration-safe, and there's a test pinning that (plus one asserting `GitMaterial` and `GitMaterialConfig` produce the same fingerprint, since the server computes from one and the agent from the other).

`equals`/`hashCode` are aligned for the same reason.

**Adjacent repos.** Happy to do the `ruby-functional-tests` work — I just don't know what "complete" looks like there. If you can point me at a comparable material-level feature that was done properly, I'll follow it. Same for docs.

**The consolidated feature.** That sounds like the better end state, and I'd rather help with it than have this sit as a partial overlap. Two things that may be useful input:

- Partial clone composes cleanly with sparse checkout — I've verified `--filter=blob:none` alongside `sparse-checkout set --no-cone`, and a server that doesn't support filtering just warns and sends everything, so it degrades safely.
- Non-cone mode matters if you want parity with a Perforce client view: cone mode can only select whole directories, while non-cone selects individual files.

I also published a standalone SCM plugin ([palencharj/gocd-git-sparse-material-plugin](https://github.com/palencharj/gocd-git-sparse-material-plugin)) so my team can use this on a stock server while the native question settles — mentioning it only because it overlaps the git-path-material plugin you referenced, so it's context for the consolidated design rather than a competing proposal.

So: happy to keep this narrowly scoped, extend it toward what you described, or close it in favour of your design — whichever is most useful. Just tell me which.

chadlwilson:
Thanks, but I’m not going to interact via prompting someone else’s Claude LLM via PR comments. That’s not why I contribute to open source and do free work for people & companies. If I wanted to do that I may as well just implement it myself with my own agent where I can prompt it properly about the way to implement it and things to worry about, test etc and I don’t have to figure out all the problems in the design from first principles.

palencharj:
Sorry I really am. :(

Issue #14475, opened 2026-07-02
Add a feature to delete pipeline runs


Ris345:
### Summary

While testing the GoCD pipeline, I noticed an excessive number of redundant runs. I wanted to remove them, but discovered that the feature was unavailable.

### Motivation / Use Case

Delete redundant pipeline runs 

### Proposal

Just a dd trash can button next to pipeline that deletes the pipeline run 

### Alternatives

leave the pipeline and waste reosurces 

### Any other info

should remove related ec2s as well but that might based on the context 

### Basic Details:
- *Go Server Version*: 
- *JVM Version*: 
- *OS Information*: 

### Additional Details _(if relevant)_:
- *Browser vendor/version*:
- *Relevant plugins in use*:

### Deployment mechanism _(e.g. static vs elastic, Docker, Kubernetes, RPM package etc)_
- *Go Server*:
- *Go Agent(s)*:

### Support Environment

```json

```

chadlwilson:
This will not be implemented as described given the resources available to the project. GoCD supports complex dependency networks of pipelines, and deleting a completed pipeline run could potentially break an indeterminate number of downstream pipelines in ways the user doing the deleting may not be aware of. All possible paths would need to be walked to determine if deleting a pipeline is "safe", and the group-based permissions model of GoCD would make it difficult to display to the user high fidelity information on why a pipeline run cannot be deleted.

Right now GoCD does not even have automated database cleanup of old pipeline/stage/job runs due to this challenge, and implementing that would be much more valuable.

What's the reason for wanting/needing this functionality? Why are there an "excessive number of redundant runs"? Being surprised that the capability does not exist does not really tell me why it'd be useful to be able to delete pipeline runs. If you explain why you want to remove them in terms of the UI or user experience, other solutions may be possible.

> should remove related ec2s

What does this mean? How are "ec2s" relevant here?

sriramnrn:
I would add:
Many of us use pipeline runs as an audit trail of actions. Having such a
button would require that the button be access controlled, and would lead
to lots of reviews and audit checks on the access control itself. This
would lead to audit headaches, uncertainty about the logs.

Most enterprise customers would vote against such a feature.

From a resources perspective, a pipeline record is a record in the database
as well as console logs on the server side. If you want, you can delete the
console logs after having the right approvals ( e.g. delete all logs older
than 6 months).

For your "EC2s", are these EC2 VM agents? What is the pain point? On my
[...]

PR #14545, opened 2026-08-05
Implement pipeline run delete feature with DAO cleanup and UI support

samikshaarora6:
Adds backend DAO cleanup, service and UI support.


chadlwilson:
I already explained at https://github.com/gocd/gocd/issues/14475#issuecomment-4895731875 why this would not be implemented, especially not this way. This is unsafe and would corrupt databases.

